PR #986, opened 2022-04-09
Fix wrong "open roles" link in README.md

Yikun:
The original link (https://www.warp.dev/hiring) is pointed to a page with `Page Not Found`, this patch fix it and link it to the right link (https://www.warp.dev/careers#open-roles).


elviskahoro:
Thanks @Yikun for catching this! Are you in our Discord server btw? So I can add the Contributor Discord role.

Yikun:
`yikun`, Thanks!

PR #1557, opened 2022-06-27
GH_Bug_temp_CI

ciyer64:
Added blocker question, updated question language.


elviskahoro (approved):

elviskahoro:
Going to merge and then check that it's working.

elviskahoro:
It didn't work because it mentioned that when we're having multi-selection the Yes and No options need to be wrapped in quotes. Just an FYI for the next one. @ciyer64

PR #1646, opened 2022-07-22
01_Template_repro

ciyer64:
Updating the Repro text from "value" to "placeholder" so that users can't file an empty response.


elviskahoro (approved):

elviskahoro:
@ciyer64 this broke the template
Screen Shot 2022-07-22 at 12 01 30
.

PR #1992, opened 2022-10-20
improvements to the new issues template

dannyneira:
Changing the order to focus more on docs/discord/issue search. Hoping this will deflect new issues for something that is known or can be resolved with self/community help.


dannyneira:
cancelling pr, working from dx repo submodule instead

PR #5481, opened 2024-11-15
PR template for ssh tmux issues

heysweet:
Add a new PR template for the upcoming ssh improvements


dannyneira:
@heysweet In my observation, folks hardly use the [current SSH template](https://github.com/warpdotdev/Warp/issues?q=is%3Aissue+is%3Aopen+sort%3Acreated-desc+label%3ASSH) and often use the Bug template for ssh problems. Instead of making another template just for labels (SSH / SSH-tmux), could we add a dropdown option to the current SSH template and assign the special labels based on a required dropdown choice?

AFAIK, GitHub issue forms cannot dynamically apply labels based on user input. But, this can be accomplished with a simple GitHub action. I can own implementing and testing this.

heysweet:
> could we add a dropdown option to the current SSH template and assign the special labels based on a required dropdown choice?

The idea with this template is there'll be a button in an error block in Warp that will say "Report to Warp" and clicking it should take them to this repo with an appropriate template. I think its hard to communicate old ssh subshell detection logic vs new ssh tmux logic, so separate templates made sense to me. Lastly, having 2 separate templates is a temporary state in my mind, where we would remove the old ssh template shortly after we launch the new feature.

That all being said, I'm supportive of however you want to move forward!

.github/ISSUE_TEMPLATE/04_ssh_tmux.yml
```diff
@@ -0,0 +1,110 @@
...
+      multiple: false
+      options:
+        - "Yes, I confirmed that this only happens in Warp, not other terminals."
+        - "No, this issue happens in Warp and other terminals."
+    validations:
+      required: true
+  - type: dropdown
+    id: "linear-label-ssh-tmux"
+    attributes:
+      label: "Warp Internal (ignore) - linear-label:7a739baa-09c3-499e-a0c9-a1a16c090597"
```
heysweet, line 96: Here I've added a new label to this template

dannyneira:
Oh i see, wasn't aware of this. Thanks for pointing that out. Since we will deprecate the older legacy ssh and you have separate questions on the new template, we should go forward with adding it as you originally suggested. No need to complicate things further when we deprecate the legacy template or have to change code in Warp for the error link. 

If I can recommend we swap the templates legacy to 04, and tmux to 03, so that when we delete the legacy one the tmux one will remain and we won't have to change any code in Warp to link to it.

heysweet:
> Oh i see, wasn't aware of this

Yeah my bad for the sparse PR description on this one. 

> If I can recommend we swap the templates legacy to 04, and tmux to 03

Great callout!

dannyneira (approved):
lgtm! Merged, added the label "SSH-tmux" and tested template

CleanShot 2024-11-15 at 11 30 25

PR #5823, opened 2025-02-11
added pre-submit checks and os options to templates

dannyneira:
Based on comments in slack thread: https://warpdev.slack.com/archives/C06MJ1612N4/p1739237604295219

Changes: 
- Added Windows OS option to templates
- Added KI / Warp Docs links as well so users can confirm they are not submitting a KI / existing feature


hongyi-chen (approved):
Overall this looks great, Danny!

I'm wondering if we want to hold off on merging this until a few days before launch day (2/26) so that we don't get users mistakingly selecting Windows as an option before launch. Perhaps maybe we can merge next Friday (2/21)? What do you think?

PR #5944, opened 2025-02-27
OS-label GitHub Action

dannyneira:
This action will add the OS-specific labels to a GitHub issue, based on the required OS dropdown selector in the body of the issue forms.


dannyneira:
Cancelling this, will work on manual OS tagging

PR #9238, opened 2026-04-28
[WAR-7378] Fix: Use fork point for PR diffs and commit message retrieval and remove parent branch logic

exzshao:
## Description
This change updates the Git dialogs to compare against the correct parent branch instead of the previous broken detect_parent_branch

Specifically, this PR:
- Introduces `detect_fork_point` to determine the point where the current branch diverged from its parent.
- Uses the fork point when no parent branch is specified, ensuring accurate diffs even when the parent branch is not explicitly set.

Fixes [WAR-7378](https://linear.app/warpdotdev/issue/WAR-7378/fix-publish-and-pr-on-diverged-parent-branch-bug)

## Testing
Tested manually by verifying that diffs and commit message generation now correctly use the fork point rather than main when working on stacked branches.

## Server API dependencies
N/A

## Agent Mode

### DIFF
--- a/app/src/code_review/git_dialog/commit.rs
+++ b/app/src/code_review/git_dialog/commit.rs
@@ -369,7 +369,6 @@ pub(super) fn start_confirm(me: &mut GitDialog, ctx: &mut ViewContext<GitDialog>
     let message_editor = state.message_editor.clone();
     let repo_path = me.repo_path().clone();
     let branch_name = me.branch_name().to_string();
-    let parent_branch = me.parent_branch_name.clone();
 
     me.set_loading(LOADING_LABEL, ctx);
 
@@ -407,7 +406,6 @@ pub(super) fn start_confirm(me: &mut GitDialog, ctx: &mut ViewContext<GitDialog>
                             create_pr_with_ai_content(
                                 &repo_path,
                                 &branch_name,
-                                parent_branch.as_deref(),
                                 ai.as_ref(),
                                 path_env_ref,
                             )
@@ -421,7 +419,6 @@ pub(super) fn start_confirm(me: &mut GitDialog, ctx: &mut ViewContext<GitDialog>
                                 &repo_path,
                                 None,
                                 None,
-                                parent_branch.as_deref(),
                                 path_env_ref,
                             )
                             .await?
--- a/app/src/code_review/git_dialog/commit.rs
+++ b/app/src/code_review/git_dialog/commit.rs
@@ -31,7 +31,10 @@ use crate::{
     },
     server::server_api::ServerApiProvider,
     ui_components::icons::Icon,
-    util::git::{FileChangeEntry, PrInfo},
+    util::git::{
+        create_pr, get_diff_for_commit_message, get_file_change_entries, run_commit, run_push,
+        FileChangeEntry, PrInfo,
+    },
     view_components::action_button::{ActionButton, ButtonSize, SecondaryTheme},
 };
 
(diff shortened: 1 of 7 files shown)


app/src/util/git.rs
```diff
@@ -193,15 +193,73 @@ pub async fn detect_main_branch(repo_path: &Path) -> Result<String> {
...
+/// `<fork>..HEAD` for "commits unique to this branch". For branch-name
+/// lookup, see [`detect_parent_branch`].
+#[cfg(not(feature = "local_fs"))]
+pub async fn detect_fork_point(_repo_path: &Path) -> Result<Option<String>> {
+    Err(anyhow!("Not supported without local_fs"))
+}
+
+/// See the no-`local_fs` stub above for documentation.
+#[cfg(feature = "local_fs")]
+pub async fn detect_fork_point(repo_path: &Path) -> Result<Option<String>> {
```
MaggieShan: Is there a way we can reduce the amount of times this is called? Maybe computed + stored alongside the parent branch?

app/src/util/git.rs
```diff
@@ -193,15 +193,73 @@ pub async fn detect_main_branch(repo_path: &Path) -> Result<String> {
...
+    // Exclude `<current>`, `origin/<current>`, and the resolved `@{u}` so
+    // the branch isn't subtracted from itself.
+    let current = detect_current_branch(repo_path).await.ok();
```
MaggieShan: Do we not already have current and upstream by the time this is called?

app/src/util/git.rs
exzshao: I think it's fine, cache seems complicated cuz a lot of things can invalidate the state and then we need to know when to re-compute. Also this is only computed for branches with no upstream.

app/src/code_review/git_dialog/commit.rs
```diff
@@ -417,14 +415,8 @@ pub(super) fn start_confirm(me: &mut GitDialog, ctx: &mut ViewContext<GitDialog>
...
                             // `gh pr create --fill`
-                            crate::util::git::create_pr(
-                                &repo_path,
-                                None,
-                                None,
-                                parent_branch.as_deref(),
-                                path_env_ref,
-                            )
-                            .await?
+                            crate::util::git::create_pr(&repo_path, None, None, path_env_ref)
```
MaggieShan: nit: can we move these imports to the top

app/src/util/git.rs
```diff
@@ -193,135 +193,65 @@ pub async fn detect_main_branch(repo_path: &Path) -> Result<String> {
...
-        if !name.contains('/') && name == "origin" {
-            continue;
-        }
-        if current == Some(name) {
-            continue;
-        }
-        if upstream == Some(name) {
-            continue;
+    current_branch_name: Option<&str>,
+    upstream_ref: Option<&str>,
```
MaggieShan: Seems like this is only called when `upstream_ref` is `None` - can we remove this param?

app/src/code_review/git_dialog/pr.rs
```diff
@@ -211,11 +204,11 @@ pub(super) async fn create_pr_with_ai_content(
             log::warn!(
                 "AI PR content generation returned empty title/body, falling back to --fill"
             );
-            crate::util::git::create_pr(repo_path, None, None, parent_branch, path_env).await
+            crate::util::git::create_pr(repo_path, None, None, path_env).await
```
MaggieShan: nit: can we move these imports to the top

specs/APP-4218/TECH.md
```diff
@@ -1,76 +1,97 @@
 # APP-4218: Git operations dialogs compare against the branch's actual parent — Tech Spec
 Product spec: `specs/APP-4218/PRODUCT.md`
```
MaggieShan, line 2: nit: can we update the tech spec, it's out of sync with the code

MaggieShan (approved):

PR #9239, opened 2026-04-28
[APP-4249] Set default cwd for project-scoped MCP servers

vkodithala:
### Motivation

File-based stdio MCP servers spawned from a project's `.mcp.json` (e.g. `pnpm run mcp:foo`, or anything using `./tooling_scripts/...`) fail with:

```
Transport creation error: No such file or directory (os error 2)
```

Root cause: when the user's `.mcp.json` doesn't include a `working_directory`, Warp's spawner inherits whatever cwd Warp was launched from rather than the directory the config was discovered in. Two failure modes share this same root cause:

- **Repo-relative commands/args** (e.g. `./tooling_scripts/foo`, `node ./src/server.js`) can't be resolved from outside the repo, so `execvp` returns `ENOENT`.
- **Workspace-aware launchers** like `pnpm`/`npm`/`yarn` walk up from cwd looking for `package.json` (and workspace manifests). When cwd isn't inside the repo, they bail out before launching the requested script — surfacing as `ENOENT` from the spawner's perspective.

`working_directory` is a Warp/VS-Code-style extension; the canonical Anthropic/Cursor MCP schemas don't include it, so most `.mcp.json` files in the wild don't set it.

### Implementation

- New helper `FileBasedMCPManager::spawn_root_for_installation(uuid)` returning the discovery root for any file-based install: the repo root for project-scoped configs, the home directory for global configs (Warp and third-party). Returns `None` for non-file-based installs (e.g. cloud-templated ones), leaving them unaffected. Global Warp installs are remapped from `~/.warp/` (Warp internal state) to `~` so all global installs share a consistent cwd.
- In `TemplatableMCPServerManager::spawn_server_impl`, default `cli_server.cwd_parameter` from this helper when unset. Single funnel — covers both auto-spawn and the manual UI opt-in path. User-supplied `working_directory` always wins.
- ENOENT-aware logging at the spawn site: when `TokioChildProcess::spawn()` fails with `NotFound`, the MCP log file now spells out the server name, the missing executable, the cwd we used, and a hint pointing the user at `working_directory`. The user-surfaced error string is unchanged.
- Documented the new defaulting behavior and `working_directory` override in `resources/bundled/skills/add-mcp-server/SKILL.md`.

## Testing

Verified spawn-site behavior manually. Demo [here](https://www.loom.com/share/ec01d98ae9114433b8ae87f5f17adfd4)!

- Added `test_parse_cli_server_preserves_explicit_working_directory` in `mod_test.rs` to lock in that an explicitly-set `working_directory` round-trips through parsing and won't be clobbered by the new defaulting logic.
- Skipped a unit test for the new helper itself — it's a thin lookup over `file_based_servers_by_root` and the sorted-pick policy is straightforward enough to validate manually.
- Spawn-site behavior (cwd actually applied, ENOENT log emission) verified manually against a `pnpm`\-based repro.

## Agent Mode

CHANGELOG-BUG-FIX: Project-scoped file-based MCP servers now spawn from the repo root by default (and global ones from `~`), so configs with relative commands/args (and workspace launchers like `pnpm`/`npm`) work without an explicit `working_directory`.

### DIFF
--- a/app/src/ai/mcp/templatable_manager/native.rs
+++ b/app/src/ai/mcp/templatable_manager/native.rs
@@ -782,6 +782,20 @@ impl TemplatableMCPServerManager {
                     value: execution_path,
                 },
             );
+
+            // For file-based MCP installations without an explicit `working_directory`,
+            // default the spawn cwd to the directory the config was discovered in
+            // (repo root for project-scoped configs, ~/.warp/ or ~ for globals). This
+            // matches user expectations for repo-relative commands in `.mcp.json`.
+            // Cloud-templated installations (lookup returns None) are unaffected and
+            // continue to inherit Warp's process cwd.
+            if cli_server.cwd_parameter.is_none() {
+                if let Some(spawn_root) =
+                    FileBasedMCPManager::as_ref(ctx).spawn_root_for_installation(installation_uuid)
+                {
+                    cli_server.cwd_parameter = Some(spawn_root.to_string_lossy().into_owned());
+                }
+            }
         }
 
         let executor = ctx.background_executor().clone();
@@ -1742,6 +1756,11 @@ async fn spawn_server(
                 }
             }
 
+            // Capture the command and configured cwd for diagnostics before they're
+            // moved into the Command builder closure.
+            let command_for_log = command.clone();
+            let cwd_for_log = cli_server.cwd_parameter.clone();
+
             // Try to spawn the child process.
             let (transport, stderr) = rmcp::transport::TokioChildProcess::builder(
                 tokio::process::Command::new(command).configure(|cmd| {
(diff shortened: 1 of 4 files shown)


app/src/ai/mcp/templatable_manager/native.rs
```diff
@@ -1768,6 +1787,17 @@ async fn spawn_server(
...
+                    let cwd_display = cwd_for_log
+                        .as_deref()
+                        .unwrap_or("<inherited from Warp's process cwd>");
+                    logger.log(format!(
+                        "[error] MCP: Failed to spawn '{server_name}': command '{command_for_log}' \
+                         not found (cwd: {cwd_display}). If your MCP server depends on a specific \
+                         working directory, set the `working_directory` field in your config to \
+                         override the default."
+                    ));
+                }
```
vkodithala, line 1800: This isn't 100p necessary, but thought it might be nice to embed some richer context around ENOENT failures to help AM debug a little better (the enterprise user whose issue this PR solves tried to debug with Oz).

After this change, most "no file/directory found" issues can be fixed by setting an explicit `working_directory` within server configurations.

resources/bundled/skills/add-mcp-server/SKILL.md
```diff
@@ -61,6 +61,24 @@ Check whether the target config file exists.
...
+  "mcpServers": {
+    "server-name": {
+      "command": "node",
+      "args": ["./tooling/mcp/server.js"],
+      "working_directory": "/absolute/path/to/repo"
+    }
+  }
+}
+```
+
```
vkodithala, line 81: Also not essential, but thought it made sense to update the bundled skill so that the agent's aware of the "working_directory" param.

peicodes (approved):

PR #9244, opened 2026-04-28
fix(context_chips): stop GitDiffStats flicker from shell fallback

### LINKED ISSUE #9228: Git diff chip flickers between tracked-only and all-files count when untracked files are present
## Summary
When running a terminal command in a repo that has both modified tracked files and untracked files, the git diff chip flickers between two values: the correct one that counts all changes (2 files | 8 edits) and an incorrect one that omits untracked files (1 file | 4 edits). The flicker is consistent and reproducible on every command run.

## Problem
The `GitDiffStats` chip has two update paths that compute the file/edit counts differently:

1. **`GitRepoStatusModel` (filesystem watcher)**: calls `DiffStateModel::diff_metadata_against_head()`, which runs `git status --untracked-files=all` and reads line counts from untracked files directly. **Includes untracked files.**
2. **`shell_git_line_changes()` (shell fallback)**: runs `GIT_OPTIONAL_LOCKS=0 git -c diff.autoRefreshIndex=false diff --shortstat HEAD`. **Excludes untracked files.**

When a command completes, the shell-command fallback fires and briefly overwrites the filesystem-watcher value with a count that omits untracked files. The filesystem watcher then fires again and restores the correct value, producing the visible flicker.

## Expected behavior
The chip consistently shows the correct count that includes both modified tracked files and untracked files (2 | 8 in this reproduction case).

## Actual behavior
Every time a command is run, the chip briefly flickers to the incorrect value (1 | 4), which excludes untracked files, before snapping back to the correct value (2 | 8).

## Reproduction steps
1. Open Warp in a local git repository.
2. Create at least one **modified tracked file** with a few edits (e.g., add 4 lines to any tracked file).
3. Create at least one **untracked file** in the same repo (e.g., `touch scripts/repro.sh` and add 4 lines).
4. Observe the git diff chip in the prompt — it should show both files/edits (e.g., `2 | 8`).
5. Run any terminal command (e.g., `echo hello`, `ls`, `git status`).
6. Watch the git diff chip: it briefly shows the lower, tracked-only value (e.g., `1 | 4`) then corrects to `2 | 8`.
7. The flicker repeats every time a command is run.

**Specific repo state used during reproduction:**
- Modified tracked file: `.mcp.json` (+4 lines, 0 deletions)
- Untracked file: `scripts/repro-mcp.sh` (4 lines)
- Correct chip value: `2 | 8`
- Incorrect (flickered) chip value: `1 | 4`

## Artifacts
Video with a repro: https://www.loom.com/share/b0f148fe1c004eb591ec82e4149fcace.

## Warp version
Unknown (please fill in from Warp → About Warp)

## Operating system
macOS

## Possible source references
- `app/src/context_chips/builtins.rs (162-180)`: `shell_git_line_changes()` runs `git diff --shortstat HEAD`, which only counts tracked file changes and excludes untracked files — this is the source of the incorrect lower value.
- `app/src/code_review/diff_state.rs (1624-1665)`: `diff_metadata_against_head()` uses `git status --untracked-files=all` and reads line counts from untracked files — this produces the correct higher value.
- `app/src/context_chips/current_prompt.rs (1428-1472)`: filesystem-watcher update path that calls `GitLineChanges::from_diff_stats(&metadata.stats_against_head)` — correct path, but races with the shell command path on block completion.
- `app/src/context_chips/current_prompt.rs (1480-1492)`: `is_updated_externally()` guard that should suppress the periodic 30-second timer when `GitRepoStatusModel` is live — but does not appear to suppress the block-completion shell command trigger.

zerone0x:
## Summary

Fixes #9228.

When a per-repo `GitRepoStatusModel` is attached to `CurrentPrompt`, the `GitDiffStats` chip is updated via filesystem-watcher events whose `DiffStateModel::diff_metadata_against_head()` count includes untracked files (it walks `git status --untracked-files=all`).

However, every time the prompt context is rebuilt (e.g. when a new block's metadata arrives after a command completes), `run_chips` was still firing the chip's periodic shell-based generator once. For `GitDiffStats` that fallback is `shell_git_line_changes()`, which runs `git diff --shortstat HEAD` and counts only tracked changes. Its result was overwriting the watcher's structured `GitLineChanges` value until the next watcher `MetadataChanged` event restored it — causing the chip to visibly flicker between the tracked-only count and the all-files count whenever untracked files were present.

## Fix

In `CurrentPrompt::run_chips`, when a chip is `is_updated_externally`, only run an `initial_value_generator` if one is provided. The shell-based `chip.generator()` is no longer used as a fallback for externally-driven chips:

- `ShellGitBranch` continues to seed its initial value from `current_environment.git_branch()` via `initial_value_generator`, then receives updates from the watcher.
- `GitDiffStats` has no `initial_value_generator`, so the periodic shell command is now skipped entirely. Initial population is handled by the watcher: `GitRepoStatusModel::new()` kicks off `refresh_metadata` immediately and emits `MetadataChanged` once metadata is computed, which `set_git_repo_status` handles by updating `GitDiffStats` with the structured `GitLineChanges`. Existing chip values survive across `clear_chips()` (which only aborts in-flight generators), so there is no transient empty state on subsequent prompt refreshes.

This keeps the diff-count semantics consistent (always all-files) without changing the behavior of remote/non-watched sessions, which still fall back to the shell generator via the periodic timer branch.

## Test plan

- [ ] In a repo with both modified tracked files and untracked files, run a command (e.g. `ls`) and observe the `GitDiffStats` chip — it should stay on the all-files count without flickering down to the tracked-only count.
- [ ] In a repo without the watcher available (e.g. remote session), confirm the chip still updates periodically via the shell fallback (unchanged code path).
- [ ] `ShellGitBranch` continues to update on branch changes via the watcher and shows the correct branch on first paint.

CHANGELOG-BUG-FIX: Fix git diff chip flickering between tracked-only and all-files count when untracked files are present

### DIFF
--- a/app/src/context_chips/current_prompt.rs
+++ b/app/src/context_chips/current_prompt.rs
@@ -1041,15 +1041,31 @@ impl CurrentPrompt {
                 }
                 RefreshConfig::Periodically { .. } => {
                     if self.is_updated_externally(chip_kind) {
-                        let initial_gen = chip_kind.initial_value_generator();
-                        let generator = initial_gen.as_ref().unwrap_or(chip.generator());
-                        self.fetch_chip_value_once(
-                            chip_kind,
-                            generator,
-                            chip.on_click_generator().cloned(),
-                            true,
-                            ctx,
-                        );
+                        // For chips updated externally (e.g. by the per-repo
+                        // git status filesystem watcher), avoid running the
+                        // periodic shell-based generator. Doing so can briefly
+                        // overwrite the structured watcher value with one that
+                        // uses different semantics (for example, the
+                        // `GitDiffStats` shell fallback runs `git diff
+                        // --shortstat HEAD`, which excludes untracked files,
+                        // whereas the watcher counts untracked files as
+                        // changes), causing the chip to flicker between the
+                        // tracked-only count and the all-files count when
+                        // untracked files are present.
+                        //
+                        // If a chip provides an `initial_value_generator` that
+                        // sources from the prompt context (rather than running
+                        // a shell command), use it for a fast initial value
+                        // until the watcher emits a metadata-changed event.
+                        if let Some(initial_gen) = chip_kind.initial_value_generator() {
+                            self.fetch_chip_value_once(
+                                chip_kind,
+                                &initial_gen,
+                                chip.on_click_generator().cloned(),
+                                true,
+                                ctx,
+                            );
+                        }
                     } else {
                         self.fetch_chip_value_at_interval(
                             chip_kind,


vkodithala (approved):
Was looking into this earlier - this is the right fix! Verified that this all works locally. Make sure that you sign our CLA and you're good to go!

captainsafia:
@zerone0x Thanks for opening this PR! When you get a chance, please sign the CLA so we can move forward with merging this PR.

zerone0x:
signed!

Issue #9138, opened 2026-04-21
Rich terminal block types: images, tables, and JSON


captainsafia:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Warp's Blocks are already a huge improvement over a raw scrollback buffer, but today a block's output is essentially a rectangle of terminal cells. I'd like Warp to recognize common *structured* output and render it as a first-class **rich block type**, with an affordance to toggle back to the raw text view.

Proposed block types:

1. **Image block**
   - Detect inline image escape sequences (iTerm2 `OSC 1337 ; File=...` and Kitty `APC G ...` graphics protocols) and render them as a proper image view inside the block, with correct sizing, selection-safe behavior, and copy/save-as actions.
   - Also detect when a command's stdout is obviously a path to an image file (`file.png`, `file.jpg`, `file.svg`) and surface an inline preview.

2. **Table block**
   - When stdout looks like CSV/TSV, a markdown table, or an ASCII box-drawing table (e.g. `psql`, `mysql`, `kubectl get ...`, `gh pr list`), render it as an interactive table: resizable columns, sortable headers, column-level find, copy-as-CSV/markdown.
   - Users should be able to trigger this explicitly via a per-block "View as table" action when auto-detection fails.

3. **JSON / structured-data block**
   - Detect JSON (and optionally YAML/TOML) in a block's output and render it as a collapsible tree with syntax highlighting, search, and "copy path to value" on right-click.
   - Support streaming JSON (pretty-print when the block completes, fall back to raw while running).

Cross-cutting behavior:

- Each rich view should have a clearly labeled toggle back to the **raw text** view — never hide the original bytes.
- Copying from a rich view should copy the underlying raw text by default (so scripts that pipe to `pbcopy` / grep continue to work).
- Rich rendering should be **opt-out**: a Settings toggle ("Render rich output in blocks") plus per-command overrides (e.g. a `WARP_RICH_OUTPUT=0` env var or a marker the user can add to a command).
- The block's serialized form (for Warp Drive sharing, AI context, and session restoration) should preserve the raw text so nothing is lost cross-client.
[...]

PR #9262, opened 2026-04-28
specs(GH9138): JSON/YAML structured-data block viewer — product + tech spec

lucianfialho:
## Description

Spec PR for issue #9138 (labeled \`ready-to-spec\`), scoped to the **JSON/YAML structured-data block** portion of the rich block types feature.

This PR adds:
- \`specs/GH9138/product.md\` — 27 testable behavior invariants covering detection, tree rendering, raw toggle, copy affordances, settings opt-out, and edge cases.
- \`specs/GH9138/tech.md\` — implementation plan grounded in the existing \`RichContent\` / blockgrid architecture, with file and line references throughout.

### Scope

This spec intentionally covers **only JSON and YAML** block rendering. Image blocks and table blocks (also described in #9138) are left as follow-ups — they are architecturally independent and can be specced and implemented separately.

### Key design decisions

- Detection runs in a \`Background\` task after block completion, with a 50 ms timeout and 5 MB cap, so it never blocks the UI thread.
- The tree view is inserted as a \`RichContent\` item (same mechanism as AI blocks and warpify blocks), keeping the raw grid intact for toggling.
- \`WARP_RICH_OUTPUT=0\` env var and a global Settings toggle provide opt-out at two granularities.
- Raw text is always the canonical serialized form (Warp Drive, share links, AI context).

## Testing

No code changes in this PR — spec only. Validation approach is documented in \`specs/GH9138/tech.md\` under "Testing and validation".

## Agent Mode

### DIFF
--- a/specs/GH9138/product.md
+++ b/specs/GH9138/product.md
@@ -72,12 +72,12 @@ Out of scope for this spec (tracked separately in #9138):
     of nesting. These limits prevent pathological inputs (e.g., a deeply nested
     1 MB JSON) from producing a tree that is impractical to render or navigate.
 
-5. If `WARP_RICH_OUTPUT=0` is present in the command's environment at the time
-   the block completes, detection is skipped for that block regardless of
-   settings. The mechanism for reading this value at block-completion time is
-   an implementation detail; if the current block model does not expose
-   per-command environment snapshots, detection falls back to reading
-   `WARP_RICH_OUTPUT` from the process environment at detection time instead.
+5. If `WARP_RICH_OUTPUT=0` is set in the **process environment** (e.g., exported
+   in the shell profile or set before launching Warp), detection is skipped for
+   all blocks in that session regardless of settings. This is a process-level
+   escape hatch, not a per-command override; inline usage such as
+   `WARP_RICH_OUTPUT=0 my-command` is not guaranteed to suppress detection for
+   that single command and is out of scope for this spec.
 
 6. Detection is not retried after failing. If the output is not valid JSON or
    YAML, the block remains raw text.
--- a/specs/GH9138/product.md
+++ b/specs/GH9138/product.md
@@ -1,0 +1,167 @@
+# PRODUCT.md — JSON / YAML structured-data block viewer
+
+Issue: https://github.com/warpdotdev/warp/issues/9138
+Figma: none provided
+
+## Summary
+
+When a terminal block's output is valid JSON or YAML, Warp renders it as an
+interactive collapsible tree instead of raw text. A toggle button on the block
+lets the user switch between the tree view and the original raw output at any
+time. The raw bytes are never modified; the tree is a read-only rendering lens.
+
+## Goals / Non-goals
+
+In scope:
+- JSON detection and tree rendering for completed blocks.
+- YAML detection and tree rendering for completed blocks.
+- Toggle between tree view and raw text per block.
+- Collapse / expand of individual nodes in the tree.
+- Keyboard navigation inside the tree.
+- "Copy value" and "Copy path" affordances on tree nodes.
+- A Settings toggle to opt out of automatic structured-data rendering.
+- A `WARP_RICH_OUTPUT=0` env-var escape hatch to suppress rendering for a
+  specific command.
+
+Out of scope for this spec (tracked separately in #9138):
+- Image block rendering.
+- Table block rendering (CSV/TSV/ASCII-box).
+- TOML rendering.
+- Editable / writable tree view.
+- Streaming JSON rendering while the command is still running (the tree renders
+  only after the block completes).
+
+## Behavior
+
+### Detection
+
+1. After a block completes (exit code is received), Warp strips ANSI escape
+   sequences from the block's output text and attempts to parse it first as JSON
+   (`serde_json`) and then, if JSON fails, as YAML (`serde_yaml`). Detection
+   runs only when the "Render rich output in blocks" setting is enabled (on by
+   default) and `WARP_RICH_OUTPUT` is not set to `0` in the command's
+   environment.
+
+2. A block is treated as JSON if the stripped output, after trimming leading and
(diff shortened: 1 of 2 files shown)


lucianfialho:
recheck

lucianfialho:
Thanks for the thorough review! Addressed all 4 important points in the latest commit:

1. **Canonical text source**: clarified in both PRODUCT.md (Summary) and TECH.md (§3) that "raw text" = `BlockGrid::contents_to_string` output — the ANSI-stripped, PTY-processed string used consistently for Copy, Share, AI context, and detection input.

2. **Immediate reversion mechanism**: TECH.md §2 now describes the concrete mechanism — `TerminalView` subscribes to settings change events; toggling off clears `block_structured_view_active`, hiding tree views on next paint; parsed values are retained so re-enabling is instant without re-parsing.

3. **WarpUI View interface**: removed the non-existent `View<TerminalView>` reference; replaced with `warpui::View` and pointed to `plugin_instructions_block.rs` as the concrete pattern to follow.

4. **Parser CPU boundedness**: TECH.md §3 now explicitly states the 5 MB `MAX_DETECT_BYTES` cap is the real guarantee (input rejected before any parser is called), that `tokio::time::timeout` would not stop a synchronous parse, and what to do if worst-case benchmarks exceed budget (lower the cap, not add cancellation).

5. **Copy-path escaping**: PRODUCT.md §17 now defines deterministic rules — simple keys use dot notation, array indices use `[n]`, and keys with dots/spaces/quotes use `["key"]` with internal `\"` escaping. Examples included.

lucianfialho:
@oss-maintainers The /oz-review triggered yesterday but the second review never posted. Could you help retrigger or check what happened?

lucianfialho:
Addressed all 5 points from the second review:

1. **50ms contradiction resolved**: removed "bounded at 50ms" from the product spec entirely. CPU boundedness is now documented as: `MAX_DETECT_BYTES` check before parse + `MAX_NODES` (10k) / `MAX_DEPTH` (50) post-parse walk. No timeout wrapper claimed.

2. **Restore/share scoped to non-goals**: product spec §15 and §25 now say restored and shared blocks always render as raw text on first load. Re-detection at restore/share time is explicitly a non-goal and follow-up.

3. **Settings reversion state machine clarified**: re-enabling the setting repopulates `block_structured_view_active` from the existing `StructuredDataBlock` `RichContent` items already in the blocklist — no re-parsing needed.

4. **`WARP_RICH_OUTPUT` mechanism made honest**: tech spec §6 now names the per-block env snapshot as primary path and process-env as explicit named fallback, acknowledges the fallback limitation (doesn't catch per-command overrides), and defers a proper fix to follow-ups.

5. **Depth/node limits added**: new Behavior #4a in product spec (10k nodes, 50 levels); reflected in tech spec §3 with `MAX_NODES` and `MAX_DEPTH` constants and corresponding unit tests.

/oz-review

lucianfialho:
Addressed all 5 points from the third review:

1. **`WARP_RICH_OUTPUT` scoped correctly**: product spec §5 and tech spec §6 now explicitly state this is a process-level env var read via `std::env::var`. Per-command inline override (`WARP_RICH_OUTPUT=0 cmd`) is out of scope and documented as such.

2. **`source_block_id` in RichContentMetadata**: tech spec §4 now adds `source_block_id: Option<BlockId>` to `RichContentMetadata`. This is set on insertion and used by the toggle handler (§7) and settings-reversion handler (§2) to enumerate and manage structured-data views by their source block.

3. **Grid suppression explained**: tech spec §6 adds `hidden_grid_blocks: HashSet<BlockId>` to `TerminalView`. The grid-rendering path in `block_list_element.rs` renders a zero-height container when the block's `BlockId` is in this set. Toggle (§7) adds/removes from both `block_structured_view_active` and `hidden_grid_blocks` together, so grid and tree are mutually exclusive.

4. **Agent Mode hook**: tech spec §6 now specifies detection at both `on_user_block_completed` and the agent block completion site; identical logic and limits apply.

5. **YAML pre-parse security**: added `MAX_YAML_BYTES = 1 MB` checked before calling `serde_yaml::from_str`, documented as the primary mitigation for anchor/alias amplification since `serde_yaml` 0.8 has no built-in alias depth limit. `MAX_NODES` post-parse walk is the secondary check.

lucianfialho:
@oss-maintainers All three /oz-review rounds are exhausted. We addressed every concern raised across the three reviews (canonical text source, limits, grid suppression, agent hook, YAML security, metadata association, env-var scoping). Could a maintainer take a look and decide if the specs are ready to proceed to implementation?

lucianfialho:
Addressed all 5 points from the latest review:

1. **`WARP_RICH_OUTPUT` aligned**: both specs now consistently state process-level only; tech spec caches `std::env::var` at `TerminalView` init.

2. **YAML 1 MB cap in product spec**: Behavior #4 now lists JSON 5 MB cap, YAML 1 MB cap, and alias rejection as separate pre-parse checks — all in the product spec, not just the tech spec.

3. **YAML alias pre-parse**: byte-scan for `*[A-Za-z_]` before any parser call. Inputs with aliases return `None` immediately — amplification is impossible because the parser is never called on such inputs. The 1 MB byte cap is a secondary defense for valid YAML without aliases.

4. **RichContentMetadata concrete**: new `StructuredDataBlock { source_block_id: BlockId, view_handle: ViewHandle<StructuredDataBlockView> }` variant, matching the pattern of existing variants. Height changes on expand/collapse call `mark_rich_content_dirty(view_handle.id())`, which feeds into `BlockList::update_block_height_indices` — same mechanism AI blocks use.

5. **Secret redaction explicit**: new Behavior #19a in product spec. Detection input is `contents_to_string` default (obfuscated). Tree values, "Copy value", "Copy path", and the block Copy button all operate on the obfuscated text. Unobfuscated secrets are never exposed through any tree affordance.

Issue #4339, opened 2024-02-26
Make Warp work with Local Language Models (like Ollama models)


hmdz105:
### Describe the solution you'd like?

Due to safety concerns, many users are getting paranoid about Warp's forced login and online AI assistance. Since terminals are used to access critical data on local machines and servers, adding the ability to use local language models like Llama 2 using Ollama instead of online AI engines would be a great plus to Warp terminal emulator. 

### Is your feature request related to a problem? Please describe.

Mostly safety concerns when using Warp for accessing password protected systems and documents. 

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

Absolucy:
Some local LLM tools even expose an OpenAI API-compatible server, which I believe could allow for rudimentary support with just a setting to change the OpenAI endpoint?

dannyneira:
Hey folks, since we already have a request tracking this over at https://github.com/warpdotdev/Warp/issues/3779

If you could please give it a 👍 and subscribe to that request, it would help us gauge interest. We'll post any updates there, so stay tuned! 

We'll be closing this as a duplicate.

dannyneira:
Duplicate of https://github.com/warpdotdev/Warp/issues/3779

kn0wmad:
This is the feature that would bring me to use Warp

Ryj3l:
As a warp user and system's administrator, this feature alone would make getting buy in from our CTO 100x easier. I love using Warp in my home lab / personal environments but no way in its current iteration could I even get it in my professional environment to demo a proof of value. Having the option to show case the AI integrations with local models would allow us to begin the discussion on an enterprise plan with zero data retention.

kentlangley:
Yes, please. Ollama is easiest. LMStudio great too. Groq (not the X model) would be rather interesting. Let people use the LLM they want to use. Warp team, this INCREASES the value of this tool greatly! Your value is NOT in being a niche gateway to popular LLMs.

adityagaddhyanphonepe:
I think this is one of the important features missing in warp. 2 reasons - Most of the big companies dont let access gpt or any other LLM which captures data. Also, if they allow, there is only so many request that can be made for free.

beamop:
Would be game changer.

mbalasko:
vscode with LmStudio is amazing, WARP could integrate very easily!

lulunac27a:
[...]

Issue #7326, opened 2025-09-05
support for Agent Client Protocol (ACP)


restot:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

[Agent Client Protocol (ACP)](https://agentclientprotocol.com/overview/introduction) its new trend in Ai development environment, and I would like to use my WARP subscription to access agents from ZED IDE [External Agents](https://zed.dev/docs/ai/external-agents) feature. 
[Gemini CLI](https://zed.dev/docs/ai/external-agents#gemini-cli) and [Claude Code](https://zed.dev/docs/ai/external-agents#claude-code) already have support for ACP and I can't wait to use WARP with ACP. I could hack-my-way trough but its would be nice to have native support such as MCP support that we already have. 

Thank you for your hard work and great product. Please let us know if you would plan to add ACP. 

### Is your feature request related to a problem? Please describe.

Also would be nice for WARP have ACP support as well. Gemini-cli for instance, I'm using it as well from WARP but its kinda hacky to make it work together now.  

### Operating system (OS)

macOS

### How important is this feature to you?

5 (Can't work without it!)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

igorgue:
Would love this one to use warp from Codecompanion in Neovim.

shrimpwtf:
+1

gigachadrustacean:
any news lol feels like warp team is not prioritizing this anyway can submit PR?

codeitlikemiley:
if there would be PR i doubt they would even add it, as they would lose revenue with it...  or it would be bundled on pro like BYOK lol which is kinda stupid,  feels like you are begging for them to add this simple stuff.  unfortunately  warp isnt open source , if it is , i could have already made a PR on this one ...

warp auto complete is okay ... them having to lock the BYOK and possibly this ACP integration on pro sucks

RonAmihai:
ACP client support for Claude Code (or prime Claude Code SDK support) would be great.

nsxdavid:
This would be super useful!

For clarification, Claude Code does not support ACP but rather uses an adaptor to make it work.

nigelpurvis:
[...]

PR #9271, opened 2026-04-28
[codex] Spec local Ollama model support

manimohans:
## Summary
- Adds product and tech specs for issue #4339 local/Ollama model support.
- Defines the data-boundary requirement: local agent prompts, terminal context, tool payloads, and transcripts do not go through Warp's model-generation service.
- Proposes a client-side generation router, local model settings, Ollama/OpenAI-compatible model discovery, and a minimal local agent runner.

## Validation
- Docs-only spec change; no code tests run.

## Notes
- Opening as draft because #4339 is triaged and under maintainer feedback, but does not currently have a ready-to-spec label.

### DIFF
--- a/specs/GH4339/product.md
+++ b/specs/GH4339/product.md
@@ -1,0 +1,103 @@
+# Local Ollama Model Support - Product Spec
+GitHub issue: https://github.com/warpdotdev/warp/issues/4339
+Figma: none provided
+
+## Summary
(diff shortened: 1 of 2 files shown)


zachlloyd:
Thanks for the spec here! Our approach to local models is going to be to support ACP (see #7326) and then support compliant harnesses like OpenCode for accessing the local models.  The other option we've considered is shipping a simple local harness to do this (our Oz harness is server-side), but the plan we'd like to try first is ACP.

knotbin:
A local harness would be amazing!

PR #9279, opened 2026-04-28
Update tab CWD and git branch from OSC 7 escape sequences

### LINKED ISSUE #9125: Vertical tabs don't update git branch when CWD changes via external tool

Warp should update its tab CWD/git-branch display whenever the shell's working directory changes — not only during the prompt rendering cycle. Ideally, Warp should respond to standard terminal escape sequences like OSC 7 (`\e]7;file://hostname/path\a`) for CWD updates, so tools can notify Warp of directory changes without waiting for the next prompt.

I use `wt switch <branch>` (from [worktrunk](https://github.com/jondot/worktrunk)) to jump between git worktrees. Each worktree is on a different branch in a different directory. The `wt` shell wrapper does a proper `cd` to the new worktree path.

When I run `wt switch` alone, the tab updates correctly — because the command finishes, fish redraws the prompt, and Warp picks up the new CWD during its prompt cycle.

But my workflow wraps `wt switch` in a function that immediately launches a long-running process (e.g. `claude`, `codex`) after switching. Since the prompt never redraws between the `cd` and the process launch, Warp's tab stays stuck on the old directory/branch for the entire session.

### What I've tried (none worked)

| Approach | Result |
|----------|--------|
| OSC 7 `\e]7;file://hostname/path\a` | Ignored — tab doesn't update |
| OSC 0 `\e]0;title\a` (same as fish_title) | Ignored |
| OSC 2 `\e]2;title\a` | Ignored |
| OSC 133 semantic prompt markers (fake command boundary) | Ignored |
| `commandline -f execute` from fish function | Doesn't fire — Warp manages input |
| `commandline -f execute` from `fish_prompt` handler | Fires but terminal input is broken (cursor position reports printed as text) |
| Running command directly from `fish_prompt` handler | Process is killed / Warp shows its own prompt over the running process |

### What does work (workarounds)

1. **Manual Enter press**: Return from the function (prompt renders, tab updates), pre-fill command in buffer via `commandline -r`, user presses Enter
2. **osascript hack**: Same as above but auto-press Enter via `osascript -e 'tell app "System Events" to key code 36'` after a 0.3s delay — fragile and requires accessibility permissions

### Warp already has the CWD data via its own plugin

The [claude-code-warp](https://github.com/warpdotdev/claude-code-warp) plugin sends the correct `cwd` in every OSC 777 notification payload (session_start, prompt_submit, stop, tool_complete, etc.). Warp receives the current working directory on every hook event but doesn't use it to update the tab's CWD/branch display. The data is already there — it just needs to be applied to the tab state.

### Root cause

Warp only reads the shell's CWD during its own prompt rendering cycle via `warp_precmd`. No escape sequence or programmatic approach can trigger a CWD re-read from within a running command or function. This affects any workflow that changes CWD and then immediately launches a long-running process without returning to the prompt:

- Git worktree managers (`wt switch` → launch editor/agent)
- `zoxide`/`autojump` in wrapper functions
- Project launchers that `cd` then start a dev server

### Proposed solution

Support OSC 7 (`\e]7;file://hostname/path\a`) as a CWD update signal, independent of the prompt cycle. This is a standard escape sequence supported by Terminal.app, iTerm2, VS Code terminal, and others. It would let any tool notify Warp of directory changes at any point — not only between commands.

Alternatively, use the `cwd` field already present in the OSC 777 `warp://cli-agent` payloads to update the tab when it changes.

### Additional context

- macOS, Warp v0.2026.04.15.08.45.stable_02
- Shell: fish
- The shell prompt and `git branch --show-current` both reflect the correct new branch after switching — only Warp's tab fails to update
- Related: #3108 (Add support for ANSI OSC) — OSC 7 is the specific sequence needed here

waynehoover:
## Description

Fixes #9125.

Warp's per-block working directory was only refreshed during the shell's prompt rendering cycle (precmd), so external tools that change directory mid-command — for example worktrunk's `wt switch` — left the tab title and git branch chip stuck on the old value until the user submitted another command. The OSC 7 escape `\e]7;file://host/path\a` is the standard way for those tools to inform the terminal of a new CWD.

This PR wires OSC 7 through the terminal handler:

- Parse the sequence in the OSC dispatcher, validating that the host is empty, `localhost`, or the local hostname (case-insensitive) and percent-decoding the path. Unescaped semicolons in the path are preserved.
- Add `Handler::set_current_working_directory`. `Block` updates its `pwd` and emits a new dedicated `BlockWorkingDirectoryUpdated` event — **not** `BlockMetadataReceived`, because that event is contracted to fire exactly once per block at precmd and several subscribers (notably the requested-command finish detector in `ai/blocklist/action_model/execute/shell_command.rs`) rely on that contract.
- `TerminalModel::set_current_working_directory` routes the call directly to `BlockList` (not through the alt-screen `delegate!`) so updates aren't swallowed when a TUI program is on the alt screen. It also drops the update entirely while inside an SSH-launching block, since a remote shell with a coincident hostname could otherwise slip through the host check.
- CWD-dependent subscribers — `ActiveSession`, `BlocklistAIContextModel`, and `TerminalView` (via the new shared `apply_block_metadata_update`) — opt in by listening to `BlockWorkingDirectoryUpdated` in addition to `BlockMetadataReceived`. Subscribers tied to precmd semantics keep listening only to `BlockMetadataReceived`.

## Testing

- New OSC 7 unit tests in `app/src/terminal/model/ansi/mod_tests.rs` covering parsing, ST/BEL terminators, percent decoding (including malformed/truncated escapes), unescaped semicolons in paths, host validation (empty / `localhost` / local hostname / case-insensitive / non-local), non-`file` schemes, and missing/empty payloads.
- New `test_set_current_working_directory_updates_pwd_and_emits_cwd_event` in `app/src/terminal/model/block_tests.rs` verifies the `pwd` update, the no-op on duplicate path, that exactly one `BlockWorkingDirectoryUpdated` is emitted per distinct path, and asserts that **no** `BlockMetadataReceived` event is emitted (preserving the once-per-block precmd contract).
- New `block_working_directory_updated_does_not_drain_finish_senders` in `app/src/ai/blocklist/action_model/execute/shell_command_tests.rs` locks in the contract that `ShellCommandExecutor`'s requested-command finish detector reacts only to `BlockMetadataReceived` (precmd) and ignores `BlockWorkingDirectoryUpdated` (OSC 7).
- Bash + zsh integration test (`test_osc7_updates_current_working_directory`) under `crates/integration/` runs `printf '\033]7;file:///tmp/osc7-test\a'` and asserts the previous block's `pwd` was updated.

`cargo fmt --check` is clean. The new code passes `cargo clippy -p warp` without introducing any new warnings.

## Server API dependencies

No server changes.

## Agent Mode

CHANGELOG-BUG-FIX: Update the tab's working directory and git branch indicator when a process emits an OSC 7 escape sequence to report a new CWD (e.g. worktrunk's `wt switch`).


waynehoover:
@cla-bot check

app/src/terminal/model/block.rs
```diff
@@ -3211,6 +3211,27 @@ impl ansi::Handler for Block {
         delegate!(self.text_area_size_chars(writer));
     }
 
+    fn set_current_working_directory(&mut self, path: String) {
+        if self.pwd.as_deref() == Some(path.as_str()) {
+            return;
+        }
+        self.pwd = Some(path);
+        self.event_proxy
+            .send_terminal_event(Event::BlockMetadataReceived(BlockMetadataReceivedEvent {
```
zachbai: I'm fairly certain that we implicitly rely on this event being emitted once per block - as that's been the case since the event was introduced years ago.

It's probably safer to introduce a new, more tightly scoped event variant and just ensure that it's probably broadcasted to dependent logic that is currently solely relying on this event to subscribe to updated working directory.

zachbai (changes requested):
Thanks for the PR!  Mostly looks good, except for the one comment i left.

app/src/terminal/model/block.rs
waynehoover: Good call — pushed 2eaf7eb that does exactly that.

OSC 7 now emits a new `BlockWorkingDirectoryUpdated` event variant instead of reusing `BlockMetadataReceived`. The new event mirrors the `BlockMetadataReceivedEvent` payload (block_metadata, block_index, is_after_in_band_command, is_done_bootstrapping) so CWD-dependent listeners can reuse the same handling, but it is opt-in:

- `TerminalView` (tab title / repo detection / git status / prompt refresh)
- `ActiveSession` (UpdatedPwd emission)
- `BlocklistAIContextModel` (directory context)

…each subscribe to **both** events via a small shared helper, so OSC 7 still drives every listener that genuinely cares about the CWD.

Subscribers tied to precmd semantics — most importantly the requested-command finish detector in `ai/blocklist/action_model/execute/shell_command.rs` — are unchanged and continue listening only to `BlockMetadataReceived`, preserving the once-per-block contract you flagged.

The block test was updated to assert that `set_current_working_directory` emits `BlockWorkingDirectoryUpdated` **and** that it does **not** emit `BlockMetadataReceived`, so any future regression on this contract will fail loudly.

app/src/ai/blocklist/action_model/execute/shell_command.rs
```diff
@@ -1,3 +1,7 @@
+#[cfg(test)]
```
zachbai: this is a mega nit but please move test modules to the bottom of the file per convention in the rest of the codebase

waynehoover: Done.

zachbai:
@waynehoover  Will give this a spin tomorrow and just do some testing locally, but change itself looks good.

waynehoover:
@zachbai any progress on this? It would be nice to land on a solution here. Others are also expressing desire for a resolution: https://github.com/warpdotdev/warp/issues/9125#issuecomment-4475899453

zachbai:
@waynehoover sorry for the delay, lost this in my inbox. Will look at this today

zachbai (changes requested):
@waynehoover I took a look at this and was seeing the correct behavior for updating git branch, but wasn't seeing the same for current working directoyr (at insofar as actually updating the user-facing UI)

https://www.loom.com/share/f00b9513ad9d4072874759f18c41f32e

Were you able to actually test this change locally and see a difference? 

I'm guessing that the working directly update is not actually being propagated to the tab UI surfaces + trriggering a re-render.

waynehoover:
@zachbai good catch — you were right that the update wasn't propagating to the UI. Just pushed a fix in 2a8186c2.

**Root cause:** the block's `pwd` and `git_repo_status` were updating fine (which is why the git branch chip worked in your loom), but the `WorkingDirectory` prompt chip text — which `display_working_directory` reads, and that's what feeds the vertical-tab subtitle / tab tooltip — is a contextual generator that reads from `CurrentPrompt::latest_context`. That context is only refreshed via `refresh_warp_prompt → current_prompt.update_context(active_block)`, and in the normal precmd flow that refresh is triggered by `BlockCompleted`. OSC 7 fires *mid-command*, so the block never completes and the chip generator never re-ran. Git branch updated independently because the repo-detection path in `apply_block_metadata_update` refreshes `git_repo_status` directly, bypassing the chip.

**Fix:** call `refresh_warp_prompt(ctx)` after dispatching `BlockWorkingDirectoryUpdated` through `apply_block_metadata_update`, gated on `!is_for_in_band_command` to match the existing in-band guard (so we don't re-fire chip generators that schedule more in-band commands).

Also extended the OSC 7 integration test to assert `view.display_working_directory(ctx)` reflects the new path, so this user-visible regression is locked in alongside the existing block-pwd assertion. Ready for another spin when you have a minute.

zachbai:
Awesome, thanks, will get to this today

zachbai (approved):

zachbai:
@waynehoover Just tested with the latest commit, working directory seems to be properly updated, running CI now then will merge!

PR #9285, opened 2026-04-28
Open "Default App" files in the running Warp channel

### LINKED ISSUE #9189: Preview: 'Default App' in Open File settings opens files in Stable Warp instead of Preview
## Summary
On Warp Preview edition, the "Default App" option in the "Open File" settings causes files to be opened in Stable Warp rather than the Preview edition that is currently running.

## Problem
When using Warp Preview, the "Open File" setting has a "Default App" option. Selecting this option is expected to open files in the currently-running Preview build. Instead, it launches or directs the file to the Stable Warp installation.

## Expected behavior
Clicking "Open File" → "Default App" in Warp Preview should open the file using the Preview edition (e.g., WarpDev.app), since that is the active application.

## Actual behavior
The file is opened in Stable Warp (Warp.app) instead of the Preview edition (WarpDev.app), suggesting the "Default App" association is resolving to the Stable channel rather than the Preview channel.

1. Install both Stable Warp and Warp Preview (WarpDev.app).
2. Open Warp Preview.
3. Navigate to Settings → "Open File" and set the option to "Default App."
4. Trigger an open-file action (e.g., click a file path in the terminal output).
5. Observe that the file opens in Stable Warp, not Warp Preview.

## Artifacts
None attached.

## Warp version
v0.2026.04.24.08.52.dev_00 (Preview)

## Operating system
macOS 26.4.1 (build 25E253)

Faizanq:
## Description

Fixes #9189. When the user has multiple Warp channels installed (e.g. Stable + Preview), LaunchServices may register the Stable build as the default handler for common file types. The "Open File → Default App" path then resolves to that sibling channel and hands files to Stable even when the user is running Preview.

The bug sits at `app/src/util/file/external_editor/mac.rs:336`: `default_app_to_open_path` returns the LaunchServices default, `Editor::new_from_identifier` has no `dev.warp.*` variants so the result is dropped, and we fall through to `ctx.open_file_path(full_path)` which is `[NSWorkspace openURL:]`. NSWorkspace then routes via the same default-app it just resolved, so the file lands in the other Warp.

The fix detects this case after the existing editor-resolution path: if the system default is a Warp bundle other than the running channel's bundle (`ChannelState::app_id()`), open the file with the running channel's bundle directly via `/usr/bin/open -b`. Non-Warp defaults (VS Code, TextEdit, etc.) are unaffected, and when the system default already matches the running channel the override is skipped.

## Testing

Added 2 unit tests in `app/src/util/file/external_editor/mac.rs` for the new `is_warp_bundle` helper:

- `dev.warp.{Warp,WarpDev,WarpPreview,WarpOss}` → matches
- `com.microsoft.VSCode`, `com.apple.TextEdit`, `dev.zed.Zed`, `invalid`, empty → does not match

Both pass under `cargo test -p warp --lib util::file::external_editor::mac::tests`.

The end-to-end behavior (file actually landing in the running channel) requires a multi-channel macOS install to verify; I didn't manually reproduce because the build I have access to is the OSS channel only. Logic is small enough that a Warp dev with Stable + Preview side-by-side should be able to confirm in seconds.

## Agent Mode

CHANGELOG-BUG-FIX: `Open File → Default App` now opens files in the running Warp channel instead of routing to a different installed Warp.


captainsafia:
@Faizanq Sorry you're running into issues re-activating the agent review here! We'll route it to a member of the team while I resolve the issues here.

alokedesai (approved):
Appreciate the fix, thanks!

PR #9277, opened 2026-04-28
Expand `~` in `warp://action/new_tab?path=` URLs

Faizanq:
## Description

Fixes #9178. `warp://action/new_tab?path=~/foo` was opening the current directory instead of `~/foo`. The query value was being passed straight into `PathBuf::from_str`, which doesn't expand `~`, so the resulting literal `~/foo` couldn't be resolved by `open_file` and the new tab silently fell back to the current location.

Routed the `path` query through `shellexpand::tilde` (already a workspace dependency) before constructing the `PathBuf`. The same call site serves both `/new_tab` and `/new_window`, so both URLs benefit. Absolute and relative paths are unchanged.

## Testing

Added 6 unit tests in `app/src/uri/uri_test.rs`:

- `~/Projects` expands to `$HOME/Projects`
- URL-encoded `%7E%2FProjects` expands the same way
- absolute path `/tmp/foo` unchanged
- relative path `relative/dir` unchanged
- missing `path=` returns `None`
- bare `~` expands to `$HOME`

Also verified the helper in a standalone harness against extra edge cases: empty value, non-leading `~` (unchanged, correct shell semantics), `~user/path` (unchanged, a shellexpand limitation), and URL-encoded spaces.

Couldn't run the in-tree test suite locally because the Metal toolchain isn't installed, so relying on CI for the full clippy / nextest pass.

## Agent Mode
- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

## Changelog Entries for Stable

CHANGELOG-BUG-FIX: `warp://action/new_tab?path=~/foo` (and `/new_window`) now expand `~` to your home directory.

PR #9300, opened 2026-04-28
Don't require login to open file from URI

seemeroland:
## Description

This gate was added ~2 years ago to prevent some panic, but I confirmed the panic doesn't happen anymore. Warp no longer requires login, so we should support opening files in warp even when not logged in.

I noticed this while testing #9277 

## Testing

Tested locally. While logged out, `open "warplocal://action/new_tab?path=/Users/rolandhuang/warp"` opens the tab in that directory correctly. 

Also tested with right click on file > open in WarpLocal while logged out works

While on the login screen, it just focuses the window.


kevinchevalier (approved):
Fix the import and it looks good.

PR #9305, opened 2026-04-28
Add /set-tab-color slash command

sebryu:
## Description

Adds a programmatic way to set or clear the active tab's color, mirroring the manual color selection in the right-click context menu. Closes the gap between the existing `/rename-tab` (programmatic tab rename) and the previously click-only color picker.

The new `/set-tab-color <color>` slash command accepts:

- `red`, `green`, `yellow`, `blue`, `magenta`, `cyan` — the same six options the right-click picker exposes (`TAB_COLOR_OPTIONS`).
- `none`, `default`, `clear`, or `reset` — clears any manual override (the tab falls back to its default directory color when `DirectoryTabColors` is on).

Implementation summary:

- New `WorkspaceAction::SetActiveTabColor(Option<AnsiColorIdentifier>)` action variant; reuses the existing `TabTelemetryAction::SetColor` / `ResetColor` events.
- New `WorkspaceView::set_tab_color(index, Option<AnsiColorIdentifier>, ctx)` method that maps `Some/None` to `SelectedTabColor::Color` / `Cleared` (or `Unset` when `DirectoryTabColors` is off) — same toggle/clear semantics the right-click flow uses.
- `SET_TAB_COLOR` registered alongside `RENAME_TAB` and dispatched from `Input::execute_slash_command`. Empty or unrecognized colors surface a toast.

## Testing

- `cargo fmt -- --check` — passes.
- `cargo clippy -p warp --all-targets --tests -- -D warnings` — passes.
- `cargo test -p warp --lib -- set_tab_color test_set_active_tab_color` — both new tests pass:
  - `commands::tests::set_tab_color_command_requires_argument` verifies the slash command is registered with the expected required argument and hint text.
  - `workspace::view::tests::test_set_active_tab_color` drives the action end-to-end: set, replace, clear, and confirms it only affects the active tab.

## Server API dependencies

No server API changes.

## Agent Mode

CHANGELOG-NEW-FEATURE: Added a `/set-tab-color` slash command for setting or clearing the current tab's color from the input bar.

### DIFF
--- a/app/src/search/slash_command_menu/static_commands/commands.rs
+++ b/app/src/search/slash_command_menu/static_commands/commands.rs
@@ -135,15 +135,23 @@ pub static RENAME_TAB: LazyLock<StaticCommand> = LazyLock::new(|| StaticCommand 
     argument: Some(Argument::required().with_hint_text("<tab name>")),
 });
 
-pub static SET_TAB_COLOR: LazyLock<StaticCommand> = LazyLock::new(|| StaticCommand {
-    name: "/set-tab-color",
-    description: "Set the color of the current tab",
-    icon_path: "bundled/svg/ellipse.svg",
-    availability: Availability::ALWAYS,
-    auto_enter_ai_mode: false,
-    argument: Some(
-        Argument::required().with_hint_text("<red|green|yellow|blue|magenta|cyan|none|default>"),
-    ),
+pub static SET_TAB_COLOR: LazyLock<StaticCommand> = LazyLock::new(|| {
+    let mut hint = String::from("<");
+    for color in crate::ui_components::color_dot::TAB_COLOR_OPTIONS {
+        hint.push_str(&color.to_string().to_ascii_lowercase());
+        hint.push('|');
+    }
+    hint.push_str("none>");
+    let hint_text: &'static str = Box::leak(hint.into_boxed_str());
+
+    StaticCommand {
+        name: "/set-tab-color",
+        description: "Set the color of the current tab",
+        icon_path: "bundled/svg/ellipse.svg",
+        availability: Availability::ALWAYS,
+        auto_enter_ai_mode: false,
+        argument: Some(Argument::required().with_hint_text(hint_text)),
+    }
 });
 
 pub static FORK: LazyLock<StaticCommand> = LazyLock::new(|| {
@@ -674,10 +682,15 @@ mod tests {
 
         assert!(!argument.is_optional);
         assert!(!argument.should_execute_on_selection);
-        assert_eq!(
-            argument.hint_text,
-            Some("<red|green|yellow|blue|magenta|cyan|none|default>")
-        );
+
+        let hint = argument
+            .hint_text
+            .expect("/set-tab-color hint text is set dynamically");
+        for color in crate::ui_components::color_dot::TAB_COLOR_OPTIONS {
+            let lower = color.to_string().to_ascii_lowercase();
+            assert!(hint.contains(&lower), "hint should mention `{lower}`");
+        }
+        assert!(hint.contains("none"), "hint should mention `none`");
     }
 
     #[test]
--- a/app/src/search/slash_command_menu/static_commands/commands.rs
+++ b/app/src/search/slash_command_menu/static_commands/commands.rs
@@ -142,8 +142,7 @@ pub static SET_TAB_COLOR: LazyLock<StaticCommand> = LazyLock::new(|| StaticComma
     availability: Availability::ALWAYS,
     auto_enter_ai_mode: false,
     argument: Some(
-        Argument::required()
-            .with_hint_text("<red|green|yellow|blue|magenta|cyan|none>"),
+        Argument::required().with_hint_text("<red|green|yellow|blue|magenta|cyan|none|default>"),
     ),
 });
 
@@ -677,7 +676,7 @@ mod tests {
         assert!(!argument.should_execute_on_selection);
         assert_eq!(
             argument.hint_text,
-            Some("<red|green|yellow|blue|magenta|cyan|none>")
+            Some("<red|green|yellow|blue|magenta|cyan|none|default>")
         );
     }
 
--- a/app/src/search/slash_command_menu/static_commands/commands.rs
+++ b/app/src/search/slash_command_menu/static_commands/commands.rs
@@ -5,6 +5,7 @@ use uuid::Uuid;
 use warp_core::features::FeatureFlag;
 
 use crate::search::slash_command_menu::{static_commands::Argument, StaticCommand};
+use crate::ui_components::color_dot;
 
 use super::Availability;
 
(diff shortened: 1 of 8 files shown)


sebryu:
@cla-bot check

sebryu:
Thanks for the review! Addressed the `default`/`reset` vs `none`/`clear` distinction in 3248216 — `default`/`reset` now map to `SelectedTabColor::Unset` (lets the directory default apply), while `none`/`clear` map to `Cleared` (suppresses everything). Updated tests, hint text, and the empty-arg error toast accordingly.

/oz-review

captainsafia:
@sebryu Thanks for addressing the feedback! I'm working on resolving some issues with re-triggers on the oz-agent so that you can we get a re-review of this PR for you soon!

app/src/search/slash_command_menu/static_commands/commands.rs
```diff
@@ -135,6 +135,17 @@ pub static RENAME_TAB: LazyLock<StaticCommand> = LazyLock::new(|| StaticCommand
...
 });
 
+pub static SET_TAB_COLOR: LazyLock<StaticCommand> = LazyLock::new(|| StaticCommand {
+    name: "/set-tab-color",
+    description: "Set the color of the current tab",
+    icon_path: "bundled/svg/ellipse.svg",
+    availability: Availability::ALWAYS,
+    auto_enter_ai_mode: false,
+    argument: Some(
+        Argument::required().with_hint_text("<red|green|yellow|blue|magenta|cyan|none|default>"),
```
harryalbert: Could we derive this from the `AnsiColorIdentifier` struct (we'd need to add a toString so that we could go from variant -> string) so that, if we ever add more identifiers, this automatically changes to accommodate them?

app/src/terminal/input/slash_commands/mod.rs
```diff
@@ -420,6 +422,45 @@ impl Input {
...
+                    );
+                    return true;
+                };
+
+                // `none`/`clear` explicitly clear the color (suppressing any
+                // directory default); `default`/`reset` remove the manual override
+                // so the directory default can apply when `DirectoryTabColors`
+                // is enabled.
+                let color = match arg.to_ascii_lowercase().as_str() {
+                    "red" => SelectedTabColor::Color(AnsiColorIdentifier::Red),
```
harryalbert: relating to the comment above, I think it would be good if we had a to and from String implementations for `AnsiColorIdentifier`. That way we could strictly enforce that each variant maps to a string and vice versa

app/src/terminal/input/slash_commands/mod.rs
```diff
@@ -420,6 +422,45 @@ impl Input {
...
+                    "green" => SelectedTabColor::Color(AnsiColorIdentifier::Green),
+                    "yellow" => SelectedTabColor::Color(AnsiColorIdentifier::Yellow),
+                    "blue" => SelectedTabColor::Color(AnsiColorIdentifier::Blue),
+                    "magenta" => SelectedTabColor::Color(AnsiColorIdentifier::Magenta),
+                    "cyan" => SelectedTabColor::Color(AnsiColorIdentifier::Cyan),
+                    "none" | "clear" => SelectedTabColor::Cleared,
```
harryalbert: I don't know if users are going to grok the difference between "cleared" and "unset" (we maintain both of these variants internally so that we can distinguish between a tab that had its color manually set to none versus a tab that just started as none and should therefore be auto-assigned a color based on directory). IMO we should just have "none" (no "reset", "default', or "clear") and that should map to Cleared.

app/src/workspace/view.rs
```diff
@@ -5065,6 +5065,41 @@ impl Workspace {
         ctx.notify();
     }
 
+    /// Programmatically sets the manual color override for a tab.
+    ///
+    /// - `Color(_)` applies that color.
+    /// - `Cleared` explicitly clears the color (also suppresses any directory default).
+    /// - `Unset` removes the manual override, letting the directory default apply.
+    pub fn set_tab_color(
```
harryalbert, line 5073: I'm guessing there's an existing function for this? We'd ideally use the same code path that we use for setting a tab color through the tab overflow menu.

harryalbert (commented):
Thanks for submitting this! This is a great idea, just had a few comments on implementation

sebryu:
Thanks for the review @harryalbert! Addressed all four threads in f8934d9:

1. **Hint text derived from `AnsiColorIdentifier`** — the slash command's hint is now built at registration time from `TAB_COLOR_OPTIONS` so it stays in sync with the tab overflow menu. New colors flow through automatically.

2. **`AnsiColorIdentifier` to/from string** — replaced the manual `Display` impl with `#[derive(strum::Display, strum::EnumString)]` and `#[strum(ascii_case_insensitive)]`. Variant names are still the canonical capitalized form for UI tooltips/labels, and `FromStr` accepts any case. Adding a new variant requires touching only the enum.

3. **Drop `default` / `reset` / `clear`** — the slash command now accepts only `none`, mapping to `SelectedTabColor::Cleared`. Removed the user-facing `Cleared` vs `Unset` distinction. The action enum still supports `Unset` for programmatic callers, but it's no longer reachable from the command line. Hint, error toasts, and tests updated.

4. **Shared code path with the menu** — refactored `toggle_tab_color` to compute the next state and delegate to `set_tab_color`, so the menu's toggle and the slash command's explicit set funnel through the same primitive (state update + telemetry). I didn't dispatch `ToggleTabColor` directly from the slash command because (a) `/set-tab-color red` typed twice would surprise users by clearing on the second invocation, and (b) `ToggleTabColor` only carries an `AnsiColorIdentifier`, so there's no way to express the `none → Cleared` case through it.

/oz-review

app/src/terminal/input/slash_commands/mod.rs
```diff
@@ -420,6 +422,54 @@ impl Input {
 
                 ctx.dispatch_typed_action(&WorkspaceAction::SetActiveTabName(name.to_owned()));
             }
+            set_tab_color if command.name == commands::SET_TAB_COLOR.name => {
+                let supported_options = || {
+                    crate::ui_components::color_dot::TAB_COLOR_OPTIONS
```
harryalbert: small nit: the convention in the codebase is to prefer imports at the top as opposed to long import chains (both for this line and other places in the PR)

app/src/search/slash_command_menu/static_commands/commands.rs
```diff
@@ -135,6 +135,25 @@ pub static RENAME_TAB: LazyLock<StaticCommand> = LazyLock::new(|| StaticCommand
...
 });
 
+pub static SET_TAB_COLOR: LazyLock<StaticCommand> = LazyLock::new(|| {
+    let mut hint = String::from("<");
+    for color in crate::ui_components::color_dot::TAB_COLOR_OPTIONS {
+        hint.push_str(&color.to_string().to_ascii_lowercase());
+        hint.push('|');
+    }
+    hint.push_str("none>");
+    let hint_text: &'static str = Box::leak(hint.into_boxed_str());
```
harryalbert: I don't love using a leak if we can avoid it. I think we could achieve something similar without the leak with something like:
```
static SET_TAB_COLOR_HINT: LazyLock<String> = LazyLock::new(|| {
    let mut hint = String::from("<");
    for color in color_dot::TAB_COLOR_OPTIONS {
        hint.push_str(&color.to_string().to_ascii_lowercase());
        hint.push('|');
    }
    hint.push_str("none>");
    hint
});

pub static SET_TAB_COLOR: LazyLock<StaticCommand> = LazyLock::new(|| StaticCommand {
    name: "/set-tab-color",
    description: "Set the color of the current tab",
    icon_path: "bundled/svg/ellipse.svg",
    availability: Availability::ALWAYS,
    auto_enter_ai_mode: false,
    argument: Some(Argument::required().with_hint_text(SET_TAB_COLOR_HINT.as_str())),
});
```

harryalbert (commented):
thanks for addressing my comments! Just a few more thoughts (plus the comment from the first round about re-using existing functions for setting the tab color)

app/src/search/slash_command_menu/static_commands/commands.rs
sebryu: Done — switched to a separate `static SET_TAB_COLOR_HINT: LazyLock<String>` and pass `SET_TAB_COLOR_HINT.as_str()` to `with_hint_text`, no more `Box::leak`.

app/src/terminal/input/slash_commands/mod.rs
sebryu: Done — added `use crate::ui_components::color_dot;` at the top of both `slash_commands/mod.rs` and `static_commands/commands.rs`, and replaced the inline `crate::ui_components::color_dot::TAB_COLOR_OPTIONS` paths.

app/src/workspace/view.rs
sebryu: I think this is now sharing the same code path: `Workspace::set_tab_color` (the new function) is the single entry point that mutates `tab.selected_color`, sends the `TabOperations` telemetry, and notifies. The tab-overflow menu still goes through `WorkspaceAction::ToggleTabColor` → `toggle_tab_color`, but `toggle_tab_color` was refactored to delegate to `set_tab_color` so both paths converge there. The slash command just dispatches a new `SetActiveTabColor(SelectedTabColor)` action that lands in the same function.

Happy to consolidate further if you had a different shape in mind — e.g. dropping the new action variant and reusing `ToggleTabColor` (though the semantics differ since "toggle" flips an already-set color back to unset/cleared, while the slash command always sets to the requested value).

harryalbert: No I think this makes sense, the new fn shape looks good!

harryalbert (approved):
nice! Thanks for iterating on this! I'll merge this after the CI checks pass

harryalbert:
Thanks for implementing this @sebryu!

PR #9308, opened 2026-04-28
fix(mcp): add size-based log rotation to prevent unbounded MCP log growth

### LINKED ISSUE #8993: [Windows] MCP server log files grow unbounded, consuming 18+ GB of disk space
## Bug Description

MCP server log files in Warp on Windows grow without any size limit or rotation, consuming massive amounts of disk space. A single log file reached **16.4 GB** and total MCP logs consumed **18.3 GB** across 1,423 log files.

## Environment

- **OS:** Windows 11 Home Single Language (10.0.26200)
- **Warp Version:** v0.2026.03.25.08.24.stable_05
- **Architecture:** x64
- **Install Method:** winget (Warp.Warp)

## Disk Usage Breakdown

```
AppData\Local\warp\Warp\data\logs\mcp\
  3015eae1-de41-4460-bedd-6b015ee59e8d.log    16,456 MB (16.4 GB!)
  b23235ab-8257-4fd5-8d13-213a51d8fa50.log     1,751 MB (1.7 GB)
  e8aad4f7-e47f-4b68-84d4-a56530df41fc.log       119 MB
  46b32854-8fc8-49e4-a5f9-1cf45b94df37.log         3.8 MB
  c78da7ed-05b0-4c94-a36e-52e533313afe.log       0.9 MB
  ...

Total: 1,423 log files = 18,340 MB (18.3 GB)
```

Other Warp data:
```
warp.sqlite                                91 MB
AppData\Roaming\warp                     25 MB
Program Files\Warp                        ~200 MB (installation)
------------------------------------------
TOTAL WARP FOOTPRINT                      ~18.7 GB
```

## Steps to Reproduce

1. Use Warp on Windows with MCP servers configured (Claude Code MCP, etc.)
2. Run development sessions that generate MCP traffic
3. Observe that log files in `AppData\Local\warp\Warp\data\logs\mcp\` grow indefinitely
4. No rotation, truncation, or size limit is applied

## Expected Behavior

- MCP log files should have a maximum size limit (e.g., 10-50 MB per file)
- Old log entries should be rotated or truncated
- Total log storage should be bounded (e.g., 500 MB max)
- User should be able to configure log retention settings

## Actual Behavior

- Log files grow unbounded to 16+ GB each
- No rotation or truncation occurs
- Total disk usage can reach 18+ GB from logs alone
- No user-configurable settings to control log size

## Workaround

Manually delete log files:
```powershell
Remove-Item "$env:LOCALAPPDATA\warp\Warp\data\logs\mcp\*.log" -Force
```

This frees up ~18 GB but logs will grow unbounded again.

## Impact

- **Severe** - Can consume entire SSD storage
- Particularly bad for developers on smaller SSDs (128GB/256GB laptops)
- No warning or notification to the user
- Silent disk space consumption

## Suggested Fix

1. **Log rotation**: Rotate log files at a configurable size limit (default 50 MB)
2. **Retention policy**: Delete logs older than N days (default 7)
3. **Total cap**: Enforce maximum total log storage (default 500 MB)
4. **Settings UI**: Add log management to Warp Settings
5. **Periodic cleanup**: Background task to clean old/large logs

## Additional Context

- The SQLite database (`warp.sqlite`) is 91 MB which seems reasonable
- This only affects MCP logs, not the main `warp.log` files which appear to have some rotation
- The log file names appear to be UUIDs suggesting they are per-session or per-MCP-server

OthmanAdi:
## Description

Fixes #8993

MCP server log files grow without any size limit or rotation. A single log file was reported at **16.4 GB**, with total MCP log storage consuming **18.3 GB** across 1,423 files on Windows. The only cleanup mechanism is a full directory purge on application startup, meaning logs accumulate unbounded during a single session.

**Root cause:** `SimpleLogger` (in `crates/simple_logger/`) opens log files with truncate on creation and writes every line via an unbounded async channel to a background task. There is zero size checking, zero rotation, and zero max file size logic. The main Warp application log (in `crates/warp_logging/`) already has rotation (up to 5 files for GUI, 10 for CLI), but this was never implemented for `SimpleLogger`.

**Fix:** Add size-based log rotation to `SimpleLogger`'s async writer task. When the current log file exceeds 50 MB, the writer:

1. Flushes and drops the async file handle
2. Deletes the oldest rotated file (`.old.4`)
3. Shifts existing rotated files up (`.old.3` -> `.old.4`, `.old.2` -> `.old.3`, etc.)
4. Renames the current file to `.old.0`
5. Opens a fresh empty log file
6. Continues logging

This mirrors the rotation pattern in `crates/warp_logging/src/native.rs` (lines 132-184).

### Bounds

| Parameter | Value | Rationale |
|---|---|---|
| `MAX_LOG_FILE_SIZE` | 50 MB | Large enough to capture extended sessions, small enough to prevent disk abuse |
| `MAX_ROTATED_FILES` | 5 | Matches the GUI rotation count for the main Warp log |
| **Total storage per server** | ~300 MB | 50 MB active + 5 x 50 MB rotated |
| **Rotation trigger** | Before write that would exceed limit | Never exceeds 50 MB + one line |

### What this changes

| Before | After |
|---|---|
| Log files grow to 16+ GB | Capped at 50 MB per file |
| No rotation at all | Up to 5 rotated copies preserved |
| Disk usage unbounded | Bounded at ~300 MB per MCP server |
| No user-configurable limits | Constants can be tuned (future: settings) |

## Testing

- [x] `cargo check -p simple_logger -p warp` passes clean
- The rotation logic follows the same pattern as `warp_logging/src/native.rs::rotate_files` (lines 154-184), which is already in production
- `std::fs::rename` and `std::fs::remove_file` are synchronous, which is correct here: rotation is rare (once per 50 MB) and must complete atomically before the new file is opened
- Byte counter tracks writes in-memory rather than calling `metadata()` per line, avoiding unnecessary syscalls on every log write

## Server API dependencies
- [x] N/A

## Agent Mode
- [ ] Warp Agent Mode

CHANGELOG-BUG-FIX: MCP server log files now rotate at 50 MB with a maximum of 5 rotated files per server, preventing unbounded disk usage on Windows and all platforms.

### DIFF
--- a/crates/simple_logger/src/lib.rs
+++ b/crates/simple_logger/src/lib.rs
@@ -10,40 +10,32 @@ use warpui::r#async::executor::{Background, BackgroundTask};
 
 pub mod manager;
 
-/// Shared state for a [`SimpleLogger`].
-///
-/// When all [`SimpleLogger`] clones are dropped, this is dropped too, which closes
-/// the logging channel and lets the background writing task finish.
-///
-/// We also support explicit shutdown via [`SimpleLogger::close`]. That allows a
-/// caller to mark a log stream as finished immediately, even if some incidental
-/// clones are still alive briefly in background tasks or callback state.
+const MAX_LOG_FILE_SIZE: u64 = 50 * 1024 * 1024;
+const MAX_ROTATED_FILES: usize = 5;
+
+fn rotated_path(path: &PathBuf, index: usize) -> PathBuf {
+    let mut s = path.as_os_str().to_owned();
+    s.push(format!(".old.{}", index));
+    PathBuf::from(s)
+}
+
 pub(crate) struct LogFileWriter {
     log_tx: Sender<String>,
     _logging_task: BackgroundTask,
 }
 
 impl LogFileWriter {
-    /// Returns true if the underlying channel has been closed.
-    ///
-    /// A closed writer is logically dead even if some [`Arc`] handles still
-    /// exist, because it can no longer accept new log lines.
     pub(crate) fn is_closed(&self) -> bool {
         self.log_tx.is_closed()
     }
 }
 
-/// A simple file-based logger for server stderr output.
-/// Writes timestamped log entries to a file asynchronously.
 #[derive(Clone)]
 pub struct SimpleLogger {
-    // Cheaply cloneable reference to the log file writer.
     writer: Arc<LogFileWriter>,
 }
 
 impl SimpleLogger {
-    /// Creates a new logger that writes to the specified file path.
-    /// The file is truncated on creation.
     pub(crate) fn new(log_path: PathBuf, executor: Arc<Background>) -> Self {
         let (log_tx, log_rx) = async_channel::unbounded::<String>();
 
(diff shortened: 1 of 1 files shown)


OthmanAdi:
Updated to address the security concern:

1. Removed the `bytes_written > 0` guard so rotation triggers even on a fresh/empty file. The empty file still gets rotated (renamed to .old.0) and a new one is opened, so the total rotated set still caps at ~300 MB.

2. Added a per-line truncation limit of 1 MB (`MAX_LOG_LINE_BYTES`). Any single log message exceeding this gets its first 1 MB written followed by `... [truncated]`. This bounds the worst case for any individual write, regardless of what the MCP server sends.

The stderr accumulation bug in `templatable_manager/native.rs` (where `read_line` appends to an uncleared buffer) is a pre-existing issue separate from log rotation. Happy to file a follow-up issue for that if it would be helpful.

OthmanAdi:
@cla-bot check

PR #9326, opened 2026-04-28
Fix read_files with out-of-bounds line ranges producing empty result

kevinchevalier:
## Description

Fixes APP-4002

Screenshot 2026-04-28 at 5 42 20 PM

This is a defensively coded UI in case something else leads to missing file contexts.

Screenshot 2026-04-28 at 5 40 44 PM

When an LLM requests a `read_files` tool call with line ranges entirely beyond EOF (e.g., lines 1891–2090 of a 1237-line file), the client returns `ReadFilesResult::Success { files: [] }` — the file silently vanishes from the result. This renders as an empty box with a green checkmark in the blocklist UI.

**Root cause:** `TextFileAccumulator::flush_range` only emitted a segment on final flush for whole-file reads (`final_flush && self.whole_file`). For ranged reads where no lines fall within the requested range, the buffer stays empty and no segment is emitted. The file ends up in neither `file_contexts` nor `missing_files`.

**Fix:**
1. **Primary:** Change `flush_range` to always emit a segment on final flush (`!self.buf.is_empty() || final_flush`), so every requested range produces a `TextFileSegment` even when entirely past EOF.
2. **Defensive:** Add rendering fallback that shows a red X error box with "Failed to read files" when `ReadFilesResult::Success` has empty `file_contexts`, preventing empty boxes for any future edge case.
3. **Tests:** Updated existing test and added two new tests covering out-of-bounds ranges on non-empty files.

## Testing

- Updated `empty_file_with_ranges_produces_no_segment` → `empty_file_with_ranges_produces_empty_segment` (expects 1 segment with empty content)
- Added `range_past_eof_produces_empty_segment` — 5-line file, range 10..15
- Added `multiple_ranges_some_past_eof` — 5-line file, ranges [1..3, 10..15]
- All 23 text_file_reader tests pass
- `cargo fmt` and `cargo clippy` clean

## Agent Mode

CHANGELOG-BUG-FIX: Fixed read_files tool showing an empty box when the LLM requests line ranges beyond the end of a file.

---

[Conversation](https://staging.warp.dev/conversation/ef9f9688-ca1e-4eb3-b29a-35c5c13e80da) | [Plan](https://staging.warp.dev/drive/notebook/lAlU2uG6KN5y96mdmk8wYL)

### DIFF
--- a/app/src/ai/blocklist/block/view_impl/output.rs
+++ b/app/src/ai/blocklist/block/view_impl/output.rs
@@ -431,11 +431,35 @@ pub(super) fn render(props: Props, app: &AppContext) -> Box<dyn Element> {
                                                 },
                                             ),
                                         ..
-                                    }) => group_file_contexts_for_display(
-                                        file_contexts,
-                                        props.shell_launch_data,
-                                        props.current_working_directory,
-                                    ),
+                                    }) => {
+                                        if file_contexts.is_empty() {
+                                            // Empty file contexts — render as a failed
+                                            // action so the user sees the error instead
+                                            // of an empty box.
+                                            let formatted_text = render_requested_action_body_text(
+                                                "Failed to read files".into(),
+                                                appearance.ui_font_family(),
+                                                app,
+                                            );
+                                            let renderable_action =
+                                                RenderableAction::new_with_formatted_text(
+                                                    formatted_text,
+                                                    app,
+                                                )
+                                                .with_icon(
+                                                    inline_action_icons::red_x_icon(appearance)
+                                                        .finish(),
+                                                );
+                                            output_items
+                                                .add_child(renderable_action.render(app).finish());
+                                            continue;
+                                        }
+                                        group_file_contexts_for_display(
+                                            file_contexts,
+                                            props.shell_launch_data,
+                                            props.current_working_directory,
+                                        )
+                                    }
                                     // if not completed/successful, generate a user message without line count
                                     _ => files
                                         .iter()
--- a/specs/read-files-out-of-bounds-fix/TECH.md
+++ b/specs/read-files-out-of-bounds-fix/TECH.md
@@ -1,0 +1,36 @@
+# TECH.md — Fix read_files with out-of-bounds line ranges
+
+## Context
+
+When an LLM requests a `read_files` tool call with line ranges entirely beyond EOF (e.g., lines 1891–2090 of a 1237-line file), the client returns `ReadFilesResult::Success { files: [] }` — the file silently vanishes from the result. This renders as an empty box with a green checkmark in the blocklist UI, and the server receives `anyFilesSuccess: {}` with zero files.
+
+The root cause is in `TextFileAccumulator::flush_range` (`crates/warp_files/src/text_file_reader.rs:131`). The condition `let should_emit = !self.buf.is_empty() || (final_flush && self.whole_file)` only emits a segment on final flush for whole-file reads. For ranged reads where no lines fall within the requested range, `buf` stays empty and no segment is emitted. The caller (`read_local_file_context` in `app/src/ai/blocklist/action_model/execute.rs:1029–1045`) extends `file_contexts` from an empty segments iterator, then `continue`s past the binary fallback — the file ends up in neither `file_contexts` nor `missing_files`.
+
+Relevant files:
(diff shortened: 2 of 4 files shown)


kevinchevalier:
* **#9326** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9326?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9326?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -422,7 +422,7 @@ pub(super) fn render(props: Props, app: &AppContext) -> Box<dyn Element> {
                                 // checks if the read file action result is completed and successful.
                                 // if successful, we have FileContext with pre-computed line counts that we use to clamp displayed file ranges to the length of the file
                                 let file_names = match agent_action_results {
-                                    // if completed and succesful, generate a user message with file info + line count
+                                    // if completed and successful, generate a user message with file info + line count
```
vkodithala, line 425: Nice lol

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -431,11 +431,43 @@ pub(super) fn render(props: Props, app: &AppContext) -> Box<dyn Element> {
...
-                                        file_contexts,
-                                        props.shell_launch_data,
-                                        props.current_working_directory,
-                                    ),
+                                    }) if !file_contexts.is_empty() => {
+                                        group_file_contexts_for_display(
+                                            file_contexts,
+                                            props.shell_launch_data,
+                                            props.current_working_directory,
+                                        )
```
vkodithala, line 461: nit: Can we match the `ReadFilesResult::Success` arm directly and then use conditions inside? This was a little hard to read on first glance, but non-blocking

specs/read-files-out-of-bounds-fix/TECH.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Proposed changes
+
+### 1. `TextFileAccumulator::flush_range` — always emit on final flush
+Change the `should_emit` condition from `!self.buf.is_empty() || (final_flush && self.whole_file)` to `!self.buf.is_empty() || final_flush`. This makes the behavior consistent: every requested range always produces a `TextFileSegment`, even when the range is entirely past EOF. The emitted segment has `content: ""`, the original requested `line_range`, and `line_count` set to the total file lines (populated by `finalize`).
+
+This is a one-line change. The `whole_file` field remains used for trailing-newline preservation logic, so it is not removed.
+
+### 2. Defensive rendering for empty `file_contexts`
+In the `ReadFilesResult::Success` rendering arm, add a match guard `if !file_contexts.is_empty()` for the normal path. Add a second arm `if file_contexts.is_empty()` that renders an error-styled action box with a red X icon and "Failed to read files" message, then `continue`s. This prevents an empty box regardless of upstream cause.
```
vkodithala, line 22: I think this change is fine for now, but out of curiosity - is there a reason we're continuing to process this as a `Success` result for read_files?  


I think the right long-term fix here might be to map read_files tool calls on out-of-bounds ranges to failures so that the agent can course-correct, but lemme know if this is naive and there's a reason you're going for the UI-only fix for now. Non-blocking, if anything it's a fast follow

vkodithala (approved):
Thanks for the fix :relaxed:

specs/read-files-out-of-bounds-fix/TECH.md
kevinchevalier: That's a great question.  The server API right now either allows `Success`, with a mixture of successfully read files and `missing_files`. If we return a failure then we can't return any file results to the server. So if one file was read from an incorrect range, it would not send the contents of other files that were successfully read.

I thought about adding to `missing_files`, but that feels semantically wrong. The file is there, but the line range is not.

The best choice is probably to add a new field to success, for missing ranges, and deprecate the missing files field. But this would be a much bigger change, and I'm wondering if the model will understand what I mean when I return am empty string as the content.

What do you think? Is it worth making the server change?

app/src/ai/blocklist/block/view_impl/output.rs
kevinchevalier: Yeah, I agree.

PR #9356, opened 2026-04-29
Fix file path detection across wrapped lines

### LINKED ISSUE #9193: Warp doesn't parse long path
### Pre-submit Checks

### Describe the bug

When the file path (.MD for example) goes beyond the width of the window, Warp will fail to parse it and propose to open a folder instead.
If you're unlucky and the path isn't ending on a folder, then no recognition will take place that you can even open it via CTRL/CMD+click

### To reproduce

Create a very long folder path:
```mkdir -p src/modules/core/internal/utils/wrappers/adapters/interfaces/implementations/factories/```
Then create a .MD file at that path:
```touch src/modules/core/internal/utils/wrappers/adapters/interfaces/implementations/factories/README.md```

### Expected behavior

I expect for warp to give me the CTRL/CMD+click option no matter the file path length

_No response_

### Operating system (OS)

Linux

43

### Shell Version

zsh 5.9

### Current Warp version

v0.2026.04.22.08.46.stable_03

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

None

### PR DESCRIPTION
**Pull request by Obed0101**
## Description
Fixes #9193.

This updates terminal file path detection so fragments split only by a soft-wrapped terminal line are treated as one continuous path. Fragments are still kept separate when either side contains a file-link separator.

A regression test covers the long wrapped path from the issue and verifies detection from both sides of the wrap boundary.

## Testing
- `cargo fmt --check`
- `cargo test -p warp terminal::model::grid::grid_handler::tests::test_possible_file_paths_across_wrapped_lines --no-default-features`

## Server API dependencies
None.

CHANGELOG-BUG-FIX: Fixed long file paths split across wrapped terminal lines not being detected as file links.

### DIFF
--- a/app/src/terminal/model/grid/grid_handler.rs
+++ b/app/src/terminal/model/grid/grid_handler.rs
@@ -111,6 +111,28 @@ impl Fragment {
     }
 }
 
+fn append_fragments_across_soft_wrap(
+    fragments: &mut Vec<Fragment>,
+    mut next_fragments: Vec<Fragment>,
+) {
+    match (fragments.last(), next_fragments.first()) {
+        (Some(last_fragment), Some(next_fragment))
+            if !last_fragment.has_separator() && !next_fragment.has_separator() =>
+        {
+            let mut fragment = fragments.pop().expect("Fragment should exist");
+
+            fragment.content.push_str(&next_fragment.content);
+            fragment.total_cell_width += next_fragment.total_cell_width;
+            fragments.push(fragment);
+
+            next_fragments.remove(0);
+        }
+        _ => (),
+    }
+
+    fragments.append(&mut next_fragments);
+}
+
 #[derive(Debug)]
 pub struct FragmentBoundary(pub Range<Point>);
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Obed0101** [comment]
@cla-bot check

**Comment by alokedesai** [comment]
Thanks for the PR! @kevinyang372 could you take a look?

**Comment by kevinyang372** [comment]
Hi @Obed0101 I am not able to repro the original issue. I also confirmed the regression test you added is in fact passing on current master too.

I don't think the bug is in the grid handler code and it might be platform related. Are you able to repro this?

**Comment by Obed0101** [comment]
Thanks for checking. You're right — if the regression test passes on current master, it doesn't prove the issue.

I based the patch on the soft-wrap fragment merge logic and the contradictory comment there, but I agree we shouldn't merge this unless it reproduces the reported behavior or has a failing regression on master.

I'll take another pass to reproduce the original issue and trace whether the bug is actually in link validation/platform behavior instead of `GridHandler`. If I can't find a failing repro, I'm happy to close this PR to avoid merging an unnecessary change.

**Comment by kevinyang372** [comment]
That sounds good. Thanks @Obed0101!

**Comment by Obed0101** [comment]
I was able to reproduce the issue in the local app with a narrow terminal. The failure mode I saw was that Cmd-click sometimes resolved a valid prefix directory, for example `.../wrappers/adapters`, instead of the full existing `README.md` path when the path wrapped across several rows.

I updated the patch to handle paths spanning more than two soft-wrapped rows. The previous regression only covered a two-row wrap, which is why it did not prove the bug. The new regression covers a file path split across four wrapped rows and verifies that the full path candidate/range is produced from hover points on each wrapped row.

Validation:

```txt
cargo test -p warp terminal::model::grid::grid_handler::tests::test_possible_file_paths_across_multiple_wrapped_lines --no-default-features
cargo test -p warp terminal::model::grid::grid_handler::tests::test_possible_file_paths --no-default-features
cargo build --bin warp-oss --features gui,skip_login
```

I also verified manually in the local app that the narrow wrapped path now opens the full `README.md` instead of the prefix directory.

kevinyang372 (commented):
Nice find! This is definitely something we should fix

My high-level feedback here is this change is trying to do too many things:
1. [Not controversial] For a link within max scan length, highlighting stops working when we have >2 softwrapped rows
2. [More controversial] Also including some changes to bump overall scan length, this will have significant performance implications since now we are doing a lot more scanning when user moves their mouse

Let's focus on 1) and make sure we are scanning forward and backward in the grid until we fully fill up the max scan length limit. 

Also I think the logic could be simplified a bit. Essentially what we want is:
1. Iterate backwards until we either 1) hit a hard-wrapped line 2) fill up max scan limit. Push line range content into fragments
2. Iterate forwards until we either 1) hit a hard-wrapped line 2) fill up max scan limit. Push line range content into fragments (some of the logic could definitely be shared with above)

Obed0101:
Thanks, agreed. I narrowed the patch to preserve the existing file-link scan budget instead of increasing it.

The updated version still scans across multiple soft-wrapped rows, but only until the existing `LINK_NUM_CHARACTER_SCAN` budget is filled in each direction. That keeps the mouse-hover work bounded the same way as before, while fixing the >2 soft-wrapped row case when it fits inside that budget.

I also adjusted the regression coverage so it verifies the multi-row behavior within the existing scan limit.

Validation:

```txt
cargo test -p warp terminal::model::grid::grid_handler::tests::test_possible_file_paths --no-default-features
```

kevinyang372 (approved):
Hey @Obed0101 Sorry this got lost in my inbox (you should be able to re-request a review!)

This looks solid. Do you mind addressing the merge conflict and I can send it to CI?

kevinyang372:
Hey @Obed0101 -- since this thread hasn't been active for a while, I cherry-picked your commit and took it over the finish line here: https://github.com/warpdotdev/warp/pull/12968

You'll still be credited for the commit!

PR #9360, opened 2026-04-29
Recognize .htm files as HTML

Abdalla-Eldoumani:
## Description

`language_by_filename` in `crates/languages/src/lib.rs` recognizes `.html` but not the legacy three-character `.htm` extension. `.htm` is widely produced by static-site generators and historical web tooling (DOS 8.3 filename limits), and is already treated as an HTML/text file elsewhere in the codebase: see `is_development_text_extension` in `crates/warp_util/src/file_type.rs:225`, which lists `"html" | "htm"` together.

Opening a `.htm` file in Warp's editor today shows "Language support is unavailable for this file type" instead of HTML syntax highlighting. This PR resolves the inconsistency.

```diff
-        "html" => language_by_name("html"),
+        "html" | "htm" => language_by_name("html"),
```

## Testing

- Added `html_extensions_resolve_to_html` in `crates/languages/src/lib_tests.rs` covering both `.html` and `.htm`. The test calls `language_by_filename(Path::new("index.htm"))` and asserts `display_name() == "HTML"`; fails on master, passes after the change.
- `cargo fmt -p languages -- --check` passes locally.

## Agent Mode

CHANGELOG-IMPROVEMENT: HTML files using the .htm extension now open with HTML syntax highlighting in Warp's editor.


szgupta (approved):
Awesome, thank you!

PR #9362, opened 2026-04-29
Get warp compiling on FreeBSD

rudrabhoj:
Most `target_os = "linux"` cfg sites in this tree gate code that already works fine on FreeBSD. wayland-client, cctk, fontconfig, x11rb, native-dialog, notify-rust, and the WindowingSystem enum all compile and run there using the existing Linux branches. Widening those cfg guards from `target_os = "linux"` to `any(target_os = "linux", target_os = "freebsd")` is enough to get a working build.

The change is mechanical: 79 files across `app/` and `crates/`, all of them just adjusting the cfg.

One Linux-only carveout: `InputFlags::IUTF8` in `app/src/terminal/local_tty/unix.rs`. nix does not expose that termios input flag on FreeBSD, and the PTY works without it.

Linux, macOS, and Windows are unaffected by construction: every widened cfg already evaluated true on Linux and continues to; neither macOS nor Windows ever matched these guards.

Tested by building `warp-oss` on FreeBSD 16-CURRENT amd64 with rust 1.92 and launching it under wayland (niri).

image

### DIFF
--- a/app/build.rs
+++ b/app/build.rs
@@ -18,7 +18,7 @@ use warp_util::path::app_target_dir;
 
 fn main() -> Result<()> {
     cfg_aliases! {
-        linux_or_windows: { any(target_os = "linux", windows) },
+        linux_or_windows: { any(any(target_os = "linux", target_os = "freebsd"), windows) },
         enable_crash_recovery: { linux_or_windows },
     }
 
--- a/app/build.rs
+++ b/app/build.rs
@@ -18,7 +18,7 @@ use warp_util::path::app_target_dir;
 
 fn main() -> Result<()> {
     cfg_aliases! {
-        linux_or_windows: { any(any(target_os = "linux", target_os = "freebsd"), windows) },
+        linux_or_windows: { any(target_os = "linux", windows) },
         enable_crash_recovery: { linux_or_windows },
     }
 
--- a/crates/lsp/src/servers/rust.rs
+++ b/crates/lsp/src/servers/rust.rs
@@ -136,45 +136,33 @@ impl LanguageServerCandidate for RustAnalyzerCandidate {
         metadata: LanguageServerMetadata,
         _executor: &CommandBuilder,
     ) -> anyhow::Result<()> {
-        #[cfg(target_os = "freebsd")]
-        {
-            let _ = (&metadata, &self.client);
-            anyhow::bail!(
-                "rust-analyzer is not auto-installable on FreeBSD: upstream \
-                 GitHub releases publish no FreeBSD asset. Install it via \
-                 `rustup component add rust-analyzer` or `pkg install \
-                 rust-analyzer` and warp will pick it up off PATH."
-            );
-        }
-        #[cfg(not(target_os = "freebsd"))]
-        {
-            let asset_kind = AssetKind::from_filename(asset_name()).ok_or_else(|| {
-                anyhow::anyhow!("Unsupported archive format for asset: {}", asset_name())
-            })?;
-            install_from_github(&self.client, &metadata, SERVER_NAME, asset_kind, None).await?;
-            Ok(())
-        }
+        anyhow::ensure!(
+            !cfg!(target_os = "freebsd"),
+            "rust-analyzer is not auto-installable on FreeBSD: upstream \
+             GitHub releases publish no FreeBSD asset. Install it via \
+             `rustup component add rust-analyzer` or `pkg install \
+             rust-analyzer` and warp will pick it up off PATH."
+        );
+        let asset_kind = AssetKind::from_filename(asset_name()).ok_or_else(|| {
+            anyhow::anyhow!("Unsupported archive format for asset: {}", asset_name())
+        })?;
+        install_from_github(&self.client, &metadata, SERVER_NAME, asset_kind, None).await?;
+        Ok(())
     }
 
     async fn fetch_latest_server_metadata(&self) -> anyhow::Result<LanguageServerMetadata> {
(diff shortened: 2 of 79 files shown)


rudrabhoj:
I have signed the Contributor License Agreement and fixed problems cited by @oz-agent kindly review again.

app/build.rs
```diff
@@ -18,7 +18,7 @@ use warp_util::path::app_target_dir;
 
 fn main() -> Result<()> {
     cfg_aliases! {
-        linux_or_windows: { any(target_os = "linux", windows) },
+        linux_or_windows: { any(any(target_os = "linux", target_os = "freebsd"), windows) },
```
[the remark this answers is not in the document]
rudrabhoj: Yes, that's understandable. minidump crate is a project external to warp and it doesn't support FreeBSD yet, it should be fine. I'll soon submit PR to minidump too until then I think this shouldn't be a blocker for a FreeBSD port of warp.

crates/warpui_core/src/clipboard_utils.rs
```diff
@@ -1,7 +1,7 @@
 #[allow(unused_imports)]
 use crate::clipboard::{Clipboard, ClipboardContent};
 
-#[cfg(any(target_os = "linux", target_os = "windows"))]
+#[cfg(any(any(target_os = "linux", target_os = "freebsd"), target_os = "windows"))]
```
vorporeal: mind cleaning up the nested `any`s where they show up in this PR?

crates/lsp/src/servers/rust.rs
```diff
@@ -136,21 +136,45 @@ impl LanguageServerCandidate for RustAnalyzerCandidate {
         metadata: LanguageServerMetadata,
         _executor: &CommandBuilder,
     ) -> anyhow::Result<()> {
-        let asset_kind = AssetKind::from_filename(asset_name()).ok_or_else(|| {
-            anyhow::anyhow!("Unsupported archive format for asset: {}", asset_name())
-        })?;
-        install_from_github(&self.client, &metadata, SERVER_NAME, asset_kind, None).await?;
-        Ok(())
+        #[cfg(target_os = "freebsd")]
```
vorporeal: nit: for both of the guards in this file, i'd suggest we do something like:

```rust
anyhow::ensure!(
  !cfg!(target_os = "freebsd"),
  "<error message here>
);
```

this avoids the need for the dual `#[cfg(...)]` blocks and the `let _ =` to appease clippy

vorporeal (approved):
hey @rudrabhoj - this is an awesome PR!

left a couple nits but overall LGTM; will run CI and merge after you address these! :)

crates/warpui_core/src/clipboard_utils.rs
rudrabhoj: Fixed!

rudrabhoj:
@vorporeal Implemented both of your suggestions, I also rebased the PR branch.

vorporeal (approved):
lgtm!

vorporeal:
happy to accept this change - i don't expect us to start building FreeBSD binaries or running CI on FreeBSD (to ensure there are no regressions here), but no harm in merging these changes. :)

aokblast:
> @vorporeal Implemented both of your suggestions, I also rebased the PR branch.

Hi, thanks for your work. 
It fails to build on FreeBSD current with yesterday main branch.

The error is as following:
  /tmp/ta555-0/gettext/libtextstyle/lib/obstack.c:351:31: error: incompatible function pointer types initializing 'void (*)(void) __attribute__((noreturn))' with an expression of type 'void (void)' [-Wincompatible-function-pointer-types]
    351 | __attribute_noreturn__ void (*obstack_alloc_failed_handler) (void)
        |                               ^
    352 |   = print_and_abort;
        |     ~~~~~~~~~~~~~~~

rudrabhoj:
> > @vorporeal Implemented both of your suggestions, I also rebased the PR branch.
> 
> Hi, thanks for your work. It fails to build on FreeBSD current with yesterday main branch.
> 
> The error is as following: /tmp/ta555-0/gettext/libtextstyle/lib/obstack.c:351:31: error: incompatible function pointer types initializing 'void (*)(void) **attribute**((noreturn))' with an expression of type 'void (void)' [-Wincompatible-function-pointer-types] 351 | **attribute_noreturn** void (*obstack_alloc_failed_handler) (void) | ^ 352 | = print_and_abort; | ~~~~~~~~~~~~~~~

You need to use your system gettext, compile it like this:
`env GETTEXT_DIR=/usr/local GETTEXT_LIB_DIR=/usr/local/lib GETTEXT_INCLUDE_DIR=/usr/local/include cargo build --release`

Oh make sure you have gettext, if you don't, you'd need to do something like this:
`pkg install gettext`

You should then get something like this:
image

I executed warp like this:
`WARP_ENABLE_WAYLAND=1 ./target/release/warp-oss`

If you aren't on wayland don't use that variable^

rudrabhoj:
> happy to accept this change - i don't expect us to start building FreeBSD binaries or running CI on FreeBSD (to ensure there are no regressions here), but no harm in merging these changes. :)

Yes totally understandable, wouldn't want to complicate your CI haha. Incorporating this PR already will go long way in letting FreeBSD community maintain a port painlessly.

vorporeal:
can you run `cargo fmt` to fix the formatting CI failures?

rudrabhoj:
> can you run `cargo fmt` to fix the formatting CI failures?

Done. Pushed the post cargo fmt changes.

PR #9371, opened 2026-04-29
Hide macOS windows instead of closing sessions

creeveliu:
## Description

Make macOS window close actions hide the window instead of closing the underlying Warp session. This updates Cmd-W / Close Window bindings, the explicit close window actions, and the native macOS window close callback so clicking the red close button also hides the window. Reopening Warp from the Dock restores an existing hidden window before creating a new one.

## Testing

- cargo fmt
- cargo test -p warp cmd_w --lib
- cargo clippy -p warp --lib --tests -- -D warnings
- TERM=xterm-256color NO_COLOR=1 ./script/run --dont-open
- Manual local verification with /Users/cl/Projects/warp/target/debug/bundle/osx/WarpOss.app: Cmd-W and the red close button hide the window; clicking the Dock icon restores the session

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -1912,19 +1912,6 @@ fn app_callbacks(is_integration_test: bool) -> warpui::platform::AppCallbacks {
             crash_reporting::uninit_sentry();
         })),
         on_should_close_window: Some(Box::new(move |window_id, ctx| {
-            let general_settings = GeneralSettings::as_ref(ctx);
-            // On Linux or Windows, if we're about to close the final window, we should quit the app instead.
-            // On Mac, we do this conditionally based on a user setting.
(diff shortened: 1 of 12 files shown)


captainsafia:
@creeveliu Thanks for taking the time to open this PR! Is there an issue associated with this PR? If not, can you create one so we can flesh out what problem you're trying to solve?

creeveliu:
> @creeveliu Thanks for taking the time to open this PR! Is there an issue associated with this PR? If not, can you create one so we can flesh out what problem you're trying to solve?

It's a new feature/UX improvement, not a Bug fix

creeveliu:
Created an issue to capture the feature request and desired behavior: #10237

This PR is the implementation for that UX improvement.

Issue #9083, opened 2026-04-15
Expose Tab Configs via URI scheme / CLI for programmatic launch


mydandyandy:
## Summary
Tab Configs are a great addition — they auto-execute commands on open (a major improvement over legacy launch_configurations YAML). But they're **UI-only**: accessible only via the `+` menu. There's no CLI flag, URI scheme, or IPC for programmatically opening a specific Tab Config.

## Request
Expose Tab Configs via a programmatic trigger. Two reasonable options:

1. **URI scheme:** `warp://action/open_tab_config?name=my-config`
2. **CLI flag:** `warp.exe --tab-config my-config` (or similar; the existing `warp.exe` binary is the Oz CLI today, so maybe a separate terminal-launch flag)

Either would suffice.

## Why this matters
The command-execution primitive already exists in Tab Configs — commands run automatically when a tab opens via the `+` menu. The only thing missing is a way to trigger that open from outside the UI.

Related: [#3959](https://github.com/warpdotdev/Warp/issues/3959) (open tab and execute a command from CLI) has been open since 2023 and is the broader ask. This feature request is a tighter, more achievable ask since the plumbing for command execution is now in place — you just need to expose an external trigger.

## Use case
Multi-session workflows where I want to programmatically launch terminal sessions with specific configurations (working directory, pre-run commands, pane splits) from a script or shortcut. Today I'm using `cmd.exe` + `schtasks /it` to open cmd windows, which isn't integrated with Warp at all. If I could trigger Tab Configs from a URI or CLI, I could keep everything in Warp.

## Tested
Confirmed the following **do NOT work** in v0.2026.04.08.08.36:
- `warp://action/open_tab_config?name=X` — no-op
- `warp://action/new_tab?config=X` — opens a blank tab, ignores the parameter
- `warp://action/split_pane?direction=right` — no-op
- `warp.exe` CLI: only exposes Oz cloud agent commands, no terminal/tab management

Only `warp://action/new_tab` (no parameters) and `warp://action/new_tab?path=X` are functional.

---
Session: Dozer (5d3a) | Task: 6g3fqhCpRWVvMFJG

zachbai:
This is a great idea, would like to get this on the roadmap. Will track this

cc @moirahuang

tisonkun:
With source code I notice that `open -a Warp "warp://action/new_tab?path=/path/project/script"` actually open the working dir and execute the script.

So a workaround is:

```
[...]

PR #9379, opened 2026-04-29
Add warp://tab_config/<name> deeplink

### LINKED ISSUE #9060: Support `warp://tab_config/<tab_configuration_path>`
### Pre-submit Checks

Warp should support starting it with a `warp://tab_config/<tab_configuration_path>` URL similar to `warp://launch/<launch_configuration_path>` ([docs](https://docs.warp.dev/terminal/more-features/uri-scheme)), which would open up a new tab in the existing Warp window with the specified tab configuration.

I'd like to open a new tab in Warp programmatically with a specific working directory and start command. Currently I use launch configurations for that and start Warp with `warp://launch/<launch_configuration_path>`. However this creates a new window instead of a new tab which becomes messy pretty quickly and is deprecated now.

### Additional context

_No response_

### Operating system (OS)

macOS

5 (Can't work without it!)

None

haha1903:
Closes #9060
Related to #9083

## Description

Add a `warp://tab_config/<name>` URL scheme handler so users can launch a saved tab config straight from a deeplink (alongside the existing `warp://launch/<name>`).

By default the tab config opens as a **new tab in the active window**. Pass `?new_window=true` (or invoke when no Warp window is open) to open in a brand-new window instead.

`<name>` is matched case-insensitively against the tab config's file stem, so both `warp://tab_config/my_tab` and `warp://tab_config/my_tab.toml` resolve to `my_tab.toml`. We key off the filename (not the `name` field inside the file) because multiple tab configs can share the same `name`.

The handler dispatches to the existing `Workspace::open_tab_config`, so the params modal and worktree branch generation flows are reused unchanged.

Gated on the existing `FeatureFlag::TabConfigs`.

## Why

Lets users (and external tools / scripts / browser bookmarks) jump straight into a configured pane layout without going through the `+` menu. Mirrors the long-standing `warp://launch/...` deeplink for launch configs.

## How

- New `UriHost::TabConfig` variant in `app/src/uri/mod.rs`, parsed from `tab_config` host string and listed in `validate_custom_uri`'s allowlist.
- New handler `handle_tab_config_uri` that resolves the config and either reuses the active window's workspace or opens a new window (via `open_new_window_get_handles`), then calls `workspace.open_tab_config(...)`.
- `Workspace::open_tab_config` promoted from private to `pub(crate)` so it can be invoked from the URI handler.
- Added a wasm-side stub for `load_tab_configs` so the cross-platform `pub(crate) use imp::load_tab_configs;` in `user_config/mod.rs` matches the pattern already used by `load_launch_configs` (no `cfg` gate needed).

## Testing

Manual:

- `open "warposs://tab_config/<name>"` opens the config as a new tab in the focused window.
- `open "warposs://tab_config/<name>?new_window=true"` opens it in a new window.
- Tab config with `params` triggers the existing params modal.
- Worktree-style tab configs still get an autogenerated branch name.
- Unknown `<name>` logs a warning and is a no-op.

`cargo fmt` and `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings` both pass locally. Full `./script/presubmit` will run on CI.

## Server API dependencies

No server changes.

## Agent Mode

CHANGELOG-IMPROVEMENT: Added a `warp://tab_config/<name>` deeplink that opens a saved tab config in the active window (or a new one with `?new_window=true`).


haha1903:
@cla-bot check

app/src/uri/mod.rs
```diff
@@ -472,6 +478,8 @@ impl UriHost {
             Self::Codex => W::default(),
             // Linear deeplink opens a new tab with agent view
             Self::Linear => W::default(),
+            // Tab config deeplink prefers existing window unless ?new_window=true.
+            Self::TabConfig => W::default(),
```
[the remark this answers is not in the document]
haha1903: I have fixed this problem.

haha1903:
Found matching issue #9060 (`Support warp://tab_config/<tab_configuration_path>`) — linked it via `Closes #9060` in the PR body and left a comment there requesting `ready-to-implement`. #9083 covers the same ground from a broader angle and is referenced as `Related to`. Happy to adjust the URL path (`tabconfig` vs `tab_config`) or any other naming to match whatever the team prefers.

app/src/uri/mod.rs
```diff
@@ -10,7 +10,7 @@ use crate::ai::agent::api::ServerConversationToken;
 use crate::drive::OpenWarpDriveObjectSettings;
 use crate::launch_configs::launch_config::LaunchConfig;
 use crate::linear::{LinearAction, LinearIssueWork};
-use crate::root_view::{open_new_window_get_handles, OpenLaunchConfigArg};
+use crate::root_view::{open_new_window_get_handles, workspace_for_window, OpenLaunchConfigArg};
```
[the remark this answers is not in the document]
haha1903: This isn't accurate — `workspace_for_window` is already defined as `pub fn` at `app/src/root_view.rs:550` (pre-existing, not added by this PR; it's also used within `root_view.rs` at lines 633 and 733). The `use crate::root_view::{..., workspace_for_window, ...}` import in `uri/mod.rs` resolves fine.

Verified locally: `cargo check -p warp --lib` passes on this branch.

haha1903:
@seemeroland when you have a minute — could I get a look from you on this one? It mirrors the existing `warp://launch/...` flow for tab configs. Oz flagged a [CRITICAL compile error](https://github.com/warpdotdev/warp/pull/9379#discussion_r3177547066) on the `workspace_for_window` import, but that function is already `pub fn` at `root_view.rs:550` (and `cargo check -p warp --lib` passes locally), so I pushed back in-thread. Thanks!

app/src/uri/mod.rs
```diff
@@ -657,6 +665,84 @@ fn find_matching_config_name<'a>(
         .find(|&config| config.name.to_lowercase() == target_name_lower)
 }
 
+/// Handles `warp://tabconfig/<name>` deeplinks.
+///
+/// Resolution rules:
+/// - `<name>` is matched case-insensitively against each tab config's `name`
```
moirahuang: It'd be better for us to use the file name, since we allow for multiple tab configs w the same name

app/src/uri/mod.rs
```diff
@@ -99,6 +101,7 @@ impl FromStr for UriHost {
             "mcp" => Ok(Self::Mcp),
             "codex" => Ok(Self::Codex),
             "linear" => Ok(Self::Linear),
+            "tabconfig" if FeatureFlag::TabConfigs.is_enabled() => Ok(Self::TabConfig),
```
moirahuang: Let's use `tab_config` instead!

app/src/uri/mod.rs
```diff
@@ -657,6 +665,76 @@ fn find_matching_config_name<'a>(
...
+    workspace.update(ctx, |workspace, ctx| {
+        workspace.open_tab_config(config, ctx);
+    });
+}
+
+/// Case-insensitive match against each tab config's file stem. Tab config
+/// `name` fields are not unique across files, so we key off the filename.
+fn find_matching_tab_config<'a>(
+    target: &str,
+    configs: &'a [crate::tab_configs::TabConfig],
```
moirahuang: we can update this and the line below to:

1. import at the top instead of the inline imports 
2. eliminate the clone at the callsite and the 'a lifetime by having the function consume the `Vec` and return an owned `TabConfig`

app/src/uri/mod.rs
```diff
@@ -657,6 +665,76 @@ fn find_matching_config_name<'a>(
...
+/// Handles `warp://tab_config/<name>` deeplinks.
+///
+/// Resolution rules:
+/// - `<name>` is matched case-insensitively against each tab config's file
+///   stem, so both `warp://tab_config/my_tab` and
+///   `warp://tab_config/my_tab.toml` work.
+/// - When `?new_window=true` (or no Warp window is open) the tab config opens
+///   in a brand-new window. Otherwise it opens as a new tab in the active
+///   window.
+fn handle_tab_config_uri(primary_window_id: Option<WindowId>, url: &Url, ctx: &mut AppContext) {
```
moirahuang, line 733: let's add some unit tests please!

moirahuang (commented):
thanks for making the changes! just a few small comments now

moirahuang:
@haha1903 I just checked out the latest code locally and it's not compiling. Mind making sure it compiles + testing locally please?

haha1903:
@moirahuang apologies for the broken merge — that was my mistake. Root cause: during the master merge I staged the conflict resolution for the `use` hunk, then edited three more call sites (`workspace_for_window` → `WorkspaceRegistry::as_ref(ctx).get(...)`, since `workspace_for_window` was removed in #9275) but forgot to `git add` them before committing the merge. The commit went out from a stale index.

Just pushed `92cdef3d` with the missing swap. `cargo check -p warp --lib` and `cargo test ... find_matching_tab_config` both pass locally; waiting on CI to confirm.

app/src/uri/mod.rs
```diff
@@ -712,6 +721,72 @@ fn find_matching_config_name<'a>(
...
+
+    workspace.update(ctx, |workspace, ctx| {
+        workspace.open_tab_config(config, ctx);
+    });
+}
+
+/// Case-insensitive match against each tab config's file stem. Tab config
+/// `name` fields are not unique across files, so we key off the filename.
+fn find_matching_tab_config(target: &str, configs: Vec<TabConfig>) -> Option<TabConfig> {
+    let stem = remove_extension(target).unwrap_or(target).to_lowercase();
```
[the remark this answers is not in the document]
haha1903: Fixed in `2a2a1fbf` — switched to the same two-step pattern `find_matching_config` uses (try the target as-is first, then with the extension stripped). `foo.bar.toml` now resolves via both `warp://tab_config/foo.bar` and `warp://tab_config/foo.bar.toml`; added a test case covering both.

moirahuang:
@haha1903 I just tried to test this locally and it wasn't opening a tab config. Mind recording and attaching you testing this locally please?

haha1903:
> @haha1903 I just tried to test this locally and it wasn't opening a tab config. Mind recording and attaching you testing this locally please?

https://github.com/user-attachments/assets/f48b8ec6-dd04-4dfd-8369-6c5770ffc31a

moirahuang (approved):

haha1903:
Thanks @moirahuang for the review and merge! 🙏

Issue #6877, opened 2025-07-22
Support text-input-method on wayland


fecet:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

If using warp on wayland, it will ignore my ime

### Operating system (OS)

Linux

### How important is this feature to you?

5 (Can't work without it!)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

mars-base:
If you don't need the latest Warp and just want a working terminal, we have a build from the OSS codebase that supports Chinese input properly.

**Download:** https://github.com/mars-base/warp/releases/tag/v0.2026.05.13.3.01

Packages: macOS aarch64 DMG, Linux x86_64 (AppImage/deb/rpm), Linux aarch64 AppImage.

**macOS users:** The DMG is not signed by Apple. After copying to Applications, run:
```bash
sudo xattr -rd com.apple.quarantine /Applications/Warp.app
```
If still blocked: `codesign --remove-signature /Applications/Warp.app`

This build is not the latest version, but Chinese input and CJK rendering work fine in our testing.

PR #9384, opened 2026-04-29
Fix IME not enabled on Linux/Wayland (set_ime_allowed missing from Linux path)

### LINKED ISSUE #9383: Bug: IME (Input Method) never enabled on Linux/Wayland — set_ime_allowed(true) missing from Linux code path
## Summary

On Linux/Wayland, Warp never activates the input method (fcitx5, ibus, etc.), making CJK and other IME-based input completely non-functional. Pressing the IME trigger key (e.g. `Ctrl+Space`) has no effect inside Warp.

## Root Cause

`winit::Window::set_ime_allowed(true)` is called **only inside a `#[cfg(windows)]` block** in `crates/warpui/src/windowing/winit/window.rs`:

```rust
// line 1363
#[cfg(windows)]
{
    // ...
    window.set_ime_allowed(true);  // ← only reached on Windows
    // ...
}
```

On Linux, this call is never made. Without it, winit never sends `zwp_text_input_v3.enable()` to the Wayland compositor, so the compositor never activates the IM session, and fcitx5/ibus never receives a keyboard grab for the window.

## Steps to Reproduce

1. Run Warp on Linux with a Wayland compositor (tested: Hyprland 0.54.3)
2. Install fcitx5 with Rime or Pinyin input method
3. Focus the Warp terminal input
4. Press `Ctrl+Space` (or your configured IME trigger key)
5. Nothing happens — fcitx5 is not activated

`fcitx5-remote` confirms it stays inactive (`1`) while Warp is focused.

## Expected Behavior

Pressing the IME trigger key activates the input method, allowing preedit and CJK character composition (same as any other Wayland-native terminal).

## Environment

- OS: Arch Linux, kernel 6.19.14
- Compositor: Hyprland 0.54.3 (Wayland)
- Input method: fcitx5 5.x with `libwaylandim.so` (text-input-v3 support confirmed)
- Warp: built from source (OSS)

## Fix

Add a `#[cfg(target_os = "linux")]` block after window creation that calls `set_ime_allowed(true)`:

```rust
#[cfg(target_os = "linux")]
if let Ok(window) = created_window.as_ref() {
    window.set_ime_allowed(true);
}
```

This causes winit's Wayland backend to call `zwp_text_input_v3.enable()`, which notifies the compositor to activate the IM session for the window. Verified fix: fcitx5 activates on `Ctrl+Space` and CJK input works correctly after the change.

Related: #6877

leozeli:
## Description

Fixes #9383 (related: #6877)

`winit::Window::set_ime_allowed(true)` was wrapped inside a `#[cfg(windows)]` block in `create_window()`, so on Linux/Wayland it was never called. Without this call, winit never sends `zwp_text_input_v3.enable()` to the Wayland compositor — meaning the compositor never activates the IM session for the window, and input methods (fcitx5, ibus) never receive a keyboard grab. The result: pressing the IME trigger key (`Ctrl+Space`) has no effect, and CJK/IME input is completely broken on Wayland.

**Fix:** add an equivalent `#[cfg(target_os = "linux")]` block immediately after the `#[cfg(windows)]` block, calling `set_ime_allowed(true)` when window creation succeeds.

```rust
#[cfg(target_os = "linux")]
if let Ok(window) = created_window.as_ref() {
    window.set_ime_allowed(true);
}
```

## Testing

- Tested on Arch Linux, Hyprland 0.54.3 (Wayland), fcitx5 + Rime
- Before fix: `fcitx5-remote` stays at state `1` (inactive) with Warp focused; `Ctrl+Space` has no effect
- After fix: `fcitx5-remote` transitions to state `2` (active); Rime preedit and CJK commit work correctly in the terminal input
- `cargo check -p warpui` passes; `cargo build --bin warp-oss` succeeds

No new automated tests added — the IME activation path is a Wayland protocol side-effect that requires a live compositor to exercise.

CHANGELOG-BUG-FIX: Fixed IME (fcitx5, ibus) not activating on Linux/Wayland, restoring CJK and other input-method-based input


leozeli:
@cla-bot check

PR #9373, opened 2026-04-29
Add Nix flake for Warp

VitalyAnkh:
## Description
Adds a first-party Nix flake for Linux users.

What changed:
- Adds `flake.nix`/`flake.lock` with `packages.default` and `packages.warp-terminal` for `x86_64-linux` and `aarch64-linux`.
- Builds the OSS Warp Linux binary from source with the repository Rust toolchain and the `release_bundle`, `gui`, and `nld_improvements` features.
- Installs bundled resources, generated settings schema, generated third-party license attribution, desktop entry, icons, and licenses into the package output.
- Wraps the executable with the runtime libraries needed on Nix systems.
- Adds a matching `devShell` for Nix-based Rust development.

## Testing
- `nix flake show --all-systems --no-write-lock-file`
- `nix flake check --all-systems --no-build --print-build-logs`
- `nix build .#warp-terminal --print-build-logs`
- `result/bin/warp-terminal --help`
- GUI smoke test with isolated HOME/XDG dirs: `timeout 20s result/bin/warp-terminal` stayed running until timeout and created Warp config/cache files.
- Verified installed bundled resources: 123 files under `result/opt/warpdotdev/warp-terminal/resources/bundled`.
- Verified generated `THIRD_PARTY_LICENSES.txt` is installed under both `opt/warpdotdev/warp-terminal/resources` and `share/licenses/warp-terminal`, with matching contents.

## Server API dependencies
No server API dependencies.

## Agent Mode
- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

## Changelog Entries for Stable
CHANGELOG-IMPROVEMENT: Added first-party Nix flake support for Linux users to build Warp from source.

Co-Authored-By: Warp <agent@warp.dev>

PR #9394, opened 2026-04-29
Add Nix flake for Warp

### LINKED ISSUE #9296: Provide a flake.nix for Nix users
### Pre-submit Checks

Please provide a flake.nix so that we can easily add it to the inputs of our NixOS configs.

Warp is not packaged for Nix which means we can't use it on NixOS (or generic Linux or macOS for the ones installing via Nix) 

### Additional context

_No response_

### Operating system (OS)

Linux

5 (Can't work without it!)

None

VitalyAnkh:
## Follow-up context
This PR supersedes #9373. The earlier PR was automatically closed by the issue-readiness workflow after it was explicitly associated with the related issue before that issue had the `ready-to-implement` label. This PR carries the same Nix flake work forward and includes follow-up review fixes around bundled resources, settings schema generation, license attribution, Linux-only scope, and experimental package naming.

## Description
Adds experimental first-party Nix flake support for Linux source builds of the OSS Warp binary.

Fixes #9296

## Platform scope
This flake currently supports `x86_64-linux` and `aarch64-linux` only. macOS support is not included in this PR and remains WIP.

What changed:
- Adds `flake.nix`/`flake.lock` with the package output `warp-terminal-experimental` for supported Linux systems.
- Builds the OSS Warp binary from source with the repository Rust toolchain and the `release_bundle`, `gui`, and `nld_improvements` features.
- Derives the main Cargo vendor directory from `Cargo.lock` with `craneLib.vendorCargoDeps`, so the app build no longer has a top-level `cargoVendorHash` to manually maintain when `Cargo.lock` changes.
- Installs bundled resources, generated stable-channel settings schema, generated third-party license attribution, desktop entry, icons, and licenses into the package output.
- Uses the stable release channel when preparing bundled resources and settings metadata so the public Nix package does not include dogfood/preview-gated resources.
- Keeps `warp-proto-apis` and `workflows` pinned through flake inputs, with exact revisions recorded in `flake.lock`.
- Applies the vendored git checkout patches needed for `warp-proto-apis` and `workflows` during Cargo vendoring.
- Wraps the executable with the runtime libraries needed on Nix systems.
- Adds a matching `devShell` for Nix-based Rust development.

## Testing
- `git diff --check -- flake.nix flake.lock`
- `nix fmt --accept-flake-config -- --check flake.nix`
- `nix eval --json --accept-flake-config .#packages`
- `nix build .#warp-terminal-experimental --accept-flake-config --print-out-paths --print-build-logs`
- Isolated XDG/HOME smoke launch of `/nix/store/9xav7pd562rhanqjyyp4qp5h5vp71m2k-warp-terminal-experimental-0.1.0+dirty/bin/warp-terminal-experimental`

Smoke result: the process stayed up until the 15-second timeout and wrote `warp-oss.log` entries for X11 app startup, OSS channel startup, ONNX classifier load, Vulkan GPU renderer selection, and an active window.

## Server API dependencies
No server API dependencies.

## Agent Mode

CHANGELOG-IMPROVEMENT: Added experimental Linux-only Nix flake support for building OSS Warp from source.


VitalyAnkh:
Context for maintainers: this PR supersedes #9373.

#9373 was the original submission for this Nix flake work, but it was automatically closed by the issue-readiness workflow after being explicitly associated with the related issue before that issue had the `ready-to-implement` label. I opened this follow-up PR to keep the contribution reviewable without using a closing keyword, and then carried over the fixes requested on the earlier thread:

- generated third-party license attribution is installed into the package output;
- bundled resources and `settings_schema.json` are generated using the stable public channel, so dogfood/preview-gated resources are not included;
- packaging now only patches shebangs for the resource helper scripts used during `postInstall`.

Oz has approved this PR after those fixes.

captainsafia:
@VitalyAnkh Thanks for addressing all the rounds of feedback from Oz here! I've gone ahead and marked https://github.com/warpdotdev/warp/issues/9296 as ready-to-implement so that we can link it with this issue.

@acarl005 Can I tap you to review this?

Eveeifyeve:
Question is does this actually build warp-terminal from source or does this build the schema? Because building warp from source requires more work including compiling the metalib seperately as xcrun doesn't actually allow compiling metal files to air. I have a feeling this is AI generated slop.

VitalyAnkh:
@Eveeifyeve What's the metalib? With this PR, I run:

```sh
nix build .#warp-terminal --print-out-paths --print-build-logs 
result/bin/warp-terminal --help
```
The compiled warp-termimal runs normally. You can run the command and verify the result.

> I have a feeling this is AI generated slop.

I utilize AI but this is not AI generated slop.

VitalyAnkh:
@Eveeifyeve I rebuilt this locally. This PR builds the real Linux `warp-oss` app binary; it is not only building the schema generator.

You can verify by yourself. Evidence from `nix build .#warp-terminal --print-out-paths --print-build-logs`:

- The flake currently targets Linux only: [`x86_64-linux` and `aarch64-linux`](https://github.com/warpdotdev/warp/blob/b742d4e79ffb2e7a8a5e50d8a1d7512d9bd8d052/flake.nix#L20-L23).
- Nix invokes Cargo with both the app binary and schema generator: [`-p warp --bin warp-oss --bin generate_settings_schema`](https://github.com/warpdotdev/warp/blob/b742d4e79ffb2e7a8a5e50d8a1d7512d9bd8d052/flake.nix#L145-L152). The schema generator is built because `settings_schema.json` is packaged as a resource; it is not the only target.
- The build log compiles `warpui`, `warp_core`, `warp_terminal`, and `warp v0.1.0 (/build/source/app)`, then finishes the optimized release profile.
- The package installs `$out/bin/warp-oss` into the app directory and exposes `bin/warp-oss` plus `bin/warp-terminal` wrapper symlinks: see the install/wrap step in [`flake.nix`](https://github.com/warpdotdev/warp/blob/b742d4e79ffb2e7a8a5e50d8a1d7512d9bd8d052/flake.nix#L180-L223).
- The final output contains `opt/warpdotdev/warp-terminal/.warp-oss-wrapped`; `file` reports it as an ELF x86-64 executable. Running `result/bin/warp-terminal` with an isolated `HOME` creates `warp-oss.log`, initializes app services, loads bundled skills, selects a GPU backend, and creates/saves app window state.

Repro:

```sh
nix build .#warp-terminal --print-out-paths --print-build-logs
file result/opt/warpdotdev/warp-terminal/.warp-oss-wrapped

tmp_home="$(mktemp -d)"
mkdir -p "$tmp_home/.config" "$tmp_home/.local/state"
HOME="$tmp_home" \
  XDG_CONFIG_HOME="$tmp_home/.config" \
  XDG_STATE_HOME="$tmp_home/.local/state" \
  timeout 20s result/bin/warp-terminal || true

tail -n 80 "$tmp_home/.local/state/warp-oss/warp-oss.log"
```

About `metalib`: that is macOS-specific. `crates/warpui/build.rs` only calls `compile_metal_shaders()` when `CARGO_CFG_TARGET_OS == "macos"`, and the `xcrun metal` / `xcrun metallib` calls live under that path: [`crates/warpui/build.rs`](https://github.com/warpdotdev/warp/blob/b742d4e79ffb2e7a8a5e50d8a1d7512d9bd8d052/crates/warpui/build.rs#L22-L107). `app/build.rs` also only links MetalKit inside the macOS branch: [`app/build.rs`](https://github.com/warpdotdev/warp/blob/b742d4e79ffb2e7a8a5e50d8a1d7512d9bd8d052/app/build.rs#L34-L47). So `metalib` is not expected to run for this Linux flake build.

I clarified the PR description as well: this PR currently supports Linux only, and macOS support is still separate WIP.

Please verify the build before calling a contribution "AI generated slop". I do use AI tooling, but this is a reproducible working build. An unverified accusation like that is human generated slop, not useful review feedback.

acarl005:
@VitalyAnkh Thank you for pushing this forward. The hold-up on my end here is auto-update. It's not going to work when building the package this way. Arguably that makes sense with NixOS's philosophy. I need to step back and think about how the app should handle this.

flake.nix
```diff
@@ -0,0 +1,313 @@
...
+          };
+          lib = pkgs.lib;
+          rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+          rustPlatform = pkgs.makeRustPlatform {
+            cargo = rustToolchain;
+            rustc = rustToolchain;
+          };
+
+          appCargoToml = builtins.fromTOML (builtins.readFile ./app/Cargo.toml);
+          version = "${appCargoToml.package.version}+${self.shortRev or "dirty"}";
```
acarl005, line 44: Unfortunately, we don't ever actually increment `package.version` in our `Cargo.toml`. The value `0.1.0` is just a static placeholder. The _actual_ version is embedded by setting the env var `GIT_RELEASE_TAG` during compilation. This is unset in this flake. I think that is fine actually. It just means places like Settings > About will show a placeholder value for the version. 

My concern, however, is that upgrading might not work as expected, e.g. I don't think `nix profile upgrade` will work as expected. That isn't _necessarily_ blocking approval as long as there is some reasonable way to upgrade. Will `nix flake update warp` work for that?

flake.nix
```diff
@@ -0,0 +1,313 @@
...
+          cargoVendorHash = "sha256-TzYSC82HVRhCxBHLmHw8BIZ4hJKCZfp+s/mfbeAjdQ4=";
```
acarl005: Is this hash going to remain valid? If I understand correctly, users could point to any revision of this flake right? The cargo dependencies (and hence the actual hash) will change as the master branch gets new commits.

flake.nix
```diff
@@ -0,0 +1,313 @@
...
+          warpProtoApis = pkgs.fetchFromGitHub {
+            owner = "warpdotdev";
+            repo = "warp-proto-apis";
+            rev = "78a78f21a75432bf0141e396fb318bf1694e47f0";
```
acarl005: Similar question to above. `warp-proto-apis` is our protobuf API definitions for talking to warp-server. If the user is tracking the master branch and this is hard-coded to a specific revision, the protobuf APIs could go stale, i.e. out of sync with the actual server.

acarl005 (commented):
I haven't used NixOS too much and I haven't really used flakes, so I'll need your help understanding some of the implications of this PR. I've put some questions in this review.

flake.nix
acarl005: To explain our versioning a bit further, the source-of-truth for our versioning comes from here: https://releases.warp.dev/channel_versions.json

Those versions point to tags [here](https://github.com/warpdotdev/warp/releases).

I'm not sure if fetching version numbers over the network meshes well with the Nix philosophy of reproducibility. What do you think?

VitalyAnkh: Yes. For an unlocked flake ref, e.g. `github:warpdotdev/warp#warp-terminal`, `nix profile upgrade warp-terminal` upgrades by re-fetching and re-evaluating the latest flake. That does not depend on `Cargo.toml` changing from `0.1.0`.

For users consuming Warp as an input in their own flake, `nix flake update warp` updates the Warp source input. The package version includes `self.shortRev`, so a new source revision still produces a distinct derivation/output even though Cargo’s package version is static.

If someone installs from a fully locked commit ref, `nix profile upgrade` will intentionally not move it. That is expected Nix behavior for locked refs.

I agree with leaving `GIT_RELEASE_TAG` unset for this source-build flake: the Nix package manager should own upgrades, not Warp’s app-level autoupdater.

flake.nix
VitalyAnkh: Yes, the Cargo vendor hash is supposed to be fixed-output for one exact source revision / `Cargo.lock` snapshot. It should remain valid for that specific input set, and it will fail loudly if the vendored Cargo graph changes.

That is the intended Nix behavior: when dependencies change, the hash is updated in the same change that updates the dependency graph, rather than drifting silently.

flake.nix
VitalyAnkh: I updated this in the latest push. `warp-proto-apis` is now a normal `flake = false` input, with the exact revision and nar hash recorded in `flake.lock` instead of being hidden as a `fetchFromGitHub` inside the package expression.

I did the same for `workflows`, since it had the same stale-input shape. This keeps the build reproducible while making updates visible through the normal flake lockfile workflow, e.g. `nix flake update warpProtoApis` or a broader `nix flake update`.

flake.nix
VitalyAnkh: I do not think fetching `channel_versions.json` during the Nix build is a good fit. It would make the build depend on mutable network state, which works against Nix reproducibility.

For this source-build flake, I think the better behavior is to leave `GIT_RELEASE_TAG` unset and let Nix own upgrades. If we later want a release-tagged Nix package, the cleaner design would be to build from an explicit Warp release tag and set `GIT_RELEASE_TAG` from that declared input, not resolve the latest version dynamically during evaluation/build.

flake.nix
VitalyAnkh: Thanks for raising this. I checked this against the Nix model, and I do not think `cargoVendorHash` belongs in `flake.lock`.

`flake.lock` is for locking flake inputs. `cargoVendorHash` is the fixed-output hash that `rustPlatform.fetchCargoVendor` uses to vendor Cargo dependencies, so it needs to stay in the Nix expression. When Cargo deps change, Nix intentionally makes that hash go stale so we notice and update it.

The source pins in this flake (`warp-proto-apis`, `workflows`) are a different case: those are flake inputs, so `flake.lock` is the right place for them.

References:
- https://releases.nixos.org/nix/nix-2.13.5/manual/command-ref/new-cli/nix3-flake.html
- https://releases.nixos.org/nix/nix-2.13.6/manual/command-ref/new-cli/nix3-flake-lock.html
- https://nixos.org/manual/nixpkgs/stable/
- https://releases.nixos.org/nix/nix-2.28.2/manual/language/advanced-attributes.html

If we want the hash easier to update, we can move it into a separate Nix file and import it, but I do not think `flake.lock` is the right mechanism here.

acarl005: Ok so this hash is valid for this revision, and only this revision. Can you explain how this will affect the developer workflow? Suppose I am working on Warp and I change a dependency causing the `Cargo.lock` to change. Do I need to manually update this hash? It's important to us that we do not impose too much tax on developers for this flake. Most people working on Warp will not know anything about Nix.

What do you propose here? Is there a way the hash can be updated in a fully automated way?

flake.nix
```diff
@@ -0,0 +1,511 @@
...
+            # Point the build script at the pinned source tree fetched by Nix.
+            protoCrate="$(dirname "$(find "$out" -path '*/warp_multi_agent_api-0.0.0/Cargo.toml' -print -quit)")"
+            if [ -z "$protoCrate" ] || [ "$protoCrate" = "." ]; then
+              echo "could not find vendored warp_multi_agent_api crate" >&2
+              exit 1
+            fi
+            substituteInPlace "$protoCrate/build.rs" \
+              --replace-fail \
+                'let proto_path = manifest_dir.parent().unwrap().parent().unwrap();' \
+                'let proto_path = std::path::PathBuf::from("${warpProtoApis}/apis/multi_agent/v1");'
```
acarl005: I really don't want this kind of hack to be necessary. It's brittle. Changing the variable name `proto_path` in that `build.rs` file would break this flake which is very unexpected and cannot be caught in warp-proto-apis' CI process.

Since both warpdotdev/warp-proto-apis and warpdotdev/workflows are both also OSS, I think you should propose changes to their `build.rs`s to take and env var that lets you override these directory locations (and fallback to the current behavior without the override). That is, assuming we can resolve the question above about the hash updates first.

acarl005 (commented):
I see you've added MacOS support which has brought with it a lot of complexity which I wasn't expecting. Before you do any more work, I want to make sure we can address the question I left on my previous review about updating the hash. I see that as a blocking issue and I want to be respectful of your time.

Although I appreciate the desire to support all platforms that Nix supports, I worry that there is a lot of added burden to support MacOS. I'm willing to bet that most Nix users are on Linux (probably NixOS). If we are officially supporting this Flake, I would prefer to do so more incrementally. Lets scope this down to only Linux support for now. I want to make sure to collect feedback on Linux before adding further complexity. I'm also hesitant about adding a CI step for this (that costs money).

Along those lines, I also don't think we should over-promise. Can you add a suffix to the description, something like "(Experimental Nix Support)" that sets the expectations. I'm not too familiar with Nix conventions, but I also wonder if naming the output something different from default is also a good idea here? Would it make sense to name the output something that implies beta/experimental status?

flake.nix
acarl005: Ok that makes sense to me. Thank you

Eveeifyeve:
Also this pr when warp is built from source in nixpkgs a follow up can be done to use the nixpkgs package with an `overrideAttrs` which will reduce a lot of complexity.

flake.nix
VitalyAnkh: Thanks, I pushed an update that scopes this PR back down to the Linux source-build path and makes the experimental naming explicit.

What changed:

- removed the Darwin/macOS build path from the flake
- updated the flake/package description to say Experimental Nix Support, Linux-only
- renamed the package output to `warp-terminal-experimental`
- removed the package-level `default` alias
- removed the `bin/warp-terminal` wrapper; the exposed package entrypoint is now `warp-terminal-experimental`

I re-verified locally with:

```sh
git diff --check
nix fmt --accept-flake-config -- --check flake.nix
nix eval --json --accept-flake-config .#packages
nix build .#warp-terminal-experimental --accept-flake-config --print-out-paths --print-build-logs
```

`nix eval` now exposes only `x86_64-linux.warp-terminal-experimental` and `aarch64-linux.warp-terminal-experimental`; there is no package-level `default` output. The Linux build completed successfully at `/nix/store/0b32qnhvzxyjj1q7nrw0z07qdnqfc7s5-warp-terminal-experimental-0.1.0+dirty`.

On the `build.rs` patching concern: I agree the current `substituteInPlace` approach is not ideal long-term. The cleaner fix is to add explicit env-var overrides in `warpdotdev/warp-proto-apis` and `warpdotdev/workflows`, with fallback to the current relative paths. I did not fold that into this scoped commit because it crosses repository boundaries, but I can take that follow-up if you want this PR to wait on it.

acarl005 (changes requested):
To make the status of this PR clear, it is currently blocked on this:

https://github.com/warpdotdev/warp/pull/9394#discussion_r3186510105

We don't have a way to keep the hash up-to-date without disrupting the developer workflow.

flake.nix
VitalyAnkh: Thanks, this is a fair concern. I changed the approach in `2bfceb9`: the flake no longer uses `rustPlatform.fetchCargoVendor` with a top-level `cargoVendorHash` for the app.

It now uses `craneLib.vendorCargoDeps { inherit src cargoLock; ... }` to derive the main Cargo vendor directory from `Cargo.lock`. If a Warp developer changes dependencies and `Cargo.lock` changes, they do not need to manually recalculate or paste a Nix-specific vendor hash into `flake.nix`. They update `Cargo.lock` as usual, and the Nix vendor derivation is generated from that lockfile. The `crane` entry in `flake.lock` is only the build helper input pin, not Cargo dependency state.

I re-verified this locally:

```sh
git diff --check -- flake.nix flake.lock
nix fmt --accept-flake-config -- --check flake.nix
nix eval --json --accept-flake-config .#packages
nix build .#warp-terminal-experimental --accept-flake-config --print-out-paths --print-build-logs
```

The build completed at `/nix/store/9xav7pd562rhanqjyyp4qp5h5vp71m2k-warp-terminal-experimental-0.1.0+dirty`. The build log still shows the real app build path:

```text
cargoBuildHook flags: ... -p warp --bin warp-oss --bin generate_settings_schema
Compiling warpui ...
Compiling warp_core ...
Compiling warp_terminal ...
Compiling warp v0.1.0 (/build/source/app)
```

I also did a smoke launch with isolated XDG dirs. It ran until the 15s timeout and wrote `warp-oss.log` showing X11 app startup, OSS channel startup, ONNX classifier load, Vulkan GPU renderer selection, and an active window.

This does not fully solve the separate `build.rs` brittleness concern. I moved those patches into `overrideVendorGitCheckout`, keeping them at the vendored git-checkout boundary. The cleaner long-term fix is env-var overrides in `warpdotdev/warp-proto-apis` and `warpdotdev/workflows`, but that is separate from the Cargo vendor-hash blocker addressed here.

VitalyAnkh:
@acarl005 I pushed the Cargo vendoring change that addresses this blocker.

The app-level `cargoVendorHash` is gone. The vendor tree now comes from `Cargo.lock` through `craneLib.vendorCargoDeps`, so dependency updates follow the normal Cargo lockfile flow instead of requiring a hand-updated hash in `flake.nix`.

I left the build and smoke-test evidence on the original review thread: https://github.com/warpdotdev/warp/pull/9394#discussion_r3214400078

acarl005 (changes requested):
We still haven't solved the developer experience problem.

I see that specific revisions of `warp-proto-apis` and `workflows` are still pinned in `flake.lock`. If someone makes changes to the protobuf APIs and bumps the revision in `Cargo.toml`, then the `flake.lock` will point to a stale version potentially breaking the app due to an out-of-date API. The developer would need to know to update `flake.lock`.

We still need a process that doesn't block devs on making changes due to keeping this flake in sync. Any ideas?

VitalyAnkh:
Updated the branch so `warpProtoApis` and `warpWorkflows` are no longer separate `flake.lock` inputs. Their sources now come from the Cargo vendor checkout, which keeps the update path aligned with `Cargo.lock` instead of asking people to remember a second pin file. I also re-ran the local Nix build and smoke check after the change.

The remaining `substituteInPlace` step on the vendored `build.rs` files is still the one bit of coupling left. The cleaner long-term path would be to teach `warp-proto-apis` and `workflows` to accept a path override or crate-local fallback, so the flake can stop patching vendored build scripts entirely.

acarl005 (approved):
Thanks for working through this with me!

acarl005:
I'm testing this locally one last time and once I verify that, I'll merge

PR #9399, opened 2026-04-29
Wake up remote Claude Code agents on new events

kjankov:
## Description

Fixes orchestration v2 parent/child agent wake-up and messaging behavior for remote child agents.

This PR updates the client-side orchestration flow so that incoming parent-agent messages wake remote child agents through the server run follow-up path instead of trying to treat them like local dormant Claude harnesses. Previously, a remote child could receive the parent’s message, but it would not be restarted correctly in a harness and could fail or hang when trying to send a message back to the parent.

Main changes:

•  Adds a remote-child wake path in the blocklist AI controller:
  ◦  detects remote child conversations with pending parent-agent message events
  ◦  submits a run follow-up to agent/runs/{run_id}/followups
  ◦  removes delivered pending message events after successful follow-up submission
  ◦  retries/logs failures instead of silently hanging
•  Keeps local dormant Claude wake behavior separate from remote child wake behavior.
•  Restores remote hidden child panes as cloud/ambient agent panes instead of local terminal-backed child panes.
•  Ensures restored remote child panes enter the existing ambient session in AgentRunning state.
•  Persists and restores remote-child conversation metadata so the client can distinguish local children from remote children across reloads.
•  Improves orchestration v2 message sending:
◦  uses task-scoped server APIs when available
◦  adds bounded timeouts and error logging for send failures
◦  surfaces failures instead of leaving action execution indefinitely pending
•  Adds regression coverage for:
◦  remote child conversation restoration
◦  remote child pane/session state
◦  task-scoped ambient agent messaging
◦  orchestration v2 message/error behavior

## Server API dependencies

## Agent Mode


crates/http_client/src/lib.rs
```diff
@@ -203,16 +203,16 @@ impl Client {
...
     }
 
-    pub fn put<U: IntoUrl + Clone>(&self, url: U) -> RequestBuilder<'_> {
+    pub fn patch<U: IntoUrl + Clone>(&self, url: U) -> RequestBuilder<'_> {
         self.builder(
-            self.wrapped.put(url.clone()),
+            self.wrapped.patch(url.clone()),
             Self::include_warp_http_headers(url),
         )
     }
```
cephalonaut, line 218: put/patch swapped in source order, bodies still correct. Noise diff, revert. (lines 206-218)

app/src/terminal/view/ambient_agent/model.rs
```diff
@@ -420,9 +420,7 @@ impl AmbientAgentViewModel {
         // Store the task ID for later use
         self.task_id = Some(task_id);
 
-        if matches!(self.status, Status::NotAmbientAgent) {
-            self.status = Status::AgentRunning;
-        }
+        self.status = Status::AgentRunning;
```
cephalonaut, line 442: `enter_viewing_existing_session` lost its `NotAmbientAgent` guard — intentional? maybe comment

cephalonaut (commented):
I think the followup based remote child wake ups ought to be driven from the server side. One good reason I think is that ultimately we will support arbitrary subscription to event sources and having the wakeups driven from the event source side seems impractical. Let me know if that doesn't make sense!

Given that, I think the claude wakeup code also ought move. Most of it is harness specific — Claude transcript envelopes, session-index files, `claude --resume` command construction, parent-bridge state staging. You could move it to a `wake_driver.rs` alongside the parent bridge and expose a single `wake_dormant_session` method to the controller. Controller keeps the `EventsReady` subscription, readiness gating, `ExecuteLocalHarnessCommand` emission, and conversation-status update; harness owns eligibility, context fetch, on-disk staging, command construction, and reopen/rollback.

app/src/ai/blocklist/controller.rs
```diff
@@ -1493,32 +1483,200 @@ impl BlocklistAIController {
...
+                        .and_then(|parent_conversation_id| {
+                            BlocklistAIHistoryModel::as_ref(ctx)
+                                .conversation(&parent_conversation_id)
+                        })
+                        .and_then(|parent_conversation| parent_conversation.run_id())
+                }),
+        ))
+    }
+
+    #[cfg(not(target_family = "wasm"))]
```
cephalonaut: **`drain_remote_child_message_events` — drop after master merge**

Becomes dead code once master lands. PR #9273 already excludes remote-child placeholders from SSE, so nothing reaches this queue. Delete the function and its call site in `handle_pending_events_ready`.

app/src/server/server_api/ai.rs
```diff
@@ -1453,6 +1532,14 @@ impl AIClient for ServerApi {
             .await?;
         Ok(response)
     }
+    async fn submit_run_followup(
```
cephalonaut, line 1560: **`submit_run_followup` / `RunFollowupRequest` — dead**

Not called anywhere now that remote-child wake moved server-side. Delete (also the struct at lines 211-213 and the trait method at 834-838).

cephalonaut (commented):
Second-pass review after the wake refactor. Most comments are small (dead code, fragile match, possible regression, asks for in-code documentation, two extraction nits). Two larger architectural notes embedded in the comments on `maybe_prepare_local_claude_wake` and `prepare_local_wake_command`: the wake decision belongs in a harness-side supervisor, and the streamer should not subscribe to dormant Claude-harness conversations at all — which collapses most of the wake's bookkeeping.

app/src/server/server_api/ai.rs
```diff
@@ -2244,9 +2332,12 @@ fn convert_harness(harness: warp_graphql::ai::AgentHarness) -> AIAgentHarness {
         warp_graphql::ai::AgentHarness::Oz => AIAgentHarness::Oz,
         warp_graphql::ai::AgentHarness::ClaudeCode => AIAgentHarness::ClaudeCode,
         warp_graphql::ai::AgentHarness::Gemini => AIAgentHarness::Gemini,
-        warp_graphql::ai::AgentHarness::Other(value) => {
+        other => {
+            if format!("{other:?}") == "Codex" {
```
cephalonaut: **Fragile harness match**

`format!("{other:?}") == "Codex"` silently breaks if the upstream variant is renamed. Use a real variant match against `warp_graphql::ai::AgentHarness`.

app/src/ai/blocklist/controller.rs
```diff
@@ -1202,40 +1223,9 @@ impl BlocklistAIController {
         slash_command: SlashCommandRequest,
         ctx: &mut ModelContext<Self>,
     ) {
-        // Slash commands are a fresh user turn; mirror `send_query`'s
-        // cancel-and-resend so we don't trip `send_request_input`'s in-flight
-        // invariant.
-        if let Some(conversation_id) = slash_command.conversation_id(self, ctx) {
-            self.cancel_active_conversation_for_follow_up(conversation_id, ctx);
-        }
         slash_command.send_request(self, /*is_queued_prompt*/ false, ctx);
```
cephalonaut, line 1243: **Slash-command pre-cancel removal**

The deleted code mirrored `send_query`'s cancel-and-resend so we don't trip `send_request_input`'s in-flight invariant. Was the removal intentional? If so, add a regression test; otherwise restore.

app/src/ai/agent_sdk/ambient.rs
```diff
@@ -50,10 +51,11 @@ use crate::ai::agent_sdk::driver::attachments::{
 use crate::cloud_object::model::persistence::CloudModel;
 use crate::server::ids::{ServerId, SyncId};
 
-use super::common::{EnvironmentChoice, ResolveConfigurationError};
+use super::common::{parse_ambient_task_id, EnvironmentChoice, ResolveConfigurationError};
 
 const MAX_LINE_WIDTH: usize = 90;
 const STREAM_RETRY_BACKOFF_STEPS: &[u64] = &[1, 2, 5, 10];
+const SEND_AGENT_MESSAGE_TIMEOUT: Duration = Duration::from_secs(15);
```
cephalonaut: **`SEND_AGENT_MESSAGE_TIMEOUT` — still needed?**

Was this added because `send_agent_message` used to do followup work synchronously on the server? If server-side followups are async to the send call now, the timeout is probably obsolete. Also asymmetric: only send gets the 15s deadline; `read`/`list`/`mark_delivered` have no equivalent timeout — intentional?

app/src/ai/agent_sdk/driver/harness/claude_code/parent_bridge.rs
```diff
@@ -28,12 +28,15 @@ use crate::ai::agent_events::{
...
 use crate::ai::agent_sdk::driver::{AgentDriver, OZ_MESSAGE_LISTENER_STATE_ROOT_ENV};
+use crate::ai::ambient_agents::AmbientAgentTaskId;
+use crate::server::server_api::ai::AIClient;
 use crate::server::server_api::ai::AgentRunEvent;
 use crate::server::server_api::ServerApi;
 
 const LEGACY_MESSAGE_LISTENER_STATE_ROOT_ENV: &str = "OZ_PARENT_STATE_ROOT";
 const PARENT_BRIDGE_DEFAULT_STATE_ROOT: &str = ".claude-code/oz-parent-bridge";
 const PARENT_BRIDGE_SURFACED_DIR_NAME: &str = "surfaced";
+const PARENT_BRIDGE_EVENT_CURSOR_FILE_NAME: &str = "event-cursor.json";
```
cephalonaut, line 39: **`event-cursor.json` — why local persistence?**

The server already persists the cursor (`update_event_sequence_on_server`) and `stage_parent_bridge_message` is idempotent. What does the local file buy us? Add a comment explaining the recovery scenario it's guarding against.

app/src/ai/blocklist/controller.rs
```diff
@@ -1493,32 +1483,200 @@ impl BlocklistAIController {
...
+    fn drain_remote_child_message_events(
+        &mut self,
+        _conversation_id: AIConversationId,
+        _ctx: &mut ModelContext<Self>,
+    ) -> bool {
+        false
+    }
+
+    #[cfg(not(target_family = "wasm"))]
+    fn pending_event_to_claude_wake_message(event: &PendingEvent) -> Option<ClaudeWakeMessage> {
```
cephalonaut: **`pending_event_to_claude_wake_message` — move to wake_driver**

Pure Claude-specific conversion with no controller state. Belongs in `wake_driver.rs` next to its sibling `From<ClaudeWakeMessage> for MessageBridgeMessageRecord` so all Claude wake types and conversions live in one place.

app/src/ai/blocklist/controller.rs
```diff
@@ -1493,32 +1483,200 @@ impl BlocklistAIController {
...
+    ) -> Option<(AmbientAgentTaskId, Option<PathBuf>, Option<String>)> {
+        let Some(conversation) =
+            BlocklistAIHistoryModel::as_ref(ctx).conversation(&conversation_id)
+        else {
+            log::info!(
+                "Skipping dormant Claude wake candidate: conversation_id={conversation_id:?} reason=conversation_missing"
+            );
+            return None;
+        };
+        if !conversation.is_child_agent_conversation() || conversation.is_remote_child() {
```
cephalonaut: **`local_claude_wake_candidate` — move to wake_driver**

Almost pure derivation. Only controller dependency is reading `active_session.current_working_directory()`. Pass `working_dir: Option<PathBuf>` in as a parameter and move the function into `wake_driver.rs`.

app/src/ai/blocklist/controller.rs
```diff
@@ -1555,6 +1713,182 @@ impl BlocklistAIController {
         }
     }
 
+    #[cfg(not(target_family = "wasm"))]
+    fn maybe_prepare_local_claude_wake(
```
cephalonaut: **`maybe_prepare_local_claude_wake` — wrong home**

The wake decision is harness knowledge dressed as controller logic — the controller infers "dormant Claude child" from `is_remote_child` / `is_child_agent_conversation` / `task_id` / `status`, but the harness already knows it directly. Cleaner: a `WakeSupervisor` in the harness module that owns a registry of dormant conversations and runs a lightweight passive SSE per registered run_id (via the shared `run_agent_event_driver`, but no `persist_cursor` override so it doesn't advance the server cursor). On first event arrival, fire the wake. The bridge takes over the real subscription once Claude boots.

Separately: the trigger should be any pending event, not just messages. Today `peek_pending_message_events` filters to `PendingEventDetail::Message`, so peer lifecycle events to a dormant Claude child fall through to `inject_pending_events_for_request` — which routes to in-process Warp agent infrastructure that doesn't exist for a Claude-harness conversation. Net: peer Errored / Blocked / Failed signals to a dormant Claude child are dropped until a peer message arrives. The harness side already gates on `task.state == Succeeded`, so a broad trigger + harness-side eligibility check is the natural shape.

app/src/ai/agent_sdk/driver/harness/claude_code/wake_driver.rs
```diff
@@ -0,0 +1,230 @@
...
+            _ => CLAUDE_WAKE_PROMPT.to_string(),
+        };
+        Ok(ClaudeWakeRemoteContext {
+            session_id: envelope.uuid,
+            envelope,
+            wake_prompt,
+        })
+    }
+
+    pub(super) async fn prepare_local_wake_command(
```
cephalonaut, line 188: **Two kinds of wake conflated**

Two distinct concerns: (1) cold-start the CLI when a dormant run gets a new event; (2) deliver events to a running CLI. `parent_bridge` already owns (2) end-to-end. `prepare_local_wake_command` is doing (2)-shaped work (`acknowledge_parent_bridge_hook_output` + `stage_parent_bridge_message` per pending message + `prepare_parent_bridge_hook_output`) eagerly, outside the bridge, before the driver boots. That work doesn't belong here — once the bridge boots, it does all of it itself.

The reason the current code shovels content through the streamer's queue is a workaround for cursor collision: the streamer is subscribed to the child's run_id while Claude is dormant and advances the server cursor as events arrive. When the bridge later resumes with `since_sequence = max(local, server)`, those events are already past the cursor and the SSE returns nothing.

Resolution: the streamer should not subscribe to dormant Claude-harness conversations at all. With that change, the server cursor stays at the bridge's last-processed sequence, case (1) collapses to "spawn the driver, write the wake prompt, flip task state, return the command," and case (2) stays entirely in the bridge. `prepare_local_wake_command` loses the staging loop; `wake_dormant_session` loses its `pending_messages` parameter; `pending_event_to_claude_wake_message` and the `ClaudeWakeMessage` type can probably go away. The supervisor restructure (other comment) lands much smaller as a result.

crates/http_client/src/lib.rs
kjankov: Verified this is no longer in the diff; `crates/http_client/src/lib.rs` is untouched by the latest update.

app/src/terminal/view/ambient_agent/model.rs
kjankov: Restored the `NotAmbientAgent` guard in `enter_viewing_existing_session`.

app/src/ai/agent_sdk/driver/harness/codex.rs
```diff
@@ -0,0 +1,447 @@
...
+    #[cfg(unix)]
+    {
+        use std::io::Write as _;
+        use std::os::unix::fs::OpenOptionsExt;
+        let mut file = fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .mode(0o600)
+            .open(path)
```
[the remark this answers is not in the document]
kjankov: Updated the Codex auth write path to set `0o600` permissions after opening/truncating the file, before writing the secret.

app/src/ai/blocklist/controller.rs
```diff
@@ -1493,32 +1483,200 @@ impl BlocklistAIController {
...
+            });
+        if pending_message_events.is_empty() {
+            return false;
+        }
+
+        let pending_message_event_ids = pending_message_events
+            .iter()
+            .map(|event| event.event_id.clone())
+            .collect_vec();
+        let removed_events = OrchestrationEventService::handle(ctx).update(ctx, |svc, _| {
```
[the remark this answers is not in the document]
kjankov: Removed the local remote-child pending-event drain path entirely now that remote-child wakes are server-driven.

app/src/server/server_api/ai.rs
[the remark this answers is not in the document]
kjankov: Added the real generated GraphQL `Codex` variant and switched conversion to an explicit variant match.

app/src/ai/blocklist/controller.rs
kjankov: Deleted `drain_remote_child_message_events` and its call site.

app/src/server/server_api/ai.rs
kjankov: Deleted the dead run follow-up request type, URL builder, client method, and API wrapper.

app/src/server/server_api/ai.rs
kjankov: Replaced the fragile debug-string fallback with an explicit `warp_graphql::ai::AgentHarness::Codex` match.

app/src/ai/blocklist/controller.rs
kjankov: Restored slash-command pre-cancel behavior before submitting the new request.

app/src/ai/agent_sdk/ambient.rs
kjankov: Removed the ambient CLI message-send timeout and its tests; the remaining timeout is in the separate in-app send-message executor path.

app/src/ai/agent_sdk/driver/harness/claude_code/parent_bridge.rs
kjankov: Added a comment describing the local cursor recovery case and how it relates to the durable server cursor.

app/src/ai/blocklist/controller.rs
kjankov: Removed `ClaudeWakeMessage` and the pending-event conversion since wake preparation no longer stages messages.

app/src/ai/blocklist/controller.rs
kjankov: Moved local Claude wake candidate derivation into `wake_driver.rs`; the controller now only passes the working directory and emits the command.

app/src/ai/blocklist/controller.rs
kjankov: Moved wake eligibility into the Claude wake driver and changed the trigger to `has_pending_events`, so lifecycle-only events can wake a dormant local Claude child too.

app/src/ai/agent_sdk/driver/harness/claude_code/wake_driver.rs
kjankov: Separated cold-start from delivery: wake prep no longer stages messages, `ClaudeWakeMessage` is gone, and the generic streamer skips dormant local Claude child SSE so the parent bridge owns delivery after boot.

app/src/ai/blocklist/action_model/execute/send_message.rs
```diff
@@ -1,28 +1,132 @@
...
+    request: SendAgentMessageRequest,
+) -> anyhow::Result<SendAgentMessageResponse, anyhow::Error> {
+    match task_id {
+        Some(task_id) => {
+            server_api
+                .send_agent_message_for_task(&task_id, request)
+                .await
+        }
+        None => ai_client.send_agent_message(request).await,
+    }
```
cephalonaut, line 118: **`send_agent_message_with_timeout` has no timeout on wasm**

Not sure this is really invoked from wasm, but maybe add a comment.

app/src/ai/blocklist/orchestration_event_streamer.rs
```diff
@@ -760,6 +928,131 @@ impl OrchestrationEventStreamer {
...
+                    }
+                    Err(err) => {
+                        log::warn!(
+                            "Dormant Claude wake listener failed for {conversation_id:?} \
+                             (gen={generation}): {err:#}"
+                        );
+                        if me.is_dormant_claude_wake_listener_eligible(conversation_id, ctx) {
+                            me.start_dormant_claude_wake_listener(conversation_id, ctx);
+                        }
+                    }
```
cephalonaut, line 1033: **Wake-listener restart has no backoff**

On `Ok(None)` or `Err` the callback immediately restarts the wake-only listener if still eligible, with no attempt counter at this layer. The inner `AgentEventDriverConfig::retry_forever` backs off across SSE failures, but a clean-but-empty stream close would loop tight. Add a small backoff or a bounded retry count for the outer restart.

app/src/ai/blocklist/orchestration_event_streamer.rs
```diff
@@ -804,14 +1094,15 @@ impl OrchestrationEventStreamer {
...
-        let hydrator = MessageHydrator::new(ai_client);
+        let hydrator = self.message_hydrator_for_run_id(&self_run_id);
 
         ctx.spawn(
             async move {
                 let mut consumer = SseForwardingConsumer {
                     tx,
                     self_run_id,
                     hydrator,
+                    hydrate_new_messages: true,
```
cephalonaut, line 1105: **`hydrate_new_messages` — dead flag in production?**

The struct field is only set `false` in the test `sse_forwarding_consumer_skips_message_hydration_when_disabled`. `start_sse_connection` (this line) hardcodes `true`. Either remove (YAGNI) or document the upcoming caller that will toggle it.

cephalonaut (commented):
Third pass after the wake-listener fix. The fix to the wake-trigger gap looks good. Three small things below.

cephalonaut (approved):

PR #9406, opened 2026-04-29
Point stable-skill instructions at resources/bundled/skills/

anshul-garg27:
### Description

The channel-gated skills README and the runtime error message in \`script/copy_conditional_skills\` both pointed contributors at \`resources/skills/\` for stable-ready skills, but that directory doesn't exist in the repo — \`script/prepare_bundled_resources\` copies the always-bundled skills from \`resources/bundled/\`, so the right destination is \`resources/bundled/skills/\`.

The result: if anyone bumped into the existing \`stable/\` guard (\`copy_conditional_skills\` line 39), the printed remediation pointed them at a path that doesn't exist.

Three updates:
- \`resources/channel-gated-skills/README.md\` — \"place them in the always-bundled \`resources/skills/\`\" → \`resources/bundled/skills/\` (in the prose blockquote and in the gating table).
- \`script/copy_conditional_skills\` — header comment + the user-facing error \`echo\`.

No behavior change in the script itself; only the comment and the error string contents.

### Testing

Doc/script-comment only — nothing to run. The actual copy logic is unchanged.

### Server API

No server changes.

### Agent Mode

Not applicable.

### Changelog Entries

None.

### DIFF
--- a/resources/channel-gated-skills/README.md
+++ b/resources/channel-gated-skills/README.md
@@ -19,7 +19,7 @@ channel-gated-skills/
 ```
 
 > **Stable-ready skills** do not belong here. Place them in the always-bundled
-> `resources/skills/` directory instead. The build script will error if a
+> `resources/bundled/skills/` directory instead. The build script will error if a
 > `stable/` directory exists under `channel-gated-skills/`.
 
 ## Progressive gating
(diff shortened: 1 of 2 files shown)


vorporeal (approved):

vorporeal:
thanks for fixing the incorrect paths!

PR #9411, opened 2026-04-29
Warp local API like server so we can use openai-compatible endpoint

regismesquita:
### **code not reviewed, use at your own risk**

It works, but it is not necessarily safe... think twice before running this.

## Server API dependencies

## Agent Mode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ desired_behavior.md
 
 # Don't include the python cache for bundled skills.
 __pycache__/
+# CocoIndex Code (ccc)
+/.cocoindex_code/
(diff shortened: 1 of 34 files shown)


regismesquita:
The github UI is not great and default to the upstream, not meant to open the PR here... closed.

regismesquita:
This PR was not meant to be open on this repo, but the github UI is not great and I selected the wrong place...

If you stumbled upon this, This works... it is a local version of their server and I am able to use openai-compatible endpoints (using at the moment with a sglang server running GLM 5.1). This code had zero reviews and was made just for fun... so think twice about actually using this.

Leaving this warning here just in case someone sees that and decided to actually use it... it will work... but who knows the sideeffects :))

PR #9412, opened 2026-04-29
Fix git diff chip flicker by not running shell fallback under GitRepoStatusModel

### LINKED ISSUE #9228: Git diff chip flickers between tracked-only and all-files count when untracked files are present

anshul-garg27:
Closes #9228.

### Description

\`GitDiffStats\` flickers between the right value and a tracked-only-count value on every block completion when both modified-tracked and untracked files exist in the repo. The cause is exactly what the issue calls out: the chip has two update paths and they disagree.

The watcher path (\`GitRepoStatusModel\` -> \`DiffStateModel::diff_metadata_against_head\`) uses \`git status --untracked-files=all\` and includes untracked files. The shell-fallback path (\`shell_git_line_changes\`) uses \`git diff --shortstat HEAD\` and excludes untracked files.

\`is_updated_externally\` already disables the *periodic* timer for chips that have a watcher attached, but \`run_chips\` was still doing one shell-driven fetch each time the prompt context updated whenever the chip didn't have a contextual \`initial_value_generator\`:

\`\`\`rust
if self.is_updated_externally(chip_kind) {
    let initial_gen = chip_kind.initial_value_generator();
    let generator = initial_gen.as_ref().unwrap_or(chip.generator()); // <-- shell fallback
    self.fetch_chip_value_once(...);
}
\`\`\`

For \`ShellGitBranch\` that's fine — its \`initial_value_generator\` is contextual, reads from \`PromptContext\`, and can't disagree with the watcher. For \`GitDiffStats\` it's \`None\`, so the call falls through to the chip's shell command and races with the watcher.

This change keeps the contextual initial fetch for \`ShellGitBranch\` but skips the fetch entirely for chips like \`GitDiffStats\` that don't have a contextual initial generator. The watcher's \`MetadataChanged\` event already populates the value (and \`refresh_metadata\` is called explicitly after block-completion events), so we don't need a parallel shell fetch.

### Testing

- Added \`test_externally_driven_git_diff_stats_does_not_run_shell_command\` in \`current_prompt_test.rs\`. It mirrors the existing \`test_externally_driven_chip_skips_periodic_timer\` setup but asserts both \`state.generator_handle.is_none()\` and that the recording \`CommandExecutor\` saw no commands.
- \`cargo fmt -p warp -- --check\` passes.
- Couldn't run the full \`cargo nextest\` locally (Metal toolchain isn't available on this machine — same situation as #9277).

### Server API

No server changes.

### Agent Mode

Not applicable.

### Changelog Entries

\`CHANGELOG-BUG-FIX\`: Stop the git diff context chip from briefly flickering to a tracked-only count when a command completes in a repo that also has untracked files.

### DIFF
--- a/app/src/context_chips/current_prompt.rs
+++ b/app/src/context_chips/current_prompt.rs
@@ -1041,15 +1041,24 @@ impl CurrentPrompt {
                 }
                 RefreshConfig::Periodically { .. } => {
                     if self.is_updated_externally(chip_kind) {
-                        let initial_gen = chip_kind.initial_value_generator();
-                        let generator = initial_gen.as_ref().unwrap_or(chip.generator());
-                        self.fetch_chip_value_once(
-                            chip_kind,
-                            generator,
-                            chip.on_click_generator().cloned(),
-                            true,
-                            ctx,
-                        );
+                        // The chip is being kept up to date by an external source (e.g. the
+                        // filesystem-watching `GitRepoStatusModel`). Only run an initial fetch
+                        // here when the chip provides a contextual `initial_value_generator`
+                        // — those compute their value from `PromptContext` and can't disagree
+                        // with the external source. Falling back to `chip.generator()` would
+                        // run the chip's shell command (e.g. `git diff --shortstat HEAD` for
+                        // `GitDiffStats`), which can produce a different value than the
+                        // external source and cause a brief flicker before the next external
+                        // update overwrites it (see issue #9228).
+                        if let Some(initial_gen) = chip_kind.initial_value_generator() {
+                            self.fetch_chip_value_once(
+                                chip_kind,
+                                &initial_gen,
+                                chip.on_click_generator().cloned(),
+                                true,
+                                ctx,
+                            );
+                        }
                     } else {
                         self.fetch_chip_value_at_interval(
                             chip_kind,
--- a/app/src/context_chips/current_prompt.rs
+++ b/app/src/context_chips/current_prompt.rs
@@ -1436,51 +1436,60 @@ impl CurrentPrompt {
                 self.git_repo_status = Some(weak);
                 ctx.subscribe_to_model(&strong, |me, event, ctx| match event {
                     GitRepoStatusEvent::MetadataChanged => {
-                        let metadata = me
-                            .git_repo_status
-                            .as_ref()
-                            .and_then(|w| w.upgrade(ctx))
-                            .and_then(|h| h.as_ref(ctx).metadata().cloned());
-
-                        let Some(metadata) = metadata else {
-                            return;
-                        };
-
-                        // Update ShellGitBranch.
-                        let new_branch = ChipValue::Text(metadata.current_branch_name.clone());
-                        let current_branch = me
-                            .latest_chip_value(&ContextChipKind::ShellGitBranch)
-                            .cloned();
-                        if current_branch.as_ref() != Some(&new_branch) {
-                            me.update_chip_value(
-                                &ContextChipKind::ShellGitBranch,
-                                Some(new_branch),
-                            );
-                            // Refresh the branch dropdown so it stays in sync.
-                            let chip_kind = ContextChipKind::ShellGitBranch;
-                            if let Some(chip) = chip_kind.to_chip() {
-                                if let Some(on_click_gen) = chip.on_click_generator().cloned() {
-                                    me.refresh_on_click_values(&chip_kind, on_click_gen, ctx);
-                                }
-                            }
-                        }
-
-                        // Update GitDiffStats with structured data directly.
-                        let new_diff_stats = ChipValue::GitDiffStats(
-                            GitLineChanges::from_diff_stats(&metadata.stats_against_head),
-                        );
-                        let current_diff_stats = me
-                            .latest_chip_value(&ContextChipKind::GitDiffStats)
-                            .cloned();
-                        if current_diff_stats.as_ref() != Some(&new_diff_stats) {
-                            me.update_chip_value(
-                                &ContextChipKind::GitDiffStats,
-                                Some(new_diff_stats),
-                            );
-                        }
+                        me.apply_git_repo_status_metadata(ctx);
                     }
                 });
+
+                // The model may already have metadata when we attach (e.g. when
+                // a session re-binds to a long-lived `GitRepoStatusModel`). No
+                // `MetadataChanged` event fires from subscribing alone, so apply
+                // any existing metadata once up front; otherwise the chips would
+                // stay empty until the next file event.
+                self.apply_git_repo_status_metadata(ctx);
             }
+        }
+    }
+
+    /// Pulls the current metadata from `git_repo_status` (if any) and pushes it
+    /// into the `ShellGitBranch` and `GitDiffStats` chip values. Used both as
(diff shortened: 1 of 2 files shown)


anshul-garg27:
Pushed [`5168a4b`](https://github.com/warpdotdev/warp/pull/9412/commits/5168a4b) addressing both concerns:

**1. Hydration on subscribe** — extracted the metadata-application body of the `MetadataChanged` closure into `apply_git_repo_status_metadata` and call it once up front in `set_git_repo_status` right after subscribing. The subscription callback now delegates to the same method, so existing-metadata hydration and on-change updates share a single code path. This covers the case where the model has metadata loaded before any subscriber attaches.

**2. Test that actually fails on the pre-fix code** — restructured `test_externally_driven_git_diff_stats_does_not_run_shell_command` to:
- Bootstrap a real session (`initialize_bootstrapped_session` + `load_external_commands`)
- Mark `git` as a known external command via the recording executor's first response, so the chip is `Enabled` and the pre-fix code would have called `fetch_chip_value_once` with the shell generator.
- Set `latest_context` to a populated `PromptContext`.
- `await_generators` after `update_states_with_new_context`.

The recording executor's command list is asserted empty after the flush — which would have been non-empty under the pre-fix code's shell fallback.

Also added a second test, `test_set_git_repo_status_applies_existing_metadata`, which exercises the hydration path directly by attaching a model whose metadata is already set and asserting both `ShellGitBranch` and `GitDiffStats` chip values are populated immediately — no `MetadataChanged` event needed.

Running `/oz-review`.

/oz-review

anshul-garg27:
Pushed `2c9ded5` addressing your second review.

**The concern:** the previous version's early return on `metadata().is_none()` left the chip values from the *previous* `GitRepoStatusModel` in place — visible whenever a session re-binds to a different repo and the new model's async fetch hasn't completed yet.

**The fix:** `apply_git_repo_status_metadata` now clears `ShellGitBranch` and `GitDiffStats` to `None` when the current `git_repo_status` has no metadata, so the only state the chips ever reflect is the *current* model's. The path is exercised both during `set_git_repo_status` (initial-attach hydration) and from the `MetadataChanged` callback (defensive — in practice the event always carries metadata, but the clearing branch is harmless if it ever doesn't).

**Test:** added `test_set_git_repo_status_clears_stale_chip_values_when_new_model_has_no_metadata` (after `test_set_git_repo_status_applies_existing_metadata`). It binds to a model with repo A's metadata, sanity-checks the chip shows `feature-x`, re-binds to a freshly-constructed model with no metadata, and asserts both chips are now `None`. The pre-fix version would have failed because the early return preserved repo A's values.

`cargo fmt -p warp -- --check` and `cargo check --features=gui --tests -p warp` both pass. `cargo nextest` skipped locally for the same Metal-toolchain reason as #9277.

Branch is rebased onto current `master` (`91dee6d`). Running `/oz-review`.

/oz-review

anshul-garg27:
Update — also confirmed locally with `cargo nextest` (Xcode is now installed on my machine, so the earlier "Metal toolchain unavailable" caveat no longer applies):

```
Starting 3 tests across 1 binary (3685 tests skipped)
    PASS [0.052s] (1/3) warp context_chips::current_prompt::tests::test_set_git_repo_status_applies_existing_metadata
    PASS [0.052s] (2/3) warp context_chips::current_prompt::tests::test_set_git_repo_status_clears_stale_chip_values_when_new_model_has_no_metadata
    PASS [0.059s] (3/3) warp context_chips::current_prompt::tests::test_externally_driven_git_diff_stats_does_not_run_shell_command
────────────
     Summary [0.062s] 3 tests run: 3 passed, 3685 skipped
```

Command: `cargo nextest run --features=gui -p warp -E 'test(test_set_git_repo_status_clears_stale_chip_values_when_new_model_has_no_metadata) | test(test_set_git_repo_status_applies_existing_metadata) | test(test_externally_driven_git_diff_stats_does_not_run_shell_command)'`

kevinchevalier:
Thank you for your submission. It looks like this was resolved by https://github.com/warpdotdev/warp/pull/9244, but let me know if you think that does not handle everything fixed here.

PR #9414, opened 2026-04-29
Remove vscode config

raydelto:
## Description
Remove the .vscode from the codebase, since it is user specific. The same way .idea user specific files are already ignored we should ignore .vscode preferences that currently contain MacOS specific configurations.

## Testing
If you cannot see .vscode directory in your codebase, the pull request was applied correctly.

## Server API dependencies
No server API changes.

## Agent Mode
This change was done manually, not an AI generated PR.

Nothing to be documented in changelog since this is not a user-facing change.


raydelto:
@cla-bot check

skal88:
Hi @raydelto 👋

Quick question — what's the motivation behind removing it? Why should it be user-specific?

According to the [official VS Code docs](https://code.visualstudio.com/docs/configure/settings#_workspace-settings), that folder is actually meant to be shared with the team. In my opinion it belongs in the project — it really helps to set up the workspace for development, debugging, etc.

Maybe I'm missing some context though — curious to hear your take! 🙂

captainsafia:
Thanks for the PR, @raydelto. I agree with @skal88's line of reasoning here. It's a common pattern to include the .vscode directory in projects and I think this is a useful convention to maintain.

I'm open to PRs that make the configuration more cross-platform or support.

Closing this one out for now. If you'd like to continue the discussion, feel free to open an issue.

raydelto:
I know that the issue is closed, but traceability, the rationale behind the idea of removing .vscode was to keep the codebase agnostic to specific code editing tools.

PR #9433, opened 2026-04-29
Reset tab name when AI CLI emits idle_prompt after /clear

mdalp:
## Description

When a user runs `/clear` inside an AI CLI (Claude Code, Codex, etc.) running in a Warp terminal, the CLI emits an `idle_prompt` OSC 777 event to signal it has returned to idle. Previously, `CLIAgentSession::apply_event` treated `IdlePrompt` as a complete no-op, so `session_context.query` and `session_context.summary` were never cleared — the tab title kept showing the last task/query even though the conversation had been reset.

This PR fixes that by:
- Clearing all title-contributing fields (`query`, `response`, `summary`, `tool_name`, `tool_input_preview`) on `IdlePrompt` in `CLIAgentSession::apply_event`, while preserving `cwd` and `project` (which reflect terminal location, not conversation state).
- Adding `IdlePrompt` to the `SessionUpdated` emission in `CLIAgentSessionsModel::update_from_event` so the workspace re-renders the tab bar immediately.
- As a bonus, also clearing the custom tab title (`group.clear_title`) in the `BlockListCleared` handler (cmd+k) so manually set tab names are reset when the buffer is wiped.

## Testing

Added unit test `idle_prompt_clears_title_context` in `mod_tests.rs` that:
- Starts a `CLIAgentSession` with all title-contributing fields populated.
- Applies an `IdlePrompt` event.
- Asserts that `query`, `response`, `summary`, `tool_name`, and `tool_input_preview` are all `None` afterward.
- Asserts that `cwd` and `project` are preserved.
- Asserts that `status` remains `Success` (no status regression).

Manual testing: run Claude Code inside Warp, submit a prompt, observe the tab title update, then run `/clear` — the tab title resets to the default.

## Server API dependencies

No server API changes.

## Agent Mode

CHANGELOG-BUG-FIX: Tab name now resets to default when `/clear` is run inside an AI CLI (Claude Code, Codex, etc.)

### DIFF
--- a/app/src/pane_group/pane/terminal_pane.rs
+++ b/app/src/pane_group/pane/terminal_pane.rs
@@ -663,6 +663,7 @@ fn handle_terminal_view_event(
                 if let Some(terminal_pane) = group.terminal_session_by_id(pane_id) {
                     terminal_pane.delete_blocks(ctx);
                 }
+                group.clear_title(ctx);
             }
             Event::ShareModalOpened(block_id) => {
                 group.terminal_with_open_share_block_modal = Some(terminal_pane_id);
--- a/app/src/terminal/cli_agent_sessions/mod.rs
+++ b/app/src/terminal/cli_agent_sessions/mod.rs
@@ -201,7 +201,15 @@ impl CLIAgentSession {
             }
             // IdlePrompt means the agent is sitting at its prompt waiting for input.
             // This should not affect status — otherwise it would override Success after a Stop event.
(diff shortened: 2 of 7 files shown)


mdalp:
Closing — the idle_prompt approach causes false positives (fires ~60s after every task completion, not just /clear). The session_id signal also proved unreliable. Needs a dedicated protocol event from Claude Code to implement reliably.

PR #9441, opened 2026-04-29
Add telemetry events for git button clicks and dialog completion

exzshao:
## Description

Adds telemetry to the git operation buttons in the code review pane so we can measure adoption, drop-off, and success/failure rates of the Commit / Push / Publish / Create PR / View PR flows.

**What:** Two new variants on `CodeReviewTelemetryEvent`:
- `GitButtonTriggered { button: GitButtonKind }` — fired once per click on the primary action button or dropdown item (covers Commit / Push / Publish / Create PR / View PR).
- `GitDialogCompleted { operation: GitOperationKind, success: bool }` — fired in each git dialog's async result callback. `GitOperationKind` distinguishes commit-dialog chained intents (`commit_only` / `commit_and_push` / `commit_and_create_pr`) from standalone push / publish / create-PR dialogs so analytics can tell those user flows apart.

**Why:** We had no signal on which git buttons users actually click, which commit-dialog intents they pick, or whether the underlying git operations succeed.

**How:**
- Added `GitButtonKind` and `GitOperationKind` helper enums in `app/src/code_review/telemetry_event.rs`.
- Emit `GitButtonTriggered` from the corresponding `handle_action` arms in `code_review_view.rs`, so primary-button and dropdown-menu clicks are both covered without threading a source flag.
- Emit `GitDialogCompleted` from each dialog mode's `start_confirm` result callback (`commit.rs` / `push.rs` / `pr.rs`). The commit dialog maps `CommitIntent` directly to `GitOperationKind` inline.
- Both events gated behind `EnablementState::Flag(FeatureFlag::GitOperationsInCodeReview)`, the same flag that gates the feature itself.
- No UGC: payloads contain only enum kinds and a success bool.

fixes: [APP-4291](https://linear.app/warpdotdev/issue/APP-4291/telemetry)

## Testing

Manually verified by running `cargo run --features log_named_telemetry_events` and exercising:
- Clicking each primary button (Commit / Push / Publish / Create PR / View PR) and each dropdown item — confirmed `CodeReview.GitButtonTriggered` fires once with the correct `button` value.
- Confirming each dialog mode for each commit intent — confirmed `CodeReview.GitDialogCompleted` fires once with the correct `operation` value and `success: true`.
- Inducing a failure (`gh` not authed) — confirmed `success: false`.

No new automated tests added. The instrumentation is straightforward emit-on-event plumbing with no algorithmic logic; existing code-review telemetry events (`PaneOpened`, `FileSaved`, `CommentAdded`, etc.) follow the same untested pattern.

## Server API dependencies

N/A — telemetry events only.

## Agent Mode

### DIFF
--- a/app/src/code_review/telemetry_event.rs
+++ b/app/src/code_review/telemetry_event.rs
@@ -267,6 +267,8 @@ pub enum CodeReviewTelemetryEvent {
         operation: GitOperationKind,
         /// Whether the underlying git operation succeeded.
         success: bool,
+        /// Raw error string when the operation failed, `None` on success.
+        error: Option<String>,
     },
 }
 
@@ -360,9 +362,14 @@ impl TelemetryEvent for CodeReviewTelemetryEvent {
             CodeReviewTelemetryEvent::GitButtonTriggered { button } => {
                 Some(json!({ "button": button }))
             }
-            CodeReviewTelemetryEvent::GitDialogCompleted { operation, success } => Some(json!({
+            CodeReviewTelemetryEvent::GitDialogCompleted {
+                operation,
+                success,
+                error,
+            } => Some(json!({
                 "operation": operation,
                 "success": success,
+                "error": error,
             })),
         }
     }
--- a/app/src/code_review/telemetry_event.rs
+++ b/app/src/code_review/telemetry_event.rs
@@ -475,7 +475,9 @@ impl TelemetryEventDesc for CodeReviewTelemetryEventDiscriminants {
             Self::GitButtonTriggered => {
                 "User clicked a git operation button in the code review header"
             }
-            Self::GitDialogCompleted => "Git operation dialog reached a terminal state (succeeded, failed, or cancelled)",
+            Self::GitDialogCompleted => {
+                "Git operation dialog reached a terminal state (succeeded, failed, or cancelled)"
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


app/src/code_review/telemetry_event.rs
```diff
@@ -215,6 +257,17 @@ pub enum CodeReviewTelemetryEvent {
...
+    /// Emitted when a user clicks a git operation button in the code review
+    /// header (primary button or dropdown item).
+    GitButtonTriggered { button: GitButtonKind },
+    /// Emitted when a git dialog finishes its async operation (success or failure).
+    GitDialogCompleted {
+        /// The git operation that actually ran (e.g. `commit_and_push` for
+        /// the commit dialog with that chained intent).
+        operation: GitOperationKind,
+        /// Whether the underlying git operation succeeded.
+        success: bool,
```
MaggieShan: If it failed, can we also include the error string on why it failed

app/src/code_review/telemetry_event.rs
```diff
@@ -215,6 +257,17 @@ pub enum CodeReviewTelemetryEvent {
         /// Number of outdated imported comments after relocation.
         outdated_count: usize,
     },
+    /// Emitted when a user clicks a git operation button in the code review
+    /// header (primary button or dropdown item).
+    GitButtonTriggered { button: GitButtonKind },
+    /// Emitted when a git dialog finishes its async operation (success or failure).
+    GitDialogCompleted {
```
MaggieShan, line 281: Let's also add an event for when the dialog was exited/cancelled

MaggieShan (approved):
Small nits but overall looks good!

PR #9443, opened 2026-04-29
Make code blocks in thinking text unselectable

seemeroland:
## Description

Thinking text is not selectable, but code blocks in thinking text would select while dragging but become unselected when you released the mouse, which feels buggy. Make them unselectable to be consistent with other thinking text for now (but consider making thinking text selectable later on).

## Testing

Tested locally, can't select text in code blocks in thinking text


harryalbert (approved):
LGTM. I feel like making thinking text selectable at some point would be nice, but agreed that right now this makes sense

PR #9444, opened 2026-04-29
Surface $CDPATH directories in cd argument completion

### LINKED ISSUE #373: Completions: Support CDPATH
### Describe the bug

```shell
I use `CDPATH` in my `.profile` to add my work root, i.e. `export CDPATH=.:~/dev/work` in order to easily switch between repos (`cd web-app`, etc.) The commands work, but I get no tab completion for them, which is a downgrade in UX compared to other terminals.
```

### To Reproduce

Steps to reproduce:
1. Add `export CDPATH=.:/some/path` to your `.profile` (or just run the command in a shell)
2. Type `cd some-dir-in-CDPATH`
3. Press Tab

### Expected behaviour

```shell
Tab completions should also be filled with directories from the CDPATH
```

### Screenshots

_No response_

### Operating System

MacOS

### OS Version

12.0.1

### Additional context

_No response_

Faizanq:
## Description

Closes #373.

`cd <Tab>` now completes against directories under `$CDPATH` in addition to the current working directory, matching what the shell itself resolves when you `cd` to a relative path.

The bash and zsh bootstrap payloads now ship the `CDPATH` value alongside `home_dir`, `path`, etc. That value flows through `BootstrappedValue` -> `SessionInfo` -> `Session::cdpath()` and is exposed on `PathCompletionContext::cdpath()`. A new `sorted_cd_directories` helper wraps `sorted_directories_relative_to`: for each non-`.` entry in `$CDPATH`, it lists directory matches under that entry using a `pwd`-overriding context wrapper, deduped against the pwd-relative results.

The cd-specific path is opt-in to the `cd` command. Other `Folders` templates (`tree`, `du`, etc.) keep the existing pwd-only completion. Absolute paths (`/foo`), home-rooted paths (`~/foo`), and explicit relative prefixes (`./foo`, `../foo`) skip CDPATH to match shell semantics.

## Testing

Four new unit tests in `crates/warp_completer/src/completer/engine/path_test.rs`:

- `test_sorted_cd_directories_no_cdpath_matches_existing_behavior`: without `cdpath`, the cd-aware path returns identical results to the current `Folders` path
- `test_sorted_cd_directories_includes_cdpath_entries`: pwd entries first, then unique cdpath entries; duplicates deduped; non-directories filtered
- `test_sorted_cd_directories_ignores_cdpath_for_absolute_token`: absolute path tokens bypass cdpath
- `test_sorted_cd_directories_skips_dot_entry_in_cdpath`: `.` in cdpath is handled by the pwd-relative pass and not double-listed

All 13 path tests pass; clippy on `warp` and `warp_completer` is clean. I couldn't run the full warp lib test suite end-to-end locally (ran into a disk-space issue mid-build), relying on CI for the broader pass.

## Agent Mode

CHANGELOG-IMPROVEMENT: `cd <Tab>` now completes against directories listed in `$CDPATH`, matching the resolution shells already perform.


zachbai (approved):
Thanks for the change!

PR #9449, opened 2026-04-29
Add experiment setup for SSH

kevinyang372:
## Description
Enable SSH remote server on stable!

This also setup the server experiment rollout so we could have a percentage of users with the setting default on and the rest with default off


kevinyang372:
* **#9449** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9449?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9449?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_features/src/lib.rs
```diff
@@ -946,6 +945,9 @@ pub const RELEASE_FLAGS: &[FeatureFlag] = &[
     // Marked text is currently only supported on MacOS.
     #[cfg(target_os = "macos")]
     FeatureFlag::ImeMarkedText,
+    // Remote server binary is not yet supported on Windows.
```
moirahuang, line 948: just to confirm my understanding, we're putting this here instead of `app/Cargo.toml` because we need `#[cfg(not(windows))]` right now?

moirahuang (approved):

crates/warp_features/src/lib.rs
kevinyang372: Ya

PR #9451, opened 2026-04-29
Improve repo picker readability for long saved-repo paths

### LINKED ISSUE #9439: Tab config repo picker truncates long paths with no way to see, scroll, or alias them
## Summary
The repo picker used by tab config `type = "repo"` parameters truncates long entries with no way to see the rest of the path, scroll horizontally, hover for a tooltip, or attach a friendly nickname/alias. When several saved repos live deep inside paths like `~/Library/Mobile Documents/iCloud~md~obsidian/Documents/...` or `~/Library/CloudStorage/GoogleDrive-<email>/...`, the visible portion of two different entries can be byte-for-byte identical, making it impossible to tell them apart at the moment of selection.

## Problem
Tab configs make it easy to launch flows like "open Claude in a project" by declaring a `[params.<name>] type = "repo"` parameter. On open, Warp shows a dropdown of saved repos so the user can pick which directory to launch in.

Today that dropdown:
- Renders each entry as the raw absolute path.
- Has a fixed dropdown width with single-line entries.
- Truncates the tail of long paths with an ellipsis.
- Does not appear to expose hover/tooltip, horizontal scroll, mid-path elision, or any user-defined display name.

In practice this means:
- Two different repos that share a long common prefix (for example two projects under the same iCloud Documents folder) collapse into visually identical rows ending in `…`. The user cannot reliably pick the right one without trial and error.
- The most distinctive part of a path (typically the trailing repo/project name) is exactly the part that gets cut off.
- The search box at the top filters by the underlying path, but the user can't always remember the exact prefix to type, especially for cloud-sync paths with vendor-specific tokens like `iCloud~md~obsidian` or `GoogleDrive-<email>`.

This is mostly a UX issue for users who keep work in nested cloud-sync folders, monorepos, or any directory with a deep parent chain.

## Expected behavior
A user opening a tab config that uses a `repo` parameter should always be able to identify the full path of each entry in the picker, and ideally label entries with a friendly name they choose.

Reasonable ways to satisfy this (any one or a combination would help):
1. **Show the full path on demand** via a tooltip on hover, by widening the dropdown to the longest entry up to a sensible cap, or by allowing the dropdown to scroll horizontally.
2. **Smart elision in the middle** of the path instead of the end, so the trailing project/repo name is always visible (e.g. `/Users/brandon/Library/…/obsidian/Documents/<project>`). macOS Finder and several IDE quick-open palettes use this pattern.
3. **User-defined nicknames / aliases** for saved repos, surfaced as the primary label in the picker, with the full path shown as secondary text underneath or on hover. This could be done either:
   - via a per-repo nickname stored alongside the saved repo entry in Warp's persisted workspace state, edited from a "rename" action in the picker, or
   - via an optional `display_name` / `alias` field declared per `[params.<name>] type = "repo"` entry in the tab config TOML schema (less likely to scale because the schema doesn't own the saved-repo list).
4. **Two-line rows** showing the basename prominently and the full path as muted secondary text, similar to VS Code's "Open Recent" picker.

Any of these would unblock the underlying need: "let me reliably distinguish saved repos at the moment of selection."

## Actual behavior
Long paths are visually cut off with an ellipsis at a fixed dropdown width, and there is no hover, scroll, alias, or alternate display mode to recover the missing characters. Entries that share a long prefix become indistinguishable.

1. Save several repos to Warp's project list whose absolute paths share a long common prefix and exceed the dropdown's visible width. Concrete examples that reproduce the issue cleanly on macOS:
   - `~/Library/Mobile Documents/iCloud~md~obsidian/Documents/<project A>`
   - `~/Library/Mobile Documents/iCloud~md~obsidian/Documents/<project B>`
   - `~/Library/CloudStorage/GoogleDrive-<email>/My Drive/<project>`
2. Create or use a tab config that declares a `repo` parameter, e.g.:
   ```toml
   name = "Open Project (Claude)"

   [[panes]]
   id = "main"
   type = "agent"
   directory = "{{project}}"
   commands = ["claude"]

   [params.project]
   type = "repo"
   description = "Choose repo"
   ```
3. Open the tab config from the `+` menu so the params modal appears.
4. Click the `project` field to open the repo picker.
5. Observe that the iCloud and Google Drive entries are truncated mid-path with `…`, and that two distinct entries can render with identical visible text.

## Artifacts
**Screenshot 1:** Tab config repo picker showing the `Open: 🧡 Open` modal with `Choose repo` dropdown; multiple entries under `~/Library/Mobile Documents/iCloud~md~obsidian/...` and `~/Library/CloudStorage/GoogleDrive-...` are truncated at the right edge with a trailing ellipsis, hiding the part of the path that would distinguish them.

Image

## Warp version

## Operating system
macOS (version not specified by reporter; inferred from `/Users/<name>/Library/...` paths in the screenshot)

## Possible source references
- `app/src/tab_configs/repo_picker.rs` — defines the picker shown for `type = "repo"` parameters; contains a `DEFAULT_DROPDOWN_WIDTH` constant (`380.0`) and renders each entry as a single line of raw path text via `path.to_string_lossy()`. This is the natural place to introduce a wider/auto-sized dropdown, mid-path elision, a hover tooltip, or a richer per-entry render that surfaces an alias.
- `app/src/tab_configs/params_modal.rs` — hosts the parameter modal that embeds the repo picker; relevant if the fix needs to coordinate with sibling controls (search input, footer).
- `app/src/tab_configs/mod.rs` — top-level tab config wiring; relevant if any new schema field (e.g. an optional alias) is added.

saucy-tech:
## Description

Fixes #9439.

Screenshot 2026-05-01 at 8 47 55 PM

The tab config `type = "repo"` picker renders saved repos as raw, single-line entries. When several saved repos live deep inside paths like `~/Library/Mobile Documents/iCloud~md~obsidian/Documents/...`, the visible portion of two different rows can be byte-for-byte identical, so the user can't reliably pick the right one.

Per @moirahuang and @alokedesai's preferred direction (#9439 comment, PR review), this is a **left-clip + `~` substitution + tooltip** approach — no soft-wrap, no schema changes:

1. **`~` substitution for the home directory** in the displayed label, via `warp_util::path::user_friendly_path` — the same helper already used in `tab_configs/action_sidecar.rs` and `tab_configs/session_config_rendering.rs`. The dropdown action and the cached `self.selected` continue to carry the raw absolute path so launched sessions resolve correctly.
2. **Left-truncation with a leading `…`** (`crate::util::truncation::truncate_from_beginning`) at a 40-character cap. The trailing repo name — the most distinctive part of the path — stays visible in every row.
3. **Hover tooltip with the full absolute path** via the existing `MenuItemFields::with_tooltip`, exposed through a new `with_tooltip` builder on `DropdownItem`. Even after `~` substitution and left-clip the user can recover the original path.
4. **Action-based preselection.** `RepoPicker::refresh_items` now calls the existing public `FilterableDropdown::set_selected_by_action(RepoPickerAction::Select(raw))` instead of `set_selected_by_name(display)`. If two raw paths happen to share the same trailing 40 characters, preselection still picks the correct backing repo. `RepoPickerAction` already derives `PartialEq`, so this is a small swap with no new public dropdown API.

The new `with_tooltip` builder on `DropdownItem` is opt-in (defaults to `None`), so other `FilterableDropdown` consumers are unaffected.

### Out of scope (left for follow-ups)

- **User-defined nicknames / aliases.** The issue body flags this as a saved-repo metadata change rather than picker UX.
- **Soft-wrap to two lines.** Considered earlier in this PR; reverted per maintainer review in favor of the left-clip approach.
- **Mid-path elision.** Left-clip supersedes this for the picker rows.

## Testing

`cargo fmt --check -p warp`, `cargo check -p warp`, and `cargo clippy -p warp --all-targets` are all clean locally.

Manual smoke verified locally with deeply-nested iCloud paths — see follow-up comment for screenshot. Short paths render unchanged with `~` substitution; long paths left-clip with a leading `…` so the trailing repo name stays visible; full absolute path appears on hover.

The change leans on already-tested primitives:

- `warp_util::path::user_friendly_path` has unit tests in `crates/warp_util/src/path_test.rs`.
- `crate::util::truncation::truncate_from_beginning` is the same helper used elsewhere in the app.
- `FilterableDropdown::set_selected_by_action` and `MenuItemFields::with_tooltip` are existing primitives.

## Agent Mode

CHANGELOG-IMPROVEMENT: Tab config repo picker now abbreviates the home directory to `~`, left-truncates long paths so the trailing repo name stays visible, and surfaces the full absolute path on hover so deeply nested saved repos stay distinguishable.

### DIFF
--- a/app/src/tab_configs/repo_picker.rs
+++ b/app/src/tab_configs/repo_picker.rs
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 
+use warp_util::path::user_friendly_path;
 use warpui::{
     elements::{Border, ChildView, Container, Hoverable, MouseStateHandle, Text},
     platform::Cursor,
@@ -18,6 +19,12 @@ const DEFAULT_DROPDOWN_WIDTH: f32 = 380.;
 
 /// Label for the sticky "Add new repo..." footer at the bottom of the picker.
 const ADD_NEW_REPO_LABEL: &str = "+ Add new repo...";
+
+/// Maximum number of lines a single repo row may render before being clipped.
+/// Two lines is enough to fit any realistic absolute path on a 412-pixel-wide
+/// dropdown without giving up the trailing repo name to a mid-string ellipsis,
+/// while keeping rows compact.
+const REPO_ROW_MAX_LINES: usize = 2;
 
 /// A filterable dropdown listing known repos (from `PersistedWorkspace`), with a
 /// sticky "+ Add new repo..." footer that is always visible even when scrolling.
@@ -147,24 +154,53 @@ impl RepoPicker {
         // workspaces() already returns entries sorted by most-recently-touched.
         // "+ Add new repo..." is a sticky footer (not a list item) so it is
         // not included here.
+        //
+        // Display labels abbreviate the user's home directory to `~` so deeply
+        // nested paths under shared parents (e.g. iCloud / Google Drive) stay
+        // distinguishable in a fixed-width dropdown. The action and the cached
+        // `self.selected` continue to carry the raw absolute path so the launched
+        // session resolves the directory correctly. The original full path is
+        // exposed on hover via the tooltip, and rows soft-wrap to two lines so
+        // the trailing repo name is preserved instead of being elided.
+        let home_dir = dirs::home_dir();
+        let home_dir_str = home_dir.as_ref().and_then(|h| h.to_str());
         let items: Vec<DropdownItem<RepoPickerAction>> = PersistedWorkspace::as_ref(ctx)
             .workspaces()
             .filter(|ws| ws.path.exists())
             .map(|ws| {
-                let path_str = ws.path.to_string_lossy().into_owned();
-                DropdownItem::new(path_str.clone(), RepoPickerAction::Select(path_str))
+                let raw_path = ws.path.to_string_lossy().into_owned();
+                let display = user_friendly_path(&raw_path, home_dir_str).into_owned();
+                DropdownItem::new(display, RepoPickerAction::Select(raw_path.clone()))
+                    .with_tooltip(raw_path)
+                    .with_max_lines(REPO_ROW_MAX_LINES)
             })
             .collect();
 
         let path_to_select = select_path
             .or(self.selected.as_deref())
             .map(|s| s.to_owned());
+        // The dropdown looks up the prior selection by display label, so the
+        // lookup key has to use the same `~`-substituted form the rows now
+        // render with. Otherwise a default value pointing at a path under
+        // $HOME would never match its own row.
+        let display_to_select = path_to_select
+            .as_deref()
+            .map(|p| user_friendly_path(p, home_dir_str).into_owned());
         self.dropdown.update(ctx, |dropdown, ctx| {
             dropdown.set_items(items, ctx);
-            if let Some(ref path) = path_to_select {
-                dropdown.set_selected_by_name(path.as_str(), ctx);
+            if let Some(ref display) = display_to_select {
+                dropdown.set_selected_by_name(display.as_str(), ctx);
             }
         });
+
+        // Mirror the raw path into `self.selected` so the `selected_value()`
+        // fallback (which reads the dropdown's selected_item_label) never
+        // returns the `~`-substituted display form to callers that intend to
+        // pass it to the shell or to a `cd`. The substituted form is purely
+        // for display.
+        if let Some(path) = path_to_select {
+            self.selected = Some(path);
+        }
 
         ctx.notify();
     }
--- a/app/src/tab_configs/repo_picker.rs
+++ b/app/src/tab_configs/repo_picker.rs
@@ -12,19 +12,20 @@ use crate::{
     ai::persisted_workspace::{PersistedWorkspace, PersistedWorkspaceEvent},
     appearance::Appearance,
     tab_configs::PickerStyle,
+    util::truncation::truncate_from_beginning,
     view_components::{DropdownItem, FilterableDropdown},
 };
 
 const DEFAULT_DROPDOWN_WIDTH: f32 = 380.;
 
+/// Max characters shown for a repo path in the picker before left-truncating
+/// with an ellipsis. Sized to fit `DEFAULT_DROPDOWN_WIDTH` (380px) and the
+/// 412px `params_modal` style without triggering the menu's `autosize_text`
+/// shrink path (`app/src/menu.rs`).
+const MAX_REPO_DISPLAY_LEN: usize = 40;
+
 /// Label for the sticky "Add new repo..." footer at the bottom of the picker.
 const ADD_NEW_REPO_LABEL: &str = "+ Add new repo...";
-
-/// Maximum number of lines a single repo row may render before being clipped.
-/// Two lines is enough to fit any realistic absolute path on a 412-pixel-wide
-/// dropdown without giving up the trailing repo name to a mid-string ellipsis,
-/// while keeping rows compact.
-const REPO_ROW_MAX_LINES: usize = 2;
 
 /// A filterable dropdown listing known repos (from `PersistedWorkspace`), with a
 /// sticky "+ Add new repo..." footer that is always visible even when scrolling.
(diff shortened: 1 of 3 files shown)


adrisys:
@oz-agent

saucy-tech:
@cla-bot check

alokedesai:
Hey @saucy-tech, thanks so much for the PR! I think this is a great start, but I think we should go with this solution instead: https://github.com/warpdotdev/warp/issues/9439#issuecomment-4345826213

I think the main change should be that we don't soft wrap and instead clip from the left instead of clip from the right. Once that change is made, this PR should be good to go!

saucy-tech:
Hey @alokedesai made the change you asked for in 69e19c0! Oz has flagged a few follow-ups since; how would you like me to handle them?

alokedesai:
Mind addressing the comments and rerunning oz-review? once it's approved i'll take a look

saucy-tech:
> Mind addressing the comments and rerunning oz-review? once it's approved i'll take a look

Oz is passing. Thank you Warp for going OSS, and allowing me to submit an issue and contribute the PR!

app/src/tab_configs/repo_picker.rs
```diff
@@ -147,22 +154,54 @@ impl RepoPicker {
         // workspaces() already returns entries sorted by most-recently-touched.
         // "+ Add new repo..." is a sticky footer (not a list item) so it is
         // not included here.
+        //
+        // Each item's `display_text` is the full `~`-substituted path,
+        // wrapped over up to `REPO_DISPLAY_MAX_LINES` lines (rather than
```
alokedesai: Hey @saucy-tech, I think the softwrapping logic is still in this PR? Mind removing it?

app/src/tab_configs/repo_picker.rs
```diff
@@ -176,11 +215,14 @@ impl RepoPicker {
...
-        self.selected
-            .clone()
-            .or_else(|| self.dropdown.as_ref(app).selected_item_label())
+    /// Returns the currently selected repo path (raw absolute path).
+    ///
+    /// `refresh_items` mirrors the selected raw path into `self.selected`
+    /// — including the implicit first-row selection used when the caller
+    /// didn't pass an explicit default — so this always matches whatever
+    /// row is visually highlighted in the dropdown.
+    pub fn selected_value(&self, _app: &AppContext) -> Option<String> {
```
alokedesai, line 211: Could you go into more detail about what necessitated the change to `selected_value` here?

app/src/view_components/dropdown.rs
```diff
@@ -117,17 +131,57 @@ where
...
+    /// Set a hover tooltip for this row. Useful when `display_text` is a
+    /// shortened form of richer underlying data (e.g. a truncated path).
+    pub fn with_tooltip(mut self, tooltip: impl Into<String>) -> Self {
+        self.tooltip = Some(tooltip.into());
+        self
+    }
+
+    /// Set a richer text for filter matching. When unset, `FilterableDropdown`
+    /// falls back to matching against `display_text`.
+    pub fn with_filter_text(mut self, filter_text: impl Into<String>) -> Self {
```
alokedesai: The reason you're adding this is so that someone can filter via absolute paths, yeah? FOr the sake of code simplicity, I'd be in favor of remove that for now

app/src/view_components/dropdown.rs
```diff
@@ -117,17 +131,57 @@ where
...
-        let menu_item = MenuItemFields::new(dropdown_item.display_text.clone())
-            .with_on_select_action(DropdownAction::SelectActionAndClose(
-                dropdown_item.action.clone(),
-            ));
+        let base = match dropdown_item.multiline_max_lines {
+            Some(max_lines) => {
+                MenuItemFields::new_multiline(dropdown_item.display_text.clone(), max_lines)
+            }
+            None => MenuItemFields::new(dropdown_item.display_text.clone()),
+        };
```
alokedesai: Related to the comment above, let's remove this

app/src/view_components/filterable_dropdown.rs
```diff
@@ -611,32 +630,38 @@ where
...
-        // We keep track of the label of the current element, and assume
-        // it won't be visible in the newly computed list of filtered items.
-        // If it isn't, we set the selected element to the first index of
-        // the new elements such that there's always a candidate element to select.
-        let current_label = self.current_selected_item_label();
-        let mut current_label_not_visible = true;
+        // We track the current selection by *action* rather than by rendered
+        // label so that two items whose `display_text` collides after
+        // truncation cannot be confused at preselection time.
+        let current_action = self.current_selected_item_action().cloned();
```
alokedesai: what necessitated this change? By definition each item in the dropdown is still unique for a path even after eliding the home directory, no?

alokedesai (commented):
Thank you so much for working on this! I left a more detailed review. Once those get addressed, I think this is good to go :D

saucy-tech:
Hey @alokedesai sorry for the back-and-forth here. I force-pushed back to `69e19c0` to get this PR back to a smaller scope.

That removes the broader Oz-driven additions from the previous version (thank you for those detailed review and comments):

| File:line | Resolution |
|---|---|
| `repo_picker.rs:159` | soft-wrap removed; repo rows are single-line with left truncation |
| `repo_picker.rs:224` | `selected_value` reverted to the original fallback behavior |
| `dropdown.rs:144` | `with_filter_text` removed |
| `dropdown.rs:178` | `with_multiline` and the `MenuItemFields::new_multiline` branch removed |
| `filterable_dropdown.rs:636` | selection preservation reverted to the existing label-based behavior |

I made one small follow-up commit on top (`bcf7c59`): preselection in `RepoPicker::refresh_items` now uses the existing `FilterableDropdown::set_selected_by_action` instead of `set_selected_by_name`. `RepoPickerAction` already derives `PartialEq`, so this is a tiny change with no new public dropdown API, and it avoids selecting the wrong backing repo if two raw paths collapse to the same truncated display label.

Oz is still asking for soft-wrap/multiline rows, but I left that out because your earlier direction was to avoid soft-wrap and clip from the left instead. If you'd prefer the multiline rendering after all, I can restore only that piece; otherwise I think the PR is now back to the intended scope.

app/src/tab_configs/repo_picker.rs
saucy-tech: Removed — back to single-line via `truncate_from_beginning` at 69e19c0.

app/src/tab_configs/repo_picker.rs
saucy-tech: Reverted to the original body. The first-row mirror is gone.

app/src/view_components/dropdown.rs
saucy-tech: Removed.

app/src/view_components/dropdown.rs
saucy-tech: Removed (along with the `MenuItemFields::new_multiline` branch).

app/src/view_components/filterable_dropdown.rs
saucy-tech: Reverted to label-based — agree the items are unique by path. The follow-up commit `bcf7c59` only swaps the call site in `repo_picker.rs` to `set_selected_by_action` for label-collision defense; no changes to `FilterableDropdown` internals.

saucy-tech:
Manual smoke on a local OSS build (`./script/run`). Saved repos include both a short path and a deeply-nested iCloud path that's well past the 40-char cap.

Screenshot 2026-05-01 at 8 47 55 PM

The first row shows `~` substitution at full length. The second row shows the left-clip in action: leading `…` plus the trailing `Brandon/2-Areas/Health` segment, which is exactly the part that used to get cut off by the right-side ellipsis.

app/src/tab_configs/repo_picker.rs
```diff
@@ -218,3 +250,11 @@ impl TypedActionView for RepoPicker {
         }
     }
 }
+
+/// Formats an absolute repo path for display in the picker: replaces a
+/// leading `home` directory with `~`, then left-truncates with an ellipsis if
+/// the result is longer than [`MAX_REPO_DISPLAY_LEN`].
+fn format_display_path(raw_path: &str, home: Option<&str>) -> String {
+    let abbreviated = user_friendly_path(raw_path, home);
+    truncate_from_beginning(&abbreviated, MAX_REPO_DISPLAY_LEN)
```
alokedesai: Instead of truncating from the beginning like this, could we expose a new optional setting on the `Dropdown` that configures the `ClipConfig` [1]. That will be cleaner than trying to approximate it based on the number of characters

[1] https://github.com/warpdotdev/warp/blob/e3514ff599cd4f02f8e5e9aa1e19d1f8be2c0080/crates/warpui_core/src/text_layout.rs#L438

alokedesai (commented):
Thanks for all of the revs on this! I have one more comment and then I think this is good to go--I'll take it spin tomorrow morning

app/src/tab_configs/repo_picker.rs
saucy-tech: Done in 330184d. Removed the fixed 40-character truncation and now pass the full `~`-abbreviated path into the dropdown, with the row text clipped at render time via `ClipConfig::start()`. Tooltip and raw-path-backed selection are unchanged.

alokedesai (approved):
Thanks for contributing!

PR #9452, opened 2026-04-29
APP-4263 Fix git ops flicker bug 

exzshao:
## Description

Fixes APP-4263: the code review git operations header briefly showed incorrect PR-related actions while PR info for the current branch was still being refreshed.

`load_metadata_for_repo` returns `pr_info: None` because PR info is fetched separately via `refresh_pr_info` (`gh pr view`). That made the UI treat unknown PR state as "no PR" while the async lookup was in flight, which could briefly show `Create PR`. With the handle-only version, switching between two branches that both have PRs could also briefly fall back from `PR #N` to disabled `Commit` before resolving to `PR #M`.

### Fix

In `app/src/code_review/diff_state.rs`:

1. Track an in-flight PR lookup with `refreshing_pr_info_handle: Option<SpawnedFutureHandle>` instead of storing PR staleness in metadata.
2. Abort any previous PR lookup before starting a new one.
3. Carry forward cached `pr_info` through metadata refreshes so the header does not clear to disabled `Commit` while the next branch's PR lookup is pending.

In `app/src/code_review/code_review_view.rs`:

1. Gate `Create PR` on `!is_pr_info_refreshing()` so an unknown/pending PR lookup is not treated as "no PR".
2. If cached PR info is being carried while a new lookup is pending, keep the previous `PR #N` label visible but disable the button/menu item with a "Refreshing PR info" tooltip so it cannot open a stale PR.
3. Once the lookup resolves, update the header to the new branch's `PR #M`, `Create PR`, or disabled `Commit` state as appropriate.

## Testing

Not run in this update per request.

Expected/manual behavior:
- Main/master → feature branch: no `Create PR` flicker while `gh pr view` is pending.
- Branch with PR → branch with PR: no disabled `Commit` interstitial; previous PR label remains disabled until the new PR info resolves.
- Branch with PR → branch without PR: previous PR label remains disabled briefly, then transitions to `Create PR` if the branch is eligible.
- Dropdown `Create PR` and commit-dialog create-PR intent remain unavailable while PR info is refreshing.

## Agent Mode

UPDATED LOOM: https://www.loom.com/share/4c94527c09914af9b295e49b4719159a

### DIFF
--- a/app/src/code_review/code_review_view.rs
+++ b/app/src/code_review/code_review_view.rs
@@ -6722,7 +6722,7 @@ impl CodeReviewView {
                 // intent (Commit and push vs Commit and publish).
                 let diff_state = self.diff_state_model.as_ref(ctx);
                 let allow_create_pr = diff_state.pr_info().is_none()
-                    && !diff_state.pr_info_stale()
+                    && !diff_state.is_pr_info_refreshing()
                     && !diff_state.is_on_main_branch();
                 let has_upstream = diff_state.upstream_ref().is_some();
                 ctx.add_typed_action_view(|ctx| {
@@ -6772,7 +6772,7 @@ impl CodeReviewView {
         let has_uncommitted_changes = self.has_uncommitted_changes(app);
         let has_upstream = diff_state.upstream_ref().is_some();
         let has_local_commits = !diff_state.unpushed_commits().is_empty();
-        let pr_info_stale = diff_state.pr_info_stale();
+        let is_pr_info_refreshing = diff_state.is_pr_info_refreshing();
         // False when upstream == main (e.g. after `git checkout -b feature origin/master`),
         // which means the branch hasn't been pushed to its own remote ref yet.
         let upstream_differs_from_main = diff_state.upstream_differs_from_main();
@@ -6785,7 +6785,7 @@ impl CodeReviewView {
             PrimaryGitActionMode::Push
         } else if diff_state.pr_info().is_some() {
             PrimaryGitActionMode::ViewPr
-        } else if !pr_info_stale
+        } else if !is_pr_info_refreshing
             && has_upstream
             && !diff_state.is_on_main_branch()
             && upstream_differs_from_main
@@ -6826,6 +6826,7 @@ impl CodeReviewView {
                     button.set_label("Push", ctx);
                     button.set_icon(Some(Icon::ArrowUp), ctx);
                     button.set_disabled(false, ctx);
+                    button.set_tooltip(None::<String>, ctx);
                     button.set_on_click(
                         |ctx| ctx.dispatch_typed_action(CodeReviewAction::OpenPushDialog),
                         ctx,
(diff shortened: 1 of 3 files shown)


app/src/code_review/diff_state.rs
```diff
@@ -348,6 +348,7 @@ struct DiffMetadata {
     unpushed_commits: Vec<Commit>,
     upstream_ref: Option<String>,
     pr_info: Option<PrInfo>,
+    pr_info_stale: bool,
```
MaggieShan: This is essentially meant to check whether there is a pr info check in progress right? Can we instead store something like a `Option<SpawnedFutureHandle>` which when `is_some()` will let us similarly know if it's in progress/stale ==> then we can also use that handle to abort the check earlier on a refresh

MaggieShan (commented):
One main comment

app/src/code_review/diff_state.rs
```diff
@@ -1531,10 +1536,6 @@ impl DiffStateModel {
             if FeatureFlag::GitOperationsInCodeReview.is_enabled() {
                 self.refresh_pr_info(ctx);
             }
-        } else if FeatureFlag::GitOperationsInCodeReview.is_enabled() && self.pr_info().is_none() {
```
MaggieShan, line 1534: Does it make sense to remove this? This means we wouldn't be able to fetch new pr info if a pr was created externally

app/src/code_review/code_review_view.rs
```diff
@@ -6919,15 +6934,22 @@ impl CodeReviewView {
             MenuItemFields::new(format!("PR #{}", pr_info.number))
                 .with_icon(Icon::Github)
                 .with_on_select_action(CodeReviewAction::ViewPr(pr_info.url))
+                .with_disabled(diff_state.is_pr_info_refreshing())
                 .into_item()
         } else {
             let is_on_main = diff_state.is_on_main_branch();
             let has_upstream = diff_state.upstream_ref().is_some();
+            let is_pr_info_refreshing = diff_state.is_pr_info_refreshing();
```
MaggieShan: nit: you can define this before the if/else block so that you don't have to call `is_pr_info_refreshing()` in both branches

app/src/code_review/code_review_view.rs
```diff
@@ -6867,6 +6881,7 @@ impl CodeReviewView {
                     button.set_label("Publish", ctx);
                     button.set_icon(Some(Icon::UploadCloud), ctx);
                     button.set_disabled(false, ctx);
+                    button.set_tooltip(None::<String>, ctx);
```
MaggieShan: nit: remove `::<String>` which is also in the other lines above ^

MaggieShan (approved):

app/src/code_review/code_review_view.rs
```diff
@@ -6915,10 +6930,12 @@ impl CodeReviewView {
     /// (e.g. a worktree branch whose tracking was auto-set to origin/master).
     fn pr_menu_item(&self, app: &AppContext) -> MenuItem<CodeReviewAction> {
         let diff_state = self.diff_state_model.as_ref(app);
+        let is_pr_info_refreshing = diff_state.is_pr_info_refreshing();
         if let Some(pr_info) = diff_state.pr_info().cloned() {
             MenuItemFields::new(format!("PR #{}", pr_info.number))
                 .with_icon(Icon::Github)
                 .with_on_select_action(CodeReviewAction::ViewPr(pr_info.url))
+                .with_disabled(is_pr_info_refreshing)
```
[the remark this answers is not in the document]
MaggieShan: No need at the moment

PR #9460, opened 2026-04-29
fix(mcp): persist token_received_at so OAuth refresh fires before expiry (#8863)

### LINKED ISSUE #8863: MCP OAuth accessTokens not being refreshed with refreshToken when they expire
### Pre-submit Checks

### Describe the bug

MCP OAuth flow works great but after logging in the first time if you wait 1hr (or whatever TTL the accessToken has) the MCP server disconnects and requires you to login again

### To reproduce

1. Connect an MCP server that uses OAuth that has a shortish TTL on accessTokens and supports refreshTokens
2. Wait until after that TTL and try to call a tool
3. Fails and needs to be reconnected

### Expected behavior

MCP tool call should get a 401 which indicates the token needs to be refreshed and Warp then uses the stored refreshToken to get a new accessToken and then runs the tool again

_No response_

### Operating system (OS)

macOS

26.3.1

### Shell Version

### Current Warp version

v0.2026.03.04.08.20.stable_02

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

None

david-engelmann:
## Description

Fixes #8863.

`PersistedCredentials` did not record when its token was received, and `install_persisting_credential_store` hardcoded `token_received_at: None` when seeding the rmcp credential store. The result: rmcp's pre-emptive refresh check at [`AuthorizationManager::get_access_token`](https://github.com/warpdotdev/rmcp/blob/c0f65dc/crates/rmcp/src/transport/auth.rs#L617-L633) — which requires both `expires_in` and `token_received_at` — was skipped for the lifetime of every OAuth-authenticated session.

After the first authorization, the cached access token was used past its TTL. When a request finally hit a 401, rmcp's [`handle_response`](https://github.com/warpdotdev/rmcp/blob/c0f65dc/crates/rmcp/src/transport/auth.rs#L710-L720) returns `AuthorizationRequired` without attempting recovery — the user sees their MCP server disconnect and is forced to re-authenticate. That matches the reproducer in #8863 exactly: "after logging in the first time if you wait 1hr… the MCP server disconnects and requires you to login again."

The cached-credentials path partially masked the bug because of the unconditional `auth_manager.refresh_token()` workaround at [`oauth.rs:272`](app/src/ai/mcp/templatable_manager/oauth.rs#L272) — that call writes back fresh credentials with `received_at` properly set, papering over the missing seed value for the duration of that session. The fresh-OAuth path has no such fallback, so users hit the bug on their first session.

### Fix

1. Add `token_received_at: Option<u64>` to `PersistedCredentials`. `#[serde(default)]` keeps existing on-disk credentials deserializable — they come back with `None` and the next refresh populates the field.
2. Forward `token_received_at` from `StoredCredentials` through `PersistingCredentialStore::save` to the persist channel so the value reaches secure storage.
3. Thread `token_received_at` as a parameter into `install_persisting_credential_store` and seed the inner store with it instead of hardcoded `None`.
4. Cached-creds call site (`make_authenticated_client`): pass the value loaded from `PersistedCredentials`.
5. Fresh-OAuth call site: stamp `token_received_at` to `now_epoch_secs()` at the point we save the credentials, and pass the same value into the seed.

Spec reference: [RFC 6749 §5.1 `expires_in`](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) and [§6 refresh-token semantics](https://datatracker.ietf.org/doc/html/rfc6749#section-6); rmcp's pre-emptive refresh implements the recommended client behavior of refreshing before expiry to avoid in-flight 401s.

### Things deliberately left alone

- The unconditional connect-time `auth_manager.refresh_token()` call at [oauth.rs:272](app/src/ai/mcp/templatable_manager/oauth.rs#L272) and the stale comment above it claiming rmcp's fork lacks expiry detection. The fork at the pinned revision (c0f65dc, "Fixes for oauth expiry and scopes") *does* now have proper detection per upstream [rust-sdk#680](https://github.com/modelcontextprotocol/rust-sdk/pull/680), so the workaround is now redundant. Removing it is a behavior change worth its own PR — out of scope here.
- 401-on-active-request retry. rmcp's `handle_response` still bubbles up `AuthorizationRequired` rather than refresh-and-retry; pre-emptive refresh covers the common case but not surprise expiries (clock skew, server-side revocation). A reactive layer is a follow-up.

## Testing

Six new unit tests in `app/src/ai/mcp/templatable_manager/oauth.rs`:

- `persisted_credentials_round_trip_through_serde_preserves_received_at` — value preservation.
- `persisted_credentials_deserializes_legacy_format_without_received_at` — **regression-protects existing users on upgrade**: credentials persisted by older Warp must still deserialize, with `received_at` defaulting to `None`. Failing this test would mean existing users lose their MCP OAuth tokens.
- `save_forwards_token_received_at_to_persist_channel` — the core regression test for the fix.
- `save_forwards_none_when_received_at_is_none` — defensive: don't substitute a fake value if rmcp passes `None`.
- `save_skips_persist_when_token_response_absent` — no regression of the existing `if let Some(token_response)` branch.
- `save_carries_forward_refresh_token_and_preserves_received_at` — combined check that the existing refresh-token carry-forward (RFC 6749 §6) and the new `received_at` propagation don't interfere.
- `now_epoch_secs_returns_recent_unix_time` — sanity check on the helper.

I was not able to run `./script/presubmit` locally (`warpui` build needs `xcrun metal` which requires full Xcode.app, only CommandLineTools is installed here). Trusting CI for `cargo clippy --workspace --all-targets --tests -- -D warnings` and `cargo nextest run --workspace`.

## Server API dependencies

N/A — client-only change.

## Agent Mode

CHANGELOG-BUG-FIX: Fix MCP OAuth servers disconnecting after access-token TTL expires; refresh tokens are now used to obtain new access tokens before expiry (#8863).


david-engelmann:
I'd rather refresh before expiration instead of after a failed call, let me know if this should be adjusted

vorporeal:
hey @peicodes - can you take a look at this?

peicodes (approved):
This is a great fix, thank you!

The part about the pre-emptive `refresh_token` request being now redundant is accurate, but yes, OOS

peicodes (approved):

peicodes:
This looks good! Let's stop making changes to it - I'll wait for CI to pass and merge.

david-engelmann:
> This looks good! Let's stop making changes to it - I'll wait for CI to pass and merge.

Sorry about that, I wasnt sure how the keeping up to master flow was handled, thought it was my responsibility

PR #9467, opened 2026-04-29
Link shared sessions to local interactive oz runs

jasonkeung:
## Description

Linear: [REMOTE-1326](https://linear.app/warpdotdev/issue/REMOTE-1326/link-shared-sessions-to-local-interactive-oz-runs)

When a user starts sharing a terminal whose active AI conversation is backed by a local Oz run (i.e. there is an `AmbientAgentTaskId` on the conversation, e.g. local child harness panes or `warp agent run`), the new shared session is now linked to that Oz run on the server. The Oz runs page can then surface the shared session link alongside the run, matching the behavior cloud agents already get for free.

### Approach
Cloud agents already link shared sessions to runs via the GraphQL mutation `updateAgentTask({ taskId, sessionId })`, which writes `ai_tasks.shared_session_id` and `ai_tasks.shared_session_uuid` (see `warp-server/model/ai_tasks.go:686`). This PR mirrors that on the client for local interactive runs:

1. In the existing `NetworkEvent::SharedSessionCreatedSuccessfully` handler in `terminal_manager.rs`, look up the active conversation on the terminal view.
2. If that conversation has a `task_id`, fire `AIClient::update_agent_task(task_id, session_id=Some(...))`.
3. Failure is non-fatal — sharing still works, the row just isn't linked. We log a warning.

The hook fires for every share entry point uniformly (share modal, "Start Remote Control" footer, command palette, right-click menu) because they all funnel through the same `attempt_to_share_session` → `Network::new` → `SharedSessionCreatedSuccessfully` chokepoint.

### What's intentionally NOT in this PR
- **No protocol-crate or server change.** The existing `updateAgentTask` mutation is fully reusable; no need to extend `SessionSourceType` or the gRPC `CreateSessionRequest.cloud_agent_run_id` field. Quota/ACL bypass for local-share-creation could be a follow-up if needed; user-owned local runs already pass the user's own quota.
- **No unlink on session end.** The link is intentionally durable — `shared_session_uuid` is denormalized in `ai_tasks` precisely so it survives `shared_sessions` GC. Mirrors cloud-agent behavior.
- **No backfill for pre-existing shared sessions.** Only sessions created after this lands get linked.

### Feature flag
Gated behind `FeatureFlag::LinkSharedSessionToLocalOzRun` (default-on for dogfood builds, off otherwise).

### Manual end-to-end
Per the local session-sharing testing notebook:
- Started session-sharing-server (`./script/server --features local_warp_server`)
- Started warp-server (`./script/server --with-client --with-local-session-sharing-server`)
- Started Warp client (`./script/run --features fast_dev,with_local_session_sharing_server`)

Then verified:
1. With an active local agent run (visible on the Oz runs page), clicking "Start Remote Control" on the agent pane creates a shared session and the corresponding `ai_tasks` row is updated with the new `shared_session_id` / `shared_session_uuid`.
2. Ending the session (via stop-sharing or inactivity) leaves the link intact on the run, matching the cloud-agent behavior.
3. With the feature flag off, no link is written.
4. Sharing a non-agent terminal (no `task_id` on the active conversation) is a no-op — no spurious mutation calls.

### Automated tests
No new unit tests added. Justification: the cloud-agent symmetric path (`SessionSourceType::AmbientAgent { task_id }` causing the harness to write the same link) has the same effective shape and is not unit-tested either; the codebase relies on integration/manual coverage for share-then-link flows. Wiring up a mock-`AIClient`-injected workspace test for this is a substantially larger lift than the change itself and would set a precedent that doesn't match the existing pattern. Behind a dogfood-only flag for safety while we exercise it.

Existing share-flow tests (`test_stop_sharing_session`, `test_stop_sharing_all_sessions_in_tab`, `test_tab_context_menu_share_session_items`, etc.) were re-run to confirm no regressions; the new code path is gated by the flag and these tests don't enable it.

## Server API dependencies

The `updateAgentTask` GraphQL mutation already exists and is in use by the cloud-agent harness today; this PR just adds another caller. No server change required.

## Agent Mode

CHANGELOG-IMPROVEMENT: Local interactive agent runs now link to their shared sessions on the Oz runs page, so you can jump from a run row back to the live shared session.

### DIFF
--- a/app/src/terminal/local_tty/terminal_manager.rs
+++ b/app/src/terminal/local_tty/terminal_manager.rs
@@ -120,6 +120,7 @@ use {
     crate::terminal::model::terminal_model::BlockIndex,
     crate::terminal::session_settings::NotificationsMode, nix::sys::termios::LocalFlags,
 };
+use crate::server::server_api::ServerApiProvider;
 
 type PtyController = writeable_pty::PtyController<mio_channel::Sender<Message>>;
 type RemoteServerController =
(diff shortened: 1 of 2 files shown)


jasonkeung:
closing to try cla again

PR #9471, opened 2026-04-29
Enhance programming language support in syntax highlighting by adding aliases and new languages

Abdalla-Eldoumani:
## Description

`ProgrammingLanguage::to_extension` in `app/src/ai/agent/mod.rs` is the choke point for syntax highlighting in AI blocks. When the agent emits a markdown code fence, the language token becomes a `ProgrammingLanguage`, and `to_extension()` is called to construct a `snippet.<ext>` path that `set_language_with_path` uses to look up the tree-sitter grammar (see `app/src/ai/blocklist/block.rs:2607-2617`, `block.rs:2675`, and `block/cli.rs:732-740`). If `to_extension` returns `None`, `set_language_with_path` is never called and the block renders as plain text.

The function omits a number of languages that the `languages` crate fully supports via `language_by_filename`, so common AI code-fence labels render with no syntax highlighting today:

- `jsx`, `tsx` — every React snippet
- `vue` — every Vue component snippet
- `xml` — every config / Android / .NET snippet
- `dockerfile` (and `docker`, `containerfile`) — every container snippet
- `starlark`, `objective-c` (and `objc`)

This PR adds those arms and folds in the most common markdown-fence aliases (`rs`, `py`, `js`, `ts`, `yml`, `kt`, `rb`, `golang`, `terraform`, `tf`) so coverage matches `language_by_filename`. The TODO comment at the top of the function (`INT-605: Refactor so we don't have to edit this function and the languages crate`) anticipates exactly this kind of sync.

The `Self::Shell(Bash | Zsh | Fish)` arms are intentionally left as-is — only PowerShell currently exposes an extension, and extending the other shells looks like a separate behavior change.

## Testing

- Added `test_programming_language_to_extension` in `app/src/ai/agent/mod_test.rs` covering 46 cases: every existing arm + every new arm + every alias + a round-trip note that each returned extension is one `language_by_filename` actually accepts. The new arms (`jsx`, `tsx`, `xml`, `vue`, `dockerfile`, `starlark`, `objective-c`, `objc`) and aliases (`rs`, `golang`, `py`, `js`, `ts`, `yml`, `c++`, `rb`, `kt`, `terraform`, `tf`, `docker`, `containerfile`) all fail on master and pass after the change.
- `cargo fmt -p warp_terminal -- --check` passes locally.

## Agent Mode

CHANGELOG-BUG-FIX: AI code blocks tagged `vue`, `xml`, `dockerfile`, `jsx`, `tsx`, `objective-c`, or `starlark` now render with syntax highlighting. Common aliases like `rs`, `py`, `js`, `ts`, `yml`, `kt`, `rb`, `golang`, `terraform`, and `docker` are also recognized.

### DIFF
--- a/app/src/ai/agent/mod.rs
+++ b/app/src/ai/agent/mod.rs
@@ -717,13 +717,19 @@ impl ProgrammingLanguage {
     #[cfg_attr(target_family = "wasm", allow(unused))]
     pub fn to_extension(&self) -> Option<&str> {
         match self {
+            // The arms below cover both canonical language names emitted by the agent (e.g.
+            // "rust", "kotlin") and common markdown code-fence aliases (e.g. "rs", "kt") to keep
+            // syntax highlighting working when the model uses either. The set of recognized
+            // languages here is kept in sync with `SUPPORTED_LANGUAGES` in the `languages` crate.
             Self::Other(language) => match language.to_lowercase().as_str() {
-                "rust" => Some("rs"),
-                "go" => Some("go"),
-                "python" => Some("py"),
-                "javascript" => Some("js"),
-                "typescript" => Some("ts"),
-                "yaml" => Some("yaml"),
+                "rust" | "rs" => Some("rs"),
+                "go" | "golang" => Some("go"),
+                "python" | "py" => Some("py"),
+                "javascript" | "js" => Some("js"),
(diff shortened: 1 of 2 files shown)


zachbai (approved):
LGTM

Issue #9233, opened 2026-04-28
Roadmap: May-June 2026


zachlloyd:
We've broken down our roadmap by:
- Warp team committed: stuff in the app that the Warp team itself plans on building this quarter
- Seeking community driver: stuff we would like to work with the community on this quarter

**Warp team committed**

ACP support (https://github.com/warpdotdev/warp/issues/7326)
- Support Warp as an Agent Client Protocol client so other coding agents can use Warp’s native agent UX
- Supporting ACP will unlock local model support through harnesses that support it (like OpenCode)

High-priority stability and performance fixes
- Warp team has a burndown list of high priority performance, stability and memory fixes that we are working on. Will share the link here shortly.

Improved worktree support + “Project” primitive
- Introduce a “project” primitive in Warp that lets users organize their work around projects
- Worktree navigation: easily switch to a specific worktree. Strawman is a worktree chip + maybe inline menu.
- Worktree management: easy deletion of a worktree when you're done with it. This can be done through the terminal but it's kind of tedious / there's something low-hanging here like a slash command menu to manage known worktrees
- LSP per worktree

Natural language detection improvements
- Improve the classifier we use for triggering our agent to have fewer false positives / negatives

Remote coding support
- Support file-tree, diff-view, LSP and other coding features on remote machines
- Persistent sessions locally and over ssh, pane detaching

Improved third-party agent support
- Continued investment in making Warp the best ADE to build with any coding agent.
- Example features are supporting conversation history, best-of-k, subagents 

**Seeking community drivers**

Tmux control mode
- Use tmux to drive Warp’s native tabs, panes and sessions

Support Warpifying more shells
- Support the native Warp input and block UX within more shells E.g. nushell, elvish

Improved code review
- Side by side diffs
- Selecting code within deleted sections

Warp CLI
- Control all aspects of the warp client app via CLI

Improved keybindings
- Improve our keybinding engine to be more flexible and configurable
- Improve the UI for settings and updating keybindings

Internationalization
- Translate UI text
- Better support for non-English keyboards

Custom prompt chips
- Arbitrary prompt customization

Theming improvements
- Per session / pane / tab / window theming

Jeffreyone:
三年了国际化终于要来了。

chan-yuu:
marked

BJXBDG:
> 三年了国际化终于要来了。

[...]

Issue #7043, opened 2025-08-10
Side by side diff view


meChrisReed:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Please let us configure the diff view between inline and side by side. I get that people with narrow monitors would prefer saving the space, but it is so much easier to read the side by side view, especially since we aren't writing the code ourselves.

### Operating system (OS)

macOS

### How important is this feature to you?

2

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

mvanhorn:
Opening a spec PR for this at #9479, referencing the May-June 2026 roadmap invitation in #9233. Happy to revise based on team feedback before any code follows.

captainsafia:
@mvanhorn Thanks for opening that spec proposal! I'll take a look at it and see what we can do to get some motion on it.

zachlloyd:
This is on our design team's radar and we will circle back shortly.

captainsafia:
@peterrajani is gonna take a stab at looking at some designs for this now that we've started to rev on a spec.

peterrajani:
@mvanhorn thanks again for taking this on. I've added some designs here for you:
https://www.figma.com/design/xfv2agz8ukUOdtOAzTOV4H/Split-diff-view?node-id=6001-24731&t=R8qGUoRbs0oD6h5u-1

and also recorded this loom to hopefully make everything as clear as possible:
https://www.loom.com/share/45967d6f53fa44fcbeb2062537be0e98

kevinchevalier:
I'm going to review the spec in relation to the mocks today.

kevinchevalier:
@mvanhorn I've marked this ready to implement, but I wasn't sure if you want me to kick off an oz implementation, or if you wanted to start from scratch.

PR #9479, opened 2026-04-29
specs(GH7043): side-by-side diff layout in code review and AI block diffs

mvanhorn:
Spec PR for #7043 (side-by-side diff view), listed under "Seeking community drivers" in the May-June 2026 roadmap (#9233).

`product.md` covers the user-facing behavior: a new `code.editor.diff_layout` setting (`inline` / `side_by_side`), exposed via the existing diff toolbar's View Options menu, applied to both AI block-list diffs and the Code Review pane. Inline stays the default; side-by-side adds a hunk-aligned two-pane renderer with synchronized vertical scrolling and per-pane selection.

`tech.md` walks the implementation: a new `DiffLayout` enum next to `DisplayMode`, a `SideBySideDiffView` sibling to `InlineDiffView`, a padding-row API on `CodeEditorView`, a `HunkAlignment` model that consumes the existing unified-diff parser, a `ScrollSyncModel` for cross-pane scroll, and View Options menu integration in `CodeReviewDiffMenu`. Every `DisplayMode` destructure site is checked for compile parity. Behind a `SideBySideDiffLayout` feature flag.

Open questions and follow-up scope are listed at the end of `tech.md` for SME triage.

### DIFF
--- a/specs/GH7043/product.md
+++ b/specs/GH7043/product.md
@@ -41,17 +41,18 @@ The two related open issues - #9017 (word wrap in diff and Markdown) and #9040 (
 
 1. When `code.editor.diff_layout` is unset or `inline`, every diff in Warp renders identically to today. No regression to the default form.
 
-2. When `code.editor.diff_layout` is `side_by_side`, every diff in Warp renders with the baseline on the left and the modified file on the right.
+2. When `code.editor.diff_layout` is `side_by_side`, every diff in Warp renders with the baseline on the left and the modified file on the right, with one explicit exception: AI block-list diffs in `InlineBanner` display (the small "Suggested fixes" banner that appears below a command block) continue to render inline regardless of the setting. The banner's vertical budget (typically 100-160px) is too small to read two columns, and the banner already collapses to inline at narrow widths today. Invariant 16 below makes this gating observable: the View Options menu's Layout radio group is hidden in the banner. Every other surface, including the full Code Review pane and the embedded AI block-list diff (`Embedded { max_height }` mode), honors the setting.
    - The split is a 50/50 vertical split with a single 1-pixel divider in the panel chrome.
    - The two panes share a horizontal scrollbar appearance and inherit horizontal scrollbars independently, since wrap-vs-no-wrap behavior is per-pane.
    - The baseline pane shows the file content prior to the diff, with deletions visible. The modified pane shows the post-diff file content, with additions visible. Unchanged context lines appear in both panes at the same vertical position.
 
-3. Hunk-aligned padding keeps corresponding lines aligned across panes:
+3. Hunk-aligned padding keeps corresponding lines aligned across panes. The alignment algorithm pairs deleted lines with added lines within a hunk so that a modification renders on a single shared row across the two panes:
    - Unchanged lines: rendered at the same row on both sides with the same content.
-   - Modifications (a `-` line replaced by a `+` line in the inline rendering): the deleted line renders on the left at row N; the added line renders on the right at row N. The right pane shows nothing for the deleted-only side; the left pane shows nothing for the added-only side.
-   - Pure additions (`+` lines with no corresponding `-` lines): the added lines render on the right at consecutive rows; the left pane shows blank padding at the same rows so the next unchanged context line still aligns.
-   - Pure deletions (`-` lines with no corresponding `+` lines): the deleted lines render on the left at consecutive rows; the right pane shows blank padding at the same rows.
+   - Modifications: within each hunk, deleted lines and added lines are paired in order. For a hunk of `D` deleted lines followed by `A` added lines, the first `min(D, A)` pairs render on shared rows: the deleted line on the left at row N, the added line on the right at row N. The shared row is what makes "before/after" review readable.
+   - Excess deletions (when `D > A`): the trailing `D - A` deleted-only lines render on the left at consecutive rows; the right pane shows blank padding at the same rows. This is the pure-deletion case for unpaired suffixes.
+   - Excess additions (when `A > D`): the trailing `A - D` added-only lines render on the right at consecutive rows; the left pane shows blank padding at the same rows. This is the pure-addition case for unpaired suffixes.
    - The padding rows are visually the same height as a normal line and use the same gutter as the matching pane, so vertical positions on the two panes always agree.
+   - Word-level or character-level highlighting on a paired modification row is out of scope (see Non-goals); a row pair shows the deleted line in full on the left and the added line in full on the right.
 
 4. Synchronized vertical scrolling:
    - A scroll wheel event in either pane drives both panes by the same delta.
@@ -108,5 +109,6 @@ The two related open issues - #9017 (word wrap in diff and Markdown) and #9040 (
     - Memory overhead for `SideBySide` is a second `CodeEditorView` per visible diff, which the existing inline path already constructs and destroys when navigating between diffs in code review. The number of simultaneously alive editors does not grow more than 2x in steady state.
 
 16. Feature flag gating:
-    - The change ships behind a `SideBySideDiffLayout` `FeatureFlag` (`app/src/features/`) that defaults to off in shipping builds and on in dogfood builds. Once stabilized, the flag is removed and the setting becomes the user-facing control.
-    - When the flag is off, the View Options menu does not show the Layout radio group, and the setting is treated as `Inline` regardless of stored value.
+    - The change ships behind a `SideBySideDiffLayout` `FeatureFlag` defined in `crates/warp_features/src/lib.rs` (the canonical flag enum, re-exported from `app/src/features.rs`) that defaults to off in shipping builds and on in dogfood/preview builds. Once stabilized, the flag is removed and the setting becomes the user-facing control.
+    - When the flag is off, the View Options menu does not show the Layout radio group, the Settings page does not render the Layout widget, and the setting is treated as `Inline` regardless of stored value.
+    - The Layout radio group is also hidden in the AI block-list `InlineBanner` display per invariant 2 (the banner is too small for two columns); the rest of the AI block-list diff surface shows the radio group when the flag is on.
--- a/specs/GH7043/product.md
+++ b/specs/GH7043/product.md
@@ -4,7 +4,7 @@ Roadmap reference: https://github.com/warpdotdev/warp/issues/9233 ("Improved cod
 Figma: none provided
 
 ## Summary
-Add a side-by-side diff layout as an alternative to the existing inline diff. The user picks the layout from the diff toolbar's View Options menu and the choice is persisted as a `code.editor.diff_layout` setting. Inline remains the default. Side-by-side renders the baseline on the left and the modified file on the right, with hunk-aligned padding so corresponding lines sit on the same row, and synchronized vertical scrolling so wheel and cursor navigation in either pane drives both. Both diff surfaces in Warp - the AI block-list embedded diff and the full Code Review pane - honor the setting.
+Add a side-by-side diff layout as an alternative to the existing inline diff. The user picks the layout in Settings -> Code under a new "Diff layout" subsection, and the choice is persisted as a synced `code.editor.diff_layout` setting. Inline remains the default. Side-by-side renders the baseline on the left and the modified file on the right, with hunk-aligned padding so corresponding lines sit on the same row, and synchronized vertical scrolling so wheel and cursor navigation in either pane drives both. Every surface that renders a diff in Warp - the Code Review pane, AI block-list diffs, and inline banner diffs - honors the setting.
 
 ## Problem
 Warp renders every file diff as a single-column inline view: deletions on `-` rows, additions on `+` rows, both interleaved. The renderer is `app/src/code/inline_diff.rs` (the `InlineDiffView` used by AI blocks) and the Code Review pane in `app/src/code_review/code_review_view.rs`. Both consume `app/src/code/diff_viewer.rs::DiffViewer` and currently have no layout knob.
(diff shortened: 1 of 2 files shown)


zachlloyd:
Excited to take a look at this!! Thank you for driving.

specs/GH7043/product.md
```diff
@@ -0,0 +1,114 @@
+# Side-by-Side Diff Layout in the AI Block-List and Code Review Pane - Product Spec
+GitHub issue: https://github.com/warpdotdev/warp/issues/7043
+Roadmap reference: https://github.com/warpdotdev/warp/issues/9233 ("Improved code review - Side by side diffs", listed under "Seeking community drivers")
+Figma: none provided
+
+## Summary
+Add a side-by-side diff layout as an alternative to the existing inline diff. The user picks the layout from the diff toolbar's View Options menu and the choice is persisted as a `code.editor.diff_layout` setting. Inline remains the default. Side-by-side renders the baseline on the left and the modified file on the right, with hunk-aligned padding so corresponding lines sit on the same row, and synchronized vertical scrolling so wheel and cursor navigation in either pane drives both. Both diff surfaces in Warp - the AI block-list embedded diff and the full Code Review pane - honor the setting.
```
captainsafia: I think because we intend on this option extending in the diff view for multiple surfaces (code review panel, AI blocks) then we should make the setting something you configure via the settings menu instead of the diff toolbar.

mvanhorn:
@captainsafia thanks for the settings-menu call - pivoted in 302ee9d. The setting now lives in Settings -> Code under a new "Diff layout" subsection and applies to every diff surface (Code Review pane, AI block-list, inline banner). The diff toolbar no longer exposes the layout toggle.

Same commit also addresses the oz-for-oss findings:
- CRITICAL hunk-alignment: replaced the deferred two-row algorithm with a v1 collapsed delete/add pairing algorithm (with pseudocode + edge cases).
- IMPORTANT items: feature-flag location fixed to crates/warp_features (was wrong before), PaneAwareDiffViewer trait with caller routing table, pane content construction bypassing apply_diffs_if_any in the side-by-side path, per-pane interaction state (baseline read-only), DiffHunk-based alignment API, ScrollSource::Mirrored suppression token, settings page widget integration, InlineBanner now switches to SideBySideDiffView constrained to viewport (no fallback), accessibility section added.

Ready for re-review when you have cycles. /oz-review

kevinchevalier:
Hey @mvanhorn Thanks for diving into this. We've put this off for too long!

I'm going to take a look at the specs today, and in addition one of our designers is going to look at this as well. We have some old designs, but they need some fresh 👀

specs/GH7043/product.md
```diff
@@ -0,0 +1,118 @@
...
+
+- Side-by-side is the de-facto standard for diff review across GitHub, GitLab, Phabricator, JetBrains IDEs, VS Code, and tools like Beyond Compare. Users with wide displays expect it, and the absence shows up repeatedly in user feedback, including the request that opened this issue: "since we aren't writing the code ourselves," the reviewer needs to read both versions in parallel rather than reconstructing them mentally from a single column.
+- Warp's review surface is shaped by AI-generated edits more than by hand-written commits. An AI agent regularly produces diffs that touch dozens of unrelated regions in one block. In an inline layout, a 30-line modification with reordered code is hard to follow, because the same logical block appears as `-` lines in one place and `+` lines in another. Side-by-side puts those next to each other.
+
+The two related open issues - #9017 (word wrap in diff and Markdown) and #9040 (auto line wrap in diff) - assume a single diff layout and argue about wrapping inside it. They are out of scope for this spec but become more useful once a side-by-side layout exists, since wrap behavior interacts with column width.
+
+## Goals
```
kevinchevalier, line 21: Can we add a goal around editing in the code review pane? Currently, we can edit any line that is either added or unmodified. So in this case, I think we should say that edits can be made only in the right hand pane.

specs/GH7043/product.md
```diff
@@ -0,0 +1,118 @@
...
+   - Excess deletions (when `D > A`): the trailing `D - A` deleted-only lines render on the left at consecutive rows; the right pane shows blank padding at the same rows. This is the pure-deletion case for unpaired suffixes.
+   - Excess additions (when `A > D`): the trailing `A - D` added-only lines render on the right at consecutive rows; the left pane shows blank padding at the same rows. This is the pure-addition case for unpaired suffixes.
+   - The padding rows are visually the same height as a normal line and use the same gutter as the matching pane, so vertical positions on the two panes always agree.
+   - Word-level or character-level highlighting on a paired modification row is out of scope (see Non-goals); a row pair shows the deleted line in full on the left and the added line in full on the right.
+
+4. Synchronized vertical scrolling:
+   - A scroll wheel event in either pane drives both panes by the same delta.
+   - A cursor-up or cursor-down keystroke in either pane moves only that pane's cursor; the other pane scrolls without moving its cursor so the corresponding line stays in view.
```
kevinchevalier: I'm wondering if we should only have a cursor in the right panel. Since we don't allow selection of deleted selections, and we don't allow editing (likely) in the left panel, I wonder if we should only have the concept of a cursor in the right panel.

mvanhorn:
Sounds great — happy to iterate on the specs once your team and designer have taken a look. Holler if you want any of it sketched in a different shape.

specs/GH7043/product.md
```diff
@@ -0,0 +1,118 @@
+# Side-by-Side Diff Layout in the AI Block-List and Code Review Pane - Product Spec
+GitHub issue: https://github.com/warpdotdev/warp/issues/7043
+Roadmap reference: https://github.com/warpdotdev/warp/issues/9233 ("Improved code review - Side by side diffs", listed under "Seeking community drivers")
+Figma: none provided
+
+## Summary
+Add a side-by-side diff layout as an alternative to the existing inline diff. The user picks the layout in Settings -> Code under a new "Diff layout" subsection, and the choice is persisted as a synced `code.editor.diff_layout` setting. Inline remains the default. Side-by-side renders the baseline on the left and the modified file on the right, with hunk-aligned padding so corresponding lines sit on the same row, and synchronized vertical scrolling so wheel and cursor navigation in either pane drives both. Every surface that renders a diff in Warp - the Code Review pane, AI block-list diffs, and inline banner diffs - honors the setting.
```
kevinchevalier: Let's start by only using the side-by-side view, if selected, in the code review panel.

specs/GH7043/tech.md
```diff
@@ -0,0 +1,507 @@
...
+`CodeEditorView` and the underlying model gain a render-only padding-row concept: a row that takes up vertical space, has a gutter, and renders empty content. The public API:
+
+```rust
+impl CodeEditorView {
+    pub fn set_padding_rows(&mut self, rows: Vec<PaddingRow>, ctx: &mut ViewContext<Self>);
+}
+
+#[derive(Clone, Debug)]
+pub struct PaddingRow {
+    pub line_index: usize,
```
kevinchevalier: What does `line_index` mean here? There are a few things it might mean:
- A line number of the baseline content.
- A line number of the modified content.

But note that a gap wouldn't correspond to a line number in either of these numbering systems. So maybe this corresponds to the line number that comes after this gap?

This could also be an index that does not correspond to a line number at all, but every line in the buffer has an index. One hard part is that when content is edited, we would have to recompute that index for every line after the edit.

kevinchevalier (changes requested):
One big architectural question that we should answer is whether this should be modeled as a `SideBySideDiffView` with two `CodeEditorView`s, or whether the `CodeEditorView` should be renderable as a side-by-side diff.

I think that there are some interactions that would be simpler if this was all internal to `CodeEditorView`. 

- Edits in the "modified" section might require updates to the baseline section to occur immediately, if a gap needs to be resized.

I see three possible solutions here:

- `SplitView` with 2 `CodeEditorView`s inside.
- `CodeEditorView` with 2 `EditorWrapper`s inside.
- `CodeEditorView` with 1 `EditorWrapper` that handles rendering the left and right sides.

Let's chat a bit about the best solution. I think that is a really important decision before we start.

kevinchevalier:
@kevinyang372 Is the code editor expert.

Do you have intuition on whether implementing the side-by-side diff would work best by:

- SplitView with 2 CodeEditorViews inside.
- CodeEditorView with 2 EditorWrappers inside.
- CodeEditorView with 1 EditorWrapper that handles rendering the left and right sides.

And then do you think it would be possible to use the same `RenderState` to render both? Or is that a recipe for disaster?

I'm leaning towards "CodeEditorView with 1 EditorWrapper that handles rendering the left and right sides with 2 `RenderState`s" at the moment.

mvanhorn:
@kevinchevalier thanks for the scope narrow and the architecture direction. Pushed e4538ab. Summary of changes:

- product.md: v1 narrowed to Code Review pane only. AI block-list and inline banner diffs are deferred to v2 with explicit non-goals.
- product.md: encoded right-pane-only editing and cursor invariants per your inline asks (#3170987216, #3171286536). Baseline pane is read-only with no insertion cursor; deleted-line ranges remain non-selectable.
- tech.md: pivoted from SideBySideDiffView to CodeEditorView with one EditorWrapper rendering both sides and two RenderStates. Scroll sync, gap resizing, hit testing, and modified-pane edits stay internal to one view.
- tech.md: replaced ambiguous `line_index` with a `RowIndex { Baseline(usize) | Modified(usize) | Gap { after_row } }` enum so the alignment producer can carry gap rows that do not correspond to either sides source line numbering (resolves your #3209320878 ask).
- tech.md: Code Review integration retargeted to `CodeReviewEditorState` / `LocalCodeEditorView`. Dropped the `InlineDiffView` migration step.

New Open Questions section flags three items pending @kevinyang372 confirm:
1. CodeEditorView + 1 EditorWrapper + 2 RenderStates is the right shape (vs SplitView+2-views or CodeEditorView+2-wrappers).
2. Intra-component scroll sync is feasible inside one EditorWrapper.
3. Screen-reader and tab-order can be expressed inside a single view exposing two panes.

Happy to re-pivot if @kevinyang372 prefers a different option once theyve had a look.

kevinchevalier:
@mvanhorn Thanks for making those updates. After discussing with Kevin, I think you were correct the first time. Here is our convo. Sorry for all of the back and forth on this, I just wanted to make sure we were heading in the correct direction before starting.

Kevin Yang: 

> The way I will model this is we have two editor views
> 
> - A select only editor view rendering the base (pre-edit) content
> - A real editor view rendering with a global buffer entry. It has a decoration config to render deleted lines as gaps instead of temporary blocks
> 
> We will have a bridge side by side wrapper that keeps the following states in sync:
> 
> - Hidden lines
> - Scroll position
> - Find states
> 
> 
> The benefit of this approach is the editor view can mostly be kept as is — there is no need for its internal model to handle whether its target is the left or right split view. We will also get selection on deleted sections for free now they're not a temporary blocks

Kevin Chevalier: 

>   I think the read-only editor still needs real time updates from the global buffer to update modified sections. Like if you delete a line in the editable editor, then we need to mark that line as deleted in the read-only editor. If we add a line we need to render a gap in the read-only view.
> 
> I guess this could work if they are both updating render states from the same global buffer.

Kevin Yang:

> Yeah I think they need to share diff states too
> 
> I don't think they should share the same buffer tho. The ‘buffer’ of the read only editor is the base content, not the latest content

mvanhorn:
Pushed d93fd86 reverting to Kevin Yang's two-editor-views + bridge model. Summary:

- tech.md: two `CodeEditorView` instances (select-only baseline + normal modified). New `SideBySideDiffBridge` wrapper owns hidden lines, line-anchored scroll, find state, and shared `HunkAlignment` diff state. Buffers stay separate per @kevinyang372's note ('the buffer of the read-only editor is the base content, not the latest content').
- `RowIndex { Baseline | Modified | Gap }` enum preserved (reviewer ask #3209320878).
- Open Questions narrowed to three: state-shape (split `CodeReviewEditorState` vs sub-states), diff-state ownership (bridge-owned vs container-owned), and cross-pane find UX (match-by-match vs pane-at-a-time).
- product.md unchanged scope; one tweak -- deleted-line ranges in the baseline pane are now selectable for copy, since they're real rows in the baseline buffer rather than synthesized blocks under the new model.

Ready for a closer look @kevinyang372.

specs/GH7043/tech.md
```diff
@@ -0,0 +1,559 @@
...
+
+Cursor movement belongs to the modified side. When cursor navigation changes the modified row, `corresponding_row` computes the nearest baseline row for visibility but does not move a baseline cursor. Hunk navigation updates the shared scroll anchor and the modified-side cursor.
+
+Resolution: the bridge synchronizes views by row anchors, not by matching pixel offsets. If implementation needs event source tagging to avoid recursive updates, that state belongs in the bridge.
+
+### 11. Code Review pane integration
+
+The Code Review path uses the existing Code Review editor host, not per-file `InlineDiffView` instances. The integration:
+
+- Represent side-by-side editor state as either two `CodeReviewEditorState` slices or one slice with `baseline_sub_state` and `modified_sub_state`. The two child views must have distinct lifecycles even when the parent reducer entry point stays shared.
```
kevinchevalier, line 406: I think this should be one `CodeReviewEditorState`.

specs/GH7043/tech.md
```diff
@@ -0,0 +1,559 @@
...
+- `app/src/code/diff_viewer.rs`: trait helpers; unchanged because `DiffLayout` is a new orthogonal axis.
+- `app/src/ai/blocklist/inline_action/code_diff_view.rs`: unchanged in v1; continues inline rendering.
+- `app/src/ai/blocklist/block/view_impl/output.rs`: match on `DisplayMode::FullPane`; unchanged.
+
+Every Code Review site that assumes one editor state should be checked. Existing write, save, cursor, and accept/reject paths should keep targeting the modified view. Rendering, comments, find, selection, and accessibility call sites should route through the bridge when side-by-side is active.
+
+## Open questions
+
+1. State management shape: do we keep `CodeReviewEditorState` as one slice with `baseline_sub_state` and `modified_sub_state`, or split into two parallel `CodeReviewEditorState` slices? Single-slice keeps the existing Code Review reducer signature; split-slice cleanly separates the two views' lifecycles.
+2. Diff result ownership: does the bridge wrapper own the diff state directly, or is the diff state hoisted into the parent Code Review container? Bridge-owned diff is encapsulated; container-owned diff is reusable by other Code Review consumers.
```
kevinchevalier, line 551: I think that the diff state would still be owned by the "modified" `CodeEditorView`.

specs/GH7043/tech.md
```diff
@@ -0,0 +1,559 @@
...
+3. Find state UX: when "Next match" crosses panes, does focus jump match-by-match, or does it stay in one pane until all matches there are visited? Confirm with @kevinyang372 whether this is designed behavior or a bridge-internal detail.
```
kevinchevalier, line 552: I don't feel strongly about this. I would prefer if it jumped from pane to pane, finding matches from top to bottom.

But if finding all matches in one pane is significantly simpler, I think that is fine.

kevinchevalier (commented):
I think this will finish all of the open questions! I'm going to mark this ready to implement.

mvanhorn:
@kevinchevalier The spec has been in the agreed state since d93fd86 (two editor views + SideBySideDiffBridge per your convo with Kevin Yang), and your follow-up marked it ready to implement. The changes-requested review from May 8 is still active though, so the stale bot is threatening auto-close. Could you dismiss it or re-approve so this doesn't get swept? Happy to start on the implementation PR whenever this lands.

kevinchevalier (approved):

kevinchevalier:
Would you mind rebasing on the current `master` branch? I think some old code is failing the test file name check.

mvanhorn:
Rebased onto current master (0af94c7). The branch only carries the two spec files (specs/GH7043/product.md + tech.md), so the rebase was clean - the failing test-filename check should clear on the rerun.

mvanhorn:
Appreciate you taking this, @kevinchevalier. Side-by-side diff specs for code review and AI blocks should be a big readability lift.

PR #9489, opened 2026-04-29
Drop unneeded fetch-depth: 0 from two workflows

sebryu:
## Summary

Two workflow checkouts request the full repo history via `fetch-depth: 0` but never use it. Falling back to the default shallow clone speeds them up at no behavioral cost.

## Analysis

I scanned every `fetch-depth: 0` site in `.github/`:

| Site | Why set | Needed? |
|---|---|---|
| `check_approvals.yml:19` | `git merge-base` + `git diff` between base/head | **Yes** — merge-base across long-lived branches needs deep history. Already mitigated with `filter: blob:none`. |
| `create_release.yml:1623` | Feeds `warpdotdev/generate-changelog` | **Yes** — changelog walks tag history. |
| `feature_flag_cleanup.yml:30` (analyze) | Agent prompt instructs `git blame` of `Cargo.toml` | **Yes** — blame needs full history. |
| `feature_flag_cleanup.yml:161` (cleanup) | None — agent prompt only says "clean up references" / "remove from enum" / "modify code" | **No — removed** |
| `feature_flag_cleanup.yml:218` (create_pr) | Applies a patch then runs `peter-evans/create-pull-request` | Probably no, but kept conservative — `peter-evans` only requires deep history when it must include prior local commits. Worth a follow-up. |
| `delete_release.yml:95` | Job only does `git push HEAD:refs/heads/<new>` then `git push --delete <old>` | **No — removed** |

## Changes

- `feature_flag_cleanup.yml` — removed `fetch-depth: 0` from the cleanup job. The Oz agent edits source files at HEAD; no git history operations occur in the prompt.
- `delete_release.yml` — removed `fetch-depth: 0` from the delete job. Renaming a branch is two server-side pushes; only the tip commit is needed.

## Test plan

- [ ] CI passes on this PR.
- [ ] Next nightly run of `feature_flag_cleanup.yml` succeeds end-to-end (analyze → cleanup → create_pr).
- [ ] Next manual `delete_release.yml` invocation succeeds (rename + delete).

### DIFF
--- a/.github/workflows/feature_flag_cleanup.yml
+++ b/.github/workflows/feature_flag_cleanup.yml
@@ -158,7 +158,6 @@ jobs:
         uses: namespacelabs/nscloud-checkout-action@938f5d2d403d6224d9a0c0dc559b1dae09c2ede4 # v8.1.1
         with:
           ref: ${{ needs.analyze.outputs.analyzed_sha }}
-          fetch-depth: 0
 
       - name: Setup Rust cache
         uses: namespacelabs/nscloud-cache-action@a90bb5d4b27522ce881c6e98eebd7d7e6d1653f9 # v1.4.2
--- a/.github/workflows/feature_flag_cleanup.yml
+++ b/.github/workflows/feature_flag_cleanup.yml
@@ -158,6 +158,7 @@ jobs:
         uses: namespacelabs/nscloud-checkout-action@938f5d2d403d6224d9a0c0dc559b1dae09c2ede4 # v8.1.1
         with:
           ref: ${{ needs.analyze.outputs.analyzed_sha }}
+          fetch-depth: 0
 
       - name: Setup Rust cache
         uses: namespacelabs/nscloud-cache-action@a90bb5d4b27522ce881c6e98eebd7d7e6d1653f9 # v1.4.2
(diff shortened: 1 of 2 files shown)


.github/workflows/feature_flag_cleanup.yml
```diff
@@ -158,7 +158,6 @@ jobs:
         uses: namespacelabs/nscloud-checkout-action@938f5d2d403d6224d9a0c0dc559b1dae09c2ede4 # v8.1.1
         with:
           ref: ${{ needs.analyze.outputs.analyzed_sha }}
-          fetch-depth: 0
```
alokedesai, line 161: I believe this one is necessary in order for the agent to run `git blame`

alokedesai (commented):
Thanks for contributing! I'd prefer if we kept `fetch-depth` for `feature_flag_cleanup.yml` but I support removing it for `delete_release.yml`

sebryu:
Thanks @alokedesai — agreed. Restored `fetch-depth: 0` on the cleanup job so the agent can still walk history if it decides to.

While I was in there, I also dropped `fetch-depth: 0` from the **create_pr** job (the third checkout in the same workflow). That job only:
1. Checks out at `analyzed_sha`
2. Applies the cleanup patch from the artifact
3. Hands the dirty working tree to `peter-evans/create-pull-request`, which makes one commit on top of HEAD and pushes a new branch

No history walk anywhere — `peter-evans/create-pull-request` works fine with a shallow clone since it only commits + pushes from the tip. Happy to revert if you'd rather keep this PR scoped to just `delete_release.yml`.

alokedesai:
Thanks! Could we keep isolated to just `delete_release.yml` for now?

sebryu:
Done — reverted the feature_flag_cleanup.yml change in 3bd890d. PR is now scoped to delete_release.yml only.

alokedesai (approved):
Thanks!

PR #9494, opened 2026-04-29
Reduce RAM requirements for release builds.

vorporeal:
## Description
Release builds (specifically the `release-lto` profile, which is `release` + `lto = "thin"`) have been OOM-killing rustc on CI. The dominant memory cost during ThinLTO + codegen is the full per-variable/type DWARF that `debug = true` emits, since rustc has to hold that DWARF for every transitively-linked rlib in memory while merging modules.

Dropping `[profile.release]` to `debug = 1` keeps line tables — which is all we need to symbolicate panics and Sentry stack traces with file/line info — and discards the type/variable DIEs that drive most of the memory pressure. The tradeoff is that attaching a debugger to a release binary will no longer show locals or rich type info; post-mortem backtraces are unaffected.

This is the lowest-side-effect change I could find that materially reduces peak rustc memory; if it isn't enough on its own, the next levers would be `codegen-units = 1` or backing off ThinLTO.

## Testing
Relying on CI to confirm the release build succeeds without OOMing. No runtime behavior changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -378,12 +378,15 @@ derivative = "2.2.0"
 parquet = { version = "55.0.0", features = ["arrow"] }
 rmcp = { git = "https://github.com/warpdotdev/rmcp.git", rev = "c0f65dc441af7d714b9c453ac5e7ef641451abe3" }
 
-# Comment this to disable building with debug symbols
-# Building with debug symbols generates a dsym which we can use to get
-# better stack traces in sentry. But there is a cost of about 20% binary size
-# increase.
 [profile.release]
-debug = true
+# Use line-tables-only (debug = 1) rather than full debuginfo (debug = 2 /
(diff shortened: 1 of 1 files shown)


vorporeal:
* **#9494** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9494?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9494?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

kevinchevalier (approved):
LOL at switching the setting from true -> 1.

vorporeal:
> LOL at switching the setting from true -> 1.

yeah fun fact - `"true" == 2` here

PR #9501, opened 2026-04-29
fix(bootstrap): warn before sudo and document install steps

### LINKED ISSUE #9421: ./scripts/bootstrap invokes sudo and installs system-wide dependencies without user acknowledgement
### Pre-submit Checks

### Describe the bug

Congrats on open sourcing Warp!  Here's a few things I think are concerning with the bootstrap process of the codebase that I have spotted so far:
 
- According to `README.md` before building Warp on a local machine, developers have to execute `./scripts/bootstrap` to install the necessary dependencies.  First and foremost, `README.md` lacks documentation explaining what the script actually does and why building needs this, common practice would be listing out the dependencies the repo needs and leave installation to the developer's discretion. 
- The `.scripts/bootstrap` conditionally execute different scripts depending on the user's platform and the underlying scripts provide no explanation of what changes they're making to the user's machine, what is being installed and how they are installing these dependencies.
- Sudo prompt triggered upon running the script with no explanation to user.  Potentially caused by switching primary Xcode version
https://github.com/warpdotdev/warp/blob/3f0ac51bc92f136c1aa7693eb07799e9a2e6f7c0/script/macos/bootstrap#L13
- At least on macOS, the script installs [cargo-bundle](https://github.com/burtonageo/cargo-bundle) from its Git repository, locked to a specific commit hash.  Although personally I do not know how to exploit this, this seems scummy and could become a supply chain risk compared to using the official install from Cargo's registry.  `cargo-bundle` is also a common dependency many developers have installed, replacing the existing installation without information the user seems bad practice to me and could introduce issues if the developer is aware of the version change and goes on to use this pinned version in other repositories.  As far as I know, the latest version of `cargo-bundle` on the registry already supports `--profile`, there's no need to install it from a separate commit.
https://github.com/warpdotdev/warp/blob/3f0ac51bc92f136c1aa7693eb07799e9a2e6f7c0/script/macos/bootstrap#L37 
- The macOS script also seems to install a lot of dependencies from various Git repositories instead of just conventionally install from registry.  Examples: https://github.com/warpdotdev/warp/blob/3f0ac51bc92f136c1aa7693eb07799e9a2e6f7c0/script/install_cargo_test_deps#L10
 https://github.com/warpdotdev/warp/blob/3f0ac51bc92f136c1aa7693eb07799e9a2e6f7c0/script/install_cargo_release_deps#L28
It would have been okay if these dependencies were installed at project scope with `cargo add` instead of overwriting the global installation with `cargo install`
- The macOS script also installs a bunch of system-wide dependencies such as `sentry-cli` and `jq` with Homebrew without asking the user, the user might have different ways of installing these dependencies such as using `curl -sL https://sentry.io/get-cli/ | sh` to install the sentry-cli instead.  Again, the script is installing these things without asking the user and proper documentation of what it's doing which is bad practice
https://github.com/warpdotdev/warp/blob/3f0ac51bc92f136c1aa7693eb07799e9a2e6f7c0/script/macos/bootstrap#L42-L49
- The script installs [Docker](https://www.docker.com/) if the user does not have it installed and assumes the user wants Docker by default, they might prefer something like [OrbStack](https://orbstack.dev/).  Not sure how Docker is used during the build process but it seems unneeded.
https://github.com/warpdotdev/warp/blob/3f0ac51bc92f136c1aa7693eb07799e9a2e6f7c0/script/macos/bootstrap#L54
- Script retrieves user's gcloud auth token and signs user in if it's not present without proper explanation of how it's going to be used and why it's needed
https://github.com/warpdotdev/warp/blob/3f0ac51bc92f136c1aa7693eb07799e9a2e6f7c0/script/macos/bootstrap#L62

### To reproduce

1. Run `./scripts/bootstrap`
2. Observe

### Expected behavior

README.md or another markdown file should explain what the script does and what dependencies it's going to install.  Would be better if developers were given the choice to install certain dependencies with an alternative package manager.

_No response_

### Operating system (OS)

macOS

15.7.4

### Shell Version

zsh 5.9 (arm64-apple-darwin24.0)

### Current Warp version

Not Relevant

### Regression

### Recent working Warp date

_No response_

### Additional context

Image

No

No, this issue happens in Warp and other terminals.

None

tautik:
- new `script/warp_sudo`: echoes each privileged command, prompts `[y/N]` from `/dev/tty`, shells to real `sudo` on confirm. `WARP_BOOTSTRAP_YES=1` or non-tty stdin skips.
- `script/bootstrap`: static `--help` install list dropped. `-y`/`--yes` exports `WARP_BOOTSTRAP_YES=1` for children.
- ten direct `sudo` call-sites in `script/{macos,linux}/bootstrap` + `script/linux/install_{build,runtime,test}_deps` switched to `warp_sudo`. windows unchanged.
- `README.md`: short "what `./script/bootstrap` does" subsection pointing at the platform scripts as the source of truth.

closes #9421.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,5 +1,16 @@
 #!/usr/bin/env bash
 
+# Bootstrap dispatches to the platform-specific install scripts. Each step
+# that needs root sources `script/warp_sudo` and asks for confirmation
+# before running. Pass -y / --yes (or set WARP_SKIP_SUDO_PROMPT=1) to skip
+# the prompts in unattended environments.
+
+for arg in "$@"; do
+  case "$arg" in
+    -y|--yes) export WARP_SKIP_SUDO_PROMPT=1 ;;
+  esac
+done
+
 OS_TYPE="$(uname -s)"
 
 if [[ "$OS_TYPE" = "Darwin" ]]; then
--- a/script/warp_sudo
+++ b/script/warp_sudo
@@ -1,0 +1,39 @@
+#!/bin/sh
+#
+# script/warp_sudo - source this file to define the `warp_sudo` function.
+#
+# `warp_sudo` is a thin wrapper around `sudo` that echoes the command it
+# is about to run and asks for confirmation before invoking it. This gives
+# the user a chance to see (and decline) every privileged action taken
+# during bootstrap, which is otherwise opaque.
+#
+# Skip the prompt by setting WARP_SKIP_SUDO_PROMPT=1 (also set automatically
+# by `./script/bootstrap -y` / `--yes`) or by running with stdin not
+# attached to a tty (e.g. CI). The prompt itself is read from /dev/tty,
+# so pipes like `curl ... | warp_sudo apt-key add -` continue to work.
+
+warp_sudo() {
+  printf '\n>>> requires root: sudo %s\n' "$*" >&2
+
+  if [ "${WARP_SKIP_SUDO_PROMPT:-}" = "1" ]; then
+    sudo "$@"
+    return
+  fi
+
+  # Probe for a usable controlling terminal. `[ -r /dev/tty ]` is unreliable
+  # on macOS (the device file is always world-readable, even for processes
+  # with no controlling tty), so actually try to open it in a subshell.
+  if (: </dev/tty) 2>/dev/null; then
+    # Write the prompt to /dev/tty (not stderr) so it stays visible even
+    # when stderr is redirected, matching where the response is read from.
+    printf '    continue? [y/N] ' >/dev/tty
+    reply=""
+    read -r reply </dev/tty || reply=""
+    case "$reply" in
+      y|Y|yes|YES) sudo "$@" ;;
+      *) printf '    aborted.\n' >&2; return 1 ;;
+    esac
+  else
+    sudo "$@"
+  fi
+}
--- a/script/linux/bootstrap
+++ b/script/linux/bootstrap
@@ -2,9 +2,11 @@
 
 set -e
 
+. "$PWD/script/warp_sudo"
+
 # Update the apt cache before installing dependencies.  It can be slow, so the install scripts
 # don't do it (to avoid doing it more than once).
-sudo apt update -y
+warp_sudo apt update -y
 
 # Install all dependencies needed to build, run, and test Warp.
 "$PWD"/script/linux/install_test_deps
--- a/script/linux/install_build_deps
+++ b/script/linux/install_build_deps
@@ -4,6 +4,8 @@
 
 set -e
 
+. "$PWD/script/warp_sudo"
+
 # Define some control sequences for text formatting.
 red="\033[0;31m"
 reset="\033[0m"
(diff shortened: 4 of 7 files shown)


script/bootstrap
```diff
@@ -1,7 +1,66 @@
...
+  case "$arg" in
+    -y|--yes) ASSUME_YES=true ;;
+    -h|--help)
+      cat <<'USAGE'
+Usage: ./script/bootstrap [-y|--yes] [-h|--help]
+
+Installs the development dependencies needed to build, run, and test
+Warp on the current platform. Several steps require sudo:
+
+  macOS    sudo xcode-select; brew (jq, sentry-cli, clang-format,
```
vorporeal: my primary concern with using a usage blurb like this is that the list of things installed will get out-of-date.

from a "product requirements" perspective, i feel like what we'd want here is something like: when we're about to execute a command that will require super-user privileges, echo out the command and ask for user confirmation before continuing.

if we had some helper script that provided a, say, `warp_sudo` alias to do this, we could use that instead of a  direct `sudo` invocation throughout these scripts.

thoughts?  do you think that's overkill?

vorporeal (commented):
thanks for opening this PR!  definitely agree we can improve the bootstrap script - it was built for internal users and we overlooked tuning it up before we open-sourced the repository.

script/bootstrap
tautik: not overkill — agreed, a static list will get out of date fast. happy to pivot to a small `warp_sudo` helper that echoes each command and asks before running it, with an env var / non-tty escape hatch for ci. pushing this on the same branch shortly.

tautik:
pushed the rewrite on the same branch (force-push).

tautik:
@cla-bot check

README.md
```diff
@@ -65,6 +57,20 @@ To build and run Warp from source:
...
+platform-specific script — `script/macos/bootstrap`,
+`script/linux/bootstrap`, or `script/windows/bootstrap.ps1` — which
+remains the single source of truth for exactly what gets installed.
+
+Several steps require `sudo` (e.g. `xcode-select` on macOS,
+`apt-get install` on Linux). Each privileged command is echoed and
+asks for confirmation before running, so you always see what you're
+agreeing to. Pass `-y` / `--yes` or set `WARP_BOOTSTRAP_YES=1` to
+auto-confirm in unattended environments such as CI.
+
```
vorporeal: i'm not sure this is necessary information to include in the top-level repository readme; i'd remove.

script/warp_sudo
```diff
@@ -0,0 +1,39 @@
...
+#
+# Skip the prompt by setting WARP_BOOTSTRAP_YES=1 (also set automatically
+# by `./script/bootstrap -y` / `--yes`) or by running with stdin not
+# attached to a tty (e.g. CI). The prompt itself is read from /dev/tty,
+# so pipes like `curl ... | warp_sudo apt-key add -` continue to work.
+
+warp_sudo() {
+  printf '\n>>> requires root: sudo %s\n' "$*" >&2
+
+  if [ "${WARP_BOOTSTRAP_YES:-}" = "1" ]; then
```
vorporeal: given that some of these scripts can and are invoked from outside the context of `./script/bootstrap`, i'd say we should name this `WARP_SKIP_SUDO_PROMPT`.

vorporeal (commented):
couple last minor comments

tautik:
dropped the readme subsection and renamed `WARP_BOOTSTRAP_YES` -> `WARP_SKIP_SUDO_PROMPT` per your review. ready for another look.

vorporeal (approved):
looks good; thanks for the PR and working through the changes with me!

PR #9508, opened 2026-04-29
[APP-3792] specs for remote codebase indexing

moirahuang:
## Description
Specs for remote codebase indexing based off of https://docs.google.com/document/d/1G4sDjVwatYMQC-BHsR3OvtlX6De27HYel7H8SSJQyM8/edit?usp=sharing

## Testing
N/A spec only

## Agent Mode


specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,251 @@
...
+- Call the app server directly for retrieval using the current root hash, then call the daemon only to hydrate fragment bytes or full file context.
+
+Backend responsibilities:
+- Store and retrieve Merkle-tree/index data and embeddings keyed by hashes.
+- Answer `get_relevant_fragments(root_hash, query, repo_metadata, embedding_config)`.
+- Rerank candidate fragments.
+- Provide codebase context config to both local client indexing and daemon-side remote indexing.
+
+### Why the client needs the root hash
+The client needs the current ready root hash so retrieval can be client → app server instead of client → daemon → app server. The root hash is the backend lookup key for the synced index; it is enough for retrieval, while avoiding a full tree sync to the client. The client should not need fragment bytes or the complete Merkle tree to decide search candidates.
```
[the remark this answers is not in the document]
kevinyang372: This is not really an issue

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+Alternative shape: the daemon receives `SearchCodebase`, calls `get_relevant_fragments`, hydrates fragments, reranks, and returns final locations.
+
+Why rejected for v1:
+- Adds an SSH hop to every retrieval query even though the client already has a valid app-server auth path.
+- Makes retrieval unavailable when the daemon's backend connection is flaky even if the client can reach the backend.
+- Couples agent retrieval latency to the remote link more than necessary.
+
+## 3. Proposed changes
+### 3.1 Extract daemon-compatible indexing code
+Create a daemon-compatible indexing crate (for example `crates/codebase_index`) by extracting the current `crates/ai/src/index/full_source_code_embedding/` modules and the minimal supporting types they need. Re-export from `crates/ai` so current client call sites can keep their public paths until cleanup.
```
kevinyang372: Do we need to do this? The server daemon model (which lives in app) already depends on the ai crate no?

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+
+The extracted crate should contain:
+- `CodebaseIndexManager`, `CodebaseIndex`, `sync_client`, `store_client`, `snapshot`, `merkle_tree`, `chunker`, `fragment_metadata`, `changed_files`, and related tests.
+- `CodeContextLocation` / indexing location types needed by retrieval.
+- Persistence adapters split so client builds can keep SQLite-backed metadata while the daemon uses a lightweight file-backed metadata store.
+
+Avoid depending on unrelated `crates/ai` agent, MCP, terminal, or UI modules in the daemon. The daemon should only pull the indexing implementation, syntax/chunking dependencies, remote-compatible filesystem/repo metadata dependencies, and GraphQL types needed by `StoreClient`.
+
+### 3.2 Add daemon-side `StoreClient`
+Add a daemon-side implementation of the `StoreClient` trait. It should use the APP-3801 daemon credential and a small HTTP/GraphQL client rather than depending on client-side `ServerApi`.
```
kevinyang372: ooc what's the reason for not using client side ServerApi

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+- Sends the same backend operations the local client sends today: config fetch, Merkle tree sync, embedding generation, intermediate-node update, cache population, relevant-fragment retrieval only if a future daemon-retrieval path needs it, and reranking only if a future daemon-retrieval path needs it.
+- Classifies errors into at least unauthenticated, backend unreachable, backend rejected, and internal/unknown so status UI can distinguish actionable failures.
+- Redacts tokens from logs and never persists them.
+
+For v1 sync, daemon-side retrieval methods may still be implemented because the trait requires them, but the normal remote retrieval path should use the client's `ServerApi` for `get_relevant_fragments` and `rerank_fragments`.
+
+### 3.3 Add daemon-side index cache and startup bootstrap
+The daemon keeps a notion of cached codebase indexes. Use an identity-scoped directory under the remote-server cache root, for example:
```
kevinyang372: Does this need to be identity scoped? Index cache should be the same for all user no?

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+2. For known ready repos, emit `Ready` status with root hash to connected clients.
+3. For known repos without a valid snapshot, emit `Not enabled` or `Failed` depending on whether the user had previously enabled indexing.
+4. When a client reconnects, respond to status requests from this cache even before the watcher has emitted new events.
+
+Snapshot parsing should follow local behavior: if a snapshot is incompatible or corrupt, delete it and rebuild from scratch rather than leaving the repo permanently failed.
+
+### 3.4 Add remote-server protocol messages
+Extend `crates/remote_server/proto/remote_server.proto` with request/response and push messages for remote indexing. Names can be adjusted during implementation, but the protocol needs these concepts:
+
+- `EnableCodebaseIndexing { repo_path, user_preferences, feature_flags }`
```
kevinyang372: What is the reasoning for this message? The indexing trigger should be entirely driven by the client right

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+- `GetCodebaseIndexStatus { repo_path }`
+- `DropCodebaseIndex { repo_path }`
+- `RetryCodebaseIndexing { repo_path }`
```
kevinyang372: Shouldn't this just be IndexCodebase { repo_path }

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+- `HydrateCodebaseFragments { repo_path, content_hashes }`
```
kevinyang372: What is this for?

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+- file-backed persisted metadata,
+- daemon-side snapshot base directory,
+- daemon-side `StoreClient`,
+- `BulkFilesystemWatcher`,
+- remote-compatible repo metadata / detected-repo dependencies already used by the daemon.
+
+In `app/src/remote_server/server_model.rs`, add handler arms for the new RPCs:
+- `EnableCodebaseIndexing`: check cache first; if miss or stale, enqueue/build index and immediately push queued/indexing status.
+- `GetCodebaseIndexStatus`: return the current cached status.
+- `DropCodebaseIndex`: remove metadata/snapshot, stop watcher registration for that repo, and push disabled/not-enabled status.
```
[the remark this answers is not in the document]
kevinyang372: Yeah we should consider this

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,254 @@
...
+9. Client caches status and enables `SearchCodebase`.
+
+### Previously seen repo
+1. Client asks daemon for status after navigation or reconnect.
+2. Daemon finds a ready cached index.
+3. Daemon returns/pushes `Ready { root_hash, embedding_config }`.
+4. Client enables retrieval without full rebuild.
+
+### Startup with known repos
+1. Daemon loads metadata and snapshots during startup.
```
kevinyang372, line 307: What's our cache invalidation strategy?

specs/APP-3792/TECH.md
moirahuang: yeah true, updated

specs/APP-3792/TECH.md
moirahuang: ah yeah that's fair, simplified to just `IndexCodebase`

specs/APP-3792/TECH.md
moirahuang: yup, updated

specs/APP-3792/TECH.md
moirahuang: right now only the daemon has the remote snapshot metadata and filesystem access needed to map the root hash back to file ranges and bytes

specs/APP-3792/TECH.md
moirahuang: updated w a cache invalidation section

specs/APP-3792/TECH.md
moirahuang: we need to handle the token but i updated this section to be more specific that we can have a wrapper around client side `ServerApi`

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,283 @@
...
+- Adds an SSH hop to every retrieval query even though the client already has a valid app-server auth path.
+- Makes retrieval unavailable when the daemon's backend connection is flaky even if the client can reach the backend.
+- Couples agent retrieval latency to the remote link more than necessary.
+
+## 3. Proposed changes
+### 3.1 Reuse daemon-compatible indexing code
+For v1, wire the daemon path to the existing `crates/ai/src/index/full_source_code_embedding/` implementation instead of creating a new crate up front. The remote-server daemon lives in `app`, and `app` already depends on `ai`, so the simplest implementation can reuse `CodebaseIndexManager`, `CodebaseIndex`, `sync_client`, `store_client`, `snapshot`, `merkle_tree`, `chunker`, `fragment_metadata`, `changed_files`, and their existing tests directly.
+
+The daemon wiring still needs daemon-specific adapters:
+- file-backed persisted metadata instead of the client's SQLite-backed metadata,
```
kevinyang372: Why use an ad-hoc file-backed persistence instead of reusing SQLite infra? That _should_ be available to the oz cli already no?

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,283 @@
...
+- daemon-side snapshot base directory,
+- remote-compatible filesystem/repo metadata dependencies,
+- daemon-compatible `StoreClient` auth plumbing.
+
+Keep daemon entrypoints narrow so the remote-server path depends only on indexing, syntax/chunking, remote filesystem/repo metadata, and backend GraphQL types. Avoid introducing daemon dependencies on unrelated `crates/ai` agent, MCP, terminal, or UI modules. Extracting the indexing implementation into a smaller crate such as `crates/codebase_index` remains a follow-up if v1 shows unacceptable daemon binary size or dependency coupling.
+
+### 3.2 Add daemon-compatible `StoreClient`
+`app/src/server/server_api/ai.rs` already implements `StoreClient` for the client-side `ServerApi`; reuse that codebase GraphQL operation and conversion logic where possible. Do not require the daemon to instantiate the full client `ServerApiProvider` or depend on the client `AuthState` refresh/event path, because daemon auth comes from APP-3801 and `ServerModel`.
```
kevinyang372: Hmmm related to my previous comment -- why do we need a wrapper instead of directly re-using ServerApi? I would like to understand the concrete blockers

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,283 @@
...
+5. Auth identity changes clear the client-side `RemoteCodebaseIndexModel` cache and reconnect through the identity-scoped daemon path, but they do not delete the shared machine-local index cache.
+6. If the backend rejects, cannot find, or no longer authorizes a previously ready root hash for a specific identity, mark that identity's repo status failed with an actionable reason and require `IndexCodebase` to rebuild, resync, or re-associate the shared cache for that identity.
+
+### 3.4 Add remote-server protocol messages
+Extend `crates/remote_server/proto/remote_server.proto` with request/response and push messages for remote indexing. Names can be adjusted during implementation, but the protocol needs these concepts:
+
+- `IndexCodebase { repo_path }`
+- `GetCodebaseIndexStatus { repo_path }`
+- `DropCodebaseIndex { repo_path }`
+- `HydrateCodebaseFragments { repo_path, content_hashes }`
```
kevinyang372: Can we name this GetFragmentMetadataFromHash? It's a bit clearer what this actually does

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,283 @@
...
+4. Backend config or embedding-config changes mark affected shared snapshots stale and re-run the necessary embedding/sync work with the new config.
+5. Auth identity changes clear the client-side `RemoteCodebaseIndexModel` cache and reconnect through the identity-scoped daemon path, but they do not delete the shared machine-local index cache.
+6. If the backend rejects, cannot find, or no longer authorizes a previously ready root hash for a specific identity, mark that identity's repo status failed with an actionable reason and require `IndexCodebase` to rebuild, resync, or re-associate the shared cache for that identity.
+
+### 3.4 Add remote-server protocol messages
+Extend `crates/remote_server/proto/remote_server.proto` with request/response and push messages for remote indexing. Names can be adjusted during implementation, but the protocol needs these concepts:
+
+- `IndexCodebase { repo_path }`
+- `GetCodebaseIndexStatus { repo_path }`
```
kevinyang372: What do we need this status request for? I thought the updates are initiated by the server side as push notifications

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,283 @@
...
+- settings controls,
+- remote tool advertisement,
+- remote dispatch branch in `SearchCodebaseExecutor`.
+
+The daemon should not independently check the feature flag. If it receives a valid `IndexCodebase` request from an authenticated client build, it should perform the requested work. This avoids requiring daemon/client flag state to be perfectly synchronized.
+
+## 4. End-to-end flows
+### New repo
+1. Client observes remote navigation into repo.
+2. Client asks daemon for index status.
```
kevinyang372: Instead of the client asking daemon on indexing status for every repo, I think the daemon should report the set of repos it has indexed after initialization

This allows us to build up the initial set of remotely indexed repos on the client side for things like settings page as well

specs/APP-3792/TECH.md
```diff
@@ -0,0 +1,283 @@
...
+
+### 3.6 Fetch and respect server-backed config on the daemon
+The daemon should call `codebase_context_config` through its `StoreClient` before sync work and at the cadence expected by the local implementation. Server-backed values such as embedding config, embedding cadence, generation batch size, and sync batching should be owned by the backend and respected on the remote host.
+
+The client should evaluate user/client-controlled gates before sending `IndexCodebase`, such as whether the remote-indexing feature flag is enabled, whether the user accepted indexing, and whether persistence is allowed. Do not send client-owned feature or preference values for the daemon to reinterpret, and do not use client-sent values for server-owned tuning knobs when the daemon can fetch them directly.
+
+### 3.7 Client-side state and UI model
+Add a client singleton such as `RemoteCodebaseIndexModel` that subscribes to `RemoteServerManager` events and tracks:
```
kevinyang372, line 219: Can we expand on what internal states and APIs RemoteCodebaseIndexModel should expose? I think it's worth diving into the weeds here as this ensures the upstream callers can easily adapt the remote case

specs/APP-3792/TECH.md
moirahuang: ah yeah good call, adding more info on how to hook that up for the daemon

specs/APP-3792/TECH.md
moirahuang: the main blocker is that `ServerApi` uses `get_or_refresh_access_token` which is different from the daemon token. for example, here https://github.com/warpdotdev/warp-internal/blob/ed4deec871cc2aefcd8d77ef8300ad878351d0c3/app/src/server/server_api.rs#L541

are you wondering if we should refactor `ServerApi` to handle the daemon token so we don't have to have a wrapper?

specs/APP-3792/TECH.md
moirahuang: combining this w the comment below, i think we should still keep this for if we are missing a codebase index status e.g. a user navigates to a new repo that wasn't part of the initial set of repos indexed after initialization

specs/APP-3792/TECH.md
moirahuang: yeah a bulk initial set of repos makes sense, will update

specs/APP-3792/TECH.md
moirahuang: updated so index cache is same for all users

kevinyang372 (approved):

PR #9509, opened 2026-04-29
Skip AgentTips with unresolved keybinding placeholders

MaggieShan:
## Description
* Reported in https://warpdev.slack.com/archives/C08KTPNQN65/p1777499214403939 
* Adds a check in  `AgentTip::is_tip_applicable` such that tips that expect a keybinding but don't have one are removed from the displayed list 
* Also updates the tip singleton model to refresh based on dependent settings and keybinding changes

## Testing
Verified locally with only the voice tip - when setting is disabled or when keybinding is not available, the tip is not applicable/not shown 
https://www.loom.com/share/dcfcf6a0b8b24be8aea1ca24dd27ddc9

## Agent Mode

### DIFF
--- a/app/src/ai/agent_tips.rs
+++ b/app/src/ai/agent_tips.rs
@@ -19,7 +19,7 @@ use std::path::Path;
 use std::sync::LazyLock;
 use std::time::Duration;
 use warpui::keymap::Keystroke;
-use warpui::r#async::SpawnedFutureHandle;
+use warpui::r#async::{SpawnedFutureHandle, Timer};
 use warpui::{AppContext, Entity, ModelContext, SingletonEntity};
 
 /// Trait for tip implementations that can be displayed to users.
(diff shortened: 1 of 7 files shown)


app/src/ai/agent_tips.rs
```diff
@@ -423,6 +423,12 @@ impl AITip for AgentTip {
         if matches!(self.kind, AgentTipKind::Handoff) {
             return AISettings::as_ref(app).is_cloud_handoff_enabled(app);
         }
+        // Tips whose description references a keybinding placeholder should only be shown
+        // when the keybinding is actually configured, so we never display the raw
+        // "<keybinding>" string to users.
+        if self.description.contains("<keybinding>") && self.keystroke(app).is_none() {
+            return false;
```
harryalbert, line 430: nit: is this extra check necessary given the other fixes in this PR?

if it is, might be worthwhile to log like a warning or error if it's an unexpected path

app/src/ai/agent_tips.rs
```diff
@@ -506,7 +512,44 @@ impl AITipModel<AgentTip> {
...
+                    .any(|tip| tip.description == current_tip.description);
+
+                !still_in_pool || !current_tip.is_tip_applicable(None, ctx)
+            })
+            .unwrap_or(true);
+
+        if should_replace {
+            let new_tip = Self::pick_random_applicable_tip(&self.tips, None, ctx);
+            if new_tip.is_some() || self.current_tip.is_some() {
+                self.current_tip = new_tip;
```
harryalbert, line 548: would we want to do this if `new_tip` is none but `current_tip` is some?

harryalbert (approved):
nice!

app/src/ai/agent_tips.rs
MaggieShan: Yes this check is still necessary since checking if voice is enabled vs. if there is a keystroke available are two separate checks - this one is for the latter. This is also an expected path for when a user has explicitly removed a setting's keybinding so I also don't think we need a log

app/src/ai/agent_tips.rs
MaggieShan: Yes we still do - as long as current tip is some() in this block, it means it's currently invalid and should either be replaced by a new tip or not be rendered at all

PR #9519, opened 2026-04-29
Don't restore vertical-tabs panel when feature is disabled

### LINKED ISSUE #9505: Window is unclickable after upgrade vertical_tabs_panel layout fails and blocks clicks (Linux/Wayland)
### Pre-submit Checks

### Describe the bug

After I upgraded to v0.2026.04.27.15.32.stable_02 on Arch Linux (Hyprland, Wayland), the Warp window stopped reacting to mouse clicks. The cursor still moves and buttons light up when I hover over them, but clicks do nothing. Keyboard shortcuts still work.

In ~/.local/state/warp-terminal/warp.log I see these warnings repeating:

[WARN] Failed to compute position for stack child element. Skipping child.
       X: Err("Position not set for \"workspace_view:vertical_tabs_panel\""),
       Y: Err("Position not set for \"workspace_view:vertical_tabs_panel\"").
[WARN] Dismiss underlay was clicked but no handler was set!

It looks like the vertical tabs panel cannot be placed correctly, but its dismiss layer stays on top of the window and catches all the clicks. That would explain why hover works but clicks don't.

As a workaround, I closed Warp and moved the workspace database aside so Warp could create a new one:

pkill -f /opt/warpdotdev/warp-terminal/warp
cd ~/.local/state/warp-terminal
mv warp.sqlite      warp.sqlite.bak
mv warp.sqlite-wal  warp.sqlite-wal.bak
mv warp.sqlite-shm  warp.sqlite-shm.bak

After that, clicks work again. So it seems the new version does not migrate the old vertical_tabs_panel layout data correctly, and this leaves a layer that blocks the clicks.

### To reproduce

1. Use Warp on the previous stable version with vertical tabs enabled, and close it normally so the layout state is saved.
2. Upgrade Warp to v0.2026.04.27.15.32.stable_02 (Arch `warp-terminal` package).
3. Open Warp again.
4. Move the mouse over the UI: hover effects work (buttons highlight).
5. Try to click anywhere in the window: nothing happens. Keyboard shortcuts still work, but the mouse is unusable.
6. Open ~/.local/state/warp-terminal/warp.log and check the latest entries: the "vertical_tabs_panel" position warnings and the "Dismiss underlay was clicked but no handler was set" message appear repeatedly.

### Expected behavior

After upgrading, the Warp window should react to mouse clicks normally. The new version should migrate or reset the old vertical tabs layout data safely, without leaving an invisible layer on top of the window that blocks clicks. Hover and click events should both reach the UI as before the upgrade.

_No response_

### Operating system (OS)

Linux

Arch Linux (rolling)

### Shell Version

bash 5.3.9

### Current Warp version

v0.2026.04.27.15.32.stable_02-1

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

webdevtodayjason:
## Description

`Workspace::initial_vertical_tabs_panel_open` (`app/src/workspace/view.rs:3726`) honored the persisted `window_snapshot.vertical_tabs_panel_open` value unconditionally when vertical tabs were not usable in the current workspace. If a user previously had vertical tabs enabled with the panel open, then later set `appearance.vertical_tabs.enabled = false`, restoring the workspace booted with `vertical_tabs_panel_open = true` while the panel itself never rendered (rendering is gated on `should_default_open`).

The panel didn't render — but its `Dismiss` underlay did. The underlay's `prevent_interaction_with_other_elements` swallowed every click while hover and keyboard input still worked, leaving the window visually responsive but functionally dead. The repeated `workspace_view:vertical_tabs_panel` position-cache warnings and `Dismiss underlay was clicked but no handler was set!` log lines reported in the issue come from this state.

### Fix

When `should_default_open` is `false` (feature flag off OR `use_vertical_tabs = false`), return `false` from `initial_vertical_tabs_panel_open` instead of falling through to the persisted snapshot.

```rust
// before
if should_default_open && *show_vertical_tab_panel_in_restored_windows {
    true
} else {
    window_snapshot.vertical_tabs_panel_open  // restores stale `true` even when disabled
}

// after
if !should_default_open {
    false  // <-- new
} else if *show_vertical_tab_panel_in_restored_windows {
    true
} else {
    window_snapshot.vertical_tabs_panel_open
}
```

Unrelated existing arms (`Empty`, `FromTemplate`, `TransferredTab`, etc.) already returned `should_default_open` directly so they were never affected — the bug was specific to the `Restored` branch.

## Testing

Adds `test_vertical_tabs_panel_closed_when_disabled_even_if_persisted_open` in `app/src/workspace/view_test.rs`, which reproduces the exact upgrade sequence from the issue:

1. Enable vertical tabs (feature flag + setting), open the panel, snapshot the workspace.
2. Disable the `use_vertical_tabs` setting.
3. Restore from the open-panel snapshot.
4. Assert `workspace.vertical_tabs_panel_open == false`.

Without the fix this test fails — the workspace boots with `vertical_tabs_panel_open == true` and the click-eating dismiss underlay would paint at runtime.

The existing tests `test_vertical_tabs_panel_visibility_restores_from_window_snapshot` and `test_vertical_tabs_panel_restored_open_when_show_in_restored_windows_enabled` both keep `use_vertical_tabs = true`, so they exercise the unchanged `else` branches and continue to pass.

**Local presubmit note:** `cargo fmt --check` passes locally. Full `./script/presubmit` is blocked by Metal Toolchain activation on this machine (downloaded asset not yet active); CI will run the full clippy/nextest matrix.

## Server API dependencies

This PR does not depend on any server API changes.

## Agent Mode

CHANGELOG-BUG-FIX: Fix Warp window becoming unclickable on Linux/Wayland after upgrading with a persisted vertical-tabs panel state — restoring no longer leaves a stale dismiss underlay above the window when vertical tabs are disabled.

Fixes #9505


vorporeal:
reassigning to @zachbai, who built vertical tabs (and is better positioned to know if this is the right fix)

app/src/workspace/view.rs
zachbai, line 3806: consider simplifying to 

`should_default_open && (TabSettings..show_vertical_tab_panel_in_restored_windows || window_snapshot.vertical_tabs_panel_open)`?

zachbai (approved):

PR #9528, opened 2026-04-30
[APP-3106] Client: preserve user query modes in CloudMode

vkodithala:
## Description

With my server changes in [this](https://github.com/warpdotdev/warp-server/pull/10678) PR, we now accept an explicit user query mode in ambient agent queries sent via the public API. This PR passes `UserQueryMode`through ambient agent requests so CloudMode preserves `/plan` and `/orchestrate` semantics, and shares prefix highlighting helpers across local and CloudMode query rendering.

Keeps local and CloudMode prompt bubbles visually consistent, and includes a drive-by change to fix `/orchestrate` prefix highlighting in rendered queries in the blocklist.

## Testing

Tested manually. Demo here.

## Server API dependencies

- [ ] Is this change necessary to make the client compatible with a desired server API breaking change?
- [ ] Does this change rely on a new server API?
- [ ] Is this change enabling the use of a server API on client channels that rely on the production server?

## Agent Mode

### DIFF
--- a/app/src/ai/agent/mod.rs
+++ b/app/src/ai/agent/mod.rs
@@ -2327,11 +2327,37 @@ impl PassiveSuggestionTrigger {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
 pub enum UserQueryMode {
     #[default]
     Normal,
     Plan,
     Orchestrate,
+}
+
+pub fn extract_user_query_mode(query: String) -> (String, UserQueryMode) {
+    if let Some(query) = commands::strip_command_prefix(&query, commands::PLAN_NAME) {
+        (query, UserQueryMode::Plan)
+    } else if let Some(query) = commands::strip_command_prefix(&query, commands::ORCHESTRATE_NAME) {
+        (query, UserQueryMode::Orchestrate)
+    } else {
+        (query, UserQueryMode::Normal)
+    }
+}
+
+/// Reconstructs the display form of a user query that has been stripped via
+/// [`extract_user_query_mode`], by re-prepending the slash-command prefix
+/// associated with [`UserQueryMode`].
+///
+/// This is the inverse of [`extract_user_query_mode`] and the canonical way
+/// for UI to render a stored `(mode, query)` pair so the displayed prompt
+/// always matches what the user originally submitted.
+pub fn display_user_query_with_mode(mode: UserQueryMode, query: &str) -> String {
+    match mode {
+        UserQueryMode::Normal => query.to_owned(),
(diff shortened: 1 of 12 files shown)


vkodithala:
* **#9528** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9528?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9528?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/blocklist/block/view_impl/common.rs
```diff
@@ -3402,13 +3403,28 @@ pub struct UserQueryProps<'a> {
...
     pub font_properties: &'a Properties,
 }
+pub(crate) fn user_query_mode_prefix_highlight_len(mode: UserQueryMode) -> Option<usize> {
+    match mode {
+        UserQueryMode::Normal => None,
+        UserQueryMode::Plan => Some(commands::PLAN.name.len()),
+        UserQueryMode::Orchestrate => Some(commands::ORCHESTRATE.name.len()),
+    }
+}
+
```
vkodithala, line 3413: Extracts prefix highlighting logic into a shared helper + includes a drive-by change to properly highlight "orchestrate" query modes in prefixes

app/src/ai/agent/mod.rs
```diff
@@ -2327,13 +2327,39 @@ impl PassiveSuggestionTrigger {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
bnavetta: Are we currently (de)serializing this anywhere? If so, this rename may break existing data

app/src/pane_group/pane/terminal_pane.rs
```diff
@@ -1371,6 +1371,8 @@ fn handle_terminal_view_event(
                             };
                             let spawn_request = SpawnAgentRequest {
                                 prompt: request.prompt,
+                                // Agents spawned during orchestrations are always run in normal mode.
```
bnavetta, line 1374: @cephalonaut @kjankov I think this is the right behavior, but FYI this is now an optional knob on new agent runs

app/src/ai/agent/mod.rs
vkodithala: We actually don't serialize this anywhere else; the only other place that `UserQueryMode`'s sent over the wire is via the multi-agent API, but this is proto-based and the query mode is converted directly into a proto type. I do agree that we should be more explicit here though to guard unexpected behavior in the future. 

Recent commit should remove this and add a custom serializer for the public API

bnavetta (approved):

PR #9536, opened 2026-04-30
[Windows] fix session restoration for maximized and fullscreen windows

acarl005:
## Description

We currently have a bug with session restoration where maximized windows won't restore into the maximized state. This is due to the "vertical adjustment" code which tries to make sure the title bar is on screen (it moves the window downward if that isn't the case). Without getting too much into the details, on Windows when windows are maximized, winit reports the physical size of the window _as if_ it had the native window decorations, even when it doesn't. It treats the native title bar as being off the screen and represents that with a negative origin (above the screen). This triggers `fn maybe_adjust_window_vertically` to return a value which triggers the caller to try to move the window which makes Windows un-maximize the window. This simple guard fixes the problem

## Testing

1. Run Warp on Windows
2. Make sure session restoration is enabled
3. Maximize the window
4. quit warp
5. Start warp again
6. The window should be maximized but it isn't

On this branch, that bug no longer occurs.


lucieleblanc (approved):
Thanks for the fix!

PR #9553, opened 2026-04-30
Attach image drops in CLI-agent sessions instead of typing the path (drag-and-drop)

### LINKED ISSUE #9545: Drag-and-drop image into CLI agent (Claude Code) inserts file path instead of attaching the image
### Pre-submit Checks

### Describe the bug

When a CLI agent (e.g. Claude Code) is the active foreground process in a Warp tab, dragging an image file (e.g. a screenshot) onto the terminal inserts the file path as text into the agent's prompt, rather than attaching the image as image context.

Pasting the same image with Cmd+V (with the image bytes on the clipboard) attaches the image correctly. The two flows should converge for image files dropped while a CLI agent is active.

### To reproduce

1. Open a Warp tab and start a Claude Code session.
2. Take a screenshot (Cmd+Shift+4) so the screenshot file lands on the Desktop.
3. Drag the screenshot file onto the Warp window with the Claude Code session active. Reproduces both when dragging from Finder and when dragging the macOS screenshot preview thumbnail directly.
4. Observe: the file path is typed into Claude Code's prompt as plain text.
5. For comparison, copy the same screenshot to the clipboard (Preview → Edit → Copy, or Cmd+Shift+Ctrl+4 to copy directly), then Cmd+V into the same Claude Code session — the image is attached.

### Expected behavior

Drag-and-drop of an image file into a tab with an active CLI agent session should attach the image as image context (the same outcome as the Cmd+V flow), not insert the file path as text.

### Operating system (OS)

macOS

macOS 26.4.1 (25E253)

### Current Warp version

v0.2026.04.22.08.46.stable_03

### Regression

### Additional context

Quick code trace from a read of the open-source repo:

- \`TerminalView::drag_and_drop_files\` (\`app/src/terminal/view.rs:23754\`) gates the image-attach branch on \`!is_in_long_running_command\` (line 23773). When a CLI agent like Claude Code is running, the active block is long-running, so the attach branch is skipped and the drop falls through to \`typed_characters_on_terminal\` (line 23830), which writes the path string into the PTY.
- \`Input::process_paste_event\` (\`app/src/terminal/input.rs:9849\`) takes a different path for clipboard image bytes — it calls \`handle_pasted_image_data\`, which does not consult the long-running-command gate. That's why Cmd+V works.
- The downstream \`can_attach_on_filepaths_paste_or_dragdrop\` (\`app/src/terminal/input.rs:9921\`) already has an explicit branch that returns \`true\` when \`CLIAgentSessionsModel::is_input_open(...)\` is true, so the gating helper is ready for this case — only the upstream long-running-command short-circuit blocks it.

This is **not** a request to revert #7032 (path-pasting in normal shell input). The proposed scope is narrowly: when a CLI agent session is active for the terminal, route image drops through the existing image-attach pipeline instead of typing the path into the PTY. Outside CLI-agent sessions, drag-drop behavior remains unchanged.

No

Ignore

SagarSDagdu:
## Description

Fixes #9545.

When a CLI agent (e.g. Claude Code) is the active foreground process in a Warp tab, dragging an image file onto the tab inserts the file path as text into the agent's PTY instead of routing the drop through the same image-paste pipeline that Cmd+V uses. Cmd+V on an image works correctly because `TerminalView::paste` writes the clipboard image bytes to the system clipboard and sends the agent's paste keystroke to the PTY (`0x16` on macOS/Linux, `ESC v` on Windows) — the agent reads its own clipboard and renders `[Image #N]` in its prompt.

This change mirrors that mechanism for drag-and-drop. When a CLI agent session is registered for the terminal, the active block is long-running, and the dropped files are all images:

### Before the fix video:
https://www.loom.com/share/32b03580946249389be828e15e5a7d8d

### After the fix video:
https://www.loom.com/share/b38e4676c5cf4a31ad985bab45fdfdfa

- Each image is read off-thread.
- Bytes are written to the system clipboard as `ImageData`.
- The agent's paste keystroke is sent to the PTY.
- A delay (`CLI_AGENT_IMAGE_PASTE_DELAY`, the existing constant used by `paste_images_then_submit_text`) separates iterations so the agent has time to read the clipboard before it's overwritten.

End user behavior is identical to copying the image to the clipboard and pressing Cmd+V over the agent's TUI.

Mixed drops (some image, some non-image paths) and drops outside CLI-agent sessions fall through to the existing path-typing behavior unchanged.

## History

This PR initially took a different approach - extending Warp's image-attach pipeline so the path wouldn't be typed into the PTY. That stopped the path from being typed but attached the image to Warp's hidden rich input, which the agent never reads. Manual testing confirmed the image didn't reach Claude Code. The second commit replaces the implementation with the clipboard + PTY paste approach above.

## Testing

Added `drag_drop_image_in_cli_agent_long_running_command_pastes_via_clipboard` in `app/src/terminal/view_test.rs`. The test creates a temp file, registers a Claude CLI-agent session, simulates a long-running active block, drops the file, and asserts (via `assert_eventually!`) that the PTY received exactly one paste-keystroke write. Without the fix, a shell-escaped path string is written instead.

`cargo test -p warp drag_drop_image_in_cli_agent` passes locally.

Manually verified on macOS 26.4.1 with WarpOss built locally: dragging a screenshot into a Claude Code session now produces `[Image #N]` in the agent's prompt, matching Cmd+V behavior.

## Agent Mode

CHANGELOG-IMPROVEMENT: Added support for drag-and-drop of image files into an active CLI agent session (e.g. Claude Code).


SagarSDagdu:
@cla-bot check

zachbai:
@advait-m mind taking a look?

advait-m:
> @advait-m mind taking a look?

Yep will take a look here!

advait-m:
Thanks for the PR!

@SagarSDagdu could you add a video showing a before/after state here? I'm not seeing this fix the linked GH issue (I'm still seeing paths for images dropped into Claude Code input directly). Did you get a chance to verify this?

Note that rich input already handles the drag-and-drop case for images correctly.

Loom walking through what I'm seeing on current Warp + running your branch locally: https://www.loom.com/share/61e75f8f66db4847bbc28c5aa82dfd7e

advait-m (changes requested):
Would like to see verification of the fix with a video of before/after since I'm not seeing this PR fix the underlying GH issue, thanks!

SagarSDagdu:
> Would like to see verification of the fix with a video of before/after since I'm not seeing this PR fix the underlying GH issue, thanks!

@advait-m Thanks for taking the time to code review!
Attaching the loom videos regarding the before and after behavior.


### Before the fix video:
https://www.loom.com/share/32b03580946249389be828e15e5a7d8d

### After the fix video:
https://www.loom.com/share/b38e4676c5cf4a31ad985bab45fdfdfa

advait-m:
Thanks for the Looms - that looks great, behavior-wise! I'll take a look at the code and try out your branch myself later today!

app/src/terminal/view/use_agent_footer/mod.rs
```diff
@@ -817,6 +833,92 @@ impl TerminalView {
...
+                            }
+                            ctx.clipboard().write(ClipboardContent {
+                                images: Some(vec![ImageData {
+                                    data: bytes,
+                                    mime_type,
+                                    filename,
+                                }]),
+                                ..Default::default()
+                            });
+                            // Windows Claude Code paste is `Alt+V` (ESC + 'v');
```
advait-m: Hmm, I think this is right though I haven't had a chance to test on Windows. Could we extract this to a helper to re-use across places?

```
fn cli_agent_paste_keystroke_bytes() -> Vec<u8> {
    if cfg!(windows) { vec![0x1b, b'v'] } else { vec![0x16] }
}
```

to reuse in `paste_images_then_submit_text`

app/src/terminal/view/use_agent_footer/mod.rs
```diff
@@ -817,6 +833,92 @@ impl TerminalView {
...
+        ctx: &mut ViewContext<Self>,
+    ) {
+        if image_filepaths.is_empty() {
+            return;
+        }
+        let spawner = ctx.spawner();
+        ctx.spawn(
+            async move {
+                for path_str in image_filepaths {
+                    let bytes = match async_fs::read(&path_str).await {
```
advait-m, line 879: There's a slight risk here of allocating a massive image, since there's no inherent limits (system clipboard has pretty high limits of images). Can we add in a check and inform user via a toast if we have an error? Would appreciate if you could test this path manually too with a massive image!

`MAX_IMAGE_SIZE_BYTES` is too low probably since CLI agents can handle the compression for us here, so maybe we can introduce a new `MAX_IMAGE_SIZE_BYTES_FOR_CLI_AGENT` constant that is like 500 MB for a reasonable very high limit?

little snippet (won't exactly look like this but directionally):
```
Ok(meta) if meta.len() as usize > MAX_IMAGE_SIZE_BYTES_FOR_CLI_AGENT => {
    let msg = format!(
        "{} is too large to send to the agent (limit {}MB).",
        Path::new(&path_str).file_name().unwrap_or_default().to_string_lossy(),
        MAX_IMAGE_SIZE_BYTES_FOR_CLI_AGENT / 1_000_000
    );
    let _ = spawner
        .spawn(move |me, ctx| me.show_error_toast(msg, ctx))
        .await;
    continue;
}
```

Oz should be able to address this pretty well

app/src/terminal/view/use_agent_footer/mod.rs
```diff
@@ -96,6 +97,21 @@ const CLI_AGENT_IMAGE_PASTE_DELAY: Duration = Duration::from_millis(300);
...
+        .map(str::to_lowercase)
+        .as_deref()
+    {
+        Some("jpg") | Some("jpeg") => "image/jpeg",
+        Some("gif") => "image/gif",
+        Some("webp") => "image/webp",
+        _ => "image/png",
+    }
+    .to_owned()
+}
```
advait-m: we don't need this - we have a helper for this already that is more robust in `infer_mime_type` (with fallback)

lets move `infer_mime_type` to a shared spot and re-use that func for use case here

can probably put it into `app/src/util/image.rs` and make it `pub`

app/src/terminal/view/use_agent_footer/mod.rs
```diff
@@ -817,6 +833,92 @@ impl TerminalView {
...
+                    let mime_type = mime_type_for_image_path(path);
+
+                    // Hop back to the view to write the clipboard + paste
+                    // keystroke. Bail if the CLI agent session disappeared,
+                    // OR if the agent's long-running block exited while we
+                    // were reading off-thread — without that second check
+                    // the paste byte would leak into the shell after the
+                    // agent quit, since the session entry can outlive its
+                    // foreground block.
+                    let cont = spawner
```
advait-m: nit: cont -> should_continue to be clearer

app/src/terminal/view.rs
```diff
@@ -23772,6 +23772,21 @@ impl TerminalView {
...
+        // CLI-agent paste path: when a CLI agent (e.g. Claude Code) is the
+        // foreground long-running process, hand image drops to the agent the
+        // same way Cmd+V does at `TerminalView::paste` — write each image to
+        // the system clipboard and send the agent's paste keystroke to the
+        // PTY. Without this branch the path string would be shell-escaped and
+        // typed into the agent's prompt.
+        if !image_filepaths.is_empty()
+            && image_filepaths.len() == paths.len()
+            && is_in_long_running_command
+            && self.has_active_cli_agent_session(ctx)
```
advait-m, line 23787: We should add a gate for rich input NOT being open - rich input handles drag and drop into image chip already.

I think smth like `CLIAgentSessionsModel::as_ref(ctx).is_input_open(self.view_id)`

app/src/terminal/view/use_agent_footer/mod.rs
```diff
@@ -5,6 +5,7 @@
 //! Gemini CLI, Codex), it displays a specialized footer with additional functionality.
 
 use crate::ai::agent::ImageContext;
+use std::path::Path;
```
advait-m: hyper nit: move into std imports group below

```
use std::sync::{Arc, LazyLock};
use std::time::Duration;
use std::path::Path;
```

app/src/terminal/view/use_agent_footer/mod.rs
```diff
@@ -817,6 +833,92 @@ impl TerminalView {
         );
     }
 
+    /// Mirrors the CLI-agent Cmd+V image-paste path in `TerminalView::paste`
+    /// for dropped image files: reads each file, writes its bytes to the
+    /// system clipboard as image data, and sends the agent's paste keystroke
+    /// to the PTY so the agent reads the image directly. This produces the
+    /// same outcome as if the user had copied the image to their clipboard
+    /// and pressed Cmd+V over the agent's TUI.
+    pub(in crate::terminal) fn paste_dropped_images_to_cli_agent(
```
advait-m: nit: I think pub(super) is fine here?

advait-m (commented):
okay, this is looking great overall! All largely smaller comments - biggest functional piece to fix is not breaking the rich input flow (and from code POV, re-using the MIME type helper we have)

ping me after addressing these comments and I'll take a look to approve/help merge this PR!

thanks again for working on this + the looms!

app/src/terminal/view/use_agent_footer/mod.rs
SagarSDagdu: Extracted `cli_agent_paste_keystroke_bytes()` and in both call sites ✅

app/src/terminal/view/use_agent_footer/mod.rs
SagarSDagdu: `MAX_IMAGE_SIZE_BYTES_FOR_CLI_AGENT = 500MB` in `util/image.rs` and stat-first to reject without loading; ephemeral error toast names the file.

app/src/terminal/view/use_agent_footer/mod.rs
SagarSDagdu: Moved `infer_mime_type` from `artifact_upload.rs` to `util/image.rs` (made `pub`); replaced my hand-rolled mapping. Original `artifact_upload.rs` site now imports the shared helper ✅

app/src/terminal/view/use_agent_footer/mod.rs
SagarSDagdu: Done ✅

app/src/terminal/view.rs
SagarSDagdu: Rich-input gate added in `TerminalView::drag_and_drop_files` so the chip-attach flow keeps working when the rich input is open.

app/src/terminal/view/use_agent_footer/mod.rs
SagarSDagdu: Done!

app/src/terminal/view/use_agent_footer/mod.rs
SagarSDagdu: Yep, done!

SagarSDagdu:
@advait-m thank you for your inputs here!
I have fixed/improved all the points you mentioned, you can take a second look and give your feedback

app/src/terminal/view/use_agent_footer/mod.rs
```diff
@@ -862,7 +887,7 @@ impl TerminalView {
                     let filename = path
                         .file_name()
                         .map(|n| n.to_string_lossy().into_owned());
-                    let mime_type = mime_type_for_image_path(path);
+                    let mime_type = infer_mime_type(path, &bytes);
```
advait-m: We should use only a small subset of bytes here rather than entire file - see example below

Should be able to reuse this kind of logic with `MIME_SNIFF_BYTES`

```
        let (file_size, mime_sniff_bytes) = file_size_and_prefix_for_path(&path, MIME_SNIFF_BYTES)?;

        Ok(Self {
            filepath: normalize_artifact_filepath(&path),
            mime_type: infer_mime_type(&path, &mime_sniff_bytes),
            file_size,
            path,
        })
```

advait-m (commented):
thanks for addressing - one comment on MIME type inference

advait-m:
and I've kicked off CI runs against this PR - would appreciate if you could fix any broken formatting, clippy or test errors (see GitHub CI run)

app/src/terminal/view/use_agent_footer/mod.rs
SagarSDagdu: Addressed in `9d9be00`. Now slicing to `MIME_SNIFF_BYTES` before calling `infer_mime_type`, and moved the constant from `artifact_upload.rs` to `util/image.rs` so both call sites share it.

advait-m (approved):
it's shipping time

advait-m:
Congrats on your first Warp contribution @SagarSDagdu - looking forward to many more! 🎉

advait-m:
This should be included in this week's Warp release, expected Thursday!

Issue #8803, opened 2026-03-02
Add support for adding custom language servers


kevinyang372:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Allow users to configure custom language servers that are not supported out-of-the-box in app

To add more color to the product behavior I think we should be building towards:
* We should directly support a neovim-like (https://neovim.io/doc/user/lsp/) config in our settings file. This adds that server into our set of globally installed servers
* The user flow here would be:
    * User opens a file editor. If this is a language we does not support natively, we show a link to documentation on how to add custom lsp configs
    * User can define the path to binary, arguments, and file types for a custom lsp binary in their settings file
    * Now whenever they open a file of the same language type, in the footer we can support showing “enable server X for repo Y”

### Operating system (OS)

macOS

### How important is this feature to you?

3

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

indigoram89:
Very great feature

MarlonDValencia:
+1. I use Warp + AI Agents for Java backend development. Without go-to-definition for `.java` files I have to jump to IDE every time I need to navigate agent-generated code.

The proposed neovim-style config would work well for `jdtls`. Looking forward to this one!

bradleyjames:
Attempt at the spec:

https://github.com/warpdotdev/warp/pull/10700

kevinyang372:
@bradleyjames Nice! Do you mind iterating with oz-review until it's in a good spot? It should assign me as the reviewer after that

bradleyjames:
@kevinyang372 so i took your comment to start on the implementation but it looks like this is tagged as ready-to-spec and not ready-to-implement.   The PR is https://github.com/warpdotdev/warp/pull/10700.

bradleyjames:
@kevinyang372, spec reviewed here: https://github.com/warpdotdev/warp/pull/11944.

nodkrot:
I would love support for php in warp because i mostly use terminal but sometimes need to see file and follow function definitions/references

jvargas:
Add Clojure language server. I would love to use Warp for all REPL interactions and avoid the need for a full blown IDE.

indigoram89:
[...]

PR #9562, opened 2026-04-30
feat: add PHP language support via Intelephense LSP (#9168)

### LINKED ISSUE #9168: Language support (LSP) - PHP
### Pre-submit Checks

## Pre-submit Checks

- [x] I have searched Warp feature requests and there are no duplicates
- [x] I have searched Warp docs and my feature is not there

## Describe the solution you'd like?

Please add **PHP** to the supported languages list in the built-in code editor:
https://docs.warp.dev/code/code-editor/language-server-protocol

**Files to support:** `*.php`, `*.phtml`, and ideally Blade templates (`*.blade.php`) for Laravel users.

**Suggested language server:** [Intelephense](https://intelephense.com/) by Ben Mewburn — the de-facto standard PHP LSP today. It's cross-platform, fast, distributed as a single npm package (`intelephense`), and is what Zed, VS Code, Neovim (coc / nvim-lspconfig), Helix, Emacs lsp-mode, and Sublime LSP all default to. A free tier covers the core intelligence features; a one-time license unlocks premium ones (rename, code actions, etc.).

Phpactor would be a reasonable open-source fallback, but shipping Intelephense by default would match what the rest of the ecosystem does and give the best out-of-the-box experience.

## Why this matters

**1. PHP is one of the largest dev populations still missing from Warp's LSP list.**
It consistently sits in the TIOBE / Stack Overflow / GitHub top 10 languages, powers a huge share of the web (WordPress alone runs ~40% of sites), and has a very active modern ecosystem — Laravel, Symfony, Filament, Livewire, Laravel Nova, Statamic. Right now `.php` files open with *"Language support is not available for this type of file"*, which sends a lot of us back to VS Code / PhpStorm / Zed just to edit.

**2. This directly affects Warp's agentic workflow story.**
I run `claude-code-cli` (and occasionally Codex / Oz) inside Warp every day for a Laravel codebase. The agent generates or modifies PHP code, and then I want to review / tweak it right there in Warp's editor — hover types, go-to-definition into `vendor/`, diagnostics. Without an LSP, I either accept agent output blind or bounce to another editor, which defeats one of the main reasons to use Warp in the first place. The "stay in one place" pitch from #8270 applies exactly the same way to PHP.

**3. Laravel + Warp is already a common combination.**
There are existing Warp issues from Laravel users around related UX gaps — artisan completions (#5620), JetBrains file-link open (#3332), double-click filename selection on `.php` paths (#1871). A working PHP LSP would be the natural next step for this audience.

**4. Integration should be straightforward.**
Intelephense is `npm install -g intelephense` and speaks standard LSP over stdio — the same shape as Warp's existing `pyright` / `rust-analyzer` / `gopls` integrations. No weird bootstrap (unlike, say, JDTLS for Java). Warp's auto-install flow for missing server binaries would work out of the box.

## Nice-to-haves (not blockers)

- Respect `~/intelephense/licence.txt` for Premium users — the same convention Zed, Sublime, and Neovim use.
- Blade highlighting / basic LSP routing for `*.blade.php` — Intelephense itself handles the PHP regions, and Laravel users would get huge value even from partial support.
- Once custom server configs are supported, let users swap Intelephense for Phpactor.

---

Happy to help test a preview build against a real Laravel / Filament / Livewire project.

Thanks! 🙏

_No response_

### Additional context

_No response_

### Operating system (OS)

macOS

4

None

lonexreb:
## Description

Closes #9168.

PHP (`.php`, `.phtml`, and Laravel `.blade.php` templates) is among the top-10 most-used languages by every measure (TIOBE, Stack Overflow, GitHub) and powers a huge share of the web — WordPress alone backs ~40% of sites. Today, opening a `.php` file in Warp's editor shows:

> Language support is not available for this type of file

This PR adds **Intelephense** — the de-facto standard PHP LSP used by Zed, VS Code, Neovim, Helix, Sublime LSP, and Emacs lsp-mode — as a new server, mirroring the existing Node.js-based Pyright integration.

## What changes

| File | Change |
| ---- | ------ |
| `crates/lsp/src/config.rs` | New `LanguageId::Php` variant. `from_path` recognises `.php` and `.phtml`; `lsp_language_identifier()` returns `"php"`; `server_type()` routes to `Intelephense`. |
| `crates/lsp/src/supported_servers.rs` | New `LSPServerType::Intelephense` variant wired into all match arms: `binary_name = "intelephense"`, `args = ["--stdio"]`, `custom_install_args = ["--stdio"]`, `languages = [Php]`, `find_installed_binary_config`, `candidate`. |
| `crates/lsp/src/servers/intelephense.rs` (new, 218 lines) | `IntelephenseCandidate` implementing `LanguageServerCandidate`. Mirrors `pyright.rs` (npm-based, runs `node node_modules/intelephense/lib/intelephense.js --stdio` via the same custom-/system-node detection as Pyright). |
| `crates/lsp/src/servers/mod.rs` | Registers the new module. |
| `crates/lsp/src/config_tests.rs` | 7 new unit tests under `php_language_detection`. |

### Blade template support

`Path::extension` returns only the **final** segment — for `welcome.blade.php` that's `"php"`. So Blade templates are routed to PHP/Intelephense via the existing `.php` arm, and Intelephense handles the embedded PHP regions exactly as it does in VS Code, Zed, etc. Covered by the `classifies_blade_php_files_via_php_extension` test.

### Project detection (`should_suggest_for_repo`)

Suggested when any of these markers are present:

- `composer.json` / `composer.lock` (universal — Composer is the PHP package manager)
- `artisan` (Laravel)
- `wp-config.php` (WordPress)

### Premium licensing

Intelephense Premium (rename, code actions, etc.) is unlocked via `~/intelephense/licence.txt`, which Intelephense itself reads — no client plumbing needed. This matches the convention used by Zed, Sublime LSP, and Neovim.

### Pre-existing pieces this builds on

- The `languages` crate already lists `php` in `SUPPORTED_LANGUAGES` (`crates/languages/src/lib.rs:42`) and already maps `.php`/`.phtml` to its tree-sitter PHP grammar at `crates/languages/src/lib.rs:165`. So syntax highlighting was already in place — only the LSP side was missing.

## Testing

7 new unit tests in `crates/lsp/src/config_tests.rs::php_language_detection`:

| Test | Asserts |
| ---- | ------- |
| `classifies_plain_php_files` | `src/Foo.php` → `LanguageId::Php` |
| `classifies_phtml_files` | `templates/header.phtml` → `LanguageId::Php` |
| `classifies_blade_php_files_via_php_extension` | `resources/views/welcome.blade.php` → `LanguageId::Php` |
| `php_routes_to_intelephense_server` | `LanguageId::Php.server_type() == LSPServerType::Intelephense` |
| `intelephense_advertises_php_language` | `LSPServerType::Intelephense.languages() == [Php]` |
| `intelephense_uses_npm_binary_name` | `LSPServerType::Intelephense.binary_name() == "intelephense"` |
| `intelephense_appears_in_all` | `LSPServerType::all()` yields `Intelephense` (so init-project + settings page surface PHP) |

### Local validation status

- `cargo fmt -- --check` clean.
- Compile-checked the wasm path with `cargo check -p lsp --target wasm32-unknown-unknown --tests` — passes.
- The native `local_fs` path was not compile-checked locally because the `lsp` crate transitively pulls in `warpui`, whose `build.rs` requires the Xcode `metal` shader compiler (full Xcode, not just CommandLineTools). Relying on CI for the full presubmit run.

The new server module is structurally a near-mirror of `pyright.rs` — identical npm-install flow, identical node-binary detection, identical config shape — so behaviour should match Pyright once CI confirms it.

## Server API dependencies

- [ ] Is this change necessary to make the client compatible with a desired server API breaking change? — No, client-only.

## Agent Mode

- [ ] Warp Agent Mode

CHANGELOG-NEW-FEATURE: Added PHP language support via [Intelephense](https://intelephense.com/) — `.php`, `.phtml`, and Laravel `.blade.php` files now get hover, go-to-definition, completions, and diagnostics in Warp's built-in editor.

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"7c330f93-d8b2-467d-a3c9-21a039512fbb","pid":12484,"procStart":"Thu Apr 30 06:20:33 2026","acquiredAt":1777537007428}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"7c330f93-d8b2-467d-a3c9-21a039512fbb","pid":12484,"procStart":"Thu Apr 30 06:20:33 2026","acquiredAt":1777537007428}
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ app/src/persistence/schema.rs.orig
 
 # Don't include personal Claude Code settings
 .claude/settings.local.json
+.claude/scheduled_tasks.lock
 
 # Tab drag development notes
 pr_cleanup.md
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7470,6 +7470,7 @@ dependencies = [
  "simple_logger",
  "strum",
  "strum_macros",
+ "tempfile",
  "tokio",
  "url",
  "warp_core",
--- a/crates/lsp/Cargo.toml
+++ b/crates/lsp/Cargo.toml
@@ -47,3 +47,6 @@ nix = { workspace = true }
 repo_metadata.workspace = true
 sha2 = { workspace = true }
 tokio = { workspace = true, features = ["process"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+tempfile = { workspace = true }
--- a/crates/lsp/src/config.rs
+++ b/crates/lsp/src/config.rs
@@ -32,6 +32,7 @@ pub enum LanguageId {
     JavaScriptReact,
     C,
     Cpp,
+    Php,
 }
 
 impl LanguageId {
@@ -52,6 +53,10 @@ impl LanguageId {
             // compile_commands.json is present, clangd will use the correct language
             // regardless of the languageId we send.
             "h" | "H" | "hh" | "hpp" | "hxx" => Some(Self::Cpp),
+            // PHP. `.blade.php` (Laravel templates) is matched here too because
+            // `path.extension()` returns only the final segment ("php"), and
+            // Intelephense handles the embedded PHP regions of Blade templates.
+            "php" | "phtml" => Some(Self::Php),
             _ => None,
         }
     }
@@ -69,6 +74,7 @@ impl LanguageId {
             LanguageId::JavaScriptReact => "javascriptreact",
             LanguageId::C => "c",
             LanguageId::Cpp => "cpp",
+            LanguageId::Php => "php",
         }
     }
 
@@ -83,6 +89,7 @@ impl LanguageId {
             | LanguageId::JavaScript
             | LanguageId::JavaScriptReact => LSPServerType::TypeScriptLanguageServer,
             LanguageId::C | LanguageId::Cpp => LSPServerType::Clangd,
+            LanguageId::Php => LSPServerType::Intelephense,
         }
     }
 }
--- a/crates/lsp/src/config_tests.rs
+++ b/crates/lsp/src/config_tests.rs
@@ -216,3 +216,67 @@ fn test_path_to_lsp_uri_rejects_relative_path() {
     assert!(result.is_err());
     assert!(result.unwrap_err().to_string().contains("must be absolute"));
 }
+
+mod php_language_detection {
+    //! Regression tests for #9168 — PHP language support via Intelephense.
+
+    use super::*;
+    use crate::config::LanguageId;
+    use crate::supported_servers::LSPServerType;
+
+    #[test]
+    fn classifies_plain_php_files() {
+        assert_eq!(
+            LanguageId::from_path(&PathBuf::from("src/Foo.php")),
+            Some(LanguageId::Php)
+        );
+    }
+
+    #[test]
+    fn classifies_phtml_files() {
+        // Legacy template extension still used by some frameworks.
+        assert_eq!(
+            LanguageId::from_path(&PathBuf::from("templates/header.phtml")),
+            Some(LanguageId::Php)
+        );
+    }
+
+    #[test]
+    fn classifies_blade_php_files_via_php_extension() {
+        // `Path::extension` returns only the final segment, so `foo.blade.php`
+        // is matched by the `php` arm — exactly what we want, since
+        // Intelephense handles the embedded PHP regions of Blade templates.
+        assert_eq!(
+            LanguageId::from_path(&PathBuf::from("resources/views/welcome.blade.php")),
+            Some(LanguageId::Php)
+        );
+    }
+
+    #[test]
+    fn php_routes_to_intelephense_server() {
+        assert_eq!(LanguageId::Php.server_type(), LSPServerType::Intelephense);
+    }
+
+    #[test]
+    fn intelephense_advertises_php_language() {
+        assert_eq!(
+            LSPServerType::Intelephense.languages(),
+            vec![LanguageId::Php]
+        );
+    }
+
+    #[test]
+    fn intelephense_uses_npm_binary_name() {
+        assert_eq!(LSPServerType::Intelephense.binary_name(), "intelephense");
+    }
+
+    #[test]
+    fn intelephense_appears_in_all() {
+        let all: Vec<LSPServerType> = LSPServerType::all().collect();
+        assert!(
+            all.contains(&LSPServerType::Intelephense),
+            "LSPServerType::all() must yield Intelephense so init_project and the \
+             settings page can offer PHP support; got {all:?}",
+        );
+    }
+}
--- a/crates/lsp/src/servers/intelephense.rs
+++ b/crates/lsp/src/servers/intelephense.rs
@@ -95,16 +95,26 @@ impl LanguageServerCandidate for IntelephenseCandidate {
     }
 
     async fn is_installed_on_path(&self, executor: &CommandBuilder) -> bool {
-        // Intelephense's CLI surface is only LSP transport flags (`--stdio`,
-        // `--node-ipc`, `--socket=N`, `--pipe=X`); `--version` is not
-        // documented and `--help` enters connection-transport setup, so any
-        // spawn-based probe is unreliable — exit codes don't distinguish a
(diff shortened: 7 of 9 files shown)


lonexreb:
Thanks for the review! Pushed bf9ac67 addressing both important findings:

- **`find_installed_binary_config`**: dropped the `--version` health probe; the presence of the npm-installed `node_modules/intelephense/lib/intelephense.js` plus a working node binary is now sufficient proof that the install completed. Any genuine failure surfaces through the LSP transport when `command_and_params` spawns the server with `--stdio`.

- **`is_installed_on_path`**: switched the probe from `--version` to `--help`, and accept any successful spawn (regardless of exit status) since we only care that the binary exists and is executable.

Also removed the now-unused `command::r#async::Command` import. `cargo fmt --check` and `cargo check -p lsp --target wasm32-unknown-unknown --tests` both pass.

lonexreb:
Thanks again for the second pass. Pushed a3eb280 + ea32caf addressing the latest finding:

> The Intelephense PATH detection returns true when the process can be spawned, even if it exits unsuccessfully. That can make Warp prefer a broken PATH binary over a working custom install.

Replaced the spawn-based probe with a pure filesystem PATH search (`binary_in_path`). Walks the `path_env_var` directories looking for an executable file named `intelephense` (with `.exe`/`.cmd`/`.bat` fallbacks on Windows since npm ships a `.cmd` shim there). No spawning, so there's no ambiguity about exit-code interpretation.

Net effect:
- Healthy intelephense on PATH → detected and used (no regression vs. last revision)
- Missing intelephense on PATH → falls through to the data_dir copy we install ourselves

Added 3 unit tests for `binary_in_path` (found in first entry, absent across all entries, tolerant of empty PATH segments) and `tempfile` as a dev-dependency.

The follow-up commit a3eb280 also drops a Claude Code session lock file that got accidentally swept into ea32caf via `git add -A`, and adds `.claude/scheduled_tasks.lock` to `.gitignore`.

`cargo fmt --check` and `cargo check -p lsp --target wasm32-unknown-unknown --tests` both pass.

lonexreb:
Pushed 5df396b addressing the third review's finding:

> The new PATH probe can report a non-executable `intelephense` file as installed.

`binary_in_path` previously accepted any regular file by name — a stray text file at `~/bin/intelephense` would pass. Now the helper requires at least one executable mode bit (`0o111`) on Unix; on Windows the `.exe`/`.cmd`/`.bat` extension match is sufficient (that's what `CreateProcessW` uses).

Added a Unix-only regression test `rejects_non_executable_file_on_unix` that creates a 0o644 file in PATH and asserts `binary_in_path` rejects it. The default-perms file is verified as non-executable up front so a regression in the test setup itself can't mask the bug.

`cargo fmt --check` and `cargo check -p lsp --target wasm32-unknown-unknown --tests` both pass.

lonexreb:
@cla-bot check

lonexreb:
Pushed cb95171 addressing the fourth review's two findings:

**1. `is_installed_on_path` was over-permissive**

> can prevent the data-dir fallback from being used when the global npm shim or Node runtime is broken.

Restored the spawn-based probe but kept it bounded:
1. `binary_in_path` filesystem check first (rejects non-executable files / dangling symlinks before any spawn).
2. Spawn `intelephense --stdio` with stdin=`Stdio::null()`. Intelephense's transport reader sees immediate EOF and exits cleanly when healthy.
3. Treat only `output().status.success()` as installed.

This matches your earlier inline suggestion verbatim. A stale npm shim, missing/old Node, or partial install now exits non-zero → `is_installed_on_path` returns false → `command_and_params` falls back to the data_dir copy.

**2. PATH-search tests hard-coded Unix `:`**

Refactored tests to use `std::env::join_paths` via a small `make_path_var` helper. On Windows, the test fixture creates the binary as `<stem>.cmd` (matching the npm shim layout) so the extension match in `binary_in_path` exercises the same code path users hit. The empty-segment test still constructs its PATH manually (since `join_paths` rejects empty components by design) but uses the platform-correct separator.

`cargo fmt --check` and `cargo check -p lsp --target wasm32-unknown-unknown --tests` both pass.

kevinyang372:
Hi @lonexreb -- thank you for contributing!

The current product direction we want to push for here is to support arbitrary, user-configurable LSPs rather than building in support for every language server out of the box.

I would close this PR for now. I would recommend taking a stab at https://github.com/warpdotdev/warp/issues/8803 if you would like to use this language server in Warp!

lonexreb:
Pushed 31285c4 addressing the remaining `[SUGGESTION]` from the [Apr 30 16:01 review](https://github.com/warpdotdev/warp/pull/9562#discussion_r3168954651):

> Bound this probe with a timeout or avoid spawning the LSP here; `--stdio` is a long-running transport mode, so a broken shim that does not exit on EOF can leave install/status detection waiting indefinitely instead of falling back to the data_dir install.

The probe is now bounded by `INSTALL_PROBE_TIMEOUT = Duration::from_secs(3)` via `warpui::r#async::FutureExt::with_timeout`. A healthy intelephense exits in tens of milliseconds on stdin EOF; 3s gives plenty of headroom while preventing a broken npm shim from hanging install detection. Three outcomes:

- `Ok(Ok(output))` — trust `status.success()` (existing behavior).
- `Ok(Err(io_err))` — treat as not installed; data_dir fallback kicks in.
- `Err(timeout)` — log at debug, return false, data_dir fallback runs. Dropped future causes `async-process` to drop its `Child` (kill signal sent).

The earlier IMPORTANT findings from this review cycle were already addressed in cb95171 / 5df396b / ea32caf. With this one in, every concern called out by oz-for-oss is now addressed. Ready for re-review whenever you are.

lonexreb:
Thanks for the steer @kevinyang372 — that's the right call. The user-configurable LSP path in #8803 is the better long-term shape; baking each language server into Warp doesn't scale.

Closing this PR. I'll take a look at #8803 — since it's labeled `ready-to-spec`, the next step is a spec PR (product.md + tech.md under `specs/GH8803/`), so I'll start there rather than another code PR. If you have prior thinking on the config schema (TOML vs JSON, where in settings, neovim-style vs simpler), happy to incorporate.

PR #9565, opened 2026-04-30
fix(bootstrap): add Node.js/yarn prerequisite checks to bootstrap scripts

### LINKED ISSUE #9544: Bootstrap scripts don't check for Node.js/yarn, leading to confusing `cargo build` failures
## Problem

yarn), but:

## Repro

## Proposed fix

  the instructions inline.

Giggitycountless:
## Summary

The `command-signatures-v2` crate compiles a TypeScript helper at `cargo build` time using `yarn`, which requires Node.js 18.14.1+ and yarn (typically via `corepack enable`) on PATH. None of the bootstrap scripts verify either is present, so a fresh checkout fails partway through `cargo build` rather than at bootstrap time.

This PR adds early checks to the bootstrap scripts so users get a clear error message before attempting to build.

Fixes #9544

## Changes

- **`script/linux/install_build_deps`**: Add Node.js version check (≥18.14.1) and yarn availability check
- **`script/macos/bootstrap`**: Add Node.js version check (≥18.14.1) and yarn availability check  
- **`WARP.md`**: Document Node.js and yarn as prerequisites in a new Prerequisites section

## How it works

The checks run during bootstrap (before Rust installation):
1. Verify `node` is installed and ≥18.14.1
2. Verify `yarn` is available (suggests `corepack enable` if missing)
3. Print detected versions on success
4. Exit with a clear error message if either check fails

## Testing

- ✅ Verified on Linux with Node.js 22 + yarn installed
- ✅ Verified error messages when Node.js/yarn are missing

### DIFF
--- a/WARP.md
+++ b/WARP.md
@@ -36,6 +36,10 @@ Environment variables:
 - `./script/run-clang-format.py -r --extensions 'c,h,cpp,m' ./crates/warpui/src/ ./app/src/` - Format C/C++/Obj-C code
 - `find . -name "*.wgsl" -exec wgslfmt --check {} +` - Check WGSL shader formatting
 
+### Prerequisites
+- **Node.js 18.14.1+** — Required by the `command-signatures-v2` crate, which compiles a TypeScript helper at build time using yarn.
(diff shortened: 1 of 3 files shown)


Giggitycountless:
@cla-bot check

PR #9564, opened 2026-04-30
feat: add JSON language support via vscode-json-languageserver (#9556)

lonexreb:
## Description

Closes #9556.

Opening a `.json` file in Warp's editor today shows:

> Language support is unavailable for this file type

This is a high-friction gap because almost every project Warp users open contains JSON config files (`package.json`, `tsconfig.json`, `composer.json`, `.vscode/settings.json`, etc.).

This PR adds the [VS Code JSON language server](https://github.com/microsoft/vscode/tree/main/extensions/json-language-features/server), distributed on npm as [`vscode-json-languageserver`](https://www.npmjs.com/package/vscode-json-languageserver). It's the same JSON LSP that ships inside VS Code, Zed, and Neovim's default configurations — schema-aware validation, hover, completion, and `$ref` go-to-definition.

Mirrors the existing Pyright integration pattern (and the parallel Intelephense PR #9562) for npm-based Node.js servers.

## What changes

| File | Change |
| ---- | ------ |
| `crates/lsp/src/config.rs` | New `LanguageId::Json` variant. `from_path` recognises `.json` and `.jsonc` (JSON-with-comments, used for `tsconfig.json` and VS Code's own config files). `lsp_language_identifier()` returns `"json"`. `server_type()` routes to `VsCodeJsonLanguageServer`. |
| `crates/lsp/src/supported_servers.rs` | New `LSPServerType::VsCodeJsonLanguageServer` variant wired into all match arms: `binary_name = "vscode-json-languageserver"`, `args = ["--stdio"]`, `custom_install_args = ["--stdio"]`, `languages = [Json]`. |
| `crates/lsp/src/servers/vscode_json_language_server.rs` (new, ~225 lines) | `VsCodeJsonLanguageServerCandidate` implementing `LanguageServerCandidate`. Runs `node node_modules/vscode-json-languageserver/dist/node/jsonServerMain.js --stdio` via the same custom-/system-node detection as Pyright/Intelephense. |
| `crates/lsp/src/servers/mod.rs` | Registers the new module. |
| `crates/lsp/src/config_tests.rs` | 7 new unit tests under `json_language_detection`. |

### Project detection (`should_suggest_for_repo`)

Suggested when any of these markers are present:
- `package.json` (Node.js / npm projects)
- `tsconfig.json` (TypeScript projects)
- `composer.json` (PHP / Composer projects)
- `.vscode/settings.json` (any VS Code-aware repo)

[...]

PR #9568, opened 2026-04-30
feat: add JSON language support via vscode-json-languageserver (#9556) — v2

### LINKED ISSUE #9556: Add JSON language support (LSP) for code editor

Add JSON language server support to Warp's code editor, similar to how JavaScript/TypeScript, Python, Rust, Go, and C/C++ are already supported.

When opening a `.json` file in the editor, Warp should:
- Provide hover information for JSON keys/values
- Support go-to-definition for `$ref` references in JSON schemas
- Show inline diagnostics for invalid JSON syntax
- Offer format-on-save

### Is your feature request related to a problem? Please describe?

Yes. When opening `.json` files in Warp's code editor, I get the message:
> "Language support is unavailable for this file type"

This is frustrating because:
- JSON is a standard file format used in almost every project (config files, package.json, tsconfig.json, etc.)
- Other languages like JavaScript already have LSP support in Warp
- The LSP protocol already supports JSON (via `vscode-json-language-server` or similar)

### Additional context

- Warp already supports TypeScript/JavaScript, Python, Rust, Go, and C/C++ with LSP
- Users can enable LSP for these languages via the editor footer when opening a supported file
- Custom language server support is tracked in issue #8803, but that feature is not yet available
- A native JSON LSP implementation would follow the same pattern as the existing built-in language servers

lonexreb:
## Description

Closes #9556. Supersedes #9564 (which the bot auto-closed before my final commit landed; this PR includes that commit plus all earlier review fixes).

Opening a `.json` file in Warp's editor today shows:

> Language support is unavailable for this file type

Adds the [VS Code JSON language server](https://github.com/microsoft/vscode/tree/main/extensions/json-language-features/server), distributed on npm as [`vscode-json-languageserver`](https://www.npmjs.com/package/vscode-json-languageserver). Mirrors the Pyright integration pattern for npm-based Node.js servers.

## What changes

| File | Change |
| ---- | ------ |
| `crates/lsp/src/config.rs` | New `LanguageId::Json` and `LanguageId::Jsonc` variants. `from_path` first checks the **filename** for well-known JSONC files (`tsconfig.json`, `jsconfig.json`, `tsconfig.<variant>.json`, `.vscode/{settings,launch,tasks,keybindings,extensions}.json`, `devcontainer.json`, `typedoc.json`); strict-JSON filenames (`package.json`, `package-lock.json`, `manifest.json`) keep going through the extension fallback. The `.jsonc` extension is also recognised. |
| `crates/lsp/src/supported_servers.rs` | New `LSPServerType::VsCodeJsonLanguageServer` variant wired into all match arms; `languages()` advertises `[Json, Jsonc]`. |
| `crates/lsp/src/servers/vscode_json_language_server.rs` (new, ~340 lines) | `VsCodeJsonLanguageServerCandidate`. Runs `node node_modules/vscode-json-languageserver/out/node/jsonServerMain.js --stdio` (the path matches the package's `package.json` `"main"` entry) via the same custom-/system-node detection used by Pyright/Intelephense. Custom-install detection is purely file-existence based. PATH detection uses `binary_in_path` — a pure filesystem PATH walk that, on Unix, requires the candidate to have at least one executable mode bit (`0o111`) set, and on Windows tries `.exe`/`.cmd`/`.bat` extensions in addition to the bare name. |
| `crates/lsp/src/servers/mod.rs` | Registers the new module. |
| `crates/lsp/src/config_tests.rs` | New `json_language_detection` test module. |
| `crates/lsp/Cargo.toml` | Adds `tempfile` as a non-wasm dev-dependency for the PATH-search tests. |
| `.gitignore` | Adds `.claude/scheduled_tasks.lock`. |

## How this PR addresses each finding from #9564

1. **`tsconfig.json` etc. were strict JSON** — `LanguageId::from_path` now checks the filename first for the canonical JSONC config files, so `tsconfig.json` and `.vscode/settings.json` route to `LanguageId::Jsonc` and emit the `jsonc` languageId in `didOpen`. New tests `known_dotjson_jsonc_filenames_route_to_jsonc` (10 filenames) and `unrelated_dotjson_stays_strict_json` (4 strict-JSON filenames) cover both directions.

2. **Wrong custom-install JS path** — `LANGSERVER_JS_PATH` now points at `node_modules/vscode-json-languageserver/out/node/jsonServerMain.js`, matching the npm package's `package.json` `"main": "./out/node/jsonServerMain"` field.

3. **Spawn-based PATH probe was unreliable** — replaced with `binary_in_path`, a pure filesystem PATH search. The npm package only documents `--stdio`/`--node-ipc`/`--socket` as transport flags; running it with `--version` or `--help` enters connection-transport setup and exits with the missing-transport error, so spawn-based probes can't reliably distinguish a healthy install from a broken one.

4. **Non-executable file falsely detected** — `binary_in_path` now requires at least one executable mode bit on Unix; on Windows the extension match is sufficient. New Unix-only test `rejects_non_executable_file_on_unix` asserts the bug fix.

## Testing

`cargo fmt -- --check` clean. `cargo check -p lsp --target wasm32-unknown-unknown --tests` passes.

The repo's full presubmit (`cargo nextest run --workspace`) was not run locally — the `lsp` crate transitively pulls in `warpui`, whose `build.rs` requires the Xcode `metal` shader compiler (full Xcode, not just CommandLineTools). Relying on CI for full presubmit verification.

13 new unit tests across `crates/lsp/src/config_tests.rs::json_language_detection` and the inline `vscode_json_language_server::tests` module:

| Test | Asserts |
| ---- | ------- |
| `classifies_plain_json_files` | `package.json`, `nested/data.json` → `Json` |
| `classifies_jsonc_files_distinctly` | `docs/example.jsonc` → `Jsonc` |
| `known_dotjson_jsonc_filenames_route_to_jsonc` | tsconfig.json + 9 others → `Jsonc` |
| `unrelated_dotjson_stays_strict_json` | package.json + 3 others → `Json` |
| `both_json_and_jsonc_route_to_vscode_json_language_server` | server routing |
| `vscode_json_language_server_advertises_both_languages` | `languages() == [Json, Jsonc]` |
| `vscode_json_language_server_uses_npm_binary_name` | binary name |
| `vscode_json_language_server_appears_in_all` | enumeration |
| `json_and_jsonc_emit_distinct_lsp_language_identifiers` | LSP identifiers |
| `finds_binary_in_first_path_entry` (`binary_in_path`) | match |
| `rejects_when_binary_absent` (`binary_in_path`) | no false positives |
| `rejects_non_executable_file_on_unix` (`binary_in_path`) | exec-bit gate |

## Server API dependencies

- [ ] No server impact.

## Agent Mode

- [ ] Warp Agent Mode

CHANGELOG-NEW-FEATURE: Added JSON language support in the built-in code editor — `.json`, `.jsonc`, and JSONC-by-convention files like `tsconfig.json` and `.vscode/settings.json` now get schema-aware validation, hover, completion, and `\$ref` go-to-definition via the [VS Code JSON language server](https://github.com/microsoft/vscode/tree/main/extensions/json-language-features/server).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ app/src/persistence/schema.rs.orig
 
 # Don't include personal Claude Code settings
 .claude/settings.local.json
+.claude/scheduled_tasks.lock
 
 # Tab drag development notes
 pr_cleanup.md
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7470,6 +7470,7 @@ dependencies = [
  "simple_logger",
  "strum",
  "strum_macros",
+ "tempfile",
  "tokio",
  "url",
  "warp_core",
--- a/crates/lsp/Cargo.toml
+++ b/crates/lsp/Cargo.toml
@@ -47,3 +47,6 @@ nix = { workspace = true }
 repo_metadata.workspace = true
 sha2 = { workspace = true }
 tokio = { workspace = true, features = ["process"] }
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+tempfile = { workspace = true }
--- a/crates/lsp/src/config.rs
+++ b/crates/lsp/src/config.rs
@@ -362,36 +362,67 @@ pub fn default_init_params(workspace_uri: &Path, client_name: String) -> Result<
     })
 }
 
-/// Returns `true` for paths whose final filename is a well-known JSON-with-comments
-/// config file. The VS Code JSON language server distinguishes `json` from `jsonc`
-/// and only allows comments under `jsonc`, but several Microsoft tools ship with
-/// `.json` extensions and JSONC contents (TypeScript's `tsconfig.json`,
-/// VS Code's own `settings.json`, etc.). Routing these by name keeps users
-/// from getting spurious "comments not allowed" diagnostics.
+/// Returns `true` for paths whose name (and, where ambiguous, parent
+/// directory) identify a well-known JSON-with-comments config file. The
+/// VS Code JSON language server distinguishes `json` from `jsonc` and only
+/// allows comments under `jsonc`, but several Microsoft tools ship with
+/// `.json` extensions and JSONC contents — routing these by name avoids
+/// spurious "comments not allowed" diagnostics.
+///
+/// Detection is split into two tiers to avoid relaxing validation for
+/// unrelated files that happen to share a common name:
+///
+/// 1. **Globally JSONC** — names that are unambiguous wherever they
+///    appear (`tsconfig.json` and its `.<variant>.json` siblings,
+///    `jsconfig.json` ditto, `devcontainer.json`, `typedoc.json`).
+///
+/// 2. **JSONC only when the parent directory is `.vscode/`** — VS Code's
+///    workspace/user files (`settings.json`, `launch.json`, `tasks.json`,
+///    `keybindings.json`, `extensions.json`). Outside `.vscode/` a file
+///    called `settings.json` is just as likely to be strict JSON in some
+///    unrelated project, so we leave those alone.
 fn filename_is_jsonc(path: &Path) -> bool {
     let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
         return false;
     };
-    // Exact filename matches that are always JSONC by convention.
-    const EXACT: &[&str] = &[
+
+    // Tier 1: globally JSONC by convention.
+    const ALWAYS_JSONC: &[&str] = &[
         "tsconfig.json",
         "jsconfig.json",
-        // VS Code workspace/user/folder settings files.
+        "devcontainer.json",
+        "typedoc.json",
+    ];
+    if ALWAYS_JSONC.contains(&name) {
+        return true;
+    }
+    // Variants like `tsconfig.build.json`, `tsconfig.app.json`,
+    // `jsconfig.app.json`, etc.
+    if (name.starts_with("tsconfig.") || name.starts_with("jsconfig.")) && name.ends_with(".json") {
+        return true;
+    }
+
+    // Tier 2: VS Code workspace files — only when the parent directory is
+    // exactly `.vscode/`. A `settings.json` somewhere else might be strict
+    // JSON in an unrelated project; we don't want to silently accept
+    // comments there.
+    const VSCODE_JSONC: &[&str] = &[
         "settings.json",
         "launch.json",
         "tasks.json",
         "keybindings.json",
         "extensions.json",
-        // Other editor configs that follow the same convention.
-        "devcontainer.json",
-        "typedoc.json",
     ];
-    if EXACT.contains(&name) {
-        return true;
+    if VSCODE_JSONC.contains(&name) {
+        let parent_is_vscode = path
+            .parent()
+            .and_then(|p| p.file_name())
+            .and_then(|n| n.to_str())
+            == Some(".vscode");
+        return parent_is_vscode;
     }
-    // Variants like `tsconfig.build.json`, `tsconfig.app.json`, etc.
-    name.starts_with("tsconfig.") && name.ends_with(".json")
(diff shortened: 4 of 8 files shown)


lonexreb:
Thanks for the review! Pushed 321af22 addressing the finding:

> The JSONC filename detection currently treats VS Code-specific filenames such as `settings.json`, `launch.json`, `tasks.json`, `keybindings.json`, and `extensions.json` as JSONC regardless of their directory.

Split `filename_is_jsonc` into two tiers:

**Always JSONC** (unambiguous regardless of location):
- `tsconfig.json`, `jsconfig.json`, `tsconfig.<variant>.json`, `jsconfig.<variant>.json`
- `devcontainer.json`, `typedoc.json`

**JSONC only when the immediate parent is `.vscode/`**:
- `settings.json`, `launch.json`, `tasks.json`, `keybindings.json`, `extensions.json`

So bare `settings.json` or `config/launch.json` now stays strict JSON, preserving full diagnostics. Nested-project layouts like `frontend/.vscode/settings.json` still resolve to JSONC because the rule checks the *immediate* parent, not the path prefix.

Two new tests cover both directions:
- `vscode_filenames_outside_dotvscode_stay_strict_json` — 8 paths including bare `settings.json`, `src/settings.json`, `config/launch.json`, `deeply/nested/extensions.json` → all `LanguageId::Json`.
- `vscode_filenames_under_dotvscode_become_jsonc` — 6 paths including the nested-project case `frontend/.vscode/settings.json` → all `LanguageId::Jsonc`.

`cargo fmt --check` and `cargo check -p lsp --target wasm32-unknown-unknown --tests` both pass.

lonexreb:
@cla-bot check

lonexreb:
Pushed d1dd21b addressing the second review's finding:

> The JSON server is enabled with the shared default initialization parameters, leaving schema URI fetching unconstrained; untrusted JSON can cause the language server to fetch remote schemas by default.

Added `LSPServerType::initialization_options()` and wired it into `command_and_params`. For `VsCodeJsonLanguageServer` it now sends:

```json
{
    "handledSchemaProtocols": ["file"],
    "provideFormatter": true
}
```

That restricts the server to local-file `$schema` URIs only. `http`/`https` schemas are no longer auto-fetched at all (they can later be delegated to the client via `vscode/content` if Warp wants to proxy them through its own HTTP client).

Other servers continue to return `None` from `initialization_options()` and behave exactly as before. Two new tests:
- `json_server_restricts_schema_protocols_to_file` — asserts the protocol allowlist is `["file"]` and nothing else.
- `other_servers_have_no_initialization_options` — sanity check that no other server kind accidentally gained init options.

`cargo fmt --check` and `cargo check -p lsp --target wasm32-unknown-unknown --tests` both pass.

lonexreb:
Pushed 270888e applying the same cross-platform fix the Intelephense PR got: `make_path_var` helper backed by `std::env::join_paths`, plus `binary_filename` that produces `.cmd` test fixtures on Windows so the extension match exercises the same code path Warp users hit. `cargo fmt --check` and `cargo check -p lsp --target wasm32-unknown-unknown --tests` both pass.

kevinyang372:
Hi @lonexreb -- thank you for contributing! 

The current product direction we want to push for here is to support arbitrary, user-configurable LSPs rather than building in support for every language server out of the box.

I would close this PR for now. I would recommend taking a stab at https://github.com/warpdotdev/Warp/issues/8803 if you would like to use this language server in Warp!

lonexreb:
Thanks @kevinyang372 — same story as #9562, closing this one too. Will pivot to #8803.

A small data point in case it helps the spec: the schema-fetching restriction work in this PR (`handledSchemaProtocols: ["file"]`) is a useful pattern for the custom-LSP world — any user-configured LSP that fetches remote resources needs the same treatment, so the eventual config schema may want a `network_access: bool` or per-protocol toggle baked into the user-facing config.

PR #9569, opened 2026-04-30
fix: run tab config commands from configured directory

### LINKED ISSUE #9130: Tab Configs: directory field not applied before commands run
## Bug

The `directory` field in a Tab Config is not applied before `commands` execute. Commands run in the default home directory regardless of what `directory` is set to.

## Steps to reproduce

1. Create a tab config:

```toml
name = "Test"

[[panes]]
id = "main"
type = "terminal"
shell = "fish"
directory = "~/projects/website"
commands = ["pwd"]
```

2. Open it from the `+` menu
3. `pwd` outputs `/Users/<user>` (home), not `~/projects/website`

## Expected

`commands` should run in the directory specified by the `directory` field.

## Workaround

Add `cd ~/projects/website` as the first entry in `commands`.

## Environment

- Warp stable channel, macOS

Jason-Bai:
Closes #9130.

## Summary
- Prefix tab config startup commands with `cd <directory>` when the pane declares a `directory`.
- Shell-quote the rendered directory path before composing the startup command.
- Add a regression test for command composition so tab config commands run from the configured directory.

## Root Cause
Tab configs passed the pane `directory` as the terminal startup directory, then separately injected `commands` through the pending-command path. If the shell was not in that startup directory by the time the pending command executed, the command ran from the default directory. Composing the pending command with an explicit `cd` makes the execution order deterministic.

## Testing
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p warp pane_group::tests::test_tab_config_pending_command_changes_to_template_cwd_first`
- `cargo test -p warp tab_configs::`

## Server API dependencies
No server API dependencies.

## Agent Mode
- [ ] Warp Agent Mode

### DIFF
--- a/app/src/pane_group/mod.rs
+++ b/app/src/pane_group/mod.rs
@@ -59,7 +59,7 @@ use std::any::Any;
 use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
 use std::ffi::OsString;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::sync::{mpsc::SyncSender, Arc};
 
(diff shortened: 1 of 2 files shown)


Jason-Bai:
@cla-bot check

Jason-Bai:
Updated this to use the same resolved startup directory for both session creation and the injected `cd` command. I also added coverage for relative cwd and missing cwd cases. Please let me know if you would like any further changes.

moirahuang:
Hi @Jason-Bai appreciate you making this PR! From the original issue, this is an issue specifically when the directory in the tab config doesn't exist. There's [discussion on the main issue](https://github.com/warpdotdev/warp/issues/9130#issuecomment-4362860789) on how we want to handle that case. I'm going to close this PR out in favor of implementing the solutions discussed on the original issue.

PR #9570, opened 2026-04-30
fix(bootstrap): add Node.js/yarn prerequisite checks to bootstrap scripts

### LINKED ISSUE #9544: Bootstrap scripts don't check for Node.js/yarn, leading to confusing `cargo build` failures

Giggitycountless:
## Summary

The command-signatures-v2 crate compiles a TypeScript helper at cargo build time using yarn, which requires Node.js 18.14.1+ and yarn (typically via corepack enable) on PATH. None of the bootstrap scripts verify either is present, so a fresh checkout fails partway through cargo build rather than at bootstrap time.

Fixes #9544

## Changes

- **script/linux/install_build_deps**: Add Node.js version check (>=18.14.1) and yarn 4+ version check
- **script/macos/bootstrap**: Add Node.js version check (>=18.14.1) and yarn 4+ version check
- **WARP.md**: Document Node.js and yarn as prerequisites

## How it works

The checks run during bootstrap (before Rust installation):
1. Verify node is installed and >=18.14.1 (proper semantic version comparison)
2. Verify yarn is available and version 4+ (rejects Yarn 1 from Homebrew)
3. Print detected versions on success
4. Exit with clear error message if either check fails

## Addressing previous review feedback

- Fixed version comparison to use proper numeric comparison (not lexicographic)
- Added yarn version check to reject Yarn 1.x (Homebrew) and require Yarn 4+ (Corepack)

### DIFF
--- a/WARP.md
+++ b/WARP.md
@@ -36,6 +36,10 @@ Environment variables:
 - `./script/run-clang-format.py -r --extensions 'c,h,cpp,m' ./crates/warpui/src/ ./app/src/` - Format C/C++/Obj-C code
 - `find . -name "*.wgsl" -exec wgslfmt --check {} +` - Check WGSL shader formatting
 
+### Prerequisites
+- **Node.js 18.14.1+** — Required by the `command-signatures-v2` crate, which compiles a TypeScript helper at build time using yarn.
+- **yarn** — Required by the `command-signatures-v2` build script. Enable via `corepack enable` after installing Node.js.
+
 ### Platform Setup
 - `./script/bootstrap` - Platform-specific setup (calls platform-specific bootstrap scripts)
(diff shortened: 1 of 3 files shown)


Giggitycountless:
@cla-bot check

Giggitycountless (commented):
fix(build): check Yarn version from crates/command-signatures-v2/js

Corepack reads packageManager from the nearest package.json. Running
yarn --version from repo root can report a different version than what
build.rs uses, since it runs from crates/command-signatures-v2/js.

Change the check to cd into the js subdir before version check so it
matches the actual build environment

PR #9428, opened 2026-04-29
feat: add AWS profile context chip

rafet:
Closes #4461

## Summary

Adds an **AWS Profile** prompt context chip, mirroring the existing Kubernetes / git-branch chip pattern. Implements the request in #4461 ("Add current aws-cli profile to warp prompt").

When the user types an AWS-related command (`aws`, `aws-vault`, `awslocal`, `sam`, `eksctl`, `terraform`, `tofu`, `cdk`, `serverless`/`sls`, `chalice`, `copilot`, `amplify`), the chip appears showing the active profile. Clicking it opens a searchable picker of profiles parsed from `~/.aws/config` and `~/.aws/credentials`. Selecting one emits a shell-appropriate command to switch profile.

## Behavior

- **Active profile resolution** (matches what AWS SDKs honor): `AWS_VAULT` → `AWS_PROFILE` → `AWS_DEFAULT_PROFILE` → `default`.
- **Tracked through precmd hooks**, same path used by `virtualenv` / `conda` / `node_version` / `kube_config`. Updates automatically after `export AWS_PROFILE=...` runs in the user's interactive shell.
- **Switch command is shell-aware**:
  - bash / zsh / fish: `unset AWS_VAULT AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN; export AWS_PROFILE='...'`
  - PowerShell: `Remove-Item Env:AWS_VAULT,Env:AWS_ACCESS_KEY_ID,Env:AWS_SECRET_ACCESS_KEY,Env:AWS_SESSION_TOKEN -ErrorAction SilentlyContinue; $env:AWS_PROFILE = '...'`
- The switch clears `AWS_VAULT` and the temporary credential triplet so that inside an `aws-vault exec` session the chip's display matches what the AWS SDK actually uses.
- **Single-quote escaping** for both POSIX (`'` → `'\''`) and PowerShell (`'` → `''`), so profile names with metacharacters can't break out of the quoted context.

## Files changed

- `app/src/context_chips/` — new `AwsProfile` chip kind, menu, click handler, render function.
- `app/src/terminal/model/` — adds `aws_profile` field to `PrecmdValue`, `Block`, and `Environment`.
- `app/assets/bundled/bootstrap/{bash,zsh,fish}_body.sh` — emit `aws_profile` in precmd JSON.
- `app/src/themes/theme.rs` — adds `input_prompt_aws` color (defaults to terminal yellow).

## Test plan

[...]

PR #9578, opened 2026-04-30
feat: add AWS profile context chip

### LINKED ISSUE #4461: Add current aws-cli profile to warp prompt

_No response_

We currently have a kube-context chip in the warp prompt that appears when typing a kubectl command, I'd love to see the same thing for aws-cli as I switch between several throughout the day, it can get confusing to track which one I'm using.

_No response_

### Additional context

_No response_

3

None

rafet:
Closes #4461 (now labeled `ready-to-implement` — re-opening per the maintainer comment on the issue).

> Replaces #9428, which was auto-closed by Oz before the issue was labeled. Code is unchanged from that PR's final state (commits `4c2e57e`, `e1b7eec`, `9e9342e`).

## Summary

Adds an **AWS Profile** prompt context chip, mirroring the existing Kubernetes / git-branch chip pattern.

When the user types an AWS-related command (`aws`, `aws-vault`, `awslocal`, `sam`, `eksctl`, `terraform`, `tofu`, `cdk`, `serverless`/`sls`, `chalice`, `copilot`, `amplify`), the chip appears showing the active profile. Clicking it opens a searchable picker of profiles parsed from `~/.aws/config` and `~/.aws/credentials`. Selecting one emits a shell-appropriate command to switch profile.

## Behavior

- **Active profile resolution** (matches what AWS SDKs honor): `AWS_VAULT` → `AWS_PROFILE` → `AWS_DEFAULT_PROFILE` → `default`.
- **Tracked through precmd hooks**, same path used by `virtualenv` / `conda` / `node_version` / `kube_config`. Updates automatically after `export AWS_PROFILE=...` runs in the user's interactive shell.
- **Switch command is shell-aware**:
  - bash / zsh: `unset AWS_VAULT AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN; export AWS_PROFILE='...'`
  - fish: `set -e VAR…; set -gx AWS_PROFILE '...'` (fish has no standard `unset`/`export`)
  - PowerShell: `Remove-Item Env:VAR,… -ErrorAction SilentlyContinue; $env:AWS_PROFILE = '...'`
- The switch clears `AWS_VAULT` and the temporary credential triplet (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN`) so that inside an `aws-vault exec` session the chip's display matches what the AWS SDK actually uses.
- **Single-quote escaping** for both POSIX and PowerShell so profile names with metacharacters can't break out of the quoted context.
- The PowerShell profile-list parses `~/.aws/config` and `~/.aws/credentials` separately, matching POSIX/fish behavior: in `config` only `[profile X]` and `[default]` are profiles; in `credentials` every `[X]` is.
- `sed` invocations on env-controlled paths use the `--` option delimiter to prevent paths beginning with `-` from being parsed as a sed program.

## Files changed

- `app/src/context_chips/` — new `AwsProfile` chip kind, menu, click handler, render function.
- `app/src/terminal/model/` — adds `aws_profile` field to `PrecmdValue`, `Block`, and `Environment`.
- `app/assets/bundled/bootstrap/{bash,zsh,fish}_body.sh` — emit `aws_profile` in precmd JSON.
- `app/src/themes/theme.rs` — adds `input_prompt_aws` color (defaults to terminal yellow).

## Test plan

- [x] `cargo check -p warp`
- [x] `cargo fmt --check -p warp`
- [x] `cargo clippy -p warp --no-deps`
- [x] `cargo test -p warp --lib context_chips::` (66/66 pass)
- [x] 8 unit tests for `format_aws_profile_command` covering POSIX / fish / PowerShell / unknown-shell paths, single-quote escaping, metacharacter injection, vault-credentials clearing, and a negative assertion that fish never emits `unset`/`export`
- [ ] Manual: run `./script/run`, type `aws s3 ls`, verify chip appears with current profile
- [ ] Manual: click chip, select different profile, verify chip updates after `export` runs
- [ ] Manual: test in zsh, bash, and fish sessions


app/src/context_chips/context_chip.rs
```diff
@@ -512,6 +537,7 @@ impl Environment {
             python_virtualenv,
             conda_environment,
             node_version,
+            aws_profile: None,
```
vkodithala, line 543: nit, non-blocking: `Environment::new` is a test-only helper, maybe add a comment here to explain why we aren't inheriting an AWS profile here, for posterity?

app/src/context_chips/display_chip.rs
```diff
@@ -879,6 +960,12 @@ impl DisplayChip {
                     return true;
                 }
             }
+            DisplayChipKind::AwsProfile { menu_open, menu } => {
+                if *menu_open {
+                    ctx.focus(menu);
+                    return true;
+                }
+            }
```
vkodithala: nit, non-blocking: Can we reduce duplication here by collapsing these match arms into a single clause, i.e. `DisplayChipKind::WorkingDirectory | DisplayChipKind::GitBranch....`

vkodithala (approved):
This looks really good! Thanks for the fix!

app/src/context_chips/context_chip.rs
vkodithala: @oz-agent Can you respond here?

app/src/context_chips/display_chip.rs
vkodithala: @oz-agent Can you respond here?

vkodithala (changes requested):
Hey @rafet - gave this a second pass and there are a few things that I'd like to see reviewed:
1/ Before I set an `AWS_PROFILE` in my env, the chip shows "default" (this makes sense), but on click I'm not seeing my available profiles. These load in (and the expected behavior begins) when I set `AWS_PROFILE`. This seems incorrect.
2/ This is non-blocking, but instead of re-running the `export` command when a user clicks on their current profile (i.e. I'm on prod, and then click prod again from the picklist), we should consider this a no-op/render some sort of toast. Alternatively, we can omit the current profile from the picklist.
3/ When we've only detected a single profile, the search bar should be disabled, since there are no entries to search through.

Here's a quick vid of me testing out this path, mostly addressing 1/ and 2/: https://www.loom.com/share/23ef2a7dde0b45e18cd9b2061ed39594.

Lmk if I'm misunderstanding anything and thanks again for the fix!

rafet:
@vkodithala thanks for the careful pass! All three points addressed in b8828bd:

1. **Empty picker before `AWS_PROFILE` is set** — root-caused this to the chip's fingerprint being stable across early-bootstrap precmd cycles. If the on-click shell command failed during the first fetch (session not fully ready yet), the empty cached result stuck around until env-driven fingerprint change. Added `ExternalCommandsState` (flips once at bootstrap completion) to force a re-fetch after bootstrap, plus `InvalidatingCommandCount` with `with_invalidate_on_commands(["aws", "aws-vault", "aws-profile", "aws2"])` so new profiles added via `aws configure` show up too.

2. **Re-running export on current profile** — went with the "omit current from picklist" alternative you mentioned. Picker now shows only profiles the user can switch *to*; selecting feels intentional.

3. **Search bar with single profile** — hidden when the AWS picker has ≤1 items. (After omitting the current profile, "1 item" is exactly when search becomes useless.)

Also addressed both inline nits — comment on `Environment::new` explaining the test-only narrow signature, and collapsed the identical `GitBranch | AwsProfile` arms in `try_focus_open_menu`.

PTAL.

rafet:
Friendly bump @vkodithala — rebased on master to resolve the conflict (test file rename + new git-branch refactor). All three review points and both nits are addressed in 10db707; tests still pass. Ready for another look when you have a moment 🙏

vkodithala:
Hey @rafet, thanks for the bump here! Apologies for the late re-review, was out-of-office last week and missed this. Taking a look!

app/src/context_chips/context_chip.rs
```diff
@@ -502,6 +527,13 @@ impl ContextChip {
 
 impl Environment {
     /// Create a new environment with the given values.
+    ///
+    /// This is a test-only constructor that takes only the historically-tracked
+    /// fields. New env-derived fields (like `aws_profile`, `git_branch`) are
+    /// not parameters here on purpose — production code should use
+    /// [`Self::from_block`] which inherits every field from a [`Block`].
+    /// If you're adding a new field, default it to `None` here and populate
+    /// it in `from_block` instead of growing this signature.
```
vkodithala, line 532: Nice!

app/src/context_chips/display_chip.rs
```diff
@@ -1817,6 +2018,49 @@ pub fn format_git_branch_command(encoded_git_branch_on_click_value: &str) -> Str
     format!("git checkout {}", shell_single_quote(&branch.branch_name))
 }
 
+pub fn format_aws_profile_command(
+    profile_name: &str,
+    shell: Option<crate::terminal::shell::ShellType>,
+) -> String {
+    use crate::terminal::shell::ShellType;
```
vkodithala: nit: We typically declare imports at the top of a file, can you move this?

vkodithala (changes requested):
Thanks so much for the changes @rafet. Mostly looks good minus a few nits. Feel free to request a re-review after fixing imports and will approve + merge this in!

rafet:
Thanks @vkodithala! Moved the `ShellType` import to the top of the file in 2115b49. Ready for re-review 🙏

vkodithala (changes requested):
Hey Rafet, thanks for the changes and apologies for the delay here. Will approve once conflicts are resolved!

rafet:
Rebased on master to resolve the conflicts that built up while waiting (test file renames + new `CreateGitBranch` struct). The 5 commits are unchanged in content, just replayed on top of latest master. Tests still pass (8/8 for the AWS-specific suite).

@vkodithala — could you take another look + approve when you have a moment? 🙏

app/src/context_chips/display_menu.rs
```diff
@@ -285,47 +290,57 @@ impl DisplayChipMenu {
...
-                            let mut options = TextOptions::ui_font_size(appearance);
-                            options.font_family_override = Some(ui_font_family);
-                            options
-                        }
-                    };
+            _ if search_disabled_due_to_few_items => None,
+            ChipMenuType::Directories
+            | ChipMenuType::Branches
+            | ChipMenuType::AwsProfiles
+            | ChipMenuType::Environments => Some(ctx.add_typed_action_view(|ctx| {
```
vkodithala, line 305: Very nice improvement!

app/src/context_chips/mod.rs
```diff
@@ -497,6 +533,31 @@ impl ContextChipKind {
...
+                const AWS_COMMANDS: [&str; 13] = [
+                    "aws",
+                    "aws-vault",
+                    "awslocal",
+                    "sam",
+                    "eksctl",
+                    "terraform",
+                    "tofu",
+                    "cdk",
+                    "serverless",
```
vkodithala: I'm a little concerned about including these, since commands like `terraform` are technically cloud-agnostic. I fear it'll be weird for GCP/Azure users, for instance. Wdyt about removing this? Feel free to push back, these are just my initial thoughts.

app/src/context_chips/display_chip.rs
```diff
@@ -1885,6 +2087,45 @@ pub fn format_create_git_branch_command(branch_name: &str) -> String {
...
+                 set -gx AWS_PROFILE '{}'",
+                profile_name.replace('\'', "'\\''")
+            )
+        }
+        // Bash, Zsh — POSIX-style single-quote escaping: `'` → `'\''`.
+        _ => format!(
+            "unset AWS_VAULT AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN; \
+             export AWS_PROFILE='{}'",
+            profile_name.replace('\'', "'\\''")
+        ),
```
vkodithala: nit: Can we avoid wildcard matching here? Pretty easy for us to be declarative.

vkodithala (commented):
I left a few very minor code nits, I think we're almost ready to ship. Letting CI run and in the meantime, going to flag this internally to double-check that the designs are correct and this is something we'd want to support by default. Thanks for the contribution Rafet!

vkodithala (changes requested):
Actually, I'm dogfooding this with `fish` and noticing a fairly large regression. See below, where I set my AWS profile, it errors, and in a new pane that AWS profile doesn't reflect what I set. I'd like you to fix this, and then send over screen recordings of this working on fish, bash, zsh, and Powershell (if you don't have a Windows machine available, lmk and we can figure something out). Feel free to use Warp [Computer Use](https://docs.warp.dev/agent-platform/capabilities/computer-use/) as needed. Thanks.

https://github.com/user-attachments/assets/7f435dd2-1222-4bf4-8129-c3a579f3a06d

rafet:
Thanks @vkodithala, addressed all 3 in 6c146567:

**1. Fish regression (the error)**: root-caused this to fish's `set -e VAR1 VAR2 …` returning exit 4 when any of the listed variables aren't set. For users who aren't inside an aws-vault session, the AWS credential triplet isn't set on the first click, so `set -e` exits non-zero and the chained `set -gx` (which actually does the work) succeeded but the overall command status was 4. Warp rendered that as a failed command. Fix: chain `; or true;` after the erase so the overall exit reflects the subsequent `set -gx`. Verified locally:
```
$ fish -c "set -e AWS_VAULT AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN; or true; set -gx AWS_PROFILE 'prod'; echo profile=\$AWS_PROFILE"
profile=prod
exit: 0
```

**2. `terraform` / `tofu` / `cdk` / `serverless` etc.**: Agree — trimmed `AWS_COMMANDS` to AWS-only CLIs (`aws`, `aws-vault`, `awslocal`, `sam`, `eksctl`, `copilot`). GCP/Azure users running `terraform plan` will no longer see the chip.

**3. Wildcard in `format_aws_profile_command`**: replaced `_ =>` with explicit `Some(ShellType::Bash | ShellType::Zsh) | None =>` so the POSIX-branch shells are declared, and future variants won't silently fall in.

**On the "new pane doesn't reflect the set"**: This is the same architectural property as the existing `node_version` / `virtualenv` / `kube_config` chips — `set -gx` / `export` mutates the current shell's env, not Warp's parent process or new sessions. A new pane spawns a fresh shell that re-reads `$AWS_PROFILE` from system/user-shell-rc, not from the previous session. Making selections persist across panes would require Warp to track and inject the env at session-spawn time, which feels like a separate, larger change than this PR. Happy to file a follow-up issue for it if useful. 

Screen recordings for fish/bash/zsh coming separately — wanted to push the fix first so you can see the regression is gone. Will reply with them shortly.

rafet:
@vkodithala

vkodithala:
Hey @rafet, thanks for the ping and sorry for the delay here. Reviewing your PR now.

app/src/context_chips/mod.rs
```diff
@@ -522,6 +558,29 @@ impl ContextChipKind {
                 })
             }
 
+            Self::AwsProfile => {
+                // Only AWS-specific CLIs trigger the chip. Cloud-agnostic
+                // tools like `terraform` / `tofu` / `cdk` / `serverless` /
+                // `pulumi` are intentionally excluded — including them would
+                // surface the AWS profile when the user is actually working
+                // against GCP or Azure.
+                const AWS_COMMANDS: [&str; 6] = [
```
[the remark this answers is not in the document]
vkodithala: Think this is wrong - we updated logic to remove Terraform and other variants. If anything we should just update the PR description here, but I don't care too much.

app/src/terminal/input.rs
```diff
@@ -947,6 +947,38 @@ fn render_prompt_chip_shell_command(
...
+            //               swallows that so the chip click doesn't render
+            //               as a failed command)
+            //   - PowerShell: `Remove-Item Env:… ; $env:AWS_PROFILE = …`
+            let quoted = shell_quote_arg(profile_name, shell_type);
+            match shell_type {
+                ShellType::Fish => format!(
+                    "set -e AWS_VAULT AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN; \
+                     or true; set -gx AWS_PROFILE {quoted}"
+                ),
+                ShellType::PowerShell => format!(
```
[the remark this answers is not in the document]
vkodithala: This seems valid, let's address this?

vkodithala (changes requested):
Thanks for the changes and apologies again for the delay. A few things:

- Re-ran /oz-review and there might be an issue with how this works in PowerShell. Can you resolve this?
- For some reason, typing `aws` into WarpLocal with a valid AWS config isn't rendering the chip. Can you re-verify these changes locally, on zsh, fish, and Powershell, and send a demo so I can verify this all works as expected?

Feel free to request a re-review after you've addressed these, thanks!

Issue #9486, opened 2026-04-29
Skills in subdirectory .agents/skills/ not discovered when git root is a parent directory


Amirault:
## Summary
Project skills stored in a `.agents/skills/` directory that sits in a **subdirectory of the git repo** are not discovered by Warp. Skills placed at the repo root work correctly. This contradicts the documented behavior: "For Git repositories, Warp includes all skills from your current directory up through the repository root."

## Problem
When the working directory is a subfolder of the git repo and `.agents/skills/` is located there (not at the repo root), Warp does not surface those skills in the slash command autocomplete or via natural language invocation. Moving `.agents/skills/` to the repo root makes them discoverable again.

## Reproduction steps or desired workflow
1. Create a monorepo with structure:
   ```
   repo-root/           ← git root
   └── sub-project/
       └── .agents/
           └── skills/
               └── my-skill/
                   └── SKILL.md   ← valid name + description frontmatter
   ```
2. Open Warp with working directory set to `repo-root/sub-project/`.
3. Open an agent conversation.
4. Type `/my-skill` in the slash command menu — skill does not appear.
5. Move `.agents/skills/my-skill/` to `repo-root/.agents/skills/my-skill/`.
6. Repeat step 4 — skill now appears correctly.

## Expected behavior
Skills in `.agents/skills/` at the current working directory (or any ancestor up to the repo root) are discovered and available via slash commands, as documented.

## Actual behavior
Only skills at the **repository root** `.agents/skills/` are discovered. Skills in a `.agents/skills/` folder inside a subdirectory of the repo are silently ignored, even when the agent session is started from that subdirectory.

## Artifacts
None attached.

## Warp version
0.2026.04.27.15.32.02

## Operating system
macOS

## Possible source references
- Git repo root: `/Users/.../Wakam/` — skills placed here ARE discovered
- Skills location: `/Users/.../Wakam/Pricing/.agents/skills/` — 5 skills (`spec-gate`, `implement`, `implementation-gate`, `test-implementation`, `execute-api-call`), all with valid YAML frontmatter, NOT discovered when working in `Pricing/`

Amirault:
Still having the issue. Is it possible to find out a solution, it’s real blocker...

PR #9583, opened 2026-04-30
fix: skill discovery in subdirectories

rajgandhi1:
Skills stored in .agents/skills/ (or other provider paths like .claude/skills/) inside a subdirectory of the repo were silently not discovered when that subdirectory — or its .agents/ folder — was gitignored. 

Root cause: The indexed file tree (RepoMetadataModel) is built with IgnoredPathStrategy::IncludeLazy: gitignored directories appear in the tree but with loaded: false and empty children. The find_skill_directories_in_tree traversal never reaches skill dirs nested inside them, so only skills at the repo root (when .agents/ is not gitignored there) were discovered.

Fix: Add find_skill_directories_on_fs which walks the filesystem directly using walkdir, bypassing the indexed tree entirely. It skips .git and short-circuits once a skill directory is found. This is called as a supplemental async pass in scan_repository_for_skills alongside the existing tree-based scan. Skills found by both passes are deduplicated by SkillManager::handle_skills_added, which already uses HashSet for skill paths, making double-registration idempotent.                                                    

Testing: Added regression tests for find_skill_directories_on_fs covering root skills, subdirectory skills (the failing scenario from #9486), and .git exclusion.

- Changelog:
      - Bug fix: "Skills in .agents/skills/ inside subdirectories of a git repo are now discovered correctly, even when the subdirectory or its .agents/ folder is gitignored."


peicodes:
We can't walk the file tree on skill reads because that's potentially a very expensive operation. The likely solution is to modify indexing.

The linked issue has been marked ready to spec. Please include your findings in that issue, and work with us to create a product and tech spec. The discussion around the spec step will help us avoid introducing performance issues or regressions. 

Once the spec is approved, we can implement.

PR #9600, opened 2026-04-30
Fix diff button when Show code review button toggle is off

### LINKED ISSUE #1: Search history by more than just command
**Is your feature request related to a problem? Please describe.**
Most shells currently only support searching through your history by the command. This is frustrating because I can't search by the output, the directory, timestamp, the box, or any other metadata. Additionally, the default history file is too small, I often try to search for commands that I ran a few months ago only to find they were flushed from my history.

**Describe the solution you'd like**
I would like to be able to be able to have a UI to infinitely search my terminal history. To start, I'd like to be able to search by:
- Command
- Output
- Timestamp
- Exit code
- Directory
- Box (e.g a specific remote box or a set of remote boxes)

**Additional context**
Once I find a command that matches my search result, I'd like to be able to find commands that I ran before and after in that specific session where the command was executed.

### LINKED ISSUE #9196: terminal prompt: diff button no longer working when code review tool is disabled
### Pre-submit Checks

## Summary

The recent update has caused a regression where the `diff` button on the terminal prompt no longer opens up the code review pane, unless the "show code review button" option is enabled

## Expected behaviour (what used to happen before the update)

- "Show code review button" disabled in settings 
- Clicking the `diff` button (on the terminal prompt) opens up the code review pane

## Version

- v0.2026.04.22.08.46.stable_03 (Note: I had two consecutive update notifications before reporting the issue, so the working version may be two versions before latest one)

### To reproduce

- disable code review button in settings 
- click on the diff button ("View changes") or hit shift, cmd, + to open the diff view 

#### expected behaviour
- the diff view opens up
#### actual behaviour
- nothing happens, unless the "show code review button" option is enabled in settings

### Expected behavior

_No response_

_No response_

### Operating system (OS)

Select an OS

macos

### Shell Version

_No response_

### Current Warp version

_No response_

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

anshul-garg27:
Closes #9196.

### Description

Two `show_code_review_button` gates were dropping panel-open requests on the floor when the user had hidden the toolbar button:

**1. Data-path gate at `Workspace::setup_code_review_panel` (`view.rs:7982`)**

```rust
if !*TabSettings::as_ref(ctx).show_code_review_button {
    return;
}
```

`update_right_panel_open_state` calls into this whenever the right panel is being opened (chip click, `Shift+Cmd+=` keybinding, etc.), so the early return silently swallowed every explicit user action.

**2. Render-path gate at `Workspace::render_config_panel` and `render_config_panel_maximized` (`view.rs:18981` / `19040`)**

```rust
if !item.is_available(app) || !item.is_panel() { return None; }
…
if !HeaderToolbarItemKind::CodeReview.is_available(app) { return None; }
```

`HeaderToolbarItemKind::is_available` for `CodeReview` returns `*TabSettings::as_ref(app).show_code_review_button.value()` (`header_toolbar_item.rs:89`). So even after fix #1 set `pane_group.right_panel_open = true` and `setup_code_review_panel` ran, the next render frame saw `is_available() == false` and returned `None` — the `right_panel_view` was never added to the layout.

This second gate is what @moirahuang flagged when their local repro still showed nothing happening after the first fix landed. The data was correct; the panel was just never composed into the UI.

### Fix

1. **Drop the early return at `setup_code_review_panel`.** The setting is meant to gate only the toolbar button's visibility (already enforced correctly by `header_toolbar_item.rs::is_available`, which feeds `render_header_toolbar_button` at `view.rs:17276`).
2. **Switch panel-render call sites from `is_available` → `is_supported`.** `is_available`'s own doc-comment says it's specifically *"Whether this item should be shown in the **toolbar** — checks both `is_supported` and user show/hide preferences."* Using it to gate panel rendering conflates two unrelated concerns. Panel rendering should only care about whether the feature is compiled in (`is_supported`), not whether the user has hidden the toolbar button.

For `CodeReview`, `is_supported` is `cfg!(feature = "local_fs")`. For the other variants in the same match (`TabsPanel`, `ToolsPanel`), `is_available` already equals `is_supported` (default `_ => true` arm in the inner match), so behaviour is unchanged. `AgentManagement` and `NotificationsMailbox` return `None` unconditionally inside `render_config_panel`, so the change is moot for them too.

### Caller audit for `setup_code_review_panel`

5 call sites in `view.rs`:

1. `view.rs:3681` — `TransferredTab` flow, only runs when the source tab already had `right_panel_open == true`.
2. `view.rs:8136` — `update_right_panel_open_state` with `should_open == true`. **The diff-button path** that #9196 is about.
3. `view.rs:13372` — `PaneFocused` event, gated on `right_panel_open` already true.
4. `view.rs:13490` — `RepoChanged` event, gated on `right_panel_open` already true.
5. `view.rs:14458` — session env update, gated on `right_panel_open` already true.

None of these need the `show_code_review_button` gate — they're either explicit user actions or gated on `right_panel_open` already being open. The toolbar button toggle continues to do its job at `render_header_toolbar_button` independently.

### Testing

Reproduced @moirahuang's test locally on macOS 26.4.1 (Apple Silicon) against `WarpOss.app` built from this branch:

1. Settings → "Show code review button" → **OFF**
2. `echo "x" >> README.md` inside a git repo
3. Click the diff stats chip on the prompt (`+1 -0`)

**Result:** Code review panel opens on the right showing the diff, while the toolbar button stays hidden — exactly the expected behaviour from issue #9196. Inverse case (toggle ON) also verified: toolbar button visible, panel still works the same.

- `cargo fmt -p warp -- --check` passes.
- `cargo nextest` skipped locally — Metal toolchain unavailable on my machine, mirroring #9277. CI will exercise the change.

### Server API

No server changes.

### Agent Mode

Not applicable.

### Changelog Entries

`CHANGELOG-BUG-FIX`: The diff button on the terminal prompt now opens the code review panel even when the toolbar's "Show code review button" toggle is disabled (regression from a recent release).

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -7979,10 +7979,6 @@ impl Workspace {
         context: Option<&CodeReviewPaneContext>,
         ctx: &mut ViewContext<Self>,
     ) {
-        if !*TabSettings::as_ref(ctx).show_code_review_button {
-            return;
-        }
-
         // If context is provided, use it directly. Otherwise, derive from active pane group.
         let context_data: Option<(
             Option<PathBuf>,
@@ -18285,6 +18281,18 @@ impl Workspace {
                     self.render_config_panel_maximized(pane_group, &config, app),
                     app,
                 );
+            } else if !config.contains_item(&HeaderToolbarItemKind::CodeReview) {
+                Self::add_panel_with_separator(
+                    &mut main_content,
+                    &mut prev_panel_added,
+                    self.render_config_panel(
+                        &HeaderToolbarItemKind::CodeReview,
+                        pane_group,
+                        &config,
+                        app,
+                    ),
+                    app,
+                );
             }
         } else if !is_right_maximized {
             main_content = main_content.with_child(Shrinkable::new(1.0, terminal_content).finish());
@@ -18915,6 +18923,18 @@ impl Workspace {
                     self.render_config_panel_maximized(pane_group, &config, app),
                     app,
                 );
+            } else if !config.contains_item(&HeaderToolbarItemKind::CodeReview) {
+                Self::add_panel_with_separator(
+                    &mut panels_view,
+                    &mut prev_panel_added,
+                    self.render_config_panel(
+                        &HeaderToolbarItemKind::CodeReview,
+                        pane_group,
+                        &config,
+                        app,
+                    ),
+                    app,
+                );
             }
         }
 
@@ -18974,7 +18994,7 @@ impl Workspace {
     }
 
     /// Renders a configurable panel for the given toolbar item, if it is open.
-    /// Returns `None` if the panel should not be rendered (item not available,
+    /// Returns `None` if the panel should not be rendered (item not supported,
     /// panel not open, or item is not a panel type).
     fn render_config_panel(
         &self,
@@ -18983,7 +19003,7 @@ impl Workspace {
         config: &HeaderToolbarChipSelection,
         app: &AppContext,
     ) -> Option<Box<dyn Element>> {
-        if !item.is_available(app) || !item.is_panel() {
+        if !item.is_supported(app) || !item.is_panel() {
             return None;
         }
         match item {
@@ -19029,7 +19049,7 @@ impl Workspace {
         if !pane_group.right_panel_open || !pane_group.is_right_panel_maximized {
             return None;
         }
-        if !HeaderToolbarItemKind::CodeReview.is_available(app) {
+        if !HeaderToolbarItemKind::CodeReview.is_supported(app) {
             return None;
         }
         Some(Shrinkable::new(1.0, ChildView::new(&self.right_panel_view).finish()).finish())
--- a/app/src/workspace/tab_settings.rs
+++ b/app/src/workspace/tab_settings.rs
@@ -269,6 +269,10 @@ impl HeaderToolbarChipSelection {
             Self::Custom { right, .. } => right.clone(),
         }
     }
+
+    pub fn contains_item(&self, item: &super::header_toolbar_item::HeaderToolbarItemKind) -> bool {
+        self.left_items().contains(item) || self.right_items().contains(item)
+    }
 }
 
 settings::macros::implement_setting_for_enum!(
--- a/app/src/workspace/tab_settings_tests.rs
+++ b/app/src/workspace/tab_settings_tests.rs
@@ -1,5 +1,6 @@
 use super::*;
 use crate::test_util::settings::initialize_settings_for_tests;
+use crate::workspace::header_toolbar_item::HeaderToolbarItemKind;
 use settings::Setting;
 use warpui::{App, SingletonEntity};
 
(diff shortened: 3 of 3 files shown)


vorporeal:
reassigning to @moirahuang, based on some historical git blame information

app/src/workspace/view.rs
```diff
@@ -7969,9 +7969,11 @@ impl Workspace {
         context: Option<&CodeReviewPaneContext>,
         ctx: &mut ViewContext<Self>,
     ) {
-        if !*TabSettings::as_ref(ctx).show_code_review_button {
-            return;
-        }
+        // Note: do NOT gate this on `show_code_review_button` — that setting
```
moirahuang: nit: i think we don't need this comment! although this is helpful context for this PR, i think we don't need a comment explaining why code is _not_ there

moirahuang:
[I just tested this locally](https://www.loom.com/share/c2578afabaa74ee69df69267607de38b) and I'm not seeing the panel open w the setting off, mind locally confirming your change please?

anshul-garg27:
@moirahuang Good catch — you were right, the original fix was incomplete. Pushed `19cf7ba` which adds the missing piece.

## What was wrong

The original commit removed the `show_code_review_button` early-return from `setup_code_review_panel` (the data path), but a **second** gate was still active in the **render path** at `Workspace::render_config_panel` (`view.rs:18981`) and `render_config_panel_maximized` (`view.rs:19030`). Both were calling `item.is_available(app)`, which for `HeaderToolbarItemKind::CodeReview` returns `*TabSettings::as_ref(app).show_code_review_button.value()` (`header_toolbar_item.rs:89`).

Net effect:
- chip click → `open_right_panel` → `update_right_panel_open_state` set `pane_group.right_panel_open = true` ✅
- `setup_code_review_panel` ran without bailing out ✅ (original fix worked)
- but on the next render frame, `render_config_panel` saw `is_available()` return `false` because the toggle was off, returned `None`, so `right_panel_view` was **never added to the layout** — the panel stayed invisible. ❌

That matches your repro exactly.

## The fix

`is_available()` doc-comment specifically says *"Whether this item should be shown in the toolbar — checks both `is_supported` and user show/hide preferences."* Using it for panel rendering bundles two unrelated concerns: "is the feature compiled in" vs "does the user want the toolbar button visible." For the panel layer we only care about the former, so I switched both call sites to `is_supported(app)`.

`is_supported()` for `CodeReview` is just `cfg!(feature = "local_fs")` — a compile-time gate that's already enforced everywhere panel rendering touches.

This is safe for the other variants in the match: `TabsPanel` and `ToolsPanel` already returned `is_supported && true` from `is_available` (the default `_ => true` arm), and `AgentManagement`/`NotificationsMailbox` return `None` unconditionally inside the match.

## Verified locally

Built `WarpOss.app` with the fix on macOS 26.4.1 (Apple Silicon). Repro your test:

1. Settings → Show code review button → **OFF**
2. `echo "x" >> README.md` in a git repo
3. Click the diff chip on the prompt (`+1 -0`)

**Result:** code review panel opens on the right with the diff visible, while the toolbar button stays hidden — exactly what #9196 reports as the expected behaviour. Toolbar button gating is preserved (`render_header_toolbar_button` still uses `is_available` at `view.rs:17276`, which is correct for that path).

Format check passes. cargo nextest skipped locally (Metal toolchain unavailable, same as #9277). Ready for re-review.

app/src/workspace/view.rs
```diff
@@ -18974,16 +18976,22 @@ impl Workspace {
     }
 
     /// Renders a configurable panel for the given toolbar item, if it is open.
-    /// Returns `None` if the panel should not be rendered (item not available,
+    /// Returns `None` if the panel should not be rendered (item not supported,
     /// panel not open, or item is not a panel type).
+    ///
+    /// Uses `is_supported` instead of `is_available` because user show/hide
```
moirahuang: similar comment as [here](https://github.com/warpdotdev/warp/pull/9600/changes#r3169962070), i think we don't need this comment. it makes sense in the context of this PR but it's not needed in the actual code

app/src/workspace/view.rs
```diff
@@ -19020,6 +19028,9 @@ impl Workspace {
     }
 
     /// Renders the maximized code review panel if it is configured and maximized.
+    /// Uses `is_supported` instead of `is_available` for the same reason as
```
moirahuang: similar request here that we don't need this comment

moirahuang (commented):
thanks for updating the PR! please address the comments and add some unit tests and then i'll approve!

moirahuang:
i just realized there's one more edge case i think we should handle. if we remove the code review panel from the toolbar, we should handle still showing the code review panel
Screenshot 2026-05-01 at 11 31 28 AM

anshul-garg27:
@moirahuang Pushed `a75e74e` addressing all three points from your review.

## 1. Removed the three explanatory comments

`view.rs:7985`, `view.rs:18982`, `view.rs:19031` — all three "why this gate isn't here" comments are gone.

## 2. Toolbar-removed edge case

You're right — `config.left_items()` and `config.right_items()` skip Code Review entirely when the user has removed it from the toolbar layout, so the panel never renders even with `right_panel_open == true`.

Added a fallback after the right-side iteration in both render paths (horizontal-tabs at `view.rs:18277` and vertical-tabs at `view.rs:18923`):

```rust
} else if !config.contains_item(&HeaderToolbarItemKind::CodeReview) {
    Self::add_panel_with_separator(
        &mut main_content,
        &mut prev_panel_added,
        self.render_config_panel(&HeaderToolbarItemKind::CodeReview, …),
        app,
    );
}
```

The fallback is gated on the same `is_supported(app)` + `right_panel_open` checks already in `render_config_panel`, so it stays a no-op when the panel is closed or the feature isn't compiled in. The maximized branch already handled this case independently via `render_config_panel_maximized`, so the new fallback only runs for the non-maximized path.

To support that without scattering the `left.contains() || right.contains()` pattern, added `HeaderToolbarChipSelection::contains_item` in `tab_settings.rs`.

## 3. Unit tests

Four new tests in `tab_settings_tests.rs`:

- `header_toolbar_chip_selection_default_contains_code_review` — sanity: the default layout has Code Review on the right.
- `header_toolbar_chip_selection_custom_without_code_review_reports_absent` — models the exact scenario you flagged: user removes Code Review from both sides; `contains_item` returns `false` so the fallback render kicks in.
- `header_toolbar_chip_selection_custom_with_code_review_on_left_reports_present` — verifies side-insensitivity (moving the item from right to left still keeps it in config).
- `header_toolbar_chip_selection_custom_empty_reports_all_absent` — covers the degenerate empty-layout case across all toolbar items.

## Local verification

- `cargo fmt -p warp -- --check` passes.
- `cargo check --features=gui --tests -p warp` passes (3m 18s) — confirms my changes compile cleanly in the test profile.
- `cargo nextest` for the `warp` crate is currently broken on `upstream/master` for unrelated reasons (`warp_graphql` cynic+serde deserialization errors and `http_client` calling a renamed reqwest API `tls_built_in_webpki_certs` that's now `tls_built_in_root_certs`). Same errors reproduce on a clean `upstream/master` checkout without any of my changes, so they're pre-existing. Trusting CI to exercise the test suite end-to-end.

Branch rebased onto `525dfb6`.

moirahuang (approved):
looks good to me! thanks for making this PR

PR #9601, opened 2026-04-30
Don't add ToggleConversationListView menu item when its flag is off

### LINKED ISSUE #9535: warp-oss debug build crashes on launch: View menu adds flag-gated 'Toggle Conversation List View' item unconditionally
## Pre-submit Checks

- [x] I have searched Warp bugs and there are no duplicates.
- [x] I have searched Warp's known issues page and my issue is not there.

## Describe the bug

`warp-oss` (OSS channel, debug profile) panics during `applicationWillFinishLaunching`, before the main window finishes its first paint. The window flashes open and immediately disappears.

```
thread 'main' panicked at 'action should have a name: ToggleConversationListView',
app/src/app_menus.rs:103
```

In release builds the same code path doesn't crash (the failing assertion is `debug_assert!`), but the **View** menu silently renders the literal string `<NO DESCRIPTION>` for that item — so this is also a user-visible regression there.

## To reproduce

```bash
git clone https://github.com/warpdotdev/warp
cd warp
git lfs install
brew install protobuf      # see "Additional context"
cargo run --bin warp-oss --no-default-features --features=gui
```

The window appears for ~2-3 s, then the process exits with code 101.

## Expected behavior

The app launches successfully. The "Toggle Agent conversation list view" entry under **View** is either:

1. shown with its real label (`"Agent conversation list view"`, per `app/src/workspace/mod.rs:797-809`), or
2. omitted from the **View** menu while its feature flag is disabled.

## Root cause

`FeatureFlag::AgentViewConversationListView` (`crates/warp_features/src/lib.rs:623`) is **not present in `DOGFOOD_FLAGS`, `PREVIEW_FLAGS`, or `RELEASE_FLAGS`** (`crates/warp_features/src/lib.rs:868-952`), so it is off by default in every channel including OSS.

The two `EditableBinding`s that register this action's description and key bindings are both gated on that flag:

- `app/src/workspace/mod.rs:738-746` — custom-action binding, description `"Left Panel: Agent conversations"`.
- `app/src/workspace/mod.rs:797-809` — `cmd-shift-A` keystroke binding with the `MAC_MENUS_CONTEXT` override (`"Agent conversation list view"`).

Both carry `.with_enabled(|| FeatureFlag::AgentViewConversationListView.is_enabled())`, so when the flag is off they are filtered out by `Keymap::editable_bindings()` (`crates/warpui_core/src/keymap.rs:441-447`).

`make_new_view_menu` at **`app/src/app_menus.rs:384`** adds the menu item **unconditionally**:

```rust
updateable_custom_item_without_checkmark(CustomAction::ToggleConversationListView, ctx),
```

The chain is then:

1. `default_name(action, ctx)` calls `ctx.description_for_custom_action(action.into(), MAC_MENUS_CONTEXT)`.
2. `default_binding_for_custom_action` finds no binding because both candidates were filtered out by `is_enabled()`.
3. `description_for_custom_action` returns `None`.
4. `unwrap_or_else { debug_assert!(false, "action should have a name: {action:?}") }` at `app/src/app_menus.rs:103` fires → panic.

This is a general class of bug: any flag-gated UI surface added unconditionally to the menu bar will reproduce it the next time a flag is shipped off-by-default.

## Proposed fix

Two changes; either alone works, both together is the right shape.

### 1. Gate the menu item on the same flag the bindings use (root cause)

```rust
// app/src/app_menus.rs::make_new_view_menu
let mut items = vec![
    updateable_custom_item_without_checkmark(CustomAction::ToggleWarpDrive, ctx),
    MenuItem::Separator,
    updateable_custom_item_without_checkmark(CustomAction::CommandPalette, ctx),
    updateable_custom_item_without_checkmark(CustomAction::NavigationPalette, ctx),
    updateable_custom_item_without_checkmark(CustomAction::LaunchConfigPalette, ctx),
    updateable_custom_item_without_checkmark(CustomAction::FilesPalette, ctx),
];

if FeatureFlag::AgentViewConversationListView.is_enabled() {
    items.push(updateable_custom_item_without_checkmark(
        CustomAction::ToggleConversationListView,
        ctx,
    ));
}

items.extend([
    updateable_custom_item_without_checkmark(CustomAction::ToggleProjectExplorer, ctx),
    updateable_custom_item_without_checkmark(CustomAction::ToggleGlobalSearch, ctx),
    MenuItem::Separator,
    // ... rest unchanged
]);
```

### 2. Make `default_name` not panic on a missing description (defense in depth)

```diff
--- a/app/src/app_menus.rs
+++ b/app/src/app_menus.rs
@@ -100,7 +100,7 @@ fn default_name(action: CustomAction, ctx: &AppContext) -> String {
     ctx.description_for_custom_action(action.into(), bindings::MAC_MENUS_CONTEXT)
         .unwrap_or_else(|| {
-            debug_assert!(false, "action should have a name: {action:?}");
+            log::warn!("custom menu action has no description: {action:?}");
             "<NO DESCRIPTION>".into()
         })
 }
```

The current `debug_assert!` turns this class of bug into a debug-only crash that doesn't show up until someone runs `cargo run` directly. Logging matches release behavior and leaves a discoverable signal in `~/Library/Logs/warp-oss.log`.

## Screenshots, videos, and logs

```
2026-04-30T02:32:43Z [INFO]  [warpui::platform::mac::window] Opening window with id 0
2026-04-30T02:32:43Z [INFO]  [warpui::platform::mac::window] Using discrete GPU for rendering new window.
2026-04-30T02:32:46Z [ERROR] [panic] thread 'main' panicked at 'action should have a name: ToggleConversationListView': app/src/app_menus.rs:103
   ...
   9: {closure#0}                              at app/src/app_menus.rs:103:13
  11: default_name                             at app/src/app_menus.rs:102:10
  12: updateable_custom_item_with_checkmark    at app/src/app_menus.rs:126:10
  13: updateable_custom_item_without_checkmark at app/src/app_menus.rs:136:5
  14: make_new_view_menu                       at app/src/app_menus.rs:384:9
  15: menu_bar                                 at app/src/app_menus.rs:68:9
  22: warp_app_will_finish_launching           at crates/warpui/src/platform/mac/app.rs:309:38
  23: -[WarpDelegate applicationWillFinishLaunching:]
                                                at crates/warpui/src/platform/mac/objc/app.m:217:22
```

Full log: `~/Library/Logs/warp-oss.log`.

## Operating system

macOS

## Operating system and version

macOS 15 (Darwin 24.6)

## Shell version

zsh 5.x

## Current Warp version

Built from `master` against the pinned `v0.2026.04.29.08.56.stable_00` release tag (cloned 2026-04-30).

## Regression

No, this bug or issue has existed throughout my experience using Warp (first OSS clone).

## Additional context

Two adjacent papercuts encountered while reproducing this on a clean macOS 15 / Xcode 16.4 box, both worth folding into `script/macos/bootstrap`:

1. **`protoc` is required but not installed by bootstrap.** `warp_multi_agent_api`'s `build.rs` invokes `protoc` and fails with `Could not find` `protoc` until `brew install protobuf` is run.
2. **OSS bundle icons are missing.** `app/Cargo.toml` references `channels/oss/icon/no-padding/{512x512.png, icon.ico}`; that directory is populated by `script/install_channel_config`, which depends on SSH access to the private `warp-channel-config` repo. OSS contributors can't fetch it. `cargo bundle --bin warp-oss` therefore fails until the icon paths are stubbed. A committed-in OSS placeholder icon (or making the icon entry optional in the bundle metadata) would let `./script/run` succeed end-to-end for OSS contributors.

anshul-garg27:
Closes #9535.

### Description

`make_new_view_menu` in `app/src/app_menus.rs` was always inserting the `CustomAction::ToggleConversationListView` menu item:

```rust
updateable_custom_item_without_checkmark(CustomAction::ToggleConversationListView, ctx),
```

But that action's keybindings and description are gated on `FeatureFlag::AgentViewConversationListView` at both keybinding sites:

- `app/src/workspace/mod.rs:738-746` — `LEFT_PANEL_AGENT_CONVERSATIONS_BINDING_NAME`
- `app/src/workspace/mod.rs:797-809` — `TOGGLE_CONVERSATION_LIST_VIEW_BINDING_NAME` (which is the binding that registers the description used by `default_name`)

The flag is registered behind `#[cfg(feature = "agent_view_conversation_list_view")]` (`app/src/lib.rs:2677-2678`), so it stays off in OSS, dogfood, preview, and release. With the menu item present but no description registered, `default_name` (lines 100-106 of the same file) hits its `debug_assert!` and crashes `warp-oss` debug builds; release builds render `<NO DESCRIPTION>`.

### Fix

Split the menu's vec literal so `ToggleConversationListView` is `push`-ed conditionally on `FeatureFlag::AgentViewConversationListView.is_enabled()`, mirroring the gating already used at the keybinding sites. Order of remaining items is preserved.

I deliberately left the `debug_assert!` in `default_name` alone — it is correctly catching a real bug class, and softening it would mask the next misalignment of this kind.

### Testing

- `cargo fmt -p warp -- --check` — passes.
- I did not add a unit test. `app/src/app_menus.rs` has no existing `#[cfg(test)]` module or sibling `_tests.rs`, and constructing an `AppContext`, registering keybindings, and walking the resulting `Menu` to assert presence/absence of an item under both flag states would balloon the diff far beyond the one-line semantic gating change. Happy to add one if a maintainer prefers.
- Could not run `cargo nextest` locally (Metal toolchain unavailable on this machine — same situation as #9277).

### Server API

No server changes.

### Agent Mode

Not applicable.

### Changelog Entries

`CHANGELOG-BUG-FIX`: The View menu no longer crashes `warp-oss` debug builds (or renders `<NO DESCRIPTION>` in release) by including a flag-gated menu item without its description.

### DIFF
--- a/app/src/app_menus.rs
+++ b/app/src/app_menus.rs
@@ -381,7 +381,21 @@ fn make_new_view_menu(ctx: &AppContext) -> Menu {
         updateable_custom_item_without_checkmark(CustomAction::NavigationPalette, ctx),
         updateable_custom_item_without_checkmark(CustomAction::LaunchConfigPalette, ctx),
         updateable_custom_item_without_checkmark(CustomAction::FilesPalette, ctx),
-        updateable_custom_item_without_checkmark(CustomAction::ToggleConversationListView, ctx),
+    ];
+
+    // The `ToggleConversationListView` action only has a registered
+    // description and key bindings when this flag is enabled (see
+    // `crate::workspace::mod`). Inserting it unconditionally caused
+    // `default_name` to trip its `debug_assert!` in OSS/dogfood/preview
+    // builds and rendered `<NO DESCRIPTION>` in release builds.
+    if FeatureFlag::AgentViewConversationListView.is_enabled() {
+        items.push(updateable_custom_item_without_checkmark(
+            CustomAction::ToggleConversationListView,
+            ctx,
+        ));
+    }
+
+    items.extend([
         updateable_custom_item_without_checkmark(CustomAction::ToggleProjectExplorer, ctx),
         updateable_custom_item_without_checkmark(CustomAction::ToggleGlobalSearch, ctx),
         MenuItem::Separator,
@@ -435,7 +449,7 @@ fn make_new_view_menu(ctx: &AppContext) -> Menu {
             },
             None,
         )),
-    ];
+    ]);
 
     let is_compact_mode = matches!(
         TerminalSettings::handle(ctx)


vorporeal:
`agent_view_conversation_list_view` is in the `default` feature set, which, tbh, we never really expect anyone to disable - it's how we've launched features once we finished developing them.  (we ought to clean that up/use a different mechanism, but that's a separate matter.)

what command did you run which produced the crash?

vorporeal:
i'm going to close this because it's not the right fix - people shouldn't be running with `--no-default-features`, and the more correct fix here is to graduate the feature flag and make the feature unconditionally enabled.

PR #9613, opened 2026-04-30
feat(uri): add warposs://pane/{uuid} deep link for pane focus

Akeuuh:
## Description

Adds a `warposs://pane/{uuid_hex}` deep link that focuses a specific terminal pane by its persistent session UUID. This enables external tools to programmatically switch to a given pane/tab/window in Warp.

### What
- **`PaneGroup::find_pane_by_session_uuid`** — new method that iterates terminal panes and matches by persistent UUID bytes, returning the `PaneId`
- **`UriHost::Pane`** — new URI host variant with full handler: decodes 32-char hex UUID, walks all workspaces/windows via `WorkspaceRegistry::all_workspaces()`, dispatches `handle_pane_navigation_event` to focus the tab, and calls `show_window_and_focus_app` to bring the window forward
- **5 unit tests** — host parsing, full URI validation, empty path safety, invalid hex safety, case-insensitive hex decoding

### Why
External tools (e.g. orchestration UIs) that know a pane's UUID need a way to switch focus to that pane without user interaction. The pane UUID is already persisted in SQLite and is stable across sessions.

### How
- Handler parses hex from URL path, converts to `[u8; 16]` via `u8::from_str_radix` (no new crate dependency)
- Cross-window lookup via existing `WorkspaceRegistry::all_workspaces()` + `tab_views()` pattern (matches `focus_terminal_view_in_other_window`)
- `WindowBehaviorHint::Nothing` since handler explicitly manages window focus
- O(n) pane iteration — deep links are rare, no cache needed

## Testing

- [x] Unit tests for URI parsing and validation (5 tests in `uri_test.rs`)
- [x] Manual testing: `open "warposs://pane/{uuid}"` correctly focuses the target pane across tabs and windows


app/src/uri/mod.rs
```diff
@@ -450,6 +454,62 @@ impl UriHost {
...
+                    (0..16)
+                        .map(|i| u8::from_str_radix(&uuid_hex[i * 2..i * 2 + 2], 16).ok())
+                        .collect()
+                } else {
+                    None
+                };
+
+                let Some(uuid_bytes) = uuid_bytes else {
+                    log::warn!(
+                        "warposs://pane/ received invalid UUID hex (safe: len={})",
```
seemeroland, line 475: nit: let's exclude the `warposs` prefix since not all builds use this name - this is only for builds built from the open source repo. Our official stable binary is just `warp`

app/src/uri/mod.rs
```diff
@@ -450,6 +454,62 @@ impl UriHost {
...
+                    if let Some(root_view_id) = ctx.root_view_id(window_id) {
+                        ctx.dispatch_action_for_view(
+                            window_id,
+                            root_view_id,
+                            "root_view:handle_pane_navigation_event",
+                            &locator,
+                        );
+                    }
+                } else {
+                    log::warn!("warposs://pane/ could not find pane with given UUID");
```
seemeroland, line 510: same nit

seemeroland (commented):
The code looks good to me, but how do you anticipate users getting the `session_uuid` to use with this URI? I don't believe we expose this in a nice way. Were you thinking of reading from sqlite?

Akeuuh:
Reopening with renamed branch (axel/ prefix per convention)

PR #9618, opened 2026-04-30
[APP-4280] Async blocklist search

### LINKED ISSUE #9757: Async blocklist search
From david:

> The synchronous path scans every block (and all rich-content / AI blocks) on the main thread whenever the query changes, a block updates, or the active block scrolls. For large histories this can stall the UI for noticeable amounts of time.

DStern has an open PR for this that I'm taking over!

---
Linear: [APP-4280](https://linear.app/warpdotdev/issue/APP-4280/async-blocklist-search)

vkodithala:
Closes #9757.

## Description (from @vorporeal):

Moves find in the blocklist off the main thread so that searching large terminal histories no longer causes UI freezes.

The synchronous path scans every block (and all rich-content/AI blocks) on the main thread whenever the query changes, a block updates, or the active block scrolls. For large histories this can stall the UI for noticeable amounts of time. This PR introduces an `AsyncFindController` that runs searches on a background task and streams results back to the main thread.

Key pieces:

- `AsyncFindController` (`app/src/terminal/find/model/async_find.rs`): controller owned by `TerminalFindModel`, hidden behind the `AsyncFind` feature flag. Owns the current config, per-block results, focused match state, a shared work queue, and the spawned task handle. Implements the same `FindModel` surface (`match_count`, `focused_match_index`, `focus_next_match`, etc.) used by the existing sync path.
- `FindWorkQueue` (`async_find/work_queue.rs`): a shared queue of `ScanFullBlock` / `ScanDirtyRange` / `ScanAIBlock` items. The background task awaits it via an `event_listener::Event`, so there is no polling. `invalidate_block` pushes dirty work to the front so newly-changed blocks are prioritized; pending full-block rescans for the same block are deduped.
- `background_task` (`async_find/background_task.rs`): the task loop pulls work items, builds `RegexDFAs` once per run, and streams `FindTaskMessage` results back over an `async_channel`. Grid scanning is chunked (`ROWS_PER_CHUNK`, `MAX_LOCK_DURATION_MS`) so the terminal model lock is never held long enough to block the main thread, and the task yields between chunks via `futures_lite::yield_now`. AI blocks are bounced back to the main thread via `ScanAIBlock` since rich-content search runs there.
- Coordinate handling: matches are stored as `AbsoluteMatch` (absolute row indices from `AbsolutePoint`). This means scrollback truncation does not require re-indexing existing results — truncated matches are filtered on read via `AbsoluteMatch::to_range`. Dirty-range updates use `BlockFindResults::update_dirty_matches` to splice replacements into the existing ordered vector.
- Rendering: a new `BlockFindRenderData` abstraction in `app/src/terminal/find/model.rs` gives `block_list_element.rs` a single interface over both the sync and async paths, and pre-converts async absolute ranges to relative `Point` ranges per grid. `GridHandler` now accumulates a `find_dirty_rows_range` during byte processing (`take_find_dirty_rows_range`) so we can hand the background task the exact rows that changed instead of rescanning whole blocks.
- UI integration: `rerun_find_on_active_grid` and a new `notify_block_completed` path on `TerminalFindModel` feed dirty ranges and block completions into `invalidate_block`. A throttled channel in the controller coalesces result bursts into at most one `FindEvent::RanFind` per 50 ms, and a monotonic `generation` counter guards against stale `Done` messages ending a newer scan. The find bar renders a "Scanning..." / `N+ ...` indicator while the background task is still running.
- Feature flag: `FeatureFlag::AsyncFind` gates creation of the controller; the sync path is the default everywhere else. For now (while @vkodithala fixes up issues in "Work still needed"), we disable this path for all builds, including dogfood. Using async find is restricted to app runs with the feature flag: `./script/run --features async_find`.

### Work still needed

- For find queries on AI blocks, up/down arrows advance the ?/xx indicator on the find bar, but don't actually highlight a match (actually not seeing any matches _ever_ highlighted). This seems like a bug, but restricted to AI blocks (terminal blocks are highlighted fine).
- Match index isn't being preserved in terminal blocks as results stream in. Repro: search "info" on `./script/run` and select the second match from the bottom of the blocklist. As new results inevitably stream in, that match index isn't preserved (defaults to the most recent match). Regression from dev.
- Visible flicker with a find query when new results stream in within a terminal block, also an adjacent app crash with error message: `Matches should be in ascending order after update_dirty_matches.`
- I think the search direction within a terminal block isn't quite right - I _think_ this is searching top-to-bottom, but we should search bottom-to-top to surface recent results as soon as possible.  (though we may want to actually search top-to-bottom for pin-to-top input mode?)
- There's a big TODO in there around optimizing query refinements - that's not a blocker but would be good to do.  Will speed things up while the user types, as we don't have to re-search things they've already searched for.

## Linked Issue

## Testing

- New unit test module `async_find_tests.rs` (~800 lines, 18 tests) covering: parity with sync find, cancellation, streaming message handling, focus navigation / wraparound ordering for both `BlockSortDirection`s, dirty-range merging (prepend / append / replace-middle / clear / empty-existing), absolute-match truncation, and `BlockFindResults` bookkeeping (`total_match_count`, `remove_block`).
- Manual dogfood via the feature flag on large blocklists to confirm the UI stays responsive while scanning and that incremental updates to the active block's output do not trigger full rescans.

## Agent Mode


vkodithala:
* **#9618** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

specs/async-find/TECH.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Two find paths in one model
+
+`TerminalFindModel` gains an `async_find_controller: Option<AsyncFindController>` field, populated in `new()` iff `FeatureFlag::AsyncFind.is_enabled()` (`app/src/terminal/find/model.rs:243`). Every existing `TerminalFindModel` method now branches:
+
+```rust path=null start=null
+if model.is_alt_screen_active() { /* alt-screen path, unchanged */ }
+else if let Some(controller) = &mut self.async_find_controller { /* async path */ }
+else { /* legacy sync path, unchanged */ }
+```
```
vkodithala, line 43: From my understanding, there are some 3rd-party agents that _do_ render in alt screens (OpenCode and Claude Code no-flicker mode, I think). It seems like we're running find on alt screens synchronously in all cases, maybe as a fast follow we can extend this logic to alt screens if not totally impossible?

app/src/terminal/find/model/async_find.rs
```diff
@@ -0,0 +1,1194 @@
...
+        &mut self,
+        options: &FindOptions,
+        block_sort_direction: BlockSortDirection,
+        ctx: &mut ModelContext<TerminalFindModel>,
+    ) {
+        // For now, we do a full rescan when the query is refined.
+        // A future optimization could filter existing matches without rescanning,
+        // and update the query for pending queue items.
+
+        // Create new config.
```
vkodithala, line 969: Ack, can probably include in a fast follow

specs/async-find/TECH.md
```diff
@@ -0,0 +1,196 @@
...
+`FindWorkQueue` (`work_queue.rs`) is a deliberately small primitive:
+
+- `enqueue_full_scan(blocks)` pushes `FullBlock` / `AIBlock` items at the back, in the newest-first order produced by `collect_block_info`.
+- `invalidate_block(idx, dirty_range)` pushes a `DirtyRange` (or `FullBlock` if no range) at the *front*, so reactive work jumps the queue ahead of the initial scan. It is also a no-op if a `FullBlock` for the same block is already pending.
+- `pop()` is async and uses `event_listener::Event` to block when empty without busy-waiting. It returns `(item, queue_drained: bool)` so the consumer can decide atomically when to emit `FindTaskMessage::Done` — no separate `is_empty()` call, no TOCTOU race.
+- `close()` flips a flag and notifies all listeners so a parked `pop()` returns `Err(QueueClosed)`.
+
+`background_task::run_find_task_loop` (`background_task.rs (56-128)`) builds `RegexDFAs` once from the config, then:
+
+- For `FullBlock { block_index }`: calls `scan_terminal_block_chunked`, which iterates the grid in the block-sort-direction-aware order and delegates each grid to `scan_grid_chunked`.
```
vkodithala, line 82: From my understanding, we match on chunks of arbitrary size 1k. Does this fail if a match falls on a chunk boundary? Should we instead end chunks on newlines (I don't think our DFA/Regex based find extends to multiple lines?)

specs/async-find/TECH.md
vkodithala: I think there's an adjacent issue, which is that we don't search above the current view for alt screens (but also probably fine to de-prio if it hasn't already been flagged)?

specs/async-find/TECH.md
```diff
@@ -0,0 +1,196 @@
...
+
+5. **Truncation during dirty-range scans.** Whenever the active block's output is updated, `prune_truncated_matches` runs on every `DirtyRangeMatches` to drop matches whose start row has been truncated; the focused index is clamped. Without this, `focused_match_index` could point past the end of the result set and `focused_terminal_match()` would return `None` even though there are matches.
+
+6. **Rollout blast radius.** Hidden behind `FeatureFlag::AsyncFind`. The cargo feature `async_find` defaults off; the runtime flag is on for dogfood only. Disabling is a one-line revert (remove from `DOGFOOD_FLAGS`). The sync path is unchanged so the worst-case fallback is "behavior identical to today".
+
+## Follow-ups
+
+- True in-place query refinement (today `filter_results_for_refinement` rescans). The plumbing — `is_query_refinement`, `current_find_options`, the queue's `invalidate_block` — is already in place.
+- Move AI block scanning off the main thread (currently still synchronous via `ScanAIBlock` round-trip).
```
vkodithala, line 195: Curious if there's any functional reason we don't also scan AIBlocks on background threads, like we do for terminal blocks? i.e., is this a design thing or just OOS?

vkodithala:
Did an initial pass via back-filled tech/product specs, think this mostly makes sense. The intent seems pretty clear, which is that we're moving from a synchronous scan of each terminal/AI block in the blocklist to a mostly async scan, and wiring up everything necessary to support this.

Noticed a few issues while dogfooding:

- For find queries on AI blocks, up/down arrows advance the ?/xx indicator on the find bar, but don't actually highlight a match (actually not seeing any matches _ever_ highlighted). This seems like a bug, but restricted to AI blocks (terminal blocks are highlighted fine)
- Match index isn't being preserved in terminal blocks as results stream in. Repro: search "info" on `./script/run` and select the second match from the bottom of the blocklist. As new results inevitably stream in, that match index isn't preserved (defaults to the most recent match). Seeing different behavior on Dev
- Visible flicker with a find query when new results stream in within a terminal block, also an adjacent app crash with error message: `Matches should be in ascending order after update_dirty_matches`

And have a few follow-up qs for ya @vorporeal (mostly non-blocking & out of curiosity). Hoping to wrap this up soon!

specs/async-find/TECH.md
vorporeal: alt screen is tiny by comparison - it never contains scrollback.  we could potentially also do async scanning of the alt screen, but it would add complexity and probably zero meaningful benefit.

specs/async-find/TECH.md
vorporeal: the terminal model is shareable across threads (it's in an `Arc<Mutex<...>>`), but AFIAK AI blocks are not - they're UI framework entities, and as such, can only be accessed from the main thread.

that said, each AI block tends to be quite small, in the scheme of things, so scanning a full one shouldn't take very long.  (by comparison, a terminal block could be as many as 1M lines long.)

specs/async-find/TECH.md
vorporeal: i believe this was something that the agent addressed and wrote a test to cover, but one of us would need to check again.

vorporeal:
> Did an initial pass via back-filled tech/product specs, think this mostly makes sense. The intent seems pretty clear, which is that we're moving from a synchronous scan of each terminal/AI block in the blocklist to a mostly async scan, and wiring up everything necessary to support this.
> 
> Noticed a few issues while dogfooding:
> 
> - For find queries on AI blocks, up/down arrows advance the ?/xx indicator on the find bar, but don't actually highlight a match (actually not seeing any matches _ever_ highlighted). This seems like a bug, but restricted to AI blocks (terminal blocks are highlighted fine)

yeah i'm not surprised there are some bugs there - correct match highlights were one piece that the agent consistently struggled with.

> - Match index isn't being preserved in terminal blocks as results stream in. Repro: search "info" on `./script/run` and select the second match from the bottom of the blocklist. As new results inevitably stream in, that match index isn't preserved (defaults to the most recent match). Seeing different behavior on Dev

yep should definitely fix this one.

> - Visible flicker with a find query when new results stream in within a terminal block, also an adjacent app crash with error message: `Matches should be in ascending order after update_dirty_matches`

might relate to the two other findings (bugs in the match order/highlight index)

> And have a few follow-up qs for ya @vorporeal (mostly non-blocking & out of curiosity). Hoping to wrap this up soon!

specs/async-find/TECH.md
vkodithala: Yeah, this was my understanding too. Appreciate the explanation!

specs/async-find/TECH.md
vkodithala: Gotcha, I can take a look here. Thanks!

specs/async-find/TECH.md
vkodithala: Yeah, that makes sense. Appreciate the context.

vorporeal (approved):

Issue #6678, opened 2025-06-26
Custom Theme Path Resolution


graysoncash:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

I have multiple Macs that I have Warp installed on. I'm using the same Warp account and have settings sync turned on for each. I also have a custom theme that I've put in the same location for each machine under `$HOME/.warp/themes`. I've noticed that with a previous machine, I didn't have an issue with setting the theme across machines because they had the same account name and home directory path. However, the new machine doesn't and when I go to set a theme, even though the underlying directory structure for the theme files are the same relative to the home directory, the machine that just synced the new setting value gets a black theme. It seems like because it was unable to resolve the file, it wasn't able to set a theme correctly.

In order to be able to sync custom themes across machines (not the actual file, just the path), it would be helpful if a custom theme's path is not synced by its fully qualified path, but rather, a relative path from the custom themes root directory. Since the custom themes path is not configurable, it does not appear to make sense to sync the fully qualified path. This would also help to keep this synced setting more platform generic.

To get the synced path value, I executed the following:

```
> defaults read dev.warp.Warp-Stable SelectedSystemThemes | python3 -c "import sys, json; print(json.dumps(json.loads(sys.stdin.read()), indent=2))"

{
  "light": {
    "Custom": {
      "name": "Some Theme Light",
      "path": "/Users/<username>/.warp/themes/some_theme/some_theme_light.yml"
    }
  },
  "dark": {
    "Custom": {
      "name": "Some Theme Dark",
      "path": "/Users/<username>/.warp/themes/some_theme/some_theme_dark.yml"
    }
  }
}
```

### Operating system (OS)

macOS

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

Gabbo7474:
Any update on this! This seems like an easy fix and would greatly improve QOL so we don't have to manually change the theme every time we change machine.

PR #9634, opened 2026-04-30
fix(themes): Handle tilde in theme paths

rrroyal:
## Description
When using settings sync between machines with different usernames, custom theme gets reset every time the settings sync. [_All settings reference_ mention `~/.warp/themes/my-theme.yaml`](https://docs.warp.dev/terminal/settings/all-settings#themes), so figured it should support it. It didn't, so I (with some help from Copilot) have added it.

## Linked Issue
Mainly targeted at my own use case, but probably also #6678.

## Screenshots / Videos
Screenshot 2026-04-30 at 19 21 19

## Testing
Verified in real-world with `./script/run`/`cargo run` using paths with and without `~`.
**Tested only on macOS** - I don't have access to a Windows or Linux machine as of right now.

Also added tests for themes (`theme_test.rs`) and `warp_core` (`theme_tests.rs`). If they seem redundant, please let me know and I'll remove one of them.

## Agent Mode
- [x] Copilot-assisted - This PR was created using Copilot assistance.

---

CHANGELOG-IMPROVEMENT: Added support for tilde (`~`) in custom theme paths.

### DIFF
--- a/app/src/themes/theme.rs
+++ b/app/src/themes/theme.rs
@@ -151,16 +151,7 @@ impl ThemeKind {
 }
 
 #[derive(
-    Debug,
-    Clone,
-    Hash,
-    PartialEq,
-    Eq,
-    Serialize,
-    Deserialize,
-    PartialOrd,
-    Ord,
-    schemars::JsonSchema,
+    Debug, Clone, Hash, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord, schemars::JsonSchema,
 )]
 #[schemars(description = "A user-provided custom theme.")]
 pub struct CustomTheme {
--- a/app/src/themes/theme.rs
+++ b/app/src/themes/theme.rs
@@ -1,9 +1,9 @@
 use super::default_themes::*;
 use anyhow::Result;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use std::collections::HashMap;
 use std::iter::FromIterator;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use warp_core::ui::color::pick_foreground_color;
 use warpui::assets::asset_cache::AssetSource;
 use warpui::{
@@ -161,14 +161,37 @@ impl ThemeKind {
     PartialOrd,
     Ord,
     schemars::JsonSchema,
-    settings_value::SettingsValue,
 )]
 #[schemars(description = "A user-provided custom theme.")]
 pub struct CustomTheme {
     #[schemars(description = "The display name of the custom theme.")]
     name: String,
+    #[serde(
+        deserialize_with = "deserialize_tilde_path",
+        serialize_with = "serialize_tilde_path"
+    )]
     #[schemars(description = "The file path to the custom theme definition.")]
     path: PathBuf,
+}
+
+// The SettingsValue derive bypasses serde, so use the serde passthrough instead.
+// This ensures the #[serde(deserialize_with/serialize_with)] annotations on
+// the path field are respected when reading/writing settings.toml.
+impl settings_value::SettingsValue for CustomTheme {}
+
+fn deserialize_tilde_path<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let path = PathBuf::deserialize(deserializer)?;
+    Ok(expand_tilde(path))
+}
+
+fn serialize_tilde_path<S>(path: &Path, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    contract_tilde(path.to_path_buf()).serialize(serializer)
 }
 
 impl CustomTheme {
--- a/app/src/themes/theme.rs
+++ b/app/src/themes/theme.rs
@@ -165,9 +165,7 @@ pub struct CustomTheme {
     path: PathBuf,
 }
 
-// The SettingsValue derive bypasses serde, so use the serde passthrough instead.
-// This ensures the #[serde(deserialize_with/serialize_with)] annotations on
-// the path field are respected when reading/writing settings.toml.
+// Use serde passthrough to respect #[serde] annotations on path field for tilde expansion.
 impl settings_value::SettingsValue for CustomTheme {}
 
 fn deserialize_tilde_path<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
--- a/app/src/themes/theme.rs
+++ b/app/src/themes/theme.rs
@@ -172,15 +172,25 @@ fn deserialize_tilde_path<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
 where
     D: Deserializer<'de>,
 {
-    let path = PathBuf::deserialize(deserializer)?;
-    Ok(expand_tilde(path))
+    // Normalize backslashes to forward slashes for cross-platform compatibility.
+    let raw = String::deserialize(deserializer)?;
+    let normalized = raw.replace('\\', "/");
+    Ok(expand_tilde(PathBuf::from(normalized)))
 }
 
 fn serialize_tilde_path<S>(path: &Path, serializer: S) -> Result<S::Ok, S::Error>
 where
     S: serde::Serializer,
 {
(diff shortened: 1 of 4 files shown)


kevinchevalier:
Thank you for the submission! This works really well.

Can you run `cargo fmt` to fix the formatting errors and then I will merge it in.

rrroyal:
> Thank you for the submission! This works really well.
> 
> 
> 
> Can you run `cargo fmt` to fix the formatting errors and then I will merge it in.

done @kevinchevalier :)

thanks for the quick review!

app/src/themes/theme_test.rs
```diff
@@ -1,5 +1,133 @@
 use super::*;
 use crate::util::color::OPAQUE;
+use settings_value::SettingsValue as _;
+
+#[test]
+#[cfg(not(target_family = "wasm"))]
+fn custom_theme_tilde_path_expansion_test() {
+    use dirs::home_dir;
```
lucieleblanc: Style nit: imports should go at the top of the file.

crates/warp_core/src/ui/theme/theme_tests.rs
```diff
@@ -339,3 +339,31 @@ fn infer_from_foreground_color_test() {
...
+
+#[cfg(not(windows))]
+#[test]
+fn contract_tilde_replaces_home_prefix() {
+    let home = dirs::home_dir().expect("home dir must exist");
+    let abs = home.join(".warp/themes/my_theme.yaml");
+    let contracted = contract_tilde(abs.clone());
+    assert_eq!(
+        contracted,
+        std::path::PathBuf::from("~/.warp/themes/my_theme.yaml")
```
lucieleblanc, line 351: Style nit: import at the top of the file instead of qualifying inline.

crates/warp_core/src/ui/theme/theme_tests.rs
```diff
@@ -339,3 +339,31 @@ fn infer_from_foreground_color_test() {
...
+    );
+}
+
+#[cfg(not(windows))]
```
lucieleblanc, line 355: We should make sure this feature works on Windows, too. Could you add Windows-specific tests too?

lucieleblanc (changes requested):
Thank you for the PR! Requesting changes because we should make sure this works on windows before merging it.

We should at least add tests to make sure tilde expansion and Windows paths (with backslashes) work too. I left some comments on the Windows portability aspect on a related PR [here](https://github.com/warpdotdev/warp/pull/9728#pullrequestreview-4223688635), in case the example there is useful.

crates/warp_core/src/ui/theme/theme_tests.rs
rrroyal: should be good now :)

lucieleblanc: The tests in this file are all non-Windows because `PathBuf::from("~").join(...)` uses native separators on Windows, making the assertions platform-dependent. This is fine for testing the function in isolation, but it means there's no test that validates the full serialize/deserialize round trip on Windows. Please add unit tests for Windows in this file too!

app/src/themes/theme.rs
```diff
@@ -151,26 +151,48 @@ impl ThemeKind {
...
+// Use serde passthrough to respect #[serde] annotations on path field for tilde expansion.
+impl settings_value::SettingsValue for CustomTheme {}
+
+fn deserialize_tilde_path<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    // Normalize backslashes to forward slashes for cross-platform compatibility.
+    let raw = String::deserialize(deserializer)?;
+    let normalized = raw.replace('\\', "/");
```
lucieleblanc, line 177: I'm not convinced this works on all platforms. The `\` is valid in UNIX paths. A valid UNIX path like `/home/user/.warp/themes/my\theme.yaml` would be silently corrupted to `/home/user/.warp/themes/my/theme.yaml`. Should this line be gated to be Windows-only?

app/src/themes/theme.rs
```diff
@@ -151,26 +151,48 @@ impl ThemeKind {
...
+
+fn serialize_tilde_path<S>(path: &Path, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    let contracted = contract_tilde(path.to_path_buf());
+    // Normalize to forward slashes for cross-platform portability.
+    // Windows would emit backslashes, which Unix treats as filename characters.
+    let portable = contracted
+        .components()
```
lucieleblanc, line 189: `Component::RootDir::as_os_str()` returns the platform's main separator — `\` on Windows, `/` on Unix. So for a Windows absolute path that isn't under the home directory (and therefore isn't tilde-contracted), like `D:\themes\my_theme.yaml`:

- Components: `["D:", "\", "themes", "my_theme.yaml"]`
- After `.join("/")`: `"D:/\/themes/my_theme.yaml"`

The backslash from `RootDir` survives the join. 
I think a simpler replacement like this could work:
```rust
let portable = contracted.to_string_lossy().replace('\\', "/");
```

lucieleblanc (changes requested):
Thanks for the update! I think we're still not quite there on Windows support, but close. I'll enable CI on this PR so we can see the Windows test failures.

Gabbo7474:
Hey @rrroyal! This is a great feature — I've been hitting the same settings sync issue. Noticed the Windows test is still failing though, and wanted to offer a fix since it's been a few weeks.

## Issues

### 1. `serialize_tilde_path` produces `C:/\/Users/...` on Windows

The `.components().join("/")` approach breaks because `Component::RootDir` on Windows is `\`, which survives the join and creates `C:/\/...`.

**Fix:** replace the components logic with a simple string replace:

```rust
fn serialize_tilde_path<S>(path: &Path, serializer: S) -> Result<S::Ok, S::Error>
where
    S: serde::Serializer,
{
    let contracted = contract_tilde(path.to_path_buf());
    #[cfg(windows)]
    let portable = contracted.to_string_lossy().replace('\\', "/");
    #[cfg(not(windows))]
    let portable = contracted.to_string_lossy().into_owned();
    portable.serialize(serializer)
}
```

### 2. `deserialize_tilde_path` corrupts valid Unix paths

On Unix, `\` is a valid filename character. A path like `~/.warp/themes/my\theme.yaml` would be silently corrupted to `~/.warp/themes/my/theme.yaml`. The backslash normalization should be Windows-only:

```rust
fn deserialize_tilde_path<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
where
    D: Deserializer<'de>,
{
    let raw = String::deserialize(deserializer)?;
    #[cfg(windows)]
    let raw = raw.replace('\\', "/");
    Ok(expand_tilde(PathBuf::from(raw)))
}
```

### 3. Missing Windows tests in `crates/warp_core/src/ui/theme/theme_tests.rs`

The existing tests there are all `#[cfg(not(windows))]`. Here are Windows equivalents:

```rust
#[cfg(windows)]
#[test]
fn contract_tilde_replaces_home_prefix_windows() {
    let home = dirs::home_dir().expect("home dir must exist");
    let abs = home.join("AppData\\Roaming\\warp\\Warp\\data\\themes\\my_theme.yaml");
    let contracted = contract_tilde(abs);
    let expected = std::path::PathBuf::from("~").join("AppData\\Roaming\\warp\\Warp\\data\\themes\\my_theme.yaml");
    assert_eq!(contracted, expected);
}

#[cfg(windows)]
#[test]
fn contract_tilde_leaves_non_home_paths_unchanged_windows() {
    let path = std::path::PathBuf::from("D:\\themes\\my_theme.yaml");
    assert_eq!(contract_tilde(path.clone()), path);
}

#[cfg(windows)]
#[test]
fn expand_contract_tilde_round_trips_windows() {
    let tilde_path = std::path::PathBuf::from("~\\AppData\\Roaming\\warp\\themes\\my_theme.yaml");
    let expanded = expand_tilde(tilde_path.clone());
    let contracted = contract_tilde(expanded);
    assert_eq!(contracted, tilde_path);
}
```

Happy to open a PR against your branch if you'd prefer — just let me know!

rrroyal:
I'm pretty sure #9728 fixes the same issue in a different way, so this PR isn't needed anymore

PR #9566, opened 2026-04-30
fix(agent): clear stale permission text from tab title after agent stops

Giggitycountless:
## Summary

When using Claude Code (or other CLI agents), the tab title continues to show the text from the last `PermissionRequest` event instead of the agent's final response after the session completes.

Fixes #9525

## Root Cause

1. `PermissionRequest` event sets `self.session_context.summary` to e.g. "Wants to run bash: ..."
2. `Stop` event sets `self.session_context.response` but never clears `summary`
3. `display_title()` checks `query` then falls back to `summary`, never consulting `response`

## Changes

- **Clear `summary` on `Stop` event**: Since the permission request text is no longer relevant after the session completes, set `summary = None` when processing a `Stop` event.
- **Add `response_text()` fallback**: Add a new `response_text()` method and use it as a fallback in `display_title()` between `query` and `summary`, so the agent's final response is shown when no user prompt is available.

## Testing

- Verified that after a Claude Code session completes, the tab title shows the agent's final response instead of stale permission request text.
- Verified that during an active session with a pending permission request, the title still shows the permission text correctly.

PR #9637, opened 2026-04-30
fix: clear stale permission text from tab title after agent stops

Giggitycountless:
When a CLI agent finishes (Stop event), the summary field still holds the last PermissionRequest text. This causes the tab title to show stale permission prompts instead of the agent's final response.

Fix: clear summary on Stop so title_like_text() returns None and the title falls back to the conversation title or terminal title.

Changes:
- cli_agent_sessions/mod.rs: Add  on Stop event
- cli_agent_sessions/mod_tests.rs: Add regression test for PermissionRequest → Stop transition

This is a cleaner version of #9566 that addresses the review feedback:
- Removed the redundant response_text() fallback
- Only clears summary (minimal fix)
- Added regression test to prevent future regressions

### DIFF
--- a/app/src/terminal/cli_agent_sessions/mod.rs
+++ b/app/src/terminal/cli_agent_sessions/mod.rs
@@ -176,6 +176,10 @@ impl CLIAgentSession {
             CLIAgentEventType::Stop => {
                 self.session_context.query = event.payload.query.clone();
                 self.session_context.response = event.payload.response.clone();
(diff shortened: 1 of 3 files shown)


Giggitycountless:
Closing this PR as it has been through multiple iterations without reaching consensus. The changes are too extensive for the scope of the fix. Thank you for the review feedback!

PR #9571, opened 2026-04-30
fix(build): verify Yarn 4+ from js crate dir in bootstrap scripts

Giggitycountless:
## Summary

This PR adds Node.js (>= 18.14.1) and Yarn 4+ prerequisite checks to both the Linux and macOS bootstrap scripts, and documents the prerequisites in WARP.md.

### Bootstrap scripts (`script/linux/install_build_deps` and `script/macos/bootstrap`):
- **Node.js check**: Verify that `node` is on PATH and is version >= 18.14.1 (required by the `command-signatures-v2` crate's `build.rs` TypeScript compilation)
- **Yarn check**: Verify that `yarn` is on PATH and is version >= 4 (Corepack-managed), rejecting Yarn 1.x from Homebrew
- **Yarn version check runs from `crates/command-signatures-v2/js`** where Corepack reads `packageManager` from `package.json`, matching the actual build environment
- Clear error messages with installation guidance

### Documentation (`WARP.md`):
- Added "Prerequisites" section listing Node.js, yarn, and protoc requirements
- Added link to `script/linux/install_build_deps` for detailed dependency info

## Motivation

The `command-signatures-v2` crate's `build.rs` runs `yarn install` and `tsc` during compilation, which requires:
- Node.js >= 18.14.1 (for the TypeScript toolchain)
- Yarn 4+ via Corepack (the crate's `package.json` pins `packageManager: yarn@4.0.1`)
- protoc >= 3.15 (already checked by the scripts)

Without these checks, `cargo build` fails with cryptic errors like "No such file or directory (os error 2)" when running the build script. This PR catches the problem early with actionable guidance.

## Testing

- Verified both scripts parse correctly with `bash -n`
- Confirmed the version comparison logic handles edge cases (e.g., `18.14.0 < 18.14.1`, `19.0.0 > 18.14.1`, `1.22.19 < 4.0.1`)
- Yarn version check runs from the js crate directory to match build.rs behavior

PR #9638, opened 2026-04-30
fix: extract Node.js/Yarn checks into shared script

Giggitycountless:
The Node.js and Yarn 4+ prerequisite checks were duplicated across bootstrap scripts. This PR extracts them into a shared script that's sourced from each location.

Changes:
- Create script/check_node_yarn_deps.sh with the validation logic
- Source it from script/linux/bootstrap and script/macos/bootstrap (NOT from install_build_deps since CI calls that before Node is installed)
- Add prerequisite docs to WARP.md

This is a cleaner version of #9571 that addresses the review feedback:
- Eliminates duplicated code
- Uses consistent version comparison logic
- Single source of truth for the checks
- Only checks in bootstrap scripts (developer-facing), not in CI-facing scripts

Fixes #9571

### DIFF
--- a/WARP.md
+++ b/WARP.md
@@ -41,6 +41,10 @@ Environment variables:
 - `./script/install_cargo_build_deps` - Install Cargo build dependencies
 - `./script/install_cargo_test_deps` - Install Cargo test dependencies
 
+### Prerequisites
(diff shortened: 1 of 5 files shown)


Giggitycountless:
Closing this PR as it's a refactoring iteration (4th version of the same feature). The shared script approach is cleaner but the earlier PRs (#9565→#9570→#9571) already established the pattern. Apologies for the noise.

PR #9455, opened 2026-04-29
implement basic local cloud handoff UI

harryalbert:
## Description

This PR implements the UI for the local -> cloud handoff flow in the client. This is done levaraging the workspace-discovery and snapshotting functionality added in the PRs below this one. On chip-click, we snapshot the conversation's touched files and repos, and create a new cloud mode pane to kick off a run. When the user submits a query, we kick that run off.

Don't over-index on the UI too much — in a follow-up PR I'll add the conversation into the cloud mode pane and also re-use the full cloud mode setup v2 UI.

The associated server PR for this client PR is here: https://github.com/warpdotdev/warp-server/pull/10777

## Testing

demo: https://www.loom.com/share/a6caa2c974e34b49b2b038a8019c062c

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

PR #9653, opened 2026-04-30
Fork conversation into local->cloud pane on-pane-creation

harryalbert:
## Description

We want to use the same new cloud mode setup UI for local -> cloud handoff that we have for regular cloud mode setup. We also want to immediately fork the entire conversation into the new cloud pane before you've actually starting the cloud run. 

The main change required to do this was to fork the conversation on-chip-click instead of on-prompt-send (see the corresponding server PR for this issue), as we don't want the forked conversation to get stale if new prompts are sent to the old conversation between clicking the chip and sending a prompt in the cloud mode window.

## Screenshots / Videos

https://www.loom.com/share/580624a891d64bec81f9642dda766583

## Agent Mode


harryalbert:
* **#10183** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10183?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9988** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9988?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#9653** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9653?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9653?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#9455** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9455?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10102** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10101** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10101?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

app/src/terminal/view/ambient_agent/view_impl.rs
```diff
@@ -144,14 +142,11 @@ impl TerminalView {
                 }
                 if FeatureFlag::CloudModeSetupV2.is_enabled() {
                     let view_model = ambient_agent_view_model.as_ref(ctx);
-                    let use_queued_prompt = view_model.is_third_party_harness()
-                        || view_model.is_local_to_cloud_handoff();
+                    let use_queued_prompt = view_model.is_third_party_harness();
                     if use_queued_prompt {
-                        // Non-oz runs and local-to-cloud handoff (REMOTE-1486) runs:
```
zachbai: I actually think we should use the queued prompt UI for cloud->cloud handoff for oz, so maybe we should revert this and continue to use queued ui for local -> cloud handoff

app/src/ai/blocklist/controller/shared_session.rs
```diff
@@ -124,6 +124,14 @@ impl BlocklistAIController {
         let existing_conversation_id =
             self.find_existing_conversation_by_server_token(&init_event.conversation_id, ctx);
         let conversation_id = existing_conversation_id
+            .inspect(|conversation_id| {
+                // The local conversation is bound to a cloud-side session, so mark it as a
+                // shared-session view; otherwise `apply_client_actions` won't reconstruct
+                // UserQuery / ActionResult inputs from the cloud agent's response messages.
+                history.update(ctx, |history, _| {
+                    history.set_viewing_shared_session_for_conversation(*conversation_id, true);
```
zachbai: It feels kind of weird that this is done here - shouldn't this be set at the point at which the shared session starts and the cloud mode pane is connected to it?

app/src/ai/blocklist/controller/shared_session.rs
```diff
@@ -249,9 +260,28 @@ impl BlocklistAIController {
         }
         drop(model);
 
-        BlocklistAIHistoryModel::as_ref(ctx)
+        // Only skip the replayed response when we already have a local exchange whose
+        // `server_output_id` matches `request_id`. New exchanges (e.g. the user's first
+        // post-handoff prompt) carry unseen request_ids and must flow through normally.
```
zachbai, line 258: i don't get this

zachbai, line 258: i don't get this

zachbai: when would we actually have a local exchagne whose server output id matches the givne `request_id` that we just got from the server?

app/src/ai/blocklist/block/view_impl.rs
```diff
@@ -159,22 +159,20 @@ fn add_slash_command_highlight(
 /// query blocks during live startup/streaming.
 ///
 /// To avoid duplicate UI, we suppress the AI block header/query only while the viewer is live
-/// (not replaying historical conversation events).
-///
-/// The prompts are rendered in the ambient-agent query block UI, so this helper only gates
-/// duplicate rendering in the AI block path when that optimistic block was actually inserted.
+/// (not replaying historical conversation events) AND the AI block's display query matches an
+/// optimistically rendered user query. The per-query check is important for forked
+/// conversations (e.g. local-to-cloud handoff) where the conversation's first exchange comes
```
zachbai: I don't get this- "the conversations first exchange comes from the source conversation and must remain visible"?

app/src/ai/blocklist/block.rs
```diff
@@ -1205,6 +1205,27 @@ impl AIBlock {
             ctx.subscribe_to_model(&agent_view_controller, |_, _, _, ctx| ctx.notify());
         }
 
+        // Re-render when the cloud agent transitions through setup phases so the response
+        // footer toggles correctly with `is_cloud_agent_pre_first_exchange`. Each event below
+        // toggles that helper's output.
```
zachbai: "toggles that helper's output" means what?

app/src/ai/blocklist/block.rs
```diff
@@ -1205,6 +1205,27 @@ impl AIBlock {
...
+                    AmbientAgentViewModelEvent::DispatchedAgent
+                        | AmbientAgentViewModelEvent::FollowupDispatched
+                        | AmbientAgentViewModelEvent::SessionReady { .. }
+                        | AmbientAgentViewModelEvent::FollowupSessionReady { .. }
+                        | AmbientAgentViewModelEvent::Failed { .. }
+                        | AmbientAgentViewModelEvent::Cancelled
+                        | AmbientAgentViewModelEvent::NeedsGithubAuth
+                        | AmbientAgentViewModelEvent::HarnessCommandStarted
+                ) {
+                    ctx.notify();
```
zachbai: This seems pretty inefficient-  it doesn't seem right to trigger render for every AI block based on status updates at the task level. Can we either unsubscribe after a certain point or listen to a more granular event

app/src/ai/blocklist/block.rs
harryalbert: it means, each of events can change the output of `is_cloud_agent_pre_first_exchange`, so we need to refresh the block on each of these events in case that output has changed.

harryalbert: This is unclear though, I'll update the wording

app/src/ai/blocklist/block.rs
harryalbert: yeah good point — I think unsubscribing makes sense

app/src/ai/blocklist/block/view_impl.rs
harryalbert: rewrote the comment to make it clearer:
```
/// Hide the AI block's header/query when its query is already shown by an optimistic
/// cloud-mode query block (live startup, not replay). The per-query match matters for
/// local-to-cloud handoff: the cloud pane inherits prior local exchanges, but only the new
/// dispatched prompt has an optimistic counterpart — the inherited blocks have none, so
/// the match leaves them visible while still suppressing the duplicate for the new prompt.
```

app/src/ai/blocklist/controller/shared_session.rs
harryalbert: wdym? IIUC `on_shared_init` is the event/function we'd look for to imply that the shared session has started and the cloud mode pane has connected to it

app/src/ai/blocklist/controller/shared_session.rs
harryalbert: will respond to the comment below

app/src/ai/blocklist/controller/shared_session.rs
zachbai: This is specifically for the init event for the MAA response stream received via session sharing-  
  
can't we set this upon forking the local conversation into the cloud mode pane, rather than here

harryalbert: ah yeah, can do

app/src/ai/blocklist/controller/shared_session.rs
harryalbert: this is covering the replay case (like when we have the local forked conversation and the session sharing server replays the events from that same conversation back to us)

app/src/terminal/view/ambient_agent/view_impl.rs
harryalbert: makes sense, can do

zachbai (approved):

PR #9655, opened 2026-04-30
feat(uri): add warposs://pane/{uuid} deep link for pane focus

Akeuuh:
## Description

Adds a `warposs://pane/{uuid_hex}` deep link that focuses a specific terminal pane by its persistent session UUID. This enables external tools to programmatically switch to a given pane/tab/window in Warp.

### What
- **`PaneGroup::find_pane_by_session_uuid`** — new method that iterates terminal panes and matches by persistent UUID bytes, returning the `PaneId`
- **`UriHost::Pane`** — new URI host variant with full handler: decodes 32-char hex UUID, walks all workspaces/windows via `WorkspaceRegistry::all_workspaces()`, dispatches `handle_pane_navigation_event` to focus the tab, and calls `show_window_and_focus_app` to bring the window forward
- **5 unit tests** — host parsing, full URI validation, empty path safety, invalid hex safety, case-insensitive hex decoding

### Why
External tools (e.g. orchestration UIs) that know a pane's UUID need a way to switch focus to that pane without user interaction. The pane UUID is already persisted in SQLite and is stable across sessions.

### How
- Handler parses hex from URL path, converts to `[u8; 16]` via `u8::from_str_radix` (no new crate dependency)
- Cross-window lookup via existing `WorkspaceRegistry::all_workspaces()` + `tab_views()` pattern (matches `focus_terminal_view_in_other_window`)
- `WindowBehaviorHint::Nothing` since handler explicitly manages window focus
- O(n) pane iteration — deep links are rare, no cache needed

## Testing

- [x] Unit tests for URI parsing and validation (5 tests in `uri_test.rs`)
- [x] Manual testing: `open "warposs://pane/{uuid}"` correctly focuses the target pane across tabs and windows

### DIFF
--- a/app/src/pane_group/mod.rs
+++ b/app/src/pane_group/mod.rs
@@ -2134,7 +2134,7 @@ impl PaneGroup {
     /// the given bytes, or `None` if no such pane exists in this group.
     pub fn find_pane_by_session_uuid(&self, uuid: &[u8]) -> Option<PaneId> {
         self.panes_of::<TerminalPane>()
-            .find(|pane| pane.session_uuid() == uuid)
+            .find(|pane| pane.session_uuid() == uuid && !self.is_pane_hidden_for_close(pane.id()))
             .map(|pane| pane.id())
     }
 
--- a/app/src/pane_group/mod.rs
+++ b/app/src/pane_group/mod.rs
@@ -2129,6 +2129,15 @@ impl PaneGroup {
         self.panes_of::<TerminalPane>()
             .any(|pane| pane.terminal_view(ctx).id() == terminal_view_id)
     }
+
+    /// Returns the [`PaneId`] of the terminal pane whose persistent UUID matches
+    /// the given bytes, or `None` if no such pane exists in this group.
+    pub fn find_pane_by_session_uuid(&self, uuid: &[u8]) -> Option<PaneId> {
+        self.panes_of::<TerminalPane>()
+            .find(|pane| pane.session_uuid() == uuid)
+            .map(|pane| pane.id())
+    }
+
     /// Iterate over the code editors in this pane group.
     pub fn code_panes<'a>(
         &'a self,
--- a/app/src/pane_group/mod.rs
+++ b/app/src/pane_group/mod.rs
@@ -2132,7 +2132,7 @@ impl PaneGroup {
 
     /// Returns the [`PaneId`] of the terminal pane whose persistent UUID matches
     /// the given bytes, or `None` if no such pane exists in this group.
-    pub fn find_pane_by_session_uuid(&self, uuid: &[u8]) -> Option<PaneId> {
+    pub fn find_terminal_pane_by_session_uuid(&self, uuid: &[u8]) -> Option<PaneId> {
         self.panes_of::<TerminalPane>()
             .find(|pane| pane.session_uuid() == uuid && !self.is_pane_hidden_for_close(pane.id()))
             .map(|pane| pane.id())
--- a/app/src/uri/mod.rs
+++ b/app/src/uri/mod.rs
@@ -1423,12 +1423,9 @@ fn validate_custom_uri(url: &Url) -> Result<UriHost> {
 /// (e.g. some `file://` URLs on certain platforms); the literal `-` is used
 /// as a placeholder in that case so the formatter never panics.
 fn safe_url_log_fields(url: &Url) -> String {
-    format!(
-        "scheme={} host={} path={}",
-        url.scheme(),
-        url.host_str().unwrap_or("-"),
-        url.path(),
-    )
+    let host = url.host_str().unwrap_or("-");
+    let path = if host == "pane" { "[redacted]" } else { url.path() };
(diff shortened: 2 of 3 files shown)


seemeroland:
Same question and nits as in the original review: https://github.com/warpdotdev/warp/pull/9613#pullrequestreview-4207041648

How do you anticipate users getting the session_uuid to use with this URI? I don't believe we expose this in a nice way. Were you thinking of reading from sqlite? Trying to understand the use case for this

Akeuuh:
> Same question and nits as in the original review: [#9613 (review)](https://github.com/warpdotdev/warp/pull/9613#pullrequestreview-4207041648)
> 
> How do you anticipate users getting the session_uuid to use with this URI? I don't believe we expose this in a nice way. Were you thinking of reading from sqlite? Trying to understand the use case for this

My bad, it's true I did not explain the usecase

I'm using ( and try to contribute to ) an open source project : Open Vibe Island (https://github.com/Octane0411/open-vibe-island), a macOS menubar app that monitors AI coding agents running inside Warp. 
When I click on a running agent session in the Open Vibe Island notch UI, I want to jump directly to that Warp pane — hence the pane/<uuid> deep link.

**How an external app retrieves pane UUIDs:**
Warp persists terminal pane state in a local SQLite database:
~/Library/Application Support/dev.warp.Warp-Stable/warp.sqlite

The terminal_panes table has a uuid column (16-byte BLOB, unique):
SELECT hex(uuid), cwd, is_active FROM terminal_panes WHERE is_active = 1;

The returned hex string is used directly in the deep link:
warp://pane/A1B2C3D4E5F6A1B2C3D4E5F6A1B2C3D4

This is the same approach Warp itself uses for persistence — the UUID is the canonical pane identifier across restarts. An external app just reads the same DB.

No new API surface is needed: the SQLite file is already world-readable on macOS, and the schema is stable (diesel-managed migrations).

Thank you for taking time on this PR ;-) 

@seemeroland

app/src/pane_group/mod.rs
```diff
@@ -2129,6 +2129,15 @@ impl PaneGroup {
         self.panes_of::<TerminalPane>()
             .any(|pane| pane.terminal_view(ctx).id() == terminal_view_id)
     }
+
+    /// Returns the [`PaneId`] of the terminal pane whose persistent UUID matches
+    /// the given bytes, or `None` if no such pane exists in this group.
+    pub fn find_pane_by_session_uuid(&self, uuid: &[u8]) -> Option<PaneId> {
```
seemeroland: nit: `find_terminal_pane_by_session_uuid`

app/src/uri/mod.rs
```diff
@@ -80,6 +81,8 @@ pub enum UriHost {
     Codex,
     /// Actions triggered from Linear integrations (e.g. work on issue).
     Linear,
+    /// Focuses a specific terminal pane by its persistent UUID.
+    Pane,
```
seemeroland: `Session`, assuming we want to find by session UUID and not `PaneId`

app/src/uri/mod.rs
```diff
@@ -1369,12 +1423,24 @@ fn validate_custom_uri(url: &Url) -> Result<UriHost> {
...
 /// as a placeholder in that case so the formatter never panics.
 fn safe_url_log_fields(url: &Url) -> String {
-    format!(
-        "scheme={} host={} path={}",
-        url.scheme(),
-        url.host_str().unwrap_or("-"),
-        url.path(),
-    )
+    let host = url.host_str().unwrap_or("-");
+    let path = if host == "pane" { "[redacted]" } else { url.path() };
```
seemeroland: nit: I know the oz review flagged this, but I don't think the path is sensitive info and it should be fine to log it. I don't think we need to change this and I don't love this special check for pane host

seemeroland (commented):
I see, we probably should expose IDs in another way rather than require querying the sqlite table, but it seems fine to add the URI regardless. The session UUID is specifically for a terminal pane, and is different from the `PaneId` which is usually used to identify a pane generically (there are other types of panes than terminal panes).

Since you need the UUID to be stable across session restarts, the session UUID seems fine. But I would rename to make it clear that this is not a pane ID. Specifically:
- The URI should be /session/{uuid}

Akeuuh:
@seemeroland Thanks for the review! All addressed in 6dd92af6:

- `UriHost::Pane` → `UriHost::Session`, route is now `warposs://session/{uuid}`
- `find_pane_by_session_uuid` → `find_terminal_pane_by_session_uuid`
- Removed the path redaction special-case in `safe_url_log_fields`

Let me know if anything else needs adjusting.

seemeroland (approved):
Thanks for iterating!

Akeuuh:
Thanks a lot @seemeroland for taking the time to review this and for being open to the feature idea! Your suggestion to rename the path to `/session/{uuid}` was the right call — it makes the semantics much clearer, especially given the distinction between `PaneId` and the persistent session UUID. Really appreciated the thoughtful feedback.

PR #9656, opened 2026-04-30
Fix issue on dev with agent history menu not working.

abhishekp106:
## Description

Fixes https://warpdev.slack.com/archives/C08KTPNQN65/p1777573039998779.

The issue was that there were two views subscribing the up arrow event and both would handle it. Both the normal history menu and the cloud mode-only history menu would emit events in rapid succession. We now make sure that only one of the views ever fires off events, since only one history menu should be active at a time.

## Testing

Before, hitting the up arrow in the agent conversation view would cause the buggy behavior Roland described: the first query would get inserted into the buffer and the history menu would appear to never open. Now, it WAI.

## Agent Mode

### DIFF
--- a/app/src/terminal/input.rs
+++ b/app/src/terminal/input.rs
@@ -2631,14 +2631,6 @@ impl Input {
         });
         if FeatureFlag::InlineHistoryMenu.is_enabled() {
             ctx.subscribe_to_view(&inline_history_menu_view, |me, _, event, ctx| {
-                // Only the active view's events should be processed. When
-                // CloudModeInputV2 is enabled and we're composing for an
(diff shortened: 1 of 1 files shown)


abhishekp106:
* **#9656** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9656?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/9656?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

liliwilson (approved):

PR #9669, opened 2026-04-30
fix(bootstrap): fail fast when Node.js / yarn are missing (#9544)

### LINKED ISSUE #9544: Bootstrap scripts don't check for Node.js/yarn, leading to confusing `cargo build` failures

lonexreb:
## Description

Resolves #9544.

`command-signatures-v2`'s `build.rs` invokes `yarn build`, so a fresh checkout without Node/yarn fails partway through `cargo build` rather than at bootstrap time. The existing build-script panic eventually surfaces the right hint, but it's buried under a long cc-rs / build-script trace and there's no canonical \"how to install Node for Warp\" guide for new contributors to follow.

This change makes the failure surface at bootstrap on all three platforms and consolidates install guidance in one place.

## Linked Issue

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. (#9544 is `ready-to-implement`.)

Resolves #9544.

## Changes

| File | Change |
|---|---|
| `script/macos/bootstrap` | After the rust check, fail fast if `node` or `yarn` is missing and point at WARP.md. |
| `script/linux/bootstrap` | Same check after `install_test_deps` / `install_linuxdeploy`. |
| `script/windows/bootstrap.ps1` | Same check using `Get-Command -Type Application` (matches the existing cargo / gcloud probes). |
| `WARP.md` | New \"Node.js setup\" section under \"Platform Setup\" covering version requirement (≥ 18.14.1), recommended install paths (nvm/fnm/volta/asdf/mise or system package manager), yarn-already-on-PATH check, and corepack/Volta caveats. |
| `crates/command-signatures-v2/build.rs` | Trimmed the panic message to point at the new WARP.md section instead of duplicating the install hints inline. |

### Why no auto-install / auto `corepack enable`

The issue author was explicit, and the rationale held up on review:

- `corepack enable` on a system-installed Node typically needs `sudo` / Administrator rights — fragile for a bootstrap script.
- Version-manager users (Volta in particular) already have yarn shimmed and `corepack enable` actively conflicts.
- A pre-existing brew yarn shadowing the corepack one is a known footgun the issue body called out.

Failing fast with a single, well-scoped pointer to WARP.md is the lowest-blast-radius fix.

## Testing

- `bash -n script/macos/bootstrap` and `bash -n script/linux/bootstrap` — both parse cleanly.
- Ran the macOS bootstrap path on a no-node shell locally (`PATH` munged to drop `node`/`yarn`); the new check fires and exits with the WARP.md pointer before any of the subsequent `brew install` / `gcloud` steps run, which is the desired behavior. With both binaries present, the check is a no-op.
- I don't have PowerShell locally, so the `.ps1` path is unrun. CI's Windows lane will exercise it; the syntax mirrors the existing `cargo` / `gcloud` `Get-Command -Type Application` probes immediately above and below it.
- No Rust code logic changed (the `build.rs` edit is a string-only change inside an existing `panic!`).

## Agent Mode

CHANGELOG-IMPROVEMENT: Bootstrap scripts now fail fast with a clear pointer to WARP.md when Node.js / yarn aren't on PATH, instead of letting the failure surface mid-`cargo build`. Thanks @lonexreb!


peicodes (approved):

peicodes:
@lonexreb Sorry for the delay in reviewing this. Once CI passes I'll merge this change.

lonexreb:
Pushed a fix for the Miscellaneous checks failure on the previous commit.

**Root cause:** PSScriptAnalyzer's `PSUseBOMForUnicodeEncodedFile` rule fired on `script/windows/bootstrap.ps1` because the comment I added contained an em-dash (`—`), and the file (per the convention shared by the other four `script/windows/*.ps1` files) ships without a UTF-8 BOM. Severity distribution was `Error = 0, Warning = 1, Information = 15`; the single Warning is what tipped the lint over its zero-warning threshold.

**Fix (commit 84e42ed):** replaced the em-dash with a colon. The file is now pure ASCII (verified with `grep -P '[^\x00-\x7F]'`), so the BOM rule no longer applies. One-character diff, lint-only, no behavior change.

CI on the new commit is currently `action_required` — needs a maintainer approve-and-run when convenient. Thanks!

peicodes:
Closing as there proof of manual testing as per our contribution guidelines

PR #9632, opened 2026-04-30
Fix schema generator binary recompilation during release bundling.

vorporeal:
## Description
The release bundling workflow runs `prepare_bundled_resources` after the main `cargo build`, and that script in turn invokes `cargo run --bin generate_settings_schema`. Until now we passed neither `--features` nor `--target` to this run, so cargo's resolver-v2 saw a different per-package feature unification than the main build and recompiled every dependency whose enabled features differed (`rust-embed` with `debug-embed`, `warp_core`'s `release_bundle`, sentry, jemalloc, etc.). This added significant time to release builds and pulled in extra system dependencies (e.g. protoc) for jobs that should otherwise just be packaging the prebuilt binary.

There was a secondary correctness issue: the schema generator iterates settings registered via `inventory::submit!`, and `#[cfg(feature = \"...\")]`-gated settings were silently omitted from the schema in any context whose feature set didn't match the main binary.

The fix threads the build's feature set (and target, where applicable) through to the schema-generator invocation, and makes package selection explicit with `-p warp`. With matching features and target, cargo reuses the existing compilation artifacts; with the same feature set, the generated schema is also faithful to what the bundled binary actually exposes.

## Testing
Verified `bash -n` syntax for the modified bash scripts and parsed the modified PowerShell files via `pwsh`. Bundling, schema generation, and CI behavior will be exercised by this PR's release builds.

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

[Conversation](https://staging.warp.dev/conversation/1bd5574f-df28-4c2e-867d-40d71ed7e827)

PR #9676, opened 2026-04-30
Revert "Fix schema generator binary recompilation during release bundling."

kevinchevalier:
Reverts warpdotdev/warp#9632

Will look at in the future. Currently causing dev build issues like https://github.com/warpdotdev/warp-internal/actions/runs/25187609052/job/73853657131


lucieleblanc (approved):
Stamp

PR #9680, opened 2026-04-30
Orchestration pill bar updates: same-pane pills, 3-dot menu, hover card, breadcrumbs

advait-m:
Demo Loom: https://www.loom.com/share/69d68745f1384c0e8c0a96602f8c1a12
Focus panes demo: https://www.loom.com/share/2c07193874df49ce868c6ca04ee5a990
Fixes https://linear.app/warpdotdev/issue/QUALITY-590/pill-bar-updates-v2

Removes old UI i.e. rows for each agent pinned at bottom of convo (we remove only if pills FF is on) and adds the pills FF to dogfood flags!

## Description

Builds out V2 of the orchestration pill bar in the agent view header — same-pane pills with avatars + click-to-switch, a 3-dot overflow menu, hover details card, split-off-only breadcrumbs that route back to the orchestrator's existing pane/tab, and an architectural "single source of truth" for which terminal view owns each child agent conversation.

Highlights:
- **Same-pane pills**: orchestrator + each child agent rendered as a pill row above the agent view. Clicking an unselected pill switches the active conversation in place via `SwitchAgentViewToConversation`. When the conversation is already open in another visible pane (this tab, another tab, or another window), the click instead focuses that pane via the same `FocusOpenedConversation` action the menu uses.
- **3-dot overflow menu**: per-child pill, anchors to the clicked pill's position. Items are context-dependent:
  - When the conversation is *not* open elsewhere: `Open in new pane`, `Open in new tab`.
  - When the conversation *is* open in another visible pane: a single `Focus pane` item that focuses the existing pane.
  - `Stop agent` / `Kill agent` items are intentionally hidden for now (the action wiring stays so they can be re-added once the underlying behaviour is reliable).
- **"Open in new tab"**: actually creates a new tab via `Event::OpenChildAgentInNewTab` plumbed up from `TerminalView` → `pane_group::Event` → `Workspace`, then enters the agent view in the fresh tab.
- **"Open in new pane"**: splits a fresh terminal pane to the right and goes through the cloud load+restore path so the new pane shows the full transcript (instead of revealing the empty hidden child pane created at agent start).
- **Hover details card**: 280-wide card overlay anchored to the hovered pill (300ms hover-in delay). Shows avatar + name, status badge (hidden for orchestrator), cwd, description, harness chip, and branch/PR chips from any `Artifact::PullRequest`.
- **Breadcrumbs (split-off panes only)**: `[Parent Avatar] Title › [Child Avatar] Name`, wrapped in a horizontal `NewScrollable` with overlaid scrollbar so a narrow pane can pan and the row stays vertically centered.
- **Focus path picker (pill clicks, menu's "Focus pane", and the breadcrumb parent crumb)** all share the same logic: resolve the canonical owner from `BlocklistAIHistoryModel`, then dispatch `TerminalAction::RevealChildAgent` when the owner pane is in the same pane group (sibling pane in this tab) or `WorkspaceAction::FocusTerminalViewInWorkspace` when it's in a different pane group (other tab / window). Going through the workspace's `focus_pane` from a foreign `ViewContext` doesn't reliably move focus to siblings, hence the split.
- **Single source of truth for child conversations**: a new `BlocklistAIHistoryEvent::ConversationOwnershipTransferred` event is emitted by `set_active_conversation_id` whenever a conversation moves from one terminal view's live list to another. `TerminalView::handle_ai_history_model_event` listens for it and drops stale `AIBlock` / `AgentViewEntry` rich content, so the previous owner's pane goes blank and the new owner is the only place that renders the transcript. Closing a split-off pane transfers its child conversations back to the parent's owning view.
- **Cosmetic fixes along the way**: clip the pill row to pane bounds (no bleeding into adjacent panes), stable label width independent of hover (siblings don't shift when the dots fade in), only highlight the truly selected pill, vertical 3-dot glyph rendered as a positioned overlay on hover.

Gated by `FeatureFlag::OrchestrationPillBar`.

## Testing

Tested locally on macOS:
- Same-pane pill click switches the active child in place; selected pill is the only highlighted one.
- 3-dot menu opens anchored to the right pill, items dispatch their respective actions, ESC / click-outside closes it cleanly. The menu collapses to a single `Focus pane` item when the conversation is already open in another visible pane.
- Pill click and menu's "Focus pane" both route through the same handler, so they behave identically. Verified focus shift works for sibling panes in the same tab, panes in another tab, and panes in another window.
- "Open in new pane" splits right and hosts the child with a populated transcript; "Open in new tab" creates a new session tab and enters the child agent view there. The orchestrator pane reverts to the parent conversation when its currently-active child gets split off.
- Closing a split-off pane transfers any child conversations back to the orchestrator pane; the orchestrator's pill bar continues to function and can re-open the children.
- Hover card appears after the 300ms delay, repositions when scrubbing across pills, and dismisses with ~80ms hover-out delay.
- Breadcrumbs only render in split-off panes; parent crumb click focuses the existing orchestrator pane (cross-tab + cross-window + same-tab sibling pane verified) and falls back to in-place switch when the orchestrator isn't open anywhere.

`cargo check -p warp`, `cargo fmt`, and `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings` all pass.


app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
cephalonaut, line 460: Agent mode suggests you might also want to prune overflow_button_mouse_states

app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
```diff
@@ -343,38 +778,532 @@ impl View for OrchestrationPillBar {
...
+        .with_corner_radius(CornerRadius::with_all(Radius::Pixels(4.)))
+        .finish()
+}
+
+/// Replaces a `$HOME` prefix in a path with `~` so the working-directory
+/// line in the hover card matches user expectations (`~/warp-internal`
+/// rather than the full absolute path). Falls back to the original path
+/// when `$HOME` is unset or doesn't prefix the path.
+fn shorten_home_path(path: &str) -> String {
+    if let Some(home) = std::env::var_os("HOME") {
```
cephalonaut: Maybe not cross-platform with windows? Conceivably some utility exists?

cephalonaut (approved):
Nothing blocking, looking good!

app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
advait-m: ah yep, thanks - fixed

app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
advait-m: ah good call, let me use `user_friendly_path`

PR #9705, opened 2026-05-01
feat: per-pane monospace font size

### LINKED ISSUE #6739: Allow for per window font size
### Pre-submit Checks

Adjusting font size via Cmd + or Cmd - (or through the settings UI) affects all Warp windows. It would be nice to be able just cmd - to shrink a single window to fit more text while keeping the other windows the same. 

We will often break up the terminal into 4 windows two on top and two on bottom so we can monitor multiple parts of an application at once. Some output is more verbose than others. So we want to shrink the top left windows text to fit the formatted print out, but the other windows information is sparse and easier to digest with larger text.  

### Additional context

_No response_

### Operating system (OS)

macOS

3

None

wpoulin:
## Description

  Adds independent monospace font-size control for each terminal pane (issue #6739). Previously the font-size shortcuts adjusted every Warp window globally; now each
  pane can have its own size, so users can shrink the text in one pane to see more output while keeping a sibling pane large for readability.

  **Ships as an opt-in Experimental feature** behind a "Pane-specific font size" toggle on the Features settings page (with a BETA chip), gated by
  `FeatureFlag::PerPaneFontSize`:

  - *Dogfood (Local/Dev):* feature force-enabled; toggle hidden.
  - *Preview/Stable:* toggle shown; default off. Users opt in.

  `PaneSettings::is_pane_specific_font_size_enabled()` returns `flag.is_enabled() || *self.pane_specific_font_size`. Promoting through `PREVIEW_FLAGS` → `RELEASE_FLAGS`
  automatically hides the toggle and force-enables the feature — no code change required.

  ## User-visible behavior

  When the toggle is **off** (the Preview/Stable default), font-size shortcuts continue to write the global `MonospaceFontSize` — existing behavior, unchanged.

  When the toggle is **on**, the shortcuts depend on whether the `UIZoom` flag is enabled (`UIZoom` repurposes `Cmd+=` / `Cmd+-` for window zoom):

  - *Without `UIZoom`:* `Cmd+=` / `Cmd+-` adjust the **focused pane's** font size (previously global). `Cmd+0` resets the focused pane's override back to the global
  default.
  - *With `UIZoom`:* `Ctrl+Shift+>` / `Ctrl+Shift+<` adjust the focused pane's font size (the `=` / `-` keys are taken by window zoom).

  The terminal grid, prompt, alt-screen, and the input editor where the user types all pick up the pane's effective font size — no glyph/cell mismatch and no leftover
  full-size input box. UI chrome (banners, menus, AI mode indicators) and **agent text** intentionally stay on the global for v1; see [Known 
  limitations](#known-limitations-out-of-scope-for-v1-follow-up-issues-to-come).

  The Settings UI's monospace font size slider continues to write the global; panes without an override continue to follow it.                                                                                                                
   
 ## Implementation

  - `FeatureFlag::PerPaneFontSize` added to the enum and to `DOGFOOD_FLAGS`. `PaneSettings::is_pane_specific_font_size_enabled()` is the composite check used everywhere.
  - The opt-in widget lives on the Features page with a BETA chip. `should_render` is gated on `!FeatureFlag::PerPaneFontSize.is_enabled()` so the toggle auto-hides on
  channels where the feature is force-enabled.
  - Per-pane override is stored as `HashMap<PaneId, f32>` on `PaneGroupFocusState`; emits a `FontSizeOverrideChanged { pane_id }` event with
  `PaneFocusHandle::is_affected` filtering to the matching pane.
  - The render path is threaded with the pane's effective size — `SizeUpdateBuilder::build`, `BlockListElement`, `BlockGridElement`, `AltScreenElement`, and the prompt
  render helpers — so cell metrics and glyph painting always change together.
  - `EditorView` already had `set_font_size`; added `clear_font_size` and made `set_font_size` idempotent. `Input` syncs the editor's override on every
  `FontSizeOverrideChanged` and once at focus-handle wiring time so a freshly split pane that inherits an override picks it up before its first render.
  - Override entries are dropped on permanent pane removal — `close_pane` (non-undo path), `cleanup_closed_pane` (undo grace expiry), `remove_pane_for_move` (pane leaves
   this group). The `UndoClosedPanes` restore path preserves the override (only forgets on permanent removal).
  - Override is in-memory only by design: `PaneId`s aren't stable across restarts, so persisting wouldn't be meaningful.
                                                                                                                              
  ## Linked Issue
  <!--                                                                                                                      
  Link the GitHub issue this PR addresses. Before opening this PR, please confirm:
  -->
  Closes #6739.                                                                                                             
  - [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. *(labeled `Ready-to-Implement`)*
  - [x] Where appropriate, screenshots or a short video of the implementation are included below (especially for            
  user-visible or UI changes).                                                                                              
                                                                                                                            
  ## Screenshots / Videos                                                                                                   
  https://github.com/user-attachments/assets/b4df90f6-be78-40b0-b75b-46ce70f65205
                                                                                                                            
  _Split the window into two panes, then increase the font size in one pane and decrease it in the other — each pane scales independently. Resetting returns both panes to the default size._                                                         
                                                                                                                            
  ## Testing

  **Unit tests added** (8 total, all passing):

  - Cover override behavior, precedence over global, cleanup, no-ops, event targeting, and editor integration.

  **Manual end-to-end testing** in a local debug build (verified on both `UIZoom` on and off):

  - Verified independent resizing across panes
  - Reset behavior (`Cmd+0`) works in both modes
  - Global font changes don't affect overridden panes
  - Works correctly in alt-screen apps (`vim`, `htop`)
  - Overrides persist through undo-close, but not permanent close

  **Presubmit:** `./script/presubmit` — `cargo fmt`, `cargo clippy --workspace --all-targets --tests -- -D warnings`, `clang-format`, and `wgslfmt --check` all clean.

  ## Known limitations (out of scope for v1, follow-up issues to come)

  - **Agent text (block contents, plan/todo lists) does not yet pick up the per-pane override.** Only terminal text and the embedded input editor are wired through for
  v1. This is the primary reason the feature ships Experimental; extending the override into the agent-text render path is a larger change that warrants its own PR.
  - Per-pane override is dropped when a pane is dragged to another window — the destination group starts at the global default. Preserving across moves would need
  override state to live on pane content metadata.                                                               
                  
## Agent Mode                                                                                                             
                                                                                                                            
 <!-- CHANGELOG-NEW-FEATURE: An experimental "Pane-specific font size" toggle is now available under Features → Settings (with a BETA chip). When enabled, the font-size
   shortcuts (or `Ctrl+Shift+>` / `Ctrl+Shift+<` when the `UIZoom` flag is on) grow or shrink only the focused pane, and `Cmd+0` resets it. -->


wpoulin:
@cla-bot check

app/src/editor/view/mod.rs
```diff
@@ -3319,10 +3319,21 @@ impl EditorView {
     }
 
     pub fn set_font_size(&mut self, new_font_size: f32, ctx: &mut ViewContext<Self>) {
+        if self.text_options.font_size_override == Some(new_font_size) {
```
vkodithala, line 3318: Nice improvement!

app/src/editor/view/mod_test.rs
```diff
@@ -4520,6 +4520,45 @@ fn test_drag_and_drop_files_applies_path_transformer() {
...
+            );
+
+            editor.set_font_size(18.0, ctx);
+            assert_eq!(editor.text_options.font_size_override, Some(18.0));
+
+            // Re-applying the same value is a no-op for the override; we
+            // can't directly observe `ctx.notify()`, but the early-return
+            // guards an unnecessary repaint.
+            editor.set_font_size(18.0, ctx);
+            assert_eq!(editor.text_options.font_size_override, Some(18.0));
```
vkodithala: nit: The agent seems to reference this, but this doesn't actually test anything functional. If we want to add a test, we should primarily just be checking if the editor's font size can be correctly overriden and cleared. Can be much less verbose.

crates/warpui/src/rendering/wgpu/shaders/glyph_shader.wgsl
```diff
@@ -25,8 +25,7 @@ struct Uniforms {
     viewport_size: vec2<f32>,
     // Padding necessary to ensure that the uniforms is 16 bytes. Some wgpu-supported devices (such as webgl) require
     // buffer bindings to be a multiple of 16 bytes.
-    padding: vec2<f32>
```
vkodithala, line 28: nit: These look like unrelated WSGL formatting diffs, can we remove?

vkodithala (approved):
Hey @wpoulin - thanks so much for the contribution. This code mostly looks good and functionally works, would like the following changes before merge:
* we should gate this feature behind a user-configurable setting, like "Pane-specific font size", that's default False

vkodithala (changes requested):
Will approve after the changes requested in my previous comment go in!

crates/warpui/src/rendering/wgpu/shaders/glyph_shader.wgsl
wpoulin: Good catch ! must've been a formatter version mismatch with CI. I'll revert those.

app/src/editor/view/mod_test.rs
wpoulin: I simplified the test to just cover the override/clear behavior

wpoulin:
Applied the changes. Also added a telemetry event matching the existing pattern for the other Settings → Appearance → Panes options. Ready for re-review

wpoulin:
Rebased on master and resolved conflicts, then force push. No behavior changes from the rebase. I've also re-tested a few things (including the feature) on my end to make sure everything's working as expected

crates/warpui/src/rendering/wgpu/shaders/glyph_shader.wgsl
```diff
@@ -88,7 +89,7 @@ fn vs_main(
         fade_alpha = 1. - fade_dist / fade_width;
     }
 
-    // Convert the position of the item from screen coordinates into normalized device coordinates
+     // Convert the position of the item from screen coordinates into normalized device coordinates
     var device_pos: vec2<f32> = pixel_pos / uniforms.viewport_size * vec2(2.0, -2.0) + vec2(-1.0, 1.0);
 
     var texture_coordinate: vec2<f32> = glyph.uv_bounds.xy + glyph.vertex_position * glyph.uv_bounds.zw;
```
vkodithala, line 95: These look like no-op diffs, can we remove these?

vkodithala (changes requested):
Thanks for gating this behind a ff @wpoulin! Was testing this out and realized that our current increase/decrease font size keybindings apply to _both_ terminal and agent text. It looks like your fix only changes the font size for terminal text (agent text size remains unchanged with `Ctrl-Shift-(</>)`. Is there a reason for this? We should ideally maintain this feature for both text types.



Also added a minor comment about some stray changes you should remove.

crates/warpui/src/rendering/wgpu/shaders/glyph_shader.wgsl
wpoulin: ahh damn.. these keep sneaking in whenever I'm allowing the AI to do edits in nearby code.. it loves to "tidy up" whitespace in the wgsl shaders even when the feature has nothing to do with rendering. reverted.

wpoulin:
You're right that it should !! Initially, I wanted to keep the scope small, but your right.. it would have been a half-baked feature.

I've wired the per-pane override through the main agent surfaces (block content,
orchestration UI, transcript entries, ambient agent screens, input indicators). Also caught a bug along the way: the pill bar was constructed before the focus handle existed and was permanently stuck on the global size. Stray WGSL changes are gone too.

One gap worth flagging: the plan/todo list rendering still uses the global, so it'll visibly mismatch the surrounding agent text on multi-step runs (which is most of them). Same with inline action cards, the agent input footer, and a couple of CLI agent bits.  happy to land here or as a follow-up @vkodithala

vkodithala (changes requested):
Hey @wpoulin, thanks for the changes and apologies for the delay. Mind fixing merge conflicts and requesting a re-review?

vkodithala (commented):
Thanks for the fixes @wpoulin, added myself as a re-reviewer and reviewing these changes now. For future reference, when you've resolved merge conflicts/responded to a "Requested changes" review, please re-request a review so that it's put back in my inbox!

app/src/workspace/view.rs
```diff
@@ -16811,18 +16828,23 @@ impl Workspace {
...
-                .set_value(new_font_size, ctx));
-        });
+        if *PaneSettings::as_ref(ctx).pane_specific_font_size {
+            focus_state.update(ctx, |state, ctx| {
+                state.set_font_size_override(pane_id, new_size, ctx);
+            });
+        } else {
+            FontSettings::handle(ctx).update(ctx, |font_settings, ctx| {
+                report_if_error!(font_settings.monospace_font_size.set_value(new_size, ctx));
+            });
```
vkodithala: Is there a reason we got rid of the `set_terminal_font_size` helper here? Would help get rid of some code duplication between `adjust_terminal_font_size` and `reset_font_size`.

app/src/workspace/view.rs
```diff
@@ -16786,6 +16799,10 @@ impl Workspace {
                 .zoom_level
                 .set_value(ZoomLevel::default_value(), ctx));
         });
+        // `Cmd+0` is bound to `ResetZoom` when the `UIZoom` flag is on, so
+        // also clear the focused pane's font-size override here — users
+        // expect the same key to reset both.
+        self.reset_font_size(ctx);
```
vkodithala: This seems like a behavior change. Tested things out locally, and in Warp currently, resetting UIZoom doesn't reset font size. Unless there's a functional reason for changing this, can we revert? (and maybe make another issue to address as needed)

wpoulin:
yeah mb.. noted ! Head's up; I'm investigation/fixing CI issues as I'm writing this message

vkodithala (changes requested):
Spent some time here, think you correctly wired the per-pane font size override through _some_ surfaces, but still seeing issues with rendering agent text, i.e. below ("Ask a follow up" is correctly small, but the core agent text remains the same size):
Screenshot 2026-05-26 at 5 43 27 PM

Left a few other drive-by comments, too.

Saw your note re: plan/todo list rendering, I'd like us to be consistent here with what we support for a v1. I think what we're running into is trying to cover all surfaces at once, which can get tricky and lead to a super large PR.

Wdyt about initially supporting this for a smaller subset of surfaces, i.e. just for terminal text, and marking this feature as "Experimental" while we continue to work on it? Some prior art on an "Experimental" toggle here: https://github.com/warpdotdev/warp/pull/11555. I can get this merged in ASAP, since I'm fairly confident this works well, and then we can open a second issue to extend this to other surfaces. Please re-request a review once responded!

wpoulin:
Totally on board with marking this "Experimental" and scoping v1 down — thanks for the #11555 pointer, I wasn't aware that pattern existed here.

I've reworked the PR so the per-pane override applies to terminal text only, gated behind FeatureFlag::PerPaneFontSize (force-on for Local/Dev; everyone else opts in via the BETA toggle on the Features page, which auto-hides once the flag promotes). I reverted all the agent-surface wiring (agent-view blocks, AI blocks, orchestration pill bar, ambient-agent surfaces, code blocks) back to the global font size — so the core-agent-text rendering 
issue you flagged is out of scope here, and the diff is much smaller. I also fixed the Reset font-size shortcut (Ctrl+Shift+0) so increase/decrease/reset all work.
  
For the follow-up: the remaining surfaces to cover before this can graduate from Experimental are the agent/AI ones I reverted — agent-view text (the core-agent-text issue), AI blocks + code blocks, the orchestration pill  bar, the zero-state block, ambient-agent surfaces, and plan/todo rendering. I was going to open a tracking issue for that, but the current issue templates are user-facing (Feature Request / Bug Report) and don't really fit an internal eng follow-up — where would you prefer I track it (Linear, a blank issue, etc.)?

Re-requesting review now — thanks for the thorough pass!

vkodithala (changes requested):
Hey @wpoulin, thanks for fixing this up. I'm gonna defer further discussion and mark REQUEST_CHANGES until I get internal alignment that we want to support this, even as experimental. Apologies for the back-and-forth, I should've clarified this earlier. Will get back to you by next week.

vkodithala (commented):
Hey @wpoulin, thanks for the patience and apologies for all the back and forth here. I just synced with our design team, we're not sure if we want to support a per-pane monospace font size and want to see if the original issue gathers interest. 

I've marked issue 6379 as `warp:tracking` for now and will continue to monitor. Closing this PR for now. Sorry about that!

PR #9706, opened 2026-05-01
Fix: validate pacman signing key before allowing update

Naveen-Boddepalli:
## Summary
This PR improves the pacman update flow by validating the signing key before proceeding with the update.

Currently, Warp checks for the presence and expiration of the pacman signing key, but does not verify whether the key is valid at the trust level (e.g., revoked, expired, or otherwise unusable). This can lead to update failures during package installation due to signature verification errors.

## Changes
- Added parsing of the `pub:` line from GPG output to extract the validity field
- Ensured only valid keys (`f` = full, `u` = ultimate) are accepted
- Return false for invalid states (`e`, `r`, `-`, `q`) to trigger key reconfiguration
- Preserved existing expiry checks and overall logic flow

## Impact
Prevents update attempts with invalid or misconfigured pacman signing keys, improving reliability and user experience on pacman-based systems.

Happy to make further changes if needed.

## Linked Issue
N/A

## Screenshots / Videos
N/A

## Testing
- Verified correctness of GPG output parsing logic
- Ensured compatibility with existing expiry validation
- Confirmed no changes to external interfaces or update flow beyond validation

No new tests were added as this is a small internal validation improvement.

## Agent Mode

### DIFF
--- a/app/src/autoupdate/linux.rs
+++ b/app/src/autoupdate/linux.rs
@@ -626,6 +626,20 @@ fn is_pacman_signing_key_installed() -> bool {
         return false;
     };
 
+    // After parsing the pub: line, also check validity field (index 1 = validity)
+    let fields: Vec<&str> = stdout
+        .lines()
+        .find(|line| line.starts_with("pub:"))
+        .map(|line| line.split(':').collect())
+        .unwrap_or_default();
+
+    // Field index 1 = validity: 'f' (full), 'u' (ultimate) are valid;
+    // 'e' (expired), 'r' (revoked), '-', 'q' = invalid
+    let validity = fields.get(1).copied().unwrap_or("");
+    if !matches!(validity, "f" | "u") {
+        return false; // Force key reconfiguration
+    }
+
     // Parse the expiry timestamp from the pub: line (field 7, 1-indexed).
     let Some(expiry_field) = stdout
         .lines()
--- a/app/src/autoupdate/linux.rs
+++ b/app/src/autoupdate/linux.rs
@@ -642,6 +642,9 @@ fn is_pacman_signing_key_installed() -> bool {
     if !matches!(validity, 'f' | 'u') {
         return false; // Force key reconfiguration
     }
+    if !matches!(validity, 'f' | 'u') {
+        return false; // Force key reconfiguration
+    }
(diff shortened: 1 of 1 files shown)


Naveen-Boddepalli:
Hi! This PR adds validation for pacman signing key validity to prevent update failures due to invalid keys. Happy to make any changes if needed.

Naveen-Boddepalli:
recheck

vorporeal:
hey @Naveen-Boddepalli, was there a particular issue you or another user ran into that this change is fixing?

Naveen-Boddepalli:
Hi! I didn’t encounter a specific user-facing failure personally, but while looking at the pacman key validation logic I noticed that we only check for key presence and expiration, and not the GPG validity state.

In cases where a key is revoked, expired, or not trusted, GPG marks it as invalid (e.g., e, r, q, etc.), but the current logic could still treat it as usable if the expiration check passes. This could allow pacman to proceed with a key that isn’t actually valid, potentially causing failures during package verification. This change ensures we catch those cases earlier and trigger key reconfiguration before attempting the update, aligning better with GPG’s trust model.

Happy to adjust the approach if you think this should be handled differently.

app/src/autoupdate/linux.rs
```diff
@@ -626,6 +626,26 @@ fn is_pacman_signing_key_installed() -> bool {
...
+    let validity = fields
+        .get(1)
+        .and_then(|field| field.chars().next())
+        .unwrap_or('\0');
+    if !matches!(validity, 'f' | 'u') {
+        return false; // Force key reconfiguration
+    }
+    if !matches!(validity, 'f' | 'u') {
+        return false; // Force key reconfiguration
+    }
```
vorporeal: can you remove the duplicated check?  (the suggested diffs sometimes don't target the correct range, leading to issues like this)

vorporeal (approved):
that makes sense!  just wanted to understand the source of the PR.  thanks for the contribution!

app/src/autoupdate/linux.rs
Naveen-Boddepalli: Removed the duplicated validity check — thanks for catching that.

Issue #9041, opened 2026-04-10
[BUG] CopyOnSelect=false setting ignored — drag selection still copies to clipboard


catallactics:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there

### Describe the bug

"Copy on Select" is disabled in Settings → Features → Terminal, and the plist confirms `CopyOnSelect = false`:

```
$ defaults read dev.warp.Warp-Stable CopyOnSelect
false
```

Despite this, **dragging to select text still copies it to the clipboard.** The setting is saved correctly to disk but Warp ignores it at runtime.

This was working correctly (not copying on select when disabled) in a previous version. The regression appeared sometime in recent updates — exact version unknown as Warp auto-updates.

### To Reproduce

1. Open Warp Settings → Features → Terminal
2. Disable "Copy on Select" (toggle off)
3. Verify via terminal: `defaults read dev.warp.Warp-Stable CopyOnSelect` → `false`
4. Copy some unrelated text to clipboard (e.g. "hello" via Cmd+C from another app)
5. In Warp, drag to select any text
6. Paste (Cmd+V) in another app
7. **Result:** The dragged-selected text is pasted, NOT "hello"
8. **Expected:** "hello" should be pasted — selection should NOT auto-copy when CopyOnSelect is disabled

### Environment

- **Warp**: v0.2026.04.08.08.36.02
- **macOS**: 26.3.1 (Apple Silicon, arm64)
- **Shell**: zsh

### Related issues

- #4580 — Copy on Select broken after alt-screen change (different symptom: enabled but misbehaving)
- #7223 — Copy on Select double-copy (reportedly fixed)

This is a distinct issue: the setting is **disabled** but the feature still activates.

MatthewJamisonJS:
Peace and Long Life 🖖🏿 
@captainsafia & @catallactics 

I submitted an attempt via PR #9708 
Whenever it's most convenient for you to take a look,
I'll be looking forward to learning more from with you all 🙂

catallactics:
Thanks for taking this on, @MatthewJamisonJS 🖖 — the diagnosis lines up with what I was seeing. Hoping it lands.

PR #9708, opened 2026-05-01
fix(user_defaults): return canonical JSON for NSNumber(BOOL) values (#9041)

MatthewJamisonJS:
## Description

Refs #9041.

`CopyOnSelect=false` set via `defaults write -bool false` is silently ignored. Root cause: `UserDefaultsPreferencesStorage::read_value` uses `stringForKey:`, which coerces an `NSNumber(BOOL)` to `"0"` or `"1"`. `serde_json::from_str::<bool>("0")` returns `Err`; the setting falls back to its default `true`; drag-to-select copies regardless of the user's preference.

Fix: call `objectForKey:` first. If the stored value is an `NSNumber` with a boolean ObjC type encoding (`c` on x86_64, `B` on arm64), return `"true"` or `"false"` directly. All other value types (NSString, non-boolean NSNumber) fall through to the original `stringForKey:` path — no behavior change for values written without `-bool`.

A second conditional path where post-migration `defaults write` edits are stranded when `FeatureFlag::SettingsFile` is enabled was also identified during the audit of this issue. That scenario has design tradeoffs and is tracked separately.

## Linked Issue

## Screenshots / Videos

Not applicable — no user-visible UI change; the fix restores an existing preference to having effect.

## Testing

- `cargo clippy -p warpui_extras --all-targets -- -D warnings` clean
- Manual: set `defaults write dev.warp.Warp-Stable CopyOnSelect -bool false`, launch Warp, drag-select text — clipboard should not be written. Toggle back with `-bool true`, verify drag-select copies again.

## Agent Mode
- [ ] Warp Agent Mode

CHANGELOG-BUG-FIX: Fix CopyOnSelect=false being ignored when the preference was set via `defaults write -bool false`

### DIFF
--- a/crates/warpui_extras/src/user_preferences/user_defaults.rs
+++ b/crates/warpui_extras/src/user_preferences/user_defaults.rs
@@ -3,6 +3,8 @@
 #![allow(deprecated)]
 
 use cocoa::base::{id, nil};
+use core_foundation::base::{CFGetTypeID, CFTypeRef, TCFType};
+use core_foundation::boolean::CFBoolean;
 use objc::{class, msg_send, rc::StrongPtr, sel, sel_impl};
 
 /// A user preferences store backed by macOS user defaults (`NSUserDefaults`).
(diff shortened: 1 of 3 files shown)


MatthewJamisonJS:
@cla-bot check

crates/warpui_extras/src/user_preferences/user_defaults.rs
```diff
@@ -61,6 +61,25 @@ impl super::UserPreferences for UserDefaultsPreferencesStorage {
...
+            // stringForKey: coerces that to "0"/"1", which serde_json cannot parse as bool
+            // and silently falls back to the setting's default. Return canonical JSON instead.
+            let raw: id = msg_send![*self.user_defaults, objectForKey: *key];
+            if raw != nil {
+                let is_number: bool = msg_send![raw, isKindOfClass: class!(NSNumber)];
+                if is_number {
+                    let obj_c_type: *const std::os::raw::c_char = msg_send![raw, objCType];
+                    let encoding = std::ffi::CStr::from_ptr(obj_c_type).to_bytes();
+                    // 'c' = signed char (BOOL on x86_64); 'B' = C++ bool (BOOL on arm64).
+                    if encoding == b"c" || encoding == b"B" {
```
grawish: This check may also match non-boolean `NSNumber` values: Objective-C type encoding `c` is signed char, and `NSNumber` can use that encoding for numeric char values as well as BOOL on some runtimes. If a future/default key stores a char-valued number, this path would coerce it to JSON `true`/`false` instead of preserving the existing `stringForKey:` behavior. Consider checking CoreFoundation boolean identity instead (for example `CFGetTypeID(raw as _) == CFBooleanGetTypeID()`) before returning canonical JSON booleans.

grawish (commented):
Reviewed with a focus on correctness and regression risk.

crates/warpui_extras/src/user_preferences/user_defaults.rs
MatthewJamisonJS: Thank you for that insight!
Truly appreciate you taking time to point me in the right direction 🙂🖖🏿

captainsafia:
*sigh* My sweet agent is having some trouble. Let's try this again.

/oz-review

MatthewJamisonJS:
@captainsafia 
No worries 😊
🖖🏿

PR #9711, opened 2026-05-01
fix(macos/ime): skip key-equivalent priority path while IME is composing (#9709)

### LINKED ISSUE #9709: macOS IME (超注音 / Yahoo Bopomofo): arrow key advances candidate selection by 2 instead of 1
### Pre-submit Checks

- [x] I have searched Warp bugs and there are no exact duplicates (related issues linked below)

### Describe the bug

On macOS, when using **超注音 (Yahoo Bopomofo / 極限注音)** — a widely used third-party Traditional Chinese input method in Taiwan — pressing the **right** or **left** arrow key during candidate selection advances the highlighted candidate by **2 positions** instead of 1. This makes it impossible to land on even-indexed candidates (#2, #4, #6, …) via arrow keys.

The bug does **not** occur with macOS built-in 注音 (Apple Bopomofo) or Apple Pinyin. It does **not** reproduce in Terminal.app, iTerm2, Ghostty, or VS Code's integrated terminal — only in Warp.

### To reproduce

1. macOS, install 超注音 (Yahoo Bopomofo) — free third-party IM, https://tw.yahoo.com/超注音
2. In Warp, switch input source to 超注音
3. Type a Bopomofo combination with many candidates, e.g. `ㄅㄨˋ` (yields 不 / 把 / 被 / 補 / 版 / 堡 / …)
4. Candidate window appears with `1不` highlighted
5. Press right arrow once
6. **Observed:** highlight jumps from `1不` to `3被` (skips `2把`)
7. Continue pressing right: `3被` → `5版` → `7表` → … (always advances by 2)

Number-key selection (`1`–`9`) works correctly as a workaround, but arrow-key navigation is broken.

### Expected behavior

Right arrow advances candidate selection by 1 position, matching behavior in Terminal.app, iTerm2, Ghostty, and every other macOS app.

Screen recording showing the candidate index jumping by 2 on each right-arrow press. Sequence: `1不` → `3被` (skipped `2把`) → `5版` (skipped `4補`).

Image

Candidate row layout: `1不 2把 3被 4補 5版 6堡 7表 8ㄅ 9並 …`

### Operating system (OS)

macOS

macOS 26.4.1 (25E253)

### Shell Version

zsh 5.9

### Current Warp version

0.2026.04.27.15.32.03

### Regression

No, this bug has existed throughout my experience using Warp.

### Recent working Warp date

N/A

#### Suspected root cause

The bug appears to originate in `keyDownImpl` in `crates/warpui/src/platform/mac/objc/host_view.m`, where every keyDown event is dispatched through **two parallel paths**:

```objc
- (BOOL)keyDownImpl:(NSEvent *)event {
    BOOL wasComposing = [self hasMarkedText];
    [textToInsert setString:@""];

    interpretingKeyEvents = YES;
    [self interpretKeyEvents:[NSArray arrayWithObject:event]];   // path 1: AppKit IME (NSTextInputClient)
    interpretingKeyEvents = NO;

    BOOL handled = NO;
    if (self.readyForWarp) {
        handled = warp_handle_view_event(self, event,
                                         wasComposing || [self hasMarkedText]);  // path 2: Rust / winit
    }
    ...
}
```

Most macOS IMEs render their candidate UI through the standard `IMKCandidates` panel (a separate `NSPanel` owned by the input method server), so they only consume events arriving through **path 1**. 超注音 hooks deeper into the input context and observes key events through **path 2** as well — so the right-arrow event reaches its internal candidate index twice, advancing it by 2.

This may also explain the long-standing #2335 (macOS Chinese candidate window misbehavior, open since 2022) and #9145 (Japanese IME popup misalignment in TUI apps): different surface symptoms (popup positioning vs. event double-counting) but plausibly the same root — inadequate isolation between the two key-event paths during IME composition on macOS.

#### Suggested fix

When `wasComposing || [self hasMarkedText]` is `YES` **and** the keycode corresponds to an arrow / page-navigation key, skip `warp_handle_view_event` (or have the Rust side early-return when the composing flag is set). This matches the contract Terminal.app and iTerm2 follow via standard NSTextInputClient delegation: while the IME is composing, the host should not act on navigation keys independently.

#### Related issues

- #2335 — macOS: candidate window doesn't follow cursor (open since 2022)
- #9145 — macOS: Japanese IME conversion candidates popup misaligned (TUI-specific)
- #9012 — Windows: pre-edit string not visible (Windows-specific, but same family of IME-path bugs)
- #9357 — Tracking issue for Chinese support layers

#### Willing to PR

I'm willing to send a PR for this. Would appreciate maintainer confirmation of the suggested fix direction (early-return on path 2 during composition for navigation keys) before opening one.

No (number-key selection works as a workaround, but arrow-key navigation flow is broken)

maxmilian:
## Description

Fixes #9709 — on macOS, third-party IMEs that listen on the input context (e.g. 超注音 / Yahoo Bopomofo) observed every arrow keystroke twice during candidate selection, advancing the highlight by 2 instead of 1.

### Root cause

Arrow keys carry the function-key flag (`modifierFlags=0xa00100`), so they qualify as key equivalents and AppKit dispatches them through `performKeyEquivalent:` before `keyDown:`. In `crates/warpui/src/platform/mac/objc/window.m::performKeyEquivalent:`, the previous implementation:

1. Called `[self.contentView keyDownImpl:event]` — which invokes `interpretKeyEvents:` and forwards to Rust
2. Rust suppresses keystrokes during composition (`is_composing=true`), so `keyDownImpl` returned `NO`
3. Fell through to `[super performKeyEquivalent:event]`
4. AppKit then dispatched `keyDown:` on the host view → `keyDownImpl` again → **`interpretKeyEvents` ran a second time**

Apple's built-in IME uses `IMKCandidates` (a separate `NSPanel`) for candidate selection, so the second delivery doesn't reach its candidate index. 超注音 draws its candidate panel itself and listens on the input context directly — it observes both deliveries and advances by 2.

### Fix

Skip the priority path in `performKeyEquivalent:` while `hasMarkedText` is `YES`. AppKit then routes the event through the standard `keyDown:` chain exactly once.

### NSLog instrumentation (before/after)

I added per-call instrumentation around `keyDownImpl`, `interpretKeyEvents`, `setMarkedText:`, `insertText:`, `unmarkText`, `doCommandBySelector:`, and `performKeyEquivalent:` and reproduced with 超注音 (Yahoo Bopomofo) on macOS 26.4.1.

**Before the fix** — single right-arrow press during candidate selection:
```
performKeyEquivalent: keyCode=124 modifiers=0xa00100
performKeyEquivalent: keystrokeIsAssigned=1
keyDownImpl#8 ENTER keyCode=124 wasComposing=1
  interpretKeyEvents BEGIN                    ← IME +1
  setMarkedText: '...' (interpretingKeyEvents=1)
  setMarkedText: '...' (interpretingKeyEvents=1)
  interpretKeyEvents END (hasMarkedText=1)
warp_handle_view_event(composing=1) handled=0
performKeyEquivalent: keyDownImpl returned NO, falling through to super
keyDown: keyCode=124                          ← AppKit fires again
keyDownImpl#9 ENTER keyCode=124 wasComposing=1
  interpretKeyEvents BEGIN                    ← IME +1 again → total +2
  ...
```

**After the fix**:
```
performKeyEquivalent: keyCode=124 modifiers=0xa00100
performKeyEquivalent: IME composing, skipping priority path
keyDown: keyCode=124
keyDownImpl#22 ENTER keyCode=124 wasComposing=1
  interpretKeyEvents BEGIN                    ← single delivery
  interpretKeyEvents END (hasMarkedText=1)
```

### Manual verification (macOS)

- 超注音 (Yahoo Bopomofo): right/left arrow now advances candidate by exactly 1 per press (was 2). ✅
- Apple built-in 注音 (Bopomofo): no regression — candidate window behaves identically. ✅
- Apple Pinyin: no regression. ✅
- Cmd-shortcut keys without IME composition (e.g. Cmd-T new tab): no regression — the `hasMarkedText` guard only triggers during active composition, so the priority path still executes for ordinary key equivalents. ✅
- Plain typing without an IME: no regression. ✅

No automated tests were added: the bug requires running an IME on macOS and observing AppKit's two-stage event delivery (`performKeyEquivalent:` → `keyDown:`), which is impractical to simulate in the existing unit/integration test harnesses. The change is small, well-scoped (`performKeyEquivalent:` only), and the guard short-circuits to `[super performKeyEquivalent:]` on the same path AppKit would fall back to anyway.

## Server API dependencies

- [ ] Is this change necessary to make the client compatible with a desired server API breaking change?
- [ ] Does this change rely on a new server API?
- [ ] Is this change enabling the use of a server API on client channels that rely on the production server?

None — this is a pure native-input-handling change.

## Agent Mode

CHANGELOG-BUG-FIX: macOS: fix third-party IMEs (e.g. 超注音 / Yahoo Bopomofo) advancing candidate selection by 2 per arrow press during composition.

### DIFF
--- a/crates/warpui/src/platform/mac/objc/window.m
+++ b/crates/warpui/src/platform/mac/objc/window.m
@@ -432,6 +432,23 @@ void init_warp_nswindow(NSWindow<WarpWindowProtocol> *window, bool testMode, boo
     // We need to bypass the default performKeyEquivalent implementation which, in the case of
     // having keybinding conflicts with MacOS itself, yields priority to the OS.
     if ([event type] == NSEventTypeKeyDown) {
+        // While the IME is composing (e.g. selecting candidates in a Bopomofo / Pinyin / Kana
+        // input method), skip the priority path entirely. AppKit will then route the event
+        // through the standard keyDown: chain exactly once, which calls interpretKeyEvents
+        // and lets the IME consume the keystroke.
+        //
+        // Without this guard, navigation keys with the function-key flag (e.g. arrow keys,
+        // 0xa00100) reach this method first because they qualify as key equivalents. We then
+        // call keyDownImpl, which invokes interpretKeyEvents and forwards to Rust. Rust
+        // suppresses keystrokes during composition, so keyDownImpl returns NO. We fall
+        // through to [super performKeyEquivalent:], AppKit fires keyDown: on the host view,
+        // and interpretKeyEvents runs a second time. IMEs that listen to every keystroke on
+        // the input context — for example 超注音 (Yahoo Bopomofo) — observe both deliveries
+        // and advance their candidate selection by 2 per arrow press. See #9709.
+        if ([(WarpHostView *)self.contentView hasMarkedText]) {
+            return [super performKeyEquivalent:event];
+        }
+
         NSApplication *application = [NSApplication sharedApplication];
 
         // If we are recording a keystroke for an EditableBinding.
--- a/crates/warpui/src/platform/mac/objc/window.m
+++ b/crates/warpui/src/platform/mac/objc/window.m
@@ -432,19 +432,11 @@ void init_warp_nswindow(NSWindow<WarpWindowProtocol> *window, bool testMode, boo
     // We need to bypass the default performKeyEquivalent implementation which, in the case of
     // having keybinding conflicts with MacOS itself, yields priority to the OS.
     if ([event type] == NSEventTypeKeyDown) {
-        // While the IME is composing (e.g. selecting candidates in a Bopomofo / Pinyin / Kana
-        // input method), skip the priority path entirely. AppKit will then route the event
-        // through the standard keyDown: chain exactly once, which calls interpretKeyEvents
-        // and lets the IME consume the keystroke.
-        //
-        // Without this guard, navigation keys with the function-key flag (e.g. arrow keys,
-        // 0xa00100) reach this method first because they qualify as key equivalents. We then
-        // call keyDownImpl, which invokes interpretKeyEvents and forwards to Rust. Rust
-        // suppresses keystrokes during composition, so keyDownImpl returns NO. We fall
-        // through to [super performKeyEquivalent:], AppKit fires keyDown: on the host view,
-        // and interpretKeyEvents runs a second time. IMEs that listen to every keystroke on
-        // the input context — for example 超注音 (Yahoo Bopomofo) — observe both deliveries
-        // and advance their candidate selection by 2 per arrow press. See #9709.
+        // Skip the key-equivalent priority path while the IME has marked text. Arrow keys carry
+        // NSEventModifierFlagFunction, so AppKit delivers them here before keyDown:. If we call
+        // keyDownImpl and Rust suppresses the keystroke (composing mode), we return NO, and AppKit
+        // proceeds to call keyDown: — running interpretKeyEvents a second time for the same event.
+        // See #9709.
         if ([(WarpHostView *)self.contentView hasMarkedText]) {
             return [super performKeyEquivalent:event];
         }


maxmilian:
@cla-bot check

vorporeal:
reassigning to @abhishekp106, who has done the most IME work

abhishekp106:
@maxmilian Thanks for opening this and diagnosing the issue! I agree with your assessment and the approach here looks sound. The only tweak I want to make is to make the comment a little more concise.

abhishekp106:
@oz-agent Can you change the comment above the new if statement added in this change to be the following?

> Skip the key-equivalent priority path while the IME has marked text. Arrow keys carry NSEventModifierFlagFunction, so AppKit delivers them here before keyDown:. If we call keyDownImpl and Rust suppresses the keystroke (composing mode), we return NO, and AppKit proceeds to call keyDown: — running interpretKeyEvents a second time for the same event. See #9709.

abhishekp106:
Update: Our oz-agent github bot does not yet run in forks yet. @maxmilian do you mind updating the comment to be what I wrote above? Then let's get this merged in!

abhishekp106 (approved):
LGTM! Just one small nit about the comment

maxmilian:
@abhishekp106 Updated the comment in 8b2311c — ready for merge.

abhishek-pandya:
Merged in! Thanks @maxmilian for your contribution :)

Issue #3944, opened 2023-12-07
Text input in multi-byte Japanese doesn't appear until confirmation


shikata1112:
### Describe the bug

Text input in multi-byte Japanese doesn't appear until confirmation

### To reproduce

Open Warp app
Input multi-byte Japanese character in command area or Warp AI

### Expected behavior

Displaying Japanese as it is typed before pressing enter.

![image](https://github.com/warpdotdev/Warp/assets/67092021/7e9bb592-8084-48c0-b6a8-52629f58ce70)

### Screenshots
bad:

![image](https://github.com/warpdotdev/Warp/assets/67092021/7a180bae-09ac-49e3-b068-d94905ac1dc7)

### Operating system

MacOS

### Operating system and version

Sonoma 14.1.2

### Current Warp version

v0.2023.11.28.08.02.stable_00

### Regression

Yes, this bug started recently or with an X Warp version

### Does this block you from using Warp daily?

Yes, this issue prevents me from using Warp daily.

### Is this a Warp specific issue? (i.e. does it happen in Terminal, iTerm, Kitty, etc.)

No, this same issue happens in Warp and other terminals.

### Warp Internal (ignore): linear-label:b8107fdf-ba31-488d-b103-d271c89cac3e

None

dannyneira:
Thanks for letting us know @shikata1112, this may be related to a known issue with rendering ligatures you can track here #239

To anyone else facing this issue, please add a :+1: to the original post at the top or comment with your details, and subscribe if you'd like to be notified.

odiak:
@dannyneira Hi, I'm also facing same problem.
Maybe it's not related to ligatures.
Inputting languages like Japanese, Chinese or Korean needs to type some keys and compose characters.

On Warp, characters while typing (and not yet finalized) is not displayed.
If there're some Chinese people in Warp Team, ask them about this problem.
Thanks.

Sonnar1824ZA:
Still reproducible in Warp **v0.2026.07.08.17.54.stable_02** on **macOS 27.0.0**.

A specific context where I can reproduce it is while editing a `.md` file in Warp:

1. Open/edit any Markdown file in Warp.
2. Switch to the Japanese IME.
3. Type a phrase that requires kana-kanji conversion.
4. Before confirming the conversion, observe the composing (preedit/marked) text.

**Actual:** The composing text is completely invisible. The characters appear only after the conversion is confirmed.

[...]

PR #9713, opened 2026-05-01
fix(ime): preserve Korean IME last syllable across commits (#8919)

### LINKED ISSUE #8919: Korean IME: Last composing character is dropped when pressing Enter
## Bug Description

When typing Korean text in Warp, the last character being composed by the Korean IME is dropped/lost when pressing Enter.

## Steps to Reproduce

1. Open Warp terminal
2. Run any interactive CLI app (e.g., `claude`, `gemini`)
3. Switch to Korean input method
4. Type "안녕하세요" (Korean for "hello")
5. Press Enter immediately

## Expected Behavior

The full text "안녕하세요" is submitted.

## Actual Behavior

Only "안녕하세" is submitted — the last character "요" is dropped because it was still in IME composing (pre-edit) state when Enter was pressed.

## Root Cause Analysis

When Enter is pressed, Warp sends the Enter keypress to the PTY **before** committing the IME pre-edit text. Native macOS apps (Safari, Notes, etc.) commit the composing character first, then process Enter. Warp skips this IME commit step.

## Environment

- **Warp version:** 0.62.2
- **OS:** macOS Darwin 25.3.0 (Apple Silicon, M1 Pro)
- **Input method:** macOS default Korean IME
- **Affected CLI apps:** Claude Code, Gemini CLI (any app reading from PTY)

## Workaround

Pressing the right arrow key (→) before Enter forces the IME to commit the composing character.

## Notes

- English input is **NOT** affected (no composition step)
- Native macOS apps are **NOT** affected (they handle IME commit properly)
- This is **NOT** specific to any CLI app — it affects all PTY-reading applications running inside Warp
- Related issues: #8786, #8736, #3127 (Korean/Japanese IME issues, but different symptoms)

ohah:
## Description

Fixes a long-standing Korean (and broader CJK) IME bug on macOS where the last composing character could be silently dropped — both when terminating composition with Enter/Arrow keys and during ongoing composition that bridges syllables.

## Linked Issue

Fixes #8919.

- [x] The linked issue is labeled `triaged` (implicitly `ready-to-implement` per CONTRIBUTING.md).
- [x] Screenshots / video

## Root cause

Two bugs in `crates/warpui/src/platform/mac/objc/host_view.m`'s `keyDownImpl:` interacted with the way Korean (and other CJK) IMEs commit pre-edit text:

1. **`!handled` guard discarded committed text on Arrow/Enter.**
   When the user pressed Enter/Arrow to terminate composition, the macOS Korean IME committed the trailing syllable via `insertText:`, but the existing guard `if ([textToInsert length] > 0 && !handled)` then discarded it because the input field had handled the key. So the very last syllable (e.g. the '요' in '안녕하세요') was lost.
2. **Commit overwrote the next marked-text selection.**
   Korean IMEs can both commit a previous syllable AND start a new composition within the same keyDown — typing 'ㅏ' after '간' produces commit '가' + new marked '나'. By the time we dispatched the commit, `setMarkedText:` had already placed the new marked text as a selection in the input field's buffer, and `insert_internal` (called by `user_insert`) overwrote that selection with the committed text — losing the new syllable.

## Fix

In `keyDownImpl:`:

1. Loosen the guard to `(!handled || wasComposing)` so a key that was both consumed by the input field AND committed previously-composed text still dispatches the committed text (it represents the user's already-typed input, not text produced by the keybinding).
2. When dispatching a commit while a new marked-text composition is already in place, temporarily clear marked text, dispatch the commit, then re-apply the marked text. The new selectedRange is captured at `setMarkedText:` time into a new instance variable `lastMarkedSelectedRange` so we can re-emit the same range afterwards.

## Testing

Manual repro on macOS with `cargo run --features ime_marked_text` (release/dogfood builds enable this flag automatically — see `crates/warp_features/src/lib.rs`):

1. Switch to macOS Korean IME.
2. Type `안녕하세요` and press **Enter** → before: `안녕하세` is sent, `요` lost. After: full `안녕하세요` is sent.
3. Type `안녕하세요` and press **→ / ↓** → before: trailing `요` dropped. After: preserved.
4. Type `가나다` slowly and watch composition during typing → before: the last syllable being composed (e.g. `다`) wasn't shown until the next syllable bumped it. After: each composing syllable is visible as marked text in real time.
5. Sanity check: dead-key flow (Option-E + g → ´g) still works; English input unaffected.

No automated tests added — `host_view.m` is Objective-C and not directly unit-testable; the `marked_text_tests.rs` suite exercises the editor model layer below this fix and continues to pass. Happy to add an integration test if reviewers want one.

## Agent Mode

https://github.com/user-attachments/assets/402f2c81-e5a9-4cc3-b1d6-e9955425e3ec

https://github.com/user-attachments/assets/6d68c0f3-52c4-4721-aa43-13979646367e

### DIFF
--- a/crates/warpui/src/platform/mac/objc/host_view.m
+++ b/crates/warpui/src/platform/mac/objc/host_view.m
@@ -59,6 +59,13 @@ void warp_marked_text_cleared(WarpHostView *);
     // The selectedRange of the most recent setMarkedText: call, kept so we
     // can re-dispatch the marked text after a commit (see keyDownImpl:).
     NSRange lastMarkedSelectedRange;
+
+    // True if setMarkedText: was actually called during the current
+    // interpretKeyEvents:. Used to distinguish a freshly-started composition
+    // from stale markedText that survived a commit-only flow — insertText:
+    // deliberately skips unmarkText while interpretingKeyEvents is true, so
+    // hasMarkedText alone cannot tell the two apart.
+    BOOL setMarkedTextDuringInterpret;
 }
 
 - (BOOL)acceptsFirstResponder {
@@ -157,6 +164,7 @@ void warp_marked_text_cleared(WarpHostView *);
 - (BOOL)keyDownImpl:(NSEvent *)event {
     BOOL wasComposing = [self hasMarkedText];
     [textToInsert setString:@""];
+    setMarkedTextDuringInterpret = NO;
 
     // Interpret the key events here so we could check whether user is composing
     // text within the IME and pass the state down to the KeyDown events.
@@ -197,7 +205,13 @@ void warp_marked_text_cleared(WarpHostView *);
         // text would then overwrite that selection (losing the next
         // character). Workaround: temporarily clear the marked text,
         // dispatch the commit, then re-apply the marked text.
-        BOOL hasNewMarked = [self hasMarkedText];
+        //
+        // We can only treat hasMarkedText as a "freshly-started composition"
+        // when setMarkedText: was actually called during this keyDown.
+        // insertText: leaves stale markedText untouched while
+        // interpretingKeyEvents, so commit-only flows (e.g. dead-key + char)
+        // would otherwise see stale marked text and incorrectly re-emit it.
+        BOOL hasNewMarked = setMarkedTextDuringInterpret && [self hasMarkedText];
         if (hasNewMarked) {
             warp_marked_text_cleared(self);
             warp_update_ime_state(self, NO);
@@ -206,6 +220,14 @@ void warp_marked_text_cleared(WarpHostView *);
         if (hasNewMarked) {
             warp_marked_text_updated(self, markedText.string, lastMarkedSelectedRange);
             warp_update_ime_state(self, YES);
+        } else {
+            // Commit-only flow (no new composition started). End IME state
+            // explicitly — insertText: deliberately skipped unmarkText while
+            // interpretingKeyEvents was true, so the stale markedText object
+            // and ime_active flag would otherwise survive into the next
+            // keyDown and break downstream input (e.g. Backspace would be
+            // treated as composing-state input and not delete the character).
+            [self unmarkText];
         }
     }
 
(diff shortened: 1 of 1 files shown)


vorporeal:
reassigning to @abhishekp106, who has done the most IME work

abhishekp106:
Hi @ohah, thanks for taking the time to write this up. I think this fix does work correctly, but I decided to go ahead with the approach in https://github.com/warpdotdev/warp/pull/9730. Similar to your approach, they set a flag when handling marked text insertions during a `interpretKeyEvents` pass. Both that PR and yours fix the same issue, if I'm understanding correctly.

I'm a bit wary of adding the `|| wasComposing` to this path as it means that even if `rust_handle_view_event` did return `true`, we might double-handle an enter key down. I did a quick sanity check and the rust app should not handle any event while the IME is open. As a result, we don't need to permit double-handled events. It also preserves the marked text instead of re-inserting it, which seems easier to maintain going forward.

That being said, thank you for looking into this!

abhishekp106:
I'm going go ahead and close this PR to keep our PR queue clean.

ohah:
@abhishekp106 — I opened #11624 as a follow-up to #9730, addressing two Korean IME flows that remained broken after the merge:

1. **Split-commit composition visibility** — the in-progress jamo is invisible until the next keystroke. Tracked as #3944 (which has `ready-to-implement`).
2. **Backspace requiring two presses** to delete a marked syllable — caused by `doCommandBySelector(deleteBackward:)` being a no-op while the IME signals it during `interpretKeyEvents:`.

I addressed the two concerns you raised when closing this PR:
- **Dropped the `(!handled || wasComposing)` guard** entirely — no double-handled events. The Enter case is already covered by #9730 + the new `deleteBackward:` handling.
- **Reused #9730's existing `imeTouchedMarkedTextDuringInterpret` flag** instead of introducing a parallel mechanism.

Would you mind taking a look when you have time? Thanks!

Issue #8988, opened 2026-04-01
Settings Sync causes custom theme to revert to default on other machine


artseabra:
## Description

When using Warp on two macOS machines (M1 and M5 MacBooks) with Settings Sync enabled and the custom themes repo cloned on both (`~/.warp/themes/`), changing to a custom theme (e.g. base16 "Summer Fruit Dark" or "Apple Park Dark") on one machine causes the other machine to immediately revert to the default dark theme.

The theme on the machine where the change was made looks correct, but the other machine's entire interface goes to what appears to be the default Warp Dark theme — as if the synced theme selection can't resolve the custom theme on the other device.

## Steps to Reproduce

1. Set up two macOS machines with Warp and Settings Sync enabled.
2. Clone the official themes repo on both machines: `git clone https://github.com/warpdotdev/themes.git ~/.warp/themes`
3. On Machine A, select a custom theme from the cloned repo (e.g. "Summer Fruit Dark" from base16, or "Apple Park Dark").
4. Observe Machine B — it immediately reverts to the default dark theme instead of matching the selected custom theme.

## Expected Behavior

Both machines should display the same custom theme, since the themes repo is cloned on both and Settings Sync should propagate the theme selection which both machines can resolve.

## Actual Behavior

The machine where the theme was NOT changed reverts to the default Warp Dark theme. The selected custom theme only displays correctly on the machine where the change was made.

## Troubleshooting Attempted

- Toggled off "Sync with OS" theme setting — no effect.
- Confirmed the themes repo is cloned and present on both machines at `~/.warp/themes/`.
- Tried multiple custom themes (base16 Summer Fruit Dark, Apple Park Dark) — same behavior.

## Environment

- **OS:** macOS (both machines)
- **Hardware:** Apple M1 MacBook and Apple M5 MacBook
- **Warp version:** 0.2026.03.25.08.24.05
- **Settings Sync:** Enabled
- **Custom themes:** Cloned from https://github.com/warpdotdev/themes.git into `~/.warp/themes/` on both machines

artseabra:
> Afaik the theme syncs only work for default themes, not custom themes, in those cases it switches back to default themes. Are the same themes installed on both machines in the same location?

Yes, i have an M1 MacBook i'm using as a headless server right next to my main M5, and i'm hopping in and out of Warp on both all day on the same account.

PR #9728, opened 2026-05-01
fix(themes): sync custom theme selections portably

### LINKED ISSUE #9723: Settings Sync stores custom theme selection as absolute paths, breaking cross-platform sync
## Summary
Settings Sync stores custom theme selections as absolute file paths. When the same Warp account is used on macOS and Linux, a custom theme selected on one platform does not work on the other because Warp's supported custom theme directories are platform-specific.

## Problem
A user syncs Warp settings between macOS and Linux and has the same custom theme YAML files installed on both systems. The selected custom theme is written to `settings.toml` under `appearance.themes.selected_system_themes` with a `custom.path` value. That path is absolute and OS-specific.

Documented theme locations differ by platform:
- macOS: `~/.warp/themes/`
- Linux: `${XDG_DATA_HOME:-$HOME/.local/share}/warp-terminal/themes/`

As a result, choosing the custom theme on macOS syncs a path like `/Users/ivan/.warp/themes/catppuccin_mocha.yml`, which Linux does not treat as the active discovered theme. Choosing the theme on Linux syncs a path like `/home/nisavid/.local/share/warp-terminal/themes/catppuccin_mocha.yml`, which macOS cannot use.

This makes Settings Sync fight itself for users who use the same custom themes across multiple OSes.

## Expected behavior
Warp should allow custom theme selection to sync across platforms when the same theme is installed on each platform, without requiring users to make fake absolute paths.

Possible acceptable product behavior would be one of:
- Sync custom theme selection by stable identity, such as theme name plus theme type, instead of absolute file path.
- Treat custom theme selection as platform-specific/device-specific, like other platform-specific settings.
- Resolve synced custom theme paths through each platform's documented theme directory when a matching theme exists.

## Actual behavior
The selected custom theme is synced using an absolute path. The setting only works on the OS whose absolute path is stored in Settings Sync.

Observed behavior from the Linux side:
- When `settings.toml` was manually changed to macOS-style paths under `/Users/ivan/.warp/themes/*.yml`, Warp did not use the custom themes on Linux.
- Creating a symlink at `/Users/ivan/.warp/themes` pointing to the Linux theme directory did not make Warp accept the macOS paths.
- Reversing the symlink so the real files lived under `/Users/ivan/.warp/themes` and Linux's theme directory pointed there also did not work.
- Bind-mounting the Linux theme directory at `/Users/ivan/.warp/themes` also did not work, even though both paths had the same device and inode and the configured files were readable.
- After selecting the themes in the Linux theme picker, Warp rewrote the synced value back to the Linux-discovered paths and updated the cloud preference with those paths.

Example values that get synced:

```toml
[appearance.themes]
selected_system_themes = {
  dark = { custom = { name = "Catppuccin Mocha", path = "/Users/ivan/.warp/themes/catppuccin_mocha.yml" } },
  light = { custom = { name = "Catppuccin Latte", path = "/Users/ivan/.warp/themes/catppuccin_latte.yml" } },
}
```

On Linux, selecting the same themes rewrites them to:

```toml
[appearance.themes]
selected_system_themes = {
  dark = { custom = { name = "Catppuccin Mocha", path = "/home/nisavid/.local/share/warp-terminal/themes/catppuccin_mocha.yml" } },
  light = { custom = { name = "Catppuccin Latte", path = "/home/nisavid/.local/share/warp-terminal/themes/catppuccin_latte.yml" } },
}
```

1. Install the same custom theme YAML files on macOS and Linux using Warp's documented custom theme directories.
2. Enable Settings Sync on the same Warp account on both OSes.
3. On macOS, select a custom theme from `~/.warp/themes/`.
4. Let Settings Sync propagate the selection to Linux.
5. Observe that Linux receives the macOS absolute path in the selected custom theme setting and the theme does not work as expected.
6. On Linux, select the same custom theme from `${XDG_DATA_HOME:-$HOME/.local/share}/warp-terminal/themes/`.
7. Let Settings Sync propagate the selection to macOS.
8. Observe that macOS receives the Linux absolute path and the theme no longer works there.

## Artifacts
None attached.

## Warp version
v0.2026.04.27.15.32.stable_03

## Operating system
Observed on Linux: CachyOS, kernel/version `7.0.1-1-cachyos`.

Also affects sync with macOS. macOS version is unknown; the macOS home directory in the user's setup is `/Users/ivan`.

nisavid:
## Summary
- Store custom theme references under Warp's theme root as relative paths so Settings Sync does not publish machine-specific theme roots.
- Resolve portable references and legacy macOS/Linux absolute theme-root paths against the local theme directory when reading synced settings.
- Allow cloud sync for custom and Base16 theme selections only when every custom theme reference can be represented portably.

Fixes warpdotdev/warp#9723.
Addresses the root cause behind warpdotdev/warp#8988 and warpdotdev/warp#6678.

## Relationship to #9634
PR warpdotdev/warp#9634 handles tilde expansion/contraction for home-relative paths. This PR is broader: it stores and resolves custom theme references relative to Warp's platform theme root, so the same selected theme can sync between macOS ~/.warp/themes and Linux XDG data theme roots.

## Notes
Settings Sync still syncs only the custom theme reference. It does not upload, download, create, copy, or manage custom theme YAML files.

## Test Plan
- [x] cargo test -p warp themes::theme::tests
- [x] cargo test -p warp settings::theme::tests
- [x] cargo fmt --check
- [x] cargo clippy -p warp --lib --tests -- -D warnings
- [x] env PATH=/tmp/codex-corepack-bin:/usr/local/sbin:/usr/local/bin:/usr/bin cargo clippy --workspace --all-targets --tests -- -D warnings
- [ ] env PATH=/tmp/codex-corepack-bin:/usr/local/sbin:/usr/local/bin:/usr/bin cargo clippy --workspace --all-targets --all-features --tests -- -D warnings (blocked locally: --all-features enables release_bundle, but warp-channel-config is not installed on PATH)


nisavid:
@cla-bot check

app/src/themes/theme.rs
```diff
@@ -148,29 +148,161 @@ impl ThemeKind {
...
+        && path
+            .components()
+            .all(|component| matches!(component, Component::Normal(_) | Component::CurDir))
+}
+
+fn legacy_theme_root_relative_path(path: &Path) -> Option<PathBuf> {
+    let components = path.components().collect::<Vec<_>>();
+
+    relative_path_after_marker(&components, &[".warp", "themes"])
+        .or_else(|| relative_path_after_marker(&components, &["warp-terminal", "themes"]))
```
rrroyal: i'm pretty sure other warp channels use different paths, i.e. preview uses `.warp-preview` instead of `.warp`. does it also handle that case?

windows also uses different paths. maybe it would be better if it used platform-specific logic with conditional compilation instead of "hardcoding" paths here?

nisavid: Good call. I agree the current suffix matching is the wrong middle ground.

A synced absolute theme path may have been written by a different platform, channel, user account, or data profile than the build now reading it. Current-platform conditional compilation can identify the current installation’s theme root, but it cannot reliably interpret every legacy absolute path that may arrive through sync.

I’m going to narrow this PR to the durable behavior:

- safe relative paths resolve under the current `themes_dir()`
- absolute paths under the current `themes_dir()` serialize as relative and are syncable
- other absolute or foreign-looking paths are preserved and are not treated as syncable

That avoids an incomplete cross-platform/channel legacy migration. New saves get the portable representation. Existing foreign absolute selections keep their current behavior until the user reselects the theme and saves the portable form.

app/src/themes/theme_test.rs
```diff
@@ -1,6 +1,196 @@
...
+    ));
+}
+
+#[test]
+fn custom_theme_absolute_parent_dir_path_under_theme_root_is_preserved() {
+    let root = PathBuf::from("/Users/example/.warp/themes");
+    let stored = root.join("../outside.yml");
+
+    assert_eq!(custom_theme_path_from_storage(&stored, &root), stored);
+}
```
lucieleblanc: This test seems like a duplicate of `custom_theme_relative_parent_dir_path_is_preserved` above?

app/src/settings/theme_tests.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+fn custom_base16(path: PathBuf) -> ThemeKind {
+    ThemeKind::CustomBase16(CustomTheme::new("Base16 Custom".to_string(), path))
+}
+
+#[test]
+fn theme_kind_syncs_custom_theme_under_theme_root() {
+    let setting = Theme::new(Some(custom(
+        crate::user_config::themes_dir().join("custom.yml"),
```
lucieleblanc: Style nit: please import `user_config` at the top of the file, instead of repeatedly qualifying inline in each test. Same goes in the other test file.

app/src/themes/theme_tests.rs
lucieleblanc, line 1: All of the tests in this file use hard-coded UNIX-style roots, so I'm not convinced this works cross-platform.

Could we add some `#[cfg(windows)]` tests that use a Windows theme root (for example, `C:\Users\example\AppData\Roaming\warp\Warp\data\themes`) as well?

lucieleblanc (changes requested):
Thank you for the PR! Requesting changes because there's some Windows details we should iron out before this merges.

My main concern is cross-platform path separators. The changes + tests here assumes the filesystem accepts backslashes `\` as a path separator. 

For example: On Windows, a theme at `C:\Users\…\themes\catppuccin\mocha.yml` would be stripped to `catppuccin\mocha.yml`. When deserializing on a Unix system, `PathBuf::from("catppuccin\\mocha.yml")` reads the backslash as part of the filename, not a separator. The path becomes a single component, and `theme_root.join(…)` produces something like `/home/user/.local/share/warp-terminal/themes/catppuccin\mocha.yml`, which seems wrong. 

I think the current setup happens to work when using Unix paths on Windows, because Windows accepts `/` as a separator. Unix systems don't accept `\` as a separator though. 

Maybe normalizing paths to use `/` during serialization, then deserializing into the native separator could work?

app/src/themes/theme_tests.rs
nisavid: Added `#[cfg(windows)]` coverage for a Windows theme root, slash-separated serialization/deserialization, syncability, and raw backslash/drive-prefix rejection.

app/src/themes/theme_test.rs
nisavid: Cleaned up the helper-level test names, made the absolute traversal case distinct, and added raw `..` settings/serde ingress coverage.

nisavid:
@lucieleblanc I pushed follow-ups addressing the review feedback.

The custom theme path storage now normalizes portable paths to `/`, validates raw stored strings before `PathBuf` parsing, and preserves/non-syncs ambiguous values like raw backslashes, `..`, and Windows drive-prefixed paths. I also added Windows-gated coverage, serde/SettingsValue ingress tests, and cleaned up the helper-level test names.

Could you take another look when you have a chance?

app/src/themes/theme.rs
```diff
@@ -148,29 +148,228 @@ impl ThemeKind {
...
+    if path_is_absolute_or_foreign_absolute(path) {
+        return portable_custom_theme_storage_string(path, theme_root)
+            .map(|path| portable_custom_theme_path_from_stored_raw(&path, theme_root))
+            .unwrap_or_else(|| path.to_path_buf());
+    }
+
+    path.to_str()
+        .map(|path| portable_custom_theme_path_from_stored_raw(path, theme_root))
+        .unwrap_or_else(|| path.to_path_buf())
+}
```
lucieleblanc: nit: please move these to the test file

lucieleblanc (approved):
Nice, thank you for the edits!

app/src/themes/theme.rs
nisavid: @lucieleblanc done

PR #9831, opened 2026-05-01
Add podman exec/run support for warpify subshells

### LINKED ISSUE #10741: Warpify does not work with podman exec -it <container> /bin/bash
### Pre-submit Checks

### Describe the bug

Warpify does not work with podman exec -it <container> /bin/bash. Three underlying issues prevent it:

 1. The subshell command regex allowlist only matches docker exec/run, not podman exec/run, so the warpify banner never appears.
 2. Even when manually triggered via the added_subshell_commands override, warpify hangs at "Starting shell ..." because the DCS escape sequences use the C1 ST byte (0x9c) as the string terminator. Podman's PTY proxy (conmon) mangles this byte since it falls in the UTF-8 continuation byte range (0x80–0xBF), so the DCS sequence never terminates and the VTE parser consumes all subsequent output.
 3. After fixing the DCS terminator, the full ~32KB bootstrap script is written to the PTY in a single burst. Podman's double-PTY architecture (host PTY → conmon → container PTY) drops data for writes larger than ~8–16KB, so the bootstrap never completes.

### To reproduce

 1. Start a container: podman run -dit --name test ubuntu:latest bash
 2. In Warp, run: podman exec -it test /bin/bash
 3. No warpify banner appears (podman not in the allowlist)
 4. If you add podman exec to Settings → Warpify → Added Subshell Commands, the banner appears but clicking "Warpify subshell" hangs at "Starting shell ..." indefinitely

### Expected behavior

 Warpify should work with podman exec the same way it works with docker exec — the warpify banner should appear and clicking it should bootstrap the subshell successfully.

_No response_

### Operating system (OS)

Linux

Fedora 44 (Linux 7.0.4-200.fc44.x86_64)

### Shell Version

bash 5.2

### Current Warp version

v0.2026.05.06.15.42.stable_04

### Regression

### Recent working Warp date

_No response_

### Additional context

Podman uses the same CLI interface as Docker but its container runtime monitor (conmon) proxies PTY I/O differently. The double-PTY architecture (host PTY → conmon → container PTY) has two characteristics that differ from Docker: (1) the 0x9c byte gets mangled in UTF-8 mode, and (2) large single writes (>8–16KB) overflow intermediate buffers. A fix is submitted in PR #9831.

None

djdanielsson:
## Description

Warpify currently only recognizes `docker exec` and `docker run` as container subshell commands. This PR adds support for `podman exec` and `podman run`, and fixes underlying issues that prevented warpify from working through podman's (and potentially docker's) double-PTY proxy layer.

### Problem

When attempting to warpify a `podman exec -it <container> bash` session, three issues prevented it from working:

1. **Regex allowlist**: The subshell command regexes only matched `docker`, not `podman`.

2. **DCS terminator byte `0x9c` mangled by podman's PTY proxy**: The bootstrap init scripts use DCS (Device Control String) escape sequences to send hooks like `InitShell` back to Warp. These sequences were terminated with the C1 ST byte (`0x9c`). Podman's PTY proxy (conmon) mangles this byte because `0x9c` falls in the UTF-8 continuation byte range (`0x80`–`0xBF`), causing the DCS sequence to never terminate. Warp's VTE parser stays in DCS passthrough mode and consumes all subsequent output — the warpify process hangs at "Starting shell ...".

3. **Bootstrap script too large for double-PTY buffer**: After the `InitShell` hook is received, Warp writes the full ~32KB bootstrap script (`bash.sh` with `bash_body.sh` inlined) to the PTY in a single burst. Podman's double-PTY architecture (host PTY → conmon proxy → container PTY) has intermediate buffers that overflow with writes larger than ~8–16KB, causing data loss. The bootstrap script never completes and the `Bootstrapped` hook is never sent.

4. **RC-file bootstrap incompatible with containers**: For shells like Fish that use RC-file based bootstrap (writing a temp file on the host and sourcing it), this method fails inside containers since the host filesystem is not accessible from within the container.

### Fix

1. **Regex**: Updated `SUBSHELL_COMMAND_REGEXES` to match `(docker|podman)` for both `run` and `exec` patterns.

2. **DCS terminator**: Replaced `\x9c` (C1 ST) with `\x1b\x5c` (7-bit ST: `ESC \`) in all bootstrap scripts. The 7-bit ST uses only ASCII bytes (`0x1b`, `0x5c`) that pass through any PTY layer reliably. Uses explicit hex escapes rather than `\e\\` to avoid backslash interpretation issues across different shells and platforms. The parser already handles both C1 and 7-bit ST — a new test confirms this.

3. **Chunked writes**: For container exec sessions (detected via `is_container_subshell()` checking if the spawning command starts with `docker ` or `podman `), the bootstrap script is written in 4KB chunks with 50ms delays between them, allowing conmon to relay data without buffer overflow.

4. **Skip RC-file bootstrap for containers**: Extracted `is_container_subshell()` as a shared helper in `bootstrap.rs`. `should_use_rc_file_bootstrap_method()` now returns `false` for container subshells, ensuring all shells (including Fish) use the chunked direct-write path instead of trying to source an inaccessible host temp file.

## Linked Issue

- This is a bug fix for podman support — podman uses the same CLI interface as Docker but its PTY proxy has different byte-handling characteristics.
- closes #10741

## Testing

- Added `parse_dcs_init_shell_7bit_st` test to verify the VTE parser correctly handles 7-bit ST (`ESC \`) as DCS terminator
- All existing DCS and bootstrap tests pass (7 DCS tests, 13 bootstrap tests)
- Verified `\x1b\x5c` produces correct bytes inside podman container using `od` hex dump
- Confirmed the ~8–16KB PTY buffer threshold by testing progressively larger scripts through `podman exec -it`
- Manually tested warpify with `podman exec -it` on Linux — bootstrap completes successfully
- `cargo fmt` and `cargo clippy` pass (no new warnings)

## Agent Mode


djdanielsson:
@cla-bot check

app/assets/bundled/bootstrap/bash_body.sh
```diff
@@ -12,8 +12,8 @@ if [ -z "$WARP_BOOTSTRAPPED" ]; then
     # Appended to $DCS_START to signal that the following message is JSON-encoded.
     DCS_JSON_MARKER="d"
 
-    # Byte used to signal the end of a DCS.
-    DCS_END="$(printf '\x9c')"
+    # Byte sequence used to signal the end of a DCS (7-bit ST: ESC \).
```
zachbai, line 15: @alokedesai @vorporeal  is there any reason you can think of we opted for 8-bit representation of ST instead of 7-bit?

app/src/terminal/bootstrap.rs
```diff
@@ -50,13 +50,27 @@ const BYTE_ORDER_MARK: &str = "\u{FEFF}";
 /// bootstrap is the only known way to bootstrap such subshells successfully.
 ///
 /// We use RC-file based bootstrap for MSYS2 because it has slow PTY throughput.
+#[cfg(feature = "local_fs")]
+pub fn is_container_subshell(session_info: &SessionInfo) -> bool {
```
zachbai: nit, can you move this so its not inserted between the doc comment and function declaration for `should_use_rc_file_bootstrap_method`

app/src/terminal/writeable_pty/pty_controller.rs
```diff
@@ -468,6 +468,19 @@ impl<T: EventLoopSender> PtyController<T> {
...
+        } else if bootstrap::is_container_subshell(pending_session_info) {
+            // Write in 4KB chunks with 50ms delays to avoid overwhelming
+            // PTY buffers in container exec sessions (podman/docker exec -it),
+            // where the double-PTY proxy drops data for large writes.
+            const CHUNK_SIZE: usize = 4096;
+            let bytes: Vec<u8> = bootstrap.into_owned();
+            let chunks: Vec<Vec<u8>> = bytes.chunks(CHUNK_SIZE).map(|c| c.to_vec()).collect();
+            for (i, chunk) in chunks.into_iter().enumerate() {
+                ctx.spawn(
+                    warpui::r#async::Timer::after(std::time::Duration::from_millis(i as u64 * 50)),
```
zachbai, line 480: how much testing did u do on the interval/chunk size here? Its not a huge deal but given our bootstrap script is ~70kb this probably means it takes about a second to feed everything in?

zachbai (approved):
thanks for the change, just a small nit on the helper fn

app/src/terminal/bootstrap.rs
djdanielsson: Good catch, fixed — moved `is_container_subshell` above the doc comment block so it stays attached to `should_use_rc_file_bootstrap_method`.

app/src/terminal/writeable_pty/pty_controller.rs
djdanielsson: We tested progressively larger single writes through `podman exec -it` (1KB, 2KB, 4KB, 8KB, 16KB, 32KB) and found the PTY buffer threshold is between 8–16KB — everything up to 8KB succeeded, 16KB+ failed. The minified bootstrap is ~32KB (comments are stripped), so 4KB chunks give solid margin below the threshold. The 50ms delay was a conservative choice — we didn't sweep different delay values, so it could likely be reduced if the ~400ms total overhead (8 chunks × 50ms) is a concern. Either way it only applies to docker/podman subshells, so normal sessions are unaffected.

djdanielsson:
I have rebased and would like to verify the CI issues are fixed, if the CI still fails then I might need to make some adjustments.

djdanielsson:
I created an issue to track #10741

app/assets/bundled/bootstrap/bash_body.sh
vorporeal: no idea; would have been before my time.  the only reason i can think we'd care is the 1 byte vs. 2, but DCS terminators are such a tiny percentage of total PTY throughput that we won't remotely notice.

app/src/terminal/writeable_pty/pty_controller.rs
zachbai: That sounds reasonable, thanks for following up

zachbai (approved):

Issue #11, opened 2021-07-16
Font rendering artifacts


zachlloyd:
There are various font rendering artifacts that make the text look suboptimal.

E.g. the tops of letters are sometimes cropped:

![image](https://user-images.githubusercontent.com/8727141/126004718-1046230c-945f-4b69-ae90-33e9adabeebd.png)

Sometimes characters are misaligned from the baseline at different zoom levels.

![New Issue · warpdotdevwarp 2021-07-16 at 4 27 56 PM](https://user-images.githubusercontent.com/8727141/126004983-c34c1913-4f9b-4a2e-a96b-51d6abb50334.jpg)

michlimlim:
We just released the fix for this today in `v0.2021.08.11.15.36.stable_00`

Issue #10, opened 2021-07-16
Visual artifacts on non-retina external display


kesne:
**Describe the bug**
There appear to be visual artifacts in a handful of places when moving the terminal from the build-in retina display to an external non-retina display. This was 

**To Reproduce**
Open warp on a non-retina display, inspect edges closely.

**Expected behavior**
Crispness.

**Screenshots**
On the Retina display, things look _generally_ fine:
<img width="72" alt="Screen Shot 2021-07-16 at 1 23 58 PM" src="https://user-images.githubusercontent.com/498479/126004698-ef5d1ea6-bb2e-42ec-94ae-a96460369a69.png">
<img width="64" alt="Screen Shot 2021-07-16 at 1 27 15 PM" src="https://user-images.githubusercontent.com/498479/126004927-98abff83-1da5-4bf7-99cf-26d555c5555b.png">

On the external display, the edges look crunchy. 
![Screen Shot 2021-07-16 at 10 47 43 AM](https://user-images.githubusercontent.com/498479/126004841-8e75b541-27ed-4334-98f1-728f41307d85.png)
![Screen Shot 2021-07-16 at 1 27 10 PM](https://user-images.githubusercontent.com/498479/126004941-4bef9662-7aba-49aa-9e1b-a5ddc22019a5.png)

**Desktop (please complete the following information):**
MacOS 11.2
External display: 3440 x 1440 display

alokedesai:
Thanks for filling! This is a known issue the team is looking into

michlimlim:
We just released the fix for this today in `v0.2021.08.11.15.36.stable_00`

PR #9847, opened 2026-05-01
Support bindkeys  #537

### LINKED ISSUE #537: Support zsh abbreviations, bindkeys, and history expansion
### Describe the bug

[Fish abbreviations](https://fishshell.com/docs/current/cmds/abbr.html) work just fine, but zsh abbreviations coming from the [zsh-abbr plugin](https://github.com/olets/zsh-abbr) don't work at all. When I press `space`, nothing happens.

### To Reproduce

Steps to reproduce:

1. Use zsh
2. Install this plugin: https://github.com/olets/zsh-abbr
3. Try to create an abbreviation: `abbr gs='git status'`
4. Try to use the abbreviation by typing: `gs` and then `space`

### Expected behaviour

```shell
The abbreviation should be expanded
```

### Screenshots

_No response_

### Operating System

_No response_

### OS Version

_No response_

### Additional context

_No response_

th1nkful:
## Description

Spec PR for #537 — honor user-defined shell bindkeys (zsh bindkey, bash readline bind / ~/.inputrc, fish bind) in Warp’s input editor, instead of silently ignoring them.
Adds two short documents under specs/GH537/:
	•	product.md — 22 numbered behavior invariants covering discovery, dynamic rebind, the widget set we honor, vi/emacs mode handling, multi-tab independence, conflict precedence (user-Warp Warp customizations > shell bindkeys > Warp defaults), surface boundaries (shell input editor only — not the AI prompt or palettes), settings/feature-flag, and performance.
	•	tech.md — implementation plan grounded in current code: bootstrap-side query into a new DProtoHook::ShellBindings DCS variant (gated by a per-tab WARP_BOOTSTRAP_NONCE plus size caps and strict schema validation so arbitrary process output can’t spoof bindings), per-tab ShellBindings storage on Shell, expanded InputAction with fine-grained ZLE/readline verbs, a third shell_bindings tier in Keymap, in-app vi-mode state machine driven by dispatched widgets (with shell-reported mode as initial state and resync), settings toggle + FeatureFlag::HonorShellBindkeys for staged rollout, and telemetry with widget-name redaction (allowlist of well-known shell vocabulary; user-defined names bucket to user-defined).
Open questions deliberately surfaced for maintainer input rather than guessed: forwarding user-defined widget bodies back to the shell (deferred), AI prompt opt-in, and default-on vs feature-flagged rollout.

## Linked Issue

Closes #537

## Agent Mode


th1nkful:
@cla-bot check

th1nkful:
@oss-maintainers — bringing this in for a maintainer pass per CONTRIBUTING.md. Have used the three /oz-review rounds and would appreciate a human review on the spec.

Recap of the rounds:
	•	Round 1 (b41c0f8) — 6 important comments: shell input-queue semantics for string macros, telemetry redaction policy, DCS payload trust boundary, shell-side re-query mechanism, fish bootstrap merge with user fish_key_bindings, in-app vi-mode state machine.
	•	Round 2 (7a2985b) — 5 important comments: nonce env-var threat model + non-exported shell-local handling, Macro(String) injection path through key resolution, toggle-on cached reuse vs re-query, bindkey -l for user-defined zsh keymaps, fish reserved-keys exception.
	•	Round 3 (740f8e7) — 5 important comments: relaxed PRODUCT #4 to align with async precmd timing, listed reserved infrastructure keys per shell (zsh/bash also have them, not just fish), fish-specific tempfile nonce path (config.fish runs before --init-command), moved ShellBinding/ShellTabBinding types out of warpui_core into the terminal layer, fixed toggle test expectations.
	•	Round 4 (9269119) — 4 important comments: diagnostic redaction matches telemetry; precmd hash state-preservation discipline ($?, set -o, KEYMAP, aliases, traps) with an integration test; replaced ContextPredicate::TabIs (which doesn’t exist — predicates are &'static str only) with a storage-tier ScopeKey API; explicit MatchOutcome::AbandonedPrefix matcher change for PRODUCT #8 prefix replay.

The spec is now grounded in real file refs (crates/warpui_core/src/keymap/<context.rs:10>-17, <matcher.rs:258>+, app/src/terminal/local_tty/<shell.rs:632>, app/assets/bundled/bootstrap/{zsh_body,bash_body,fish}.sh). Open questions (rollout default-on vs flag, AI prompt opt-in, user-defined widget forwarding) are explicit and tagged for maintainer input.

specs/GH537/product.md
```diff
@@ -0,0 +1,383 @@
...
+   driven shell-side at every `precmd`, so the change is detected when
+   the prompt next redraws. The user-visible invariant: a binding
+   declared at the shell prompt is honored starting with the first
+   keystroke after Warp has parsed the next `ShellBindings` payload
+   from that prompt. Keystrokes pressed during the small async window
+   between the prompt firing and the payload being parsed use the
+   previous keymap (consistent with the non-blocking guarantee in #26);
+   declarations never block typing.
+
+5. Each tab tracks its own bindings independently. Changing bindings in one
```
acarl005, line 136:
```suggestion
5. Each session/pane tracks its own bindings independently. Changing bindings in one
```

A Warp "tab" can contain multiple sessions or panes. It would be more exact to say this is per-pane or per-session state.

specs/GH537/product.md
```diff
@@ -0,0 +1,383 @@
...
+- Discovery via the user's live shell session (querying the shell for its
+  current binding table), so dynamic and conditionally-declared bindings are
+  picked up — not by parsing rc files.
+- Best-effort coverage of the action set: any shell widget / readline function
+  / fish input function that has a Warp-input equivalent is honored. Widgets
+  with no clean Warp equivalent degrade gracefully (see below) rather than
+  silently stealing the keystroke.
+- Keymap modes: emacs vs vi (insert/command/visual) for the shells that
+  expose them. Mode switches initiated by the user (e.g. `bindkey -v`,
+  `set -o vi`, vi-mode plugins, fish bind modes) take effect without restart.
```
acarl005, line 33: I would put this as very low priority. Mutating your keybindings mid-session seems like a very rare case. It's fine if that works without much additional complexity, but optimizing for this case isn't worth a lot of effort IMO. We should focus on what more users are requesting, e.g. `fzf`.

Furthermore, Warp has its own vim keybinding implementation (which IMO is quite good) so I don't find this specific example to be very compeling.

specs/GH537/product.md
```diff
@@ -0,0 +1,383 @@
...
+      diagnostic. Telemetry records unsupported-widget hits under the
+      same rule; user-defined or otherwise unknown widget names are reported as
+      the bucket `user-defined` with no further identifying information,
+      since user-defined widget names can be arbitrary or private. Key
+      contents, key sequences, and binding bodies are never recorded.
+    - **Open question:** for user-defined shell-function widgets (e.g. zsh
+      `zle -N my-widget; bindkey '^X' my-widget`), v1 treats these as
+      unsupported. A future iteration could forward the keystroke to the
+      shell to let it execute the widget. Confirm v1 = unsupported is
+      acceptable.
```
acarl005: How would `fzf-history-widget` be treated? As a user-defined widget? `fzf` is the primary reason people request bindkey so I don't think it makes sense to proceed without supporting `fzf`.

specs/GH537/product.md
```diff
@@ -0,0 +1,383 @@
...
+- Best-effort coverage of the action set: any shell widget / readline function
+  / fish input function that has a Warp-input equivalent is honored. Widgets
+  with no clean Warp equivalent degrade gracefully (see below) rather than
+  silently stealing the keystroke.
+- Keymap modes: emacs vs vi (insert/command/visual) for the shells that
+  expose them. Mode switches initiated by the user (e.g. `bindkey -v`,
+  `set -o vi`, vi-mode plugins, fish bind modes) take effect without restart.
+- Conflict policy with Warp's own keybindings (see Behavior #14).
+
+Out of scope for this spec:
```
acarl005, line 36: What about PowerShell? It's not blocking if we mark that out-of-scope, but we should explicitly mention it.

specs/GH537/product.md
```diff
@@ -0,0 +1,383 @@
...
+      `vi-up-line-or-history`, `vi-down-line-or-history`, `vi-goto-mark`,
+      `vi-set-mark`, `vi-replace-chars`, `vi-substitute`,
+      `vi-change-whole-line`, plus fish-mode equivalents.
+    - Completion: `complete-word`, `expand-or-complete`,
+      `expand-or-complete-prefix`, `menu-complete`, `reverse-menu-complete`.
+      These trigger Warp's existing completion UI (not the shell's), but
+      from whichever key the user has bound them to. Behavior of the
+      completion UI itself is unchanged by this spec.
+
+11. Widgets that have no Warp equivalent (examples: `quoted-insert` in some
```
acarl005: The stated goal is to honor "user-defined keybindings". However, I'm not convinced this proposal actually solves that pain point. Let's take a simple example of a user-defined keybinding (kill-line but it copies to the system clipboard):

```zsh
function kill-line-to-clipboard() {
  # Kill the line (zsh built-in)
  zle kill-line
  # Copy the killed text to clipboard
  echo -n "$CUTBUFFER" | pbcopy
}

zle -N kill-line-to-clipboard
bindkey '^k' kill-line-to-clipboard
```

In #10 you have proposed essentially an "allowlist" of known line editor functionality that Warp is able to reproduce. This custom widget `kill-line-to-clipboard` won't be in that list. Here in #11 you state that this is not something Warp can introspect or fulfill. If Warp can't perform this, how are we solving the original pain point?

specs/GH537/product.md
```diff
@@ -0,0 +1,383 @@
...
+    - Any modal overlay rendered above the input editor (file palette,
+      command palette, suggestions popover focus, etc.).
+
+21. Switching focus from the input editor to another surface and back does
+    not require re-querying. The binding table from the most recent query
+    remains valid for the duration of the tab unless invalidated by #4.
+
+### AI / agent prompt input
+
+22. The AI prompt input editor does not honor shell bindkeys by default —
```
acarl005, line 577: How would this work with our "natural language auto-detection" feature? In agent conversations, we run a custom model on the input on every keypress to classify whether the current buffer is a shell command or natural language. Sometimes this flickers between the 2 as the user is typing. If a shell command is misclassified as natural language it would disable the shell bindkeys which would be frustrating.

acarl005 (changes requested):
Thanks for working on this. It's one of the top requested Warp features. It's also very complex, so this won't be easy to align on.

I haven't looked at the tech spec yet since I think there are blocking issues with the product spec. I want to emphasize that the most important thing is to make sure we solve common user pain points. What most users actually want when this issue comes up is:

1. Integration with popular libraries, mostly `fzf`, but also `atuin`, copilot CLI, and others.
2. Their custom line editor bindings/widgets to work.

My understanding of this product spec is that neither can actually be supported. I left comments and questions inline about those. 

I do like this idea of the keybinding precedence chain. I also like the idea of introspecting the shell for its keymaps. I'm just not sure about only supporting a small list of known widgets.

vkodithala:
@lucieleblanc's handling this, she mentioned she'll get to it later this wk! In the meantime, @th1nkful mind re-running /oz-review and ensuring that those issues are fixed up?

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+inline. See #11.6.
+
+### Discovery and lifecycle
+
+1. When a Warp tab starts a supported shell (zsh, bash, fish), Warp queries
+   that shell for its current keybinding table once the shell is ready to
+   accept commands but before the first user keystroke is processed by the
+   input editor. Until the table arrives, the input editor uses Warp's
+   default keymap; once the table arrives, user bindings take effect on the
+   next keystroke.
```
lucieleblanc, line 105: I'm curious what this intermediate stage would look like. Do we want a way to signal to users that their keybindings have loaded? Is the current "starting <shell name>..." loading text in the terminal output during bootstrap enough?

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+
+3. If the shell fails to start, exits before the query completes, or returns
+   an unparseable response, Warp logs a diagnostic and falls back to its
+   default keymap for that tab. The tab remains usable; no user-facing error
+   toast is required.
+
+4. When the user changes their bindings inside an existing session
+   (`bindkey '^X^E' edit-command-line`, `bind '"\C-x\C-e": edit-and-execute-command'`,
+   sourcing a new rc file mid-session, switching emacs/vi mode), Warp picks
+   up the change without requiring a restart of the tab. Discovery is
```
lucieleblanc, line 126: Echoing @acarl005's comment above, I'm not sure this is a priority. Picking up changes from `precmd` is clever, but I'm concerned this would be a lot of overhead to add to every block.

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+   user shell bindings for the active keymap → Warp's default
+   keybindings → default character insertion). Shell bindings are
+   consulted only after the two higher tiers have been checked and have
+   not produced a match. When the matched action is a shell binding and
+   the bound widget has a Warp equivalent, Warp performs that action
+   and consumes the keystroke. When the bound widget is unsupported
+   (#11), the keystroke continues down the ladder to Warp's defaults.
+
+8. Multi-key sequences (`^X^E`, `^[f`, `gg`, fish `\\cx\\ce`) are honored
+   as a single action. Resolution rules:
```
lucieleblanc, line 157: Supporting keyboard shortcut sequences in Warp would be awesome. We have a state machine that handles this in Vim mode, but don't currently have that set up for the app as a whole. Do we need to prioritize supporting this? (As someone who doesn't use sequences outside of Vim emulation, I am curious how useful this feature would be without sequences.)

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+     complete binding (e.g. partial `^X` of `^X^E`), pending keys are
+     held without timeout — readline / ZLE both wait indefinitely on
+     pure prefixes. The user pressing any non-extending key abandons
+     the prefix immediately (next rule).
+   - **Abandonment.** When a non-matching key arrives mid-sequence,
+     the prefix is abandoned: Warp replays the buffered keys plus the
+     just-received key through normal handling, in arrival order. Any
+     of those replayed keys may itself trigger a single-key binding;
+     none of them re-enter prefix accumulation until the replay
+     finishes. This matches readline / ZLE behavior.
```
lucieleblanc, line 180: Opinion of one: I'm surprised this is what `readline` and `ZLE` do -- the sudden replay seems much more surprising and potentially error-prone than dropping invalid sequences the way `vim` does. I would definitely prefer the Vim-style abandonment with no replay.

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+   - **No keystroke is ever silently dropped.** Either a binding fires,
+     or the buffered keys are replayed.
+   - **Focus loss / window blur** mid-sequence abandons the prefix
+     (replay path); on refocus, accumulation starts fresh.
```
lucieleblanc, line 184: Agreed on this.

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+   and consumes the keystroke. When the bound widget is unsupported
+   (#11), the keystroke continues down the ladder to Warp's defaults.
+
+8. Multi-key sequences (`^X^E`, `^[f`, `gg`, fish `\\cx\\ce`) are honored
+   as a single action. Resolution rules:
+
+   - **Mid-sequence buffering.** While Warp has received one or more
+     keys that match a prefix of a longer binding but not yet a complete
+     binding, no action fires and no character is inserted; Warp waits
+     for the next key.
```
lucieleblanc, line 162: Do we want to show anything to the user when this is happening? Warp's Vim mode has a small status icon and buffer text to show pending keystrokes; is that something we'd want to reuse here?

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+      is synced back into Warp's input editor and the user is
+      returned to block-mode editing. The cursor position the widget
+      left behind is preserved.
+    - If the widget calls `accept-line` (i.e. submits the command
+      itself, as some atuin configurations do), Warp treats the
+      submission the same as if the user had pressed Enter in
+      block mode — the resulting command is run as a Warp block.
+    - The widget's stderr / stdout (anything it writes outside its
+      alt-screen) renders as terminal output, like any other
+      command. It does not appear in Warp's input editor.
```
lucieleblanc, line 309: When I read this, I'm assuming the widget's stdout/stderr will appear in a background output section, not as a full-fledged command block, since there's no associated prompt or input line. Is that right?

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+    fast-syntax-highlighting, fish abbreviations, zsh-vi-mode's
+    visual mode indicators), Warp honors them. Concretely, while the
+    user types in the input editor on a tab where these plugins are
+    active:
+
+    - **Inline suggestions appear.** If zsh-autosuggestions or an
+      equivalent is loaded and would have suggested a completion at
+      the current buffer state, that suggestion is visible in the
+      input editor in dimmed text after the cursor, exactly as it
+      would render in the user's terminal without Warp.
```
lucieleblanc, line 352: Does this mean that Warp's own command autosuggestions won't appear at all if the shell's autosuggestions are enabled? We should add a note under the corresponding Warp setting or some documentation around the override behavior here.

Beyond the terminal input mode, how would this interact with natural language autosuggestions, which the shell can't provide?

specs/GH537/product.md
```diff
@@ -0,0 +1,755 @@
...
+    These match the keys Warp's existing bootstrap already installs in
+    each shell. Lifting the exception (re-implementing each integration
+    point without bind-level interception) is a tracked follow-up; the
+    integrations exist today and replacing them is out of scope here.
+
+15. When a user shell binding shadows a Warp default, no warning, banner, or
+    toast appears. The user already declared this binding in their shell
+    config; the override is the desired behavior. Diagnostics for shadowed
+    Warp defaults may be available through verbose logging but are not
+    user-facing.
```
lucieleblanc, line 530: I don't completely agree with silent overrides. I think it'd be good to surface some kind of warning in the keybindings settings page, so that users are aware that whatever keybindings they set there won't be respected in the terminal input editor. Even a small note like "this keybinding is currently being overridden by your shell (`zsh`) configuration" with a little info icon ( :information_source: ) when the user uses a binding that's overridden could prevent major confusion.

lucieleblanc (changes requested):
Exciting and very ambitious feature, with lots of details and decisions to iron out.

lucieleblanc:
@th1nkful Thank you for the specs! This product spec has lots of details and design decisions we’ll need to consider when building this feature.

We've discussed this proposal internally. The keybinding system needs work, and this will be a major technical undertaking whether it's part of a refactor or not. We’ll probably want to start with a proof-of-concept (small working demo/screen recording) of the shell integration piece, then use that to refine the product spec + write the tech spec. If you get a demo working, we’d love to see it. Otherwise, we’re planning to tackle this internally.

I’m going to close this PR for now — thank you again!

PR #9848, opened 2026-05-01
spec: user-configurable language servers (#8803)

lonexreb:
## Description

Spec PR for #8803 ("Add support for adding custom language servers"), per @kevinyang372's direction on #9562 and #9568 — closing those PRs in favor of a user-configurable LSP path. This converts that product direction into testable behavior invariants plus a concrete implementation plan grounded in the existing `crates/lsp/` infrastructure.

## Linked Issue

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. (#8803 is `ready-to-spec`.)

## What's in this PR

| File | Lines | What |
|---|---|---|
| `specs/GH8803/product.md` | ~140 | 15 numbered testable invariants. TOML config shape (`[[lsp.servers]]` array-of-tables with `name` / `command` / `file_types` / `root_files` / `initialization_options` / `start_timeout_ms`). Explicit non-goals (auto-install, version pinning, marketplace). User experience walkthrough (add → chip → enable → use → disable). Open questions (`~` expansion, example-configs doc, network defaults). |
| `specs/GH8803/tech.md` | ~270 | File-by-file implementation plan. New `UserConfiguredLanguageServer` impl of `LanguageServerCandidate`. New `LSPServerType::UserConfigured(...)` variant. Per-workspace `enabled_custom_servers` / `dismissed_custom_servers` state. Settings-reload diff semantics. Per-invariant test plan. Risks called out. |

## Reuses lessons from the closed LSP PRs

This spec deliberately leans on patterns that survived bot review on #9562 and #9568, so the implementation phase doesn't relitigate them:

- **`binary_in_path` PATH search with executable-bit check on Unix** — from #9562's `crates/lsp/src/servers/intelephense.rs`. Used here as `is_installed_on_path`.
- **`warpui::r#async::FutureExt::with_timeout` for bounded `initialize`** — from #9562's follow-up commit `31285c4`. Used here as the `start_timeout_ms` enforcement path.
- **`std::env::join_paths` + `binary_filename` for cross-platform tests** — from #9562/#9568's test suites. Reused for the new `user_configured_tests.rs`.
- **`initialization_options` as the security-knob surface** — from #9568's `handledSchemaProtocols` discovery. Documented as the recommended pattern for user-managed network-fetching restrictions.

## Scope discipline

- **Not implementing** — this is spec-only, no Rust code yet. Implementation will land as a follow-up PR after spec approval.
- **Doesn't replace built-in servers** — user-configured servers run alongside any built-in server matching the same file extension. The LSP client merges responses (existing behavior).
- **Doesn't auto-install** — the user owns binary lifecycle for these (npm, brew, cargo, etc.). Warp's job is to discover, prompt, and connect.
- **Per-workspace enablement** — defining a server is global; running it requires explicit opt-in per workspace via the editor footer chip.

## Open questions for maintainers (also documented in product.md)

1. **`~` expansion in `command[0]`** — Cmd-O / `/open-file` already use `shellexpand::tilde`; recommend yes for `~`, defer `$VAR` to a follow-up. Confirm?
2. **Example configs doc** — recommend shipping `docs/custom-lsp-examples.md` (intelephense, lua-language-server, zls, bash-language-server) alongside the implementation. Worth gating behind this spec or a separate docs PR?
3. **Schema-fetching restriction defaults** — should we mandate / document that user-configured servers run with their network-fetching options disabled by default? Hard to enforce generically (each server has its own keys); spec recommends per-server `initialization_options` and an examples doc rather than a generic flag. Confirm?

## Testing — for the spec itself

- Both files render correctly on GitHub.
- All 15 product invariants have a corresponding row in tech.md's testing matrix.
- Cross-references to closed PR commits (`31285c4`, etc.) and existing files (`crates/lsp/src/servers/intelephense.rs`, `app/src/settings/input.rs`) verified at write time.

## Agent Mode

CHANGELOG-NEW-FEATURE: Spec for user-configurable language servers (#8803) — defines the TOML config shape, per-workspace enablement model, and lifecycle for custom LSPs that aren't shipped built-in. Implementation PR to follow. Thanks @lonexreb!

### DIFF
--- a/specs/GH8803/product.md
+++ b/specs/GH8803/product.md
@@ -5,7 +5,7 @@
 
 ## Summary
 
-Let users add language servers Warp does not ship out of the box by declaring them in a settings file (binary path, arguments, file types they apply to, optional root-file globs, optional initialization options). When the user opens a file matching a configured server's file types, Warp offers to enable that server for the workspace and — once enabled — uses it for code intelligence (diagnostics, hover, goto, completions) on that file type, alongside any built-in servers.
+Let users add language servers Warp does not ship out of the box by declaring them in a settings file (binary path, arguments, file types, LSP language identifier). When the user opens a file matching a configured server's file types, Warp offers to enable that server for the workspace. Once enabled, the user-configured server **takes over** code intelligence (diagnostics, hover, goto, completions) for those file types in that workspace, replacing any built-in server that would otherwise handle them.
 
 This is the contributor-facing alternative to baking each language server into the Warp binary. It directly addresses why PRs adding specific built-in LSPs (e.g. PHP Intelephense in #9562, JSON in #9568) were closed in favor of this product direction.
 
@@ -23,8 +23,8 @@ This scales poorly and pulls language-specific install logic into the core. User
 ## Goals
 
 - A user can add a language server to Warp by editing a settings file — no binary modifications, no agent involvement.
-- Configured servers run side-by-side with built-in servers and follow the same code-intelligence surface (diagnostics, hover, goto, completions, semantic tokens).
-- Configuration is portable across workspaces and discoverable in Warp's settings UI.
+- A user-configured server **supersedes** any built-in server for the file extensions it declares (in workspaces where it is enabled). Two servers do not run for the same file type in the same workspace.
+- Configuration is portable across workspaces and discoverable in Warp's existing Settings UI.
 - Enablement is per-workspace by default with a clear opt-in moment, not silent global activation.
 - Mis-configuration produces a visible, actionable error — not a silent disabled state.
 
@@ -32,10 +32,11 @@ This scales poorly and pulls language-specific install logic into the core. User
 
 - **Auto-installation of user-configured servers.** The user installs the binary themselves (npm, cargo, system package manager, brew, etc.). Warp does not run package managers on the user's behalf for these.
 - **Version pinning / auto-update of user-configured servers.** Out of scope; the user owns the lifecycle.
-- **Per-file dynamic switching.** A single file is associated with at most one user-configured server, plus any built-in servers that already match. No "try server A, fall back to server B" runtime logic.
-- **Replacing built-in servers.** A user-configured server with the same `file_types` as a built-in does **not** disable the built-in. Both run; the LSP client merges their results.
+- **Per-file dynamic switching.** Within a workspace, a single file extension is associated with at most one user-configured server. No "try server A, fall back to server B" runtime logic.
+- **Coexisting with a built-in server on the same file extension.** When a user-configured server is enabled in workspace W and matches file type X, the built-in server that would otherwise serve X in W is **not** spawned for W (per @kevinyang372's review of the original draft — running both is undesirable).
 - **Cross-workspace global enablement on first open.** A configured server is *defined* globally but *enabled* per-workspace.
 - **Marketplace / discovery of community configs.** Out of scope; users find configs themselves.
+- **Per-server initialization options (V0).** `initializationOptions` forwarding is deferred to a follow-up; warp does not yet have settings-file shape for arbitrary nested config payloads.
 
 ## User experience
 
@@ -43,7 +44,7 @@ This scales poorly and pulls language-specific install logic into the core. User
 
 1. User edits their Warp settings file (TOML, located at the standard Warp settings path) and adds an `[[lsp.servers]]` entry. (See the "Configuration shape" section below.)
 2. Warp detects the new entry on settings reload. No restart required.
-3. If the entry is malformed (missing `name`, missing `command`, empty `file_types`), Warp shows a non-blocking notification: *"Custom LSP `<name>` is misconfigured: <reason>. See settings."* with a button to open the settings file at the offending line.
+3. If the entry is malformed (missing `name`, missing `command`, empty `file_types`, missing `language_id`, duplicate `name`, or duplicate `file_types` across user-configured entries), Warp shows a non-blocking notification: *"Custom LSP `<name>` is misconfigured: <reason>. See settings."* with a button to open the settings file at the offending line.
 
 ### First time opening a matching file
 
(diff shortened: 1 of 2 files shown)


specs/GH8803/product.md
```diff
@@ -0,0 +1,125 @@
...
+- Configuration is portable across workspaces and discoverable in Warp's settings UI.
+- Enablement is per-workspace by default with a clear opt-in moment, not silent global activation.
+- Mis-configuration produces a visible, actionable error — not a silent disabled state.
+
+## Non-goals
+
+- **Auto-installation of user-configured servers.** The user installs the binary themselves (npm, cargo, system package manager, brew, etc.). Warp does not run package managers on the user's behalf for these.
+- **Version pinning / auto-update of user-configured servers.** Out of scope; the user owns the lifecycle.
+- **Per-file dynamic switching.** A single file is associated with at most one user-configured server, plus any built-in servers that already match. No "try server A, fall back to server B" runtime logic.
+- **Replacing built-in servers.** A user-configured server with the same `file_types` as a built-in does **not** disable the built-in. Both run; the LSP client merges their results.
```
kevinyang372: User-configured server should overwrite built-in servers. It doesn't make sense to have both servers running at the same time

specs/GH8803/product.md
```diff
@@ -0,0 +1,125 @@
...
+5. If `Dismiss` is pressed, the chip is suppressed for this workspace until the user re-opens it from settings.
+
+### Subsequent opens in an enabled workspace
+
+1. User opens any file matching `file_types` in a workspace where the server is already enabled.
+2. The server is already running; no UI surfaces. Diagnostics, hover, goto, completions all behave as they do for built-in servers.
+
+### Disabling a server in a workspace
+
+1. User opens settings → "Code intelligence" → "Custom language servers".
```
kevinyang372: "Code intelligence" doesn't exist in our settings

To start, I also think it's probably fine to not have a dedicated custom language servers section in UI

specs/GH8803/product.md
```diff
@@ -0,0 +1,125 @@
...
+2. Each configured server lists which workspaces it is enabled for.
+3. User clicks the workspace row's `Disable` button. Warp shuts down that server's process for that workspace and removes the per-workspace enablement record.
```
kevinyang372: Please make sure we are following the existing settings > code > indexing and projects UI

specs/GH8803/product.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Misconfiguration scenarios
+
+1. **Binary not on PATH:** When a user enables a server whose `command[0]` is not on PATH, Warp shows: *"Could not start `<name>`: binary `<cmd>` not found on PATH."* The chip's `Enable` button is replaced with `Open settings`.
+2. **Binary on PATH but spawn fails:** Warp shows: *"`<name>` exited with status `<n>`. Last 200 bytes of stderr: `<...>`."* with `Open settings` and `Retry` buttons.
+3. **Spawn hangs:** A configurable `start_timeout` (default 5s) bounds the LSP `initialize` request. On timeout, Warp shows: *"`<name>` did not respond to `initialize` within 5s."* The server's process is killed.
```
kevinyang372: I don't think we need to do anything special for custom LSPs here

specs/GH8803/product.md
```diff
@@ -0,0 +1,125 @@
...
+
+The user's Warp settings TOML grows a new `[lsp]` table. Multiple servers via array-of-tables `[[lsp.servers]]`:
+
+```toml
+[[lsp.servers]]
+name = "intelephense"
+command = ["intelephense", "--stdio"]
+file_types = ["php", "phtml"]
+root_files = ["composer.json", "composer.lock"]      # optional; default: file's parent dir
+initialization_options = { storagePath = "/tmp/intelephense" }   # optional, opaque to Warp
```
kevinyang372: We don't have support for these configs yet. Fine to skip for V0

specs/GH8803/product.md
```diff
@@ -0,0 +1,125 @@
...
+## Configuration shape
+
+The user's Warp settings TOML grows a new `[lsp]` table. Multiple servers via array-of-tables `[[lsp.servers]]`:
+
+```toml
+[[lsp.servers]]
+name = "intelephense"
+command = ["intelephense", "--stdio"]
+file_types = ["php", "phtml"]
+root_files = ["composer.json", "composer.lock"]      # optional; default: file's parent dir
```
kevinyang372: We should use our current root repo detection logic

specs/GH8803/product.md
```diff
@@ -0,0 +1,125 @@
...
+initialization_options = { storagePath = "/tmp/intelephense" }   # optional, opaque to Warp
+start_timeout_ms = 5000                                          # optional, default 5000
```
kevinyang372: Can remove this

kevinyang372:
Thanks for writing this up! Left some comments on the product spec

lonexreb:
Thanks for the thorough reviews! Pushed [`4f85a49`](https://github.com/warpdotdev/warp/pull/9848/commits/4f85a49) revising both specs to address every comment. Summary of how each was resolved:

### @kevinyang372 — maintainer comments

| Comment | Resolution |
|---|---|
| User-configured servers should overwrite built-ins (product.md:36) | Reframed as a Goal; new invariants 8 and 9 specify built-in suppression in the enabled workspace + restoration on disable. Tech spec adds `handled_extensions()` to the built-in trait and a suppression filter in `all_candidates_for(workspace)`. |
| "Code intelligence" section doesn't exist (product.md:66) | Removed. Disable flow now lives under Settings → Code → Indexing/Projects per your direction. |
| Follow existing Settings → Code → Indexing/Projects pattern (product.md:68) | Tech spec §8 is now "Settings UI (in scope for V0)" and explicitly extends that surface — no new section introduced. |
| No special timeout handling needed for custom LSPs (product.md:74) | `start_timeout_ms` is no longer a config field; a fixed 5s default lives in code. |
| Use existing root repo detection (product.md:85) | `root_files` removed from V0 config. `should_suggest_for_repo` defers to Warp's existing root-repo detection. |
| Skip `initialization_options` for V0 (product.md:86) | Field removed from config + invariants; tracked under Follow-ups. |
| Remove `start_timeout_ms` (product.md:87) | Done. |

### oz-for-oss findings

| Finding | Resolution |
|---|---|
| 🚨 Settings UI cannot be follow-up (tech.md:279) | Moved into V0 as Tech §8, hooked into the existing Settings → Code surface per @kevinyang372. Invariants 11/12/14 are now reachable. |
| ⚠️ Crate boundary violation for `UserLspServerConfig` (tech.md:87) | Type now defined in `crates/lsp/src/user_config.rs` and re-exported; `app/src/settings/lsp.rs` imports it. New "Crate boundaries" section in tech.md states the rule explicitly. |
| ⚠️ File extension insufficient as `languageId` (tech.md:164) | New required `language_id` field on every `[[lsp.servers]]` entry. The closed `LanguageId` enum is no longer extended; the `didOpen` dispatch site branches on candidate type. New invariant 15 covers forwarding. |
| ⚠️ [SECURITY] auto-restart on `command` change (tech.md:191) | `enabled_custom_servers` is now a `HashMap<name, command_fingerprint>` keyed on the user's last confirmation. Live ≠ stored fingerprint demotes the server to "pending re-confirmation"; chip re-appears with a "command changed — re-enable?" affordance; the new binary is **never** spawned without explicit user click. New invariant 11 covers this. |
| ⚠️ `Drop` of `Child` not a portable kill (tech.md:206) | Switched to `tokio::process::Command` with `kill_on_drop(true)` plus explicit `child.kill().await; child.wait().await;` on timeout, JSON-RPC error, and shutdown paths. |
| ⚠️ Duplicate `file_types` resolution (product.md:35) | With overwrite semantics, overlapping `file_types` is now a parse-time error. Cross-entry validation rejects the entire `[lsp]` block; invariant 2 covers it. |
| 💡 `[[lsp.servers]]` ↔ `custom_servers` serde rename (tech.md:37) | Documented in tech §2 with the explicit `#[serde(rename = "servers")]` mapping. |

### Net spec shape

V0 config is now minimal:

```toml
[[lsp.servers]]
name = "intelephense"
command = ["intelephense", "--stdio"]
file_types = ["php", "phtml"]
language_id = "php"
```

Four required fields, no optional V0 fields. `root_files`, `initialization_options`, `start_timeout_ms`, and `~`/`$VAR` expansion are all tracked under Follow-ups.

Re-requesting review when you have a chance — happy to iterate further.

lonexreb:
Quick heads-up on commit [`140434e`](https://github.com/warpdotdev/warp/pull/9848/commits/140434e) — this is a **self-review pass only**, not a response to additional feedback. Found four spec-internal inconsistencies in `4f85a49` and fixed them proactively:

1. Settings UI section only described `Enabled → Disable` rows, but invariants 11/12/14 also require Settings UI affordances for `Dismissed → Re-enable` and `Pending re-confirmation → Re-enable`. Expanded into a 3-state table.
2. Disable button silently broke invariant 12 ("chip reappears on next file open") when the server had been previously dismissed. Spelled out that Disable clears both `enabled_custom_servers[name]` and any `dismissed_custom_servers` entry.
3. Chip-state cascade ordering was inconsistent between §7 (alternatives) and the end-to-end flow (`binary not on PATH` placed last, contradicting product misconfig 1). Reordered to: command-changed (security override) → dismissed → binary missing → Enable default → already-running.
4. The "command-changed chip overrides dismissal" behavior was implicit in the cascade ordering but not in any invariant. Tightened invariant 11 to state the security override explicitly.

Net diff: +26 / -18. No design semantics changed; same 4-field config, same crate boundary, same kill/wait semantics, same 16 invariants (with #11 made more precise).

dor-tzur-lmnd:
Can we move this one? It's a critical feature missing

Issue #6665, opened 2025-06-25
Linux: Launching Warp through "Open in Terminal" in context menu opens it in $home path instead of context path


phenomen:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [x] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/support-and-billing/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When I launch  Warp through "Open in Terminal" in context menu, it opens in $home directory path instead of the one where it was invoked.

### To reproduce

1. Set Warp as a Preferred Application for Terminal
2. Open some directory
3. Right Click -> Open in Terminal

### Expected behavior

Warp opens in the same path where the context menu was invoked.

### Operating system (OS)

Linux

### Operating system and version

Mint 22.1

### Shell Version

GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)

### Current Warp version

v0.2025.06.20.22.47.stable_05

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

Yes, this issue prevents me from using Warp daily.

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

LadyNay:
Hi there, having the same issue. Windows 11, In the directory (or right clicking on the folder), Right clicking to access the context menu, Select Open in New Tab or Open in new Window does not open the directory in Warp but the Home ~.

Currently using the latest version from the website. The version before this didn't have the issue. New Windows install, using bash as the session default.

Issue #6143, opened 2025-03-18
Windows 10: "Open Warp Here" opens 2 instances.


phy0bab:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [x] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/help/sending-us-feedback#gathering-warp-logs)

### Describe the bug

If Warp is not already open, using the right click context menu in Windows file explorer to "Open Warp Here", 2 instances of Warp are opened.
[warp-logs.zip](https://github.com/user-attachments/files/19319822/warp-logs.zip)

### To reproduce

1. Ensure no Warp instance is already open.
2. Right click in a folder location using Windows file explorer and choose "Open Warp Here".

### Operating system (OS)

Select an OS

### Operating system and version

Windows 10

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

tipani86:
I have a close relative issue to this one when launching from taskbar shortcut, also spawns two windows. Mentioned this issue in my own bug report too.

acarl005:
Related to #6004

PR #9849, opened 2026-05-01
fix(windows): context menu "Open Warp in new tab" navigates to home dir (#9844)

### LINKED ISSUE #9844: Bug: 'Open Warp in New Tab' from Explorer context menu doesn't navigate to current directory on Windows
### Pre-submit Checks

- [x] I have searched Warp bugs and there are no exact duplicates

### Describe the bug

When right-clicking in Windows File Explorer and selecting **"Open Warp in new tab"**, the new Warp tab opens at `~` (USERPROFILE / home directory) instead of the directory where the right-click occurred.

This appears to be the **Windows counterpart** of issue #6665 (which reports the same behavior on Linux via "Open in Terminal" context menu).

### To reproduce

1. Open any folder in Windows File Explorer (e.g. `C:\Projects\my-app`)
2. Right-click on empty space in the folder
3. Select **"Open Warp in new tab"** from the context menu
4. Warp opens a new tab, but `pwd` is `~` (e.g. `C:\Users\<username>`) instead of `C:\Projects\my-app`

### Expected behavior

Warp should open the new tab with the working directory set to the folder where the right-click context menu was invoked.

### Actual behavior

Warp opens at `USERPROFILE` (e.g. `C:\Users\<username>`), ignoring the directory context entirely.

### Operating system (OS)

Windows

Windows 10 (Build 26200.4061)

### Current Warp version

Latest stable (as of May 2025)

### Regression

No, this bug has existed throughout my experience using Warp on Windows.

### Technical analysis

Based on a review of the Warp source code, here is a likely root cause:

1. **URI handler parsing**: The `on_open_urls` handler parses incoming URLs via `Url::parse()` and `extract_filepath()`. For `file://` URIs with no host component, it attempts to prepend a drive letter by scanning for the first `:` or `\`.

2. **Shell extension format mismatch**: The Windows shell extension likely passes **bare Windows paths** (e.g. `C:\Users\user\folder`) rather than properly formatted `file:///` URIs. When `Url::parse()` receives a bare path like `C:\Users\user\folder`, it fails because the `C:` is interpreted as a scheme rather than a drive letter.

3. **Silent fallback**: When URL parsing fails, the path is silently dropped. The window spawn logic then falls back to `std::env::var_os("USERPROFILE")` when `initial_directory` is `None`, resulting in the home directory being used.

4. **Possible mitigation**: The `warp://warp/new_from_path` endpoint exists for shell integration and may be the intended mechanism, but it appears the shell extension's context menu action doesn't invoke it correctly (or at all).

**Suggested investigation areas:**
- Check how the Windows shell extension serializes the directory path when invoking "Open Warp in new tab"
- Verify whether `Url::parse()` handles bare Windows paths (with drive letters) correctly
- Consider adding a fallback path parser that handles bare `C:\...` paths when `Url::parse()` fails
- Compare the "Open Warp Here" flow (#6143) vs "Open Warp in new tab" to see if they share the same path-passing mechanism

### Additional context

- Related issue: #6665 (same bug on Linux — context menu "Open in Terminal" opens at `$HOME`)
- Related issue: #6143 (Windows "Open Warp Here" opens 2 instances — different bug but same shell extension surface)
- This is filed as a bug report; full investigation likely requires a Windows development environment to trace the shell extension → URI handler flow

KuaaMU:
## Summary

- `parse_tab_path` trims trailing whitespace from the decoded path, catching an edge-case where `%1`/`%V` expansion appends invisible characters
- `path_if_directory` falls back to `std::fs::canonicalize()` when `is_dir()` returns false, resolving symlinks, `.`/`..` segments, and other minor path normalization issues before giving up
- `on_open_urls` has a Windows-only fallback: if `Url::parse()` fails on a bare `C:\...` path, it converts to a `file:///` URL before retrying

## Root cause

The Windows Explorer context menu passes raw filesystem paths via `%1`/`%V` without URL-encoding them into the `warp://` URI query string. While the `url` crate handles bare backslashes and drive-letter colons correctly in most cases, the resulting `PathBuf` can still fail the `is_dir()` gate in `path_if_directory()` — for example when the OS needs a canonicalized form (symlinks, trailing separators, `\?\` prefix, etc.). When that happens the initial directory is silently dropped and the new tab falls back to `~`.

Verified by running actual Rust tests:
```
Input: warp://action/new_tab?path=C:\Projects\my-app
  decoded: "path"="C:\Projects\my-app"   ← correct
  PathBuf: "C:\Projects\my-app"           ← correct
```

The URL parsing itself works fine — the issue is downstream in `path_if_directory` where `is_dir()` can fail for paths that need canonicalization.

## Test plan

- [ ] Right-click a folder in Windows Explorer → "Open Warp in new tab" → verify it opens at the selected directory
- [ ] Right-click empty space inside a folder → "Open Warp in new tab" → verify it opens at the current directory
- [ ] Right-click a folder with spaces in the name → verify it works
- [ ] Right-click a folder that is a symlink/junction → verify it resolves correctly
- [ ] "Open Warp in new window" variants also work

Closes #9844
Related: #6665 (Linux counterpart), #6143 (duplicate instances)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -2131,7 +2131,22 @@ fn app_callbacks(is_integration_test: bool) -> warpui::platform::AppCallbacks {
         })),
         on_open_urls: Some(Box::new(move |urls, ctx| {
             for url in &urls {
-                let parsed_url = Url::parse(url);
+                let parsed_url = Url::parse(url).or_else(|_| {
+                    // On Windows, a bare path like "C:\folder" fails URL parsing
+                    // because "C:" looks like a URI scheme. Convert to a file:// URL.
+                    #[cfg(windows)]
+                    {
+                        let path = url.trim();
+                        if path.len() >= 2
+                            && path.as_bytes()[1] == b':'
+                            && (path.as_bytes()[0].is_ascii_alphabetic())
+                        {
+                            let forward = path.replace('\\', "/");
+                            return Url::parse(&format!("file:///{forward}"));
+                        }
+                    }
+                    Err(url::ParseError::EmptyHost)
+                });
                 match parsed_url {
                     Ok(url) => uri::handle_incoming_uri(&url, ctx),
                     Err(e) => log::warn!("Unable to parse received url: {e}"),
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -2141,8 +2141,8 @@ fn app_callbacks(is_integration_test: bool) -> warpui::platform::AppCallbacks {
                             && path.as_bytes()[1] == b':'
                             && (path.as_bytes()[0].is_ascii_alphabetic())
                         {
-                            let forward = path.replace('\\', "/");
-                            return Url::parse(&format!("file:///{forward}"));
+                            // from_file_path handles path separators, special chars (#, %), and UNC paths
+                            return Url::from_file_path(path);
                         }
                     }
                     Err(url::ParseError::EmptyHost)
--- a/app/src/root_view.rs
+++ b/app/src/root_view.rs
@@ -914,8 +914,22 @@ fn open_from_restored(arg: &OpenFromRestoredArg, ctx: &mut AppContext) {
     }
 }
 
-fn path_if_directory(path: &Path) -> Option<&Path> {
-    path.is_dir().then_some(path)
+fn path_if_directory(path: &Path) -> Option<PathBuf> {
+    if path.is_dir() {
+        return Some(path.to_path_buf());
+    }
+    // On Windows, paths from Explorer context menu may arrive without URL-encoding
+    // (e.g. `warp://action/new_tab?path=C:\Projects\my-app`). The url crate handles
+    // this correctly, but the resulting path can still fail `is_dir()` when the OS
+    // needs a normalized form (canonicalized, trailing-separator-stripped, etc.).
+    // Try canonicalizing as a fallback — this resolves symlinks, relative segments,
+    // and other minor path differences that `is_dir()` alone would reject.
+    if let Ok(canonical) = std::fs::canonicalize(path) {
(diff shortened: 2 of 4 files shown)


KuaaMU:
@cla-bot check

KuaaMU:
@oz-for-oss[bot] Thanks for the review! All three points addressed in the latest push:

1. **lib.rs file URL construction** — Replaced `Url::parse(&format!(...))` with `Url::from_file_path(path)`, which properly handles `#`, `%`, UNC paths, and other special characters.

2. **mod.rs trim removing leading whitespace** — Changed to `trim_end()` to preserve leading whitespace that may be part of a legitimate path (e.g. a folder starting with a space).

3. **uri_test.rs weak assertion** — Test now asserts the exact `PathBuf` value instead of just `result.is_some()`.

KuaaMU:
@oz-for-oss[bot] Thanks for the re-review! All three points were already addressed in the latest push:

1. **lib.rs URL construction** — Now uses `Url::from_file_path(path)` instead of string interpolation, which properly handles `#`, `%`, and UNC paths.

2. **uri/mod.rs trim()** — Already uses `trim_end()` (trailing only), preserving legitimate leading whitespace in paths like `%20Project`.

3. **uri_test.rs assertions** — Tests already assert exact `PathBuf` values (e.g. `Some(PathBuf::from(r"C:\Projects\my-app"))` and `Some(PathBuf::from(r"C:\Projects"))`), not just `Some`.

The inline comments reference the original diff before these fixes were pushed.

KuaaMU:
Confirmed this issue is now resolved in the latest Warp release — the "Open Warp in New Tab" context menu action correctly navigates to the selected directory. Closing this PR as it's no longer needed. Thanks!

Issue #6393, opened 2025-04-30
Terminal does not support clickable hyperlinks (e.g. Click to view scan report) via CMD+Click


tjpreston96:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/help/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When using the terminal to run CLI tools such as `wizcli`, the output occasionally includes anchor-style text, such as `Click to view scan report`. In other terminals like iTerm2, this text renders as a clickable hyperlink, and `CMD+Click` will open the URL in the default browser.

However, in this terminal environment (see attached screenshot), `CMD+Click` does not open the associated link, nor does the text behave like a hyperlink. This limits usability, especially when tools provide dynamic links to reports or dashboards directly in terminal output.

### To reproduce

Steps to Reproduce:
- Run a CLI tool (e.g., `wizcli iac scan`) that outputs clickable links in its final message.
- Attempt to `CMD+Click` on text such as `Click to view scan report` at the bottom of the output.
- Observe that nothing happens.

### Expected behavior

`CMD+Click` on the hyperlink-like text should open the target link in the default web browser, as it does in iTerm2 and similar terminal environments.

Actual Behavior:
The link is not interactive, and `CMD+Click` has no effect.

Environment:
- Terminal app in use: Warp tested against ITerm2
- macOS version: MacOS Sonoma 14.7.2 (23H311)
- CLI tool used: `wizcli` (private access)
- CLI output format: Standard, no redirection
- Working behavior confirmed in: iTerm2 (latest version)

### Screenshots, videos, and logs

Refer to the attached image and gif. Note the line at the bottom stating `Click to view scan report`, which appears as plain text and not an interactive link:
![Image](https://github.com/user-attachments/assets/f330fb33-8417-4eb9-9db2-c1c9c5a39c4b)

My apologies for the 12sec scan time delay but visualizes the behavior as clearly as possible:
![Image](https://github.com/user-attachments/assets/d3713877-286d-4785-9839-e0f436f0e48f)

### Operating system (OS)

macOS

### Operating system and version

[...]

Issue #176, opened 2021-09-09
hyperlink support 


elviskahoro:
### Describe the solution you'd like?

```
hyperlink support
```
FYI:
- Hyperlinks in terminal emulator https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
- Detect the terminal emulator supports hyperlinks or not https://github.com/jamestalmage/supports-hyperlinks
- A terminal hyperlink app with Node.js https://github.com/sindresorhus/terminal-link

### Is your feature request related to a problem? Please describe.

```
core
```

### Additional context

```
Sourced from the discord (feature requests channel; 08/27/2021)
How about add support for Hyperlinks?
```

dannyneira:
@elviskahoro just want to confirm we okay to close this as URLs supported.

ruudk:
I'm not sure what you mean? Because this doesn't seem to work?
```bash
printf '\e]8;;http://example.com\e\\This is a link\e]8;;\e\\\n'
```
<img width="580" alt="Screenshot 2022-12-29 at 20 05 40@2x" src="https://user-images.githubusercontent.com/104180/209997642-38e36717-196b-4585-89c4-6bc1f07d9b56.png">

ruudk:
@dannyneira Could you confirm that this is not solved? Because having hyperlink support like this is so helpful.

birkirb:
Also in #3108 and #4194

mattrq:
This feature looks like a natural fit for warp regardless of the current interest. 

Seems like a quick win?! 🤔

aljos:
@dannyneira any chance to implement this feature?

strickvl:
Registering my interest in this as well!

wdthem:
Another plus one from me. Please don't forget support for local file links as well, (file://)

ivpadim:
+1

rbarbadillo:
+1

Naragorn:
+1

RafaB15:
+1

joshle298:
+1

CurtisScopulovic:
+ 1 will need to use other terminals until this basic feature exists lol

CurtisScopulovic:
@captainsafia can this please happen? It's been 5 years of people asking for OSC 8 hyperlink support. It's not really just an inconvenience anymore, not having this enhancement impacts the ability of the core product to facilitate a high-productivity experience with AI terminal tools like Claude Code.

[...]

Issue #123, opened 2021-08-14
Tab completion on `curl -I` results in another I being added (repeatedly)


RSO:
**Describe the bug**
When I put `curl -I` in the terminal and press tab to autocomplete another `I` is added to the command, if I press tab again it repeats

**To Reproduce**
1. In any folder
2. Type `curl -I`
3. Press tab

**Expected behavior**
I expect either to see a list of different options I can add to the command, or nothing to happen (if there's no suggestions).

**Screenshots**
<img width="227" alt="CleanShot 2021-08-14 at 11 03 50@2x" src="https://user-images.githubusercontent.com/139663/129440998-d80aceea-1a67-4f73-8c2c-79998d9ddd3b.png">

**Desktop (please complete the following information):**
 - OS Version: MacOS 11.5

zachlloyd:
Thanks for the report - still working through a number of completion issues.

@charlespierce something to check with your parser change.

gagata:
Looks like it's been fixed, so closing this one!

PR #9850, opened 2026-05-01
Support OSC 8 hyperlinks (GH6393)

### LINKED ISSUE #4194: Add support for OSC 8: Hyperlinks in the terminal

_No response_

Support text formatted with links in the terminal (OSC 8) without the URL needing to be visible.

Allow output like "[This is a link](http://example.com/)" using
`printf '\e]8;;http://example.com\e\\This is a link\e]8;;\e\\\n'"`

image
[Image source](https://github.com/xtermjs/xterm.js/issues/1134) 

**Relevant info**
- Related to Issue [3108](https://github.com/warpdotdev/Warp/issues/3108)
- [OSC 8 adoption](https://github.com/Alhadis/OSC8-Adoption/) in terminal emulators
- Info on [Hyperlinks  in the terminal](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)
- More[ OSC codes](https://chromium.googlesource.com/apps/libapps/+/nassh-0.8.41/hterm/doc/ControlSequences.md#OSC)

### Additional context

Nice to have

1 (Not too important)

None

th1nkful:
> [!NOTE]
> the issue was labelled bug + ready-to-implement but the scope of the change seemed pretty big, so I’m creating a spec first to make sure it aligns fully with the warp team  

Warp does not currently parse OSC 8 hyperlink escape sequences, so output that uses the standard `ESC ] 8 ; params ; URI ESC \` form (e.g. wizcli's "Click to view scan report") is unclickable. The spec proposes parsing OSC 8, attaching the URI to the cells covered by the span via a per-grid hyperlink registry, and routing Cmd+click through the existing `ctx.open_url` path used for auto-detected URLs.

product.md captures the behavior surface as 18 numbered, testable invariants (recognition / closing / params, hover and click affordances, right-click menu, copy semantics, coexistence with auto-detected URLs, block boundaries, streaming, scrollback, sharing as markdown, AI context, robustness, and a scheme allow-list).

tech.md grounds the implementation in the existing OSC dispatcher (app/src/terminal/model/ansi/mod.rs:811-1230), the Handler trait's default-impl pattern, both cell-storage variants (FlatStorage and Cell/CellExtra), and the GridHighlightedLink hover/click flow in app/src/terminal/view/link_detection.rs. The plan is feature-flagged behind FeatureFlag::OscHyperlinks so the layers can land independently.

## Linked Issue

This PR closes #4194, #6393, #176

## Screenshots / Videos

CleanShot 2026-05-07 at 20 47 31@2x

CleanShot 2026-05-07 at 20 47 37@2x

CleanShot 2026-05-07 at 20 47 15@2x

## Agent Mode

CHANGELOG-BUG-FIX: Add support for hyperlinks rendered using OSC8


th1nkful:
@oss-maintainers — bringing this in for a maintainer pass per CONTRIBUTING.md. Have used the three oz review rounds and would appreciate a human review on the spec.

specs/GH6393/tech.md
```diff
@@ -0,0 +1,461 @@
...
+2. **Close form.** `params == []`, or `params == [b""]`, or the rejoined URI field is empty → `Ok(None)`. Three accepted shapes because real emitters send all three.
+3. **Open form.** Rejoined URI field is non-empty → parse the params field and return `Ok(Some(Hyperlink {...}))`. The params field is split on `:`; recognized keys are `id=...`, all others ignored. A params entry without `=` is `Err(MalformedParam)`.
+4. **Error cases.** Non-UTF-8 bytes in the URI → `Err(InvalidUtf8)`. URI exceeding `MAX_URI_BYTES` (a `pub const` defined in this module — see §3) → `Err(InvalidUtf8)`, **checked on the raw `&[u8]` rejoin before allocating the URI `String`**, so a 1 GB OSC 8 sequence never produces a 1 GB allocation. Empty params slice with no opening byte at all isn't reachable because `osc_dispatch` already guards against `params.is_empty()`.
```
acarl005: You're mapping exceeding the max length to malformed UTF-8 error?

> URI exceeding `MAX_URI_BYTES` (a `pub const` defined in this module — see §3) → `Err(InvalidUtf8)`

specs/GH6393/tech.md
```diff
@@ -0,0 +1,461 @@
...
+
+**Cell storage — block list path (`FlatStorage`).** `crates/warp_terminal/src/model/grid/flat_storage/` uses a run-length-encoded `AttributeMap<T>` per attribute: `FgColorMap` and `BgAndStyleMap` (`flat_storage/style.rs:18-21`). This representation deduplicates long runs of cells that share the same value into a single map entry — a perfect fit for hyperlinks, where one OSC 8 span typically covers many adjacent cells with the same URI. Used by `GridHandler` (`grid_handler.rs:381`).
+
+**Cell storage — alt-screen / row-based path (`Cell` + `Row`).** `crates/warp_terminal/src/model/grid/cell.rs:144` defines `Cell` (24 bytes, deliberately tightly packed; the doc comment at line 122 calls out that adding bytes to `Cell` itself bumps the struct to 32 bytes — a 33% memory hit). Optional rare attributes go in `CellExtra` (`cell.rs:113-120`), a boxed side allocation that already holds `cell_with_zero_width` and `end_of_prompt`. Hyperlinks belong in `CellExtra`, not in the base `Cell`.
```
acarl005: It's not quite as simple as "BlockList is backed by flat storage" and "AltScreen is backed by 2D array/grid"

The block list stores characters across both. Things move from the 2d-array storage into flat storage once they enter the scrollback.

acarl005 (commented):
Great specs! I've been wanting this feature for a while. Feel free to update your PR with the actual code now. I left a couple of small comments inline.

th1nkful:
Thanks for the review — addressed all four items in fb70400:

- **grid_handler.rs `bounds_to_markdown_string`** — single-row and end-row column ranges now use `end.col.saturating_add(1).min(row_length)` so the final selected column is included, matching the inclusive `end` semantics used elsewhere.
- **view.rs OSC 8 right-click** — added an **Open link** menu item above **Copy link** that dispatches `TerminalAction::OpenGridLink`, so opens route through the same `link_security` scheme allow-list as Cmd/Ctrl-click.
- **integration/step.rs `CmdClickOnce`** — synthetic direct-open click now picks the modifier per platform (Cmd on macOS, Ctrl on Linux/Windows) to match `should_directly_open_link`, so the test exercises the shipped shortcut on every platform.
- **.claude/scheduled_tasks.lock** — removed from the index and added to `.gitignore`.

acarl005:
@cla-bot check

seemeroland:
Will leave review to @acarl005!

vkodithala:
Hey @th1nkful, thanks so much for the contribution. Andy's out of office this week, we'll get to this review when he's back! Haven't forgetten about this :)

app/src/integration_testing/url/assertion.rs
```diff
@@ -0,0 +1,35 @@
...
+/// subsequent assertions in the same test see a clean slate.
+pub fn assert_url_opened(expected: impl Into<String>) -> AssertionCallback {
+    let expected = expected.into();
+    Box::new(move |_app, _window_id| {
+        let urls = read_captured_urls();
+        if urls.iter().any(|u| u == &expected) {
+            // Drain so this assertion is idempotent and the buffer is
+            // cleared for any subsequent assertion in the test.
+            let _ = drain_captured_urls();
+            return warpui::integration::AssertionOutcome::Success;
```
acarl005:
```suggestion
            return AssertionOutcome::Success;
```

Try to keep paths to 1 or 2 `::` separators when it's practical to do so.

app/src/terminal/model/ansi/mod.rs
```diff
@@ -939,6 +939,25 @@ where
                 unhandled(params);
             }
 
+            // OSC 8: Hyperlink (terminal anchor) sequences.
+            // Format: OSC 8 ; params ; URI ST. An empty URI closes the active
+            // hyperlink. Gated behind FeatureFlag::OscHyperlinks; when off,
+            // the bytes fall through to `unhandled(params)` and behavior
+            // matches pre-feature.
```
acarl005:
```suggestion
            // hyperlink.
```

We don't need comments explaining how Rust itself works, i.e. that match arms fall through when their conditions return false. 

The meta-comment would be that the comments are too verbose in this diff generally.

crates/warp_terminal/src/model/ansi/control_sequence_parameters.rs
```diff
@@ -708,6 +708,259 @@ impl TryFrom<&[u8]> for PromptKind {
     }
 }
 
+/// Maximum byte length of a URI carried by an OSC 8 hyperlink. Sequences
+/// whose URI exceeds this cap are dropped at parse time before the URI
+/// `String` is allocated, so a hostile sender can't trigger an arbitrarily
+/// large heap allocation.
+pub const MAX_URI_BYTES: usize = 4096;
```
acarl005:
```suggestion
pub const MAX_URI_BYTES: usize = 2083;
```

We should copy VTE here.

app/src/integration_testing/url/step.rs
```diff
@@ -0,0 +1,47 @@
...
+/// since the last `install_open_url_capture()` step. Cleared by that
+/// step so each test starts with a clean buffer.
+fn captured_urls() -> &'static Mutex<Vec<String>> {
+    static CAPTURED: OnceLock<Mutex<Vec<String>>> = OnceLock::new();
+    CAPTURED.get_or_init(|| Mutex::new(Vec::new()))
+}
+
+/// Drain the captured URLs into a fresh Vec. Used by assertions.
+pub(crate) fn drain_captured_urls() -> Vec<String> {
+    let mut guard = captured_urls().lock().expect("captured_urls poisoned");
```
acarl005: Prefer `parking_lot` for this since poisoning is not an issue.

app/src/terminal/model/ansi/mod.rs
```diff
@@ -939,6 +939,25 @@ where
...
+            //
+            // Reference: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
+            b"8" if FeatureFlag::OscHyperlinks.is_enabled() => {
+                match Hyperlink::parse_osc_params(&params[1..]) {
+                    Ok(hyperlink) => self.handler.set_hyperlink(hyperlink),
+                    Err(err) => {
+                        // Do NOT pass `params` to the generic `unhandled` formatter:
+                        // OSC 8 payloads are untrusted terminal data and can be
+                        // arbitrarily large. Log only the bounded error variant.
+                        debug!("[osc 8 parse error]: {:?}", err);
```
acarl005: I'm not convinced avoiding `unhandled` has any value. Hasn't the parser already allocated a slice for all the bytes regardless whether they have exceeded the length limit?

crates/warp_terminal/src/model/grid/hyperlink_registry.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+impl GetSize for HyperlinkId {}
+
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct HyperlinkRegistry {
+    /// Reverse map: hyperlink → id. Lets `intern` dedupe.
+    by_link: HashMap<Hyperlink, HyperlinkId>,
+    /// Forward array: id → hyperlink. The id's `NonZeroU32` value is
+    /// `index_in_by_id + 1`.
```
acarl005: What is `index_in_by_id`?

app/src/terminal/model/grid/grid_handler.rs
```diff
@@ -811,6 +833,204 @@ impl GridHandler {
...
+    /// Returns the contiguous OSC 8 hyperlink span at `displayed_point`, if
+    /// the cell there is part of one. The returned range is contiguous —
+    /// cross-run grouping by `id` is intentionally out of scope (see
+    /// `specs/GH6393/tech.md` Follow-ups).
+    ///
+    /// Mirrors the shape of [`Self::url_at_point`] but skips
+    /// `urlocator`: the URI doesn't live in the visible cell text, so
+    /// detection reduces to "walk left/right while the next adjacent cell
+    /// carries the same `HyperlinkId`."
+    pub fn hyperlink_at_point(&self, displayed_point: Point) -> Option<Link> {
```
acarl005, line 919: This can use some unit tests

app/src/terminal/model/grid/grid_handler.rs
```diff
@@ -970,13 +1198,50 @@ impl GridHandler {
...
+                            }
+                        });
+                        if cell_uri != osc8_open_uri {
+                            // Close previous span, if any.
+                            if osc8_open_uri.is_some() {
+                                text.push_str("\x1b]8;;\x1b\\");
+                            }
+                            // Open new span, if any.
+                            if let Some(ref uri) = cell_uri {
+                                text.push_str("\x1b]8;;");
```
acarl005: We shouldn't be hard-coding these values. Use these instead

https://github.com/warpdotdev/warp/blob/7076885b3e744d493775c88178038f12f0257a04/crates/warp_terminal/src/model/escape_sequences.rs#L18-L18

app/src/terminal/grid_renderer.rs
```diff
@@ -795,6 +812,14 @@ fn render_grid_without_ligatures<'a>(
...
                 && visible_cursor_shape == Some(CursorShape::Block))
             .then(|| theme.cursor().into_solid());
+            cache_osc8_hyperlink_position(
+                cell.hyperlink_id(),
+                &mut seen_hyperlink_ids,
+                osc8_cache_scope,
+                grid_origin + cell_size * vec2f(col as f32, offset_row as f32),
+                cell_size,
+                ctx,
+            );
```
acarl005: This makes me nervous, as the grid renderer is a very hot code path. It's a risk of taxing performance. The fact that this is only for integration tests makes me think it isn't work the risk. Can we figure out a way to click on links in integration tests without this?

app/src/bin/oss.rs
```diff
@@ -24,6 +24,10 @@ fn main() -> Result<()> {
     if cfg!(debug_assertions) {
         state = state.with_additional_features(warp_core::features::DEBUG_FLAGS);
     }
+    // Features that should be enabled for OSS builds. The OSS channel has
+    // its own (small) flag-set rather than inheriting from dogfood/preview,
+    // so additions live here when they're ready for OSS users.
+    state = state.with_additional_features(&[warp_core::features::FeatureFlag::OscHyperlinks]);
```
acarl005:
```suggestion
```

The oss channel should be treated the same as the stable channel, i.e. generally the default features should apply. It should stick to only `DOGFOOD_FLAGS` for now. Please remove from `PREVIEW_FLAGS`. That will limit this to being deployed to our internal-only nightly release. This change affects very core functionality so it's important that we catch bugs internally first.

In order to make it work when you test locally, e.g. via `cargo run`. For that, you'll want to use e.g. `cargo run --features osc_hyperlinks`. To make that work, you'll need to add this feature flag here:

- https://github.com/th1nkful/warp/blob/4e923d3d90a0ec07e341a7e6fb11d0e9a5199bc6/app/src/features.rs#L493-L493
- https://github.com/th1nkful/warp/blob/4e923d3d90a0ec07e341a7e6fb11d0e9a5199bc6/app/Cargo.toml#L970-L970

crates/warp_terminal/src/model/grid/flat_storage/hyperlink.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use super::attribute_map::AttributeMap;
+
+/// Map holding each cell's OSC 8 hyperlink id (or `None` for cells that
+/// aren't part of a hyperlink).
+pub type HyperlinkIdMap = AttributeMap<Option<HyperlinkId>>;
+
+impl From<&cell::Cell> for Option<HyperlinkId> {
+    fn from(value: &cell::Cell) -> Self {
+        value.hyperlink_id()
```
acarl005: This doesn't seem like a semantically correct use of the `From` trait. Also, it looks unused. Can you remove?

crates/warp_terminal/src/model/grid/flat_storage/hyperlink.rs
```diff
@@ -0,0 +1,28 @@
...
+    }
+}
+
+impl PartialEq<&cell::Cell> for Option<HyperlinkId> {
+    fn eq(&self, other: &&cell::Cell) -> bool {
+        *self == other.hyperlink_id()
```
acarl005: Similar comment to above.

crates/warp_terminal/src/model/ansi/control_sequence_parameters.rs
```diff
@@ -708,6 +708,259 @@ impl TryFrom<&[u8]> for PromptKind {
...
+                    // Bound the id length before allocating: it's untrusted
+                    // terminal data and the `Hyperlink` it ends up in is used
+                    // as the registry key. Oversized ids are dropped (the URI
+                    // remains usable as a hyperlink without an id).
+                    if value.len() > MAX_ID_BYTES {
+                        continue;
+                    }
+                    let value =
+                        str::from_utf8(value).map_err(|_| HyperlinkParseError::InvalidUtf8)?;
+                    id = Some(value.to_owned());
```
acarl005, line 811: The [official spec](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) says an empty ID should be treated as the same as omitting the ID parameter. Looks like that isn't the case in this PR?

crates/warp_terminal/src/model/ansi/control_sequence_parameters.rs
```diff
@@ -708,6 +708,259 @@ impl TryFrom<&[u8]> for PromptKind {
...
+        let uri = String::from_utf8(uri_bytes).map_err(|_| HyperlinkParseError::InvalidUtf8)?;
+
+        let mut id: Option<String> = None;
+        if !params_field.is_empty() {
+            for pair in params_field.split(|byte| *byte == b':') {
+                if pair.is_empty() {
+                    continue;
+                }
+                let Some(eq_idx) = pair.iter().position(|byte| *byte == b'=') else {
+                    return Err(HyperlinkParseError::MalformedParam);
```
acarl005: The [official spec](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) doesn't say malformed params are an error. It says unknown params should be silently ignored. The link should ultimately still be clickable, just skip the invalid params.

crates/warp_terminal/src/model/grid/hyperlink_registry.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl HyperlinkId {
+    /// Returns the underlying integer value of this id. Intended for callers
+    /// that need a stable string identity for the hyperlink (e.g. caching a
+    /// rendered cell position by id for integration tests); other callers
+    /// should treat ids as opaque.
+    #[inline]
+    pub fn get(self) -> u32 {
+        self.0.get()
+    }
```
acarl005: Will be unused if you get rid of position caching.

crates/warpui_core/src/integration/step.rs
```diff
@@ -210,6 +210,10 @@ pub enum IntegrationTestEvent {
 
 pub enum MouseEvent {
     ClickOnce,
+    /// Like `ClickOnce`, but with the Command (Cmd) modifier held during
+    /// both mouse-down and mouse-up. Used to test Cmd-click affordances
+    /// like opening hyperlinks.
+    CmdClickOnce,
```
acarl005: This is causing duplication. Can we extend the original `ClickOnce` variant with a field that stores `ModifiersState`?

app/src/terminal/model/grid/grid_handler.rs
acarl005: Also, out of an abundance of caution, please short-circuit this method if `FeatureFlag::OscHyperlinks.is_enabled()` is false. Should be a no-op but I just want to make sure that if we need to disable it that it shuts off any potential for bugs remaining in the app.

crates/warp_terminal/src/model/grid/cell.rs
```diff
@@ -276,7 +282,31 @@ impl Cell {
...
+    /// Set or clear the OSC 8 hyperlink id on this cell. Allocates `extra`
+    /// only when setting a non-`None` id; clearing an unset cell is free.
+    #[inline]
+    pub fn set_hyperlink_id(&mut self, id: Option<super::HyperlinkId>) {
+        match (id, self.extra.as_deref_mut()) {
+            (Some(id), Some(extra)) => extra.hyperlink_id = Some(id),
+            (Some(id), None) => {
+                self.extra.get_or_insert_with(Default::default).hyperlink_id = Some(id);
+            }
+            (None, Some(extra)) => extra.hyperlink_id = None,
```
acarl005:
```suggestion
        match (self.extra.as_deref_mut(), id) {
            (Some(extra), id) => extra.hyperlink_id = id,
            (None, Some(id)) => {
                self.extra.get_or_insert_with(Default::default).hyperlink_id = Some(id);
            }
```
nit: you can reduce the number of arms here

app/src/integration_testing/url/step.rs
```diff
@@ -0,0 +1,47 @@
...
+        // don't poison our assertions.
+        captured_urls()
+            .lock()
+            .expect("captured_urls poisoned")
+            .clear();
+        app.update(|ctx| {
+            ctx.set_before_open_url(|url, _ctx| {
+                let mut guard = captured_urls().lock().expect("captured_urls poisoned");
+                guard.push(url.to_owned());
+                url.to_owned()
```
acarl005: This is cool!

app/src/terminal/model/grid/grid_handler.rs
acarl005: If you see others in the existing code, can you drive-by and fix those too? 🙏

crates/warp_terminal/src/model/grid/hyperlink_registry.rs
```diff
@@ -0,0 +1,204 @@
...
+/// Past this cap, [`HyperlinkRegistry::intern`] returns `None`; existing
+/// entries continue to resolve.
+pub const MAX_DISTINCT_ENTRIES: usize = 4096;
+
+/// Opaque, dense, non-zero handle into a [`HyperlinkRegistry`]. Stored in
+/// each [`crate::model::grid::cell::Cell`] that's part of an OSC 8 span.
+///
+/// `NonZeroU32` lets `Option<HyperlinkId>` fit in 4 bytes thanks to niche
+/// optimization, which keeps the registry handle compact alongside the
+/// other per-cell attributes.
```
acarl005, line 35: Super cool

acarl005 (changes requested):
Thanks for this PR! This is certainly one of the most complex contributions we've gotten so far. I also really appreciate the commitment to testing. I tested it locally and it's very cool seeing this working.

I've left some inline comments with some changes. Also here are some more coarse changes I'm requesting:

- I don't think we should add position caching to grid rendering. It's too risky of a code path and making integration tests more convenient isn't a good justification for taking on this risk. I'd prefer a different way to click these in integration tests.
- Remove the “copy as markdown” feature. This bloats the scope of this change. I know I approved it in the product spec but there were a lot of details omitted from the tech spec and I'm realizing that belongs in a separate PR. There's just too much ambiguity with how to represent everything that can appear in a block as markdown. Please delete that code and remove from these specs.
- Similarly, remove "Copy as terminal bytes” from your spec.
- Remove URL scheme allowlist. I don't actually think this is necessary. For example, Ghostty doesn't do that. It's also affecting other URL-related features and isn't feature-flagged. That is implying that it's exceeding the scope of this PR. We can consider adding it back in a future PR.
- Make sure `file://` works, as the current implementation doesn't allow it. I don't think we need to enforce the hostname as the [official spec](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) suggests. Integration tests for this would be useful.
- There are a lot of references to the specs, e.g. `See specs/GH6393/tech.md §3c.`. I appreciate you wanting to explain the code. However, the specs are intended to be prompting artifacts. They shouldn't be a crutch for understanding the code itself. If it's difficult to understand without the code itself plus comments, that's a code smell.
- Use imports more. There are a lot of very long absolute paths. Notice that we usually keep references to 1 or maybe 2 `::` separators. 
- Remove a lot of the unnecessary comments. I know LLM-generated code often does this, but I find these comments counter-productive to long-term maintainability. To explain this a bit further:
    - There are several length limits you've added. All of them explain that without a limit, malicious/buggy programs can do so-and-so and we need to defend against it. That is a general principle for length limits and doesn't need to be re-explained for each and every one.
    - Similar to above, you explain some threat models in places where I don't think it's necessary.
    - You also don't need to explain language functionality, e.g. [this](https://github.com/warpdotdev/warp/pull/9850/changes#r3383343879)
    - You don't need to explain operations that are similar to other operations in close proximity. For example, you don't need to explain a delegator method in a sea of other delegator methods, or explain resetting the hyperlink ID when a bunch of other fields are also being reset in the same function body.
    - You don’t need to explain what a function does at a _call site_ for the function, e.g. `cursor_cell.set_hyperlink_id(hyperlink_id)`. Prefer doc comments on the function definition.

acarl005:
Sorry about the bot and its warnings. It's having some logic errors (which we'll fix today).

th1nkful:
> Sorry about the bot and its warnings. It's having some logic errors (which we'll fix today).

My bad, I hit rerequest review a little early. Hoping to get this sorted soon, bunch of the changes you requested are done. Just a few more to sort out 👍🏻

th1nkful:
Thanks for the thorough review @acarl005! I've addressed all the coarse points from your review. Summary below; per-line replies are on each thread.

**Coarse changes**
- **Position caching removed** — dropped from grid rendering entirely; integration tests now locate links via the model's `hyperlink_at_point` lookup and dispatch `OpenGridLink` directly (no synthetic pixel click). (`46039bda`)
- **Copy as markdown removed** — deleted the code and removed it from the specs (now noted as out of scope). (`46039bda`)
- **Copy as terminal bytes removed** from the specs. (`46039bda`)
- **URL scheme allowlist removed** — `open_url` is no longer gated by a scheme allowlist; no longer feature-flagged or affecting other URL features. (`46039bda`)
- **`file://` works** — no scheme restriction, and no hostname enforcement (`file:///path` is accepted). Added an integration test (`test_osc8_file_scheme_opens_url`) that opens a hostname-less `file://` link. (`aa0231de`)
- **Spec references removed from code** — dropped every `specs/GH6393/…` reference from code comments. (`07bb833d`, `aa0231de`)
- **Imports / long paths** — shortened the longest absolute paths (e.g. `HyperlinkId`/`HyperlinkRegistry` now imported). (`aa0231de`)
- **Unnecessary comments trimmed** — removed Rust-explaining comments, repeated threat-model rationale on the length limits, and call-site/delegator explanations. (`07bb833d`, `aa0231de`)

I've also merged latest `master` into the branch.

From Claude ✨

app/src/integration_testing/url/assertion.rs
th1nkful: The URL test-helper module (`url/assertion.rs`) was removed entirely when position caching went away, so this path no longer exists.

— Fixed in `46039bda`. From Claude ✨

app/src/terminal/model/ansi/mod.rs
th1nkful: Trimmed this to a single line and cut the over-explaining comments throughout this diff (Rust-language explanations, repeated threat-model rationale, call-site/delegator notes).

— Fixed in `07bb833d`, `aa0231de`. From Claude ✨

crates/warp_terminal/src/model/ansi/control_sequence_parameters.rs
th1nkful: Done — capped at 2083 to match VTE.

— Fixed in `07bb833d`. From Claude ✨

app/src/integration_testing/url/step.rs
th1nkful: The URL-capture step module was removed along with position caching, so this is moot now.

— Fixed in `46039bda`. From Claude ✨

app/src/terminal/model/ansi/mod.rs
th1nkful: Agreed — reverted to passing `params` to `unhandled` on parse error rather than special-casing it.

— Fixed in `07bb833d`. From Claude ✨

crates/warp_terminal/src/model/grid/hyperlink_registry.rs
th1nkful: Reworded: an id's `NonZeroU32` value is its index in `by_id` plus one (the `+1` keeps the first id non-zero).

— Fixed in `aa0231de`. From Claude ✨

app/src/terminal/model/grid/grid_handler.rs
th1nkful: Added unit tests in `grid_handler_tests.rs`: contiguous-span detection, adjacent links with different URIs not merging, and the flag-disabled short-circuit.

— Fixed in `aa0231de`. From Claude ✨

app/src/terminal/model/grid/grid_handler.rs
th1nkful: The only code emitting these literal escape bytes was the copy-as-markdown path, which has been removed, so there are no hard-coded OSC 8 bytes left in production code.

— Fixed in `46039bda`. From Claude ✨

app/src/terminal/grid_renderer.rs
th1nkful: Agreed. Removed the position caching from the grid renderer entirely. Integration tests now locate the link via the model's `hyperlink_at_point` lookup and dispatch `OpenGridLink` directly, so the hot path is untouched.

— Fixed in `46039bda`. From Claude ✨

app/src/bin/oss.rs
th1nkful: Done: removed the forced enable in `oss.rs`, dropped `OscHyperlinks` from `PREVIEW_FLAGS` (it's `DOGFOOD_FLAGS` only now), and added an `osc_hyperlinks` cargo feature plus the compile-time bridge so it can be enabled locally via `cargo run --features osc_hyperlinks`.

— Fixed in `aa0231de`. From Claude ✨

crates/warp_terminal/src/model/grid/flat_storage/hyperlink.rs
th1nkful: Removed the `From<&Cell>` impl (it was unused and not a semantically correct use of `From`).

— Fixed in `aa0231de`. From Claude ✨

crates/warp_terminal/src/model/grid/flat_storage/hyperlink.rs
th1nkful: Removed the `PartialEq<&Cell>` impl as well.

— Fixed in `aa0231de`. From Claude ✨

crates/warp_terminal/src/model/ansi/control_sequence_parameters.rs
th1nkful: Good catch — an empty `id=` is now treated the same as omitting the param (added a unit test `empty_id_is_treated_as_omitted`).

— Fixed in `aa0231de`. From Claude ✨

crates/warp_terminal/src/model/ansi/control_sequence_parameters.rs
th1nkful: Fixed — malformed/unknown params are now silently ignored per the spec; the link stays clickable. Removed the `MalformedParam` error variant and updated the tests.

— Fixed in `aa0231de`. From Claude ✨

crates/warp_terminal/src/model/grid/hyperlink_registry.rs
th1nkful: Removed — `HyperlinkId::get()` went away with position caching.

— Fixed in `46039bda`. From Claude ✨

crates/warpui_core/src/integration/step.rs
th1nkful: The `CmdClickOnce` variant was removed entirely along with the synthetic-click approach, so there's no longer any duplication with `ClickOnce`.

— Fixed in `46039bda`. From Claude ✨

app/src/terminal/model/grid/grid_handler.rs
th1nkful: Done — both `hyperlink_at_point` and `hyperlink_uri_at_point` now short-circuit to `None` when `FeatureFlag::OscHyperlinks` is disabled (covered by a unit test).

— Fixed in `aa0231de`. From Claude ✨

crates/warp_terminal/src/model/grid/cell.rs
th1nkful: Simplified `set_hyperlink_id` to an `if/else` (allocates `extra` only when setting a non-`None` id), which is fewer arms than the original match.

— Fixed in `07bb833d`. From Claude ✨

app/src/terminal/model/grid/grid_handler.rs
th1nkful: The hard-coded escapes only existed in the now-removed copy-as-markdown path, so there's nothing left to migrate here.

— Fixed in `46039bda`. From Claude ✨

th1nkful:
Thanks for the patience @acarl005 can you review this one again? I have run the review skill locally, looking good. Let me know if I should just run Oz review in CI again

laaaaaaaaaxxxx:
+1 from a daily Warp user on macOS — really hoping this one makes it in.

This is exactly the thing that bugs me every day: I've got CLI coding agents running in Warp that spit out stuff like `file.rs:120` or `PR #123`, and none of it's clickable right now. The same output opens fine in VS Code's terminal but just sits there dead in Warp. Being able to cmd+click those would be so good.

Also gotta say @th1nkful — this is seriously impressive. ~3k lines and you kept grinding on it for months. The registry design and all the perf/flag care look really solid, and it's wild it's basically green with just a couple threads left. Thanks for sticking with such a big one 🙏

Not trying to rush the review or anything — just wanted to throw in that there are real people out here who actually want this.

app/src/terminal/view.rs
```diff
@@ -18239,9 +18258,16 @@ impl TerminalView {
...
+                    .model
+                    .lock()
+                    .link_at_range(url, RespectObfuscatedSecrets::No);
+                ctx.notify();
+                ctx.open_url(&uri);
+            }
+            GridHighlightedLink::Hyperlink { link, uri } if link.contains(position) => {
                 ctx.notify();
-                ctx.open_url(&model.link_at_range(url, RespectObfuscatedSecrets::No));
+                ctx.open_url(uri);
```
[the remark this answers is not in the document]
acarl005: Ignore the feedback about an explicit scheme policy. We've already decided that's not needed.

However, it suggests validating that the URL can pass `Url::parse` first before attempting to open it. I think that's actually a good idea. We should check that here and also [there](https://github.com/warpdotdev/warp/pull/9850/changes#diff-8503e5871bad515725ab6aa5e21c7b1611dacbc476bde23311f4e8d68d444bd8R477)

crates/warp_terminal/src/model/grid/flat_storage/row_iterator.rs
```diff
@@ -114,6 +116,7 @@ impl Iterator for RowIterator<'_> {
             cell.fg = fg;
             cell.bg = bg;
             cell.flags = flags;
+            cell.set_hyperlink_id(hyperlink_id);
```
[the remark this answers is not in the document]
acarl005: This seems valid. the spacer needs to be hoverable as well. We should make sure this case is covered in tests too.

crates/warp_terminal/src/model/grid/flat_storage/mod.rs
```diff
@@ -228,6 +242,12 @@ impl FlatStorage {
                     self.bg_and_style_map
                         .push_attribute_change(offset.., bg_and_style);
                 }
+                if hyperlink_id != cell.hyperlink_id() {
+                    needs_processing = true;
+                    hyperlink_id = cell.hyperlink_id();
+                    self.hyperlink_id_map
```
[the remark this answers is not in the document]
acarl005: This comment is a false positive. However, it might be another good test to add.

specs/GH6393/tech.md
```diff
@@ -0,0 +1,375 @@
...
+#### 6a-ii. Warp Drive (cloud objects, shared sessions)
+
+Warp Drive blocks store the same `SerializedBlock` shape (`app/src/terminal/model/block.rs:474-480`), with `stylized_output: String`. Same reasoning as sqlite: OSC 8 bytes round-trip transparently in the string and re-parse on load.
+
+- **Forward and backward compat both ride for free** for the byte-stream encoding. No protocol-version bump is needed for blocks whose hyperlink state can be recovered from `stylized_output`.
+- **Risk: a future representation change.** If Warp Drive ever moves from the byte-stream encoding to a structured per-cell encoding (e.g. for performance), the cell-side `hyperlink_id` field becomes the wire field of record. The `serde(default, skip_serializing_if)` defaults above already cover both directions. We name this in the spec so anyone making that representation change tests both compatibility directions.
+
+#### 6a-iii. Session-sharing protocol (`session-sharing-protocol` workspace crate)
+
+The session-sharing protocol streams **events**, not byte streams (set cursor, write char, set SGR, etc.). Adding OSC 8 means a new event:
```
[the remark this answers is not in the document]
acarl005: We can leave this as a follow-up. This PR is already becoming pretty huge

acarl005:
Hi @th1nkful thanks for another revision. The intended flow is for you to call another Oz review after revising based on the human-reviewer-requested changes. I went ahead and kicked that off. Most of the bot's feedback is valid.

.agents/skills/review-pr/scripts/resolve_spec_context.py
```diff
@@ -0,0 +1,723 @@
+from __future__ import annotations
```
acarl005: I assume this is unintentional

app/src/terminal/view.rs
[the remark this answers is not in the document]
th1nkful: Per @acarl005 we are not adding a scheme allow-list. Instead, all three OSC 8 open paths — `maybe_open_link`, the `HyperlinkClick` action, and `open_highlighted_link` (the `OpenGridLink` / context-menu "Open link" handler) — now route through a shared `TerminalView::open_hyperlink_uri` helper that validates the URI with `Url::parse` and skips it if it does not parse before calling `open_url`. Fixed in e901fffb and 561fadcf.

crates/warp_terminal/src/model/grid/flat_storage/row_iterator.rs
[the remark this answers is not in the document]
th1nkful: Fixed. When `cell_width == 2` the spacer cell now inherits the wide glyph`s id (`row[idx + 1].set_hyperlink_id(hyperlink_id)`), matching how the live input path stamps both cells, so both halves stay hoverable after a flat-storage roundtrip. Covered by the new `test_wide_char_hyperlink_spacer_survives_roundtrip` test. Fixed in 312d2c8a.

crates/warp_terminal/src/model/grid/flat_storage/mod.rs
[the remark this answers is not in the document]
th1nkful: Per @acarl005 this is a false positive: the hyperlink attribute change is pushed at the linked cell`s offset, and the backfilled blank cells precede that offset so they stay `None`. Added `test_blank_cells_before_hyperlink_are_not_clickable` to lock the behavior in. Added in 312d2c8a.

specs/GH6393/tech.md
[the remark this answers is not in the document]
th1nkful: Deferred to a follow-up per @acarl005 — this PR is already very large. Note that the byte-stream persistence paths (sqlite and Warp Drive `stylized_output`) already round-trip OSC 8 for free since the escape bytes re-parse on load; only the event-streamed session-sharing path in §6a-iii is unimplemented, and it will be handled in a separate PR.

app/src/terminal/view/link_detection.rs
```diff
@@ -94,6 +101,7 @@ impl GridHighlightedLink {
             #[cfg(feature = "local_fs")]
             GridHighlightedLink::File(_) => "Open file",
             GridHighlightedLink::Url(_) => "Open link",
+            GridHighlightedLink::Hyperlink { .. } => "Open link",
```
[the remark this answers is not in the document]
th1nkful: Fixed — the hover tooltip now shows the URI itself for OSC 8 links instead of the generic "Open link" label, so the hidden destination is visible before clicking (product invariant 5). Auto-detected URLs are unchanged since their destination is already the visible text. Commit 7844c426.

specs/GH6393/product.md
```diff
@@ -0,0 +1,76 @@
...
+    - Restoring a session from history / Warp Drive / shared session preserves the URI on hyperlink spans so they remain clickable in the restored block.
+    - Searching within a block matches against the visible text, not the URI.
+
+13. **Sharing a block.** The default text-copy path never produces OSC 8 escape sequences (invariant 8 wins for that case).
+    - **Selection text-copy** (Cmd+C / context menu "Copy"): produces visible text, never OSC 8 bytes, never a markdown link. This is invariant 8 — restated here so there is no ambiguity.
+    - Dedicated "Copy as markdown" and "Copy as terminal bytes" export actions are out of scope for this iteration.
+
+14. **AI / agent context.** Block content fed to the AI assistant or agents includes the URI for any OSC 8 hyperlink span as **untrusted metadata**, distinct from the visible text and clearly labeled as such (e.g., as a structured field, an `<untrusted-uri>` tag, or an equivalent marker the agent cannot mistake for an instruction). The agent must treat OSC 8 URIs identically to any other untrusted output coming from the terminal:
```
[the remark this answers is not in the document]
th1nkful: Marked as an explicit follow-up / out of scope for this PR — the AI/agent context formatter is not changed here. product.md invariant 14 now carries that note so the spec matches the diff. Commit 4b2ec9d8.

specs/GH6393/tech.md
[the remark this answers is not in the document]
th1nkful: Marked as an explicit follow-up in §6a-iii and reverted the "no longer deferred" line — this PR does not touch the session-sharing protocol. Local history and Warp Drive persistence remain in scope since they ride the existing byte stream and re-parse OSC 8 on load. Commit 4b2ec9d8.

app/src/terminal/model/ansi/mod.rs
```diff
@@ -864,6 +864,17 @@ where
...
             }
 
+            // OSC 8: Hyperlink (terminal anchor) sequences.
+            // Format: OSC 8 ; params ; URI ST. An empty URI closes the active
+            // hyperlink.
+            // Reference: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
+            b"8" if FeatureFlag::OscHyperlinks.is_enabled() => {
+                match Hyperlink::parse_osc_params(&params[1..]) {
+                    Ok(hyperlink) => self.handler.set_hyperlink(hyperlink),
+                    Err(_) => unhandled(params),
```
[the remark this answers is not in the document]
th1nkful: Fixed — a parse error (malformed, non-UTF-8, or over-length) now clears the active hyperlink via `set_hyperlink(None)`, so later unrelated output cannot inherit a stale URI. Added `parse_osc8_malformed_sequence_clears_active_hyperlink` covering open-then-malformed. Commit is on the branch and lands on the next push.

app/src/terminal/view/link_detection.rs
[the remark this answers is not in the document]
th1nkful: Already addressed — the hover tooltip now shows the URI itself for OSC 8 links instead of the generic "Open link" label (product invariant 5). Commit 7844c426.

specs/GH6393/tech.md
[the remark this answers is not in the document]
th1nkful: Already addressed — §6a-iii is now marked an explicit out-of-scope follow-up and the "no longer deferred" line was reverted; this PR does not touch the session-sharing protocol. Commit 4b2ec9d8.

specs/GH6393/tech.md
```diff
@@ -0,0 +1,375 @@
...
+- **sqlite round-trip** (`app/src/persistence/block_list.rs` test module): persist a block with `stylized_output` containing an OSC 8 sequence; reload from sqlite; assert the reloaded `Block` has cells with non-None `hyperlink_id` and the URI matches.
+- **Warp Drive round-trip:** simulate the cloud serialization round-trip and verify the same.
+- **Old-client read-of-new-payload (session-sharing):** a CI matrix test that pins the old client to a tag-before-this-PR, runs the new client as the producer, and asserts the old client renders cells correctly (just not clickable). Lives in `crates/integration` and runs as part of the protocol-compat CI job.
+- **New-client read-of-old-payload (session-sharing):** the inverse — new client consumes a recorded session from before the PR; asserts the deserializer doesn't error and cells render plain.
+
+This subsumes the earlier "session persistence" follow-up; it is no longer deferred.
+
+### 7. AI context
+
+- **AI context** (invariant 14). The block→agent context formatter inlines `visible text (URI)` for hyperlinked spans so an agent reading wizcli output sees the URI without losing the visible label.
```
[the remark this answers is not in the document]
th1nkful: Already addressed — the AI-context section (§7) is now marked an explicit out-of-scope follow-up; the block-to-agent formatter is unchanged in this PR. Commit 4b2ec9d8.

app/src/terminal/model/ansi/mod.rs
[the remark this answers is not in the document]
th1nkful: Fixed — a parse error (malformed, non-UTF-8, or over-length) now clears the active hyperlink via `set_hyperlink(None)` in the OSC 8 dispatch, so subsequent output cannot inherit a stale URI. Covered by `parse_osc8_malformed_sequence_clears_active_hyperlink`. Commit d229f5fd (lands on next push).

app/src/terminal/model/grid/ansi_handler.rs
```diff
@@ -868,6 +876,9 @@ impl ansi::Handler for GridHandler {
                 // grid has been cleared.
                 self.clear_secrets();
                 self.clear_displayed_rows_and_filter_matches();
+                // Same reason as in `reset_state`: don't carry an unclosed
+                // OSC 8 link past a full grid clear.
+                self.active_hyperlink_id = None;
```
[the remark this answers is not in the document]
th1nkful: Fixed — `ClearMode::All` on the primary screen now resets `active_hyperlink_id`, so output after `clear` / ESC[2J cannot inherit an unclosed hyperlink. Alt-screen redraws intentionally keep the active link (they manage their own spans and are not persisted/shared). `ClearMode::Saved` only clears scrollback, not the active write state, so it is left as-is. Covered by `test_full_grid_clear_drops_active_hyperlink`. Commit df5934aa (lands on next push).

crates/warp_terminal/src/model/grid/flat_storage/mod.rs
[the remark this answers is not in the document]
th1nkful: This is a false positive — the hyperlink change is recorded at the linked cell's offset, not the row start.

`offset += grapheme.len()` (mod.rs:260) runs for every non-spacer cell, **including empty ones**, and it runs *before* the `if needs_processing` backfill block below. So for a row like `[blank, blank, linked]`, by the time the linked cell pushes its `Some(id)` change (line 248), `offset` has already advanced past the two blanks — the change lands at the linked cell's offset. The backfill loop (263-269) then inserts the blank graphemes at the *earlier* offsets, where `hyperlink_id_map` still reads its `None` default. So the backfilled blanks rematerialize as non-clickable.

This is covered by `test_blank_cells_before_hyperlink_are_not_clickable` (flat_storage/mod_tests.rs), which builds exactly `[blank, blank, linked, blank]`, round-trips it through `FlatStorage`, and asserts the two leading blanks come back with `hyperlink_id() == None` while only the linked cell carries the id. @acarl005 reached the same conclusion on the earlier identical comment. No code change needed.

th1nkful:
@acarl005 this one is ready for re-review! :D the latest oz review only flagged the same false positive from earlier

app/src/terminal/view/tooltips.rs
```diff
@@ -229,7 +229,13 @@ impl TerminalView {
             }
 
             links.push(GridTooltipLink {
-                text: link.tooltip_text().to_owned(),
+                // OSC 8 links hide the destination behind arbitrary visible
+                // text, so show the URI itself in the tooltip; auto-detected
+                // URLs already show their destination as the visible text.
```
acarl005, line 233: This is 👍

app/src/terminal/view/link_detection.rs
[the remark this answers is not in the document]
acarl005: Ah this is a great suggestion. Thanks for doing that.

app/src/terminal/model/grid/ansi_handler.rs
```diff
@@ -848,11 +856,20 @@ impl ansi::Handler for GridHandler {
...
                 } else {
-                    self.clear_viewport();
+                    if self.full_grid_clear_behavior == FullGridClearBehavior::Clear {
+                        self.clear_visible_rows_in_place(bg);
+                    } else {
+                        self.clear_viewport();
+                    }
+                    // Don't carry an unclosed OSC 8 link past a full-screen clear
+                    // of the primary screen, so output after `clear` can't inherit
+                    // a stale URI.
```
acarl005:
```suggestion
```

I have similar remarks about over-commenting in this revision.

app/src/terminal/model/grid/ansi_handler.rs
```diff
@@ -869,6 +886,9 @@ impl ansi::Handler for GridHandler {
                 // grid has been cleared.
                 self.clear_secrets();
                 self.clear_displayed_rows_and_filter_matches();
+                // Same reason as in `reset_state`: don't carry an unclosed
+                // OSC 8 link past a full grid clear.
```
acarl005:
```suggestion
```

This is superfluous given that just 2 lines above we have already explained, "Clear out state that will no longer be valid now that the grid has been cleared."

app/src/terminal/model/grid/ansi_handler.rs
```diff
@@ -900,6 +920,11 @@ impl ansi::Handler for GridHandler {
 
         self.clear_secrets();
 
+        // Drop any unclosed OSC 8 hyperlink so a hostile sender can't leave a
+        // link open, force a terminal reset, and have subsequent unrelated
+        // output inherit the stale URI.
```
acarl005:
```suggestion
```

app/src/terminal/model/grid/ansi_handler.rs
```diff
@@ -160,6 +160,14 @@ impl ansi::Handler for GridHandler {
         report_error!("Handler method GridHandler::set_title should never be called. This should be handled by TerminalModel.");
     }
 
+    fn set_hyperlink(&mut self, hyperlink: Option<warp_terminal::model::ansi::Hyperlink>) {
+        // `and_then` is intentional: if `intern` returns None (registry cap
+        // hit), `active_hyperlink_id` stays None and subsequent `input(c)`
+        // writes plain non-clickable cells, so a cell's `hyperlink_id` never
+        // references an entry the registry doesn't hold.
```
acarl005:
```suggestion
```

app/src/terminal/model/grid/ansi_handler.rs
```diff
@@ -1598,6 +1624,10 @@ impl GridHandler {
         cursor_cell.fg = fg;
         cursor_cell.bg = bg;
         cursor_cell.flags = flags;
+        // Stamp the active OSC 8 hyperlink id (if any) onto the freshly-written
+        // cell. `set_hyperlink_id(None)` is a no-op when extra is None, so this
+        // only allocates `extra` when a hyperlink is actually active.
```
acarl005:
```suggestion
```
Don't explain a function's behavior at one of its *call sites*. This is already adequately explained in the doc comment.

app/src/terminal/model/alt_screen.rs
```diff
@@ -387,6 +395,12 @@ impl ansi::Handler for AltScreen {
         self.ansi_handler().input(c);
     }
 
+    fn set_hyperlink(&mut self, hyperlink: Option<warp_terminal::model::ansi::Hyperlink>) {
+        // AltScreen wraps a GridHandler that owns active_hyperlink_id and the
+        // per-screen HyperlinkRegistry — delegate.
```
acarl005:
```suggestion
```

None of the other delegator methods needs to explain delegation. This doesn't need to either.

app/src/terminal/model/block.rs
```diff
@@ -3123,6 +3123,11 @@ impl ansi::Handler for Block {
         delegate!(self.input(c));
     }
 
+    fn set_hyperlink(&mut self, hyperlink: Option<warp_terminal::model::ansi::Hyperlink>) {
+        // Delegate to BlockGrid -> GridHandler.
```
acarl005:
```suggestion
```

Similar comment to above. None of the other delegator methods needs to explain delegation. This doesn't need to either.

app/src/terminal/model/blockgrid.rs
```diff
@@ -728,6 +728,12 @@ impl ansi::Handler for BlockGrid {
         self.ansi_handler().input(c);
     }
 
+    fn set_hyperlink(&mut self, hyperlink: Option<warp_terminal::model::ansi::Hyperlink>) {
+        // Delegate to the inner `GridHandler` — that's where the active id
+        // is stored and where `input(c)` reads it.
```
acarl005:
```suggestion
```
Similar comment to above. None of the other delegator methods needs to explain delegation. This doesn't need to either.

app/src/terminal/model/blocks.rs
```diff
@@ -3611,6 +3643,13 @@ impl ansi::Handler for BlockList {
         delegate!(self.input(c));
     }
 
+    fn set_hyperlink(&mut self, hyperlink: Option<warp_terminal::model::ansi::Hyperlink>) {
+        // Delegate to the active block (or early-output handler) so the OSC 8
+        // active state lands on the same `GridHandler` that `input(c)` will
+        // stamp the cells on.
```
acarl005:
```suggestion
```
Similar comment to above. None of the other delegator methods needs to explain delegation. This doesn't need to either.

app/src/terminal/model/terminal_model.rs
```diff
@@ -2536,6 +2554,13 @@ impl ansi::Handler for TerminalModel {
         delegate!(self.set_cursor_shape(shape));
     }
 
+    fn set_hyperlink(&mut self, hyperlink: Option<warp_terminal::model::ansi::Hyperlink>) {
+        // Routes to the active AltScreen or BlockGrid, both of which delegate
+        // to their inner GridHandler — which owns the active id and the
+        // per-grid HyperlinkRegistry.
```
acarl005:
```suggestion
```
Similar comment to above. None of the other delegator methods needs to explain delegation. This doesn't need to either.

crates/warp_terminal/src/model/grid/cell.rs
```diff
@@ -276,7 +282,29 @@ impl Cell {
...
+        } else if let Some(extra) = self.extra.as_deref_mut() {
+            // Don't allocate `extra` just to clear an already-unset id.
+            extra.hyperlink_id = None;
+        }
+    }
+
+    /// Free all dynamically allocated cell storage. Preserves EndOfPromptMarker
+    /// if present. NOTE: this does NOT preserve `hyperlink_id` — that field is
+    /// content-bound and is cleared whenever the cell's content is reset
+    /// (erase/clear/reset_state).
```
acarl005, line 300:
```suggestion
```

acarl005 (changes requested):
Looks great! I think the actual code is all solid. I've got additional requests to remove some of the over-commenting in inline suggestions. Looks like there are also some lint errors to fix. Once those are handled then I'll approve.

By the way, you can skip the oz review. I'll approve after the requested changes.

acarl005:
@oz-agent Please perform the edits I requested in my most recent review

acarl005:
@oz-agent I asked you to delete some comments and you ended up doing other stuff in that last commit you pushed. You also added attributes to appease some lint rules. However, I didn't want you to do that. Please revert those. I ONLY wanted the edits I explicitly requested in my last review

crates/warp_terminal/src/model/grid/cell.rs
```diff
@@ -276,7 +282,28 @@ impl Cell {
...
+            self.extra.get_or_insert_with(Default::default).hyperlink_id = id;
+        } else if let Some(extra) = self.extra.as_deref_mut() {
+            extra.hyperlink_id = None;
+        }
+    }
+
+    /// Free all dynamically allocated cell storage. Preserves EndOfPromptMarker
+    /// if present. NOTE: this does NOT preserve `hyperlink_id` — that field is
+    /// content-bound and is cleared whenever the cell's content is reset
+    /// (erase/clear/reset_state).
```
acarl005, line 300:
```suggestion
```

This was supposed to get deleted.

crates/warp_terminal/src/model/grid/cell.rs
```diff
@@ -117,6 +117,12 @@ struct CellExtra {
     /// base character and zerowidth characters).
     cell_with_zero_width: Option<String>,
     end_of_prompt: Option<EndOfPromptMarker>,
+    /// OSC 8 hyperlink id. Resolves to a `Hyperlink { id, uri }` via the
+    /// owning grid's `HyperlinkRegistry`. Defaults to `None` (cell is not
+    /// part of a hyperlink span). `#[serde(default)]` so older payloads
+    /// without this field deserialize cleanly into `None`.
```
acarl005:
```suggestion
```
redundant comment

crates/warp_terminal/src/model/grid/flat_storage/row_iterator.rs
```diff
@@ -126,10 +129,13 @@ impl Iterator for RowIterator<'_> {
             }
 
             // If the grapheme takes up two cells, mark the following cell as
-            // a spacer.
+            // a spacer. The spacer carries the same hyperlink id as the wide
+            // glyph so both halves stay hoverable after rematerialization,
+            // matching how the live input path stamps both cells.
```
acarl005:
```suggestion
            // a spacer.
```
revert this addition

acarl005 (changes requested):
Requesting a bit more light comment cleanup.

acarl005:
@oz-agent Please apply the requested edits from my most recent review, which is just some light comment cleanup. Also, fix the CI failure for the WASM lint job, I've inlined the error here:

```
error[E0433]: failed to resolve: use of undeclared type `GridHighlightedLink`
   --> app/src/terminal/view/tooltips.rs:236:21
    |
236 |                     GridHighlightedLink::Hyperlink { uri, .. } => uri.clone(),
    |                     ^^^^^^^^^^^^^^^^^^^ use of undeclared type `GridHighlightedLink`
    |
help: consider importing this enum through its public re-export
    |
  3 + use crate::terminal::view::GridHighlightedLink;
    |

For more information about this error, try `rustc --explain E0433`.
```

Please do not make any other edits apart from what I've requested

acarl005:
@th1nkful Ok all the superficial changes are taken care of. Looks like there's just one failing test `test_osc8_copy_block_yields_visible_text_only`. Interestingly it only failed in the Linux job

th1nkful:
> @th1nkful Ok all the superficial changes are taken care of. Looks like there's just one failing test `test_osc8_copy_block_yields_visible_text_only`. Interestingly it only failed in the Linux job

@acarl005 I have pushed what should be a fix (I run on MacOS, so harder to verify). The cause appears to be that the test was sending CMD + C on MacOS but the helper it used sends CTRL + C on other OSes, which caused a cancel and not a copy.

acarl005 (approved):

acarl005:
@th1nkful thanks for the hard work! We'll dogfood this internally for a week at minimum and then I'll flip the feature flag once it's validated

dominik-stagtech:
Thanks for building this!
I am waiting eagerly for this to drop, so I can switch to warp. Not being able to click on my OSC links is the only thing stopping me from switching over right now.

PR #9855, opened 2026-05-01
fix: sync shell history with standard history files

mitre88:
Implements bidirectional sync between Warp command history and standard shell history files. Fixes issue #3422.


mitre88:
Closing this to keep my open PR queue focused on contributions that are currently reviewable and likely to proceed. This one has a blocking signal (changes-requested-cla) and would need a deeper rework before it is worth maintainer time. Thanks for the review/context.

PR #9864, opened 2026-05-01
Distinguish between agent conversation model update types

### LINKED ISSUE #10076: Distinguish between agent conversation model update types
Tracking issue for PR #9864.

Differentiates between types of agent conversation model updates to avoid costly full-list rebuilds in the agent management view.

lucieleblanc:
Closes #10076

## Description

The "Runs" agent management view rebuilds its entire card list whenever any conversation's status changes. Most of those events are visually no-ops (re-emitted statuses or status changes that don't cross the active filter), but subscribers can't tell because the event carries no detail.

This PR is the first of two and only changes the shape of the conversation status update event. Subscribers continue to behave exactly as before. The follow-up PR uses the new payload to skip rebuilds when possible.

Concretely:
- `BlocklistAIHistoryEvent::UpdatedConversationStatus` now carries the previous and new `ConversationStatus`. Restoration events report `prev_status: None` since restoration doesn't change the underlying status.
- `AgentConversationsModelEvent::ConversationUpdated` now carries a `conversation_id` and a typed `ConversationUpdateKind` of either `Restored` or `StatusSet { prev_filter, new_filter }`. Filter buckets are precomputed at emission time so subscribers don't have to recompute membership.
- A small helper `conversation_status_filter` maps `ConversationStatus` to its `StatusFilter` bucket.
- All existing subscribers were updated to the new event shape with `{ .. }` patterns and ignore the new fields. The variant is annotated with `#[allow(dead_code)]` until the follow-up PR lands.

## Testing

Added/updated four unit tests on the model: `Restored` emission, changing filter buckets, same-bucket re-emission, and the `ConversationStatus` --> `StatusFilter` mapping.

## Server API dependencies

No server dependencies.

### DIFF
--- a/app/src/ai/agent_conversations_model.rs
+++ b/app/src/ai/agent_conversations_model.rs
@@ -4,7 +4,9 @@ use crate::ai::agent::conversation::{AIConversationId, ConversationStatus};
 use crate::ai::ambient_agents::AmbientAgentTaskId;
 use crate::ai::ambient_agents::{AgentSource, AmbientAgentTask, AmbientAgentTaskState};
 use crate::ai::artifacts::Artifact;
-use crate::ai::blocklist::{format_credits, BlocklistAIHistoryEvent, BlocklistAIHistoryModel};
+use crate::ai::blocklist::{
+    format_credits, BlocklistAIHistoryEvent, BlocklistAIHistoryModel, ConversationStatusUpdate,
+};
 use crate::ai::cloud_environments::CloudAmbientAgentEnvironment;
 use crate::ai::conversation_navigation::ConversationNavigationData;
 use crate::auth::auth_manager::{AuthManager, AuthManagerEvent};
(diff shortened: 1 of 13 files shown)


app/src/ai/agent_conversations_model.rs
```diff
@@ -873,9 +872,38 @@ pub enum AgentConversationsModelEvent {
...
+    TaskManuallyOpened,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConversationUpdateKind {
+    /// The conversation was re-loaded into a terminal view.
+    Restored,
+    /// The conversation's status was set.
+    StatusSet {
+        prev_filter: StatusFilter,
```
liliwilson, line 890: Thoughts on storing `AgentRunDisplayStatus` here instead of `StatusFilter`?  We can use `from_conversation_status` to get it, and means we don't need a new `conversation_status_filter` method since `AgentRunDisplayStatus` already has a `status_filter`

app/src/ai/blocklist/history_model.rs
```diff
@@ -2109,6 +2096,11 @@ pub enum BlocklistAIHistoryEvent {
         conversation_id: AIConversationId,
         terminal_view_id: EntityId,
         is_restored: bool,
+        /// The conversation's status before this update, if known.
+        /// Restoration events do not have a previous status.
+        prev_status: Option<ConversationStatus>,
+        /// The conversation's status after this update.
+        new_status: ConversationStatus,
```
liliwilson: nit: what do you think about something like 
```rust
enum ConversationStatusUpdate {
    Restored,
    Changed { prev_status: ConversationStatus },
}

UpdatedConversationStatus {
  // ...
  update: ConversationStatusUpdate,
  new_status: ConversationStatus
}
```
since `prev_status` and `is_restored` are pretty coupled

liliwilson (approved):
Nice fix!

app/src/ai/blocklist/history_model.rs
lucieleblanc: I like this a lot! Changed, thanks

app/src/ai/agent_conversations_model.rs
lucieleblanc: Changed, thank you!

PR #9866, opened 2026-05-01
Handle agent management view updates based on event type

lucieleblanc:
## Description

The "Runs" agent management view was rebuilding its entire card list every time any conversation's status changed. While an agent is streaming, status events fire many times per second, and a recent heap profile attributed over 2 GB of allocations (plus ~700 MB of allocator churn) to this rebuild path.

Most of those events don't actually change the visible list: they re-emit the same status (typical during streaming) or change the status within the same active-filter bucket. In both cases the list doesn't need to rebuild — cards read their status fresh at render time, and on this path (local conversations) the action buttons don't depend on status.

This PR makes the agent management view branch on the typed payload introduced in #9864:
- **Restored:** no card-visible change. Only refresh the open details panel.
- **Status set crossing the active filter:** full rebuild, since membership in the visible list changed.
- **Status set in the same bucket** (or `All` filter): refresh the open details panel and re-render so the status icon picks up the new value. No list rebuild and no per-card work.

A small `StatusFilter::is_membership_crossed` helper centralizes the "did this status transition cross the active filter?" check.

This PR also drops the `conversation_id` field from the `ConversationUpdated` event, since no subscriber needs it now that the view doesn't target a specific card on this path.

## Testing

Started a conversation in the background. Opened the agent management view with the `All` filter selected and confirmed:
- The list doesn't re-sort while the agent is streaming.
- Toggling the status filter across the `Working`/`Done`/`Failed` boundary still updates the visible cards.

## Server API dependencies

No server dependencies.

## Agent Mode

CHANGELOG-IMPROVEMENT: Reduced memory usage and CPU work in the agent runs management view while a conversation is streaming.


app/src/ai/agent_management/view.rs
```diff
@@ -1258,6 +1259,95 @@ impl AgentManagementView {
...
+                    self.refresh_details_panel_if_needed(ctx);
+                    ctx.notify();
+                }
+            }
+        }
+    }
+
+    /// Update the action-buttons config for the card backing a given conversation, without
+    /// rebuilding the entire list. If no matching card is currently visible, this is a no-op.
+    fn update_card_for_conversation(
```
liliwilson: I don't actually think this function is necessary—we read the status for the card at render time, so if we haven't crossed a filter boundary, we shouldn't actually need to rebuild the card at all. A status change for a local conversation (which is the only type triggered from the `BlocklistAIHistoryModel`) won't change the action buttons available, so we should be good to just refresh the details panel in the else branch above

app/src/ai/agent_conversations_model.rs
```diff
@@ -99,6 +99,25 @@ pub enum StatusFilter {
     Failed,
 }
 
+impl StatusFilter {
+    /// Returns `true` if a status transition from `prev_bucket` to `new_bucket` flips
+    /// whether an item is included by this filter. `All` matches every bucket so it
+    /// is never crossed; the other variants are crossed when exactly one of the buckets
+    /// equals this filter.
+    pub(crate) fn is_membership_crossed(
```
liliwilson, line 109: Nice!

liliwilson (approved):
Thank you for fixing!! 🙌

app/src/ai/agent_management/view.rs
lucieleblanc: That makes sense, thank you!

app/src/ai/agent_management/view.rs
```diff
@@ -1258,6 +1259,95 @@ impl AgentManagementView {
...
+                let crossed_active_filter = self
+                    .filters
+                    .status
+                    .is_membership_crossed(prev_filter, new_filter);
+
+                if crossed_active_filter {
+                    self.get_tasks_from_model(ctx);
+                    self.refresh_details_panel_if_needed(ctx);
+                } else {
+                    self.update_card_for_conversation(conversation_id, ctx);
```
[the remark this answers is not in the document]
lucieleblanc: Fixing

PR #9560, opened 2026-04-30
fix: strip linked-worktree '+' marker from branch picker (#9170)

lonexreb:
## Description

Fixes #9170.

The shell git-branch chip in the prompt input shows a list of branches that the user can click to run `git checkout <branch>`. The chip's data source is the porcelain output of:

```sh
git --no-optional-locks branch --no-color --sort=-committerdate
```

`git branch` prefixes each line with one of three markers:

| Marker | Meaning |
| ------ | ------- |
| `*`    | Branch checked out in **this** worktree |
| `+`    | Branch checked out in **another linked worktree** |
| (space) | Every other branch |

`filter_git_branch_on_click_values` in `app/src/context_chips/current_prompt.rs` was stripping only the `*` marker. When a user clicked a `+`-prefixed entry, the value (still containing the `+`) was passed straight into `format!("git checkout {branch_name}")`, producing:

```sh
git checkout + 273-improvement-suggestion-agent
```

which fails:

```text
error: pathspec '+' did not match any file(s) known to git
error: pathspec '273-improvement-suggestion-agent' did not match any file(s) known to git
```

(matching the screenshots in the bug report exactly).

### Change

Single-line behavioral fix:

```rust
- .map(|s| s.trim_start_matches('*').trim().to_string())
+ .map(|s| s.trim_start_matches(['*', '+']).trim().to_string())
```

The function is also converted from `&self` to an associated `fn` (it never used `self`) so it can be tested directly without spinning up a full `CurrentPrompt` model. The single caller is updated from `self.` to `Self::`.

A `+`-marked branch is one that's already checked out in another worktree; running `git checkout` on it from the current worktree will still surface git's own (more meaningful) error. This change just stops Warp from generating a malformed command before git ever sees it.

## Testing

Added five unit tests in `app/src/context_chips/current_prompt_test.rs`:

- `filter_git_branch_strips_star_marker` — pre-existing behavior is preserved.
- `filter_git_branch_strips_plus_marker_for_linked_worktrees` — explicit regression for #9170, with a final assert that no value retains a leading `+` or `*`.
- `filter_git_branch_handles_mixed_marker_set` — full `git branch` sample (`*`, `+`, space) verified end-to-end including sort order.
- `filter_git_branch_passes_through_none` — `None` input returns `None`.
- `filter_git_branch_drops_blank_lines` — pre-existing behavior is preserved.

[...]

PR #9440, opened 2026-04-29
Strip linked-worktree '+' marker from git branch picker

tkshsbcue:
Fixes #9170.

The ShellGitBranch prompt chip's on-click menu is populated from the output of:

    git --no-optional-locks branch --no-color --sort=-committerdate

That command prefixes the current branch with '*' and any branch checked out
in another linked worktree with '+'. The on-click filter
(filter_git_branch_on_click_values) only stripped '*', so when a user picked
a '+'-marked branch the resulting command was literally:

    git checkout + 273-improvement-suggestion-agent

which fails with:

    error: pathspec '+' did not match any file(s) known to git
    error: pathspec '273-improvement-suggestion-agent' did not match any file(s) known to git

This PR strips both '*' and '+' so the real branch name is passed through to
git checkout. Git itself will still report a clear "is already checked out at
<path>" error if the user selects a branch held by another worktree, which is
the expected and useful failure mode.

What changed

app/src/context_chips/current_prompt.rs:

- Extracted the value-cleaning logic out of
  CurrentPrompt::filter_git_branch_on_click_values into a free
  filter_git_branch_values helper. This makes it directly unit-testable
  without standing up a CurrentPrompt model.
- Strip '+' in addition to '*' from the leading characters.
- Added a filter_git_branch_values_tests module covering:
    - linked-worktree '+' is stripped (the regression case)
    - current-branch '*' is stripped and promoted to the top
    - recency order is preserved for non-current branches
    - empty / whitespace-only lines are skipped
    - internal '+' / '*' characters in branch names are not touched
    - input made up entirely of '+'-marked branches still works
    - empty input returns empty

Testing

- cargo test -p warp filter_git_branch_values -> 7 passed, 0 failed
- cargo clippy -p warp --tests --lib -- -D warnings -> clean

Reproduction (before the fix)

1. Create a repo with a linked worktree (git worktree add ../other branch-x)
2. From the parent repo, open the branch chip in the prompt
3. Click on "+ branch-x"
4. The terminal runs "git checkout + branch-x" which fails

Manual test plan

- In a repo with linked worktrees, confirm the branch chip menu lists
  '+'-prefixed branches
- Selecting one runs "git checkout <name>" (no leading '+')
- If the branch is held by another worktree, git's standard
  "already checked out at" error surfaces
- Selecting '*'-marked / unmarked branches still works as before

PR #9264, opened 2026-04-28
fix: strip linked-worktree '+' marker from git branch chip output

JasonLovesDoggo:
## Description

Fixes #9170.

When a branch is checked out in another linked git worktree, `git branch --no-color` prefixes its line with `+ ` (instead of `* ` for the current branch, or `  ` for regular branches). The branch-chip on-click parser in `filter_git_branch_on_click_values` only stripped `*`, so clicking a `+`-marked entry inserted `git checkout + branchname` and failed with:

```
error: pathspec '+' did not match any file(s) known to git
error: pathspec '273-improvement-suggestion-agent' did not match any file(s) known to git
```

### Changes
- `app/src/context_chips/current_prompt.rs` — `filter_git_branch_on_click_values` now strips both `*` and `+` markers via `trim_start_matches(['*', '+'])`. Also dropped the unused `&self` so the helper is testable as an associated fn.
- `app/src/terminal/model/session.rs` — `git_branches_for_command_corrections` had the same bug class for typo-correction suggestions; fixed alongside.
- `app/src/context_chips/current_prompt_test.rs` — added two unit tests covering marker stripping with realistic `git branch --no-color --sort=-committerdate` output, plus None/empty edge cases.

The `BranchPicker` itself (`app/src/tab_configs/branch_picker.rs`) was not affected — it goes through `git for-each-ref --format=%(refname:short)`, which never includes the marker.

### Approach note
Per the issue's first suggested resolution, the marker is stripped rather than blocking checkout. If the branch is in fact checked out in another worktree, git's own error (`fatal: '<branch>' is already used by worktree at '<path>'`) is much clearer than the original `pathspec '+'` failure.

## Testing

- `cargo fmt` clean.
- New unit tests verify both the happy path (mixed `*`, `+`, and unmarked entries) and edge cases (None input, empty/whitespace-only lines).
- Manual verification of the reproduction conditions described in the issue is gated on a running build; the `git branch --no-color` output format used in the tests is taken verbatim from real git output for a repo with linked worktrees.

## Server API dependencies

No server API changes.

## Agent Mode

- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

## Changelog Entries for Stable

CHANGELOG-BUG-FIX: Fixed git branch chip click failing with "pathspec '+' did not match" when selecting a branch that is checked out in another linked git worktree.

PR #9905, opened 2026-05-02
Fix linked-worktree branch checkout

### LINKED ISSUE #9170: Branch picker includes linked-worktree '+' marker in git checkout
### Pre-submit Checks

- I have searched Warp bugs and there are no duplicates.
- I have searched Warp known issues page and my issue is not there.

### Describe the bug

When selecting a branch from Warp's branch picker/search UI, if that branch is checked out in another linked git worktree, Warp appears to include Git's leading `+` marker in the checkout command.

For example, the branch is shown as:

```bash
+ 273-improvement-suggestion-agent
```

Warp then appears to execute:

```bash
git checkout + 273-improvement-suggestion-agent
```

which fails with:

```bash
error: pathspec '+' did not match any file(s) known to git
error: pathspec '273-improvement-suggestion-agent' did not match any file(s) known to git
```

This looks like Warp is parsing raw `git branch` display output and treating the leading `+` linked-worktree marker as part of the checkout arguments.

### To reproduce

1. Create a repository with linked git worktrees.
2. Check out a branch in another worktree.
3. In Warp, open the branch picker/search UI from a different worktree of the same repo.
4. Select the branch that is shown with a leading `+`.
5. Warp attempts checkout and the command fails.

### Expected behavior

Warp should not pass the leading `+` marker to `git checkout`.

It should either:

- strip the marker and use only the real branch name, or
- prevent checkout with a clear message that the branch is already checked out in another linked worktree.

I have screenshots showing:

- the branch picker showing `+ 273-improvement-suggestion-agent`
- the resulting error after Warp attempts checkout:

```bash
error: pathspec '+' did not match any file(s) known to git
error: pathspec '273-improvement-suggestion-agent' did not match any file(s) known to git
```

Image

Image

### Operating system (OS)

macOS

26.3.1

### Shell Version

### Current Warp version

v0.2026.04.15.08.45.stable_02

### Regression

Unknown

### Recent working Warp date

No response

### Additional context

This repository uses linked git worktrees.

Local `git branch --list` output includes:

```bash
+ 263-catalog-page-allow-sorting
+ 269-adapt-ui-dataset-subset
+ 271-move-super-admin
+ 273-improvement-suggestion-agent
+ 274-static-config
+ 284-handle
```

The `+` is Git's marker for a branch checked out in another linked worktree, not part of the branch name.

kranthik10:
## Description

Teach the prompt branch picker to treat Git's `+` branch marker as a linked-worktree action, not a checkout target. `git branch` marks the current branch with `*` and branches checked out in another worktree with `+`; clicking a `+` branch should move the user into that worktree instead of running `git checkout <branch>`, which Git rejects.

The branch list command now also fetches `git worktree list --porcelain`, and Warp keeps the worktree path as hidden action data for `+` branches. Normal branches still run `git checkout '<branch>'`, quoted for shell safety. Linked-worktree branches run `cd '<worktree-path>'`; if the path cannot be resolved, Warp prints a clear message instead of attempting a checkout that is expected to fail.

## Linked Issue

Fixes #9170

## Screenshots / Videos

Demo showing the git chip with a linked worktree branch and selecting it to move into that worktree:

https://github.com/user-attachments/assets/5b31d204-9e13-4855-8977-cbd5bc0ad1bd

## Testing

- `cargo fmt -- --check`
- `git diff --check`
- `env TOOLCHAINS=com.apple.dt.toolchain.Metal.32023.864 cargo test -p warp test_format_git_branch_command_checks_out_normal_branch --lib`
- `env TOOLCHAINS=com.apple.dt.toolchain.Metal.32023.864 cargo test -p warp test_git_branch_on_click_value_round_trips_through_encode_decode --lib`
- Added parser coverage for resolving `+` branches to worktree paths
- Added `GitBranchOnClickValue` encode/decode round-trip coverage
- Added command formatting coverage for normal branch checkout quoting, linked-worktree branch clicks, and the missing-path fallback
- GitHub CI is passing

## Agent Mode

CHANGELOG-BUG-FIX: Fixed prompt branch picker clicks for branches checked out in linked worktrees. Thanks to @JasonLovesDoggo (#9264), @tkshsbcue (#9440), and @lonexreb (#9560) for earlier fixes and investigation.

### DIFF
--- a/app/src/context_chips/display_chip.rs
+++ b/app/src/context_chips/display_chip.rs
@@ -1789,8 +1789,8 @@ fn format_change_directory_command(dir_name: &str) -> String {
     format!("cd {}", shell_single_quote(dir_name))
 }
 
-pub fn format_git_branch_command(value: &str) -> String {
-    let branch = GitBranchOnClickValue::decode(value);
+pub fn format_git_branch_command(encoded_git_branch_on_click_value: &str) -> String {
+    let branch = GitBranchOnClickValue::decode(encoded_git_branch_on_click_value);
     if let Some(worktree_path) = branch.worktree_path {
         return format_change_directory_command(&worktree_path);
     }
@@ -1805,7 +1805,7 @@ pub fn format_git_branch_command(value: &str) -> String {
         );
     }
 
-    format!("git checkout {}", branch.branch_name)
+    format!("git checkout {}", shell_single_quote(&branch.branch_name))
 }
 
 pub(crate) fn chip_container(
--- a/app/src/context_chips/display_chip.rs
+++ b/app/src/context_chips/display_chip.rs
@@ -448,6 +448,15 @@ impl GitBranch {
     fn command(&self) -> String {
         format_git_branch_command(&self.0)
     }
+
+    fn icon_for_menu(&self) -> Icon {
+        let branch = GitBranchOnClickValue::decode(&self.0);
+        if branch.is_linked_worktree {
+            Icon::Dataflow02
+        } else {
+            Icon::GitBranch
+        }
+    }
 }
 
 impl GenericMenuItem for GitBranch {
@@ -460,7 +469,7 @@ impl GenericMenuItem for GitBranch {
     }
 
     fn icon(&self, _app: &AppContext) -> Option<Icon> {
-        Some(Icon::GitBranch)
+        Some(self.icon_for_menu())
     }
 
     fn action_data(&self) -> String {
--- a/app/src/context_chips/display_chip.rs
+++ b/app/src/context_chips/display_chip.rs
@@ -1782,7 +1782,7 @@ impl ActionButtonTheme for EnterAgentViewButton {
 }
 
 fn shell_single_quote(value: &str) -> String {
-    format!("'{}'", value.replace("'", "'\\'''"))
+    format!("'{}'", value.replace("'", "'\\''"))
 }
 
 fn format_change_directory_command(dir_name: &str) -> String {
--- a/app/src/context_chips/display_chip.rs
+++ b/app/src/context_chips/display_chip.rs
@@ -10,6 +10,7 @@ use crate::ai::{
 use crate::code::editor::{add_color, remove_color};
 use crate::code_review::code_review_view::CODE_REVIEW_TOOLTIP_TEXT;
 use crate::code_review::diff_state::DiffStats;
+use crate::context_chips::git_branch_on_click::GitBranchOnClickValue;
 use crate::context_chips::node_version_popup::{NodeVersionPopupEvent, NodeVersionPopupView};
 use crate::context_chips::spacing;
 use crate::settings::{AISettings, AISettingsChangedEvent, InputSettings};
@@ -443,13 +444,19 @@ pub struct DisplayChipConfig {
 #[derive(Debug, Clone)]
 pub struct GitBranch(String);
 
+impl GitBranch {
+    fn command(&self) -> String {
+        format_git_branch_command(&self.0)
+    }
+}
+
 impl GenericMenuItem for GitBranch {
     fn as_any(&self) -> &dyn std::any::Any {
         self
     }
 
     fn name(&self) -> String {
-        self.0.clone()
+        GitBranchOnClickValue::decode(&self.0).branch_name
     }
 
     fn icon(&self, _app: &AppContext) -> Option<Icon> {
@@ -557,7 +564,7 @@ impl DisplayChip {
                         };
 
                         ctx.emit(PromptDisplayChipEvent::TryExecuteCommand(
-                            format_git_branch_command(&git_branch.name()),
+                            git_branch.command(),
                         ));
                         me.close_git_branch_menu(ctx);
                         ctx.notify();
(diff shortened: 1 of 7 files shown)


unrevised6419:
Heads up: this PR strips the `+ ` marker but the underlying checkout will still fail because the branch is already used by a linked worktree (`fatal: '<branch>' is already checked out at ...`). See https://github.com/warpdotdev/warp/issues/9170#issuecomment-4363968730 for details and suggested fixes.

kranthik10:
Thanks for calling this out. You were right that just stripping the marker only changes the failure mode: Git will still reject checkout when the branch is already used by another worktree.

I updated this PR so the branch picker also reads `git worktree list --porcelain`. For `+` branches, Warp now keeps the linked worktree path as action data and emits `cd '<path>'` when that item is selected. Normal branches still use `git checkout <branch>`. If Git marks a branch with `+` but we can’t resolve the path, Warp now prints a clear message instead of attempting a checkout that should fail.

I added parser coverage for the worktree path mapping and command-formatting coverage for the `cd` behavior/fallback. Local `cargo test` is still blocked on my machine by the missing Apple Metal Toolchain, but `cargo fmt -- --check` and `git diff --check` pass.

kranthik10:
Quick update: I pushed a follow-up for the failed unit test.

The test caught that the fallback message was using an extra single-quote escape, which could also make the generated shell command invalid when the message contains words like couldn't. It now uses the normal POSIX escape for single quotes.

Checked locally:
- cargo fmt -- --check
- git diff --check
- a small zsh smoke test for the generated command

The focused Rust test still does not reach the assertion on my machine because this local build is missing protoc and the Xcode Metal Toolchain, so CI is the source of truth for the full test run.

app/src/context_chips/git_branch_on_click.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        match parts.next() {
+            Some(WORKTREE_TAG) => {
+                let worktree_path = parts
+                    .next()
+                    .filter(|path| !path.is_empty())
+                    .map(str::to_string);
+                Self::linked_worktree(branch_name, worktree_path)
+            }
+            _ => Self::new(value.to_string()),
```
moirahuang, line 62: should we be using `branch_name` instead of `value` here?

app/src/context_chips/display_chip.rs
```diff
@@ -1774,12 +1781,31 @@ impl ActionButtonTheme for EnterAgentViewButton {
...
+        return format!(
+            "echo {}",
+            shell_single_quote(&format!(
+                "Branch '{}' is already checked out in another worktree, but Warp couldn't find its path.",
+                branch.branch_name
+            ))
+        );
+    }
+
+    format!("git checkout {}", branch.branch_name)
```
moirahuang: should we use `shell_single_quote` here as well?

app/src/context_chips/display_chip.rs
```diff
@@ -1774,12 +1781,31 @@ impl ActionButtonTheme for EnterAgentViewButton {
...
+}
+
 fn format_change_directory_command(dir_name: &str) -> String {
-    format!("cd '{}'", dir_name.replace("'", "'\\'''"))
+    format!("cd {}", shell_single_quote(dir_name))
 }
 
-pub fn format_git_branch_command(branch_name: &str) -> String {
-    format!("git checkout {branch_name}")
+pub fn format_git_branch_command(value: &str) -> String {
```
moirahuang: let's use a more descriptive name than `value` here to make it clear that we're expecting a `GitBranchOnClickValue` encoded string

app/src/context_chips/display_chip_tests.rs
```diff
@@ -40,6 +42,36 @@ fn test_github_pr_chip_display_value_falls_back_to_raw_value() {
     );
 }
 
+#[test]
```
moirahuang, line 56: i know the normal `git checkout` case wasn't added by you but it'd be great if we can also add some tests here for the normal branch case

and also a test for `decode(x.encode()) == x` to guard against drift there would be great!

moirahuang (commented):
hi @kranthik10, really appreciate you making this PR and making the updated fix so that cd-ing works! 

1. do you mind updating the CHANGELOG to give attribution as well to other people who attempted a fix? The relevant authors and PRs:

- @JasonLovesDoggo https://github.com/warpdotdev/warp/pull/9264
- @tkshsbcue https://github.com/warpdotdev/warp/pull/9440
- @lonexreb https://github.com/warpdotdev/warp/pull/9560

2. in the PR description, in the testing section you say "Not applicable; this is a prompt-chip branch-name parsing fix." however, I think this has visible testing changes. mind recording a video showing the updated git chip with worktreees and switching to a worktree please?

unrevised6419:
Question: Will we still see the `+ <branch-name>` in the dropdown? I would expect the `+` to remain, so we know that is a workspace.

moirahuang:
> Question: Will we still see the `+ <branch-name>` in the dropdown? I would expect the `+` to remain, so we know that is a workspace.

My proposed behavior here is that we should strip `+` from the dropdown (I think that reads better) but swap out the icon for branch (first screenshot) to be the worktree icon (second screenshot) instead. This will have your intended effect of knowing what are worktrees vs. regular git branches, while visually looking better

Screenshot 2026-05-05 at 12 29 00 PM


Screenshot 2026-05-05 at 12 29 06 PM

unrevised6419:
@moirahuang with icon should be much better! Thanks!

kranthik10:
Updated the PR description with the demo video and the latest test results. CI is passing now too.

moirahuang:
> Updated the PR description with the demo video and the latest test results. CI is passing now too.

I'm seeing no movement in the demo video, only a static view of a Warp window. Also mind handling [this comment](https://github.com/warpdotdev/warp/pull/9905#issuecomment-4382356724) too please?
Screenshot 2026-05-05 at 1 49 55 PM

unrevised6419:
> I'm seeing no movement in the demo video

@moirahuang for me the video is working fine 🤔

moirahuang:
> > I'm seeing no movement in the demo video
> 
> @moirahuang for me the video is working fine 🤔

for some reason a refresh fixed it for me!

moirahuang:
@kranthik10 I'm going to loop in oz-agent to make the icon change so we can try to land this PR soon!

@oz-agent please update the git branch icon in the git chip to be the worktree icon if the branch is tied to a worktree

moirahuang:
@oz-agent please update the git branch icon in the git chip to be the worktree icon if the branch is tied to a worktree

moirahuang:
here's with the worktree icons!
Screenshot 2026-05-07 at 6 06 17 PM

i'm going to try to resolve the conflicts and merge this PR

moirahuang (approved):
@kranthik10 thanks for your work on this PR!

Issue #2618, opened 2023-02-07
Set warp theme in launch configuration 


2kohm:
### Discord username (optional)

Yama#1576

### Describe the solution you'd like?

```text
I would like to select different warp themes for launch configurations under `.warp/launch_configurations`


# Warp Launch Configuration for TEF PROD
---
name: DEV
windows:
  - tabs:
      - title: DEV
        layout:
          - theme: MyWarpThemeName
          cwd: /
          commands:
            - exec: clear
            - exec: ssh connect-dev
```

### Is your feature request related to a problem? Please describe.

```text
No
```

### Additional context

This would help working with PROD and DEV systems an have a clear and different theme per config <3 

### How important is this feature to you?

3

### Warp Internal (ignore) - linear-label:770f6576-d6c0-4e4f-a259-fc64b5156087

None

dannyneira:
Thanks for this, we'll add any updates to the status on this thread.

Issue #21, opened 2021-07-17
Closing tab with active process quits Warp entirely


ccpost:
**Describe the bug**
Closing any tab with a process that's still controlling the terminal causes the entire Warp app to quit.

**To Reproduce**
1. Open a new tab
2. Run a program that keeps control of the terminal (ex: `man ls` / `vi` / etc)
3. Close the tab through either the X button or Cmd+W
4. The entire Warp app quits

**Expected behavior**
Expect all the tab's children to be terminated, but Warp to stay running.

Bonus: might be worth considering an "undo close" grace period like iTerm (https://gitlab.com/gnachman/iterm2/-/issues/2877#note_335567160), though that'd be a separate FR.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS Version: macOS 10.15.7

**Additional context**
N/A

donaldguy:
Duplicate of #18 

Hi, Chris 👋

ccpost:
LGTM on the 2021.07.21 release; thanks team!

(Hi Donald! 👋 )

Issue #18, opened 2021-07-16
Warp crashes on `exit` with multiple tabs open


lucieleblanc:
**Describe the bug**
When I open multiple tabs, typing `exit` on the terminal prompt crashes Warp.

**To Reproduce**
1. Open Warp.
2. Open a new tab.
3. In either tab, type `exit` at the prompt.

**Expected behavior**
When I type `exit`, I expect only the current tab to close.

**System info:**
 - OS: `MacOS 11.4`
 - Shell: `zsh`
 - Warp version: `0.1.0`

**Additional information**
I'm happy to send over the full stack trace if needed.

alokedesai:
Thanks for reporting! We already have a fix for this in, it will be in the next release that will go live next week

lucieleblanc:
Got it, thanks for the super fast reply!

zheng:
The fix was in our last release! :)

Issue #13, opened 2021-07-16
Missing dylib on startup (Mac OS X 10.13.6 - High Sierra)


andrasio:
Looks like the `/System/Library/Frameworks/UserNotifications.framework/Versions/A/UserNotifications` dylib isn't there.

Here is the report:

```Process:               stable [89506]
Path:                  /Applications/Warp.app/Contents/MacOS/stable
Identifier:            dev.warp.Warp-Stable
Version:               0.1.0 (20210714.005705)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           stable [89506]
User ID:               501

Date/Time:             2021-07-16 14:40:36.372 -0500
OS Version:            Mac OS X 10.13.6 (17G14042)
Report Version:        12
Anonymous UUID:        EFF49F03-6ED0-347A-6DF5-560AB4793A35

Sleep/Wake UUID:       2A0FC56C-3277-4CD2-B852-474976BF7AE7

Time Awake Since Boot: 300000 seconds
Time Since Wake:       28000 seconds

System Integrity Protection: enabled

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    DYLD, [0x1] Library missing

Application Specific Information:
dyld: launch, loading dependent libraries

Dyld Error Message:
  Library not loaded: /System/Library/Frameworks/UserNotifications.framework/Versions/A/UserNotifications
  Referenced from: /Applications/Warp.app/Contents/MacOS/stable
  Reason: image not found

Binary Images:
       0x1062bb000 -        0x107f82fcf +dev.warp.Warp-Stable (0.1.0 - 20210714.005705) <4A6EA122-961B-3EB9-BEDF-4E509C11AD8A> /Applications/Warp.app/Contents/MacOS/stable
       0x10910a000 -        0x109169ff7 +io.sentry.Sentry (7.0.3 - 7.0.3) <77709E89-C368-3369-9CD7-6408F63CC857> /Applications/Warp.app/Contents/Frameworks/Sentry.framework/Versions/A/Sentry
       0x1123e7000 -        0x112431adf  dyld (551.5) <CB9BFB56-4511-36F1-A546-891FF770C01C> /usr/lib/dyld
[...]

PR #9910, opened 2026-05-02
Spec: per-tab theme overrides driven by directory and launch configurations (GH478)

### LINKED ISSUE #478: Theming: Allow for different theme selections per tab

```shell
Similar to the native Mac OS terminal, it would be useful to allow for each tab in the Warp app to have it's own theme.  This would allow for quick visual differentiation between tabs. It appears that right now all theming is done across the entire application.
```

```shell
The only way to differentiate between tabs is based on the tab order, which does not provide a visual cue for what the tab is being used for.
```

### Additional context

_No response_

gulsahsarsilmaz:
## Description

Spec PR for [#478](https://github.com/warpdotdev/warp/issues/478) — *Allow for different theme per session, agent conversation, ssh, etc.* — which is labeled `ready-to-spec`. Per [`CONTRIBUTING.md`](https://github.com/warpdotdev/warp/blob/master/CONTRIBUTING.md), feature requests at this stage need a product + tech spec under `specs/GH<issue>/` before any code change.

This PR adds:

- `specs/GH478/product.md` — product spec with 15 numbered testable invariants, scope/non-scope, user-visible failure modes, and two open questions.
- `specs/GH478/tech.md` — tech spec grounded in real file:line references, structured per `CONTRIBUTING.md` (Context → Proposed changes → Testing and validation → Risks → Follow-ups).

### Scope of the proposal

Adds an optional `theme:` field to launch-configuration tab and window templates, persisted on `TabSnapshot`, and resolved at render time through a new `Appearance::theme_for(Option<&ThemeKind>)` accessor that falls back to the global theme when no override is present. Window chrome stays on the global theme; terminal cells and in-tab UI follow the override. Saving a layout round-trips overrides, with window-level coalescing when every tab in a window shares the same override.

### What this spec deliberately does **not** propose

These items are raised in the `#478` thread but are scoped out so this spec stays focused; each is listed as a follow-up that *consumes* the override field this spec introduces:

- Auto-theme by SSH host / hostname / cwd (`stevenchanin`, `pyronaur`, `zethon`, `janderegg`).
- Runtime escape-code or shell-hook protocol for setting a tab's theme (`yatharth`, for Claude Code session signaling).
- In-app per-tab theme picker submenu (open question in the product spec).
- Per-pane theming and per-tab wallpapers (`scottaw66`, `SheepDomination`).

### Relationship to #2618

[#2618](https://github.com/warpdotdev/warp/issues/2618) (*Set warp theme in launch configuration*) describes the same feature at a narrower scope and would close as a duplicate of #478 once this spec lands.

## Linked Issue

Closes #478 (spec-only — implementation will follow on this PR or a linked code PR per `CONTRIBUTING.md`).

> Spec-only PR: no implementation yet, so no screenshots. Visual verification is enumerated in `tech.md` under *Manual verification* and will be attached on the implementation PR.

## Screenshots / Videos

N/A — spec-only PR.

## Testing

No code changes in this PR. The tech spec's *Testing and validation* section enumerates the unit tests (launch-config YAML round-trip, `TabSnapshot` persistence, `Appearance::theme_for` fallback semantics), integration tests under `crates/integration/` (multi-tab launch configuration, global-theme change scoping, "Reset theme" menu, session-restore), and manual verification artifacts that the implementation PR will need to satisfy.

## Agent Mode

### DIFF
--- a/specs/GH478/product.md
+++ b/specs/GH478/product.md
@@ -195,10 +195,11 @@ bit-for-bit identical to today.
     so that reopening the saved configuration produces the same effective
     themes (modulo cwd matching, which re-runs against the current
     `directory_overrides`). Specifically, for each tab the *preserved
-    override* is the tab's manual pin if any, otherwise the tab's
-    window-level default if any; tabs whose effective theme came only
-    from directory matching have no preserved override. The save rule
-    is then:
+    override* is the first non-empty value of, in priority order: the
+    tab's menu pin (layer #1), the tab's launch-configuration manual
+    pin (layer #2), or the tab's window-level default (layer #4). Tabs
+    whose effective theme came only from directory matching (layer #3)
+    have no preserved override. The save rule is then:
 
     - If every tab in the window has the same preserved override
       `Some(X)` and no tab has a manual pin different from the others,
@@ -317,18 +318,24 @@ bit-for-bit identical to today.
 
 ### Accessibility
 
-21. The override does not change any text content, accessible labels, or
-    focus order. Screen readers continue to report tab titles and
-    contents identically. The new menu entries have accessible labels
-    "Pin theme" (with a submenu) and "Reset theme".
+21. The override does not change any text content, accessible labels,
+    or focus order. Screen readers continue to report tab titles and
+    contents identically. The three new right-click menu entries have
+    accessible labels "Pin theme" (with a submenu of theme names),
+    "Reset theme", and "Forget launch config theme"; each is announced
+    as a menu item, and each follows the menu's existing visibility
+    rules (Reset theme appears only when the tab has a menu pin;
+    Forget launch config theme appears only when the tab has a
+    launch-config manual pin).
 
 ## User-visible failure modes
 
 - **Unknown theme name** — anywhere it appears, the entry is skipped at
   apply time, a one-line warning is written to the Warp log identifying
-  the source (launch configuration filename + tab title or index;
-  `directory_overrides` key), and the rest of the configuration loads
-  normally.
+  the source (launch configuration filename + tab title or index for
+  launch-config sources; **redacted entry identifier** for
+  `directory_overrides` sources, never the raw path key — see #7b),
+  and the rest of the configuration loads normally.
 - **Custom theme file missing** — same fallback the global theme uses
   today: tab opens with the next-layer theme, warning logged.
 - **Two `directory_overrides` keys are equivalent after tilde expansion**
--- a/specs/GH478/product.md
+++ b/specs/GH478/product.md
@@ -1,0 +1,184 @@
+# PRODUCT.md — Per-tab theme override via launch configurations
+
+Issue: https://github.com/warpdotdev/warp/issues/478
+Related: https://github.com/warpdotdev/warp/issues/2618 (set warp theme in launch configuration)
+
+## Summary
+
+Today the Warp theme is a single global value (`appearance.themes.theme` in
+`settings.toml`). Switching themes affects every open window and tab at once.
+Users who keep many tabs open across distinct contexts — different projects,
+local vs. remote machines, production vs. development environments — have asked
+for years (`#478`, 55+ upvotes; `#2618`) for a way to give specific tabs a
+different theme so context is visible at a glance.
+
+This spec covers a focused first cut of that capability: a tab can carry a
+**theme override** that comes from the launch configuration that opened it (or
+from a tab restored from a previous session). The override scopes the theme to
+that single tab; the global theme setting is unchanged and continues to apply
+to every tab that does not have an override. A window-level default override is
+also defined so a launch configuration can theme all of its tabs at once.
+
+This spec deliberately scopes out automatic theme switching (SSH host
+detection, agent conversation type, escape-code-driven runtime changes). Those
+appear in `#478` discussion and warrant their own product spec once the
+override mechanism this spec defines exists to build on.
+
+Figma: none provided.
+
+## Goals / Non-goals
+
+In-scope surfaces:
+
+- The launch configuration YAML schema (`~/.warp/launch_configurations/*.yaml`)
+  gains an optional theme field at the tab level and at the window level. The
+  values are theme identifiers — the same names users already see in the theme
+  picker (`"Dark"`, `"Solarized Dark"`, `"Dracula"`, `"Dark City"`, etc.) and
+  the same custom-theme references the rest of the theme system already
+  accepts.
+- Tabs opened from a launch configuration that specifies a theme render with
+  that theme: terminal background, foreground, ANSI palette, and any
+  theme-derived UI surfaces inside the tab follow the override.
+- The override persists with the tab so that on restart / session restore the
+  tab continues to render with the same theme without depending on the user
+  re-running the launch configuration.
+- A user can clear a tab-level override (returning the tab to the global theme)
+  through the same right-click tab menu that already exposes per-tab
+  attributes.
+
+Out of scope (explicitly **not** part of this spec):
+
+- Automatic theming based on SSH host, hostname, working directory, or
+  `whoami` (raised in `#478` comments). These require separate detection and
+  policy mechanisms and are tracked as follow-ups.
+- A new escape-code or shell-side protocol for setting a tab's theme at
+  runtime (raised in `#478` comments by users wanting hooks to color Claude
+  Code sessions). Once a per-tab override field exists internally a runtime
+  setter is a small follow-up; defining the protocol is its own surface.
+- Per-pane theming. Panes inside a tab continue to share one theme; the
+  override is a tab-level concept.
+- Changes to the global theme storage path
+  (`appearance.themes.theme` in `settings.toml`), the theme picker UI, or
+  custom-theme loading. The override reuses the existing theme identifier type
+  unchanged.
+- Onboarding or settings-page surfacing of the new field. Discoverability lives
+  in the launch configuration docs and the right-click tab menu only.
+
+## Behavior
+
+1. A launch configuration YAML file may include a `theme:` field on any tab
+   entry. The accepted value is a theme identifier exactly as it appears in
+   `settings.toml` today (e.g. `theme: "Dark City"`, `theme: "Solarized Dark"`,
+   `theme: "Dracula"`). Custom theme references use the same form the global
+   theme setting accepts. Omitting the field leaves the tab using whatever
+   theme would otherwise apply (per behavior #3).
+
+2. A launch configuration YAML file may include a `theme:` field on any window
+   entry. When present, every tab in that window with no tab-level `theme:`
+   inherits this window-level value. A tab-level `theme:` always wins over a
+   window-level one.
+
+3. Theme resolution for a tab is, in order: (a) the tab's own override, if
+   any; (b) the window-level override of the window that opened the tab, if
+   any and inherited at open time; (c) the global theme as derived from
+   `ThemeSettings` and the system theme, exactly as today. If none of the
+   override sources apply the tab's behavior is bit-for-bit identical to a
+   tab opened without this feature.
+
+4. When a tab has an override the override applies to: the terminal cell
+   foreground/background, the ANSI 16-color palette used by the terminal grid,
+   and any in-tab UI surfaces whose colors are derived from the active theme
+   (block backgrounds, command output styling, accent colors). The window
+   chrome (title bar, sidebar, settings views, the tab strip itself) continues
+   to follow the global theme so that windows holding mixed-theme tabs remain
+   visually coherent at the window level.
+
+5. Switching tabs is instant: there is no flash, no progressive paint, and no
+   redraw of unrelated tabs. The previously-active tab's content does not
+   change appearance because of the switch — only the rendering of the
+   newly-active tab reflects its (possibly different) theme. A user
+   alt-tabbing through ten tabs with different overrides sees each tab in its
+   own theme as it becomes active.
+
+6. Changing the global theme (via the theme picker, settings page, or system
+   light/dark switch) updates every tab that does **not** have an override.
+   Tabs with overrides are unaffected by the global change; their rendering
+   stays on the override value.
+
+7. The existing per-tab `color:` field on a tab template (the small colored
+   indicator next to the tab title) is independent of the new `theme:` field.
+   Both can be set on the same tab; both are honored. Neither implies the
+   other.
+
+8. A tab opened with a theme override surfaces a "Reset theme" entry in the
+   right-click tab context menu, alongside the existing per-tab attributes
+   (color, title, etc.). Choosing "Reset theme" clears the tab's override; the
+   tab immediately re-renders using whatever theme falls out of the resolution
+   order in #3 (typically the global theme). The menu entry is hidden for
+   tabs that have no override.
+
+9. Tab overrides survive session restore. A tab that had an override when the
+   user last quit Warp opens with the same override on relaunch. A tab with
+   no override opens with no override (i.e. it tracks the current global
+   theme).
+
+10. Saving a window's current state to a launch configuration (`File → Save
+    Layout as Launch Configuration` and equivalents) emits `theme:` entries on
+    each tab that has an override. Tabs without overrides emit no `theme:`
+    field. Window-level `theme:` is emitted only when every tab in the window
+    shared the same explicit override at save time, in which case the per-tab
+    fields are coalesced up to the window. (This keeps round-trip save/load
+    clean for the common "themed launch config" case.)
+
+11. An unknown theme identifier in a launch configuration (e.g. a theme that
+    has been deleted, or a typo) is treated the same way an unknown global
+    theme is treated today: the tab falls back to the global resolved theme,
+    a warning is logged, and opening the launch configuration does not fail.
+    Other tabs in the same launch configuration are unaffected.
+
+12. Custom themes referenced by an override behave identically to custom
+    themes used as the global theme — they are loaded from the user's themes
+    directory, fail-soft to the global theme if the file is missing, and
+    obey the same trust/validation rules already in the theme loader.
+
+13. The override persists per-tab, not per-launch-configuration. If a user
+    edits a launch configuration's theme after first opening it, already-open
+    tabs that came from that launch configuration keep their original
+    override; new tabs opened by re-running the launch configuration get the
+    new value. This matches how the existing per-tab `color:` field behaves.
+
+14. Accessibility: the theme override does not change any text content,
+    accessible labels, or focus order. Screen readers continue to report tab
+    titles and contents identically. The "Reset theme" menu entry has the
+    accessible label "Reset theme" and is announced as a menu item.
+
+15. The feature applies on every supported platform (macOS, Linux, Windows).
+    It is gated behind no feature flag and is on by default once shipped.
+
+## User-visible failure modes
+
+- **Unknown theme name in YAML** — tab opens with the global theme; a one-line
+  warning is written to the Warp log identifying the launch configuration
+  filename, the tab title (or index, if untitled), and the unrecognized theme
+  string. The launch configuration as a whole still opens.
+- **Custom theme file missing** — same fallback as the global theme exhibits
+  today: tab opens with the global theme, warning logged.
+- **Window-level theme set, tab-level not set, tab opens with override
+  inherited; user later edits the YAML to remove the window-level value** —
+  already-open tabs keep their existing override (per #13). New tabs opened
+  from the edited launch configuration get no override.
+
+## Open questions
+
+- Should the right-click "Reset theme" entry also offer a submenu of theme
+  names so a user can change a tab's override without editing YAML? This spec
+  takes the more conservative position of "reset only" so the launch
+  configuration file remains the single source of truth for setting overrides;
+  a richer in-app picker is a candidate follow-up.
+- Should saving a launch configuration record the *resolved* theme of each
+  tab, or only the explicit override the tab carries? This spec records only
+  explicit overrides (#10) so a saved launch configuration that contains no
+  `theme:` fields continues to track the global theme on every machine it is
+  loaded on. The alternative (record resolved theme always) would lock the
+  saved configuration to a snapshot of the user's current global theme, which
+  is rarely the desired behavior for a portable launch configuration.
--- a/specs/GH478/product.md
+++ b/specs/GH478/product.md
@@ -131,6 +131,16 @@ bit-for-bit identical to today.
    skipped for matching purposes, and the rest of the map continues to
    work.
 
+7a. **`directory_overrides` is stored locally and never synced to Warp's
+    cloud.** Directory paths can encode employer, customer, and project
+    names (`~/Work/<client>/<engagement>/...`); cloud-syncing the keys
+    would push that organizational context off-machine. Users on
+    multiple machines who want shared themes today set them per-machine.
+    An opt-in cloud-sync mode is a candidate follow-up. The global
+    theme setting (`appearance.themes.theme`) and per-tab pins set via
+    the right-click menu remain user-controllable surfaces; only this
+    map is local-only.
+
 ### Launch-configuration overrides
 
 8. A launch configuration YAML may include `theme:` on any tab entry.
@@ -165,11 +175,15 @@ bit-for-bit identical to today.
     directory, fail-soft to the next layer if the file is missing, same
     trust/validation rules as the global theme loader.
 
-13. The override persists per-tab through session restore. A tab whose
-    effective theme came from a manual pin keeps that pin on relaunch. A
-    tab whose effective theme came from directory matching is restored
-    with no manual pin; on relaunch its theme is recomputed from the
-    current `directory_overrides` and the restored cwd.
+13. Overrides persist per-tab through session restore, by source:
+    - **Manual pin** — kept on relaunch.
+    - **Launch-configuration window-level default** — kept on relaunch.
+      The launch configuration that opened the tab is not necessarily
+      reopened on restore, so the value travels with the tab.
+    - **Directory match** — not stored; recomputed on relaunch from the
+      current `directory_overrides` and the restored cwd. Editing
+      `directory_overrides` between sessions therefore takes effect on
+      the next launch.
 
 14. The accepted form for any theme reference (in
     `directory_overrides`, in launch-config tab `theme:`, in launch-config
--- a/specs/GH478/product.md
+++ b/specs/GH478/product.md
@@ -101,9 +101,27 @@ bit-for-bit identical to today.
    `appearance.themes.theme` (the parser tolerates both display names like
    `"Dark City"` and snake-case like `"dark_city"`; see #14 below).
 
-2. Keys are tilde-expanded to absolute paths at match time. Trailing slashes
-   are normalized away. Symlinks are not resolved — the cwd as the shell
-   reports it is what's matched.
+2. Keys are tilde-expanded to absolute paths at match time. Trailing
+   slashes are normalized away. Symlinks are not resolved — the cwd as
+   the shell reports it is what's matched. Path normalization rules,
+   per platform:
+
+    - **Linux:** matching is case-sensitive (matches the filesystem
+      semantics on standard ext4/xfs). Path separators are `/`.
+    - **macOS:** matching is case-insensitive (matches the default
+      HFS+/APFS case-insensitive setting). Path separators are `/`.
+    - **Windows:** matching is case-insensitive. Both `/` and `\` are
+      accepted as separators in `directory_overrides` keys and are
+      normalized internally to a single canonical form before
+      comparison. Drive letters are normalized to uppercase (`c:\Work`
+      and `C:\Work` are equivalent keys; the second-written one wins
+      per TOML duplicate-key semantics). Tilde expands to
+      `%USERPROFILE%`.
+
+   Component-boundary matching (the rule in the previous paragraph
+   that prevents `~/Work/medone` from matching `~/Work/medone-archive`)
+   uses the platform's path-component definition — on Windows that
+   means the boundary follows either `\` or `/` after normalization.
 
 3. Match resolution: a key matches if it is a prefix of the active pane's
    cwd at a path-component boundary. `~/Work/medone` matches both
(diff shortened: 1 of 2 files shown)


gulsahsarsilmaz:
@cla-bot check

gulsahsarsilmaz:
Thanks for the review @oz-for-oss. Pushed v2 (7460929) addressing all three findings, plus a scope expansion that I should flag explicitly.

### Oz findings, addressed

1. **`ThemeKind` serde vs. product spec's display-name promise.** The `theme:` field on `TabTemplate` and `WindowTemplate` is now `Option<String>`, not `Option<ThemeKind>`. A new `resolve_theme_ref(&str) -> Option<ThemeKind>` helper (specced in `tech.md` §2) accepts both display form (`"Dark City"`) and snake_case (`"dark_city"`), case-insensitive, whitespace-tolerant. The YAML surface is now decoupled from the enum's internal representation.

2. **Unknown theme handling.** Direct consequence of #1 — deserialization never fails on theme strings. Resolution runs at apply time; an unresolved name logs one warning identifying the source (launch config + tab, or `directory_overrides` key) and the affected entry falls through to the next resolution layer. Other tabs in the same file and other entries in the same map are unaffected. Pinned in product behavior #11 and in the integration tests under `tech.md` *Testing and validation*.

3. **`Appearance::theme_for` ownership / lifetime.** `Appearance` now owns a lazy `theme_cache: parking_lot::RwLock<HashMap<ThemeKind, Arc<WarpTheme>>>` alongside the existing global theme. `theme_for(Option<&ThemeKind>) -> Arc<WarpTheme>` returns the global `Arc` clone for `None` or for the active global theme; otherwise a hit on the cache or a load via the existing global-theme loader path on miss. Invalidation: `AppearanceManager`'s existing `ThemeSettings` subscription clears the cache; the existing custom-theme file watcher does the same. Specced in `tech.md` §5 with the full type signatures and cache rules.

### Scope change since v1

v1 limited the user-visible surface to launch-configuration-set overrides and listed cwd-driven theming as a follow-up. v2 promotes cwd-driven theming to in-scope — that's the path most users in #478 actually describe ("cd into project A → theme A").

Mechanism in v2:

- New settings group `[appearance.themes.directory_overrides]`: directory path → theme name, longest-prefix wins, tilde-expanded, component-boundary match.
- `theme_override` on `TabSnapshot`/`TabData` is now `Option<ThemeOverride>` where `ThemeOverride` is `Manual(ThemeKind) | Cwd(ThemeKind)`. The variant exists so "Reset theme" can clear a manual pin without dismissing a still-applicable directory match.
- Resolution order (product spec): manual > directory match > launch-config window default > global.
- The cd-driven re-evaluator hooks into the existing focused-pane cwd event already used for tab title generation; no new shell-side plumbing.
- `Cwd(_)` overrides are not persisted across sessions; they re-derive on startup from current settings + restored cwd.

SSH-host detection and runtime escape-code protocols remain follow-ups; both consume the override field this spec defines.

PR title updated to reflect the broader scope.

/oz-review

gulsahsarsilmaz:
v3 pushed (3cad044) addressing all three findings.

### Critical: window default outranking cwd

You were right — expanding window-level launch-config theme into the `Manual` slot made it top priority, which contradicts the resolution order. Replaced the single `Option<ThemeOverride>` with a `TabThemeState` struct holding three independent slots:

```rust
pub struct TabThemeState {
    pub manual: Option<ThemeKind>,
    pub window_default: Option<ThemeKind>,
    pub cwd_resolved: Option<ThemeKind>,
}

impl TabThemeState {
    pub fn effective<'a>(&'a self, global: &'a ThemeKind) -> &'a ThemeKind {
        self.manual.as_ref()
            .or(self.cwd_resolved.as_ref())
            .or(self.window_default.as_ref())
            .unwrap_or(global)
    }
}
```

`effective()` walks slots in the product-spec priority order, so window default sits below cwd. "Reset theme" now only clears `manual`; if a cwd match was still applicable underneath, the tab redraws with the cwd theme rather than the global. Pinned by both unit tests on `effective()` and an integration test that opens a launch config with a window-level theme into a `directory_overrides`-matched cwd.

Persistence: `manual` and `window_default` are serialized; `cwd_resolved` is not (it's recomputed on startup from current settings + restored cwd). Window default has to persist because the launch config that set it isn't necessarily reopened on session restore.

### Important [SECURITY]: directory paths leaking off-machine

Right call. Switched to:

```rust
sync_to_cloud: SyncToCloud::Locally,
private: true,
```

Added an explicit *Privacy model* section in `tech.md` §4 documenting:

- Local-only storage (no sync of keys or values).
- No telemetry on map contents (the match function emits only a count metric, never path keys).
- No round-trip into shareable launch-config YAML — saved launch configs do not emit `directory_overrides`. Launch configs are designed to be shareable; this map is not.
- Opt-in cloud sync is a follow-up requiring its own spec covering UI, encryption-at-rest of keys, and admin-policy controls.

Also added user-visible product behavior #7a so the no-sync rule is part of the contract, plus three privacy invariant tests under *Privacy invariant tests* in the validation section.

### Important: settings edits re-evaluating open tabs

Missed this in v2. The directory-pattern matcher is now invoked from **three** places (`tech.md` §4):

1. Tab creation.
2. Focused-pane cwd-change events (existing tracker).
3. **`DirectoryThemeOverrides` settings-change events** — new subscription, walks every open tab in every window, recomputes `cwd_resolved` against the new map, redraws tabs whose effective theme changed. Pinned by an integration test that adds, edits, and removes keys while Warp is running.

The same subscription handles validation: any value that fails `resolve_theme_ref` is logged once with its key and treated as absent for matching. Edits that fix the value re-trigger validation.

### Other v3 cleanups

- Product spec #13 spells out per-source persistence explicitly (manual and window_default persist; cwd recomputes).
- Risks section adds three new entries: cwd_resolved persistence guard, path-key leak guard, and window-default-priority guard.
- Follow-ups updated: SSH-host detection would add a fourth `ssh_resolved` slot to `TabThemeState` and a corresponding rung in `effective()`.

/oz-review

specs/GH478/product.md
```diff
@@ -0,0 +1,279 @@
...
+for years (`#478`, 55+ upvotes; `#2618`) for tabs to render with different
+themes when they represent different contexts — different projects, local vs.
+remote machines, production vs. development.
+
+This spec covers a focused first cut of per-tab theme overrides driven by
+**three** sources, in priority order:
+
+1. A user-visible **manual** override on a tab (set via the launch
+   configuration YAML or via a right-click menu).
+2. A **directory-pattern** auto-match: the user maps directory paths to
```
zachlloyd, line 19: the other match that would be nice is based on hostname.  folks want different themes for ssh sessions

zachlloyd: lol, i see below that hostname is deliberately out of scope.  that's probably a good call for this PR.

specs/GH478/product.md
```diff
@@ -0,0 +1,279 @@
...
+      when the tab has a manual override. After reset, the tab falls
+      through to the directory-match / window-default / global layers in
+      that order; if a directory match applies, the tab redraws with the
+      directory-matched theme.
+
+19. The existing per-tab `color:` field on a tab template (the small
+    colored indicator next to the tab title) is independent of the new
+    theme override. Both can be set; both are honored.
+
+20. The feature applies on every supported platform (macOS, Linux,
```
zachlloyd, line 300: let's gate this behind a flag pls

zachlloyd: @captainsafia we should adjust our product spec to prefer adding flags for significant features

specs/GH478/product.md
```diff
@@ -0,0 +1,279 @@
...
+- **Two `directory_overrides` keys are equivalent after tilde expansion**
+  — last-write-wins per TOML semantics; a warning is logged identifying
+  the duplicate.
+- **A pane's cwd is unavailable** (non-shell pane content) — that pane
+  contributes no cwd to directory matching; if it is the focused pane the
+  tab falls through to the window/global layers.
+
+## Open questions
+
+- Should `directory_overrides` keys support glob patterns (`~/Work/*`)?
```
zachlloyd: i think prefix-match is fine to start

specs/GH478/product.md
```diff
@@ -0,0 +1,279 @@
...
+  This spec proposes prefix-matching only because that covers the
+  most-requested use case ("a folder of projects, each in a subfolder")
+  and avoids the edge cases of glob semantics on Windows paths. Globs are
+  a candidate follow-up.
+- Should "Reset theme" also clear a launch-config-set manual override
```
zachlloyd: i don't think it should

specs/GH478/tech.md
```diff
@@ -0,0 +1,673 @@
...
+The field is `Option<String>` rather than `Option<ThemeKind>` for two
+reasons that directly address Oz's first-pass concerns:
+
+- **Serde-compat (Oz concern 1).** `ThemeKind` derives `Deserialize`,
+  but its accepted string form (variant names / snake-case) is not
+  guaranteed to match the human-readable display strings the product
+  spec promises (`"Dark City"`, `"Solarized Dark"`). Going through
+  `String` decouples the YAML surface from the enum's internal
+  representation.
+- **Field-level fallback (Oz concern 2).** A direct `ThemeKind`
```
zachlloyd, line 115: is it possible that post serialization we convert this to a more constrained type than string?

specs/GH478/tech.md
```diff
@@ -0,0 +1,673 @@
...
+        supported_platforms: SupportedPlatforms::ALL,
+        // **Local-only.** Directory paths can encode employer, customer,
+        // and project names (e.g. `~/Work/<client>/<engagement>/...`).
+        // Cloud-syncing the keys would leak that organizational context
+        // off-machine; cloud-syncing the values without the keys is
+        // useless. The setting is therefore not synced. Per-tab themes
+        // remain user-controllable through the right-click "Pin theme"
+        // menu, which writes to the (non-synced) tab snapshot, not to
+        // this map. See *Privacy model* below.
+        sync_to_cloud: SyncToCloud::Locally,
```
zachlloyd: should this be SyncToCloud::Never?

zachlloyd: I forget what SyncToCloud::Locally means

zachlloyd:
@gulsahsarsilmaz this is looking pretty good.  left a few small comments, but moving to ready-to-implement so you can assign to Oz to build

gulsahsarsilmaz:
v4 pushed (28f79f1) addressing all four findings.

### Important: save layout dropping window defaults

Replaced v3's "only emit manual" save rule with a `preserved_override` helper:

```rust
fn preserved_override(state: &TabThemeState) -> Option<&ThemeKind> {
    state.manual.as_ref().or(state.window_default.as_ref())
}
```

`From<WindowSnapshot> for WindowTemplate` then:

- If every tab's preserved override is the same `Some(X)` and no manual differs, emit a single window-level `theme: X` and clear per-tab.
- Otherwise emit each tab's preserved override at the tab level; window-level `theme:` omitted.
- Cwd-matched themes are still never emitted (they re-derive on reopen).

Pinned by two new integration tests: window-level-coalescing round-trip, mixed-pinning case.

### Important: Windows path normalization

Added explicit per-platform rules in product spec #2 and tech spec §4:

- **Linux:** case-sensitive matching, `/` separator only.
- **macOS:** case-insensitive matching (matches HFS+/APFS default).
- **Windows:** case-insensitive, both `/` and `\` accepted as separators, drive letters normalized to uppercase, tilde expands to `%USERPROFILE%`.

Component-boundary matching uses `Path::components()` (not string operations), with `unicase::eq` for case-insensitive platforms. Cross-platform integration test matrix.

### Important [SECURITY]: log redaction

Added `redacted_key_id(raw_key) -> String` (6-hex-char FxHash, non-cryptographic, stable per key) and a *Diagnostic redaction* contract in tech §4: every warning, error, or diagnostic emitted by the directory_overrides machinery references entries by hash, never by raw path. The theme value is included verbatim (theme names are non-sensitive and needed for the user to grep their `settings.toml`).

Example warning text:

```
directory_overrides[hash=8a3f9c]: unknown theme "Drakula" — matching this entry will be skipped until the value is corrected
```

Pinned by a redaction test asserting that the warning for `~/Work/AcmeCorp/2026` → `"Drakula"` contains the hash and `"Drakula"` but no substring of the raw path. Product behavior #7b makes the rule contractual.

### Important: renderer call-site inventory

There are 862 `Appearance::*::theme()` call sites. Exhaustive enumeration in the spec would be unreviewable and would rot. Instead:

1. **Categorical rule + table** in tech §5 covering each top-level path (`app/src/terminal/**`, `app/src/{tab,root_view,voltron,modal,menu,search_bar,input_suggestions}.rs`, `app/src/settings/**`, `crates/onboarding/**`, `crates/ui_components/**`, `crates/editor/**`). Each entry marked per-tab or global with rationale.

2. **Custom lint `appearance_theme_in_tab_path`** that fails CI when a global `theme()` call appears under a per-tab path. The PR opens with the lint enabled and zero warnings. Future call sites added in per-tab paths are caught at PR time, not in retrospect.

The lint is the deliverable. Reviewers verify scope by reading lint config (one entry per top-level area), not by walking 862 lines. Stronger guarantee than enumeration: correct by construction, robust to refactors.

### Other v4 cleanups

- Risks section adds entries pointing at the test that pins each fix.
- Product behaviors added: #2 (per-platform path rules), #7b (diagnostic redaction); #10 (save rule) rewritten to spell out coalescing.

/oz-review

gulsahsarsilmaz:
> @gulsahsarsilmaz this is looking pretty good. left a few small comments, but moving to ready-to-implement so you can assign to Oz to build

Aww, thank you so much! I will clean some feedback up and get to it :)

specs/GH478/product.md
gulsahsarsilmaz: Done in v5 (4013e42) — feature is now gated behind `appearance.themes.per_tab_overrides`, default **off** in stable and **on** in dev/preview per the rollout pattern. Tech §7 enumerates exactly which surfaces are gated when off (matching, launch-config theme application, menu entries, theme catalog) and which are not (the lint and persisted state, so flipping the flag is non-lossy). Five flag-state integration tests added.

specs/GH478/product.md
gulsahsarsilmaz: Ack — moved from Open questions to a resolved record in v5. Prefix-only ships; globs are a clean follow-up that consumes the same matcher.

specs/GH478/product.md
gulsahsarsilmaz: Updated in v5. The `manual` layer is now split into two independently-clearable slots — `menu_pin` (cleared by Reset theme) and `launch_config_pin` (cleared only by a new Forget launch config theme entry). `TabThemeState` is now 4 slots (menu_pin > launch_config_pin > cwd_resolved > window_default), and the resolution order in product.md is correspondingly 5 layers. Tests cover menu vs launch independence: a tab with both menu_pin=Light and launch_config_pin=Dracula resolves to Light; Reset clears only menu_pin and the tab redraws Dracula; subsequent Forget then clears the launch_config_pin.

specs/GH478/tech.md
gulsahsarsilmaz: You're right — I invented `Locally`; the actual variants are `Globally`, `Never`, `PerPlatform`. Switched to `SyncToCloud::Never` in v5 (4013e42). The privacy-model section in tech §4 already documented the intent (no off-machine transmission of path keys); `Never` is the correct mechanical expression of that.

gulsahsarsilmaz:
v5 pushed (4013e42) addressing both Oz's v3 review and @zachlloyd's first-pass.

**Summary of changes in v5:**

| Source | Finding | v5 fix |
| --- | --- | --- |
| Oz v3 [IMPORTANT] | Save rule dropped window_default | `preserved_override()` now walks both manual slots before window_default; coalescing test added |
| Oz v3 [IMPORTANT] | Windows path semantics undefined | Per-platform rules (case-sensitivity, separators, drive-letter normalization) added to product spec #2; cross-platform test matrix |
| Oz v3 [IMPORTANT][SECURITY] | Local logs could leak path keys | `redacted_key_id()` helper + Diagnostic redaction contract; product behavior #7b makes it user-visible |
| Oz v3 [IMPORTANT] | Renderer call-site list deferred (862 sites) | Categorical rule + table by top-level path + custom lint `appearance_theme_in_tab_path` that fails CI on missed migrations |
| Zach | "let's gate this behind a flag pls" | Feature flag `appearance.themes.per_tab_overrides`, default off in stable / on in dev+preview, gated surfaces enumerated in tech §7, five integration tests |
| Zach | "i think prefix-match is fine to start" | Globs moved out of Open questions; prefix-only ships first |
| Zach | "i don't think it should" (Reset clearing launch-config pins) | Manual layer split into `menu_pin` and `launch_config_pin`; right-click menu gains "Forget launch config theme"; resolution order is now 5 layers |
| Zach | `SyncToCloud::Locally` → `SyncToCloud::Never` | Fixed (the variant I cited didn't exist); the real `Never` matches the privacy-model intent |

`TabThemeState` is now 4 slots (menu_pin > launch_config_pin > cwd_resolved > window_default), with `effective()` walking them in priority order. Tests pin every adjacent priority transition.

All 8 review threads resolved. Holding the last `/oz-review` retrigger in case another iteration produces further design questions.

specs/GH478/product.md
```diff
@@ -0,0 +1,359 @@
...
+9. A launch configuration YAML may include `theme:` on any window entry.
+   Tabs in that window with no tab-level `theme:` and no directory match
+   inherit this window-level value (per the resolution order: #3 sits
+   below #2).
+
+10. Saving a window's current state to a launch configuration preserves
+    every theme override that was *not* derived from directory matching,
+    so that reopening the saved configuration produces the same effective
+    themes (modulo cwd matching, which re-runs against the current
+    `directory_overrides`). Specifically, for each tab the *preserved
```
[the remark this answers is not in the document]
gulsahsarsilmaz: Fixed in v6 (22a8b7d). Product spec #10 now explicitly says "first non-empty value of menu_pin (layer #1), launch_config_pin (layer #2), or window_default (layer #4)." The tech-side `preserved_override()` helper already walked all three slots — this aligns the product text with that.

specs/GH478/product.md
```diff
@@ -0,0 +1,359 @@
...
+21. The override does not change any text content, accessible labels, or
+    focus order. Screen readers continue to report tab titles and
+    contents identically. The new menu entries have accessible labels
+    "Pin theme" (with a submenu) and "Reset theme".
+
+## User-visible failure modes
+
+- **Unknown theme name** — anywhere it appears, the entry is skipped at
+  apply time, a one-line warning is written to the Warp log identifying
+  the source (launch configuration filename + tab title or index;
```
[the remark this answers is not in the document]
gulsahsarsilmaz: Fixed in v6. Product failure mode now reads: "redacted entry identifier for `directory_overrides` sources, never the raw path key (see #7b)." Pinned by the diagnostic redaction tests in tech §6.

specs/GH478/product.md
```diff
@@ -0,0 +1,359 @@
...
+    - An empty `directory_overrides` map (the default) plus no
+      launch-config theme fields plus no pinned themes — even with the
+      flag on — is bit-for-bit identical to current behavior.
+
+### Accessibility
+
+21. The override does not change any text content, accessible labels, or
+    focus order. Screen readers continue to report tab titles and
+    contents identically. The new menu entries have accessible labels
+    "Pin theme" (with a submenu) and "Reset theme".
```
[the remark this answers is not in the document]
gulsahsarsilmaz: Done in v6 — accessibility behavior #21 now lists all three menu entries (Pin theme, Reset theme, Forget launch config theme) with their visibility rules and aria semantics.

specs/GH478/tech.md
```diff
@@ -0,0 +1,975 @@
...
+
+Helper, alongside the settings group:
+
+```rust
+/// Stable short hash of a `directory_overrides` key suitable for
+/// inclusion in user-facing warnings. **Not cryptographic.** Purpose
+/// is identification of an entry across runs without including the
+/// raw path. 6 hex chars (24 bits) is enough to disambiguate the
+/// O(10) entries a typical user will have.
+pub fn redacted_key_id(raw_key: &str) -> String {
```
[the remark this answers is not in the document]
gulsahsarsilmaz: Fair finding — the unsalted FxHash was dictionary-guessable from candidate paths. Switched in v6 to HMAC-SHA256 keyed by a per-installation salt: 32 bytes generated on first launch, stored in `~/.warp/redaction_salt` (mode 0600), never synced. The truncated 24-bit identifier is now non-reversible without the salt, which only ever exists on the originating machine. Cross-installation determinism is broken on purpose (different salts → different ids) so logs leaked across users cannot be correlated. Tests added: salt-file-mode, different-salt-different-id, missing-salt-fallback (warning shows `[unsalted]` rather than any derived id).

specs/GH478/tech.md
```diff
@@ -0,0 +1,975 @@
...
+  `define_settings_group!` macro emits change events the same way
+  `ThemeSettings` does (`app/src/appearance.rs:51` is the existing
+  pattern). A new subscription on `DirectoryThemeOverrides::handle(ctx)`
+  walks every open tab in every window, recomputes
+  `cwd_resolved` from the new map and the tab's current cwd, and
+  re-renders each tab whose effective theme changed. Implements
+  product behavior #6.
+
+The same subscription handles validation: any value in the new map
+that fails `resolve_theme_ref` is logged once (with its key) and
```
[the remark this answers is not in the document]
gulsahsarsilmaz: Fixed in v6. The validation path text now reads "logged once via the `redacted_key_id` helper (the raw key is **never** included in the warning)." The contract is uniform across the matcher, settings subscriber, and value validator.

specs/GH478/tech.md
```diff
@@ -0,0 +1,975 @@
...
+| `app/src/reward_view.rs`, `wasm_nux_dialog.rs`, `word_block_editor.rs` | global | full-window modals/views |
+| `crates/onboarding/**` | global | onboarding flow runs outside a tab |
+| `crates/ui_components/**` | global | generic widgets — they do not know about tabs |
+| `crates/editor/**` | global by default; per-tab when invoked from `terminal/view/**` (the call site, not the crate, decides) | editor is reusable |
+| `crates/integration/**` | tests; updated to assert both forms |
+
+**Migration enforcement — custom lint**
+
+To make the migration mechanical, exhaustive, and durable across future
+additions, the implementation adds a small `clippy_lints`-style check
```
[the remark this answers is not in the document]
gulsahsarsilmaz: Picked `dylint` in v6 — it is the standard for project-specific Rust lints (rust-fuzz, solana, others), runs against a normal cargo build, no toolchain fork required. CI hook: `cargo dylint --lib appearance_theme_in_tab_path -- -D warnings` after the existing clippy step in `./script/presubmit`. If the Warp team would rather not add dylint as a dependency, the spec lists a ripgrep-in-presubmit fallback that gives 90% of the guarantee with no new tooling; switching is a one-file change.

specs/GH478/tech.md
```diff
@@ -0,0 +1,975 @@
...
+| --- | --- |
+| `directory_overrides` matching | Skipped entirely. The settings group is still parsed (so users who set up the map under preview do not lose data on stable) but `directory_theme_for` is never invoked. |
+| Launch-config `theme:` fields | Deserialized as today (`Option<String>`) but ignored — `launch_config_pin` and `window_default` are not populated when applying templates. |
+| Right-click menu entries | "Pin theme...", "Reset theme", and "Forget launch config theme" are hidden. |
+| Theme catalog (§5) | `theme_for(None)` always returns the global theme; `theme_for(Some(_))` is unreachable because no slot is ever populated. The catalog cache stays empty. |
+| Settings-change subscription on `DirectoryThemeOverrides` | Subscribed but the handler short-circuits when the flag is off. |
+| `appearance_theme_in_tab_path` lint (§5) | Still enforced — the call sites need to be migrated regardless of runtime behavior because the flag must be flippable without recompiling. The migrated call sites pass `theme_state.effective(&global)` which falls through to `global` when every slot is `None`, which is exactly what the flag-off state guarantees. |
```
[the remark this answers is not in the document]
gulsahsarsilmaz: You are right — the table and the persistence section were inconsistent. Resolved in v6 by introducing a wrapper `theme_state_effective(state, global, ctx)` that short-circuits to `global` when the flag is off, regardless of slot contents. Persisted slots stay on disk through flag toggles; rendering only reflects them when the flag is on. The 862 migrated call sites all call the wrapper (not `state.effective()` directly), and the lint flags both the bare global accessor and any direct `state.effective()` inside a per-tab module. Flag-toggle round-trip test pins the non-lossy persistence.

specs/GH478/tech.md
```diff
@@ -0,0 +1,975 @@
...
+  from current settings.
+- Toggle the flag from **on** to **off** mid-session. Assert all
+  themed tabs redraw with the global theme (or the cwd-stripped
+  fallback chain — but with the flag off, `effective()` collapses to
+  global). Persisted slot data is not cleared.
+
+### Privacy invariant tests
+
+- `DirectoryThemeOverrides` settings group has `private == true` and
+  `sync_to_cloud == SyncToCloud::Locally`. Pinned by a settings-system
```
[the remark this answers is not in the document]
gulsahsarsilmaz: Fixed in v6 — the privacy invariant test now asserts `SyncToCloud::Never` (the v5 fix did not make it into the test text). Added a salt-file invariant test alongside: created with mode 0600 on first launch, never synced.

gulsahsarsilmaz:
@oss-maintainers — requesting an Oz cloud agent for implementation per CONTRIBUTING.md ("Approved requests run for free on complimentary Oz credits"). The spec is settled at v6 (`22a8b7d`); all review threads from Oz's four passes and @zachlloyd's first pass are resolved.

### Implementation surface (derived from `specs/GH478/tech.md`)

The PR opens with the feature flag `appearance.themes.per_tab_overrides` — default **off** in stable, **on** in dev/preview. Implementation work fans out across:

1. **Schema** — `Option<String>` `theme:` field on `TabTemplate` and `WindowTemplate` in `app/src/launch_configs/launch_config.rs:180–187`.
2. **Theme-name resolver** — `resolve_theme_ref(&str) -> Option<ThemeKind>` in `app/src/themes/theme.rs`, accepts display and snake_case forms.
3. **Per-tab state** — new `TabThemeState` (4 slots: `menu_pin`, `launch_config_pin`, `cwd_resolved`, `window_default`) on `TabSnapshot` (`app/src/app_state.rs:61–69`) and `TabData` (`app/src/tab.rs:134`); `effective(global)` resolver implementing the 5-layer priority order.
4. **Directory-pattern settings group** — new `app/src/settings/directory_overrides.rs` with `DirectoryThemeOverrides` (`SyncToCloud::Never`, `private: true`); `directory_theme_for(map, &Path)` matcher with longest-prefix component-boundary semantics, per-platform path normalization (Linux case-sensitive; macOS/Windows case-insensitive; Windows accepts `\` and `/`, normalizes drive letters).
5. **Three matching invocation paths** — tab creation, focused-pane cwd-change events (existing tracker), `DirectoryThemeOverrides` settings-change subscription.
6. **Theme catalog** — `Appearance` gains `theme_cache: RwLock<HashMap<ThemeKind, Arc<WarpTheme>>>` + `theme_for(Option<&ThemeKind>) -> Arc<WarpTheme>`, invalidated on `ThemeSettings` change and on custom-theme file changes.
7. **Renderer migration** — wrapper `theme_state_effective(state, global, ctx)` short-circuits to global when the flag is off; ~169 call sites under `app/src/terminal/**` migrate to call the wrapper, ~700 sites in chrome/onboarding/ui_components stay on global `Appearance::theme()`. Migration enforced by a new `dylint` lint `appearance_theme_in_tab_path` (with a ripgrep-in-presubmit fallback if the team prefers no new tooling).
8. **Save round-trip** — `From<WindowSnapshot> for WindowTemplate` updated with `preserved_override()` helper that walks `menu_pin → launch_config_pin → window_default`; window-level coalescing when shared, otherwise per-tab emission. `cwd_resolved` is never emitted.
9. **Right-click menu** — three new entries (Pin theme / Reset theme / Forget launch config theme) with the visibility rules in product behavior #18.
10. **Privacy / redaction** — `redacted_key_id(raw_key, &salt)` using HMAC-SHA256 truncated to 24 bits, salt from `~/.warp/redaction_salt` (mode 0600, generated on first launch, never synced); used in every diagnostic emitted by directory_overrides machinery.
11. **Tests** — unit + integration coverage per `tech.md` §Testing (priority-order, save round-trip, settings-change recompute, cross-platform path matching, redaction invariants, flag-off short-circuit, salt file invariants).

### CI requirements

`./script/presubmit` (cargo fmt, clippy with `-D warnings`, nextest, dylint) must pass; integration tests under `crates/integration/` pin the user-facing flows. Lint enabled with zero warnings on first push.

### PR-description fields the implementation will fill in

- `Closes #478` — auto-closes the parent issue on merge.
- `CHANGELOG-NEW-FEATURE: Per-tab theme overrides via launch configurations and directory matching`.
- Screenshots / short video of the three menu entries and a multi-themed-tab session for the *Screenshots / Videos* section.
- `Warp Agent Mode` checkbox checked.

Implementation can continue on this same branch (`gulsah/spec-per-tab-theme`) per `CONTRIBUTING.md`'s "implementation generally continues on the same PR" guidance, or move to a linked PR if Oz prefers to keep the spec diff isolated. I'll review whatever Oz pushes and iterate.

Thanks to @oz-for-oss for the four review passes and to @zachlloyd for the design feedback.

zachlloyd:
@gulsahsarsilmaz you are free to ask oz agent to implement this by tagging it.

after it generates the PR, please share a screencast demonstrating the behaviors

zachlloyd:
@oz-agent please implement the specs

zachlloyd:
@oz-agent please please implement the specs

gulsahsarsilmaz:
> @oz-agent please please implement the specs

Ooh! That's how to do it, sorry! Getting on it 😍

zachlloyd:
Nah there's a bug here. Oz isn't doing it for some reason. We are looking

captainsafia:
@gulsahsarsilmaz Thanks for working on this spec! Oz is having a bit of trouble here because the branch containing the specs exists on your fork (which Oz doesn't have permissions to push to).

To get things moving here, I'd recommend that we merge this spec-only PR to the repo then assign Oz to do the implementation in a fresh branch. That branch will exist in warpdotdev/warp which Oz has permission to push to.

gulsahsarsilmaz:
> @gulsahsarsilmaz Thanks for working on this spec! Oz is having a bit of trouble here because the branch containing the specs exists on your fork (which Oz doesn't have permissions to push to).
> 
> To get things moving here, I'd recommend that we merge this spec-only PR to the repo then assign Oz to do the implementation in a fresh branch. That branch will exist in warpdotdev/warp which Oz has permission to push to.

Sounds good. Thank you for getting back to me :) Is there anything I can do to assist?

captainsafia (approved):
Marking this spec as approved so we can merge and get Oz working on this!

gulsahsarsilmaz:
Thanks again! :)

Issue #7495, opened 2025-09-18
Support i(Pad)OS.


RokeJulianLockhart:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

I want Warp to provide a terminal client for iOS and iPadOS (the XNU-based Darwin derivatives).

### Is your feature request related to a problem? Please describe.

Current shells lack intelligent auto-completion and account synchronisation.

### Additional context

The most difficult part of implementing this would be locating a suitable basis, so that the Warp development team need not reinvent the wheel for the platform (instead, merely adding what makes Warp unique atop a stable base). To that end, https://github.com/holzschu/a-shell and https://github.com/blinksh/blink appear usable.

### Operating system (OS)

macOS

### How important is this feature to you?

1 (Not too important)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

dannyneira:
Thanks for this feature request! 

To anyone else interested in this feature, please add a 👍 to the original post at the top to signal that you want this feature, and subscribe if you'd like to be notified.

RokeJulianLockhart:
> Also note we're tracking iPhone support seperately here: https://github.com/warpdotdev/Warp/issues/6391

@dannyneira, glad to see such granularity, although need they be? They utilise the same OS, merely with different brand names.

RokeJulianLockhart:
@dannyneira, we're unable to see whether the codebase was forked, but since 90% of the platform APIs are standardised across them (and the rest can just be disabled with configuration files), I'd presume so. Unless you've reason to believe that any client would be architecturally distinct on each, *I* would track it in one.

dannyneira:
I prefer to keep it separate for now since they are different with regard to OS and form factor and would require different design and architecture considerations, etc. Again if anything changes on this front we'll consolidate.

Issue #6391, opened 2025-04-30
iPhone (iOS) support


selkordy:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

It would be lovely if Warp would work on my iPhone. And even better if history/settings synced with my Mac 😋

### Operating system (OS)

macOS

### How important is this feature to you?

1 (Not too important)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

dannyneira:
Thanks for this feature request! We have a related request for iPad support were tracking here https://github.com/warpdotdev/Warp/issues/2083


To anyone else interested in this feature, please add a 👍 to the [original post at the top](https://github.com/warpdotdev/Warp/issues/6391#issue-3030365041) to signal that you want this feature, and subscribe if you'd like to be notified. 

Please avoid spam comments like "+1"

kevinisyoung:
This would be fantastic. Please add support!

Xytronix:
Hi, can we get an iOS app with support for remote control of agent threads?

@captainsafia you may be able to offer limited support  only for remote control chats and later expand it into a terminal client?

The remote control service is such a great feature, but lacks clients on iOS and Android.

tinystorm:
Actually, it's very important now.

xiaochunyong:
I really need it. When can I offer some support

olegbal:
Please !

Issue #3328, opened 2023-07-07
Support Android (AOSP).


RokeJulianLockhart:
### Discord username

[`discord.com/users/638098902570237965`](https://discord.com/users/638098902570237965) 

### Describe the solution you'd like?

Like [`issues/3004`](https://github.com/warpdotdev/Warp/issues/3004#issue-1691077264), but for Android. We already have Termux, but it's barebones. Having Warp would be the difference between Google Keep and [`VSCode.dev`](https://VSCode.dev/).

I'm not requesting, at least initially, that Warp duplicate Termux's efforts in creating a `chroot` (although utilising it as a base would prevent reinvention of the wheel, so-to-speak).

### Is your feature request related to a problem? Please describe.

I don't currently like using terminal-based software much on Android. I genuinely would prefer it with this.

### Additional context

The WASM support mentioned at [`e37f458cf274e8452bbf7eee61fe805db610d23e/README.md#supported-platforms`](https://github.com/warpdotdev/Warp/blob/e37f458cf274e8452bbf7eee61fe805db610d23e/README.md#supported-platforms) might be able to indirectly provide this to an extent, but not to the extent that [`termux/termux-app/blob/3b5018b4c79a2c2d218eb6fd570b2f2bebb8415a/README.md#termux-app-and-plugins`](https://github.com/termux/termux-app/blob/3b5018b4c79a2c2d218eb6fd570b2f2bebb8415a/README.md#termux-app-and-plugins) does.

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:770f6576-d6c0-4e4f-a259-fc64b5156087

None

dannyneira:
Thanks for this feature request! 

Please add a :+1: to the original post at the top to signal that you want this feature, and then hit the subscribe button if you'd like to be notified.

crispybegs:
+ 1 for this

DEAD1nsane:
> Thanks for this feature request!
> 
> Please add a :+1: to the original post at the top to signal that you want this feature, and then hit the subscribe button if you'd like to be notified.

this is definitely important for me :) I have hit the 👍🏻

Vortecsmaster:
+1

d-throttle:
[Desktop mode now being GA in Android 16 QPR3](https://android-developers.googleblog.com/2026/03/android-devices-extend-seamlessly-to.html) is reason enough to justify this ticket!

Yavmarto:
+1

[...]

PR #9921, opened 2026-05-02
fix: prevent error toasts from overflowing the viewport

### LINKED ISSUE #9733: Error toasts can render outside of the viewport, without a way to scroll them.
### Pre-submit Checks

### Describe the bug

> When a session fails to share many times, the consequent error toasts can render outside of the viewport, without a way to scroll them.

### To reproduce

Reproduce what [`issues/9735`](https://github.com/warpdotdev/warp/issues/9735#issue-4364707732) describes, until the consequent error toasts render outside of the viewport.

### Expected behavior

They should be possible to scroll.

<p></p>Image<p></p>

### Operating system (OS)

Linux

`fedora-release-identity-kde-desktop-43`

### Shell Version

`$HOME/.dotnet/tools/.store/powershell/7.5.4`

### Current Warp version

`v0.2026.04.27.15.33.preview_02`

### Regression

### Recent working Warp date

I don't know whether it ever worked.

### Additional context

As requested:

1. ~~~pwsh
   #!/usr/bin/env pwsh
   loginctl show-session "$Env:XDG_SESSION_ID" -p Type | jc --ini | jq | yq -P .Type
   ~~~

1. > `wayland`

No

None

henrytran1803:
## Description

Error toasts (e.g. from repeated session share failures) could accumulate without bound, causing the toast stack to overflow the visible window with no way to scroll or dismiss older toasts.

The fix applies a three-layer defense:

1. **Cap persistent toast count** (`MAX_PERSISTENT_TOASTS = 5`): When the limit is exceeded, the oldest toast is evicted from the stack.
2. **Deduplication by `object_id`**: `show_persistent_toast()` now sets an `object_id` based on the toast text, so repeated identical error messages replace older toasts instead of stacking. The existing `dismiss_older_toasts` mechanism handles this automatically.
3. **Layout safety net** (`ConstrainedBox` + `Clipped`): The rendered toast stack is constrained to a `500px` max height with clipping, ensuring toasts never overflow the viewport even in edge cases.

### Changes

- **`app/src/view_components/dismissible_toast.rs`**:
  - Add `MAX_PERSISTENT_TOASTS` constant (5) and `TOAST_STACK_MAX_HEIGHT` constant (500px)
  - Cap persistent toast count in `add_persistent_toast()` by removing oldest toasts when limit exceeded
  - Wrap toast stack render in `ConstrainedBox` with `Clipped` to prevent viewport overflow
- **`app/src/terminal/view.rs`**:
  - Add `object_id` to persistent toasts in `show_persistent_toast()` for deduplication

## Linked Issue

- Fixes #9733

## Testing

The fix is defensive — it caps count, deduplicates, and constrains layout. Manual verification:
- Repeatedly trigger session share failures → older duplicate toasts are replaced (dedup)
- After 5+ unique persistent toasts → oldest evicted (cap)
- Toast stack never exceeds 500px height (layout constraint)

No automated test was added because the toast stack depends on UI rendering infrastructure that requires the full Warp app bootstrap. The logic changes (cap + dedup) are straightforward guards that are easy to verify visually.

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -14775,8 +14775,9 @@ impl TerminalView {
         ctx: &mut ViewContext<Self>,
     ) {
         let window_id = ctx.window_id();
+        let object_id = format!("persistent_toast:{}", &text);
         ToastStack::handle(ctx).update(ctx, |toast_stack, ctx| {
-            let toast = DismissibleToast::new(text, flavor);
+            let toast = DismissibleToast::new(text, flavor).with_object_id(object_id);
             toast_stack.add_persistent_toast(toast, window_id, ctx);
         });
     }
--- a/app/src/view_components/dismissible_toast.rs
+++ b/app/src/view_components/dismissible_toast.rs
@@ -11,7 +11,7 @@ use warpui::keymap::Keystroke;
 use warpui::r#async::Timer;
 use warpui::{
     elements::{
-        Border, ChildAnchor, ConstrainedBox, Container, CornerRadius, CrossAxisAlignment,
+        Border, ChildAnchor, Clipped, ConstrainedBox, Container, CornerRadius, CrossAxisAlignment,
         DispatchEventResult, EventHandler, Flex, Hoverable, Icon, MainAxisAlignment, MainAxisSize,
         MouseStateHandle, OffsetPositioning, ParentElement, PositionedElementAnchor,
         PositionedElementOffsetBounds, Radius, SavePosition, Shrinkable, Stack,
@@ -34,6 +34,8 @@ const VERTICAL_PADDING: f32 = 8.;
 const HORIZONTAL_PADDING: f32 = 12.;
 const ICON_RIGHT_MARGIN: f32 = 8.;
 const CLOSE_BUTTON_SIZE: f32 = 16.;
+const MAX_PERSISTENT_TOASTS: usize = 5;
+const TOAST_STACK_MAX_HEIGHT: f32 = 500.;
 
 const SUCCESS_ICON_PATH: &str = "bundled/svg/check-skinny.svg";
 const ERROR_ICON_PATH: &str = "bundled/svg/alert-circle.svg";
@@ -116,6 +118,10 @@ impl<A: Action + Clone> DismissibleToastStack<A> {
             uuid: Uuid::new_v4(),
         });
 
+        while self.toasts.len() > MAX_PERSISTENT_TOASTS {
+            self.toasts.remove(0);
+        }
+
         ctx.notify();
     }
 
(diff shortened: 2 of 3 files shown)


RokeJulianLockhart (commented):
> 1. **Cap persistent toast count** (`MAX_PERSISTENT_TOASTS = 5`): When the limit is exceeded, the oldest toast is evicted from the stack.

@henrytran1803, that might be problematic if more than 5 unique toasts are rendered. (Is having them be scrollable infeasible?)

> 2. **Deduplication by `object_id`**: `show_persistent_toast()` now sets an `object_id` based on the toast text, so repeated identical error messages replace older toasts instead of stacking. The existing `dismiss_older_toasts` mechanism handles this automatically.

Rendering a counter, for each time a duplicate toast is replaced, so that the user knows how many times that toast has been produced, would be helpful. [^1]

[^1]: [`invent.kde.org/plasma/spectacle/-/merge_requests/464#note_1244210`](https://invent.kde.org/plasma/spectacle/-/merge_requests/464#note_1244210)

> 3. **Layout safety net** (`ConstrainedBox` + `Clipped`): The rendered toast stack is constrained to a `500px` max height with clipping, ensuring toasts never overflow the viewport even in edge cases.

Can we not, instead, calculate the geometry of the interface, rather than define an arbitrary constant? This might not be sufficient on low-DPI (1 024 × 720 px) displays:

- [`issues/3328`](https://github.com/warpdotdev/warp/issues/3328#issue-1794527007)

- [`issues/7495`](https://github.com/warpdotdev/warp/issues/7495#issue-3429800827)

- [`issues/6391`](https://github.com/warpdotdev/warp/issues/6391#issue-3030365041)

henrytran1803:
@RokeJulianLockhart great points, all three — pushed a fix that addresses each of them!

**1. Scrollable instead of capped** — yep, agreed. I dropped the `MAX_PERSISTENT_TOASTS` cap entirely. Now the toast stack is just constrained to whatever vertical space the tab content actually has, so all unique toasts stay visible (and don't get silently evicted).

**2. Counter for duplicates** — nice idea! Instead of replacing the duplicate, I now increment a counter on the existing toast and display it as `(x3)` in the text. That way the user can see something's going wrong repeatedly without the toast stack growing.

**3. Viewport geometry instead of 500px** — good catch on low-DPI, that would've been embarrassing 😅 Replaced the hardcoded constant with a computed height from `TAB_CONTENT_POSITION_ID`'s actual rect (`height - 32px` for margins). Should scale correctly on any display size now.

Let me know if anything else needs tweaking!

RokeJulianLockhart (commented):
@henrytran1803, I've not time to locally test the changes (nor to produce enough output, from something like `strace`, to reproduce the original problem). However, those are very impressive changes, and the code looks good, from a cursory glance. <em>Many</em> thanks!

vkodithala (changes requested):
Thanks for the fix, @henrytran1803! Confirming with our designers that this is the behavior we want to see for excessive toasts. In the meantime, can you please attach a video demo showing how this works for the case that we have many duplicate toasts and excessive toasts that'd otherwise extend past the viewport?

I figure you can probably make a local, test-only code change that produces error toasts on some trigger to repro this behavior, in lieu of true errors.

vkodithala:
Hey @henrytran1803  - any update here? This is an awesome fix, would love to get it in. A video demo would be helpful!

PR #9926, opened 2026-05-02
Hide Dock icon when using dedicated hotkey window (GH-1154)

### LINKED ISSUE #1154: Hide Warp icon from the Dock when using Hotkey window

_No response_

```text
Like in xterm2, we can hide the icon from showing in the dock cause if we use the shortcuts to toggle up the terminal. There's no reason. 

It's always working in the background and in the settings there must be the option to show/hide the icon to maintain the cleanest dock we can have.
```

_No response_

### Additional context

_No response_

ShankyJS:
## Description
Implements issue #1154 per spec PR #9349.

Adds a macOS-only opt-in setting under Settings → Features → Global hotkey → Dedicated hotkey window: **"Hide Warp from the Dock"**. When the user has selected the dedicated hotkey window mode and configured a keybinding, enabling the switch flips Warp's `NSApplicationActivationPolicy` to `Accessory` so the Dock icon disappears while Warp keeps running and responds to the global hotkey.

Recovery is built into the effective-state rule, not the UI: switching away from Quake Mode, clearing the keybinding, or turning the setting off all restore the Dock icon immediately. Non-macOS platforms ignore the setting entirely (default no-op on the platform delegate trait).

### Files
- `app/src/settings/mod.rs` — `hide_dock_icon: bool` field on `QuakeModeSettings` (default `false`).
- `app/src/terminal/keys_settings.rs` — setters, `should_hide_dock_icon`, `apply_effective_dock_icon_visibility`, and a pure `compute_should_hide_dock_icon` helper that the unit tests target. Re-applied from the existing `set_global_hotkey_mode_…` and `set_quake_mode_keybinding_…` paths so mode/keybinding changes immediately update Dock visibility.
- `crates/warpui_core/src/platform/mod.rs` — `Delegate::set_dock_icon_visible(&self, bool)` with default no-op.
- `crates/warpui/src/platform/mac/delegate.rs` — macOS impl calls `[NSApp setActivationPolicy:]` on the main thread.
- `crates/warpui_core/src/core/app.rs` — `AppContext::set_dock_icon_visible`.
- `app/src/root_view.rs` — applies effective state at startup right after `maybe_register_global_window_shortcuts`.
- `app/src/settings_view/features_page.rs` — new `FeaturesPageAction::QuakeEditorToggleHideDockIcon`, telemetry mapping, action handler, mouse-state field, and `render_quake_mode_hide_dock_icon_row` rendered only inside the `QuakeMode` branch and gated with `#[cfg(target_os = "macos")]`.

## Testing
**Automated.** Six new unit tests in `app/src/terminal/keys_settings.rs` cover each invariant from `specs/GH1154/product.md`:
- happy path (macOS + Quake Mode + opt-in + keybinding → hides)
- default setting never hides (behaviors 2/3)
- ActivationHotkey / Disabled mode → visible (behavior 8)
- missing keybinding → visible (behavior 9)
- non-macOS → visible (behavior 13)

`cargo fmt`, `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`, `clang-format`, and `cargo test -p warp` are all clean locally. `wgslfmt` and the full-workspace `cargo nextest` run were not executed locally (the diff doesn't touch shaders or the broader workspace), so reviewers should rely on CI for those.

**Manual.** A full sweep of the product-spec invariants 1–17 has not been performed. Recommended manual validation before merging:
- enable Quake Mode + a keybinding, toggle the new switch, confirm the Dock icon disappears immediately
- press the hotkey while hidden — window still toggles
- clear the keybinding while the setting is on — Dock icon returns
- switch global hotkey to "Show/hide all windows" or Disabled — Dock icon returns
- restart with the setting on and a keybinding configured — Dock icon stays hidden from launch
- run a long-lived command, toggle the setting on/off, confirm the session/pane survives
- Short video for testing attached in here https://github.com/warpdotdev/warp/pull/9926#issuecomment-4513313891

## Server API dependencies
None.

## Agent Mode

CHANGELOG-NEW-FEATURE: macOS: added an option to hide Warp from the Dock while using the dedicated hotkey window (Settings → Features → Global hotkey).

Closes #1154
Spec: #9349


ShankyJS:
@cla-bot check

ShankyJS:
Hey @oz-agent thanks for the review, I've attached a video testing the feature. In the description (check for the manual testing done)

ShankyJS:
Tested locally with the OSS bundle (`warp-oss`) on macOS. Confirmed end-to-end behavior matches the spec:

- Toggling **Settings → Appearance → Show Warp in Dock** off immediately removes Warp from the Dock and from Cmd-Tab, and the Warp status bar icon appears in the menu bar.`
- Toggling it back on restores the Dock icon and Cmd-Tab entry and removes the status bar icon — no restart needed.
- The status bar menu (Show Warp / New Window / Settings / Quit Warp) all work as expected:
  - **Show Warp** with dedicated hotkey enabled focuses the hotkey window without toggling it closed.
  - **Show Warp** with no global hotkey and no visible windows opens a new normal window.
  - **Settings** opens the Appearance page in an existing or new normal window.
  - **Quit** goes through the standard confirmation path.
- The preference persists across restart and the hidden-Dock state is applied early on launch (read from `prefs_for_public_settings` before `app_builder.run`).
- Setting is hidden on non-macOS via `SupportedPlatforms::MAC`.

Per the spec (`product.md` L8/L18/L37, `tech.md` §6), the toggle is now a general Appearance preference next to the app icon controls rather than gated under quake mode, and the menu bar icon is the mandatory recovery surface whenever the Dock icon is hidden.

Here's a recording showcasing the changes. 

https://github.com/user-attachments/assets/302ab766-c7c2-4d36-b846-95e87dc4b97d

ShankyJS:
@kevinyang372 a review would be awesome, I think this PR covers the spec pretty good, I've tested the app locally and it works fine using the OSS build.

kevinyang372:
@ShankyJS Ugh sorry about that. I just added the ready-to-implement label to the issue. You should be able to re-request an oz review now

ShankyJS:
Addressed the latest review: the hide-Dock path now checks the result of `setActivationPolicy:NSApplicationActivationPolicyAccessory`. If the switch fails after the status item was created, we remove the status item, restore `NSApplicationActivationPolicyRegular`, and return NO so the setting isn't left persisted as off while Warp remains visible in the Dock.

`cargo check -p warpui` and `cargo test -p warpui --lib` both pass.

/oz-review

ShankyJS:
Hey @kevinyang372 OZ is finally happy with the code in this PR lol.

I posted another video recording on the feature, would love if you could take a look!

Thanks!

ShankyJS:
@alokedesai / @kevinyang372 just gently nudging, I've been running this oss version in my local for a couple of days and it looks good, I would love this feature upstream if it's looking good!

alokedesai:
Sorry for delay, will take a look today!

app/src/settings/app_icon.rs
```diff
@@ -132,4 +132,37 @@ define_settings_group!(AppIconSettings, settings: [
...
+        sync_to_cloud: SyncToCloud::Never,
+        private: false,
+        storage_key: "ShowDockIcon",
+        toml_path: "appearance.icon.show_dock_icon",
+        description: "Whether Warp is shown in the macOS Dock and Cmd-Tab switcher.",
+    },
 ]);
+
+#[cfg(test)]
+mod tests {
```
alokedesai: Not sure these tests add much value, mind just removing?

app/src/settings_view/appearance_page.rs
```diff
@@ -2838,12 +2848,36 @@ impl SettingsWidget for CustomAppIconWidget {
             &view.app_icon_dropdown,
         );
 
+        #[cfg(target_os = "macos")]
```
alokedesai: please gate this on `is_supported_on_current_platform` instead of cfg gating (here and below)

app/src/root_view.rs
```diff
@@ -262,6 +262,11 @@ pub fn init(app: &mut AppContext) {
 
     app.add_global_action("root_view:open_from_restored", open_from_restored);
     app.add_global_action("root_view:open_new", open_new);
+    app.add_global_action("root_view:show_primary_window", show_primary_window);
+    app.add_global_action(
+        "root_view:open_settings_from_status_item",
+        open_settings_from_status_item,
```
alokedesai: Mind adding more explanation for why we're adding mac menu items? It seems somewhat unrelated to the actual change unless i'm misunderstanding

specs/GH1154/product.md
```diff
@@ -0,0 +1,116 @@
+# Hide Warp Dock Icon with Menu Bar Fallback — Product Spec
+GitHub issue: https://github.com/warpdotdev/warp/issues/1154
+Figma: none provided
+
+## Summary
+Add a macOS-only setting that lets users hide Warp from the Dock and Cmd-Tab app switcher while Warp continues running. When the Dock icon is hidden, Warp must show a minimal menu bar icon in the system status area so users can bring Warp back, open a window, access settings, or quit even if they do not have a global hotkey configured.
```
alokedesai: Can you show a video / image of where this icon lives? I don't actually see it when playing around with this branch.

specs/GH1154/tech.md
```diff
@@ -0,0 +1,207 @@
...
+
+Extend `warpui::platform::mac::AppExt` with a builder method such as `set_show_dock_icon_on_launch(bool)` and call it from the macOS block next to `set_activate_on_launch`, `set_dev_icon`, `set_menu_bar_builder`, and `set_dock_menu_builder`.
+
+This lets the AppKit layer apply accessory mode as early as practical during launch, reducing visible Dock flicker for users who have already hidden the Dock icon.
+
+### 3. Add macOS platform support for activation policy and status item
+Extend `crates/warpui/src/platform/mac/app.rs` and the Objective-C bridge in `crates/warpui/src/platform/mac/objc/app.m` / `app.h`.
+
+Recommended shape:
+- Add `show_dock_icon_on_launch: bool` to the macOS `App` backend, defaulting to `true`.
```
alokedesai: I feel like this should just be a `show_dock_icon` field that can be set at launch or changed at runtime

specs/GH1154/tech.md
```diff
@@ -0,0 +1,207 @@
...
+
+1. If dedicated hotkey mode is enabled:
+   - If a quake window exists and is hidden, show it via `ctx.windows().show_window_and_focus_app`.
+   - If a quake window exists and is already open or pending open, focus it without hiding it.
+   - If no quake window exists, create it using the same creation path as `toggle_quake_mode_window`.
+2. If dedicated hotkey mode is not enabled:
+   - If a normal Warp window exists, activate the app and bring a normal window forward.
+   - If no normal window exists, call `open_new`.
+
+Do not reuse `toggle_quake_mode_window` directly for the status item Show Warp command, because toggling would hide the hotkey window when the user is trying to recover or foreground Warp.
```
alokedesai: not sure I'm following why we need a separate function for enabling quake mode because of the dock icon. They seem like two unrelated features to me, mind explaining in more detail?

alokedesai (commented):
Thanks for making this change! Did a high level pass on the specs and left a few questions. Can do a closer pass on the actual code afterwards 🍾

app/src/root_view.rs
ShankyJS: Yes, essentially this was requested by @kevinyang372 . The idea is that when we use the new feature to hide the icon from the app switcher, users may have a hard time getting back to the Warp window without using the hotkey. By adding an icon to the status bar, we can improve the user experience and make the app more easily accessible.

specs/GH1154/product.md
ShankyJS: It should be in the top bar in your MacOS it will only show when you run it using the OSS Build, I think this is because it needs the logo in order to get built

I think this video shows it here https://github.com/warpdotdev/warp/pull/9926#issuecomment-4513313891
 
Also including a screenshot in this comment 

image

app/src/settings_view/appearance_page.rs
ShankyJS: Updated in 1440c61e to gate the row with `AppIconSettings::show_dock_icon.is_supported_on_current_platform()` instead of compile-time `cfg` checks. The UI now follows the setting metadata (`SupportedPlatforms::MAC`) consistently.

app/src/settings/app_icon.rs
ShankyJS: Removed the test in 09fb457f. The setting definition remains unchanged.

app/src/root_view.rs
ShankyJS: Added a code comment in 6c88a99c explaining that when macOS Dock/Cmd-Tab visibility is disabled, the status item menu is the required recovery surface for foregrounding Warp or opening Settings.

specs/GH1154/tech.md
ShankyJS: Good callout — I clarified this in 9852bd56. Dock visibility is still independent from quake mode; quake only comes up here because the status item’s **Show Warp** action needs to foreground the dedicated hotkey window without going through the existing toggle path and accidentally hiding it.

specs/GH1154/tech.md
ShankyJS: Yep, agreed. I cleaned this up in 9852bd56 so the spec talks about a single `show_dock_icon` presentation state. The saved setting initializes it during launch, and runtime setting changes update that same state afterward.

ShankyJS:
Hey @alokedesai I think I've replied to all the comments, have a good one!

app/src/root_view.rs
alokedesai: gotcha, I have a slight preference with removing the status bar changes from this PR. In reality, the main use case for this is really via the hotkey window. Users can still get into Warp by swiping up and seeing all applications. Ghostty also doesn't have an icon in the top bar.

alokedesai (commented):
Have one comment about removing the status bar changes for now (I don't think the extra surface area of Warp is worth it for this specific feature).

Other than that, this should be good to go!

app/src/root_view.rs
ShankyJS: Ok, I'll save that code in another branch just in case we want to revive it at any point. Pushing the changeds

ShankyJS:
Hey @alokedesai, done! Removed the status bar / menu bar recovery changes; the dock-icon toggle now just switches the activation policy. 

Saved the status bar implementation on a branch in my fork (`gh1154-status-bar-recovery-saved`) in case we want to revisit it later. Updated the specs to match too. Ready for another look!

app/src/lib.rs
```diff
@@ -900,7 +900,12 @@ fn run_internal(mut launch_mode: LaunchMode) -> Result<()> {
     // the TOML-backed store. When disabled, they live in the platform-native
     // store (same backend as private). Use the correct one for pre-app reads.
     #[cfg_attr(
-        not(any(enable_crash_recovery, any(target_os = "linux", target_os = "freebsd"))),
+        not(any(
+            enable_crash_recovery,
+            target_os = "linux",
+            target_os = "freebsd",
+            target_os = "macos"
```
alokedesai, line 931: what warranted this change?

app/src/root_view.rs
```diff
@@ -982,8 +982,6 @@ fn open_settings_page_in_new_window(section: &SettingsSection, ctx: &mut AppCont
     });
 }
 
-/// MCP servers need to wait for initial load to complete, so we have this action in addition
```
alokedesai, line 986: Why remove the comments here? Mind reverting this?

app/src/root_view.rs
```diff
@@ -1424,11 +1427,6 @@ fn toggle_quake_mode_window(global_resource_handles: &GlobalResourceHandles, ctx
     };
 }
 
-/// This action will show or hide all of Warp's windows except the quake window
-///
-/// - If Warp is active and has any windows, hide those windows.
```
alokedesai, line 1430: Same here, let's keep the comment

app/src/root_view.rs
```diff
@@ -1330,59 +1328,64 @@ fn get_quake_mode_state(ctx: &mut AppContext) -> Option<QuakeModeState> {
     }
 }
 
-fn toggle_quake_mode_window(global_resource_handles: &GlobalResourceHandles, ctx: &mut AppContext) {
```
alokedesai, line 1334: can you summarize why we still need this change? it seems orthogonal to the actual change

alokedesai (commented):
Have a few small comments and then it should be good to go!

app/src/lib.rs
ShankyJS: We need this condition because prefs_for_public_settings is now used on macOS to read ShowDockIconState before app launch, so macOS must be excluded from the expect(unused) lint or the build fails with an "unfulfilled expectation" error

ShankyJS: Would you like me to make it one-liner instead? I feel like after 3 or more target_os it becomes harder to read in a single line.

app/src/root_view.rs
ShankyJS: Restored in https://github.com/warpdotdev/warp/pull/9926/commits/9c202e993e37011d90cbd7daf49ba07e44afac51

app/src/root_view.rs
ShankyJS: Restored in https://github.com/warpdotdev/warp/pull/9926/commits/9c202e993e37011d90cbd7daf49ba07e44afac51

app/src/root_view.rs
ShankyJS: Good point. This refactor was leftover from the previous status/menu-bar recovery path and is not needed anymore. Reverted in https://github.com/warpdotdev/warp/pull/9926/commits/836655c22e5101ec4b6f94a91feda5e2c89187ef to keep the PR focused on the dock icon preference.

ShankyJS:
Sorry for the extra iterations this took; applied the requested changes @alokedesai

app/src/lib.rs
alokedesai: no need, makes sense.

alokedesai (approved):

alokedesai:
@ShankyJS Looks like tests are failing btw

ShankyJS:
@alokedesai I think this was because my branch was heavily out-dated against master, I'm trying to run those test locally because It doesn't seem like Github action is running

ShankyJS:
A full `cargo test` fails on 13 tests locally both in my branch and on `master` so I think there might be something wrong  when running those locally. 

Everything else seems to be passing fine locally, ready for a `workflow` approval so we can check if everything looks good in GHA.

alokedesai:
Looks like clang format is still failing

ShankyJS:
Pushed the lint fix @alokedesai

alokedesai (approved):

ShankyJS:
Lol. I fix one two break, is there a reliable way to run all of these locally?

alokedesai:
`./script/presubmit` should do the job

ShankyJS:
Yeah after the last commit I should have ran `cargo fmt` fix upcoming

ShankyJS:
Ok there's at least one extra change we'll have to do, it seems like some things moved upstream during the development of this feature, the rebase brought to my attention that now `objc` has moved to `objc2`so I will need to patch that.

ShankyJS:
Ok one more try; lol. Hopefully this is the good one; I ran everything locally and the presubmit script. Hope this one is good! FYI check the last commit for the objc to objc2 refactor. https://github.com/warpdotdev/warp/pull/9926/commits/ef87ca760cc08357e1af18eac9fcb1543ea21e5c @alokedesai

alokedesai (approved):

ShankyJS:
Heck yeah I think we green, I don't have write access so can't merge, I'll leave you the honors @alokedesai

alokedesai:
Thanks for contributing!

PR #9928, opened 2026-05-02
feat(orchestrator): task→agent boundary tracker (PDX-40)

aloewright:
## Summary

Implements [PDX-40](https://linear.app/pdx-software/issue/PDX-40/a24-implement-boundary-enforcement-no-mid-task-switching) — the orchestrator invariant that an agent cannot be swapped mid-task; switches happen only at task boundaries.

- New `crates/orchestrator/src/boundary.rs` exposing `TaskBoundary`, `BoundaryGuard`, `BoundaryError`.
- `TaskBoundary::begin(task_id, agent_id)` returns an RAII guard whose `Drop` releases the binding. Re-binding the same `TaskId` (to any agent — same or different) fails with `BoundaryError::AlreadyBound` until the guard drops.
- Synchronous `std::sync::Mutex` is deliberate: O(1), no `.await`, and the only option inside `Drop`. Lock poisoning is recovered transparently.
- Re-exported from `orchestrator` crate root; integrates cleanly with the existing `Router` without modifying it (the user's parallel-pane rule).

## Test plan

- [x] `cargo check -p orchestrator --tests`
- [x] `cargo test -p orchestrator --test boundary` — 8 passing (happy path; same-agent and different-agent rebind errors; drop-and-rebind across boundary; multi-task isolation; shared-state cloning; 16-thread race asserting exactly one winner; `Send + Sync`).
- [x] No clippy warnings introduced on changed files.
- [x] No rustfmt drift on changed files.

Pre-existing failures in `crates/orchestrator/tests/mcp_forwarder.rs` (9 tests) and pre-existing clippy errors in `budget.rs` (`std::time::Instant` disallowed for wasm) exist on master and are out of scope.


aloewright:
Wrong base — reopening against aloewright/warp:master (this work is fork-internal, not for upstream warp).

PR #9932, opened 2026-05-02
feat: surface Claude tool calls as inline text annotations (Slice B)

danieljohnmorris:
## Summary

- Switches `claude -p` from `--output-format text` to `--output-format stream-json --verbose`, enabling structured event parsing
- Parses NDJSON stream events and renders tool activity as inline text annotations in the agent panel:
  - `[tool: Bash] echo hello` when Claude invokes a tool
  - `[result] hello` when the tool returns output
- Assistant text blocks are forwarded verbatim as before
- System, rate_limit, and result events are silently dropped
- Codex `--json` parsing is unchanged

## What's new vs the previous text-only path

Before: only the final assistant text reply appeared in the panel. Tool calls (file edits, shell commands, searches) were invisible.

After: each tool invocation and its result appear as a line in the panel as Claude works. Users see the agent's tool loop in real time.

## Scope

This is Slice B of the tool-loop bridge. The previous state was effectively Slice 0 (text-only).

Deferred (Slice C): Warp's native tool-block UI - the rich block rendering in cloud agent runs. That requires mapping each `tool_use` event to an `api::message::ToolCall` protobuf message, which is a larger protocol surface change.

## Event types handled

| Claude stream-json type | Panel output |
|---|---|
| `assistant` / `text` block | text forwarded verbatim |
| `assistant` / `tool_use` block | `[tool: Name] input-summary` |
| `user` / `tool_result` block | `[result] stdout-excerpt` |
| `system`, `result`, `rate_limit_event` | dropped |

Tool input summaries are tailored for common tools (Bash shows command, Read/Write/Edit show path, Grep shows pattern + path) and fall back to truncated JSON for everything else.

## Test plan

- Build with `./script/run` and set `WARP_BYPASS_AUTH=1`
- Open the agent panel (Cmd+I), ask a question that causes Claude to run a shell command, e.g. "what files are in the current directory?"
- Verify `[tool: Bash] ls ...` and `[result] file1 file2 ...` lines appear as Claude works
- Verify assistant text is still shown correctly after the tool loop completes
- Verify a text-only prompt (no tool use) still works correctly

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,22 @@
+# Local-AI bypass for the warpdotdev/warp fork.
+#
+# Loaded by `dotenvy::from_filename(".env")` at the very top of `run()` in app/src/lib.rs.
+# Copy to `.env` (gitignored) and uncomment the lines you want active.
+
+# Boot straight into a logged-in fake-user state. Skips the browser-check / paywall modal.
+# Implied (auto-on) whenever WARP_LOCAL_AI is set.
+# WARP_BYPASS_AUTH=1
+
+# Route agent runs to a local provider. One of: ollama | claude | codex.
+# - claude: shell out to the `claude` CLI (must be on PATH and authenticated).
+# - codex:  shell out to the `codex` CLI (must be on PATH and authenticated).
+# - ollama: NOT YET IMPLEMENTED (chat-only HTTP harness, planned for phase 2).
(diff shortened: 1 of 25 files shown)


danieljohnmorris:
Wrong target repo - this PR was created against the upstream by mistake.

PR #9934, opened 2026-05-02
feat(doppler): 401-triggered re-fetch logic (PDX-54)

aloewright:
## Summary

Implements [PDX-54](https://linear.app/pdx-software/issue/PDX-54/a46-401-triggered-re-fetch-logic) — when a downstream provider returns 401, invalidate the cached secret and refetch via Doppler before retrying.

Two layers, both in the new `crates/doppler/src/refetch.rs`:

1. **Primitive** — `DopplerClient::refetch(name, cwd)`: drops the cache entry, re-`get`s from the CLI. Use directly when the caller wants to drive its own retry.

2. **Higher-order helper** — `with_refetch_on_unauthorized(client, name, cwd, op, is_unauthorized)`: runs `op(secret)` once, and if it returns an error the caller classifies as unauthorized, invalidates and retries exactly once with the fresh secret. Single-retry budget bounds damage if the credential really is broken — no infinite loops.

`is_unauthorized` is a caller-supplied predicate (`Fn(&E) -> bool`) over the provider's error type — typically `|e| e.status() == 401` or `|e| matches!(e, MyError::Unauthorized)`. No coupling between Doppler and the caller's HTTP error type.

`RefetchError<E>` is generic over the provider's error type and distinguishes:
- `Doppler(DopplerError)` — initial fetch or refetch failed (Doppler-side problem).
- `Provider(E)` — non-unauthorized first error, or post-refetch retry also failed.

## Test plan

- [x] `cargo test -p doppler --test refetch` — 8 passing:
  - happy path — `op` called once, no refetch
  - unauthorized → refetch → success: 2 op calls, 2 doppler spawns, cache holds the fresh value
  - unauthorized twice — returns `Provider(Unauthorized)` from the second attempt; exactly one refetch
  - non-401 error short-circuits — no refetch
  - Doppler error during initial fetch — `op` never runs
  - Doppler error during refetch — `op` ran once, propagates as `Doppler(_)`
  - `refetch` primitive invalidates cache and returns the fresh value
  - helper path uses fresh secret (warm cache → 401 → refetch → success)


aloewright:
Wrong base — reopening against aloewright/warp:master (this work is fork-internal, not for upstream warp).

PR #9974, opened 2026-05-03
fix: resolve ls DIR/ clickable filenames against listed directory

### LINKED ISSUE #9908: Clickable filenames in `ls DIR/` output resolve against CWD, silently opening wrong files
### Pre-submit Checks

### Describe the bug

When a block's command lists the contents of a directory other than the block's working directory — for example `ls -la some/dir/`, `ls subdir`, `ls /etc/` — clickable filename detection in the block's output uses the **block's CWD** as the resolution root instead of the **directory that was listed**.

This produces three user-visible symptoms:

1. **Silent misresolution (most harmful).** If a file by the same name as one listed in `ls DIR/` also exists in the block's CWD, clicking the filename in the output opens the *wrong* file (the one from CWD) without any indication. The most obvious tell is `README.md`: clicking `README.md` in `ls subdir/` output opens the root `README.md` in whatever application is configured for Markdown files, not `subdir/README.md`.
2. **False negatives for files.** If `DIR/foo.txt` exists but `CWD/foo.txt` does not, `foo.txt` is not clickable at all in the `ls DIR/` output.
3. **False negatives for directories.** Subdirectories listed by `ls DIR/` are not clickable. (In contrast, directories listed by `ls` with no argument are clickable and open in Finder via `SystemGeneric` routing.)

Symptoms 2 and 3 are related to (but distinct from) the enhancement request in #4636. This report focuses on the silent-misresolution root cause that #4636 does not surface.

Related but distinct prior art: #4636 is an enhancement request for clickability of files outside CWD; this bug is about *silent misresolution* when a same-named file exists in CWD. #3439 is a feature request for right-click menus on `ls` output.

### To reproduce

**Symptom 1 — silent misresolution (the most important case):**

```bash
# Set up
mkdir -p /tmp/warp-repro/subdir
echo "root readme"      > /tmp/warp-repro/README.md
echo "subdir readme"    > /tmp/warp-repro/subdir/README.md
cd /tmp/warp-repro

# Run in Warp
ls -la subdir/
# Output includes:
#   -rw-r--r-- … README.md

# Click README.md in the ls output.
```

- **Actual:** Warp opens `/tmp/warp-repro/README.md` ("root readme") in the Markdown viewer.
- **Expected:** Warp opens `/tmp/warp-repro/subdir/README.md` ("subdir readme").

**Symptom 2 — false negative for unique files:**

```bash
cd /tmp/warp-repro
ls subdir/          # output includes 'README.md'
rm README.md        # now only subdir/README.md exists
ls subdir/          # clicking README.md in the new block: nothing happens
```

**Symptom 3 — directories in `ls DIR/` output not clickable:**

```bash
ls -la /tmp/warp-repro/        # subdir is clickable → opens Finder
ls -la /                       # /tmp is clickable → opens Finder
ls -la subdir/                 # any subdirectory row: NOT clickable
```

**Scrollback reproduction** (to show the bug is stable across block history):

```bash
ls -la subdir/     # block A (CWD = /tmp/warp-repro)
cd /tmp            # move away
# scroll back to block A, click README.md
```

- **Actual:** Same wrong `README.md` opens (from `/tmp/warp-repro`, not from `subdir/`).

This is consistent with how per-block resolution works today: each block stores its own pwd, so scrollback is internally correct — the bug is that the stored pwd is the wrong resolution root when the command listed a different directory.

### Expected behavior

When a block's command is a directory-listing command with a directory argument, clickable path detection in that block's output must use that directory (joined to the block's stored pwd) as the primary resolution root.

Specifically:

1. Given block command `ls DIR/` (or `ls -la DIR/`, `ls --color=always DIR/`, etc.) with `block.pwd = CWD`, bare filename `NAME` in the output must resolve to `CWD/DIR/NAME` if that path exists.
2. If both `CWD/NAME` and `CWD/DIR/NAME` exist, the block's listed-argument wins. (There is no scenario where a user running `ls DIR/` wants clicks on its output to open files outside `DIR`.)
3. If only `CWD/DIR/NAME` exists, `NAME` must be clickable.
4. Directory entries in `ls DIR/` output must be clickable and open in Finder, matching the existing behavior for `ls` with no argument.
5. The behavior must hold for scrollback blocks, including blocks restored from a previous session, because each block carries its own pwd and command.
6. The fix must not affect blocks whose output is already rooted (e.g. `find DIR/ -name '*.md'` output contains `DIR/sub/foo.md`, which already resolves correctly against `CWD`). Avoid double-joining.

#### Initial command coverage

The fix should start with `ls` (the most common case). A user-configurable list of listing-aware commands accommodates `exa`, `eza`, `lsd`, and similar modern replacements, and allows users to add their own. `tree` is a natural extension but is blocked by a separate tokenizer bug (box-drawing characters not recognized as link separators — tracked in a sibling issue).

Out of scope for this issue (explicitly defer to follow-ups):

- `ls -R DIR/` — recursive listings have multiple root contexts marked by `DIR/SUB:` headers; needs per-section resolution.
- `ls DIR1/ DIR2/` — multi-argument listings; needs per-section resolution.
- Piped/redirected output (`ls DIR/ > file; cat file`) — argument context is lost at the pipe boundary.

#### Root cause (from source reading)

Pointers to the code paths so maintainers can sanity-check the framing:

- `app/src/terminal/view/link_detection.rs`, `scan_for_file_path`: for BlockList clicks, the working directory passed to the resolver is `block.pwd().map(String::from)` — the CWD at the time the command ran. The block's command is not consulted.
- `app/src/util/file.rs`, `absolute_path_if_valid`: joins the candidate to the `working_directory` passed in, or treats it as absolute. No awareness of any directory argument on the block's command.
- `app/src/terminal/model/grid/grid_handler.rs:1097`, `possible_file_paths_at_point`: tokenizes the clicked row into candidate substrings. The tokenizer works correctly for bare filenames like `README.md`; the bug is strictly in the resolution step, not tokenization (for `tree`-style output, tokenization is also broken — see sibling issue).
- `app/src/terminal/model/block/serialized_block.rs`: blocks serialize `pwd` and `stylized_command`, so the fix works durably across scrollback and session restore.
- `app/src/terminal/view/open_in_warp.rs`, `check_openable_in_warp`: already uses `warp_completer::parsers` to extract positional arguments from commands in `FILE_VIEWER_COMMANDS`. The primary fix can reuse this parsing path for the new listing-command set.

#### Proposed fix — implementation summary

I've prototyped a fix on a local branch (`rndjams/fix-ls-dir-link-resolution`). Summary:

**What's in the proposed PR:**

- A new pure-logic helper in `crates/warp_util/src/listing_command.rs`:

  ```rust
  pub const DEFAULT_LISTING_COMMANDS: &[&str] = &["ls", "exa", "eza", "lsd"];

  pub fn listing_command_argument_dir(
      command: &str,
      pwd: &Path,
      listing_commands: &[&str],
  ) -> Option<PathBuf>;
  ```

  Uses `shlex::split` for shell tokenization (handles `'...'`, `"..."`, backslash escapes). Skips leading `KEY=VALUE` env-var prefixes (`LS_COLORS=auto ls DIR`), skips flags (`-*`), returns the first path-like positional joined to pwd, gated on `is_dir()`.

- 22 unit tests covering realistic `ls` variants: bare path, trailing slash, absolute path, tilde expansion, quoted args with spaces, multi-arg (first wins), flag+value combos like `--color=always DIR`, env-var prefixes, malformed input, custom command sets, negative tests for `find`/`cat`/`git`, plus 5 alias-specific tests exercising the resolved-name override path.

- Wire-up in `app/src/terminal/view/link_detection.rs`:
  - `scan_for_file_path` reads the block's command + pwd and computes `listing_dir_to_scan` via the helper.
  - The helper also accepts an alias-resolved command name via `Block::top_level_command(sessions)`, so user aliases like `alias ll='ls -l'` trigger the fix without needing `ll` in `DEFAULT_LISTING_COMMANDS`.
  - `compute_valid_paths` tries `listing_dir` resolution first via a new `try_resolve` closure; if that returns nothing, falls through to the existing `ShellNative(working_directory)` path. Behavior is strictly additive — non-listing commands and blocks without a directory argument are unaffected.

**Deliberate scope boundaries (V1):**

- **Aliases with baked-in positional args are a known limitation.** Single-level alias *names* are resolved via `Block::top_level_command(sessions)` so `alias ll='ls -l'` → `ll DIR/` correctly triggers the fix. But if an alias bakes in a positional (`alias lsd='ls /tmp'` → `lsd DIR/`), the helper picks the user-typed `DIR/` as the first positional, not the aliased `/tmp`. Rare in practice; documented in the helper's doc-comment.
- **No user-facing setting in V1.** `DEFAULT_LISTING_COMMANDS` is a `const` slice in `warp_util`. Making it a TOML setting is a separate, orthogonal concern and felt out of scope for a bug fix. Happy to add it in the same PR if maintainers prefer.
- **Recursive listings (`ls -R DIR/`, `tree DIR`) are out of scope.** Each section header (`SUB:`) changes the resolution root; needs per-section tracking.
- **Multi-arg listings (`ls DIR1/ DIR2/`) are out of scope.** First arg wins today.
- **Piped/redirected output (`ls DIR/ > file; cat file`) is a permanent limitation.** The argument context is lost at the pipe boundary.
- **`tree` output tokenization is blocked on a separate tokenizer bug** — box-drawing characters (`│`, `├`, `└`, `─`) aren't in `FILE_LINK_SEPARATORS`. Tracked in a sibling issue so `tree` benefits immediately once both fixes land.

**Validation (local):**

- `cargo test -p warp_util --lib listing_command` — all 22 tests pass (17 original + 5 alias-specific).
- `cargo test -p warp --lib util::` — 98 existing tests pass. 1 unrelated pre-existing flake in `util::git::tests::detached_tag_display_returns_short_sha` (environment-sensitive, unrelated to this change).
- `cargo clippy -p warp --all-targets -- -D warnings` — clean.
- `cargo fmt --all --check` — clean.

Happy to open the PR once this issue is triaged.

The repro steps above are minimal and deterministic; happy to record a short screen capture on request.

### Operating system (OS)

macOS

macOS 26.4.1 (Darwin 25.4.0, arm64)

### Shell Version

zsh 5.9

### Current Warp version

v0.2026.04.27.15.32.stable_03

### Regression

### Recent working Warp date

_No response_

### Additional context

Sibling issue: tokenizer does not treat box-drawing characters (`│`, `├`, `└`, `─`) as link separators, preventing `tree` and `eza --tree` leaf filenames from being tokenized as clickable candidates. That issue is complementary to this one — even after listing-aware resolution lands, `tree` output will still be unclickable until the tokenizer is fixed.

No

None

rndjams:
## Description

Fixes clickable filename resolution in `ls DIR/` output. Previously, bare filenames were resolved against the block's CWD, causing silent misresolution (opening the wrong same-named file) or false negatives (file not clickable at all). Now, when a block's command is a directory-listing command with a directory argument, bare filenames resolve against that directory first.

**Components:**
1. `crates/warp_util/src/listing_command.rs` — pure-logic helper that parses a command line to extract the listing directory argument.
2. `app/src/terminal/view/link_detection.rs` — wire-up: reads block command + alias-resolved name, tries listing-dir resolution before CWD fallback.
3. 22 unit tests covering flags, quoting, env-var prefixes, recursive/multi-dir rejection, aliases, and negative cases.

## Linked Issue

Closes #9908

## Testing

- `cargo clippy -p warp_util --lib -- -D warnings` — clean.
- `cargo fmt -- --check` — clean.
- Unit tests: 22 tests in `listing_command_test.rs` covering all edge cases from the product spec.
- Manual testing: `ls subdir/` → clicking `README.md` opens `subdir/README.md` (not CWD's copy).
- Full CI suite will validate the integration with `link_detection.rs`.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed clickable filenames in `ls DIR/` output resolving against CWD instead of the listed directory, which could silently open the wrong file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14878,6 +14878,8 @@ dependencies = [
  "regex",
  "serde",
  "serde_json",
+ "shlex",
+ "tempfile",
  "thiserror 2.0.17",
  "typed-path 0.10.0",
  "warpui",
--- a/app/src/terminal/view/link_detection.rs
+++ b/app/src/terminal/view/link_detection.rs
@@ -548,13 +548,23 @@ impl super::TerminalView {
         // resolve into DIR, not into CWD — fixing the silent-misresolution bug where
         // a same-named file in CWD would otherwise win.
         let try_resolve = |clean_path: &CleanPathResult| -> Option<PathBuf> {
+            // Only attempt listing-dir resolution for bare entry names (no path
+            // separators). Candidates like `subdir/README.md`, `./foo`, or absolute
+            // paths already resolve correctly against CWD and must not be double-joined
+            // with the listing directory.
             if let Some(listing_dir) = listing_dir {
-                if let Some(resolved) = absolute_path_if_valid(
-                    clean_path,
-                    ShellPathType::PlatformNative(listing_dir.to_path_buf()),
-                    shell_launch_data.as_ref(),
-                ) {
-                    return Some(resolved);
+                let path_str = &clean_path.path;
+                let is_bare_name = !path_str.contains('/')
+                    && !path_str.contains('\\')
+                    && !path_str.starts_with('~');
+                if is_bare_name {
+                    if let Some(resolved) = absolute_path_if_valid(
+                        clean_path,
+                        ShellPathType::PlatformNative(listing_dir.to_path_buf()),
+                        shell_launch_data.as_ref(),
+                    ) {
+                        return Some(resolved);
+                    }
                 }
             }
             absolute_path_if_valid(
--- a/app/src/terminal/view/link_detection.rs
+++ b/app/src/terminal/view/link_detection.rs
@@ -27,6 +27,9 @@ cfg_if::cfg_if! {
             util::openable_file_type::FileTarget,
         };
         use std::path::PathBuf;
+        use warp_util::listing_command::{
+            listing_command_argument_dir, DEFAULT_LISTING_COMMANDS,
+        };
         use warp_util::path::CleanPathResult;
         use warp_util::path::LineAndColumnArg;
     }
@@ -443,15 +446,49 @@ impl super::TerminalView {
     ) {
         // For AltScreen we scan for relative path with the current working directory.
         // For BlockList we scan for relative path with the pwd of the hovered block.
-        let pwd_to_scan_for = match position {
-            WithinModel::AltScreen(_) => self.pwd_if_local(ctx),
-            WithinModel::BlockList(inner) => self
-                .model
-                .lock()
-                .block_list()
-                .block_at(inner.block_index)
-                .filter(|block| !self.is_block_considered_remote(block.session_id(), None, ctx)) // Don't scan for file links if the block is on remote sessions
-                .and_then(|block| block.pwd().map(String::from)),
+        //
+        // For BlockList we also look at the block's command: if the block ran a
+        // directory-listing command (`ls DIR/`, `eza DIR/`, etc.), bare filenames in
+        // its output are rooted at DIR, not at the block's pwd. `listing_dir_to_scan`
+        // holds the resolved argument directory (`pwd.join(DIR)`) in that case, and is
+        // tried first during candidate resolution. If the block's command is not a
+        // listing command, or has no directory argument, this is `None` and the
+        // existing pwd-only resolution is used.
+        //
+        // `top_level_command` resolves shell aliases (e.g. `ll` → `ls`) via
+        // `Block::top_level_command`, which delegates to the active session's alias
+        // table. We pass the resolved name to `listing_command_argument_dir` so
+        // user-aliased listing commands trigger the fix without needing their alias
+        // name in the `DEFAULT_LISTING_COMMANDS` list.
+        let (pwd_to_scan_for, listing_dir_to_scan) = match position {
+            WithinModel::AltScreen(_) => (self.pwd_if_local(ctx), None),
+            WithinModel::BlockList(inner) => {
+                let sessions = self.sessions.as_ref(ctx);
+                let model_guard = self.model.lock();
+                let (pwd, command, top_level) = model_guard
+                    .block_list()
+                    .block_at(inner.block_index)
+                    .filter(|block| !self.is_block_considered_remote(block.session_id(), None, ctx))
+                    .map(|block| {
+                        (
+                            block.pwd().map(String::from),
+                            block.command_to_string(),
+                            block.top_level_command(sessions),
+                        )
+                    })
+                    .unwrap_or((None, String::new(), None));
+                drop(model_guard);
+
+                let listing_dir = pwd.as_deref().and_then(|pwd_str| {
+                    listing_command_argument_dir(
+                        &command,
+                        top_level.as_deref(),
+                        std::path::Path::new(pwd_str),
+                        DEFAULT_LISTING_COMMANDS,
+                    )
+                });
+                (pwd, listing_dir)
+            }
         };
 
         match pwd_to_scan_for {
@@ -473,6 +510,7 @@ impl super::TerminalView {
                     .spawn(move || {
                         let paths = Self::compute_valid_paths(
                             &path,
+                            listing_dir_to_scan.as_deref(),
                             possible_paths,
                             max_columns,
                             shell_launch_data,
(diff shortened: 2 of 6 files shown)


zachbai (commented):
@lucieleblanc mind taking a look at this one?

lucieleblanc (changes requested):
Hi @rndjams, thank you for the PR! There are many design decisions and technical limitations documented in comments throughout the code. These should all be collected and discussed up-front in a spec (see [CONTRIBUTING.md](https://github.com/warpdotdev/warp/blob/master/CONTRIBUTING.md#opening-a-spec-pr)) instead of needing to be discovered by reading through the implementation.

Some of the cases I'm a little concerned about from a product/user experience perspective:
- Excluding `ls DIR FILE` : if we always exclude invocations with multiple arguments, some entries that currently get resolved won’t get resolved at all anymore.
- Excluding ls -d: I’m not sure I understand why we need to exclude directories. The user can click on a resolved directory to open it in the OS’s native file navigator, for example.
- Do we handle aliases that include a positional argument, for example `lt='ls /tmp';` ? What about shell variable expansion?

On the technical side, my main concern is that this feature should build on our existing command parsing infrastructure, not introduce a parallel one. We already have a command parser in `warp_completer::parsers` that handles env-var prefixes, quoting, subshells, and compound commands. `Block::top_level_command` already calls into it. We should reuse and expand that. Plus, if we reuse the existing command completion parsers, we could also have access to the command completion specifications defined in https://github.com/warpdotdev/command-signatures, which could be useful for classifying arguments.

Let's discuss further through a spec.

PR #9304, opened 2026-04-28
Don't run SSH integration tests for PRs from forks.

vorporeal:
## Description
PRs from external forks have been failing CI because `google-github-actions/auth` cannot complete OIDC auth: our GCP Workload Identity Federation provider only trusts the base repository, so the OIDC token issued for a fork-PR run (which carries the fork's `repository` claim) is rejected. See [run #25071797593](https://github.com/warpdotdev/warp/actions/runs/25071797593/job/73455200680?pr=9280) for an example failure.

The auth is only needed by SSH integration tests, which use `gcloud compute start-iap-tunnel` to reach a GCP-hosted Ubuntu VM. As a short-term unblock, this skips the `auth` and `setup-gcloud` steps on fork-PR runs and filters the SSH-dependent tests out of the test runs (they all match `_ssh_` in their test names). All other tests continue to run as before, and SSH integration coverage is unchanged on `master`, `workflow_dispatch`, `workflow_call`, and same-repo PRs — those tests still validate post-merge.

A more complete fix (e.g. a `workflow_run`-triggered job that runs SSH tests in the base repo's context against the fork's SHA) is a possible follow-up if we decide fork PRs need full SSH coverage.

## Testing
Validated via YAML parse. Will verify behavior on this PR's CI run (same-repo PR — SSH tests should still run) and on a fork PR before relying on it.

## Server API dependencies
N/A — CI-only change.

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

PR #9503, opened 2026-04-29
Run executable shell scripts in the terminal instead of opening in the editor

amriksingh0786:
## Description

`is_file_openable_in_warp` returns `Some(Text)` for any non-binary file, which shadowed the existing executor branch in `open_file` when handling a `file://` URL — so executable shell scripts were opened in the editor instead of being run. The URI handler now classifies the action up front via a new pure helper `classify_open_file_action`, and routes executable shell scripts to the executor path that the rest of the agent flow already uses.

Detection rules for "runnable shell script":

- **Unix:** user-execute bit set AND (extension in `{sh, bash, zsh, fish, ksh}` OR file starts with a `#!` shebang).
- **Windows:** extension in `{ps1, bat, cmd}` (no x-bit concept).

Non-executable shell scripts still open in the editor — the routing is gated on the user-execute bit on Unix, so this only changes behavior for files the user marked executable.

Fixes #9005.

## Testing

Unit tests added at both layers:

- **`is_runnable_shell_script`** — 7 tests in `app/src/util/openable_file_type.rs`: executable `.sh`, non-executable `.sh`, alternate executable extensions (`.bash`, `.zsh`, `.fish`, `.ksh`), shebang-with-no-extension (with and without x-bit), plain-text rejection, symlink to executable.
- **`classify_open_file_action`** — 6 tests in `app/src/uri/uri_test.rs`: executable `.sh` → execute, non-executable `.sh` → editor, executable `.bash`/`.zsh`/`.fish` → execute, markdown → notebook, Rust source → editor, directory → session.

`./script/presubmit` is clean (`cargo fmt`, `cargo clippy --workspace --all-targets --tests -- -D warnings`, `clang-format`, `wgslfmt`, `cargo nextest`). The 5 `shell_integration_tests::*ssh*` failures observed locally are unrelated to this change — they require a Docker SSH container — and are skipped on fork PRs per #9304.

## Agent Mode

- [ ] Warp Agent Mode — This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: Executable shell scripts opened from a \`file://\` URL now run in the terminal instead of opening in the editor.

PR #9979, opened 2026-05-03
Skip and log unreadable pane trees during tab restore

### LINKED ISSUE #9109: Breaking regression: tabs cannot be restored after closing in v0.2026.04.15.08.45.stable_02
### Pre-submit Checks

### Describe the bug

Version v0.2026.04.15.08.45.stable_02 is a breaking release. It not only forces the UI into a vertical tab layout, but previously opened tabs can no longer be restored after they are closed and reopened.

The Save to New Config feature also appears limited to saving only a single tab.

Can you imagine how disastrous this is for users on managed corporate machines like ours, where all apps are automatically closed and logged out after 30 minutes of inactivity?

If this cannot be fixed, I’ll have no choice but to cancel my subscription.

Restoring tabs is a fundamental feature of any terminal tool, unless Warp no longer considers itself a terminal tool.

### To reproduce

Open some tabs.
Close Warp.
Open again.
All tabs lost!

### Expected behavior

_No response_

Restore tabs

### Operating system (OS)

macOS

26.4.1 

### Shell Version

_No response_

### Current Warp version

v0.2026.04.15.08.45.stable_02

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

None

amriksingh0786:
## Description

`read_sqlite_data` was using `filter_map(read_root_node(...).ok()?)` to assemble the saved tabs of a window, which silently dropped every tab whose pane tree failed to deserialize. After the `2026-04-14-150000_add_code_pane_tabs` and `2026-04-17-020439_add_custom_vertical_tabs_title_to_pane_leaves` migrations, a single bad leaf (typically an orphan `code_panes` row) would short-circuit the iterator and cause the user to lose **every** tab in that window — not just the broken one.

This PR replaces the silent drop with skip-and-log: a `log::warn!` keyed by the offending tab id, and the iterator continues with the remaining tabs.

```rust
let root = match read_root_node(conn, tab_id_for_log) {
    Ok(root) => root,
    Err(err) => {
        log::warn!(
            "Skipping tab id={tab_id_for_log} during restore: failed to read pane tree: {err}"
        );
        return None;
    }
};
```

**No schema change. No new migration.** Read-side fix only — safe to roll back via revert.

The fix prevents future tab loss but cannot recover sessions already overwritten by `save_app_state` on the broken stable_02 build; that data is gone before this code runs.

Fixes #9109.

## Testing

Added 3 behavioural tests in `app/src/persistence/sqlite_tests.rs`:

- `restore_skips_bad_pane_tree_keeps_others` — DB with [good, bad, good] tabs returns the two good tabs and emits the warn for the bad one.
- `restore_logs_error_on_bad_blob` — the warn-and-skip path runs without panicking.
- `restore_with_all_bad_blobs_returns_empty_not_panic` — graceful degradation when every tab is bad: empty list, no panic.

Plus three small fixture helpers (`terminal_tab_snapshot`, `code_tab_snapshot`, `window_with_tabs`) and `orphan_all_code_panes` to deliberately break pane trees the same way the April migrations did.

`./script/presubmit` is clean: `cargo fmt`, `cargo clippy --workspace --all-targets --tests -- -D warnings`, `clang-format`, `wgslfmt`, and the targeted `cargo nextest run -p warp 'persistence::sqlite'` all green. The 5 `shell_integration_tests::*ssh*` failures observed locally are unrelated to this change — they require a Docker SSH container — and are skipped on fork PRs per #9304.

Manually verified on macOS: dev build (`cargo run`) restored an existing 3-tab window cleanly, with no `Skipping tab id=` warnings (correct behaviour on a healthy DB — that branch only fires on bad blobs).

## Agent Mode

- [ ] Warp Agent Mode — This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: Tabs no longer disappear on restart when one pane fails to read from the persistence DB; the bad tab is skipped and logged instead.

### DIFF
--- a/app/src/persistence/sqlite.rs
+++ b/app/src/persistence/sqlite.rs
@@ -2681,7 +2681,20 @@ fn read_sqlite_data(
             let saved_tabs: Vec<_> = tabs_for_window
                 .into_iter()
                 .filter_map(|tab| {
-                    let root = read_root_node(conn, tab.id).ok()?;
+                    // Skip-and-log instead of silent-drop: see issue #9109. A single bad
+                    // pane tree (e.g. orphan code_panes row after the 2026-04-14/2026-04-17
+                    // migrations) used to drop every tab in the window. Now we drop only
+                    // the offending tab and surface the error so the rest survive.
+                    let tab_id_for_log = tab.id;
+                    let root = match read_root_node(conn, tab_id_for_log) {
+                        Ok(root) => root,
+                        Err(err) => {
+                            log::warn!(
+                                "Skipping tab id={tab_id_for_log} during restore: failed to read pane tree: {err}"
+                            );
+                            return None;
+                        }
+                    };
                     let panel = db_panels.get(&tab.id);
 
                     let left_panel = panel
--- a/app/src/persistence/sqlite_tests.rs
+++ b/app/src/persistence/sqlite_tests.rs
@@ -1,5 +1,6 @@
 use std::{path::PathBuf, sync::Arc};
 
+use diesel::{RunQueryDsl, SqliteConnection};
 use warp_core::features::FeatureFlag;
 use warp_graphql::scalars::time::ServerTimestamp;
 
@@ -11,7 +12,7 @@ use crate::{
     cloud_object::{CloudObjectPermissions, Owner},
     code::editor_management::CodeSource,
     notebooks::{CloudNotebook, CloudNotebookModel},
-    persistence::{model::ObjectPermissions, BlockCompleted, ModelEvent},
+    persistence::{model::ObjectPermissions, schema, BlockCompleted, ModelEvent},
     server::ids::ClientId,
     tab::SelectedTabColor,
     terminal::model::block::SerializedBlock,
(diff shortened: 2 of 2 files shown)


amriksingh0786:
@captainsafia @seemeroland, flagging since you both helped land #9503. Would value your call here.

@oz-for-oss you're right. I missed that `.ok()?` inside `filter_map` already returns `None` for just the failing item and continues iteration. The behavioral tests in this PR pass on the old code too, so they don't demonstrate the regression. Apologies for the misanalysis.

Looking at the wider failure surface, the actual all-tabs-loss path is at `app/src/persistence/sqlite.rs:164-174`: any `Err` from `read_sqlite_data` (including from the early `?` propagations on `windows.load`, `Tab::belonging_to(...).load`, or `panels.load` at lines 2659-2671) bubbles up to the caller, which sets `app_state = None` and wipes every window. The April migrations don't directly touch `windows`/`tabs`/`panels`, but a partial-migration or downgrade scenario could leave Diesel-expected columns missing on those tables. The original reporter is on stable_02 specifically; a maintainer on the issue thread also couldn't reproduce, which is consistent with a DB-state-dependent failure.

Without a reproducer DB I'd be guessing at the exact failing query. Two options:

1. Close this PR as misdiagnosed; the issue stays open for someone with a repro.
2. Narrow this PR to keep the warn-and-skip as a defensive observability improvement only. Iterator behavior is unchanged, but a future failure leaves a log line keyed by the bad tab id instead of a silent drop. Description, commit message, and changelog get rewritten to be honest about scope: this does not fix #9109; it just makes the next report easier to triage. Issue stays open.

Leaning toward (1) since (2) feels like salvaging a misframed PR. Happy to do either. What would you prefer?

seemeroland:
Yea I agree with 1! Thanks for the attempt

amriksingh0786:
Closing per @seemeroland's preference on the option-1 path. Issue #9109 stays open for someone with a reproducer DB. Thanks @oz-for-oss for catching the misanalysis.

PR #9999, opened 2026-05-03
fix: pane separator moves at half mouse speed during drag (#6468)

### LINKED ISSUE #6468: Pane separator resizing horizontal divider moves very slowly
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/help/sending-us-feedback#gathering-warp-logs)

### Describe the bug

Before, when I was using multiple panes in the same window, I used to be able to resize the panes just fine. Since a few versions ago, the separator moves extremely slow, compared to the mouse. If I move the mouse 200 px, the separator will move perhaps half of that, or less

### To reproduce

1. Open a Warp window
2. Create a second pane by dividing (on mac, command D)
3. click the divider between the two panes, and move horizontally to resize them
4. the movement of the divider is much shorter than the mouse movement

### Expected behavior

The pane divider should move at the same speed as the mouse pointer

[warp-logs.zip](https://github.com/user-attachments/files/20231822/warp-logs.zip)

The following is actually an .mp4 video but Github won't let me upload with the .mp4 extension. No idea why. Just change the name to divider.resize.mp4

[divider resize.zip](https://github.com/user-attachments/files/20231831/divider.resize.zip)

### Operating system (OS)

macOS

14.7 (23H124)

### Shell Version

zsh 5.9 (x86_64-apple-darwin23.0)

### Current Warp version

v0.2025.05.07.08.12.stable_02

### Regression

### Recent working Warp date

Probably three months ago

### Additional context

_No response_

No

None

yingxuel:
Closes #6468

## Description

When dragging the pane separator, the divider moved at roughly half the speed of the mouse cursor — a 200 px mouse movement produced only ~100 px of divider movement.

**Root cause:** `PaneBranch::adjust_pane_size` called `pane_size()` (which returns last-rendered element bounds) on every drag event to compute the flex redistribution. Mouse drag events can fire at ~120 Hz while the app renders at ~60 Hz, meaning two events fire per render frame. Both events read the same stale rendered sizes and compute the same flex value, so the second event overwrites the first with an identical result — only one event's worth of movement is applied per frame, halving the effective speed.

**Fix:** Add `cached_total_pixel_size: Option<f32>` to `DraggedBorder`. On the first drag event after a render, the total pixel size of the two adjacent panes is read from the renderer (when fresh) and cached. All subsequent events in the same drag session derive individual pane sizes from the current flex ratio instead of re-reading stale rendered bounds, so each delta accumulates correctly.

Extract the redistribution math into `compute_new_flex`, a pure function with no `ViewContext` dependency, enabling unit testing.

## Linked Issue

## Screenshots / Videos

https://github.com/user-attachments/assets/7c1a65c1-a32f-439c-94c4-6d42899548b3

## Testing

Added 6 unit tests in `app/src/pane_group/tree_tests.rs` covering `compute_new_flex`:

- `test_compute_new_flex_single_event` — basic single drag event correctness
- `test_compute_new_flex_accumulates_across_events` — 3 × 10 px events yield 430 px, not 410 px
- `test_stale_size_bug_would_undercount_deltas` — explicitly compares old formula (410 px) vs new (430 px) for N events before a re-render, proving the bug existed and is fixed
- `test_compute_new_flex_minimum_pane_size_respected` — clamping at minimum pane size
- `test_compute_new_flex_ignores_near_zero_delta` — sub-epsilon delta returns `None`
- `test_compute_new_flex_asymmetric_flex` — asymmetric flex values (600 px / 200 px split)

**Pre-existing presubmit failures (unrelated to this PR):** 6 tests fail on `master` before applying these changes, confirmed by the PR author:
- `shell_integration_tests::test_legacy_ssh_into_bash`
- `shell_integration_tests::test_legacy_ssh_into_zsh`
- `shell_integration_tests::test_ssh_into_ash`
- `shell_integration_tests::test_ssh_into_sh`
- `ui_tests::test_ssh_with_shell_override`
- `settings::init::tests::test_migration_does_not_rerun_when_marker_present`

These require SSH infrastructure or specific local state not available in a standard dev environment and are not caused by this PR.

## Agent Mode

### DIFF
--- a/app/src/pane_group/tree.rs
+++ b/app/src/pane_group/tree.rs
@@ -503,9 +503,11 @@ impl PaneData {
         &mut self,
         border_id: EntityId,
         delta: f32,
+        cached_total_pixel_size: &mut Option<f32>,
         ctx: &mut ViewContext<PaneGroup>,
     ) {
-        self.root.adjust_pane_size(border_id, delta, ctx);
+        self.root
+            .adjust_pane_size(border_id, delta, cached_total_pixel_size, ctx);
     }
 
     pub fn adjust_pane_size_by_id(
@@ -728,11 +730,14 @@ impl PaneNode {
         &mut self,
         border_id: EntityId,
         delta: f32,
+        cached_total_pixel_size: &mut Option<f32>,
         ctx: &mut ViewContext<PaneGroup>,
     ) -> bool {
         match self {
             PaneNode::Leaf(_) => false,
-            PaneNode::Branch(branch) => branch.adjust_pane_size(border_id, delta, ctx),
+            PaneNode::Branch(branch) => {
+                branch.adjust_pane_size(border_id, delta, cached_total_pixel_size, ctx)
+            }
         }
     }
 
@@ -1107,6 +1112,7 @@ impl PaneBranch {
         &mut self,
         border_id: EntityId,
         delta: f32,
+        cached_total_pixel_size: &mut Option<f32>,
         ctx: &mut ViewContext<PaneGroup>,
     ) -> bool {
         if let Some(idx) = self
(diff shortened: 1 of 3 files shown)


yingxuel:
@cla-bot check

app/src/pane_group/tree.rs
```diff
@@ -1107,48 +1112,43 @@ impl PaneBranch {
...
+                let pane_size_1 = self.nodes[idx].1.pane_size(ctx);
+                let pane_size_2 = self.nodes[idx + 1].1.pane_size(ctx);
+                match self.axis {
+                    SplitDirection::Horizontal => pane_size_1.x() + pane_size_2.x(),
+                    SplitDirection::Vertical => pane_size_1.y() + pane_size_2.y(),
+                }
+            });
 
-            // Omit noise in dragging.
             let minimum_pane_size = get_minimum_pane_size(ctx);
```
vkodithala, line 1139: nit: `minimum_pane_size` is used only in `compute_new_flex`, can we derive it within `compute_new_flex` and remove this parameter/variable declaration?

app/src/pane_group/tree_tests.rs
```diff
@@ -1,5 +1,58 @@
...
+) -> f32 {
+    let mut flex_1 = initial_flex_1;
+    for _ in 0..n {
+        // Old formula: always divides by stale total, overwrites with the same
+        // value on every iteration before a re-render.
+        flex_1 = ((stale_size_1 + delta) / (stale_size_1 + stale_size_2) * total_flex)
+            .clamp(0., total_flex);
+    }
+    flex_1
+}
```
vkodithala, line 25: Your tests generally look good, but I'd remove things that directly check old/stale functionality (basically every caller of `old_buggy_flex_after_n_events`). Tests should check that _new_ functionality works as expected, not prove the bugginess of an earlier approach.

vkodithala (changes requested):
Thanks for the contribution @yingxuel! These changes mostly look good, left a few comments on extraneous parameters and test hygiene. 

Awesome that you included a demo, and I'm pretty sure that your solution is correct, but I'd like to see something that compares drag speeds from the old implementation vs. your approach to be confident that this is a real improvement. Will take another look once these comments are resolved and you've included an updated demo!

PR #9667, opened 2026-04-30
feat: recognize Mistral Vibe as a CLI agent (#9607)

lonexreb:
## Description

Resolves #9607. Adds `CLIAgent::Vibe` so running `vibe` (or `vibe-acp`) in a Warp pane is detected as a known CLI agent on par with Claude Code, Codex, Gemini CLI, OpenCode, etc. Mistral Vibe is the open-source ACP-compatible CLI shipped alongside Devstral 2.

The change is purely additive and follows the existing `CLIAgent` pattern in `app/src/terminal/cli_agent.rs`:

| Property | Value |
|---|---|
| `command_prefix` | `"vibe"` |
| `display_name` | `"Mistral Vibe"` |
| `brand_color` | Mistral orange `#FA520F` (new `MISTRAL_ORANGE` constant) |
| `brand_icon_color` | white (default arm) |
| `supported_skill_providers` | `&[SkillProvider::Agents]` |
| `icon` | `None` — see below |

Telemetry: `CLIAgentType::Vibe` added in lockstep so `From<CLIAgent> for CLIAgentType` stays exhaustive.

### Why `icon()` returns `None`

Shipping `Icon::MistralLogo` requires an officially licensed SVG that I don't have rights to bundle. The brand-color tile still renders, and the agent is fully detected. A follow-up PR can wire the `Icon::MistralLogo` variant + asset once the Warp team / Mistral provide one — the issue author explicitly suggested this draft-PR approach.

## Linked Issue

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. (#9607 is `ready-to-implement`.)

Resolves #9607.

## Testing

Extended the existing `test_detect_known_agents` table-driven test to cover `vibe` -> `CLIAgent::Vibe`. The pre-existing `test_serialized_name_round_trips_known_agents` exercises every variant via `enum_iterator::all`, so it now also covers Vibe's serde round-trip without code changes.

Local builds blocked on missing Xcode `metal` compiler in this environment; CI's MacOS / Linux / Windows lanes will validate `cargo build`, `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings`, and `cargo nextest run`.

## Agent Mode

- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-IMPROVEMENT: Recognize Mistral Vibe (`vibe` / `vibe-acp`) as a known CLI agent in the terminal, with brand-colored toolbar tile. Thanks @lonexreb!

PR #10022, opened 2026-05-04
fix: detect Node.js / Python shebang scripts as their agent identity (#9870)

### LINKED ISSUE #9870: Codex agent toolbelt does not appear on Linux because Node.js script process shows as 'node' not 'codex'
## Summary

On Linux, the Codex CLI agent toolbelt never appears when running `codex` inside a Warp terminal. The root cause is that `codex` is a Node.js shebang script (`#!/usr/bin/env node`), so the OS process name (`/proc/PID/comm`) is always `node`, not `codex`. Warp's agent detection matches against `comm`/`argv[0]`, which resolves to `node` for any Node.js script — so the toolbelt is never triggered.

## Problem

Warp detects CLI agents by observing the foreground process name in the terminal session. On Linux, `/proc/PID/comm` is initialised from the basename of the **executable** (i.e. `node`), not from `argv[0]` or the symlink name used to invoke the script. Because `codex` is installed as:

```
/usr/local/bin/codex -> ../lib/node_modules/@openai/codex/bin/codex.js
```

...and `codex.js` has the shebang `#!/usr/bin/env node`, the foreground process that Warp observes is:

- `/proc/PID/comm` = `node`
- `argv[0]` = `node`
- `argv[1]` = `/usr/local/bin/codex`

Warp never sees a process named `codex`, so the toolbelt does not activate. The issue does not reproduce on macOS where process naming for shebang scripts behaves differently.

Additionally, `codex.js` does not call `process.title = 'codex'` anywhere in its source, which is the Node.js mechanism that calls `prctl(PR_SET_NAME, ...)` on Linux and would update `/proc/PID/comm` to `codex` at runtime.

## Expected behavior

The Codex agent toolbelt should appear at the bottom of the Warp pane when `codex` is launched inside a Warp terminal on Linux, matching the behaviour on macOS.

## Actual behavior

The toolbelt never appears. All other settings are correctly configured:
- `agents.third_party.should_render_cli_agent_toolbar = true`
- `agents.third_party.cli_agent_toolbar_enabled_commands.codex = "Codex"`
- `TERM_PROGRAM=WarpTerminal`, `WARP_CLI_AGENT_PROTOCOL_VERSION=1` are set in the shell environment
- Warp shell integration marker (`SourcedRcFileForWarp`) is present in `~/.bashrc`

## Reproduction steps

1. Install Warp Preview on Ubuntu Linux (apt, x86_64).
2. Install Codex CLI via `npm install -g @openai/codex`.
3. Open a new Warp terminal window.
4. Run `codex` in the foreground.
5. Observe that no agent toolbelt appears at the bottom of the pane.

To confirm the process name issue independently:
```bash
# Create a shebang script and symlink, identical to how codex is installed
TMPSCRIPT=$(mktemp /tmp/codex_sim_XXXXXX.js)
printf '#!/usr/bin/env node\nsetTimeout(()=>{}, 5000);\n' > "$TMPSCRIPT"
chmod +x "$TMPSCRIPT"
TMPSYM=$(mktemp -u /tmp/codex_XXXXXX)
ln -s "$TMPSCRIPT" "$TMPSYM"
"$TMPSYM" &
SIM_PID=$!
sleep 0.3
cat /proc/$SIM_PID/comm   # prints: node
tr '\0' ' ' < /proc/$SIM_PID/cmdline  # prints: node /tmp/codex_...
kill $SIM_PID
```

## Artifacts

None attached.

## Warp version

`0.2026.04.27.15.33.preview.02` (Warp Preview, installed via apt from `releases.warp.dev/linux/deb preview/main`)

## Operating system

Ubuntu Linux, kernel `7.0.0-14-generic`, x86_64. Wayland (GNOME Shell). Node.js shebang scripts present this process-naming behaviour on all Linux distributions.

## Possible source references

- Agent detection logic likely reads `/proc/PID/comm` or `argv[0]` to identify the foreground process. Extending the match to also check `argv[1]` (the script path) when `argv[0]` is `node` (or another JS/Python runtime) would cover this case.
- The `cli_agent_toolbar_enabled_commands` setting in `settings.toml` maps `codex = "Codex"`, suggesting the detection key is the literal string `codex` — which will never match `/proc/PID/comm` for a Node.js script on Linux.
- Workaround: adding `process.title = 'codex'` to `/usr/local/lib/node_modules/@openai/codex/bin/codex.js` causes Node.js to call `prctl(PR_SET_NAME, 'codex')` on Linux, updating `/proc/PID/comm` to `codex` and unblocking detection. The correct fix is either in Warp's detection logic or upstream in the Codex CLI package.

lonexreb:
Fixes [#9870](https://github.com/warpdotdev/warp/issues/9870).

## The bug

On Linux, agents distributed as Node.js or Python shebang scripts surface as `node /usr/local/bin/<agent>` because `/proc/PID/comm` is initialised from the executable basename, not from `argv[0]` or the symlink name. The bare prefix match in `CLIAgent::detect` saw `node` and missed the agent entirely, so the third-party CLI agent toolbar never appeared.

The canonical reproduction in #9870: the `@openai/codex` npm package installs `/usr/local/bin/codex` as a symlink to `codex.js` whose shebang is `#!/usr/bin/env node`. Detection failed for every Linux user of Codex.

## The fix

Two new private helpers in `app/src/terminal/cli_agent.rs`, plus a small extension to `detect()` to consider their output:

- **`path_basename_token`** peels a leading path component so `/usr/local/bin/codex` → `codex` for matching purposes. Covers fully-qualified invocations.
- **`shebang_script_basename`** recognises a known script-runtime invocation (`node`, `nodejs`, `bun`, `deno`, `python`, `python3`) and pulls the script basename from `argv[1]`, stripping common script extensions (`.js`, `.mjs`, `.cjs`, `.ts`, `.py`). Skips leading flags (`node --inspect /usr/local/bin/codex`).
- `detect()` matches `command_prefix()` against the exact first word OR either of the two derived tokens.

The change is **purely additive** — every existing detect path still matches exactly as before, including the Uber-team `aifx agent run claude` special case. The new candidates only fire when the exact-word match fails, so there's no false-positive surface for unrelated commands.

## Test coverage

`cli_agent_tests.rs` gets three new test functions covering:

- `node /usr/local/bin/codex` → `CLIAgent::Codex` (the canonical #9870 case)
- `node --inspect /usr/local/bin/codex` (runtime flag before script path)
- `node /path/to/codex.js` (extension stripped)
- `nodejs`, `bun` alternative runtimes
- Path-prefixed binaries: `/usr/local/bin/codex`, `./goose`, `/opt/homebrew/bin/claude`
- **Negative tests** ensure no false positives: `node`, `node ./my-app.js`, `python /home/user/script.py`, `node --version` all return `None`.

## Why not patch upstream `process.title`?

The reporter notes that adding `process.title = 'codex'` to the Codex CLI's entry script would update `/proc/PID/comm` to `codex` on Linux. That's a fix in the wrong tree — every shebang-script-distributed agent would need the same upstream patch, including agents Warp doesn't ship and can't influence. Recognising the pattern in Warp's detection layer covers all of them at once.

## Test plan

- [ ] On Ubuntu Linux with `npm install -g @openai/codex`, run `codex` in a Warp terminal — third-party agent toolbar appears (was: never appeared).
- [ ] On macOS with the same Codex install, behavior is unchanged (the fix's new code paths are exercised but produce the same answer as today's exact-word match).
- [ ] On Linux, run `node ./my-random-app.js` — no agent toolbar appears (negative case).
- [ ] All existing `cli_agent_tests.rs` tests continue to pass.

## Scope

- Two files: `app/src/terminal/cli_agent.rs` (+77/-2), `app/src/terminal/cli_agent_tests.rs` (+103).
- No public API changes.
- No new dependencies.
- Builds on top of master; no conflict with the in-flight Mistral Vibe PR (#9667) — both modify `detect()` but in non-overlapping arms (Vibe adds a `vibe-acp` exact-word check, this adds basename / runtime fallbacks). Either can rebase cleanly onto the other.


lonexreb:
The two important issues from the previous bot review are now addressed in commit 409b61b:

- **Generic-prefix false positives:** introduced `GENERIC_AGENT_PREFIXES` and `prefix_eligible_for_basename_recovery()` so basename / shebang-runtime recovery only applies to unambiguous agent names. The generic `agent` prefix (CLIAgent::CursorCli) is now gated out of fallback matching.
- **Env-var prefix in shebang scripts:** `shebang_script_basename` now uses `is_env_var_assignment()` to skip `KEY=value` tokens, aligning with `extract_first_command`'s shell-aware parsing.

The bot's re-review at 09:30 UTC on May 4 completed without requesting further human review, but the PR still shows `CHANGES_REQUESTED` from the original review. Could a maintainer dismiss or re-approve so this can move forward? Happy to address anything else flagged.

harryalbert:
> The two important issues from the previous bot review are now addressed in commit [409b61b](https://github.com/warpdotdev/warp/commit/409b61bd334074cb7c233b57f4b00f9fd605aa3f):
> 
> * **Generic-prefix false positives:** introduced `GENERIC_AGENT_PREFIXES` and `prefix_eligible_for_basename_recovery()` so basename / shebang-runtime recovery only applies to unambiguous agent names. The generic `agent` prefix (CLIAgent::CursorCli) is now gated out of fallback matching.
> * **Env-var prefix in shebang scripts:** `shebang_script_basename` now uses `is_env_var_assignment()` to skip `KEY=value` tokens, aligning with `extract_first_command`'s shell-aware parsing.
> 
> The bot's re-review at 09:30 UTC on May 4 completed without requesting further human review, but the PR still shows `CHANGES_REQUESTED` from the original review. Could a maintainer dismiss or re-approve so this can move forward? Happy to address anything else flagged.

You're able to re-request review from the agent by using the /oz-review command. Once you feel this PR is ready for review by a maintainer, please feel free to @/tag me! I'd be happy to take a look

harryalbert (commented):
@lonexreb, after looking over this PR a bit more, I think I'm a little hesitant about the solution here. While it would potentially address the issue, it's depending on a lot of special node, python, etc. state that could change with new versions or older versions (and is not generalizable to other runtimes that folks might be using). I'd prefer a less specialized solution if possible.

I chatted with the agent a bit, and I think that a more robust solution shape might look something like:

Persist the shell preexec command
◦ Add preexec_command: Option to Block.
◦ In Block::preexec at app/src/terminal/model/block.rs:3247, store data.command before/while handling PreexecValue.
◦ Add a narrow getter, e.g. preexec_command(&self) -> Option<&str>.
Use preexec command for CLI-agent detection
◦ In detect_cli_agent_from_model at app/src/terminal/view/use_agent_footer/mod.rs:352, prefer:
▪ active_block.preexec_command()
▪ fallback to active_block.command_with_secrets_obfuscated(false) at app/src/terminal/view/use_agent_footer/mod.rs:373
◦ This makes detection use what the shell was about to execute, not whatever command text/process shape Warp later reconstructed.
Simplify CLIAgent::detect
◦ Keep:
exact first-token match: codex
alias expansion
path basename match: /usr/local/bin/codex
existing special cases like aifx agent run claude / vibe-acp
◦ Remove the broad runtime parser for Node/Python/Deno/Bun flags from app/src/terminal/cli_agent.rs.
Curious on your thoughts on this shape/if you agree!

lonexreb:
Thanks for the detailed shape @harryalbert — agreed, this is cleaner. Using `preexec_command` as the input source sidesteps the post-exec reconstruction loss entirely, and the simplified `CLIAgent::detect` core (first-token / alias / basename / existing special cases) is the right shape for what the PR is actually trying to do (shebang scripts → basename match).

Quick clarifying question before I rewrite:

For the Deno case oz flagged in the most recent review (`deno run <script>` shape), do you want it left unsupported in this PR — matching your simplification preference — or handled via a small special case alongside the existing aifx / vibe-acp ones? I'm fine either way; just want to align on scope before I open the rewrite.

Will plan to push this as two commits on the same branch:

1. `block: add preexec_command storage + getter, route detection through it`
2. `cli_agent: simplify detect; drop runtime parser`

Let me know on Deno and I'll get the rewrite up.

harryalbert:
> For the Deno case oz flagged in the most recent review (`deno run <script>` shape), do you want it left unsupported in this PR — matching your simplification preference — or handled via a small special case alongside the existing aifx / vibe-acp ones? I'm fine either way; just want to align on scope before I open the rewrite.

Of course! I think we should generally avoid special casing for specific runtimes if possible

lonexreb:
@harryalbert thanks for the clear steer — agreed, no special-casing for Deno. The latest oz review verdict is **Approve with nits** with the Deno suggestion being the sole remaining item, which we're consciously leaving out of scope per your guidance. To keep the supported-runtimes list honest, I can drop Deno from `KNOWN_RUNTIME_INTERPRETERS` if you'd prefer it not appear at all (since `deno run <script>` won't actually recover the basename). Let me know whether you'd like that follow-up or whether the current state is good for a maintainer pass.

lonexreb:
Thanks @harryalbert — that settles it. I'll drop the runtime parser entirely (Node/Python/Deno/Bun handling all goes), keep the simplified `CLIAgent::detect` core (first-token / alias / basename / existing `aifx` / `vibe-acp` special cases), and add `preexec_command` to `Block` plus the routing in `detect_cli_agent_from_model`.

Will push the rewrite as two commits on this branch shortly. Pinging you once it's up.

lonexreb:
@harryalbert rewrite is up on the same branch — pushed as the two commits we discussed:

1. **`block: add preexec_command storage + getter, route CLI-agent detection through it`** ([c755acde](https://github.com/warpdotdev/warp/pull/10022/commits/c755acde)) — adds `preexec_command: Option<String>` to `Block`, populates it from `data.command` inside `Block::preexec` at `app/src/terminal/model/block.rs`, exposes `preexec_command()` returning `Option<&str>`. `detect_cli_agent_from_model` at `app/src/terminal/view/use_agent_footer/mod.rs:378` prefers the preexec form with `command_with_secrets_obfuscated(false)` as fallback. Custom toolbar command patterns continue to match against the reconstructed form for backward compatibility.

2. **`cli_agent: simplify detect; drop runtime parser`** ([e3821b7a](https://github.com/warpdotdev/warp/pull/10022/commits/e3821b7a)) — drops `shebang_script_basename`, `is_env_var_assignment`, and their supporting `SCRIPT_RUNTIMES` / `VALUE_TAKING_RUNTIME_FLAGS` constants. `CLIAgent::detect` now uses just: exact first-token / alias / `path_basename_token` (still gated through `prefix_eligible_for_basename_recovery` for generic prefixes) / `aifx agent run claude` / `vibe-acp`. No Deno special case, per your guidance.

**Net diff vs prior tip of branch: −243 lines.**

### Coverage

- All 150 tests under the `cli_agent` filter pass.
- `cargo clippy -p warp --lib --no-deps` is clean.
- Dropped 4 tests that exercised the removed runtime parser; added `test_detect_shebang_script_via_preexec_path` covering the new path that shebang invocations take (`./claude.py` / `/usr/local/bin/codex.js` resolve through basename + extension stripping after preexec preserves the path form). Updated `test_detect_generic_prefix_no_basename_false_positive` to verify `/tmp/agent.js` / `/tmp/agent.py` stay gated against false-positiving on `CLIAgent::CursorCli`.

Ready for re-review whenever you have a moment.

harryalbert:
Hi @lonexreb, thanks for contributing! Going to close this for now because it's missing visual verification, but please feel free to re-open if you've been able to record a video of the fix working. Thank you!

Issue #5286, opened 2024-09-08
Image protocols


thesombady:
### Dupe Check

- [X] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates

### Describe the solution you'd like?

So I've university student and use LATEX every day. I usually live in the terminal but as of right now, after compiling a pdf I have to open it in Preview (Mac OS). Today, however, I found a cli pdf viewer, https://github.com/itsjunetime/tdf, and was ecstatic about having the possibility to view it in a Panel next to my terminal editor (helix). However, it seems like Warp does not support the image protocols. I would like having those protocols!

### Is your feature request related to a problem? Please describe.

The problem is that the PDF does not render correctly, and it seems to be a problem with Warp itself. I've tried viewing it in Iterm2 and it works fine. But in Warp it renders to the attached file.
![Screenshot 2024-09-08 at 11 28 08](https://github.com/user-attachments/assets/54e6ee24-e6d1-42ad-8d04-c9c7de327dc1)

### Additional context

Since other terminal emulators have this feature, Iterm2, Kitty and many more, has this feature I think it would be a great addition to Warp. Otherwise I would most likely change to using Iterm2 or kitty with tmux :/

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

Ignore

alerque:
Probably a duplicate of #26.

dannyneira:
closing as duplicate of https://github.com/warpdotdev/warp/issues/26 which has been resolved with kitty image protocol support.

Issue #3953, opened 2023-12-09
Display image with viu


kayac-chang:
## Background
As a front-end developer, I have to deal with images very frequently.
I use terminal and neovim for most of my job.
It would be handy to display the image preview in the terminal and neovim, so I don't have to leave the terminal.
So I found the [tool][1], which can help me display images on the terminal.

## Issues
The following uses the same command to display images, 
but I got totally different results in different terminals.

### Iterms2 
<img width="1552" alt="Screenshot 2023-12-09 at 9 35 42 PM" src="https://github.com/warpdotdev/Warp/assets/32451546/2c347130-3ed7-443d-b5f2-e808647df43a">

### Warp
<img width="1440" alt="Screenshot 2023-12-09 at 9 35 52 PM" src="https://github.com/warpdotdev/Warp/assets/32451546/f81d8567-adb8-40be-a07c-d6a642ba477e">

[1]: https://github.com/atanunq/viu

### To reproduce

Follow the install instructions to setup [viu]

[viu]: https://github.com/atanunq/viu

### Expected behavior

The image should be displayed like iterm2.

### Operating system

MacOS

### Operating system and version

Ventura Version 13.6.1

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this a Warp specific issue? (i.e. does it happen in Terminal, iTerm, Kitty, etc.)

No, this same issue happens in Warp and other terminals.

### Warp Internal (ignore): linear-label:b8107fdf-ba31-488d-b103-d271c89cac3e

None

Issue #3271, opened 2023-06-28
Ability to configure Warp to open certain types of through the MacOS "Open with...' context menu


westernwontons:
### Describe the solution you'd like?

```Text
A tab in the settings where I can configure a file extension to be opened with a custom command. For example, when I go to 'Open with...' in Finder, Warp would be there to open markdown files with a custom command like 'nvim <file>'
```

### Is your feature request related to a problem? Please describe.

```Text
The above task is easily achievable by navigating to the folder in Warp, and opening the file with the command. Another way to do it is by using the "New Warp tab here" option at the bottom of Finder.

Having the ability to open certain files with certain commands would be a lot more customizable though.
```

### How important is this feature to you?

1 (Not too important)

### Warp Internal (ignore) - linear-label:770f6576-d6c0-4e4f-a259-fc64b5156087

None

Issue #3035, opened 2023-05-07
Support Image preview in ranger


wombmmm:
### Describe the bug

I can't preview images in ranger. It will show up in other terminals. Setting in rc.conf and scope.sh does not solve the problem.

### To Reproduce

ranger → Image preview doesn't work

### Operating System

MacOS

### Operating System and Version

13.3.1

### Warp Version

v0.2023.05.02.08.03.stable_00

### Does this block you from using Warp daily?

Yes, this issue prevents me from using Warp daily.

### Is this a Warp specific issue? (i.e. does it happen in Terminal, iTerm, Kitty, etc.)

Yes, this I confirmed this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b8107fdf-ba31-488d-b103-d271c89cac3e

None

dannyneira:
Hi @copula sorry to hear about this blocker for you. I can't quite reproduce your issue here. I tried to open a test image with `ranger` and the image preview was opened correctly.

https://user-images.githubusercontent.com/16809145/236936550-c603b636-7a10-4737-b4de-7e8ff7ba3512.mp4

Are you using any other image apps by any chance? Does Warp have the permissions it needs to do things like [notifications](https://docs.warp.dev/features/notifications#troubleshooting-warp-notifications) and [updates](https://docs.warp.dev/help/updating-warp#auto-update-issues)?

LiGaOg:
@dannyneira Image preview shows picture on the rightmost column like this.
![5570fa9dd41185fa](https://github.com/warpdotdev/Warp/assets/72175888/3e6df2b1-7af6-4b40-bbd9-f58b9784b56f)

dannyneira:
Please share a copy of your rc.conf and other relevant files to repro this. I cant seem to get the image preview to work on any terminal at the moment. I tried both `ueberzug` and `w3m` as image preview methods.

LiGaOg:
@dannyneira Use iTerm2 and set true for `preview_images_method`.
<img width="540" alt="image" src="https://github.com/warpdotdev/Warp/assets/72175888/32c072b5-f8c1-4aed-b8d4-6a29bdd7e6c7">

LiGaOg:
@dannyneira Use iTerm2 and set preview method to iTerm2
<img width="540" alt="image" src="https://github.com/warpdotdev/Warp/assets/72175888/924994e2-87b0-4b5a-b35e-ba1cd0945104">

[...]

PR #10024, opened 2026-05-04
Image preview in the Code Editor file pane (GH9729)

### LINKED ISSUE #9729: Image preview in the Code Editor file pane, with directory thumbnail strip
### Pre-submit Checks

- [x] I have searched Warp feature requests and there are no duplicates. Closest existing issues are #1693, #5286, #3953, and #3035, which all target the terminal grid path (escape-sequence rendering inside a shell). #2206 was about an external "Open with" submenu and was closed as not planned. None cover the GUI File Tree to Code Editor preview surface this request is about.
- [x] I have searched Warp docs and my feature is not there. The Code Editor docs and File Tree docs describe text-file behavior only; the Markdown viewer is the closest precedent for a non-text view inside a Code Editor tab.

When clicking an image file in the File Tree (Project Explorer), open it in an Image Preview tab in the Code Editor instead of a text editor tab. The preview should behave as a lightweight photo browser.

**Main view**

- Fit-to-pane by default, preserving aspect ratio; no upscale beyond 100% by default.
- Cmd+= / Cmd+- / Cmd+0 for zoom in / out / reset; trackpad pinch zoom; click-and-drag pans when zoomed.
- Status footer always visible: filename, pixel dimensions, file size, format.
- Animated GIF, animated WebP, and animated AVIF auto-play and loop, with a play/pause toggle that persists across navigation.
- SVG renders as vector (re-rasterized at the current zoom level), sharp at any zoom.
- EXIF orientation honored for both the main view and thumbnails.

**Thumbnail strip below the image**

- Horizontal, virtualized strip listing supported images in the same directory, sorted by case-insensitive natural filename order (so `img2` precedes `img10`).
- Current image highlighted; strip auto-scrolls to keep it visible.
- Left and Right arrows navigate previous and next; Home and End jump to first and last; navigation stops at the ends, no wrap.
- Click swaps the preview in place (single-tab photo-browser feel); Cmd+click opens the thumbnail in a new tab.
- Thumbnails decoded on a background thread pool, cached on disk keyed by (canonical path, mtime, size, target dimension); LRU eviction with a configurable cap (proposed default 256 MB).

**Supported formats:** `.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`, `.svg`, `.heic`, `.heif`, `.bmp`, `.tiff`, `.tif`, `.avif`, `.ico`. Magic-byte sniff overrides the extension when they disagree.

**Reuse:** the existing Agent-context image pipeline (drag-drop, paste, decode, EXIF, downscale, shipped throughout the 2025-07 and 2025-08 changelog) already does the heavy lifting. This feature is largely a new view, a thumbnail-cache layer, and a router from File Tree click to view kind.

**Out of scope (explicit non-goals):** terminal-grid inline rendering (#1693, #5286), editing / cropping / annotation, RAW formats (CR2, NEF, ARW, DNG), PDFs, network or remote URLs, slideshow / full-screen presentation mode, image diffs across git revisions.

Sketch of the layout:

````text
┌─────────── Code Editor tab: "logo@2x.png" ─────────────┐
│                                                          │
│                  [   image content   ]                   │
│                                                          │
│  Fit  100%  -  +    logo@2x.png  1024x1024  124 KB  PNG  │
├──────────────────────────────────────────────────────────┤
│ [thumb] [thumb] [THUMB*] [thumb] [thumb] [thumb] [thumb] │
└──────────────────────────────────────────────────────────┘
````

Today, clicking a `.png`, `.jpg`, `.svg`, `.heic`, or similar in the File Tree opens it as a text file in the Code Editor and shows binary bytes (or, for SVG, raw XML). To actually look at an image asset the user has to leave Warp for Finder, Files, or an external viewer, which breaks flow inside the agentic development environment Warp positions itself as.

This affects practically every front-end, mobile, design-systems, docs, and ML repo. A developer hits this within minutes of using the File Tree on any real project.

The terminal-grid image protocol path (iTerm2 / Kitty) does not solve it: that path only renders when a CLI tool emits escape sequences inside a shell, and a GUI File Tree click never goes through a shell. The "Open with" external-app route was closed as not planned in #2206. The remaining gap is an in-app preview in the Code Editor, and the Markdown viewer already establishes the precedent for a non-text view in the same tab system.

### Additional context

**Behavior invariants and a tech sketch are pre-drafted.** I have a `product.md` (numbered testable behavior invariants covering format support, magic-byte routing, zoom/pan/fit math, thumbnail strip ordering and virtualization, animation, EXIF and color profiles, drag-and-drop into Agent prompts, error states, performance budget, accessibility, and tab-system integration) and a `tech.md` (MediaTypeRouter, ImagePreviewView, ZoomPanController, ThumbnailStripController, ThumbnailCache, settings additions, keybindings, file-touch list, testing matrix, risks around HEIC on Linux and animated AVIF). Both are ready to drop into `specs/GH<issue-number>/` as soon as `ready-to-spec` is applied; happy to open the spec PR immediately after labeling.

**Proposed performance budget.** Thumbnail strip first paint under 100 ms with a cold cache for 50 images. Strip virtualized for directories with more than 500 images. Decoded full-size image memory dropped on tab close. Concurrent full-size decodes capped at one per tab. Thumbnail decode thread pool capped at `min(num_cpus, 4)`.

**Cross-platform notes.**

- HEIC on Linux requires libheif. Needs a packaging and licensing note before commit.
- Color profile fidelity outside macOS documented as best-effort sRGB for v1.
- Animated AVIF may fall back to first-frame for v1 if the existing decoder lacks frame iteration; static AVIF stays in scope.

**Reuse and integration.**

- Decode, EXIF, downscale, color resample: existing Agent-context image pipeline.
- Tab system, drag-split-close, session restore: existing Code Editor tab controller.
- Directory change notifications: existing File Tree watcher (do not instantiate a second one).
- Drag from preview into Agent prompt: emit the same payload type as the existing Agent context attach drag; sink unchanged.

**Related issues:** #1693, #5286, #3953, #3035 (terminal-grid surface, different problem), #2206 (Open with, closed as not planned). Reference precedent: docs.warp.dev/terminal/more-features/markdown-viewer for a non-text view inside the Code Editor tab system.

### Operating system (OS)

macOS (with Linux and Windows explicitly in scope per the proposed spec).

4

NorfeldtKnowit:
## Description

Spec PR for [#9729](https://github.com/warpdotdev/warp/issues/9729) (*Image preview when clicking image files in the File Tree*), which is labeled `ready-to-spec`. Per the maintainer's guidance in the issue thread (@peicodes: "we think this should be implemented using the Lightbox component, which would open up a preview overlay over your pane with the image"), the spec routes File Tree image clicks through the existing `Lightbox` overlay rather than introducing a new tab variant.

This PR adds:

- `specs/GH9729/product.md`: product spec covering opening, navigation, dismiss/focus restoration, error handling, decoder limits, accessibility, multi-window behavior, and edge cases (hidden files, identity click, file-system mutation during preview).
- `specs/GH9729/tech.md`: tech spec grounded in real `file:line` references, structured per the house convention (Context, Relevant code, Proposed changes, End-to-end flow, Risks and mitigations, Testing and validation, Follow-ups).

### Scope of the proposal

1. New `FileTarget::ImagePreview` variant, plus a step-0 short-circuit in `resolve_file_target_with_editor_choice` (`app/src/util/openable_file_type.rs`) so PNG/JPEG/JPEG/GIF/WebP/SVG are captured ahead of both the markdown/code probe (so SVG no longer opens as raw XML) and the binary fall-through to `SystemGeneric` (so raster formats no longer hand off to the OS).
2. A new `Workspace::open_file_with_target` arm that scans sibling images via a new `list_sibling_images_natural_sorted` helper (capped at `MAX_SIBLING_IMAGES = 1024`, hidden-file filter mirrors the clicked file), builds a `Vec<LightboxImage>` of `Resolved { LocalFile { path } }` entries, and dispatches `WorkspaceAction::OpenLightbox`. The existing handler at `app/src/workspace/view.rs:21710-21737` already handles "reuse the open `LightboxView` via `update_params`," so no handler change is needed.
3. Decoder safety in `crates/warpui_core/src/image_cache.rs`: introduce `image::Limits` (max dimension, max alloc) plus an explicit `MAX_DECODED_PIXELS` cap and an `MAX_SVG_BYTES` input cap, so a malformed or maliciously oversize image cannot OOM the process via `into_rgba8()` or animated-frame collection.
4. New `LightboxImageSource::Error { message }` variant in `crates/ui_components/src/lightbox.rs` so per-image read/decode failures render the filename inline rather than spinning forever (the artifacts call site at `app/src/ai/artifacts/mod.rs:362-365` works around this today by stuffing "Failed to load" into the description; that workaround drops once the variant exists).

### What this spec deliberately does **not** propose

These are enumerated as Follow-ups in `tech.md`:

- Zoom (`Cmd+=`/`-`/`0`), trackpad pinch-zoom, click-drag pan.
- Status footer (filename, dimensions, file size, format string).
- Animation play/pause control.
- EXIF orientation rotation and ICC color-profile honoring.
- Visible thumbnail strip (the sibling list exists internally for arrow navigation; no UI strip is rendered in v1).
- Format support beyond PNG/JPEG/GIF/WebP/SVG (HEIC, HEIF, AVIF, BMP, TIFF, ICO).
- Magic-byte content sniffing when extension and content disagree.
- Right-click context menu and drag-out to Agent input.
- Disk-backed thumbnail cache.
- Image diff across git revisions, remote URL preview, RAW formats, slideshow mode.
- Terminal-grid inline images (already tracked in #1693, #5286).

### Adjacent issues

- #1693, #3953, #5286, #3035: terminal inline-image / ranger-preview behavior. Distinct surface; out of scope.
- #2206, #3271: external "Open With" flows. Distinct surface; out of scope.

## Linked Issue

Closes #9729 (spec-only; implementation will follow on a subsequent PR per the house spec-driven flow).

> Spec-only PR; no implementation yet, so no screenshots. Visual verification is enumerated in `tech.md` under *Manual validation* and will be attached on the implementation PR.

## Screenshots / Videos

N/A; spec-only PR.

## Testing

No code changes in this PR. The tech spec's *Testing and validation* section enumerates:

- Unit tests in `app/src/util/openable_file_type.rs` for the new resolver branch (each supported extension; precedence over markdown; non-image binaries still `SystemGeneric`), the natural-sort helper, the hidden-file filter, and the `MAX_SIBLING_IMAGES` cap.
- Unit tests in `crates/warpui_core/src/image_cache.rs` for the decoder-limit guard (huge dimensions rejected, normal photos accepted, garbage returns `Unrecognized`, oversize SVG rejected).
- Manual validation against fixtures including a 10000×10000 PNG (decoder cap), corrupt PNG (per-entry error), animated GIF/WebP, SVG, broken symlink, large directory smoke test, and split-pane / multi-window scrim coverage.
- Standard `cargo fmt --check`, `cargo clippy --workspace --all-targets --tests -- -D warnings`, and `cargo nextest run --no-fail-fast --workspace --exclude command-signatures-v2`.

## Agent Mode

- [x] Warp Agent Mode; This PR was created via Warp's AI Agent Mode


NorfeldtKnowit:
@cla-bot check

NorfeldtKnowit:
Thanks @oz-for-oss. Pushed `faa4860` addressing all three findings:

**🚨 CRITICAL — `tech.md` sibling-list windowing.** Rewrote `list_sibling_images_natural_sorted` to window around the clicked image (up to 512 entries before, 512 after, clamped at the directory's first/last entries) before any truncation. The clicked image is now guaranteed to be inside the kept slice for any directory size. Updated the unit-test list to cover left-edge, right-edge, and centered cases against a 2000-image fixture; dropped the misleading prose paragraph in the Notes section that described the algorithm without implementing it.

**⚠️ IMPORTANT — `product.md` replace-in-place flow.** The "click another image to replace overlay in place" behavior was internally inconsistent with the scrim-blocks-input invariant. Rewrote the section as "Opening a second image while one is already open": the file-tree click steals focus, fires `LightboxViewEvent::FocusLost`, the workspace handler tears down the open Lightbox, and the second click goes through cold open. The `update_params` reuse path in the `OpenLightbox` handler still exists and is preserved for non-focus-stealing dispatch sites (artifact / screenshot Lightboxes from agent block panes), but the file-tree path no longer relies on it. Manual-validation step renamed and reworded accordingly.

**⚠️ IMPORTANT [SECURITY] — `tech.md` eager 1024-sibling decodes.** Restructured change 3 to build only `current ± PRELOAD_RADIUS` (default `PRELOAD_RADIUS = 2`, so 5 entries) as `LightboxImageSource::Resolved`; the rest are `LightboxImageSource::Loading`. `LightboxView::start_asset_loads` only loads `Resolved` entries, so on initial open at most 5 parallel reads/decodes are queued regardless of directory size. On Left/Right, a small navigation hook in `LightboxView::handle_action` walks the new `±PRELOAD_RADIUS` window and dispatches `UpdateLightboxImage` to promote each newly-visible neighbour from `Loading` to `Resolved`, which routes through the existing `update_image_at` → `start_asset_load` path. Peak transient allocation is now bounded at `~5 * MAX_DECODED_PIXELS * 4 bytes ≈ 1.2 GB` (combined with change 4's per-decode cap), versus the pre-fix `~262 GB` worst case. The Risks section is updated to call this out as the chosen mitigation rather than relying on cache eviction reacting to allocation pressure.

/oz-review

NorfeldtKnowit:
Thanks for the second pass @oz-for-oss. Pushed `10e1835` addressing all five findings.

**1. Animated GIF/WebP needs `enable_animation_with_start_time` to repaint** (`product.md:56`). v1 is now scoped to **first-frame static preview** for animated formats. Continuous playback would require wiring `Image::enable_animation_with_start_time(Instant)` (`crates/warpui_core/src/elements/image.rs:128`) into the Lightbox image element plus a per-frame redraw loop, which is enough surface area (animation timing, navigation reset semantics, eventual pause/play interaction) to belong in a follow-up. Files still load successfully and the user sees a static, unambiguous preview. Behavior 9/10, the Success Criteria bullet, and the manual-validation step are all updated; the follow-up entry is now spelled out as "continuous playback via `enable_animation_with_start_time`" with play/pause as the next layer.

**2. Lazy promotion data model is underspecified** (`tech.md:154`). Picked one: extend `LightboxImage` in `crates/ui_components/src/lightbox.rs:38-43` with `path: Option<PathBuf>`. Every entry on the file-tree path carries `Some(p)`; existing call sites (artifacts, screenshots, agent block paths) pass `None` and continue working. The navigation hook in `LightboxView::handle_action` reads `image.path` directly, so there's no filename-reconstruction or parent-dir-stashing needed; the data model is the source of truth.

**3. `collect_frames()` allocates everything before the cap** (`tech.md:191`). Replaced with a streaming `collect_animated_frames_with_limits` helper that iterates `decoder.into_frames()` directly, accumulates `width * height` per frame, and `bail!`s as soon as either the per-frame dimension cap or the running total-pixel cap is exceeded. The remaining iterator items are dropped along with the function's stack frame; peak allocation is bounded by `MAX_DECODED_PIXELS * 4 bytes ≈ 256 MB` plus one in-flight frame, instead of the entire animation. Both the WebP-animated and GIF arms are updated.

**4. `ImageType::Unrecognized` leaves the Lightbox spinning** (`tech.md:222`). `try_from_bytes` now returns `Err` for any unknown format (`Ok(other) => bail!("unsupported image format: {other:?}")` and `Err(_) => bail!("could not detect image format from header bytes")`) instead of `Ok(Unrecognized)`. The `Unrecognized` variant is removed entirely (single-call-site cleanup in `size_in_bytes`). The asset cache then surfaces `AssetState::FailedToLoad`, which `LightboxView` maps to the new `LightboxImageSource::Error { message }` variant from change 5; the user sees a per-entry inline error with the filename. This also catches the extension/content-mismatch case (a `.png` containing tarball bytes) that was previously silent. New unit-test bullet covers the new `Err` behavior.

**5. End-to-end flow contradicts the bounded preload** (`tech.md:368`). Rewrote step 5 to describe building both `Resolved` (preload window) and `Loading` (rest of siblings) entries with `path: Some(p)` carried on every entry. Rewrote step 6 to say `start_asset_loads` queues loads only for `Resolved` entries and noted the `update_params` reuse path is not exercised on the file-tree path. Rewrote step 8 to describe lazy promotion via `UpdateLightboxImage` after each Left/Right and to thread the new `Unrecognized → Err → AssetState::FailedToLoad → LightboxImageSource::Error` failure flow.

/oz-review

NorfeldtKnowit:
Pushed `2aff764` addressing all three notes from the third pass.

**1. Animated decode still materializes every frame up to the cap (`tech.md:255`).** Real bug; the previous round scoped product.md to first-frame-only display but left the decoder in change 4 happily collecting up to `MAX_DECODED_PIXELS` worth of frames per animated input. Collapsed both the WebP-animated and GIF-animated arms to first-frame-only decode, returning `StaticBitmap` directly:

- WebP: always go through `decode_with_limits(data, image::ImageFormat::WebP)` (same path as static WebP). The `WebPDecoder` yields the first frame whether or not the file is animated.
- GIF: read dimensions, cap-check, then `DynamicImage::from_decoder(decoder).into_rgba8()` for one frame; never call `into_frames().collect_frames()` or iterate frames at all.

`ImageType::AnimatedBitmap` is unconstructed in v1 (no in-tree caller actually drives animation today, since `Image::enable_animation_with_start_time` is not wired anywhere consuming it); the variant stays in the enum for the continuous-playback follow-up. The `collect_animated_frames_with_limits` helper from the previous round is dropped from the spec since it is no longer needed. Per-decode envelope is now `MAX_DECODED_PIXELS * 4 bytes` regardless of input being animated or static.

**2. 1.2 GB transient is too generous (`tech.md:177`).** Tightened both knobs:

- `PRELOAD_RADIUS = 2 → 1` (window of `2 * 1 + 1 = 3` entries: prev, current, next).
- `MAX_DECODED_PIXELS = 64M → 32M` (~122 MB RGBA per decode).

Worst-case transient envelope is now `~3 * 122 MB ≈ 366 MB` per file-tree click, down from ~1.2 GB. For typical photo directories (4000×3000 = 12M pixels, well under cap) actual allocation is a small fraction of that. Risks-section math is updated; reference "below cap" fixture for manual validation moves from 7000×7000 to 5000×5000 (25M pixels, just under the new cap); the rejection-path fixture stays at 10000×10000. Tighter envelopes (radius 0, smaller cap, or serializing decodes via a semaphore) are enumerated as a follow-up if real-world telemetry shows 366 MB is still too generous, but I want to leave one neighbour preloaded by default so Right-arrow is responsive in normal use.

**3. A11y label requirement not implemented in tech (`tech.md:311`).** Added a sub-section under change 5 spelling out the implementation:

- `Lightbox::render` per-image render branch sets the rendered `Image` element's accessible label to `current_image.description.as_deref().unwrap_or("image")`.
- `Loading` entries get an accessible label of `"loading {description}"`.
- `Error` entries get an accessible label of `"image preview failed: {description}: {message}"`.
- Implementer should verify the GPUI element accessor name against `crates/warpui_core/src/elements/image.rs` and use the same accessor that other in-tree labeled elements (e.g. `lightbox.rs:128-143` close button) use; the spec mandates the *what* not the exact API name.

Added a manual validation step: enable VoiceOver, click `screenshot.png`, confirm the announcement includes the filename instead of a generic "image" label.

This is the third `/oz-review` so the auto-retrigger budget is exhausted. Leaving the PR up for human review at this point; happy to iterate on anything reviewers raise.

NorfeldtKnowit:
Thanks for the fourth-pass review @oz-for-oss. Pushed `112e581`, which is a substantive narrowing rather than a touch-up; quick summary so you have context if you pick it up.

**Critical (modality contradiction).** Cut sibling navigation, the bounded-preload window, and the click-to-swap flow entirely. v1 is now a single, fully modal, single-image Lightbox: while open, the workspace (file tree, terminals, code editor, tab bar) is inert; opening another file means dismiss-first. That removes the inert-vs-click-through contradiction at its source rather than papering over it.

**Important (pre-read DoS surface).** Added a unified pre-read file-size cap and a bounded `AssetCache` read, so a multi-gigabyte file or a symlink to `/dev/zero` is rejected at the metadata stat before any bytes are read, and a TOCTOU grow-after-stat is bounded by the read cap. Files above the cap surface as the per-entry error state without ever allocating the bytes.

**Important (animated frame materialization).** The previous round's first-frame-only collapse would have regressed the changelog's animated rendering, which I missed. Reverted that: `AnimatedBitmap` frame collection is unchanged so the changelog still animates. Hardened it instead with `image::Limits` plus an explicit frame-count and total-pixel budget enforced *during* iteration (`bail!` mid-stream, drop the iterator), so a pathological animated WebP/GIF cannot allocate every frame before the cap is checked.

**Important (SVG decompression bomb).** Added an SVG intrinsic-dimension cap at parse time, so a 200-byte SVG declaring `<svg width=\"200000\" height=\"200000\">` is rejected before rasterization.

**Suggestion (a11y label honesty).** Verified against `crates/warpui_core/src/elements/image.rs`: GPUI's `Image` element does not currently expose an accessibility-label API. Descoped the rendered-image screen-reader label from v1 and listed it as the first follow-up rather than mandating an API the codebase does not have.

Quick procedural question: my read of the prompt is that the "up to 3 times on the same pull request" budget is aimed at re-review spam where nothing has changed between requests. Does it apply identically when the retrigger follows a substantive code push, or only to re-reviews against the same revision? Happy to leave this for human review either way; mostly want to make sure I am reading the rule correctly.

/oz-review

NorfeldtKnowit:
Thanks for the fifth-pass review @oz-for-oss. Pushed `b6349ff` addressing both findings.

**Important (product/tech consistency on global decoder).** The honest answer is that change 4 *does* harden the shared `ImageType::try_from_bytes` path globally, and centralizing the caps in the decoder is the right place for them rather than a bug. The `product.md` line claiming "no global decoder behavior changes" was the inconsistency. Reworded the Inside-the-Lightbox bullet to state plainly that v1 hardens the shared decoder against pathological inputs (oversized dimensions, animated frame-count and total-pixel budgets, SVG intrinsic-dimension cap), that the budgets are sized well above any in-tree consumer's legitimate inputs, and that the only user-visible global effect is pathological/malicious files surfacing as decode errors at every consumer of `ImageType::try_from_bytes` instead of silently allocating gigabytes of memory. The changelog's animated rendering continues unchanged. Existing Success Criterion 8 and the changelog regression-check in Validation step 9 are already aligned with this wording.

**Important / Security (post-open regular-file revalidation).** Real TOCTOU gap. Updated change 5 in `tech.md` to require a post-open `is_file()` check on the **opened descriptor** (`fstat` against the open file, not a fresh path-based stat) before any byte is read. This closes the swap-between-stat-and-open window: even if the path is replaced with a FIFO, character device, or directory between change 2's pre-read path stat and change 5's open syscall, the check runs against the already-open descriptor and rejects before `read_to_end` can hang on a FIFO or stream until OOM on `/dev/zero`. Spelled out in the change-5 prose, the code block, the Risks > Pre-read size envelope section, a new unit test `local_file_read_rejects_post_open_non_regular_file`, and a new manual validation step 7a (race-replace the path with a FIFO between stat and open). The post-open check is in addition to, not a replacement for, the pre-read size cap and the bounded `MAX + 1` read; together they cover oversize regular files, file-grows-after-stat, file-type-swaps-after-stat, and FIFO-was-the-original-target.

/oz-review

NorfeldtKnowit:
Thanks for the sixth-pass review @oz-for-oss. Pushed `9a043b5` addressing all three findings.

**Important / Security (open() blocks on FIFO before is_file check).** Real gap; the round-5 post-open check only runs if `open()` itself returns. On Linux/macOS, `open(O_RDONLY)` of a FIFO with no writer blocks indefinitely. Updated change 5 in `tech.md` to apply `O_NONBLOCK` via `OpenOptions::custom_flags(libc::O_NONBLOCK)` gated `#[cfg(unix)]`. POSIX `read(2)` on a regular file completes fully regardless of the flag, so no fcntl-clear step is needed before the bounded `take()`. Documented why on Windows the post-open `is_file()` is sufficient on its own (named pipes do not present as regular files there). Added a `#[cfg(unix)]` unit test `local_file_read_does_not_block_on_fifo` that creates a writer-less FIFO via `mkfifo`, passes the path to the asset-cache load future, and asserts the future returns `Err` quickly. Added manual validation step 11b for an explicit FIFO race-replace and updated the Risks > Pre-read size envelope section to enumerate the three coordinated mitigations (pre-read stat in change 2, `O_NONBLOCK` + post-open `is_file()` in change 5, bounded `MAX + 1` read).

**Important / Security (SVG parser CPU/memory under 64 MB cap).** Layered defense in `tech.md`. (a) Introduced `MAX_SVG_BYTES = 4 * 1024 * 1024` (4 MB) in change 5; the bounded read picks `MAX_SVG_BYTES + 1` for `.svg` paths and `MAX_ASSET_LOCAL_FILE_BYTES + 1` for raster. 4 MB is comfortable for any legitimate icon, diagram, or illustration SVG (real-world is usually a few hundred KB). (b) Added a `looks_like_svg_xml` content-sanity prefix check in change 4 that, before `usvg::Tree::from_data` is invoked, peeks the first 1 KB of bytes and confirms a UTF-8-BOM-tolerant whitespace-tolerant `<?xml` or `<svg` prefix. This catches the "binary blob renamed to `.svg`" case and is bounded-time / no-allocation. (c) Added unit tests `local_file_read_caps_svg_at_smaller_limit`, `decode_svg_rejects_non_xml_prefix`, and `decode_svg_accepts_xml_prelude_with_bom_and_whitespace`. (d) Added manual validation step 11a for a 3.5 MB SVG containing 50000 nested `<g>` elements (rejected by the byte cap before parse) and a binary-blob renamed `.svg` (rejected by content-sanity). Tracked a `usvg`/`resvg` parse-budget API as a follow-up; if it ever lands, it becomes the right primary defense and the 4 MB cap can relax.

**Suggestion (product spec conflates pre-read with decode-time caps).** Reworded the Performance posture and Limits sections in `product.md` to call out the two distinct caps. Pre-read size cap rejects oversize *files* before any read; decode-time dimension and pixel caps reject oversize *decoded images* during decode. Added concrete examples covering the matrix: a 3 MB JPEG with normal pixel content passes both; a 100 MB PNG with normal pixel content fails pre-read; a 3 MB JPEG header decoding to 100000×100000 passes pre-read but fails decode; a 100 MB PNG decoding to 100000×100000 fails pre-read (decode never runs). Updated Success criteria 5, 6, 7, and 10 with explicit cap attribution. Extended the Validation fixtures list with FIFO, deeply-nested SVG, and binary-blob `.svg` cases.

/oz-review

NorfeldtKnowit:
Thanks for the seventh-pass review @oz-for-oss. Pushed `1b79a5d` addressing both findings.

**Important / Security (extension-keyed SVG cap is bypassable).** Real bypass; the round-6 design picked the cap from path extension, so an `evil.png` containing 50 MB of SVG XML would have read at the 64 MB raster cap. Switched the cap to **content-keyed** at the asset-cache layer in change 5: after `open` + post-open `is_file()`, peek the first 1 KB, run `looks_like_svg_xml` on the peek, and pick `MAX_SVG_BYTES + 1` (4 MB) if the prefix looks XML/SVG-shaped, otherwise `MAX_ASSET_LOCAL_FILE_BYTES + 1` (64 MB). Extension is no longer used to choose the cap. The reused helper (`looks_like_svg_xml`) is the same one change 4 uses to gate `usvg::Tree::from_data`, so the asset-cache cap and the parser gate share one signal by construction. Reconciled the spec text in change 4's SVG section and Risks > Pre-read size envelope to describe the cap as content-keyed and to spell out why extension-keying was discarded. The parser route in `try_from_bytes` remains extension-keyed (only `.svg` reaches `usvg`) AND content-checked (`looks_like_svg_xml` must pass), so `evil.png` with SVG content still cannot reach `usvg`; the content-keyed read cap is defense-in-depth even if `try_from_bytes` is ever changed to content-sniff. Added unit tests `local_file_read_caps_svg_content_under_png_extension` and `local_file_read_uses_raster_cap_for_non_svg_content`, plus a manual-validation fixture for the same case.

**Important (global LocalFile cap under-surveyed).** Surveyed every current `AssetSource::LocalFile` consumer in the codebase and documented each in a new "Other `LocalFile` consumers and rollout impact" subsection in change 5: image preview (this PR), markdown image embeds in the editor and notebooks, custom theme background images, theme-deletion preview, settings profile image, blocklist images. For each consumer the typical legitimate input envelope (icons / avatars / screenshots / theme backgrounds in the 0.5–5 MB range) sits at least an order of magnitude below the 64 MB raster cap and well below the 4 MB SVG cap, so a single global cap does not regress any current consumer. Decision: keep the global cap as the simpler design. The parameterized `AssetSource::LocalFile { path, max_bytes: Option<u64> }` shape is documented as a follow-up for if/when a video-preview or large-export consumer ever lands and legitimately needs higher caps.

/oz-review

NorfeldtKnowit:
Thanks for the eighth-pass review @oz-for-oss. Pushed `759a5bd` addressing all three consistency findings from the round-7 content-keyed switch.

**Important (product.md describes SVG cap as pre-read).** The cap is post-open, after the first-1-KB content peek; the pre-read metadata cap is uniform at 64 MB across formats because no bytes are read at that point. Reworded the relevant Limits and Performance-posture bullets in `product.md` to state plainly that the SVG-specific cap (`MAX_SVG_BYTES = 4 MB`) fires at the asset-cache bounded read step, not at the pre-read metadata stage, and that the two caps are complementary defense-in-depth (a 50 MB file is rejected by the pre-read cap whether or not it is SVG; a 5 MB SVG is rejected by the post-open content-keyed cap; a 5 MB JPEG is read normally and proceeds to decode). Adjusted Success criterion 10 to match.

**Important (end-to-end flow describes extension-keyed cap).** End-to-end-flow step 7 in `tech.md` previously read "`cap` is `MAX_SVG_BYTES` for `.svg` paths and `MAX_ASSET_LOCAL_FILE_BYTES` otherwise" — pure extension-keyed wording carried over from the round-6 design. Replaced with content-keyed wording: after `open` + post-open `is_file()`, peek the first 1 KB of the file, run `looks_like_svg_xml` on the peek, and select the cap from content regardless of extension. The asset-cache code block in change 5 was already content-keyed since round 7; only the narrative around it needed reconciling. Audit grep confirms no stale design-as-current statements survive; remaining `extension`-keyword hits are anti-pattern contrast wording ("content-keyed, not extension-keyed"), historical context for the round-6→7 switch, or unrelated supported-format-list strings.

**Important (manual fixture below the 4 MB cap).** Bumped the SVG parser-DoS fixture in Manual validation step 11a from "around 3.5 MB" (which would have parsed successfully and not exercised the cap) to "4.5 MB above `MAX_SVG_BYTES`", with the explicit expected outcome "rejected by the content-keyed asset-cache cap, no parse attempted, per-entry error state shown". Added an `evil.png`-renamed parallel (same 4.5 MB SVG payload under a `.png` extension) so the manual run also exercises the content-keyed-not-extension-keyed contract end-to-end, and an under-cap counter-fixture (100 KB legitimate SVG, expected to render normally). Verified the existing unit test `local_file_read_caps_svg_content_under_png_extension` already uses a 4.5 MB fixture so it actually exercises the cap.

/oz-review

NorfeldtKnowit:
Thanks for the approve-with-nits @oz-for-oss. Pushed `6bc517b` addressing the one suggestion.

**Suggestion (SVG prefix predicate broadening).** Real concern: legitimate SVGs in the wild often start with authoring-tool comments (`<!-- Generated by Inkscape -->`) or DOCTYPE declarations (`<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" ...>`), and the round-6 predicate only accepted `<?xml` or `<svg` after optional BOM and whitespace, so those files would have hit the per-entry error path instead of rendering. Broadened `looks_like_svg_xml` to accept four prelude tokens after the optional UTF-8 BOM and bounded ASCII whitespace scan: `<?xml`, `<svg`, `<!--`, and `<!DOCTYPE`. Added an explicit "Supported prelude subset" subsection in `tech.md` enumerating the four accepted forms, noting that the predicate is a coarse content sniff (not a full XML lexer) and that `usvg::Tree::from_data` remains the actual XML/SVG validator. Out-of-scope forms (lowercase `<!doctype`, non-`<?xml` processing instructions, leading-data attributes, etc.) are documented as falling through to the per-entry decode error rather than being rejected at the content-sanity gate. Same helper is used at both layers (asset-cache cap selection in change 5, parser gate in change 4) so broadening covers both. Added unit tests `decode_svg_accepts_doctype_prelude` and `decode_svg_accepts_xml_comment_prelude`; the BOM-and-whitespace and binary-prefix tests are retained.

Leaving the PR for human review at this point. The retrigger budget across these passes has been generous and substantive — thanks for the careful read on the threat model in particular (the FIFO open-blocks-before-fstat finding and the content-keyed-vs-extension-keyed SVG cap bypass were both real bugs in the spec that I would not have caught on my own).

NorfeldtKnowit:
## Pushed: implementation + manual verification

Going from spec-only to spec+implementation on this PR per the documented flow ("Once the specs are approved, implementation generally continues on the same PR"). 41 commits pushed: 17 `GH9729(impl):`, 18 `GH9729(review):`, 2 fixes from today's local manual testing, and 4 spec/scaffolding commits.

### Implementation

The implementation is structured around a Ralph loop checklist at `specs/GH9729/IMPLEMENTATION_TODO.md`, with each bullet landing as one small atomic commit. All bullets through `FINAL` are checked. Items map to the tech spec changes:

- **Change 1** (resolver): `FileTarget::ImagePreview` variant + short-circuit ahead of markdown probe and binary fall-through, with unit tests for each supported extension
- **Change 2** (workspace arm): `FileTarget::ImagePreview` arm with synchronous metadata pre-read (regular-file + 64 MB cap), `truncate_message` helper, `build_image_preview_entry` extracted for unit-testability, plus the four `image_preview_arm_*` tests
- **Change 3** (Lightbox `Error` variant): `LightboxImageSource::Error { message }` + inline render path
- **Change 4** (decoder caps): static-decode (`image::Limits`, `MAX_DECODE_PIXELS = 67M`), animated (`MAX_ANIMATED_FRAMES`, `MAX_ANIMATED_TOTAL_PIXELS`), SVG content-sanity prefix + intrinsic-dimension cap
- **Change 5** (asset-cache): content-keyed read cap (raster vs SVG by 1 KB peek), post-open `is_file()`, `O_NONBLOCK` regression guard, and the `lightbox_view.rs` post-load callback rewriting `FailedToLoad`/`Unrecognized` to `Error`
- **Change 7** (telemetry): acceptance test for `target = "image_preview"` on the existing `CodePanelsFileOpened` event

### Two-reviewer review trail

In parallel with implementation, a separate Ralph loop produced two reviews per item against the spec: **R1 = correctness lens** (spec fidelity, edge cases, security/DoS/resource caps), **R2 = quality lens** (idiomatic Rust, naming, structure, test rigor). Notes live under `specs/GH9729/reviews/<item>-r{1,2}.md`. The tracker in IMPLEMENTATION_TODO.md shows verdicts per item; everything is `pass` or `pass-with-nits` after final pass.

### Today's two follow-up commits (`58e428a`, `410a10e`)

Manual testing of the debug build surfaced one runtime gap that the unit tests did not catch:

- `~/Library/Logs/warp-oss.log` showed the action arm dispatching `WorkspaceAction::OpenLightbox` with the right payload (resolver and arm worked end-to-end, including the 64 MB size cap), but: `Action OpenLightbox { ... } was dispatched, but no view handled it`.
- Root cause: the dispatcher in `crates/warpui_core/src/core/app.rs:1490-1493` walks the focused-view responder chain leaf-to-root and "explicitly does not propagate to parent views". `Workspace::open_file_with_target` is invoked via subscription callbacks (file tree → left panel → workspace), so at the moment of dispatch `Workspace` is not in the chain. The action handler at `view.rs:21725` is registered but never reached.
- The spec cited `app/src/ai/artifacts/mod.rs:311` as the precedent for this dispatch pattern. That precedent is correct for that site (artifact thumbnail click is itself a focused-view action, so `Workspace` is in the chain), but does not generalize to a workspace self-call from a subscription callback.

Fix in `58e428a`: extract the `OpenLightbox` action-arm body into `Workspace::open_lightbox(images, initial_index, ctx)` and call it directly from the new arm. The existing action arm becomes a 1-line forward to the same helper, so `mod.rs:311` and `block.rs:6371` are unchanged. This matches the pattern of the other arms in `open_file_with_target` (`open_code`, `open_file_notebook`, `attach_path_as_context`, `cd_to_directory`) which all use direct method calls.

`410a10e` updates `tech.md` (lines 17, 37, 121, 164, 545, 546, 557, 620–623) to describe the helper-call shape and document the responder-chain rationale, with a pointer to the dispatcher comment in `app.rs`.

### Manual verification (debug `warp-oss` build, macOS)

Tested against fixtures at `/tmp/gh9729-fixtures/`:

| Scenario | Outcome |
|---|---|
| Small PNG, JPEG, static GIF | Lightbox opens with image |
| Valid SVG | Lightbox opens with rendered SVG |
| Animated GIF | Lightbox opens to first frame static (per spec; v1 does not call `enable_animation_with_start_time`) |
| 65 MB `.png` | Inline `Error`, "image is too large to preview" |
| SVG with garbage byte prefix | Inline `Error` from the change-5 post-load callback |
| Markdown file | Routes to existing markdown editor (resolver precedence preserved) |
| Non-image binary | Falls through to `SystemGeneric` (pre-GH9729 behavior, unchanged) |
| Escape / scrim click / × | All dismiss; focus returns to file tree |

### Presubmit (already run as part of the FINAL impl commit `8a5c2e6`)

- `cargo fmt` clean
- `cargo clippy --workspace --exclude command-signatures-v2 --all-targets --tests -- -D warnings` clean
- `cargo nextest run --no-fail-fast --workspace --exclude command-signatures-v2`: 5933 tests, all 38 GH9729-touched tests passing. The 7 non-GH9729 failures (SSH integration, git tag display, settings migration marker) reproduce on master and are unrelated to this work.

CI will rerun on this push. Happy to address any nits from the next review pass.

NorfeldtKnowit:
## Pushed: rebased onto master + Tier 2 review trail (39 commits)

Force-pushed the branch after rebasing onto `origin/master` to clear the conflicting state. Two conflicts resolved during rebase, both mechanically additive:

- `app/src/workspace/view_tests.rs` — upstream PR #10373 renamed `view_test.rs` → `view_tests.rs`. Kept both upstream's `test_tab_mru_order` and our four `image_preview_arm_*` tests.
- `app/src/util/image.rs` — upstream PR #9553 added `use std::path::Path; use mime_guess::from_path;`; our `t2-9` added `ImageDecoder` and `metadata::Orientation` to the `image::` import line. Resolution is the superset of both.

Local verification: `cargo check -p warp --tests` clean, `cargo nextest run -E 'test(image_preview_arm_)'` 4/4 pass.

New commits since the previous Oz pass on 2026-05-06: the **Tier 2 review trail** for 21 items (`t2-1` ... `t2-21`), each with R1-correctness + R2-quality self-reviews under `specs/GH9729/reviews/tier2-*.md` and follow-up fix commits where a verdict warranted it. Highlights: t2-9 wires EXIF orientation in both the lightbox and the agent decoder, t2-19/t2-20 ship the custom `PanClippedImage` element with persistent drag state, t2-21 adds the 1.25× zoom step + double-tap zoom-and-center. Full tracker in `specs/GH9729/TIER2_TODO.md`.

/oz-review

NorfeldtKnowit:
## Heads-up: first contribution

This is my first contribution to `warpdotdev/warp`, so I want to flag that upfront. I tried my best to follow the contribution guide and the documented spec-first flow:

- Opened spec-only first (`product.md` + `tech.md`), iterated through 8 rounds of `@oz-for-oss` review until approve-with-nits.
- Crossed over to implementation in the same PR per the documented "Once the specs are approved, implementation generally continues on the spec PR" guidance.
- Worked the post-v1 follow-up list (`tech.md` §688-713) as a Tier 2 UX-polish loop with per-item Impl → R1-correctness → R2-quality cycles; tracker and review files are checked in under `specs/GH9729/`.
- Rebased onto fresh `origin/master`, then landed 5 manual-review polish commits on top (vertical thumbnail rail, animated thumbs, zoom-pill centering, scrim and rail fixes).

If I've missed any convention or step that a Warp team member would normally expect, I'd appreciate the correction and I'll fix it. Happy to split, squash, or restructure however reviewers prefer.

PR #10057, opened 2026-05-04
Fix find focus drifting when output streams into active block

### LINKED ISSUE #9542: Find (cmd+f) selection jumps to a different match when new output streams in, instead of staying on the same text
## Summary
When using cmd+f to search a block whose output is still streaming (for example, logs from a long-running server), the currently selected match jumps to a different piece of text as new matches are appended, instead of staying anchored to the originally selected match.

## Problem
The find UI appears to identify the "currently selected" match by its index in an ordered list of matches rather than by a stable handle to the actual matched span. When new matches arrive earlier in the list (or the list is rebuilt to include the new output), the selection stays at the same numeric index (e.g. still "5 of N") and the highlighted text visibly changes, even though the user's selection should have shifted to a higher index to remain on the same text.

This makes find effectively unusable for inspecting a specific occurrence of a string in a block whose output is still being appended: the user keeps losing the location they navigated to whenever new matches arrive.

## Expected behavior
While the match list grows due to new live output, the selection should remain anchored to the same underlying matched text. If new matches appear before the current one, the index should update accordingly (e.g. "5 of 100" becomes "6 of 101"), but the highlighted text should not change.

## Actual behavior
The selection stays at the same numeric index as new matches stream in (e.g. "5 of 100" → "5 of 101") and visually jumps to whatever match now occupies that position. The text the user had navigated to is no longer highlighted, and they have to re-find it manually.

1. Run a long-running command in a Warp block whose stdout/stderr produces lines containing some search term (for example, a server that periodically logs the word `request`).
2. While output is still streaming, press cmd+f and search for that term.
3. Use the find navigation to step to a specific match (e.g. result 5 of 100) and note which line is highlighted.
4. Wait for new output containing additional occurrences of the search term to be appended.
5. Observe that the highlighted match changes to a different line while the counter stays at "5 of N+k". The originally selected text is no longer the selected match.

## Artifacts
None attached

## Warp version
Unknown

## Operating system
macOS

amazansky:
## Description

When using cmd+f to search a block whose output is still streaming, the currently selected match jumps to a different piece of text as new matches are appended. This happens because `BlockListFindRun::rerun_on_block()` preserves the raw focused-match index when splicing updated matches for the active block — as new matches stream in, the same numeric index points to a different span.

**Fix:** Before splicing, snapshot the currently focused `BlockListMatch`. After splicing, relocate it in the updated vector by comparing span identity (`block_index`, `grid_type`, `range`) via a new `same_span()` helper. If the focused match was outside the rerun block's slice, shift the index by the delta in match count. Also handles the "no old matches" (new block) case by shifting the index forward.

## Linked Issue

Closes #9542

- [x] The linked issue is labeled `ready-to-implement` (triaged bug — implicitly ready-to-implement per CONTRIBUTING.md).

## Testing

Added two regression tests in `block_list_test.rs`:
- `test_rerun_on_block_preserves_focused_match_in_active_block` — focuses the prompt match inside a streaming block, appends new output with additional matches, verifies focus stays on the same span.
- `test_rerun_on_block_preserves_focused_match_in_older_block` — focuses a match in an older finished block, appends output to the active block, verifies the focused match and its UI index shift correctly.

All 10 terminal find tests pass. `cargo fmt` and `cargo clippy -p warp --tests` pass clean.

## Agent Mode

[Conversation link](https://app.warp.dev/conversation/1129ea71-41d4-4f5c-a0c7-771b3bc714e4)

CHANGELOG-BUG-FIX: Fixed find (cmd+f) selection jumping to a different match when new output streams into the active block.
-->

### DIFF
--- a/app/src/terminal/find/model/block_list.rs
+++ b/app/src/terminal/find/model/block_list.rs
@@ -288,6 +288,40 @@ impl BlockListMatch {
             _ => false,
         }
     }
+
+    /// Returns `true` if `self` and `other` refer to the same matched span, ignoring transient
+    /// state like `is_filtered`.
+    fn same_span(&self, other: &BlockListMatch) -> bool {
+        match (self, other) {
+            (
+                BlockListMatch::CommandBlock(BlockGridMatch {
+                    grid_type: g1,
+                    range: r1,
+                    block_index: b1,
+                    ..
+                }),
+                BlockListMatch::CommandBlock(BlockGridMatch {
+                    grid_type: g2,
+                    range: r2,
+                    block_index: b2,
+                    ..
+                }),
+            ) => g1 == g2 && r1 == r2 && b1 == b2,
+            (
+                BlockListMatch::RichContent {
+                    match_id: id1,
+                    view_id: v1,
+                    index: i1,
+                },
+                BlockListMatch::RichContent {
+                    match_id: id2,
+                    view_id: v2,
+                    index: i2,
+                },
+            ) => id1 == id2 && v1 == v2 && i1 == i2,
+            _ => false,
+        }
+    }
 }
 
 /// Represents the result of a find "run" on the blocklist.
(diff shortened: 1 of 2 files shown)


seemeroland (approved):
Nice find and fix - I was able to reproduce the issue and confirm that this branch fixes it locally. Thanks for the contribution!

PR #10109, opened 2026-05-05
Fix Panes-mode vertical tab search for custom tab names

### LINKED ISSUE #9666: Sidebar Panes-mode search doesn't match custom tab names shown in the group header
## Summary

In the vertical-tabs sidebar's "search tabs…" input, typing a tab's custom name returns no results when the sidebar is in **Panes** display mode — even though that custom name is rendered on screen as the group header above the pane rows.

## Repro

1. Open a workspace with multiple tabs.
2. Switch the vertical-tabs sidebar to **Panes** display mode.
3. Rename a tab (e.g. to `deploy`).
4. Type `deploy` in the sidebar search input.

**Expected:** the tab and its panes remain visible.
**Actual:** zero results, even though `deploy` is visible on screen as the group header above the tab's panes.

## Where in the code

Two filter sites in `app/src/workspace/view/vertical_tabs.rs` drop the tab's `custom_title` from per-pane search fragments in Panes mode via a `!uses_outer_group_container(display_granularity)` gate:

- `matching_tab_indices` — vertical_tabs.rs:1022-1024 (drives keyboard tab navigation under search)
- `render_groups` — vertical_tabs.rs:1547-1549 (drives the visible filtered Panes-mode list)

The same expression at vertical_tabs.rs:1777-1779 is correct on the **render** side: it prevents the tab name from being repeated on every pane row in Panes mode (the name lives in the group header above instead). But the two filter sites build a throwaway `PaneProps` whose only job is to feed `pane_matches_query` — those instances are never rendered. The shared-field design (`display_title_override` is read by both render and search) made the search-side drop an accidental consequence of the render rule.

## Proposed fix

Drop the gate at the two filter sites; leave the render gate at line 1777 untouched.

\`\`\`rust
// before (both filter sites)
let title_override = (!uses_outer_group_container(display_granularity))
    .then(|| pane_group.custom_title(app))
    .flatten();

// after
let title_override = pane_group.custom_title(app);
\`\`\`

## Scope

- Sidebar Panes-mode tab-name search only.
- Summary and FocusedSession modes already include the tab custom name in search; pane-name search is unaffected.
- Render path unchanged — tab name still appears only as the group header in Panes mode.

## Relationship to #9155

#9155 is broader: extending custom tab/pane name search across the command palette and sidebar. This issue is the narrow sidebar-only Panes-mode hole, which can ship independently.

CarlAmko:
## Summary
- include custom tab group titles in vertical-tabs search matching without replacing generated pane-row search fragments
- keep Panes-mode row rendering unchanged so pane rows do not duplicate the group header title
- add regression coverage for search-vs-render title behavior

## Issue
Fixes #9666

## Test plan
- cargo fmt --check
- cargo test -p warp custom_tab_title
- cargo test -p warp panes_granularity_search_adds_group_title_without_replacing_pane_title

CHANGELOG-BUG-FIX: Fixed vertical tabs Panes-mode search so renamed tabs can be found by their visible group title.


CarlAmko:
@cla-bot check

zachbai (approved):
Nice, thanks for the change

PR #10116, opened 2026-05-05
Fix proper sentry initialization

kevinyang372:
## Description
Tech spec here: https://github.com/warpdotdev/warp/blob/70a0bed2f31a126495f8190e9cb5252cb9fe3ee6/specs/daemon-sentry-initialization/TECH.md

## Testing
Tested locally and confirmed sentry event was produced: https://warpdotdev.sentry.io/issues/7459771029/?project=5660375&query=is%3Aunresolved&referrer=issue-stream

### DIFF
--- a/app/src/remote_server/server_model.rs
+++ b/app/src/remote_server/server_model.rs
@@ -545,6 +545,7 @@ impl ServerModel {
     ///
     /// Also configures Sentry crash reporting based on the user's identity
     /// and preferences supplied by the connecting client.
+    #[cfg_attr(not(feature = "crash_reporting"), allow(unused_variables))]
     fn handle_initialize(
         &mut self,
         msg: Initialize,
@@ -570,8 +571,6 @@ impl ServerModel {
                 crate::crash_reporting::uninit_sentry();
             }
         }
-        #[cfg(not(feature = "crash_reporting"))]
-        let _ = ctx;
 
         let server_version = ChannelState::app_version().unwrap_or("").to_string();
         HandlerOutcome::Sync(server_message::Message::InitializeResponse(
--- a/app/src/remote_server/server_model.rs
+++ b/app/src/remote_server/server_model.rs
@@ -131,6 +131,37 @@ impl PendingFileOps {
     }
 }
 
+/// Client-supplied auth credentials and user identity for the daemon.
+///
+/// Populated by `Initialize` and `Authenticate` messages. The auth token
+/// is used for server API calls; the user identity is forwarded to Sentry
+/// so crash reports from the daemon are attributed to the connecting user.
+///
+/// All fields are intentionally retained across proxy connection teardown
+/// and cleared only by daemon process exit.
+struct DaemonAuthContext {
+    /// Bearer credential set by Initialize / Authenticate.
+    auth_token: Option<String>,
+    /// User ID from the most recent `Initialize` handshake (Firebase UID).
+    #[cfg(feature = "crash_reporting")]
+    sentry_user_id: String,
+    /// User email from the most recent `Initialize` handshake.
+    #[cfg(feature = "crash_reporting")]
+    sentry_user_email: String,
+}
+
+impl DaemonAuthContext {
+    fn new() -> Self {
+        Self {
+            auth_token: None,
+            #[cfg(feature = "crash_reporting")]
+            sentry_user_id: String::new(),
+            #[cfg(feature = "crash_reporting")]
+            sentry_user_email: String::new(),
+        }
+    }
+}
+
 /// The top-level server-side orchestrator model.
 ///
 /// Receives `ClientMessage`s from connected proxy sessions and routes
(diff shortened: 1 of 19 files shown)


kevinyang372:
* **#10116** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10116?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10116?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/remote_server/server_model.rs
```diff
@@ -130,6 +130,37 @@ impl PendingFileOps {
...
+/// so crash reports from the daemon are attributed to the connecting user.
+///
+/// All fields are intentionally retained across proxy connection teardown
+/// and cleared only by daemon process exit.
+struct DaemonAuthContext {
+    /// Bearer credential set by Initialize / Authenticate.
+    auth_token: Option<String>,
+    /// User ID from the most recent `Initialize` handshake (Firebase UID).
+    #[cfg(feature = "crash_reporting")]
+    sentry_user_id: String,
```
alokedesai: Why is this prefixed with `sentry_`? It's really just the user_id, no?

app/src/remote_server/server_model.rs
```diff
@@ -130,6 +130,37 @@ impl PendingFileOps {
...
+}
+
+impl DaemonAuthContext {
+    fn new() -> Self {
+        Self {
+            auth_token: None,
+            #[cfg(feature = "crash_reporting")]
+            sentry_user_id: String::new(),
+            #[cfg(feature = "crash_reporting")]
+            sentry_user_email: String::new(),
```
alokedesai: if there's no user, this should be unset instead of an empty string

app/src/remote_server/server_model.rs
```diff
@@ -562,7 +585,25 @@ impl ServerModel {
...
+        if !self.auth.sentry_user_id.is_empty() {
+            crate::crash_reporting::set_user_id(
+                crate::auth::UserUid::new(&self.auth.sentry_user_id),
+                if self.auth.sentry_user_email.is_empty() {
+                    None
+                } else {
+                    Some(self.auth.sentry_user_email.clone())
+                },
+                ctx,
+            );
```
alokedesai, line 602: Can we not directly populdate the `AuthManager` / `AuthStateProvider` upon receiving the initialize message? I doubt this is the only model that we need for remote code that reads those two models

app/src/terminal/writeable_pty/remote_server_controller.rs
```diff
@@ -94,15 +95,29 @@ impl<T: EventLoopSender> RemoteServerController<T> {
...
             auth_state,
             ServerApiProvider::as_ref(ctx).get_auth_client(),
-            crash_reporting_enabled,
+            crash_reporting_enabled.clone(),
         ));
+
+        // Keep the shared crash-reporting flag in sync with the user's
+        // privacy settings so that future daemon handshakes send the
+        // current value rather than a stale snapshot.
+        let privacy_settings = PrivacySettings::handle(ctx);
```
alokedesai: It's not just future daemon handshakes thuogh...we'll need to send an explicit message to the server when the user changes their crash reporting setting right?

app/src/remote_server/server_model.rs
kevinyang372: This is a follow-up Moira is already working on. I will keep it minimal for now but this can hook into her work

app/src/terminal/writeable_pty/remote_server_controller.rs
kevinyang372: This code path is for initial connection -- the explicit notification on settings change is handled in app/src/remote_server/mod.rs

Although I agree this is unnecessarily complicated here, I cleaned up the code path to compute this on the fly instead

app/src/remote_server/server_model.rs
```diff
@@ -513,16 +543,36 @@ impl ServerModel {
...
+            }
+
+            if msg.crash_reporting_enabled {
+                self.apply_sentry_user_id(ctx);
+            } else {
+                crate::crash_reporting::uninit_sentry();
+            }
         }
+        #[cfg(not(feature = "crash_reporting"))]
+        let _ = ctx;
```
alokedesai: not blocking but this is a clunky. Just use it as `_ctx` or allow the clippy unused params rule

app/src/remote_server/ssh_transport.rs
alokedesai: not blocking but tests should go in its own file

crates/remote_server/src/auth.rs
```diff
@@ -2,30 +2,44 @@ use std::sync::Arc;
...
+    get_user_email: Arc<GetStringFn>,
+    crash_reporting_enabled: bool,
 }
 
 impl RemoteServerAuthContext {
     pub fn new(
         get_auth_token: impl Fn() -> BoxFuture<'static, Option<String>> + Send + Sync + 'static,
         remote_server_identity_key: impl Fn() -> String + Send + Sync + 'static,
+        get_user_id: impl Fn() -> String + Send + Sync + 'static,
+        get_user_email: impl Fn() -> String + Send + Sync + 'static,
```
alokedesai: Can you explain why we need a closeure here? It's not clear at all to me why we can't directly pass the value?

crates/remote_server/src/client/mod.rs
```diff
@@ -186,12 +186,18 @@ impl RemoteServerClient {
     pub async fn initialize(
         &self,
         auth_token: Option<&str>,
+        user_id: &str,
+        user_email: &str,
+        crash_reporting_enabled: bool,
```
alokedesai: not blocking

1) this should take owned strings, not `str`s
2) It should probably be a struct with named fields (very easy to confused `user_id` and `user_email` as currently writte

alokedesai (approved):

Issue #9869, opened 2026-05-01
Voice input hotkey picker should support Caps Lock, individual F-keys, and custom key bindings


tnederlof:
## Summary
The voice input hotkey picker in **Settings > Agents > Warp Agent > Voice** offers a fixed list of modifier/function keys. Users cannot bind voice activation to other ergonomic keys such as Caps Lock or individual F-keys (F1–F12), even though those keys are well-suited to a push-to-talk use case and are commonly used for this purpose in other apps.

## Problem
The current pre-defined key list is:
- None
- Fn
- Option (Left / Right)
- Control (Left / Right)
- Command (Left / Right)
- Shift (Left / Right)

Keys like **Caps Lock** and **individual F-keys (F1–F12)** are absent. Both are low-conflict and ergonomically convenient: Caps Lock is commonly remapped for push-to-talk in voice/video apps, and F-keys are isolated from the normal typing flow and a natural hold-to-record key. There is no way to supply a custom key outside the fixed list.

## Desired workflow
1. Open **Settings > Agents > Warp Agent > Voice**.
2. Open the hotkey picker dropdown.
3. Select **Caps Lock**, an individual **F-key (e.g. F13)**, or any other bindable key not currently listed — or press a custom key to bind it directly.
4. Voice recording activates/deactivates on press-and-hold of the chosen key as expected.

## Expected behavior
Users can bind voice input to any key that macOS exposes as bindable, including at minimum **Caps Lock** and **individual F-keys (F1–F12)**, either through an expanded dropdown or a press-to-capture custom binding field.

## Actual behavior
The hotkey picker only exposes a fixed list of ten modifier/function keys. Caps Lock and individual F-keys are not available, and there is no free-form binding option.

## Warp version
v0.2026.04.27.15.32.stable_03

## Operating system
macOS 26.4.1 (Build 25E253)

PR #10127, opened 2026-05-05
spec: voice input hotkey picker — Caps Lock, F-keys, custom (#9869)

lonexreb:
## Summary

Spec for issue #9869 — the voice input hotkey picker in **Settings → Agents → Voice** offers a fixed list of 10 modifier/function keys, missing **Caps Lock** and **individual F-keys (F1–F12)**, and providing no escape hatch for any other key.

## Investigation

- Setting is `VoiceInputToggleKey` enum at [`app/src/settings/ai.rs:110`](app/src/settings/ai.rs). Closed enum with hand-written `display_name`, `to_key_code`, `keystroke`, and `tooltip_message`.
- Underlying [`KeyCode` at `crates/warpui_core/src/platform/keyboard.rs`](crates/warpui_core/src/platform/keyboard.rs) already exposes `CapsLock` and `F1`–`F35`; the enum is artificially narrower than what the platform layer can identify.
- Persistence is `SyncToCloud::Never` (intentional — same person uses different keyboards on different devices). Preserved.
- Action dispatch at [`block_list_element.rs:3072`](app/src/terminal/block_list_element.rs) and [`alt_screen_element.rs:620`](app/src/terminal/alt_screen/alt_screen_element.rs) already compares incoming `key_code: KeyCode` to the configured one, so new variants don't require dispatch-layer changes.

## What's in the spec

**`product.md`** — 8 testable behavior invariants (B1–B8), 6 acceptance criteria (A1–A6), explicit non-goals (chord hotkeys, per-tab/profile, cloud sync), and 5 risks with concrete TECH decisions to make.

**`tech.md`** — picks the data-model approach with explicit reasoning, names every file and line range touched, and lays out 6 unit tests + 4 modal tests + 3 integration tests.

## Architectural choices

- **`Custom(KeyCode)` variant on the existing enum, not a separate override field.** Single source of truth for "what key fires voice input"; avoids a "two settings disagree" failure mode.
- **Sectioned dropdown** ("Modifier keys" / "Function keys" / "Other") to keep 22 entries scannable. Existing modifiers stay at the top so existing users see no reorder.
- **Press-to-capture modal** for the "Custom key…" entry, with a reject-list (app-quit shortcut, Enter/Escape/Tab/Backspace/Space) and an >800ms held-modifier carve-out so users can still bind a bare modifier through this path.
- **macOS Caps Lock uses press-to-toggle** semantics (one tap on, next tap off) because macOS Caps Lock fires only on toggle events, not press/release. Discord and Krisp both take this approach.
- **Validation reset** for hand-edited TOML with unbindable values, with a one-time toast.

## Test plan

- [ ] Unit T1–T6 (TOML round-trip, display name, key mapping, toggle-mode, validation reset)
- [ ] Modal T7–T10 (capture flow, reject-list, Escape, held-modifier carve-out)
- [ ] Integration IT1–IT3 (full modal flow, Caps Lock toggle on macOS, F-key hold-to-talk)
- [ ] Manual: existing user with `voice_input_toggle_key = "alt_left"` in TOML sees no behavior change after upgrade
- [ ] Manual: hand-edited `voice_input_toggle_key = "enter"` shows toast and resets to None

## Open questions for maintainer review

1. Does the `SettingsValue` derive macro support enum variants with payloads (`Custom(KeyCode)`), or do we need a manual `Serialize`/`Deserialize` impl?
2. Is there an existing `DropdownItem::heading()` constructor for sectioning, or should this PR add one?
3. Confirm the macOS Caps Lock press-to-toggle decision (vs documenting the difference and not handling).
4. Should the press-to-capture modal expose the captured `KeyCode` spelling (debug aid) or only the friendly display name?

Closes (spec-only) #9869 — implementation PR will follow once spec direction is confirmed.

### DIFF
--- a/specs/GH9869/product.md
+++ b/specs/GH9869/product.md
@@ -131,9 +131,15 @@ replace or rename any existing variant.
 The press-to-capture flow rejects (with an inline error message and
 no save) any key that would brick the user's keyboard:
 - The currently-bound app-quit shortcut.
-- The currently-bound voice-input shortcut itself (no infinite loop).
 - Enter, Escape, Tab, Backspace, Space (would block the agent input).
 - Modifier-only without the >800ms held-modifier rule (B3 carve-out).
+
+> **Correction (re-review #10127):** the previous draft included
+> "the currently-bound voice-input shortcut itself" as a reject-list
+> entry. That would have prevented users from changing an existing
+> custom hotkey — the literal opposite of the modal's purpose. The
+> tech spec already removed the rule (capture is accepted; same-key
+> re-confirm is a no-op short-circuit). Product B5 now matches.
 
 ### B6 — "None" remains the disable option
 
--- a/specs/GH9869/product.md
+++ b/specs/GH9869/product.md
@@ -56,18 +56,40 @@ escape hatch for everything else.
 
 ## Behavior contract (V1)
 
-### B1 — Caps Lock as a first-class option
+### B1 — Caps Lock as a first-class option (platform-specific semantics)
+
+> **Correction (review #10127):** earlier drafts described Caps Lock
+> as hold-to-talk on every platform while tech.md required macOS
+> press-to-toggle. Resolved to a single source of truth below.
 
 Caps Lock appears in the dropdown on all platforms (macOS, Windows,
-Linux). When selected, holding Caps Lock activates voice input and
-releasing it deactivates it, identical to the current modifier-key
-behavior. The OS-level Caps Lock toggle (the LED, the
-uppercase-letters effect) is NOT consumed: holding Caps Lock for
-voice input still toggles the OS Caps Lock state. (Suppressing the OS
-behavior would require platform-specific event interception which is
-out of V1 scope; users who want the suppression can remap Caps Lock
-to F19 at the OS level and bind that, which is the current best
-practice in Discord/Krisp.)
+Linux). The activation semantics are **platform-specific** and
+documented to the user in the dropdown's tooltip:
+
+- **macOS — press-to-toggle:** macOS reports Caps Lock as a single
+  key event when the lock state changes, not as a sustained press.
+  Holding-to-talk is impossible without OS-level event interception
+  (out of V1 scope). One tap starts voice input recording; the next
+  tap stops it. The dropdown tooltip on macOS reads
+  *"Tap to start, tap again to stop."* This matches Discord and
+  Krisp.
+- **Windows — hold-to-talk:** standard behavior. Holding Caps Lock
+  activates voice input; releasing it deactivates. The OS-level Caps
+  Lock toggle (the LED, the uppercase-letters effect) still fires —
+  V1 does not suppress it. The dropdown tooltip on Windows reads
+  *"Hold to talk."*
+- **Linux — hold-to-talk:** same as Windows. Tooltip reads
+  *"Hold to talk."*
+
+Acceptance criteria A4 is split accordingly:
+- A4-mac: tap, then tap again, on macOS — voice activates then
+  deactivates.
+- A4-win-lin: hold and release on Windows or Linux — voice activates
+  while held, deactivates on release.
+
+Users on macOS who want hold-to-talk semantics can remap Caps Lock
+to F19 at the OS level and bind F19; this is the current best
+practice in Discord/Krisp.
 
 ### B2 — F1 through F12 as first-class options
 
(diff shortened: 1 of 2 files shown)


lonexreb:
Pushed corrections addressing all 6 bot review concerns (1 critical + 4 important + 1 security):

- **Critical: macOS Caps Lock contradiction.** B1 split into platform-specific semantics: macOS uses press-to-toggle (matches the macOS Caps Lock event model and Discord/Krisp UX), Windows/Linux use hold-to-talk. A4 split into A4-mac and A4-win-lin.
- **Important: `Custom(KeyCode)` breaks `EnumIter`/`all_possible_values()`.** Added a "Replacing the EnumIter flow" section with a concrete `predefined_options()` returning curated unit variants. `Custom` is constructed at capture time, not iterated. `EnumIter` derive is dropped.
- **Important: serialization plan.** Added "Settings serialization" section recommending a hand-written `Serialize`/`Deserialize` impl over a one-off `implement_setting_for_enum!` extension.
- **Important: TOML format conflict.** Canonicalized to `{ custom = "f19" }` with the `KeyCode` payload as snake_case lowercased. Tests assert exact strings.
- **Important: app-quit reject-list.** Now compares full `Keystroke` (key + modifiers), so `Cmd+Q` is rejected without banning bare `Q`. Static structural rejects (Enter/Escape/Tab/Backspace/Space) remain bare-key.
- **Important: validation flow.** A6 split into Path 1 (raw parse failure intercepted in the hand-written `Deserialize` impl via a side-channel `Mutex<Option<String>>`) and Path 2 (parsed-but-unbindable post-parse `validate()`). Both fire the same toast.
- **Security: telemetry guardrails.** B8 now explicitly limits telemetry to the final saved value — never raw captures, rejected keys, or cancelled captures. Respects global opt-out.

Each correction includes a "Correction (review #10127)" callout in the spec for cross-reference. Ready for re-review.

/oz-review

lonexreb:
Round-2 corrections pushed (commit 7387e88):

- **A4 / Caps Lock contradiction:** A4 split into A4-win-lin (hold-to-talk for all keys) and A4-mac (Caps Lock tap-to-toggle, all other keys hold-to-talk). Mirrors B1's existing platform breakdown.
- **Self-conflict rule preventing custom-hotkey changes:** removed the "reject voice-input self-conflict" rule from the modal reject-list. Same-key re-confirm is handled by accepting the capture and short-circuiting the dispatch when the captured `KeyCode` equals the current one. The dispatch handler at `block_list_element.rs:3072` only fires voice activation, never re-binds — no infinite-loop risk.
- **Field deserializer feasibility:** replaced the side-channel-Mutex-from-inside-`Deserialize` design (not feasible Serde) with a `#[serde(untagged)]` wrapper enum `VoiceInputTomlValue { Valid(...), Invalid(toml::Value) }`. Standard Serde, no container-mutation from the field deserializer. A single post-init pass on `AISettings::initialize` drains the `Invalid` branch, logs, rewrites the TOML, fires the toast.
- **TOML rewrite vs. left-unchanged:** product.md A6 updated to "TOML is rewritten" so the toast fires once, not on every launch. Now matches tech.md.

/oz-review

lonexreb:
Round-3 corrections pushed (commit e8eca85):

- **Self-conflict in B5:** removed "currently-bound voice-input shortcut" from the product reject-list. Tech round-2 already removed it; product now matches.
- **Persisted Custom(KeyCode) display:** new "Displaying a persisted Custom(KeyCode) in the dropdown" section. Virtual "Custom: <friendly-name>" row prepended when the current value isn't in `predefined_options()`; clicking re-opens the modal pre-populated. No silent fallback to index 0 (which would have caused data loss).
- **Open question 4 resolved:** friendly display name primary; protocol-level `KeyCode` in hover tooltip only. Same string on dropdown row and modal.

/oz-review

lonexreb:
Round-4 corrections (commit ccbec29):
- Modal body now shows friendly name only with KeyCode in hover tooltip — matches the round-3 resolution of Open Question 4. Removed the contradictory `Selected: F19 (KeyCode::F19)` inline form.
- Open Question 1 closed: committed to manual Serialize/Deserialize impl per round-2 Settings serialization section.
- VoiceInputTomlValue wrapper now explicitly an internal implementation detail. All external accessors (set_value, value(), tests, dropdown reads, dispatch) continue to see the unwrapped VoiceInputToggleKey. After post-init the field is always Valid(...).
/oz-review

Issue #9889, opened 2026-05-02
Show per-message timestamps in Agent Mode for prompts and responses


ryanentwarp:
## Summary
Add per-message timestamps to Agent Mode so users can see when each prompt was sent and when the agent responded.

## Problem
In Agent Mode, there is currently no way to see when a prompt was submitted or when the agent finished responding. For users who run long agent tasks or reference past conversations, the lack of timestamps makes it difficult to understand duration, sequence, and recency of interactions.

## Reproduction steps or desired workflow
1. Open Agent Mode in Warp.
2. Send one or more prompts to the agent.
3. Desired: each user prompt and each agent response shows a timestamp (e.g., "5:47 AM" or a relative time like "2 min ago") so the user can tell when it was sent and how long the agent took to respond.

## Artifacts
None attached.

## Warp version
0.2026.04.25.08.29.00

## Operating system
macOS 26.4

PR #10128, opened 2026-05-05
spec: per-message timestamps in Agent Mode (#9889)

lonexreb:
## Summary

Spec for issue #9889 — Agent Mode shows the conversation as a sequence of prompts and responses with no temporal information. Users cannot tell when a prompt was submitted, how long the agent took, or how recent any given exchange is — particularly relevant for long-running tasks, comparing two attempts, or reviewing what an autonomous agent did overnight.

## Investigation

The data already exists at the model layer:

- `AIAgentExchange` carries `start_ts: Option<DateTime<Local>>` and `completed_ts: Option<DateTime<Local>>` ([`app/src/ai/agent/conversation.rs:3218-3219`](app/src/ai/agent/conversation.rs)).
- `Conversation::start_time_from_exchange_messages` derives the start time from the user input's `AIAgentContext::CurrentTime` ([conversation.rs:556](app/src/ai/agent/conversation.rs)).
- The corresponding finish-time helper at lines 538-553 derives completion from the latest message timestamp on the exchange.
- Conversation restoration replays the underlying timestamps, so restored conversations get the same derivation for free.

**Only the view layer omits this information.**

## What's in the spec

**`product.md`** — 8 testable behavior invariants (B1–B8), 8 acceptance criteria (A1–A8), explicit non-goals (per-token timestamps, server/client time reconciliation, CPU/credit accounting), and 5 risks with concrete TECH decisions to make.

**`tech.md`** — picks the data-derivation site (pure helper at the view layer, no new model state), the UI integration site (existing bubble metadata row, not a new gutter), the timer strategy (single shared 1Hz timer gated to visible in-progress exchanges), and lays out 5+6 unit tests + 8 integration tests.

## Architectural choices

- **Pure derivation helper at the view layer**, reusing the existing model-level helpers. No new state on the model, no view-layer caching of derived times.
- **Inline in the existing bubble metadata row** (recommended over a new gutter) — leverages existing density and avoids a new layout slot.
- **Single shared 1Hz `TimestampTickService`** for all in-progress duration counters, gated to "panel visible AND in-progress exchange present." A long conversation with 50 visible exchanges costs one timer, not 50.
- **Format auto-promotion via 30s tick**: "just now" → "Xm ago" → "HH:MM" → "Day HH:MM" → "YYYY-MM-DD HH:MM". The 30s rule is applied internally to a 1Hz timer to avoid layout churn.
- **New setting `agents.show_message_timestamps`** (default `true`, `SyncToCloud::Always`). When `false` the widgets are not added to the layout — zero off-state cost (A5 invariant).
- **Missing-timestamp fallback** to em-dash with a once-per-exchange `log::warn!` so production data gaps are visible without flooding the log.
- **Cancellation handling**: "cancelled at HH:MM • Xs" instead of "HH:MM • Xs."

## Test plan

- [ ] Data-derivation unit T1–T5 (`exchange_times.rs`)
- [ ] Format unit T6–T11 (`timestamp_widgets.rs`)
- [ ] Integration IT1–IT8 including a snapshot test for the off-state, a clock-advance test for format auto-promotion, and a yaml-restore test for historical timestamps
- [ ] Manual: long-running prompt; verify "running for Xs" ticks once per second and freezes on completion
- [ ] Manual: hover any timestamp; verify ISO 8601 tooltip
- [ ] Manual: toggle setting off; verify pixel-equivalent layout to current build

## Open questions for maintainer review

1. Default: `true` (opt-out, recommended) or `false` (opt-in)?
2. Inline-bubble (recommended) vs. gutter rendering. Defers to design — issue is `needs-mocks`.
3. `TimestampTickService` lifetime: per-conversation (if conversations can render in multiple places simultaneously) or singleton (if not)?
4. Tooltip format: ISO 8601 (unambiguous) or localized "Tuesday, May 4, 2026 at 3:47:23 PM CDT" (friendlier)?

Closes (spec-only) #9889 — implementation PR will follow once spec direction is confirmed.

### DIFF
--- a/specs/GH9889/product.md
+++ b/specs/GH9889/product.md
@@ -233,17 +233,17 @@ A8. The display format auto-promotes from "just now" → "Xm ago" →
      labels (B2's max 30s update rule), OR
    - Render-on-scroll only.
    Recommendation: the 30s tick, gated to "while panel is visible
-   AND the oldest visible timestamp is in a relative-format
-   bucket" so we don't waste cycles on conversations whose
-   timestamps are all absolute.
+   AND any visible timestamp is in a relative-format bucket" so we
+   don't waste cycles on conversations whose timestamps are all
+   absolute.
 
 4. **Setting default.** Default `true` (opt-out) per B5. Confirm
    with maintainers before implementation; this is the most
    reversible decision.
 
-5. **Telemetry.** Add a one-time `setting_changed` event when the
-   user disables the new setting, so the team can see the opt-out
-   rate. No timestamp-display telemetry beyond that.
+5. **Telemetry.** Add only the existing `setting_changed` event when
+   the user toggles the new setting, so the team can see the opt-out
+   rate. No timestamp-display telemetry is added.
 
    **Privacy guardrails (security review #10128):**
    - The `setting_changed` payload is `{ setting: "agents.show_message_timestamps",
--- a/specs/GH9889/product.md
+++ b/specs/GH9889/product.md
@@ -40,6 +40,9 @@ across long-running and restored conversations.
 - **Editing or backdating timestamps.** Read-only display.
 - **Time-zone selection.** Use `Local` (matches existing
   `Local.timestamp_opt(...)` derivation in conversation.rs:549).
+  *Time-zone* selection is out of scope. **12h-vs-24h preference**
+  is in scope (see B2) and is read from the OS — that is not a
+  user-facing setting in V1, just an OS preference read at startup.
 - **Showing timestamps in the conversation export / yaml.** Out of
   scope — the export already preserves the underlying timestamps,
   this spec is about the in-app view.
@@ -97,11 +100,21 @@ support tickets, regression diffs).
 
 ### B5 — Hidden by default behind a setting; default ON
 
+> **Correction (re-review #10128):** the previous draft cited a
+> non-existent `SyncToCloud::Always`. The actual variants in
+> `app/src/settings/cloud_preferences.rs` are `Never`,
+> `Globally(RespectUserSyncSetting)`, and `PerPlatform(...)`. The
+> right value here is
+> `SyncToCloud::Globally(RespectUserSyncSetting::Yes)` — sync
+> across the user's devices, but respect their global cloud-sync
+> opt-out.
+
 A new setting `agents.show_message_timestamps` (boolean, default
-`true`, `SyncToCloud::Always`) gates the entire feature. When
-`false`, the view layer is identical to today (no timestamp glyphs,
-no tooltip hooks, no extra layout). Existing keyboard shortcuts and
-focus handlers are unchanged in either state.
+`true`,
+`SyncToCloud::Globally(RespectUserSyncSetting::Yes)`) gates the
+entire feature. When `false`, the view layer is identical to today
+(no timestamp glyphs, no tooltip hooks, no extra layout). Existing
+keyboard shortcuts and focus handlers are unchanged in either state.
 
 The default is `true` because the feature is opt-out (the issue is
 about a missing affordance) but power users who want a denser view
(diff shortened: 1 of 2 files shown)


lonexreb:
Pushed corrections addressing all 4 bot review concerns:

- **Codebase facts:** corrected `AIAgentExchange` location to [`app/src/ai/agent/mod.rs:2835`](app/src/ai/agent/mod.rs) with the actual fields `start_time: DateTime<Local>` and `finish_time: Option<DateTime<Local>>` (not the `start_ts`/`completed_ts` I'd written). Updated all helper references to the right line numbers.
- **UI integration site:** moved from `agent_view/agent_message_bar.rs` (which renders the bottom input bar) to the actual prompt/response renderers — `ai/blocklist/block/view_impl/query.rs::render_query` for the user prompt and `view_impl/output.rs` for the agent response. New helper module relocated to `view_impl/` alongside the consumers.
- **Locale-aware formatting:** replaced the hard-coded English 12-hour format with a `ClockFormat` struct sourced from the OS at startup (NSLocale on mac, LOCALE_ITIME on Windows, LC_TIME/LANG on Linux). Tests T8/T9 are duplicated as T8b/T9b for non-English/24h locales.
- **Telemetry privacy:** added explicit guardrails to product.md §Risk 5 — payload shape, no IDs/timestamps/message content, respects global telemetry opt-out, no new server-side channels.

Each correction includes a "Correction (review #10128)" callout in the spec so future reviewers can cross-reference. Ready for re-review.

/oz-review

lonexreb:
Round-2 corrections pushed (commit 523d7f2):

- **Prompt path data plumbing:** I was wrong — `render_query` in `view_impl/query.rs` takes primitive params via `Props` (`query: &str`, `user_display_name`, etc.), not an `AIAgentExchange`. Spec now plans an explicit `submitted_at: DateTime<Local>` field on `Props` plus the call-site builder. One field on each of two structs; no new layer.
- **Locale formatter:** dropped `chrono::Locale` / `format_localized` (the workspace `chrono = { version = "0.4.38", features = ["serde"] }` doesn't enable `unstable-locales`). V1 keeps English weekday names from default `%a` and reads only the 12h-vs-24h preference from the OS — works without the feature flag. Localized weekdays called out as an explicit V2 follow-up with two implementation options.
- **Missing-timestamp fallback:** the model's recompute call sites at `conversation.rs:1777/1895/1962` do `unwrap_or_else(Local::now)`, so finished exchanges always have `Some(finish_time)`. The dash glyph is now scoped to defensive guards (state-machine-bug only) with one-time warn logs. The happy path uses `ProgressDurationLabel` for in-progress and `TimestampLabel` for finished — no dash in normal operation.

/oz-review

lonexreb:
Round-3 corrections pushed (commit cdbd9c0):

- **B7/A7 dash conflict:** the model already synthesizes `Local::now()` at `conversation.rs:1777/1895/1962`, so finished restored exchanges always have `Some(finish_time)`. Promising the dash there contradicted the model. Resolved: dash is now defensive-guard-only (state-machine-bug invariant). In normal operation it never renders.
- **Non-existent `SyncToCloud::Always`:** real variants are `Never`, `Globally(RespectUserSyncSetting)`, `PerPlatform(...)`. Updated to `SyncToCloud::Globally(RespectUserSyncSetting::Yes)` — same value as every UI preference in `code.rs:19/29/40/48`.
- **12h/24h non-goal contradiction:** clarified time-zone selection is out, 12h-vs-24h preference is in (read from OS, not a user setting in V1).

/oz-review

lonexreb:
Round-4 corrections (commit 51fba7c):
- `is_visible` hook identified concretely: `BlockListViewModel::is_panel_visible` / `terminal_view().is_active` (same checks the agent view's existing 1Hz status refresh uses). No new lifecycle API.
- Cancellation rendering: three structurally distinct shapes (in-progress / cancelled / finished) instead of swapping strings in one slot. Cancelled shape is one composite label that owns both time + duration so they cannot duplicate.
- IT7 corrected: tests force-strip finish_time on a Finished output_status (since the model normally synthesizes Local::now()). Defensive-guard path only.
- Non-goals updated: time-zone out, 24h/12h preference IN scope per B2.
/oz-review

PR #10130, opened 2026-05-05
fix(bootstrap): detect Xcode installs with non-default app names on macOS

### LINKED ISSUE #10125: `script/macos/bootstrap` fails when Xcode is installed via Xcodes app or under a non-default name
### Summary

`script/macos/bootstrap` hardcodes a check for `/Applications/Xcode.app` and a `sudo xcode-select --switch /Applications/Xcode.app/Contents/Developer`. This breaks for users who install Xcode via the [Xcodes app](https://github.com/XcodesOrg/XcodesApp), where installations are named like `Xcode-26.3.0.app` or `Xcode-beta.app`. The script exits with "Please install Xcode from the App Store before continuing." even though a fully functional Xcode is installed.

### Steps to reproduce

1. Install Xcode via the Xcodes app (so it lands at `/Applications/Xcode-26.3.0.app` rather than `/Applications/Xcode.app`).
2. Ensure `xcode-select -p` points at that install (e.g. `/Applications/Xcode-26.3.0.app/Contents/Developer`).
3. Run `./script/macos/bootstrap`.

### Expected behavior

The script detects the existing Xcode installation (regardless of its `.app` filename) and proceeds.

### Actual behavior

The script fails immediately at the `[ -d "/Applications/Xcode.app" ]` check and prints "Please install Xcode from the App Store before continuing."

### Environment

- macOS, Xcode installed via Xcodes app.
- Affected file: `script/macos/bootstrap` (lines 8–13).

### Notes

A complete fix should: (a) accept any `/Applications/Xcode*.app` name, (b) detect when only Command Line Tools is selected and either switch to a discovered full Xcode or fail with the intended error, and (c) preserve the existing auto-`xcode-select --switch` behavior so users with a full Xcode installed but CLT currently selected don't have to fix selection by hand.

SagarSDagdu:
## Description

`script/macos/bootstrap` hardcoded `/Applications/Xcode.app` for both the existence check and the `xcode-select --switch` call, rejecting valid Xcode installs managed by the Xcodes app (e.g. `Xcode-26.3.0.app`, `Xcode-beta.app`) even when `xcode-select -p` already pointed at one.

This change:

- Respects an existing full Xcode developer dir already selected via `xcode-select -p`.
- Falls back to scanning `/Applications/Xcode*.app` when nothing is selected or only Command Line Tools is selected. Each candidate is validated by checking for `Contents/Developer/usr/bin/xcodebuild`, so the Xcodes launcher app (`/Applications/Xcodes.app`) is correctly skipped.
- Preserves the auto-`xcode-select --switch` so users with a full Xcode installed but CLT currently selected don't have to fix selection by hand.

Fixes #10125

## Testing

Validated the detection block in isolation against three states on a machine with `/Applications/Xcode-26.3.0.app` (Xcodes-managed) and `/Applications/Xcodes.app` present:

1. Xcode already selected via `xcode-select` — fast path is taken, no re-switch.
2. Command Line Tools selected — fallback finds `Xcode-26.3.0.app` and issues the `xcode-select --switch`.
3. Nothing selected (`xcode-select -p` errors) — same fallback succeeds.

In all cases the `Xcodes.app` launcher is skipped because it lacks `Contents/Developer/usr/bin/xcodebuild`.


SagarSDagdu:
Addressed in 1bc11b0: the fallback now prefers `/Applications/Xcode.app` explicitly when present and only scans `/Applications/Xcode*.app` when it isn't, so stable+beta/versioned setups stay on the stable install.

/oz-review

script/macos/bootstrap
```diff
@@ -7,12 +7,27 @@ set -e
...
+            if [ -x "$candidate/Contents/Developer/usr/bin/xcodebuild" ]; then
+                XCODE_APP="$candidate"
+                break
+            fi
+        done
+    fi
+    if [ -z "$XCODE_APP" ]; then
+        echo "Please install Xcode before continuing."
+        exit 1
+    fi
```
captainsafia, line 28: Small nit if you're able to get to it: can we echo the candidate XCode instance that was picked for visibility here?

captainsafia (approved):
PR looks solid overall! I left one small comment inline if you're able to push a quick change to it. I'll circle back to merge it later.

script/macos/bootstrap
SagarSDagdu: Done!

Issue #8410, opened 2026-01-10
Add support for git diff in WSL


juanfradb:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Just make it possible to view diff in WSL

### Operating system (OS)

Windows

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

WakaTaira:
#13793 (merged today) should resolve the core ask here: the diff / Code Review panel's git invocations all run through the function that now routes to `wsl.exe` when the working directory is a `\\wsl$` UNC path — the PR's after-screenshot is exactly this scenario, diffs rendering for a WSL repository. Worth re-testing once a release containing d464735 ships.

One caveat: the `gh`-based GitHub integration (PR metadata in Code Review) is not routed yet, so that part may still fail for WSL repos and would be follow-up material.

Issue #7909, opened 2025-10-27
Git repository detection not working on WSL


httpdss:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agents/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-billing/sending-us-feedback#gathering-warp-logs)

### Describe the bug

@ command and all things related to git integration dont seem to work correctly. folder is not detected as git repository and it has everything to be a git repo on github.com: https://github.com/httpdss/struct
so im not understunding why this doesnt work. 

### To reproduce

prerequisites: you are using WSL on windows machine. (in my case, ubuntu with zsh)

1. go to a git repo folder
2. click the "git changes button" at the topleft 

### Expected behavior

the expected behaviour is that warp detects this as a git repo

### Screenshots, videos, and logs

<img width="1898" height="624" alt="Image" src="https://github.com/user-attachments/assets/718b0c1e-6b32-4105-98e6-014f33e94410" />

### Operating system (OS)

Windows

### Operating system and version

windows 11 with wsl (ubuntu 24.04)

### Shell Version

zsh 5.9 (x86_64-ubuntu-linux-gnu)

### Current Warp version

v0.2025.10.22.08.13.stable_02

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

Issue #6645, opened 2025-06-24
Git Repositories not detected for @ attachment in WSL


TheOneTrueAnt:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/support-and-billing/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When working in a WSL tab in warp, git repositories do not appear to be completely detected, the branch information shows up in the prompt but the `@` menu does not appear and the button remains disabled. This also appears to be a problem in Git bash (`bash.exe`) tabs. The feature works fine in powershell however.

![Image](https://github.com/user-attachments/assets/e9666cd4-d168-4c16-8ba9-a3258144da97)

### To reproduce

1. Open a new tab in warp, in WSL
2. navigate to a git repository
3. hit `@` and no menu appears / `@` button remains disabled with "only available in git repository" tooltip

### Expected behavior

pop up menu of files to select should appear and button should become activated

### Operating system (OS)

Windows

### Operating system and version

11

### Shell Version

GNU bash, version 5.2.21(1)-release (on Ubuntu 24.04 WSL)

### Current Warp version

v0.2025.06.20.22.47.stable_05

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

PR #10137, opened 2026-05-05
wsl: prefer Linux-side git/gh over Windows interop binaries

JonRC:
## Description

When Warp runs as a Linux ELF inside WSL on Windows, WSL's default `appendWindowsPath = true` (in `/etc/wsl.conf`) puts directories like `/mnt/c/Program Files/Git/cmd/` on `PATH`. That makes a bare `Command::new("git")` (or `"gh"`) potentially resolve to the Windows `.exe` through WSL interop. The interop call is dramatically slower (cross-OS), can mishandle Linux paths, and breaks Linux-side hooks (e.g. LFS). Affected features include the GitDiffStats context chip, code-review per-file diffs, branch detection, file-search status, AI skill resolution, and passive suggestions.

This PR adds a small `warp_util::wsl` module with three cached helpers:

- `is_wsl()` — the existing `/proc/sys/fs/binfmt_misc/WSLInterop` probe, deduplicated from two prior copies in `crates/warpui/src/platform/linux/mod.rs` and `app/src/crash_reporting/linux.rs`.
- `git_binary()` / `gh_binary()` — return the literal program name on non-WSL, and on WSL walk `PATH` while skipping any entry under `/mnt/`, returning the first Linux-side executable match. If nothing is found on a filtered PATH the helper logs once and falls back to the literal name (so behavior is no worse than before).

Every Warp-internal `Command::new("git"|"gh")` site is rewired through the helpers, including the test/debug helpers under `integration_testing/`, the snapshot tests, and `crates/integration/src/test/code_review.rs`. The `/mnt/*` filter mirrors the existing precedent at `app/src/terminal/model/session/command_executor/wsl_command_executor.rs:60-65` used for `compgen`.

User-typed shell git invocations are unchanged — those remain governed by the user's shell PATH.

## Linked Issue

Tracks [#8410](https://github.com/warpdotdev/warp/issues/8410). The issue is currently labeled `enhancement` without `ready-to-spec` / `ready-to-implement`. The change here is a behavior fix for an existing feature rather than new functionality, so I'd appreciate maintainers / Oz triaging it as a bug fix; happy to file a separate bug report and re-link if a fresh issue is preferred.

### Other open WSL issues likely fixed by this PR

These reports describe symptoms of the same root cause (Warp-internal `git` falling through to `git.exe` via WSL interop). Maintainers may want to verify and close them once this lands:

- [#7909 — Git repository detection not working on WSL](https://github.com/warpdotdev/warp/issues/7909) (`bug`, `ready-to-implement`): the "git changes" button fails to surface diffs because the underlying `git diff` invocation goes through `git.exe`.
- [#6645 — Git Repositories not detected for `@` attachment in WSL](https://github.com/warpdotdev/warp/issues/6645) (`bug`, `ready-to-implement`): the `@` menu stays disabled because `git ls-files` (now rewired through `git_binary()` in `app/src/ai/agent_sdk/driver/snapshot.rs` and `app/src/ai/skills/resolve_skill_spec.rs`) fails under WSL interop.

- [x] No UI changes; screenshots/videos N/A.

## Screenshots / Videos

Not applicable — change is entirely in subprocess invocation plumbing.

## Testing

- New unit tests in `crates/warp_util/src/wsl_tests.rs` cover the pure resolver `resolve_binary_in_wsl_safe_path` across 8 cases: WSL prefers Linux-side dirs over `/mnt/...`; non-WSL passes through; only-`/mnt` falls back to None; user-local bin is found via PATH walk; symlinks resolve; non-executables are skipped; `PATH=None` is safe; non-UTF-8 PATH bytes don't panic.
- Locally verified: `cargo fmt --all -- --check`, `cargo check` and `cargo clippy --all-targets --tests -- -D warnings` for `warp_util`, `warpui`, `warp`, and `integration`; full `cargo test -p warp_util` (66 tests + 3 doc tests, all green). Full-workspace clippy/nextest weren't run on my machine because of an unrelated missing system dep (`libasound2-dev`).
- Manual end-to-end check: in WSL, build and run inside a git repo with both `/usr/bin/git` and Windows git on PATH; the GitDiffStats chip and code-review diff view continue to work, and the `[GIT OPERATION]` debug logs no longer warn.

## Agent Mode


JonRC:
Thanks for the review! Addressed in fce8b74:

`resolve_or_warn` now short-circuits when `is_wsl()` is false and returns the bare program name, so `Command::new` performs its normal PATH lookup at spawn time — restoring the per-call `cmd.env("PATH", …)` override behavior that callers like `run_git_command_with_env` (LFS hooks) and `run_gh_command` (Homebrew `gh` on macOS) depend on. Module docs updated accordingly. The pure resolver and existing tests are unchanged; the WSL-only path is exercised explicitly with `is_wsl=true`.

/oz-review

JonRC:
@cla-bot check

JonRC:
@cla-bot check

crates/warpui/src/platform/linux/mod.rs
```diff
@@ -56,11 +56,7 @@ pub fn user_windowing_system() -> WindowingSystem {
 }
 
 pub fn is_wsl() -> bool {
-    use std::sync::OnceLock;
-    static IS_WSL: OnceLock<bool> = OnceLock::new();
-    IS_WSL
-        .get_or_init(|| std::path::Path::new("/proc/sys/fs/binfmt_misc/WSLInterop").exists())
-        .to_owned()
+    warp_util::wsl::is_wsl()
```
vorporeal: this dependency edge is in the wrong direction - the UI framework shouldn't be dependent on "app code".  (in general, `warp_`-prefixed crates should all sit above the UI framework.  also, as time goes on, i'm not loving the `warp_util` grab-bag crate, so i'm generally trying to avoid putting new things in there.)

given `warpui` already depends on `command` (on non-macOS platforms), if you agree that we should move the git/gh resolution logic into the `command` crate, i'd say we also move the `is_wsl()` function in there.  (it's not my _favorite_ home for that function, but i also don't necessarily feel like making a `wsl` crate just to hold `is_wsl()` is necessary.)

vorporeal (commented):
thanks for the PR!  some thoughts:

at a high level, i'm concerned that "leaking" this detail to invokers of `Command` will make it too easy for someone to fail to use this at a future call site.

if we want to _always_ be doing this additional logic when using `Command` and `BlockingCommand`, should this be an implementation detail of those wrappers?  i.e.: we still call `Command::new("git")` but internally, it uses your helper functions to perform the translation, when needed.

JonRC:
Thanks @vorporeal! Both points make a lot of sense — the "every new call site has to remember the helper" risk is real, and the dep direction critique on `warpui → warp_util` is fair. I'll move the WSL detection and the git/gh resolution into the `command` crate so `Command::new("git")` / `BlockingCommand::new("git")` transparently do the right thing, drop the new module from `warp_util` entirely, and have `warpui::platform::linux::is_wsl()` delegate into `command`. Will push and ping you once it's done.

JonRC:
@vorporeal — addressed in a811077:

- The WSL detection + git/gh resolution now live in `crates/command/src/wsl.rs`. `command::r#async::Command::new` / `command::blocking::Command::new` (plus `new_with_session` / `new_with_process_group`) call an internal `translate_program_for_spawn` so call sites continue to write `Command::new("git")` and the substitution happens automatically — no per-callsite remembering.
- Only bare-name programs in `KNOWN_NAMES = ["git", "gh"]` are touched; path-qualified or unknown programs pass through unchanged, so any caller that already specified `/usr/bin/git` is preserved.
- The new `warp_util::wsl` module is gone — every previous `warp_util::wsl::git_binary()` / `gh_binary()` callsite reverted to bare `Command::new("git" | "gh")`. The `warpui → warp_util` and `integration → warp_util` Cargo edges are gone with it.
- `warpui::platform::linux::is_wsl()` and `app::crash_reporting::linux` now delegate to `command::wsl::is_wsl()`. Dep direction is `warpui → command` (which already existed on non-macOS) and `app → command` — no `warp_*` crate depending on `warp_util` for this.
- 11 unit tests live alongside the new module: 8 cover the pure resolver, 3 cover the bare-name detector (recognizing `git`/`gh`, skipping paths, skipping unknowns).

Local checks (touched crates): `cargo fmt --check`, `cargo clippy --all-targets --tests -- -D warnings`, and `cargo test` for `command`, `warp_util`, `warpui`, `warp`, and `integration` all green.

vorporeal (approved):

JonRC:
@vorporeal could you approve the CI runs for 39f1d1a when you get a moment? They're sitting at `action_required` so the Windows clippy / wasm jobs that were failing previously haven't been able to re-run yet. The fix in 39f1d1a cfg-gates the unix-only test helpers (`std::fs`, `join`) so Windows clippy stops failing `-D warnings` on unused imports/dead code; verified locally.

PR #10140, opened 2026-05-05
Fix `tail` path in script/macos/run --open_with_launchd for OSS channel

sebryu:
## Summary

`./script/run --open_with_launchd` (macOS, OSS channel) ends with:

```
tail: /Users/.../Library/Logs/warp_local.log: No such file or directory
```

The script hardcodes `tail -f ~/Library/Logs/warp_local.log`, but the OSS binary writes to `~/Library/Logs/warp-oss.log` (see `app/src/bin/oss.rs` — `logfile_name: "warp-oss.log"`). External contributors who can't install `warp-channel-config` always build the OSS channel and so always hit this.

This branches on `$WARP_CHANNEL` so:
- **OSS** tails `~/Library/Logs/warp-oss.log` (verified path).
- **Local** keeps tailing `~/Library/Logs/warp_local.log` (unchanged — local channel `logfile_name` lives in the closed `warp-channel-config` repo, so I didn't touch it).

Also swaps `tail -f` → `tail -F` so the tail survives the brief window before launchd has created the log file.

## Test plan
- [x] `./script/run --open_with_launchd` on OSS channel → app launches and `warp-oss.log` is tailed correctly.
- [ ] Internal: `./script/run --open_with_launchd` on local channel → unchanged behavior, still tails `warp_local.log`.

### DIFF
--- a/script/macos/run
+++ b/script/macos/run
@@ -134,7 +134,12 @@ if [ "$DONT_OPEN" = false ] ; then
     if  [ "$OPEN_WITH_LAUNCHD" = true ]; then
         echo "Launching with MacOS application launcher"
         PATH="" /usr/bin/open "./$WARP_APP_PATH"
-        tail -f ~/Library/Logs/warp_local.log
+        if [ "$WARP_CHANNEL" = "local" ]; then
+            LOG_FILE=~/Library/Logs/warp_local.log
+        else
+            LOG_FILE=~/Library/Logs/warp-oss.log
+        fi
+        tail -F "$LOG_FILE"
     else
         echo "Opening app at ./$WARP_APP_PATH/Contents/MacOS/$WARP_BIN_NAME"
         "./$WARP_APP_PATH/Contents/MacOS/$WARP_BIN_NAME" "${WARP_ARGS[@]}"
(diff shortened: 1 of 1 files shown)


script/macos/run
```diff
@@ -134,7 +134,12 @@ if [ "$DONT_OPEN" = false ] ; then
     if  [ "$OPEN_WITH_LAUNCHD" = true ]; then
         echo "Launching with MacOS application launcher"
         PATH="" /usr/bin/open "./$WARP_APP_PATH"
-        tail -f ~/Library/Logs/warp_local.log
+        if [ "$WARP_CHANNEL" = "local" ]; then
+            LOG_FILE=~/Library/Logs/warp_local.log
+        else
```
peicodes, line 141: I would prefer we `elif` here and be specific about the OSS channel, and in the else case log a warning that the channel is unrecognized

peicodes (changes requested):

script/macos/run
sebryu: Done in 679b8cb — added an explicit `elif` for `oss` and a warning on unrecognized channels.

peicodes (approved):

peicodes:
Thanks the the changes! I'll merge once CI passes

Issue #10115, opened 2026-05-05
Independent zoom level per session (Cmd +/-/0 affects focused window only)


Eridanus117:
### Pre-submit Checks

- [x] I have searched Warp feature requests and identified several related but non-matching issues — see *Relation to existing issues* below.
- [x] I have searched Warp docs and this is not configurable today.

### Describe the solution you'd like?

`Cmd +` / `Cmd -` / `Cmd 0` should adjust the **focused OS window's** zoom level only, leaving other Warp windows at their existing zoom. All tabs and splits inside the same window share that window's zoom level (a window-scoped baseline, not per-pane override).

In-memory only is acceptable (zoom resets to global default on app restart) — same trade-off PR #9705 makes for per-pane font size.

### Is your feature request related to a problem? Please describe.

I run multiple OS-level Warp windows across monitors of different sizes and viewing distances (e.g. a small reference window on a side monitor, a primary window on the main display). I want each window at its own readable zoom. Today `Cmd +/-` is global — adjusting one window changes all of them, which is the opposite of what I want.

VS Code, iTerm2, Chrome, and most macOS apps with multi-window support behave this way already.

### Relation to existing issues (please review before triage)

I went through the zoom/font-size cluster carefully. None of the existing issues cleanly match a *per-OS-window zoom baseline*:

- **#1394** "Different font size per window/split/tab" — closest match in title, but the scope bundles three independent design decisions (window vs split vs tab) and has been stuck at `triaged` for ~4 years. The window-level slice alone is the smallest meaningful step.
- **#6739** "Allow for per window font size" / **PR #9705** — title says "window" but the body describes splitting into "4 windows two on top and two on bottom," i.e. panes, not OS windows. PR #9705 correctly implements per-*pane* in-memory override. That's an orthogonal feature (pane override on top of a baseline), not a substitute for a window-scoped baseline.
- **#2573** "Isolate zoom levels per pane / tab" (`ai:ready-to-spec`) — covers pane/tab isolation but does not include the window dimension.
- **#1439** per-pane font size — different layer.

### Architecture note

Currently `ZoomLevel` is a global `SettingsBackedValue`. This proposal moves the *owner* from global to `WindowState` (with global as the default for new windows). PR #9705's per-pane override would naturally layer on top: `effective_zoom = pane_override ?? window_zoom ?? global_default`.

### Question for @oz-agent

Given the above mapping, is the cleanest path:

[...]

Issue #1394, opened 2022-05-23
Different font size per window/split/tab


divan:
### Describe the solution you'd like?

```text
Currently font size adjustement (Cmd-+/-) seems to be synchronized for all Warp windows/splits/tabs. But often it's desirable to have separate font sizes per window (i.e. one window shows logs or verbose output and would benefit from smaller font size).

It would be nice to adjust font sizes for each window separately.
```

elviskahoro:
Great call out. Thanks for reporting!

dannyneira:
Hey Folks, Please add a :+1: to the original post at the top to signal that you want this feature, and then hit the subscribe button if you'd like to be notified.

wmendes-ionos:
How many likes does this post needs in order to have this feature kicked-off ? It's been more than one year of just voting... ?

yossi-trullion:
This is a deal breaker

rickgladwin:
Would really like to see this implemented. It's a more important and fundamental feature to me as a user than whatever AI integrations have been getting developer time over the last few months, to be honest. 

Different command line outputs work best with different line lengths (docker information, for example, requires long lines and is hard to read when the lines wrap), other output can stay as big and readable as possible. Setting font size per window is the easiest way to achieve this, and not having the feature means we have to repeatedly adjust the font size up and down when _looking at a different session_. It's a workflow hindrance.

Font size per window is the default behaviour in iTerm, and it's very convenient. Not sure why font sizes haven't been implemented this way in Warp.

umwwwelt:
Any news? Two years later…

divan:
I bought a second computer and second screen, so now I can have two Warp windows with different font sizes!

Eridanus117:
Drive-by note in case it helps unstick this — the title bundles three independent design decisions (window vs split vs tab), and that may be why it's been at `triaged` since 2022. The minimal actionable slice is **window only**, since the variants for split/tab/pane are already separately tracked:

- per-pane: #1439, #6739 + PR #9705 (in-flight, in-memory `HashMap<PaneId, f32>`)
- per pane/tab zoom isolation: #2573 (`ai:ready-to-spec`)
- per OS-window baseline: filed as #10115 with an architecture note proposing a layered model (`effective_zoom = pane_override ?? window_zoom ?? global_default`)

@oz-agent re-reviewed the relationship at #10115 and recommended consolidating the OS-window baseline here under #1394, keeping the pane/tab variants in their own trackers.

[...]

PR #10145, opened 2026-05-05
Per-OS-window zoom override (refs #1394)

Eridanus117:
Refs #1394.

## Summary

`Cmd+=` / `Cmd+-` / `Cmd+0` in one Warp OS window only affects that window. Other windows keep their current zoom, which falls back to the app-wide default (`WindowSettings.zoom_level`) when the window has no override. The Settings UI "Default Zoom" dropdown still drives the app-wide default; windows without a per-window override follow it.

Conceptually mirrors VS Code's `window.zoomLevel` (global) + per-window override via `window.zoomPerWindow`, and Kitty's `change_font_size` (defaults to current OS window).

Demo: two windows, left at default, right after Cmd+= x3

## Implementation shape

Three commits on top of `a057a10`, `+191 / -28` total across 9 files:

| commit | scope |
|---|---|
| `c3a6997` | `Window::zoom_factor_override: Option<ZoomFactor>` + new AppContext methods `window_zoom_factor` / `set_window_zoom_factor` / `reset_window_zoom_factor`; renderer + Cmd-keys route through per-window override |
| `6d38c6b` | Cover layout/chrome zoom readers: native macOS titlebar height, traffic-light reservation, tab-bar left padding, theme chooser margin; explicit `update_titlebar_height` call in adjust/reset because per-window path no longer fires `WindowSettingsChangedEvent::ZoomLevel`; tightened clamp; unit tests |
| `16cff28` | Align `set_zoom_factor` clamp `[0.5, 4.0]` → `[0.5, 3.5]` so neither setter can leave `adjust_zoom`'s VALUES lookup outside the discrete table |

Resolution model:

```
window_effective_zoom = window.zoom_factor_override ?? AppContext.zoom_factor
```

Pane-level overrides (PR #9705 implementing per-pane font for #6739) are on a separate axis and untouched:

```
pane_effective_font = pane.font_size_override ?? font_from(window_effective_zoom)
```

## Why it's app-wide today (current master)

| File:line | Why |
|---|---|
| `crates/warpui_core/src/core/app.rs:1188` `set_zoom_factor` | Doc-comment explicit: _"All views in every window are invalidated when this is invoked."_ Writes AppContext field, calls `invalidate_all_views()`. |
| `crates/warpui_core/src/presenter.rs:346/347/520` | Renderer reads `ctx.zoom_factor()` from AppContext (every window same value). |
| `app/src/window_settings.rs:73` `zoom_level` | Field lives inside `WindowSettings`, but `WindowSettings` is an app-wide singleton (`WindowSettings::handle(ctx)` / `as_ref(ctx)` accessors take no `WindowId`). |
| `app/src/workspace/view.rs:15545-15580` | `increase_zoom`/`decrease_zoom`/`reset_zoom`/`adjust_zoom` write through that singleton. |
| `app/src/settings_view/appearance_page.rs:925` | Settings subscriber relays into `set_zoom_factor`, closing the global path. |

Four additional zoom-aware sites in chrome/layout read the same singleton — any per-window scheme that only touches the renderer would leave native macOS titlebar / traffic-light spacing misaligned, so the patch covers those too:

- `app/src/workspace/view.rs:11990` `update_titlebar_height`
- `app/src/workspace/view.rs:17494` traffic-light reservation
- `app/src/workspace/view.rs:17516` `compute_tab_bar_left_padding`
- `app/src/themes/theme_chooser.rs:593` traffic-light header margin

## Behavioral notes

- Settings UI "Default Zoom" changes the app-wide default; existing windows without an override follow the new default; windows with an override are not reset (matches VS Code's `window.zoomLevel` semantics — reset is via `Cmd+0` which clears the override).
- Per-window override is in-memory; closing a window discards its override, app restart restores all windows to the saved global default. Persistence is a separate follow-up.

## Out of scope (follow-ups)

- **Persistence** of per-window overrides across (a) window close→reopen, (b) workspace reload, (c) app restart — needs stable per-window identity beyond a single process lifetime.
- **Status bar indicator** when a window's zoom differs from the default.
- **`window.zoomPerWindow: false` toggle** for users who prefer global-sync legacy behavior.
- **DPI-aware auto-scale** across monitors with different scale factors.

## Test plan

- [x] `cargo build --bin warp-oss --features gui` (build clean)
- [x] `cargo test --package warpui_core zoom_tests` (unit tests pass: per-window isolation, global fallback, override reset, missing-window noop, clamp range)
- [x] Manual: two OS windows side by side, `Cmd+=` × N in window-A only changes window-A; `Cmd+0` returns to default; new window opens at default; native macOS titlebar height tracks per-window zoom.

## Posture

Filed as **draft** — wanting to surface implementation shape against #1394 (canonical per-OS-window tracker per the earlier bot routing on #10115 and #1394) before requesting review. Happy to mark ready-for-review with any specific design preference on the resolution shape, scope, or clamp/range.


Eridanus117:
Signed.
@cla-bot check

CC
@warp-agent 
@oss-maintainers

Eridanus117:
@oz-agent

Eridanus117:
Thanks for the review @oz-agent. The concern is valid — even with the [0.5, 3.5] clamp, `adjust_zoom` would still no-op when the effective zoom landed on a non-`VALUES` percentage (e.g. 105%, 230%) because the old implementation looked the current percent up by `position()` in `ZoomLevel::VALUES` and stepped by index.

Pushed `e7a783f`: replace the index lookup with a directional search for the next/prev `VALUES` entry strictly above/below `current_percent`, falling back to the boundary value if there is no further step. Behaviour for values already in VALUES is unchanged; non-VALUES values now snap to the nearest entry in the chosen direction.

Extracted the step computation as a free `next_zoom_step` so it's testable in isolation. Added 8 unit tests covering in-VALUES step, non-VALUES snap, and boundary fallback (at min, at max, above max, below min).

`cargo test --package warp next_zoom_step` → 8 passed.

`/oz-review`

app/src/workspace/view.rs
```diff
@@ -1055,6 +1055,94 @@ pub struct Workspace {
...
+#[cfg(test)]
+mod next_zoom_step_tests {
+    use super::next_zoom_step;
+    use crate::window_settings::ZoomLevel;
+
+    #[test]
+    fn step_up_from_in_values() {
+        assert_eq!(next_zoom_step(100, true), 110);
+        assert_eq!(next_zoom_step(110, true), 125);
+    }
```
alokedesai: this is a nit but mind adding new tests  in their own file (`view_tests.rs`) following the convention of this codebase?

alokedesai (commented):
@Eridanus117 Thanks for contributing!

My preference would be to make a setting for this that configures whether the zoom commands are per window or across all windows (akin to vscode's `window.zoomPerWindow` setting). The default for this setting would still be across all windows. 

I'd be open to making `zoomPerWindow` the default but I think we'd need to make the per-window zoom persist across app restarts to consider making that change.

Eridanus117:
Thanks for the review @alokedesai! Pushed three commits addressing your feedback (`99bed1a` → `dbcd973` → `b9c883a`):

**Setting + branch behaviour** (`99bed1a`)

Added `appearance.window.zoom_per_window: bool` (default `false`). When `false`, `Cmd+=` / `Cmd+-` / `Cmd+0` route through `WindowSettings::zoom_level` and the existing global path (legacy behaviour preserved). When `true`, the shortcuts read/write the focused window's per-window override and recompute the native macOS titlebar height explicitly.

**Persistence across app restarts** (same commit)

`WindowSnapshot` gains `zoom_factor_override: Option<f32>`; new SQLite migration `2026-05-06-170107_add_zoom_factor_override_to_windows` adds a nullable `FLOAT` column. The save path captures the override at snapshot time, the load path restores it onto the new `Window` after tab restoration. New `test_sqlite_round_trips_zoom_factor_override` covers the `None` / `Some(_)` round-trip.

With persistence in place I left the default at `false` so this PR keeps current behaviour for everyone unchanged. Flipping the default to `true` can land as a follow-up at your discretion.

**Toggle-off semantics** (`b9c883a`)

To make the setting truly mean "across all windows" when disabled, subscribed to `WindowSettingsChangedEvent::ZoomPerWindow`: switching the setting to `false` clears each window's override (and drops it from the next snapshot), and the restore handler now applies a snapshot override only when the setting is currently enabled. So toggling `zoom_per_window` off retroactively brings every window back to the global default, and restarting with the setting off doesn't resurrect overrides from earlier sessions.

**Test relocation** (`dbcd973`)

Moved the eight `next_zoom_step` unit tests from the inline `#[cfg(test)] mod` block in `view.rs` to sibling `view_test.rs`, with a `next_zoom_step_*` prefix. Same commit applies `cargo fmt --all` (red CI was rustfmt only).

`cargo build --bin warp-oss --features gui`, `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --features gui`, and `cargo test --package warp --lib zoom` (9 passed) all clean locally.

`/oz-review`

app/src/window_settings.rs
```diff
@@ -79,6 +79,15 @@ define_settings_group!(WindowSettings, settings: [
         toml_path: "appearance.window.zoom_level",
         description: "The zoom level for the window, as a percentage.",
     },
+    zoom_per_window: ZoomPerWindow {
```
alokedesai, line 82: Let's also add a way to configure this in settings (specifically, via a dropdown in the settings page)

Image

app/src/workspace/view.rs
```diff
@@ -15999,33 +16062,62 @@ impl Workspace {
...
+        // - When false (default), step `WindowSettings::zoom_level` within
+        //   the discrete `ZoomLevel::VALUES` percentages; the settings
+        //   subscriber relays the value into `ctx.set_zoom_factor` and every
+        //   window updates together (legacy app-wide behaviour preserved for
+        //   existing users).
+        // - When true, step the focused window's per-window override; other
+        //   windows are unaffected. The native macOS titlebar height is
+        //   recomputed explicitly because the per-window path does not fire
+        //   `WindowSettingsChangedEvent::ZoomLevel`.
+        //
```
alokedesai: The difference between these two flows is a little confusing. Instead, could we:

1) Unconditionally read the current zoom level from the window instead of from `WindowSettings::zoom_level`
2) Determine the next step by iterating through the set of values, finding the last index where the value is <= the current zoom level, and then choosing the next step after that

Once we have the new zoom value, we can determine whether it's a window override or a new global setting based on the value of the setting

app/src/workspace/view.rs
```diff
@@ -15999,33 +16062,62 @@ impl Workspace {
...
+        // the nearest VALUES entry in the chosen direction via
+        // `next_zoom_step` rather than indexing into VALUES directly.
+        if *WindowSettings::as_ref(ctx).zoom_per_window {
+            let window_id = ctx.window_id();
+            let current_factor = ctx.window_zoom_factor(window_id);
+            let current_percent = (current_factor.as_f32() * 100.0).round() as u16;
+            let next_percent = next_zoom_step(current_percent, increase);
+            let next_factor = next_percent as f32 / 100.0;
+            ctx.set_window_zoom_factor(window_id, next_factor);
+            self.update_titlebar_height(ctx);
```
alokedesai: Can we emit an event from the `windowing::State` model instead so _any_ change to a zoom level from the window is event driven? Then, we can listen to events from the model to determine whether we need to update the titlebar_height or not instead of needing to remember to update it at the callsite like this

app/src/window_settings.rs
```diff
@@ -79,6 +79,15 @@ define_settings_group!(WindowSettings, settings: [
...
+        type: bool,
+        default: false,
+        supported_platforms: SupportedPlatforms::ALL,
+        sync_to_cloud: SyncToCloud::Globally(RespectUserSyncSetting::Yes),
+        private: false,
+        toml_path: "appearance.window.zoom_per_window",
+        description: "Whether the zoom keyboard shortcuts (Cmd+=, Cmd+-, Cmd+0) scope to the focused OS window only. When false (the default), the shortcuts change the app-wide zoom level. When true, each OS window keeps its own zoom and the shortcuts only affect the focused window; the value of `zoom_level` is then used as the default zoom for windows without an explicit per-window override.",
```
alokedesai: Let's describe them as "Zoom commands" instead of keyboard shortcuts (they can also be invoked via the command palette and referencing them as keyboard shortcuts is confusing because the exact shortcuts depend on which OS you're on).

Something like:

"Whether the Zoom comands (Zoom In, Zoom Out, Reset Zoom)  are scoped to the active window only. When false (the default), these commands change the zoom level across all windows. When true, each window keeps its own zoom and the commands only affect the current active window."

crates/warpui_core/src/core/app.rs
```diff
@@ -1178,19 +1178,74 @@ impl AppContext {
...
+    /// [`AppContext::set_window_zoom_factor`] are not affected; windows that
+    /// have no override will follow this new default.
     ///
     /// All views in every window are invalidated when this is invoked.
     ///
     /// ## Validation
-    /// The zoom factor is clamped to the range [0.5, 4.0].
+    /// The zoom factor is clamped to the range [0.5, 3.5], matching
+    /// `WindowSettings::ZoomLevel::VALUES` (50%–350%). This range is shared
+    /// with [`AppContext::set_window_zoom_factor`] so that
```
alokedesai: related to the comment above, leaking implementatino of the app into the UI framework is semantically wrong and shouldn't really be needed if we change how we compute the next step

alokedesai (commented):
This is great! I have a few comments to simplify the approach slightly, but nothing major.

Really appreciate you taking this on

Eridanus117:
Pushed two commits to address the latest review:

- `3618c3a` — Address per-window zoom review feedback
  - Added a **Zoom commands** dropdown (Across all windows / Active window only) in Settings → Appearance → Window, wired to `WindowSettings.zoom_per_window` and `WindowSettingsChangedEvent::ZoomPerWindow`.
  - Reworked `adjust_zoom` to unconditionally read the current window's effective zoom, step via `next_zoom_step`, then route the write to the per-window override or global `zoom_level` based on the setting.
  - Made window zoom changes event-driven: `windowing::State` now emits `WindowZoomFactorChanged { window_id }` from `set_window_zoom_factor` / `reset_window_zoom_factor`; Workspace subscribes and updates titlebar height, removing the manual call-site updates.
  - Updated the setting description to refer to "Zoom commands" rather than keyboard shortcuts.
  - Removed the `ZoomLevel::VALUES` coupling from `warpui_core` and restored the core `[0.5, 4.0]` clamp.
- `562f786` — Fix zoom step boundary handling
  - Fixed `next_zoom_step` so that when the current zoom is below the minimum or above the maximum `VALUES`, the first step lands on the boundary value instead of skipping past it.
  - Added `next_zoom_step_up_below_min_starts_at_min` and `next_zoom_step_down_above_max_starts_at_max`.

Verified locally:
- `git diff --check`
- `cargo test --package warp --lib zoom` — 11 passed
- `cargo test --package warpui_core zoom` — 2 passed

app/src/window_settings.rs
Eridanus117: Done in `3618c3a`. Added a `Zoom commands` dropdown under Settings → Appearance → Window with `Across all windows` / `Active window only`, bound to `WindowSettings.zoom_per_window` and kept in sync via `WindowSettingsChangedEvent::ZoomPerWindow`.

app/src/workspace/view.rs
Eridanus117: Done in `3618c3a`. `adjust_zoom` now always reads the current window effective zoom, steps via `next_zoom_step`, then writes either the per-window override or `WindowSettings.zoom_level` depending on `zoom_per_window`. Boundary stepping is covered by the new tests in `562f786`.

app/src/workspace/view.rs
Eridanus117: Done in `3618c3a`. `windowing::State` now emits `WindowZoomFactorChanged { window_id }` from `set_window_zoom_factor` / `reset_window_zoom_factor`, and Workspace subscribes to update the titlebar height for that window. The manual update at the call site is gone.

app/src/window_settings.rs
Eridanus117: Done in `3618c3a`. Description now reads "Zoom commands" instead of referring to keyboard shortcuts.

crates/warpui_core/src/core/app.rs
Eridanus117: Done in `3618c3a`. Dropped the doc/clamp reference to `ZoomLevel::VALUES` from `warpui_core` and restored the original `[0.5, 4.0]` clamp; the discrete-step logic stays in the app layer.

Eridanus117:
image

alokedesai (approved):

PR #10184, opened 2026-05-05
Fix file tree refresh logic

kevinyang372:
## Description

Fix local file tree blinking and reshuffling after connecting to an SSH session.

**Root cause:** When a remote SSH session is active, the remote server streams frequent
`FileTreeEntryUpdated { Remote }` events for every filesystem change on the remote host.
The previous code called `rebuild_flattened_items()` on each such event, which re-flattened
*all* roots (including local ones), causing the local file tree to re-render on every remote
filesystem change — visible as constant blinking/reshuffling.

**Fix:** Add an optional `target_root` parameter to the core `rebuild_flatten_items_impl`
method. When `Some`, only the specified root is re-flattened; all other roots keep their
existing items untouched. Three thin wrappers provide the public API:

- `rebuild_flattened_items_for_root(path)` — single-root rebuild
- `rebuild_flattened_items()` — full rebuild (all roots)
- `rebuild_flattened_items_without(path)` — full rebuild excluding a deleted path

Updated the following event handlers to use per-root rebuilds instead of full rebuilds:

- `FileTreeEntryUpdated { Local }` — rebuilds only the affected local root(s)
- `FileTreeEntryUpdated { Remote }` — rebuilds only the affected remote root
- `RepositoryRemoved { Remote }` — no-op rebuild (root already removed), avoids touching remaining roots

Skipping unchanged roots is safe because the rebuild is fully deterministic given the same
`entry`, `expanded_folders`, and `item_states` — re-flattening an unmodified root produces
an identical items list and selection index.

## Testing
- `cargo check -p warp --lib` passes cleanly
- Verified the three updated event handlers and the rename flow (editing.rs) all route through `rebuild_flatten_items_impl` correctly

## Agent Mode

### DIFF
--- a/app/src/code/file_tree/view.rs
+++ b/app/src/code/file_tree/view.rs
@@ -549,13 +549,15 @@ impl FileTreeView {
                 if !root_paths.is_empty() {
                     let id = RepositoryIdentifier::Local(std_path.clone());
                     if let Some(state) = RepoMetadataModel::as_ref(ctx).get_repository(&id, ctx) {
-                        for root_path in root_paths {
-                            if let Some(root_dir) = self.root_directories.get_mut(&root_path) {
+                        for root_path in &root_paths {
+                            if let Some(root_dir) = self.root_directories.get_mut(root_path) {
                                 root_dir.entry = state.entry.clone();
                             }
                         }
 
-                        self.rebuild_flattened_items();
+                        for root_path in &root_paths {
+                            self.rebuild_flattened_items_for_root(root_path);
+                        }
                         self.apply_pending_focus_target();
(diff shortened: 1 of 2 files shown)


kevinyang372:
* **#10184** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10184?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10184?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

moirahuang (approved):

PR #10186, opened 2026-05-05
Fix editor panic from overlapping V4A diff deltas (WARP-CLIENT-DEV-NYY)

kevinyang372:
## Problem

Sentry issue `WARP-CLIENT-DEV-NYY`: `panic: Invalid edit range 4042..3982` in `Buffer::edit`.

When the AI agent produces a multi-hunk V4A diff where one hunk's matched range subsumes another's (e.g. a large deletion whose context window overlaps a nearby single-line edit), `fuzzy_match_v4a_diffs` produces `DiffDelta`s with overlapping `replacement_line_range` values. When `CodeEditorModel::apply_diffs` later converts these to char offsets and feeds them to `insert_at_offsets`, the editor's `Buffer::edit` panics on the invalid range.

Confirmed via MAA conversation `d71bf84b` (request `b621adb3`): a 17-hunk V4A diff against `mod.rs` produced overlapping deltas (lines 46..71 subsuming 64..65).

## Fix

**Diff layer** (`crates/ai/src/diff_validation/mod.rs`):
- Added `deduplicate_overlapping_deltas()` — after matching all hunks, sort deltas by start line and drop any delta whose range overlaps with a preceding one.

**Editor safeguard** (`crates/editor/src/content/core.rs`):
- In `apply_core_edit_actions()`, skip any action whose anchor-resolved range has `start > end` instead of panicking. Returns `EditResult::default()` if all actions are skipped.

## Tests

- `test_v4a_maa_crash_d71bf84b_no_overlapping_deltas` — uses the exact V4A hunks from the crash to verify the subsumed delta is dropped (1 delta survives, not 2).
- `test_insert_at_offsets_overlapping_ranges_skipped` — passes inverted char-offset range (`4042..3982`) to `insert_at_offsets` and verifies the buffer is unchanged (edit gracefully skipped).

Fixes WARP-CLIENT-DEV-NYY

### DIFF
--- a/crates/ai/src/diff_validation/mod.rs
+++ b/crates/ai/src/diff_validation/mod.rs
@@ -402,6 +402,14 @@ pub fn fuzzy_match_v4a_diffs(
         }
     }
 
+    // Sort by start line and remove overlapping deltas. When the LLM produces
+    // multiple hunks targeting the same region (e.g. a large deletion whose
+    // matched range subsumes a nearby single-line edit), the overlapping delta
+    // must be dropped — applying both would produce an invalid edit range in
+    // the editor buffer (see WARP-CLIENT-DEV-NYY).
+    deltas.sort_by_key(|d| d.replacement_line_range.start);
+    deltas = deduplicate_overlapping_deltas(deltas);
+
     let update_deltas_empty = deltas.is_empty();
     let failures = if failures.fuzzy_match_failures > 0
         || failures.missing_line_numbers > 0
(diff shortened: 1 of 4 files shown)


kevinyang372:
* **#10186** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/ai/src/diff_validation/mod.rs
```diff
@@ -420,6 +428,33 @@ pub fn fuzzy_match_v4a_diffs(
...
+/// Adjacent ranges (`A.end == B.start`) are kept.
+fn deduplicate_overlapping_deltas(sorted_deltas: Vec<DiffDelta>) -> Vec<DiffDelta> {
+    let mut result: Vec<DiffDelta> = Vec::with_capacity(sorted_deltas.len());
+
+    for delta in sorted_deltas {
+        let dominated = result.last().is_some_and(|prev| {
+            delta.replacement_line_range.start < prev.replacement_line_range.end
+        });
+        if dominated {
+            log::warn!(
```
harryalbert, line 444: should this be an error? Just wondering if this is a thing quality should be aware of/worrying about or if it doesn't really matter

crates/editor/src/content/core.rs
```diff
@@ -293,6 +306,12 @@ impl Buffer {
             new_range_anchors.push(new_anchors);
         }
 
+        // If every action in the batch was skipped (e.g. all had inverted ranges),
+        // there is nothing to commit — return early.
+        if new_range_anchors.is_empty() {
```
harryalbert, line 311: could we surface this in some way to the agent (maybe in the form of a failed tool call or something?). Just feels like it should be aware if the edit wasn't applied

harryalbert (approved):

crates/ai/src/diff_validation/mod.rs
kevinyang372: I think of error as an unexpected state we shouldn't have gotten into

This is totally a reasonable state that could have happened if LLM outputs the diffs this way. It also self-heals since LLM will see the echoed back resulting state. I am not worried about putting this as warn level

crates/editor/src/content/core.rs
kevinyang372: Related to my comment above. The agent will see the resulting file state after the applied diff

crates/ai/src/diff_validation/mod.rs
harryalbert: cool, makes sense

PR #10188, opened 2026-05-05
Avoid parallel precheck execution

kevinyang372:
## Description
We were running three steps in parallel with `futures::join!`
1. Check binary
2. Check old binary for autoupdate
3. Detect platform

This will end up using 3 control master connections. Since each connection is a limited resource, this could cause these open failed issues.

For now I am moving these to be sequential. I considered combining them into one command and have some client side parsing to reconstruct the results but I am not confident about that change with different shells and potential error states. We could observe if latency with sequential execution becomes a problem and decide if we want to further optimize

### DIFF
--- a/crates/remote_server/src/manager.rs
+++ b/crates/remote_server/src/manager.rs
@@ -482,16 +482,16 @@ impl RemoteServerManager {
             ctx.background_executor()
                 .spawn(async move {
                     // Run platform detection, binary check, and old-binary
-                    // check concurrently. The old-binary check lets the
-                    // controller distinguish fresh install (no prior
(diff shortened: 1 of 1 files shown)


kevinyang372:
* **#10188** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10188?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10188?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

MaggieShan (approved):

PR #10196, opened 2026-05-05
[APP-3792] handshake for codebase indexing for remote envs

moirahuang:
## Description
Handshake for codebase indexing for remote envs, based off of the specs from https://github.com/warpdotdev/warp/pull/9508

## Testing
Locally checked for logs that confirm the handshake is happening
Screenshot 2026-05-05 at 4 55 26 PM

## Agent Mode

### DIFF
--- a/app/src/remote_server/server_model.rs
+++ b/app/src/remote_server/server_model.rs
@@ -5,6 +5,7 @@ use std::collections::{HashMap, HashSet};
 use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
 use warp_core::channel::ChannelState;
 use warp_core::safe_error;
 use warp_core::SessionId;
@@ -19,12 +20,13 @@ use warp_util::file::FileId;
 
 use super::proto::{
     client_message, delete_file_response, run_command_response, server_message,
-    write_file_response, Abort, Authenticate, ClientMessage, DeleteFile, DeleteFileResponse,
-    DeleteFileSuccess, ErrorCode, ErrorResponse, FailedFileRead, FileContextProto,
-    FileOperationError, Initialize, InitializeResponse, NavigatedToDirectory,
-    NavigatedToDirectoryResponse, ReadFileContextResponse, RunCommandError, RunCommandErrorCode,
-    RunCommandRequest, RunCommandResponse, RunCommandSuccess, ServerMessage, SessionBootstrapped,
-    WriteFile, WriteFileResponse, WriteFileSuccess,
+    write_file_response, Abort, Authenticate, ClientMessage, CodebaseIndexStatusUpdated,
+    CodebaseIndexStatusesSnapshot, DeleteFile, DeleteFileResponse, DeleteFileSuccess, ErrorCode,
+    ErrorResponse, FailedFileRead, FileContextProto, FileOperationError, Initialize,
+    InitializeResponse, NavigatedToDirectory, NavigatedToDirectoryResponse,
+    ReadFileContextResponse, RunCommandError, RunCommandErrorCode, RunCommandRequest,
+    RunCommandResponse, RunCommandSuccess, ServerMessage, SessionBootstrapped, WriteFile,
+    WriteFileResponse, WriteFileSuccess,
 };
 
 /// How long the daemon waits with no connections before exiting.
@@ -32,6 +34,9 @@ pub const GRACE_PERIOD: std::time::Duration = std::time::Duration::from_secs(10 
 
 /// Unique identifier for a connected proxy session in daemon mode.
 pub type ConnectionId = uuid::Uuid;
+use super::codebase_index_proto::{
+    statuses_to_snapshot_proto, RemoteCodebaseIndexState, RemoteCodebaseIndexStatus,
+};
 use super::protocol::RequestId;
 use crate::ai::agent::FileLocations;
 use crate::ai::blocklist::{read_local_file_context, ReadFileContextResult};
@@ -198,6 +203,8 @@ pub struct ServerModel {
     pending_file_ops: PendingFileOps,
     /// Daemon-wide auth credentials and user identity.
     auth: DaemonAuthContext,
+    /// Remote codebase-index statuses reported to connected clients.
+    codebase_index_statuses_by_repo: HashMap<String, RemoteCodebaseIndexStatus>,
 }
 
 impl Entity for ServerModel {
@@ -223,6 +230,7 @@ impl ServerModel {
             executors: HashMap::new(),
             pending_file_ops: PendingFileOps::new(),
             auth: DaemonAuthContext::new(),
+            codebase_index_statuses_by_repo: HashMap::new(),
         };
         // Subscribe to FileModel and RepoMetadataModel events
         // file operation results and repo metadata pushes are forwarded to all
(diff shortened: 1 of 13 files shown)


crates/remote_server/src/manager.rs
```diff
@@ -1215,16 +1354,47 @@ impl RemoteServerManager {
...
+                    let status_count = statuses.len();
+                    // The daemon intentionally pushes the initial codebase
+                    // status snapshot immediately after `InitializeResponse`.
+                    // The client reader resolves the initialize request and
+                    // forwards push messages on independent async paths, so the
+                    // snapshot can arrive while the manager still records this
+                    // session as `Initializing`. Buffer it instead of dropping
+                    // it; `mark_session_connected` will have the negotiated
+                    // host/identity and can emit the normal manager event.
+                    self.pending_initial_codebase_index_statuses_snapshots
```
moirahuang: i added this to track pending codebase index statuses because i was seeing issues where we're still initializing and losing the snapshot

app/src/remote_server/server_model.rs
```diff
@@ -198,6 +203,8 @@ pub struct ServerModel {
     pending_file_ops: PendingFileOps,
     /// Daemon-wide auth credentials and user identity.
     auth: DaemonAuthContext,
+    /// Remote codebase-index statuses reported to connected clients.
+    codebase_index_statuses_by_repo: HashMap<String, RemoteCodebaseIndexStatus>,
```
kevinyang372: What value does this state provide? I thought the canonical state should live in crates/ai/src/index/full_source_code_embedding/codebase_index.rs?

app/src/remote_server/server_model.rs
```diff
@@ -463,6 +476,9 @@ impl ServerModel {
         match outcome {
             HandlerOutcome::Sync(message) => {
                 self.send_server_message(Some(conn_id), Some(&request_id), message);
+                if push_codebase_status_snapshot_after_response {
```
kevinyang372: nit: instead of having `push_codebase_status_snapshot_after_response` bool, we could just match on message == server_message::Message::InitializeResponse here yeah?

app/src/remote_server/server_model.rs
```diff
@@ -474,6 +490,61 @@ impl ServerModel {
         }
     }
 
+    fn send_codebase_index_statuses_snapshot(
+        &self,
+        conn_id: Option<ConnectionId>,
+        request_id: Option<&RequestId>,
```
kevinyang372: What are the other cases where we need to send back all of the codebase index statuses?

For dedicated indexing requests for a specific repo, we just need to send back the status for that repo no?

app/src/remote_server/server_model.rs
```diff
@@ -474,6 +490,61 @@ impl ServerModel {
         }
     }
 
+    fn send_codebase_index_statuses_snapshot(
+        &self,
+        conn_id: Option<ConnectionId>,
```
kevinyang372: When would conn_id be None?

app/src/remote_server/server_model.rs
```diff
@@ -440,6 +450,9 @@ impl ServerModel {
             Some(client_message::Message::LoadRepoMetadataDirectory(msg)) => {
                 self.handle_load_repo_metadata_directory(msg, &request_id, ctx)
             }
+            Some(client_message::Message::ListCodebaseIndexStatuses(_)) => {
```
kevinyang372: Erm I don't remember this proto message in the tech spec -- what is this for?

app/src/remote_server/server_model.rs
```diff
@@ -861,6 +937,30 @@ impl ServerModel {
                     ),
                 );
 
+                // Remote codebase indexing is repo-scoped, so only git repo
+                // navigations get a status entry.
+                if is_git {
```
kevinyang372: I thought indexing is driven by the client? We don't know if the client has auto-indexing on here for example right?

crates/remote_server/src/manager.rs
```diff
@@ -417,6 +450,17 @@ pub struct RemoteServerManager {
     /// remote server daemon on every (re)connect. Persists until
     /// `deregister_session`.
     session_bootstrap_info: HashMap<SessionId, SessionBootstrapInfo>,
+    /// Initial remote codebase-indexing status snapshots received before a
```
kevinyang372: Emmm is this true? The delivery of messages are guaranteed to be sequential and on the client side, once we get the initialized message, we synchronously update the connection status

If it is actually true, we have a bigger problem we need to address here

crates/remote_server/src/manager.rs
kevinyang372: As mentioned above, we should fix the root cause of this. Transitioning from initializing to connected should be synchronous once we receive the initialization response

app/src/remote_server/server_model.rs
moirahuang: it's in the spec! https://github.com/warpdotdev/warp/blob/master/specs/APP-3792/TECH.md?plain=1#L162

mostly used to re-sync status. not sure if this is what you're imagining but maybe i can just reuse `CodebaseIndexStatusesSnapshot`?

edit: actually hm i'm not sure if that'll work because it needs to allow the client to request a refresh

app/src/remote_server/server_model.rs
moirahuang: it shouldn't be, i can update to not be `Option`

app/src/remote_server/server_model.rs
moirahuang: yeah this isn't the canonical state, it's just the current known statuses. it's relevant for `ListCodebaseIndexStatuses` so i can simplify `ListCodebaseIndexStatuses` and here if we end up deciding we don't need it in the discussion below

edit: i can also remove this once we have the canonical state in, i'll make that more clear in the comments

app/src/remote_server/server_model.rs
moirahuang: yeah, this is just to report back the not enabled status it doesn't actually start indexing

crates/remote_server/src/manager.rs
moirahuang: ok i tried to fix this, see `crates/remote_server/src/manager.rs` line 198. lmk if you think that isn't the right approach

app/src/remote_server/server_model.rs
moirahuang: similar to the list status discussion, we would need this for reconnection but maybe we don't need it entirely?

app/src/remote_server/server_model.rs
moirahuang: we talked over zoom, i'm going to remove `ListCodebaseIndexStatuses` and enforce the invariant that we'll always push the status

app/src/remote_server/server_model.rs
moirahuang: going to keep this for now but make it more clear that it's not the canonical source and that we will remove it in later PRs

crates/remote_server/src/manager.rs
moirahuang: we synced, going to move `event_rx` to be returned from `run_connect_and_handshake`

app/src/remote_server/server_model.rs
kevinyang372: Erm why do we need to report back the not enabled status? Shouldn't the client already know this from the initial snapshot push?

kevinyang372 (approved):

app/src/remote_server/server_model.rs
moirahuang: removed

Issue #10036, opened 2026-05-04
Add setting to maximise review panel


DPirate:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

A setting which lets me open review panel maximised by default.

### Is your feature request related to a problem? Please describe.

Right now when I click on code review button on top right, a new panel is added on the right and then if I want to maximise it I have to click on another button.

Since laptop screen is small for multiple panels (file browser, agent, opened file panel, review panel) I almost always have to maximise the review panel.

### Operating system (OS)

macOS

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

peicodes:
Thanks for the feature suggestion! We'd like to see how the community responds to this request before deciding whether it's product aligned.

PR #10223, opened 2026-05-06
spec: open review panel maximised by default (#10036)

lonexreb:
Spec for #10036. New `code.editor.review_panel.open_maximized` boolean setting (default false). When enabled, clicking the code-review button opens the panel pre-maximised in one click. Default-off preserves today's behavior exactly.

Closes (spec-only) #10036.

### DIFF
--- a/specs/GH10036/SPEC.md
+++ b/specs/GH10036/SPEC.md
@@ -9,13 +9,17 @@ flow is friction for the most common case.
 
 ## Goal
 
-Add a boolean setting `editor.review_panel_open_maximized` (default
-`false`) that, when enabled, opens the review panel pre-maximised.
-The non-maximised default flow is unchanged.
+Add a boolean setting `code.editor.review_panel.open_maximized`
+(default `false`) that, when enabled, opens the review panel
+pre-maximised. The non-maximised default flow is unchanged.
+
+The TOML key uses the existing `code.editor.*` namespace already in
+use by other code-editor settings in this surface, with the
+`review_panel` sub-namespace grouping panel-specific preferences.
 
 ## Behavior contract
 
-- B1. New setting `editor.review_panel_open_maximized: bool`,
+- B1. New setting `code.editor.review_panel.open_maximized: bool`,
   default `false`, `SyncToCloud::Globally(RespectUserSyncSetting::Yes)`
   (matches the convention for UI prefs in
   `app/src/settings/code.rs:19/29/40/48`).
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) renamed setting key to `code.editor.review_panel.open_maximized` to match the existing `code.editor.*` TOML namespace; (2) added runtime-toggle test covering the A3 acceptance criterion.

lonexreb:
Round-2 fix: audited every mention of the setting key across Behavior, Settings, Acceptance, Tests, Implementation Pointers, and Open Questions; renamed any stragglers to the single canonical key `code.editor.review_panel.open_maximized`. Added explicit 'Canonical setting key' callout at top of Settings section to prevent future drift.

lonexreb:
Addressed all 2 important + 2 suggestion concerns from latest oz-for-oss review:

**Settings UI path corrected (important #1)**
The spec previously claimed 'Settings → Editor → Code review' — that path does NOT exist. Verified against `app/src/settings_view/code_page.rs`: the actual category is `Category::new("Code Editor and Review", code_editor_review_widgets)` at lines 314 and 438. B4 now states the toggle is appended to `code_editor_review_widgets` adjacent to `show_code_review_button`, `show_code_review_diff_stats`, and `auto_open_code_review_pane_on_first_agent_change`. Both `cfg(...)` branches must be updated.

**Implementation sketch grounded in real code (important #2)**
Removed the bogus 'review_panel anchor in code.rs' reference (no such anchor exists — sibling settings are `show_project_explorer`, `show_global_search`, etc.). New sketch shows the exact `define_settings_group!` field shape, points at `setup_code_review_panel` (line 8001) and `open_code_review_panel_from_arg` (line 8055) as the runtime entry points, and explains that we seed `PaneGroup::is_right_panel_maximized` rather than introducing a new action variant.

**Maximize/restore state-machine conflict resolved (suggestion)**
B5 now has 4 sub-rules + concrete acceptance criteria A4 and A5 covering the 'previously maximised before being closed' case the bot called out:
- Tab restore (`RightPanelSnapshot`) overrides the setting — the saved per-tab state is authoritative.
- Close-then-reopen within a session uses the setting (because the cross-close persistence is only `RightPanelSnapshot`, which only fires on tab restore).
- Runtime toggle never retroactively maximises an already-open panel.
- New tests T5 and T6 enforce these rules.

**PR description (suggestion) — note for maintainer**
Spec is now the source of truth and forbids any variant other than `code.editor.review_panel.open_maximized` (with grep test T8). The PR body still uses the older `editor.review_panel_open_maximized` spelling and should be updated by a maintainer with edit permission on the PR description; my role here is limited to the spec file and review comments.

lonexreb:
Addressed the latest oz-for-oss review:

- **Settings UI path**: Corrected to `GeneralSettings` in `app/src/terminal/general_settings.rs` (line 159 — appended after `auto_open_code_review_pane_on_first_agent_change`). Verified that code-review settings are split across `TabSettings` (per-tab) and `GeneralSettings` (global), and the new global flag belongs with the latter. Removed references to the non-existent `define_settings_group!(CodeSettings, ...)` block.
- **State-machine conflict**: Reconciled the OFF default with the existing `is_right_panel_maximized` / `RightPanelSnapshot` state machine. The setting + snapshot are orthogonal: snapshot always wins on tab/session restore, the setting always wins on fresh opens. Close-then-reopen is treated as a fresh open (no `RightPanelSnapshot` is created on close). A1 now spells out the previously-maximised-then-closed case.
- **PR description key drift**: Edited the PR description to use the canonical `code.editor.review_panel.open_maximized` (was `editor.review_panel_open_maximized`). Spec now mandates a grep of PR descriptions/commits for non-canonical variants.

Issue #9965, opened 2026-05-03
Show running weekly USD spend in a persistent status bar at the bottom of the Warp window


nslonim:
## Summary
Add a persistent indicator at the bottom of the Warp window that shows the user's running weekly USD spend on Agent usage, so users can budget their AI spend and avoid surprises without having to navigate into Settings > Billing and usage.

## Problem
Today, AI spend visibility in Warp is split across two surfaces:

- A per-turn credit count chip at the bottom of an Agent response, which only covers the most recent turn.
- The Settings > Billing and usage dashboard, which covers monthly totals and is several clicks away from a normal terminal/Agent workflow.

Neither surface gives a continuously visible signal of how much money has been spent recently. For users on Build/Business plans (and especially anyone using Add-on Credits, where AI usage maps directly to dollars), this makes it hard to:

- Track spend over a shorter horizon than a full billing cycle.
- Notice mid-week when usage is running hot before it becomes a surprise on the bill.
- Stay within a personal weekly budget without proactively opening Settings.

The user filing this feedback explicitly said: "I want to budget my usage and avoid surprises."

## Reproduction steps or desired workflow
Desired workflow:

1. Use Warp normally throughout the week (Agent conversations, Generate, AI Autofill, etc.).
2. At any time, glance at the bottom of the Warp window and see something like `This week: $4.20` (USD).
3. Optionally hover or click the indicator to expand a small breakdown by model (e.g., Claude Opus, Claude Sonnet, GPT-5), since model choice is the dominant driver of credit/dollar consumption per the credits docs.
4. The weekly total resets at the start of each week (a clearly documented boundary, e.g., user's local Monday or the start of the billing week).

Acceptance-style notes:

- Indicator must be persistent (always visible) at the bottom of the Warp UI, not gated behind opening Settings.
- Default unit is USD spend (dollars), not raw credits or tokens. Credits are non-deterministic and harder to reason about as a budget; the user explicitly asked for dollars.
- Default view shows only the weekly total. An optional expanded/hover view can break the total down by model.
- Setting to hide the indicator would be welcome for users who don't want it visible (e.g., screen sharing, pair programming).
- Should respect team vs. personal accounting where applicable (a user on a team with Add-on Credits should see their personal slice, not the entire team pool).

## Artifacts
None attached

## Warp version
Unknown

## Operating system
macOS (specific version not collected)

## Possible source references
[...]

PR #10224, opened 2026-05-06
spec: persistent weekly USD spend status bar (#9965)

lonexreb:
Spec for #9965. Opt-in (`agents.show_weekly_spend_status`, default false) status-bar indicator showing rolling 7-day Agent spend. Click opens Settings → Billing. Hidden when billing data is unavailable (avoids misleading $0.00).

Closes (spec-only) #9965.

### DIFF
--- a/specs/GH9965/SPEC.md
+++ b/specs/GH9965/SPEC.md
@@ -48,11 +48,13 @@ window/workspace.
   key keeps the existing `weekly` token for backwards-compat with
   any in-progress drafts; user-facing copy uses "7-day spend".
 - B2. When ON, a status-bar item shows "$XX.XX · 7-day spend" in the
-  same status-bar slot used by other agent indicators (verify slot
-  by grepping the existing status-bar render path).
+  right-hand status-bar group, immediately to the left of the
+  context-remaining chip. See "Status-bar placement" below for the
+  precise slot, ordering, and width-constrained collapse rules.
 - B3. The indicator updates on every agent turn finish AND on a
-  60-second timer (cheap; the spend value comes from the same
-  billing model that already polls).
+  60-second refresh timer. Failure handling uses a separate
+  exponential-backoff retry timer; the 60 s cadence governs only the
+  success path. See "Refresh vs failure-retry timers" below.
 - B4. Hover tooltip shows: active scope name, rolling 7-day total
   (USD with 2 decimal places), current month total, the caption
   "rolling 7 days", and a "View details →" link that opens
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) added Billing Scope subsection clarifying personal vs team/workspace selection via workspace selector; (2) renamed UI label from 'Weekly spend' to '7-day spend' (rolling) to match requirements; (3) added Data Source subsection specifying Billing & Usage backend, ≤2s TTL after turn completion, stale-on-error, hide-on-30s-failure; (4) added tests for hidden-unavailable, refresh timing, and tooltip total formatting.

lonexreb:
Round-2 fixes: (1) data source — added explicit dependency on `usage_rolling_7d_usd` field in existing `BillingUsage` response (additive, no new endpoint); flagged in Open Questions until backend confirms; (2) separated refresh vs failure-retry timers — refresh on 60s cadence + post-turn event; failures use exponential backoff 5/10/20/40/80s cap; ≥30s since success → 'stale' tooltip; ≥5 min consecutive failures → hide segment; reconciled the prior 30s-conflict; (3) status-bar placement now names concrete slot — same right-group as context-remaining indicator, ordering `[7d spend][context%][others]`, with explicit collapse priority on width constraint.

lonexreb:
Addressed both concerns from latest oz-for-oss review:

**Cross-scope leak (important + security)**
Added new 'Cross-scope leak prevention' subsection with 5 mandatory implementation rules:
1. Synchronous clear — cache invalidation happens *before* the workspace-selector mutation commits to the UI tree, so no frame ever contains the old value alongside the new scope label.
2. Scope-keyed cache — cache lookups returning a different-scope value are treated as misses, not displayed.
3. Immediate scoped refresh — dispatched in the same code path as the synchronous clear, NOT waiting for the 60 s tick.
4. In-flight fetch cancellation + scope-id verification on response handlers.
5. Tooltip parity — scope label and dollar value read from the same cache entry.

Added new B6.1 specifying the loading-state placeholder (`— · 7-day spend`) used between the synchronous clear and the new scope's first successful fetch. Width-equivalent so the right cluster does not reflow. Existing 5-minute hide threshold absorbs prolonged loading.

New tests T9a (synchronous clear, asserted on the same UI commit), T9b (cross-scope cache miss), T9c (late-response discard with fault-injecting fake), and T9d (immediate scoped refresh). T9a is the explicit security-regression guard.

**Tooltip B4 coverage (suggestion)**
T7 expanded to cover ALL of B4's requirements:
- (a) active scope name
- (b) rolling 7-day USD with 2 decimals
- (c) **current-month total** (the missing item) — sourced from existing `BillingUsage.usage_current_month_usd`, no client-side aggregation
- (d) 'rolling 7 days' caption
- (e) **'View details →' link** (the missing item) — both click and keyboard activation dispatch `OpenSettingsPage(BillingAndUsage)`. Asserted with click-handler unit test plus DOM snapshot verifying all 5 fields.

Issue #10027, opened 2026-05-04
Redaction on matched command outputs and/or multiline redaction


micahnz:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

One of the biggest things holding be back from using this more right now in my SysOps/DevOps workflow is what is being transmitted to the agent which means I have to be very careful about commands I use during an agent session and their outputs, or limit my use.

Secret Redaction works great for single line API keys etc but there cases where I want to redact the entire output of a command like `env` or `kubectl get secret(s)?` so I don't have to change terminals when performing  commands that would create a sensitive output, so it would be helpful to have a secret redaction option that will match a command and redact the entire output of that command.

This could also be achieved allowing multi-line redaction so I can write rules like .*Kind: Secret.* or something and have it redact the entire output.

Both would be useful, the command output redaction probably the most useful since I could add common patterns like `ssh/id_ed25519`  or `id_rsa` and other sensitive files or directories so they never accidentally leak to an agent, I would also maintain a list of commands for my team for which all outputs should be considered secret which would be super handy.

### Operating system (OS)

Linux

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

micahnz:
Adding to the idea by mentioning it would be useful to be able to mark blocks for exclusion in an existing agent session, essentially the opposite of "attach as agent context".

Would also be useful to be able to be able to specify when running a command if it should be excluded from context for one offs, or by way of adding a prefix that will run the command as secret and keep the command and output from any llm contexts.

These solve two problems that the command matching redaction doesn't, 1. One offs where you need to quickly run something that you know is secret to get on with what you are doing and don't want to go create a rule, or it is truly a one off and doesn't make sense to create a rule for and 2. In the case where a command was run in an existing agent enabled session but no follow up questions have been asked and the command and result have not been sent on a context, marking the item as secret after the fact will prevent accidental secret leakage in that case.

PR #10225, opened 2026-05-06
spec: multiline + command-keyed redaction (#10027)

lonexreb:
Spec for #10027. Two new redaction mechanisms: command-keyed output rules (e.g. `kubectl get secret` redacts entire block output) and multiline-bracketed rules (e.g. PEM BEGIN/END blocks). Defaults for common DevOps cases shipped in V1.

Closes (spec-only) #10027.

### DIFF
--- a/specs/GH10027/SPEC.md
+++ b/specs/GH10027/SPEC.md
@@ -81,28 +81,79 @@ is redacted.
 #### B1.1 — Command matching normalization
 
 `command_pattern` does **not** apply to the raw command string
-verbatim. Before matching, Warp normalizes the command:
+verbatim. Before matching, Warp normalizes the command into a
+*subcommand path string* and a separate *argv set*. The
+`command_pattern` regex matches ONLY against the subcommand path
+string. Trailing resource-name positionals are NOT included in the
+string `command_pattern` matches against — they are separated into
+the argv set.
 
 1. **Argv parse:** Split using shell-aware tokenization (the same
    parser used elsewhere in Warp). Quoting and escaping are
    respected.
 2. **Subcommand path extraction:** For known multi-word CLIs
    (`kubectl`, `aws`, `gcloud`, `vault`, `op`, `gh`, `doppler`,
-   `bw`), extract the ordered sequence of non-flag positionals
-   that form the subcommand path (e.g.,
-   `kubectl get secret`, `aws secretsmanager get-secret-value`).
-3. **Argument-order independence:** Flags (`-n ns`, `--namespace=ns`,
-   `-o yaml`) and additional positionals (resource names) are
-   collected into a *set*. The match operates on the tuple
-   `(subcommand_path, normalized_argv_set)`. This means
-   `kubectl -n ns get secret foo` and `kubectl get secret -n ns foo`
-   both match a rule keyed on subcommand path `kubectl get secret`.
-4. **Pattern semantics:** `command_pattern` is a Rust regex matched
-   against the normalized subcommand path string (a single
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) added Command Matching Normalization with argv-set matching and 12 default tool rules; (2) reconciled V1 scope to include PEM-style private-key delimiter redaction in both output AND pasted input; (3) added Default Override Semantics defining id-based merge, disable flag, and precedence; (4) fixed Buffer Cap Behavior to keep redaction state active post-cap until end delimiter or flush.

lonexreb:
Round-2 fixes: (1) all rule ids harmonized to `default-<tool>-<purpose>` naming convention with full table listed; user rules cannot use `default-` prefix; (2) every default command pattern rewritten as Rust regex with anchors (replaces glob `*`); pattern syntax subsection added; per-rule regex tests; (3) `inclusive` field defined (true = include delimiters in redaction, false = between only; default false); invalid user rules drop individually with warning; duplicate ids drop subsequent; reserved prefix rejected; built-in defaults stay active even on full config load failure (fail-closed invariant); (4) added Centralized Agent-Context Boundary enumerating every agent-bound path (block output, prompt, paste, tab-config commands, history replay, tool-call inputs, file reads) with single `redact_for_agent` choke-point and clippy-lint enforcement.

lonexreb:
Addressed both concerns from latest oz-for-oss review. **The CRITICAL is fully resolved.**

**CRITICAL — PEM bare-PKCS#8 unredacted (closed)**
The round-2 `start_pattern` was `-----BEGIN [A-Z ]+PRIVATE KEY-----` — `[A-Z ]+` requires at least one char before `PRIVATE`, so the bare `-----BEGIN PRIVATE KEY-----` (PKCS#8, no algorithm tag) form **did not match** and a common private-key block was silently leaked to the agent on both output and pasted-input paths.

Fix: changed quantifier to `[A-Z ]*` (zero-or-more). Spec now contains an explicit algorithm-form coverage table proving every PEM private-key variant is matched: bare PKCS#8, RSA, EC, OPENSSH, ENCRYPTED, DSA. Added inline comment in the TOML example explaining why the quantifier matters.

New test **T16** is the regression guard with six BEGIN..END fixtures (one per form), each tested on BOTH the output stream and the pasted-input path. The test MUST fail under `[A-Z ]+` and MUST pass under `[A-Z ]*`. CI fails if any fixture leaks a single key byte. The bare PKCS#8 case is the explicit closure of this critical gap.

**Important — Command-matching contract internal inconsistency (closed)**
Rewrote B1.1 with a single canonical answer: `command_pattern` is matched ONLY against the normalized **subcommand path string**, NEVER against a string that includes trailing resource-name positionals. Added concrete (subcommand_path, argv_set) splits showing exactly what each input becomes:

- `kubectl get secret foo` → path `kubectl get secret` + argv_set `{foo}`
- `kubectl -n ns get secret foo` → path `kubectl get secret` + argv_set `{-n, ns, foo}`
- `aws secretsmanager get-secret-value --secret-id x` → path `aws secretsmanager get-secret-value` + argv_set `{--secret-id, x}`

Added explicit clarification that the trailing `(\\s|$)` boundary in default patterns is for ZERO-OR-MORE FUTURE SUBCOMMAND TOKENS, not resource names — those are in the argv set, not in the matched string. Resource-name constraints use `argv_contains` / `argv_excludes`.

New test **T_command_match_contract** asserts the contract end-to-end with a recording test double on the matcher: it fails if the matcher EVER sees a string containing a trailing resource positional. Covers all 12 default rules in B1.2.

Issue #9953, opened 2026-05-03
The default proportional font (used in the Settings and to render Markdown) is hard to read


tredondo:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

The default font of Markdown previews looked too cramped to me, and difficult to read. The same font is used in the Warp Settings.

### Is your feature request related to a problem? Please describe.

I took a screenshot of Warp's rendering of its own README.md, side by side with GitHub, and asked Claude to analyze the readability of the text.

<img width="1700" height="649" alt="Image" src="https://github.com/user-attachments/assets/5d323b09-65aa-4923-a252-b66f1c595cea" />

Below is the raw reply.

------

Looking at the two renderings side by side, the right version is significantly more readable. Here's a breakdown of why:

### 1. GitHub-style alert/admonition rendering
- **Left:** The `> [!TIP]` syntax is rendered literally as raw markdown text with blockquote markers (`>`), leaving the user to mentally parse what it means.
- **Right:** It's rendered as a proper styled callout — a green vertical accent bar, a 💡 icon, and a "Tip" label. This instantly communicates *"this is supplementary advice, not part of the main flow."*

### 2. Typography and spacing
- **Right:** Generous line-height, comfortable paragraph spacing, and what looks like a slightly larger or better-hinted font. Text feels airy.
- **Left:** Tighter line-height and cramped paragraph gaps make the wall of text harder to scan. The eye doesn't get natural resting points.

### 3. Line length and measure
- **Right:** Wider column, but still within a comfortable reading measure. Fewer awkward line breaks mid-thought (e.g., "ad-hoc" gets hyphenated across a line on the left).
- **Left:** Narrower column forces more wrapping, including the unfortunate `ad-/hoc` hyphenation, which interrupts reading flow.

### 4. Inline code styling
- Both render inline code (`#oss-contributors`, `ready-to-spec`) with a pill background, but on the right the contrast and padding feel more deliberate, making the tokens pop without being noisy.

### 5. Visual hierarchy
- **Right:** Headings (`Open Source & Contributing`, `Issue to PR`, `Building the Repo Locally`) feel weightier relative to body text, creating clearer section boundaries.
- **Left:** Headings are present but the overall density flattens the hierarchy.

[...]

PR #10226, opened 2026-05-06
spec: improve default proportional font readability (#9953)

lonexreb:
Spec for #9953. New `appearance.proportional_font_{family,size,line_height}` settings with GitHub-baseline defaults (system-ui/Segoe UI Variable/Inter, 14px, 1.55). Settings → Appearance subsection with live preview.

Closes (spec-only) #9953.

### DIFF
--- a/specs/GH9953/SPEC.md
+++ b/specs/GH9953/SPEC.md
@@ -153,30 +153,61 @@ fall through to the appended Default stack on every render.
         ASCII space. Tabs and other whitespace inside an unquoted
         identifier sequence are rejected at parse time per CSS
         Fonts L4 grammar.
-     2. **Case is preserved verbatim from user input.** No
-        case-folding is applied to family names at any layer —
-        storage, sync, settings UI, render. CSS family names are
-        case-insensitive at *match* time (handled by the font
-        engine, not by us), but the *stored* value preserves the
-        user's chosen case so a peer device sees the same string
-        the user typed. This rule applies to both unquoted and
-        quoted families:
+     2. **Case is preserved verbatim from user input for class-3
+        named families.** No case-folding is applied to class-3
+        family names at any layer — storage, sync, settings UI,
+        render. CSS family names are case-insensitive at *match*
+        time (handled by the font engine, not by us), but the
+        *stored* value preserves the user's chosen case so a peer
+        device sees the same string the user typed. This rule
+        applies to both unquoted and quoted class-3 families:
           - Unquoted `inter`, `Inter`, `INTER` are stored
             byte-for-byte as the user wrote them.
           - Quoted `"helvetica neue"`, `"Helvetica Neue"`, and
             `"HELVETICA NEUE"` are stored byte-for-byte as the user
             wrote them.
-     3. **Generic-family keywords are normalized to lowercase.**
-        The CSS generic identifiers enumerated in class 1 above
-        (`system-ui`, `sans-serif`, etc.) are matched
-        case-insensitively at parse time and stored
-        lowercase-canonical. This is the ONE exception to rule
-        1.ii and is necessary because CSS generic identifiers are
-        a closed set with a defined lowercase spelling; storing
-        `Sans-Serif` would be a category-confusion bug. Vendor-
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) Linux default now uses system-available stack (Cantarell/Ubuntu/Liberation Sans) since Inter isn't bundled; (2) added font-family validation rules, missing-font cascade behavior, per-platform sync schema with reset-to-default semantics; (3) renamed `proportional_line_height` to `proportional_font_line_height` for naming consistency; (4) added explicit size-24 + clamp-boundary acceptance and tests; (5) added Surface Scope subsection limiting V1 to Settings panes + rendered Markdown; (6) all three new settings now share persistence/sync metadata; (7) tests cover clamp, synced-unsupported-family fallback, and visual comparison.

lonexreb:
Round-2 fixes: (1) per-platform defaults explicitly stated WITHOUT Inter — macOS uses `-apple-system, SF Pro Text`, Windows uses `Segoe UI Variable, Segoe UI`, Linux uses `Cantarell, Ubuntu, Liberation Sans`; spec note clarifies Inter must be opted-in by user config; (2) canonical serialization for font-family per CSS Fonts L4 — unquoted for simple identifiers, double-quoted with backslash-escaping for special chars; storage/sync use canonical form with round-trip identity; (3) render path is parsed-token-list -> CSS-string via dedicated serializer (no raw-string interpolation), preventing CSS injection; family-list parser/serializer module exposed and renderer takes `&FamilyList` not `&str` to enforce at the type level; added tests T_canonical_serialization, T_canonical_quoting, T_inter_not_in_default, T_render_no_injection.

lonexreb:
Addressed all 3 important concerns from latest oz-for-oss review:

**Generic-family resolution defined (important #1)**
Added 'Generic-family entries in the cascade' subsection classifying every family-list entry into 3 classes:
- **Class 1** (CSS generic keywords: `system-ui`, `sans-serif`, etc.) — emitted as-is, font engine resolves; **NOT subject to the installed-font check** (this was the gap the bot flagged).
- **Class 2** (vendor-prefixed: `-apple-system`) — recognized on matching platform, treated missing elsewhere.
- **Class 3** (named families) — the ONLY class B1b's installed-font cascade applies to.

Walks the macOS Default stack on Linux as a worked example. New test `T_generic_families_resolve` exercises each platform Default stack plus deliberate misses to confirm the cascade does NOT fall through past class-1 entries on the basis of 'not installed.'

**Canonical case-normalization disambiguated (important #2)**
B1a now has explicit, ordered normalization rules:
- Internal whitespace in unquoted identifiers collapsed to single ASCII space.
- **Case preserved verbatim from user input** for both unquoted and quoted named families. (The round-2 'or the user's preferred case' parenthetical was the ambiguity.)
- Class-1 generic and class-2 vendor-prefixed identifiers normalized to lowercase (closed-set canonicalization).

Full case-rule table added showing all combinations. A7 rewritten: `inter, ...` stays `inter`. Writing `SYSTEM-UI` stores `system-ui`. New test `T_canonical_case` enforces byte-for-byte the canonicalized values.

**CSS-injection guard at actual render path (important #3 / security)**
B1e now has two layers: Layer 1 (CSS quoting) is necessary; Layer 2 (insertion-context safety) is REQUIRED. Two safe paths:
- (a) **CSSOM / style-property assignment** — V1 default. Browser parses through CSS parser, never HTML tokenizer; `</style>` is inert.
- (b) **`<style>`-text-node fallback** — only allowed when the serializer additionally emits `\\3c\\2f` (or equivalent) for every `</` inside quoted family tokens, defeating the HTML tokenizer's `</style` lookahead.

Two entry points: `to_css_for_style_property()` and `to_css_for_inline_style_tag()`. Type system enforces declaration of which path. A8 rewritten to test the actual render path. `T_render_no_injection` split into `T_render_no_injection_property_path` (drives real DOM render, asserts `window.__injected_alert__` is unset) and `T_render_no_injection_style_tag_path` (asserts `\\3c\\2f` escape, no HTML breakout).

Issue #10171, opened 2026-05-05
Feature request: Allow tab configs to specify an agent profile for agent-mode panes


gh-owongfigma:
## Summary
Allow tab configs to specify an agent profile, so that when a tab opens in Agent Mode, the agent automatically uses the configured profile (e.g. full autonomy, specific model, custom allowlists) without requiring manual selection.

## Problem
Tab configs currently support `type = "agent"` to open a pane in Agent Mode, but there is no way to associate a specific agent profile with that pane. Users who have custom profiles — for example, a "Coder devbox" profile configured for full autonomy on a remote dev machine — must manually switch profiles every time they open the tab. This adds friction to workflows where the tab config is specifically designed for a trusted environment (like an SSH session to a dev container) where the agent should run without asking for approval.

## Reproduction steps or desired workflow
1. Create an agent profile in **Settings > Agents > Profiles** (e.g. "Coder devbox" with all permissions set to "Always allow").
2. Create a tab config with `type = "agent"` and a new field like `profile = "Coder devbox"` (or a profile ID).
3. Open the tab config from the `+` menu.
4. **Expected:** The agent pane opens with the "Coder devbox" profile already active.
5. **Actual (today):** The agent pane opens with whichever profile was last used; the user must manually switch.

## Artifacts
None attached.

## Warp version
Unknown

## Operating system
macOS

dhruvac29:
@oz-agent 
This looks like a useful enhancement, especially for repeatable trusted environments like dev containers or SSH-based devboxes.

One implementation detail that seems worth deciding early: should tab configs reference the agent profile by display name, stable profile ID, or support both? A display name is easier to write by hand, but a stable ID would be safer if profiles can be renamed.

Example config shape:

```toml
[[tabs]]
title = "Dev Agent"

[[tabs.panes]]
type = "agent"
profile = "Coder devbox"

zachlloyd:
I like this idea a lot.  @moirahuang @zachbai thoughts?

moirahuang:
> I like this idea a lot. [@moirahuang](https://github.com/moirahuang) [@zachbai](https://github.com/zachbai) thoughts?

Yeah I like this too! I think we don't expose the profile ID in the UI. So for now, let's use the profile name but if there are duplicate profile names, we should error instead of trying to choose one of the profiles. I'm going to move this to `ready-to-implement`!

rgatx:
[...]

PR #10229, opened 2026-05-06
spec: tab configs specify agent profile (#10171)

lonexreb:
Spec for #10171. New `profile` field on agent-typed panes in tab configs. When the tab opens, the named profile is applied before any agent input is dispatched. Missing profile = default + one-time toast.

Closes (spec-only) #10171.

### DIFF
--- a/specs/GH10171/SPEC.md
+++ b/specs/GH10171/SPEC.md
@@ -14,6 +14,24 @@ open is a trust hazard. The spec must define a consent model so
 that imported / shared configs cannot escalate autonomy without a
 visible user gate.
 
+## Security threat model and mitigations
+
+This section maps the four standing concerns from oz-for-oss
+review onto the binding clauses in the rest of the spec, so the
+reviewer can verify coverage at a glance.
+
+| # | Threat | Concrete mitigation in this spec |
+| - | ------ | -------------------------------- |
+| S1 | An attacker ships a deceptive imported tab config whose `prompt` or `commands` field contains ANSI control sequences, OSC 8 hyperlinks, bidi overrides, or zero-width characters that hide what would actually run. | B-CONSENT-INERT enumerates an explicit inert rendering contract: HTML-escaped, ANSI surfaced as visible escaped glyphs, bidi/RTL controls replaced with visible placeholders, zero-width and confusable characters surfaced as visible placeholders, control characters rendered as escape sequences, whitespace preserved, no click handlers, no execution, read-only typography. T-CONSENT-INERT-ANSI / T-CONSENT-INERT-BIDI / T-CONSENT-INERT-ZWSP enforce this in tests. |
+| S2 | An attacker tricks a user into trusting a config from the settings row without ever showing the user the prompt + commands. | B-TRUST-UPGRADE-DISCLOSURE requires a modal Trust-review dialog presenting the SAME B-CONSENT-CONTENT verbatim payload (rendered inert per B-CONSENT-INERT) before any sidecar write from the settings path. B-TRUST-UPGRADE-NO-BYPASS forbids ANY other code path from writing `"trusted"` to the sidecar without first showing one of the two qualifying disclosure surfaces (consent card or Trust-review modal) in the same atomic UI interaction. |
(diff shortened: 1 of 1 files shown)


dhruvac29:
Thanks for writing this up. A few details seem worth aligning with the issue thread before this spec lands:

1. The issue thread now has maintainer guidance that tab configs should reference profiles by display name, and if multiple profiles share the same display name Warp should error instead of choosing one. Could we add that to the behavior contract, acceptance criteria, and test plan?

2. B3 says omitting `profile` preserves today’s behavior, but the parenthetical says “default profile.” The original issue says today’s behavior is whichever profile was last used. It would be good to clarify whether omitted `profile` truly preserves the current last-used behavior or intentionally switches to default.

3. Since this is spec-only, I’d avoid a closing keyword for #10171 unless maintainers want the implementation issue closed by the spec PR. `Refs #10171` may be safer than `Closes ... #10171`.

lonexreb:
Round-1 fixes for oz-for-oss review: (1) added Trust & Consent Model — imported/shared configs require inline confirmation before agent dispatch; user-saved configs are pre-trusted; (2) defined missing-profile behavior with Disabled-Agent state holding initial input AND commands; (3) resolved duplicate-name lookup via (name, source) pairs and qualified `source:name` syntax; (4) extended confirmation gate to cover all open-time agent inputs (initial prompt, commands array, etc.); (5) added explicit ordering test for profile-bind-before-dispatch; (6) updated acceptance criteria to cover ambiguous-name, missing-profile, imported vs owned config paths.

lonexreb:
Round-2 fixes: (1) **CRITICAL resolved** — single-source name uniqueness is now an enforced invariant (B-IDENTITY); the profile store rejects create/import/rename collisions. Only cross-source ambiguity remains and is resolved deterministically by precedence (user > team > builtin); (2) trust marker is LOCAL-ONLY in `~/.config/warp/tab_config_trust.json`, keyed by `(canonical_path_hash, content_hash)`; imported files always default to `imported`; in-file trust markers are stripped on import (B-TRUST-STRIP); (3) consent card shows exact verbatim inputs (full prompt text + commands array + all open-time agent inputs); (4) Cancel rolls back import (no input retention, no draft); 'Open with default profile' clears all agent inputs to prevent piggyback elevation; (5) `source:name` parser splits on first unescaped `:`; backslash-escape rules defined; tests T_qualified_name_with_colon and T_qualified_name_with_backslash added.

lonexreb:
Round-3 review fixes (commit 9c587af1):

- **Trust-upgrade-from-settings disclosure (B-TRUST-UPGRADE-DISCLOSURE)**: clicking 'Trust this tab config' from the settings row now requires a Trust-review modal with the SAME verbatim B-CONSENT-CONTENT (profile, autonomy, prompt, commands, other inputs) rendered inert per B-CONSENT-INERT. Sidecar writes from settings are forbidden without that gate. Closes the bypass where settings could promote a malicious imported config without ever showing its open-time inputs.
- **Multi-pane consent / cancel (B-CONSENT-MULTIPANE, B-CANCEL-MULTIPANE, B-CANCEL-DEFAULT-MULTIPANE, B-CONSENT-MULTIPANE-PROMOTE)**: per-pane consent cards in source order with 'Pane K of N' header; ALL agent processes are gated until every consent-gated pane resolves; `[Cancel]` on any card rolls back the entire tab open (no partial opens); `[Open with default profile]` is per-pane; the 'Trust this tab config' checkbox is shared across panes and the sidecar write is deferred until all panes resolve with `[Open with profile]`.
- **Inert rendering contract (B-CONSENT-INERT)**: every verbatim field on the consent card and the Trust-review modal is rendered inert: no execution, no live links, ANSI escapes (incl. OSC 8 hyperlinks) surfaced as visible escaped glyphs, bidi/RTL-override controls (U+202A–U+202E, U+2066–U+2069) replaced with visible placeholders, zero-width / confusable characters (ZWJ, ZWNJ, ZWSP) flagged, control characters escaped, read-only typography. Tests T-CONSENT-INERT-ANSI / -BIDI / -ZWSP cover the spoof vectors.
- **Imported-until-trusted reconciled (B-CONSENT-EVERY-OPEN)**: the prior 'session-scoped consent' wording is replaced. Imported configs require consent on EVERY open (same session, new session, multi-pane open) — there is NO per-session memory of prior consent. Only sidecar trust per B-TRUST-UPGRADE persists; this resolves the conflict with the imported-until-trusted model.

lonexreb:
Round-4 review addressed: tightened B-TRUST-UPGRADE to enumerate the only two trust-write surfaces (consent-card checkbox + settings modal) and explicitly forbid all other sidecar write paths. The settings surface now mandates B-TRUST-UPGRADE-DISCLOSURE as a precondition. Multi-pane consent/cancel (B-CONSENT-MULTIPANE, B-CANCEL-MULTIPANE), inert/escaped rendering (B-CONSENT-INERT), and session-scope reconciliation (B-CONSENT-EVERY-OPEN) remain in place from prior rounds.

Issue #10031, opened 2026-05-04
When I create a worktree with `claude --worktree`, it doesn’t update the metadata in Warp, such as the branch, PR, and changes


matheuspimentaa:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [x] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [x] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When I create a new worktree using Claude Code with the command `claude --worktree`, it creates a worktree in the directory `<project>/.claude/worktree/<worktree>`. This configuration cannot be changed — it’s a default behavior of Claude Code.

Automatically, this creates a worktree directory and a new branch with the same name. However, the metadata is not updated for the current worktree as shown in Warp, such as:

* Branch name
* PR
* View changes

Everything remains the same as in the project root, which would be `<project>/`.

This needs to be fixed, as I use `claude --worktree` a lot, especially to quickly create disposable worktrees.

### To reproduce

1. I enter the project `<project>/`
2. In the terminal, I run the command `claude --worktree <worktree-name>`
3. It opens Claude Code inside the worktree at the path `<project>/.claude/worktree/<worktree-name>`

### Operating system (OS)

Windows

### Operating system and version

11

### Shell Version

PowerShell 7

### Current Warp version

v0.2026.04.29.08.57.stable_01

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

Yes, this issue prevents me from using Warp daily.

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

peicodes:
It's currently uncertain how we can or should implement this. More research is needed. We would love to see a proposal in the form of a spec.

goli4th-1:
[...]

PR #10230, opened 2026-05-06
spec: refresh repo metadata after claude --worktree (#10031)

lonexreb:
Spec for #10031. CWD-tracking hook gains a 'repo-root changed' signal driven by (toplevel, git-common-dir) tuple change. Existing repo-metadata subsystem refreshes on event. Throttled to 2/sec/pane. Clears metadata when CWD leaves a git repo.

Closes (spec-only) #10031.

### DIFF
--- a/specs/GH10031/SPEC.md
+++ b/specs/GH10031/SPEC.md
@@ -1,0 +1,79 @@
+# Spec: Refresh repo metadata after `claude --worktree` (GH-10031)
+
+## Problem
+
+When the user runs `claude --worktree`, it creates a new git
+worktree and the user's CWD inside the running pane changes.
+But Warp's per-pane repo metadata (branch label, PR pill, diff
+chip) does not refresh — it keeps showing the old worktree's
+metadata until the user manually triggers a refresh (e.g., by
+focusing another tab and back).
+
+## Goal
+
+Detect CWD changes within a single pane and re-derive repo
+metadata when the pane's CWD lands inside a different git repo
+or a different git worktree.
+
+## Behavior contract
+
+- B1. The terminal's existing CWD-tracking hook (already used to
+  feed prompt PWD into Warp) gains a "repo-root changed"
+  derived signal: re-run `git rev-parse --show-toplevel` and
+  `git rev-parse --git-common-dir` (the latter resolves to the
+  shared `.git` directory across worktrees, identifying the
+  worktree group).
+- B2. When (`toplevel`, `git-common-dir`) tuple changes, fire a
+  `RepoContextChanged` event that the existing repo-metadata
+  subsystem already responds to. (It currently fires only on
+  pane creation and tab switch.)
+- B3. Throttle the re-derivation: at most once per 500ms per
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) single `git rev-parse --show-toplevel --git-common-dir` invocation reconciles throttle with derivation; (2) tuple comparison now uses canonical (symlink-resolved, absolute) paths to avoid false positives on subdir/symlink moves; (3) throttle is leading-edge with mandatory trailing-edge re-derivation to capture final CWD after bursts; (4) Non-Git Clear Path now fires explicit `RepoMetadataCleared` event resetting the FULL metadata surface (branch, ahead/behind, PR, diff cache, file-tree overlays, status bar, open panels).

lonexreb:
Round-2 fixes: (1) every `git rev-parse` invocation MUST run with explicit CWD (B-BASEDIR); relative outputs are resolved against that same CWD before canonicalization (deterministic key); (2) non-git clear now invalidates cached tuple (B5-INVALIDATE) — next CWD update re-runs `git rev-parse` even if returning to a previously-tracked repo (B5-RETURN); added T_return_to_repo_after_leaving; (3) throttle budget tightened — 250ms window, sustained max 4 derivations/sec/pane, leading + at-most-one-trailing per window, no cross-window amortization; T_throttle_burst asserts 10-CWD-changes-in-100ms produces ≤2 derivations.

lonexreb:
Round-3 review fixes (commit 5a21bd00):

- **Stale-result handling (B-ORDER, B-ORDER-1..4)**: each derivation now carries a monotonically increasing per-pane `derivation_seq`; the coordinator tracks `last_applied_seq` and DISCARDS any result whose `seq <= last_applied_seq`. Tuple cache and event firing both gate on this comparison, so when a slow leading-edge `git rev-parse` finishes after the trailing-edge derivation has already applied the latest CWD, the leading result is silently dropped instead of overwriting the metadata. Cancellation is allowed but not required for correctness; `derivation_seq` is monotonic across pane lifetime so a late leading result cannot be re-applied across a clear (T_stale_after_clear).
- **Throttle-budget contradiction resolved (B3-budget)**: replaced the conflicting wording. Single canonical rule: at most 2 derivations per 250ms window per pane (1 leading + 1 trailing); per-window peak rate = 8/s; sustained throughput across continuously busy windows averages to 4 trailing-derivations/s/pane because subsequent leading edges are short-circuited by 'skip if CWD unchanged'. Implementations MUST NOT amortize trailing edges across windows.
- Added A-STALE-REJECT acceptance criterion + T_stale_leading_overrun and T_stale_after_clear tests covering exit-order interleavings.

lonexreb:
Round-4 review addressed: rewrote B3-budget to clearly distinguish PEAK (8/s, per-window worst case) vs STEADY-STATE (4/s, long-run under churn) and explicitly noted the per-window cap of 2 is the ONLY normative bound — the two numbers are derived informatively, not contradictory. Stale-result handling (B-ORDER + tests T_stale_leading_overrun, T_stale_after_clear) was already in place from round-3.

PR #10238, opened 2026-05-06
Fix race condition causing permanent loss of active rule files

szgupta:
## Description

Fix a race condition in `ProjectContextModel's file watcher stream handler that can permanently lose active rule files until client restart.

### Root cause

The stream handler processes each `RepositoryUpdate` by spawning an async task. The original code called `path_to_rules.remove()` before spawning, then re-inserted the updated rules in the task's completion callback. This created two distinct race windows:

**Race 1 — concurrent task overwrites (original bug)**

Consider two updates, U1 and U2, arriving in quick succession for the same path:

1. U1 arrives → `remove()` extracts the current rules → async task T1 spawned with a snapshot of those rules.
2. U2 arrives while T1 is in flight → `remove()` returns `None` (the key is gone) → the `if let Some(rules) = existing_rules` guard fires → **U2 is silently dropped**.
3. T1 completes and re-inserts its result. If U1 was a deletion and U2 was the corresponding re-addition (common during an editor's atomic save: delete-then-write), the file is now permanently absent from `path_to_rules` until restart.

**Race 2 — stale-snapshot overwrites (Oz review finding)**

Even after fixing Race 1 by keeping rules in the map during async processing, a second race remained: if T1 and T2 both read the rules _before_ either writes back, T2's write silently discards T1's changes.

### Fix

**For Race 1:** Replace `remove()` with `get().cloned()` so the rules stay in `path_to_rules` throughout async processing. A `pending_updates` queue (keyed by path) is introduced: if a task is already in flight for a path, incoming updates are queued instead of spawned concurrently. When a task completes, it drains the queue — processing all accumulated updates sequentially against the freshest rules — before clearing the in-flight marker.

**For Race 2:** The drain loop combines multiple sequential deltas using a new `RulesDelta::merge()` method that is order-preserving: for each path, the _last_ operation wins. This correctly handles (add → delete) → net deletion and (delete → add) → net addition. The previous `deduplicate()` was symmetric — it cancelled both sides regardless of order — which would silently drop real state changes observed by persistence consumers.

### Changes

- `ProjectContextModel`: adds `pending_updates: HashMap<PathBuf, Vec<RepositoryUpdate>>` to serialize per-path processing.
- `RulesDelta::merge()`: replaces `deduplicate()` with an order-preserving combinator.
- `apply_update_result()`: extracted helper shared by the stream handler and the drain loop.
- Unit tests for all `merge()` orderings (add→delete, delete→add, add→delete→add, etc.).

## Linked Issue

N/A - discovered via debugging

## Testing

- Manual verification of the race condition scenario.
- Unit tests covering all `RulesDelta::merge()` orderings in `model_tests.rs`.

## Agent Mode

[Conversation link](https://staging.warp.dev/conversation/6cf1e2ca-9c21-4a00-9751-5fc9d7136473)

### DIFF
--- a/crates/ai/src/project_context/model.rs
+++ b/crates/ai/src/project_context/model.rs
@@ -183,6 +183,35 @@ pub struct RulesDelta {
     pub deleted_rules: Vec<PathBuf>,
 }
 
+impl RulesDelta {
+    /// Merge another delta into this one, preserving the ordering of operations.
+    ///
+    /// When the same path appears across sequential deltas the *last* operation
+    /// wins. For example:
+    ///   - (add A, delete A) → net effect is **delete**
+    ///   - (delete A, add A) → net effect is **add**
+    ///
+    /// This is important because consumers (e.g. persistence) apply the delta
+    /// incrementally; a symmetric "cancel both sides" approach would silently
+    /// drop real state changes.
+    fn merge(&mut self, other: RulesDelta) {
+        // Each newly-discovered path supersedes any prior deletion or earlier
+        // discovery of the same path.
+        for discovered in &other.discovered_rules {
+            self.deleted_rules.retain(|p| *p != discovered.path);
+            self.discovered_rules.retain(|r| r.path != discovered.path);
+        }
+        // Each newly-deleted path supersedes any prior discovery or earlier
+        // deletion of the same path.
+        for deleted in &other.deleted_rules {
+            self.discovered_rules.retain(|r| r.path != *deleted);
+            self.deleted_rules.retain(|p| *p != *deleted);
+        }
+        self.discovered_rules.extend(other.discovered_rules);
+        self.deleted_rules.extend(other.deleted_rules);
+    }
+}
+
 /// Events emitted by the ProjectContextModel
 pub enum ProjectContextModelEvent {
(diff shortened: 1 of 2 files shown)


crates/ai/src/project_context/model.rs
```diff
@@ -345,27 +350,85 @@ impl ProjectContextModel {
...
+        ctx.spawn(
+            async move {
+                let mut current_rules = rules;
+                let mut combined_delta = RulesDelta::default();
+                for update in updates {
+                    let (updated_rules, delta) =
+                        Self::process_repository_updates(update, current_rules, repo_path.clone())
+                            .await;
+                    current_rules = updated_rules;
+                    combined_delta
```
cephalonaut: Looks good overall, nice find! I wonder if there's a little potential for trouble in the fact that we break updates into discovered and deleted rules? This loses the ordering of those updates so an (add, delete) of the same rule ends up looking just like a (delete, add) in the emitted event. Might be better before adding to either list to dedupe?

cephalonaut (approved):

crates/ai/src/project_context/model.rs
szgupta: Fixed

PR #10241, opened 2026-05-06
Fix race condition causing requested commands to be auto-cancelled.

szgupta:
## Description

When two agent conversations run in parallel in different terminal views, `set_active_conversation_id` was called during automatic follow-ups (tool call → result cycles). This function has **transfer semantics**: it removes the conversation from every *other* terminal view's live list. When two views both trigger follow-ups concurrently, each view's follow-up could steal the other view's conversation, causing requested commands to be silently dropped with:

```
No conversation ID found for requested command
```

### Root cause

`set_active_conversation_id` was designed for **explicit user navigation** (opening a conversation in a different pane from the command palette), but was also being called in two automatic code paths:

1. `send_follow_up_for_conversation` (controller.rs) — after tool call results are sent back
2. `send_request_input` (controller.rs) — when sending any non-passive request

These paths only need to update the "most recently streamed" pointer within their own view. They should never touch other views.

### Fix

**`history_model.rs`**: Added `mark_active_conversation_id` — a non-transferring variant that updates the active conversation pointer for a single terminal view without removing the conversation from other views' live lists. The existing `set_active_conversation_id` (with transfer semantics) is preserved for explicit navigation call sites.

**`controller.rs`**: Switched both follow-up call sites to use `mark_active_conversation_id` instead of `set_active_conversation_id`.

**`history_model.rs`** (`conversation_id_for_action`): Added a global fallback search across all in-memory conversations when the per-view lookup fails. This is a safety net — if a conversation is ever removed from a view's live list while actions are in flight, the command still executes instead of being silently dropped.

## Linked Issue

N/A — discovered via internal debugging of intermittent command cancellation during parallel agent conversations.

## Testing

Manual testing with parallel agent conversations in split panes. The race condition is timing-dependent and not consistently reproducible, but the fix is structurally sound: the follow-up paths no longer call the transferring variant, so the race is eliminated by design.

## Agent Mode

[Conversation](https://staging.warp.dev/conversation/ff1292cc-8e8e-4c53-9f5d-6fcfa23089aa)

### DIFF
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -1384,9 +1384,18 @@ impl BlocklistAIController {
             return;
         }
 
-        BlocklistAIHistoryModel::handle(ctx).update(ctx, |history, ctx| {
-            history.set_active_conversation_id(conversation_id, self.terminal_view_id, ctx);
-        });
+        // Only update the active conversation pointer if this view still owns
+        // the conversation. Without this guard, the transfer loop inside
+        // set_active_conversation_id can steal the conversation from the view
+        // that is actually processing it when two conversations race.
+        if BlocklistAIHistoryModel::as_ref(ctx)
+            .all_live_conversations_for_terminal_view(self.terminal_view_id)
+            .any(|c| c.id() == conversation_id)
+        {
+            BlocklistAIHistoryModel::handle(ctx).update(ctx, |history, ctx| {
+                history.set_active_conversation_id(conversation_id, self.terminal_view_id, ctx);
+            });
+        }
 
         if !FeatureFlag::AgentView.is_enabled() && trigger == FollowUpTrigger::Auto {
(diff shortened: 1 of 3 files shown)


szgupta:
@zachbai this is a short-term fix. Real fix is consolidating these terminal view <> convo ID maps and their access patterns. It's a big mess right now and will require a good amount of work to avoid regressing stuff. So this is an in-between to get the halting problem fixed.

zachbai (approved):

zachbai:
> @zachbai this is a short-term fix. Real fix is consolidating these terminal view <> convo ID maps and their access patterns. It's a big mess right now and will require a good amount of work to avoid regressing stuff. So this is an in-between to get the halting problem fixed.

We should be able to get rid of the `active_conversation_id` map. It's redundant with the active conversation_id in the `AgentViewController`, which is more correctly the source of truth for "conversation actively displayed in this terminal pane"

PR #10293, opened 2026-05-06
Style rich text editor newlines as line breaks, not paragraph breaks

alokedesai:
## Description
The rich text editor would treat each newline as a paragraph break instead of line break, causing the user-visible line height to look unusually large.

This PR changes the styling of the rich text editor to more closely mimic a normal markdown editor instead of an editor like Notion.

Specifically:
1) I changed the margin of text blocks to be `0px`.  There's now no difference in padding between hard-wrapped and soft-wrapped lines.
2) I reduced the padding of some of the other block styles (including headers)
3) I removed the `post_process` notebook step that strips newlines. This processing step was just wrong, the filesystem should be the source of truth.  Removing newlines can change how a document reads, for ex: `foo\n\n\nbar` is different from `foo\nbar`.

I considered fixing this deeper in the stack by not creating a new paragraph block when a user creates a newline after an existing paragraph block. That ended up being a much more complicated change than I intended, so opted for this now given priorities. Happy to keep on investigating that if you feel strongly we should go down that approach.

## Testing
I did a bunch of monkey testing with AI-generated markdown. See screenshots and looms below

Left is before, right is after:
image

**Loom**
https://www.loom.com/share/7d90d47720314eaa9aba7040935b8d51?from_recorder=1&focus_title=1


alokedesai:
* **#10295** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10295?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10294** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10294?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10293** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10293?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10293?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/code/editor/comment_editor.rs
```diff
@@ -423,8 +423,8 @@ impl View for CommentEditor {
                     )
                     .with_child(
                         Container::new(footer_row)
-                            .with_vertical_padding(8.)
-                            .with_horizontal_padding(8.)
+                            .with_vertical_padding(4.)
```
alokedesai, line 427: This just reduces the padding of the comment editor, which looks much better now that we've reduced the padding of the editor

Before:

Image

After:

Image

app/src/notebooks/editor/mod.rs
```diff
@@ -40,9 +44,7 @@ mod omnibar;
 pub mod view;
 
 pub use block_insertion_menu::BlockInsertionSource;
-use warpui::elements::ListIndentLevel;
-
-const NOTEBOOK_LINE_HEIGHT_RATIO: f32 = 1.6;
+const NOTEBOOK_LINE_HEIGHT_RATIO: f32 = 1.5;
```
alokedesai, line 42: Bumped this down just a tad to overall reduce the size of any given line

crates/editor/src/render/model/mod.rs
```diff
@@ -107,7 +107,7 @@ const TABLE_SCROLL_REVEAL_MARGIN: Pixels = Pixels::new(8.);
 pub const EMBEDDED_ITEM_FIRST_LINE_HEIGHT: f32 = 24.;
 
 pub const TEXT_SPACING: BlockSpacing = BlockSpacing {
-    margin: Margin::uniform(4.).with_right(16.),
+    margin: Margin::uniform(0.).with_right(16.),
```
alokedesai, line 93: This is the key change. By setting the margin to `0px`, we just use line height. 

The downside is there may be slightly less margin between different a non-text block and a text block but I've bashed this pretty extensively and I don't see any visual issues.

crates/editor/src/render/model/mod.rs
```diff
@@ -139,8 +139,8 @@ pub const BROKEN_LINK_SPACING: BlockSpacing = BlockSpacing {
 
 pub const HEADER_SPACING: BlockSpacing = BlockSpacing {
     margin: Margin::uniform(4.)
-        .with_top(12.)
-        .with_bottom(12.)
+        .with_top(4.)
```
alokedesai, line 125: I reduce this, the header spacing was way too big (I think the post_process step was obscuring this because we'd often remove newlines between a header and the actual text)

bnavetta (approved):
> I considered fixing this deeper in the stack by not creating a new paragraph block when a user creates a newline after an existing paragraph block. That ended up being a much more complicated change than I intended, so opted for this now given priorities. Happy to keep on investigating that if you feel strongly we should go down that approach.

I think the approach here, rather than conditionally folding paragraphs together, is safest. There are a _ton_ of assumptions in the editor around newline handling

PR #10300, opened 2026-05-06
Add remote backed global buffer

kevinyang372:
## Description
This adds the data modeling for the remote backed global buffer. It also implements the syncing protocol for client side and server side pushed changes

Full tech spec here: https://github.com/warpdotdev/warp/blob/cbb64c5fa2751a023580f80d8dac750f2b6a6bd5/specs/APP-3802/TECH.md

## Testing
This adds comprehensive unit test cases to verify the syncing behavior with a mocked server and client

See detailed discussion on how we construct the test cases in the tech spec


kevinyang372:
* **#10300** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10300?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10300?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/code/buffer_location.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Last version acknowledged from the server (file-watcher side).
+    pub server_version: ContentVersion,
+    /// Last version acknowledged from the client (user-edit side).
+    pub client_version: ContentVersion,
+}
+
+impl SyncClock {
+    pub fn new() -> Self {
+        Self {
+            server_version: ContentVersion::new(),
```
moirahuang: should we also start `server_version` at 0 to be explicit that it's the start?

app/src/code/buffer_location.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    /// Bump the server version after a file-watcher change.
+    pub fn bump_server(&mut self) -> ContentVersion {
+        self.server_version = ContentVersion::new();
+        self.server_version
+    }
+
+    /// Bump the client version after a local edit.
+    #[expect(dead_code)]
+    pub fn bump_client(&mut self) -> ContentVersion {
```
moirahuang: nit: let's just remove this and `client_edit_matches` below if we're not using it

app/src/code/global_buffer_model.rs
```diff
@@ -82,6 +201,22 @@ pub enum GlobalBufferModelEvent {
         file_id: FileId,
         error: Rc<FileSaveError>,
     },
+    /// A remote buffer update conflicted with local edits.
+    /// The UI should present a resolution dialog.
+    #[allow(dead_code)]
```
moirahuang: `RemoteBufferConflict` is used right? i think we can remove this

app/src/code/global_buffer_model.rs
```diff
@@ -143,15 +325,18 @@ impl GlobalBufferModel {
         // Collect paths for didClose before removing entries.
         let paths_to_close: Vec<PathBuf> = ids_to_remove
             .iter()
-            .filter_map(|id| self.path_to_id.get_by_right(id).cloned())
+            .filter_map(|id| match self.location_to_id.get_by_right(id) {
+                Some(BufferLocation::Local(path)) => Some(path.clone()),
+                _ => None,
```
moirahuang: let's be explicit here about all the cases so it's clear what the expected behavior is for each location

app/src/code/global_buffer_model.rs
```diff
@@ -1186,10 +1488,466 @@ impl GlobalBufferModel {
...
+            buffer.replace_all(client_content, ctx);
+            buffer.set_version(new_version);
+        });
+
+        // Save to disk.
+        let content = client_content.to_string();
+        let save_version = ContentVersion::new();
+        state.set_base_content_version(save_version);
+        FileModel::handle(ctx).update(ctx, |file_model, ctx| {
+            file_model.save(file_id, content, save_version, ctx)
```
moirahuang, line 1796: what happens if this save fails?

app/src/code/global_buffer_model.rs
```diff
@@ -1186,10 +1488,466 @@ impl GlobalBufferModel {
...
+    pub fn sync_clock_for_server_local(&self, file_id: FileId) -> Option<&SyncClock> {
+        let state = self.buffers.get(&file_id)?;
+        match &state.source {
+            BufferSource::ServerLocal { sync_clock, .. } => Some(sync_clock),
+            BufferSource::Local { .. } | BufferSource::Remote { .. } => None,
+        }
+    }
+
+    /// Returns whether a buffer is a `ServerLocal` source.
+    #[allow(dead_code)]
```
moirahuang: idt is dead code?

app/src/code/global_buffer_model.rs
```diff
@@ -1186,10 +1488,466 @@ impl GlobalBufferModel {
...
+        let client_for_sub = {
+            let manager = remote_server::manager::RemoteServerManager::handle(ctx);
+            manager.as_ref(ctx).client_for_host(&host_id).cloned()
+        };
+        if let Some(client) = &client_for_sub {
+            let client = client.clone();
+            let path_for_edit = path_str.clone();
+            ctx.subscribe_to_model(&buffer, move |me, event, ctx| {
+                use warp_editor::content::buffer::BufferEvent;
+                if let BufferEvent::ContentChanged { delta, .. } = event {
```
moirahuang: do we need to differentiate here server originated vs. user originated buffer change events? otherwise something like `BufferUpdatedPush` will trigger another trip right?

app/src/remote_server/server_model.rs
```diff
@@ -1159,6 +1311,153 @@ impl ServerModel {
...
+        // Not yet loaded — stash request info so the GlobalBufferModelEvent
+        // subscription can send the response when content arrives.
+        self.buffers
+            .insert_pending(file_id, request_id.clone(), conn_id);
+        HandlerOutcome::Async(None)
+    }
+
+    /// Handles `BufferEdit` notification (fire-and-forget).
+    /// Delegates to `GlobalBufferModel::apply_client_edit`. On rejection
+    /// (stale server version), pushes the server's current state back to
```
moirahuang: this differs from the comment at line 1378, are we silently dropping when an edit is rejected or do we push the server's current state back to all connections?

app/src/code/global_buffer_model.rs
```diff
@@ -1186,10 +1488,466 @@ impl GlobalBufferModel {
...
+        acknowledged_server_version: ContentVersion,
+        client_content: &str,
+        ctx: &mut ModelContext<Self>,
+    ) -> Result<(), FileSaveError> {
+        let Some(state) = self.buffers.get_mut(&file_id) else {
+            return Err(FileSaveError::RemoteError("Buffer not found".to_string()));
+        };
+
+        if let BufferSource::ServerLocal { sync_clock, .. } = &mut state.source {
+            sync_clock.server_version = acknowledged_server_version;
```
moirahuang, line 1770: should we also update the client version?

specs/APP-3802/TECH.md
```diff
@@ -0,0 +1,158 @@
...
+The `is_loaded()` check differs by variant: `Local`/`ServerLocal` use `base_content_version.is_some()`, while `Remote` uses `sync_clock.is_some()` to distinguish the "waiting for OpenBufferResponse" state from a loaded buffer.
+
+### 2. Syncing protocol and conflict handling
+
+The protocol uses a two-component version vector (`SyncClock`) where each side owns one counter:
+
+- **server_version** (S) — bumped by the daemon when the file changes on disk.
+- **client_version** (C) — bumped by the client when the user edits the buffer.
+
+All edits on the wire use 0-indexed character offsets (`TextEdit { start_offset, end_offset, text }`), avoiding line/column conversion overhead. The buffer internally uses 1-indexed `CharOffset`; the +1/−1 conversion happens at the boundary.
```
moirahuang: what does "avoiding line/column conversion overhead" mean? i'm a little worried this conversion is brittle and can result in many one off errors

moirahuang (commented):
some smaller comments; i think the tech spec flow makes sense but i'm a bit worried about code edge cases

app/src/code/global_buffer_model.rs
kevinyang372: I changed this to a request type instead of fire-and-forget

This allows the client to show and act on the failures

specs/APP-3802/TECH.md
kevinyang372: This was from a leftover previous change. Tho I do agree we don't need to do the 0-index / 1-index dance when it is a syncing protocol internal to buffers. I changed it so everything is 1-indexed

app/src/code/global_buffer_model.rs
```diff
@@ -46,19 +51,133 @@ struct PendingDiffParse {
...
-    /// when the file is loaded in the app. We track this version so we don't submit
-    /// a document updated event to LSP when the file is first loaded. (We do need to
-    /// submit document updated if the file is auto-reloaded).
-    initial_content_version: Option<ContentVersion>,
     /// Tracks the latest buffer version we've attempted to sync with LSP.
     /// Used to detect if previous versions were synced successfully.
     latest_buffer_version: Option<usize>,
     /// Tracks any active background diff parsing for auto-reload.
     pending_diff_parse: Option<PendingDiffParse>,
+    source: BufferSource,
```
moirahuang, line 87: shoudl `BufferSource` maybe live on `Buffer`?

app/src/code/global_buffer_model.rs
```diff
@@ -1186,10 +1486,468 @@ impl GlobalBufferModel {
...
+    ///
+    /// If `expected_server_version` matches the buffer's current server version,
+    /// the edits are applied to the in-memory buffer (no disk write) and the
+    /// client version is updated. Returns `true` if accepted, `false` if rejected
+    /// (stale edit — silently discarded).
+    #[cfg(feature = "local_fs")]
+    pub fn apply_client_edit(
+        &mut self,
+        file_id: FileId,
+        edits: &[super::super::remote_server::proto::TextEdit],
```
moirahuang, line 1685: nit: topline import?

app/src/code/global_buffer_model.rs
```diff
@@ -1186,10 +1486,468 @@ impl GlobalBufferModel {
...
+    ) -> Result<(), FileSaveError> {
+        let Some(state) = self.buffers.get(&file_id) else {
+            return Err(FileSaveError::RemoteError("Buffer not found".to_string()));
+        };
+        let Some(buffer) = state.buffer.upgrade(ctx) else {
+            return Err(FileSaveError::RemoteError("Buffer deallocated".to_string()));
+        };
+        let content = buffer.as_ref(ctx).text().into_string();
+        let version = ContentVersion::new();
+        FileModel::handle(ctx).update(ctx, |file_model, ctx| {
```
moirahuang, line 1749: nit: maybe let's add a comment similar to 1787 here

app/src/code/global_buffer_model.rs
```diff
@@ -1186,10 +1486,468 @@ impl GlobalBufferModel {
...
+                let char_edits: Vec<(std::ops::Range<CharOffset>, String)> = edits
+                    .iter()
+                    .map(|edit| {
+                        let start = std::cmp::min(edit.start, max_offset);
+                        let end = std::cmp::min(edit.end, max_offset);
+                        (start..end, edit.text.clone())
+                    })
+                    .collect();
+
+                buffer.insert_at_char_offset_ranges(char_edits, new_version, ctx);
```
moirahuang, line 1888: maybe not worth it but i think we can pull out some shared logic here w line 1728ish

app/src/remote_server/server_buffer_tracker.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// Maps wire path strings to `FileId` for open server-local buffers.
+    open_buffers: HashMap<String, FileId>,
+    /// Tracks which connections have each buffer open.
+    /// File-watcher pushes go to all connections in the set.
+    buffer_connections: HashMap<FileId, HashSet<ConnectionId>>,
+    /// Tracks in-flight OpenBuffer / SaveBuffer / ResolveConflict requests so
+    /// `GlobalBufferModelEvent`s can be correlated back to the originating
+    /// request and connection. Uses a `Vec` to support concurrent requests
+    /// for the same buffer from different connections.
+    pending_requests: HashMap<FileId, Vec<(RequestId, ConnectionId, PendingBufferRequestKind)>>,
```
moirahuang, line 34: can we maybe put this in an enum

app/src/code/global_buffer_model.rs
moirahuang: i think i'm still a little worried here. if we fail to save, it sounds like we're stuck until the next reconciliation. is there any risk if there being a big delay before the next reconciliation or continued failures to save?

moirahuang (approved):
approving + going to suggest we synchronously talk about the failed to sync case

app/src/code/global_buffer_model.rs
kevinyang372: This semantically doesn't belong to Buffer tho. Buffer should mainly just care about the content it holds

app/src/code/global_buffer_model.rs
kevinyang372: I actually think the inline import is valuable here since it signifies this is using the proto TextEdit type. Tho agreed super::super::remote-server is a bit excessive

PR #10310, opened 2026-05-06
Add platform credits counting to client

jefflloyd:
## Description

For certain teams, the server may report that an AI query cost platform credits; this just makes the client sum up both inference + platform credits so the cost totals shown in the UI are accurate.

## Testing

Verified that the total credits being reported matched the sum of inference + AI credits in the backend.

## Agent Mode

### DIFF
--- a/app/src/ai/agent/api/convert_to.rs
+++ b/app/src/ai/agent/api/convert_to.rs
@@ -808,6 +808,8 @@ fn convert_context(context: &[AIAgentContext]) -> api::InputContext {
                 api_context.git = Some(api::input_context::Git {
                     head,
                     branch: branch.unwrap_or_default(),
+                    repository: None,   // TODO: populate?
+                    pull_request: None, // TODO: populate?
                 });
             }
             AIAgentContext::Skills { skills } => {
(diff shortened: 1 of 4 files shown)


app/src/ai/agent/api/convert_to.rs
```diff
@@ -808,6 +808,8 @@ fn convert_context(context: &[AIAgentContext]) -> api::InputContext {
                 api_context.git = Some(api::input_context::Git {
                     head,
                     branch: branch.unwrap_or_default(),
+                    repository: None, // TODO: populate?
+                    pull_request: None, // TODO: populate?
```
jefflloyd: @yuanben-warp just a heads up that I had to add these fields once I brought in the new version of the protos - think they're related to your work, but leaving them None for now!

yuanben-warp: Got it, thanks I'm working on the PR to fill them out, might take a sec to land it though!

jefflloyd: No worries/rush :)

yuanben-warp: I got the PR here: https://github.com/warpdotdev/warp-internal/pull/25423

IsaiahWitzke (approved):

PR #10315, opened 2026-05-06
Don't check server for new versions on wasm or during onboarding

seemeroland:
## Description

Don't check the server endpoint for new client versions on wasm, because wasm can't update anyway.

Also don't do this check during onboarding because we can't update at that time

## Testing

Tested manually

### DIFF
--- a/app/src/root_view.rs
+++ b/app/src/root_view.rs
@@ -1798,7 +1798,21 @@ impl RootView {
             ctx.focus(onboarding_view);
         }
 
+        // For users who bypass onboarding (already logged in, or onboarding flags not active),
+        // start autoupdate polling immediately. For new users in onboarding, this is a no-op;
+        // polling will be started once onboarding completes.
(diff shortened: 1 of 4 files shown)


app/src/root_view.rs
```diff
@@ -1802,9 +1798,23 @@ impl RootView {
             ctx.focus(onboarding_view);
         }
 
+        // For users who bypass onboarding (already logged in, or onboarding flags not active),
+        // start autoupdate polling immediately. For new users in onboarding, this is a no-op;
+        // polling will be started once onboarding completes.
+        root_view.start_autoupdate_polling(ctx);
```
jefflloyd, line 1804: OOC - why do we wait until after onboarding to autoupdate?

seemeroland: A not insignificant number of people abandon warp before ever finishing the onboarding. This will just decrease the rate our server gets hit at

jefflloyd (approved):

PR #10317, opened 2026-05-07
remote_server: fix bash-3.2 tilde expansion in install script

### LINKED ISSUE #10316: remote-server SSH wrapper install fails on macOS due to bash-3.2 tilde expansion
### Pre-submit Checks

- [x] I have included the logs

### Describe the bug

Warp's remote-server SSH wrapper (the new agent-based one, not the legacy `warpify`/tmux wrapper) fails to install its `oz` agent binary on any remote macOS host where `bash` resolves to Apple's stock `/bin/bash` (version 3.2.57 — i.e. every macOS host that hasn't put homebrew bash on PATH for non-login shells, which is the default).

User-visible symptom in `~/Library/Logs/warp.log`:

```
[INFO]  Checking for remote server binary at ~/.warp/remote-server/oz-v0.2026.04.29.08.57.stable_01
[INFO]  Installing remote server binary to ~/.warp/remote-server/oz-v0.2026.04.29.08.57.stable_01
[INFO]  Starting remote server connection for session SessionId(...)
[INFO]  [remote_server] zsh:1: no such file or directory: /Users/<user>/.warp/remote-server/oz-v0.2026.04.29.08.57.stable_01
[INFO]  Reader task: server disconnected (EOF)
[ERROR] Connection failed for session SessionId(...): initialize: Response channel closed before receiving a reply
[WARN]  Remote server connection failed for session SessionId(...); flushing bootstrap to unblock SSH session
```

Warp falls back to ControlMaster/legacy mode and the user sees `Initialize: Response channel closed before receiving a reply` — a misleading symptom (the protocol channel didn't close prematurely; the agent simply doesn't exist).

### Root cause

`crates/remote_server/src/install_remote_server.sh:30`:

```bash
install_dir="${install_dir/#\~/"$HOME"}"
```

`run_ssh_script` in `crates/remote_server/src/ssh.rs` invokes the install script with `ssh ... bash -s` and pipes the script body into stdin. On macOS the resolved `bash` is `/bin/bash` (3.2.57 — Apple has not shipped a newer bash since the GPLv3 transition).

In bash 3.2, the `"$HOME"` *inside the replacement* of `${var/pattern/replacement}` is treated as **6 literal characters including the double-quotes**, not expanded. Bash 4+ strips them. Empirical confirmation:

```
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

$ HOME=/Users/me /bin/bash -c 'x=~/foo; echo "${x/#\~/"$HOME"}"'
"/Users/me"/foo                      # ← literal quotes

$ HOME=/Users/me /bin/bash -c 'x=~/foo; echo "${x/#\~/$HOME}"'
/Users/me/foo                        # ← correct
```

Downstream consequences in the install script:

- `mkdir -p '"/Users/<user>"/.warp/remote-server'` succeeds, creating a directory tree under `$HOME` whose first component is literally `"`.
- The 278 MB agent tarball downloads and extracts cleanly.
- `mv` lands the binary at `$HOME/"/Users/<user>"/.warp/remote-server/oz-v0...`.
- `install_binary()` returns `Ok(())`.
- `connect()` then asks zsh to exec `~/.warp/remote-server/oz-v0...` (the *correct* path), which doesn't exist → `zsh:1: no such file or directory`.

You can verify on any affected host: there will be a directory literally named `"` in the user's home folder containing a misplaced `oz-v0...` binary. On my Mac Studio that directory was dated to my very first connection attempt — the bug has been silent since the first install.

### To reproduce

1. From a Mac running Warp, `ssh` into another Mac (or any host where `bash -s` resolves to bash 3.2 — i.e. any macOS without `/opt/homebrew/bin/bash` shadowed via login shell config).
2. Wait for the remote-server bootstrap.
3. Observe the `initialize: Response channel closed before receiving a reply` error and the misplaced binary at `~/"/Users/<user>"/.warp/remote-server/oz-v0...` on the remote.

### Expected behavior

Install lands at `$HOME/.warp/remote-server/oz-v0...`; remote-server agent launches; full Warp UI works over SSH.

### Proposed fix

One character: drop the inner quotes around `$HOME` in the replacement. Bash 3.2 and 4+ both expand `${install_dir/#\~/$HOME}` correctly.

Happy to open the PR — I have it ready locally with two regression tests:

1. A live test that pipes the relevant snippet into `/bin/bash` and asserts the resolved path contains no literal `"` characters.
2. A static check over `INSTALL_SCRIPT_TEMPLATE` that grep-matches the buggy `/#\~/"$HOME"` pattern and explains why in the failure message.

### Operating system (OS)

macOS

macOS 26.2 (build 25C56), Darwin 25.2.0 arm64 (Mac Studio M2 Max). Reproduces against any Mac running default `/bin/bash` 3.2.57.

### Shell Version

Local: zsh 5.9. Remote `bash --version`: `GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)`.

### Current Warp version

v0.2026.04.29.08.57.stable_01

### Regression

### Additional context

- The connection works fine when Warp falls back to the legacy ControlMaster path. The bug only affects the new remote-server wrapper.
- Workaround for affected users: pre-place the agent at the correct path so `binary_check_command` (`test -x ~/.warp/remote-server/oz-v0...`) returns 0 and the broken install step is skipped:
  ```
  ssh <host> 'mv "\"/$HOME\"/.warp/remote-server/oz-v0..." ~/.warp/remote-server/oz-v0...'
  ```
- The bug doesn't happen on Linux remotes because Linux bash is 4+ (Apple is the outlier — they freeze /bin/bash at 3.2 to avoid GPLv3).
- Diagnostic timeline I went through, in case useful for triage: shell-init noise from missing brew tools (red herring — `bash -s` doesn't source rc files); URL guessing against `releases.warp.dev` (wrong base — actual URL is `app.warp.dev/download/cli`, served from GCS); `fs_usage`-style polling watcher confirmed zero filesystem writes in `~/.warp/remote-server/` during failure (red herring — writes were going to `~/"/Users/.../"...`); reading `crates/remote_server/src/install_remote_server.sh` and `ssh.rs:run_ssh_script` made the bash-3.2 incompatibility obvious.

No

lucasdoell:
## Description

Fixes the remote-server SSH wrapper install failure on macOS remotes whose `bash -s` resolves to Apple's stock `/bin/bash` (3.2.57). User-visible symptom is `Initialize: Response channel closed before receiving a reply`; underlying cause is the install script silently steering the agent binary into a directory tree literally named `"`.

`crates/remote_server/src/install_remote_server.sh` resolved a leading `~` with:

```bash
install_dir="${install_dir/#\~/"$HOME"}"
```

In bash 3.2, the inner double-quotes around `$HOME` are **not** stripped from the replacement — they're kept as literal characters. Bash 4+ strips them. `run_ssh_script` in `crates/remote_server/src/ssh.rs` invokes the install script via `ssh ... bash -s`, which on macOS remotes resolves to bash 3.2 (Apple has not shipped a newer bash since the GPLv3 transition). So `~/.warp/remote-server` expands to the *relative* path `"/Users/<user>"/.warp/remote-server`, `mkdir -p` and the rest of the script happily run against it, and the 278 MB agent binary lands at `$HOME/"/Users/<user>"/.warp/remote-server/oz-v0...`. The subsequent proxy launch in `app/src/remote_server/ssh_transport.rs` looks at the correct `$HOME/.warp/remote-server/oz-v0...` path, finds nothing, and reports the misleading "Response channel closed" error.

The one-character fix is to drop the inner quotes; bash 3.2 and 4+ both expand `${install_dir/#\~/$HOME}` correctly.

Direct repro on the affected interpreter:

```
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)

$ HOME=/Users/me /bin/bash -c 'x=~/foo; echo "${x/#\~/"$HOME"}"'
"/Users/me"/foo                # ← buggy

$ HOME=/Users/me /bin/bash -c 'x=~/foo; echo "${x/#\~/$HOME}"'
/Users/me/foo                  # ← fixed
```

Also adds a `# Pipe the script through bash 3.2 before changing this line.` comment so a future contributor doesn't reintroduce the quotes thinking they're harmless.

## Linked Issue

Fixes #10316.

  (Per CONTRIBUTING.md, triaged bug reports are implicitly `ready-to-implement`. Issue #10316 carries `bug`, `triaged`, `repro:high`, `os:mac`, `area:ssh`.)

## Testing

Two regression tests in `crates/remote_server/src/setup_tests.rs`:

1. **`install_script_tilde_expansion_works_in_bash_3_2`** — runs the relevant snippet under `/bin/bash` (or `bash` from PATH) with a fixed `HOME=/Users/test`, asserts the resolved `install_dir` is exactly `/Users/test/.warp/remote-server` and contains no literal `"` characters. This catches behavioural regressions on the bash version we actually invoke at install time.
2. **`install_script_does_not_quote_home_in_tilde_substitution`** — static grep over `INSTALL_SCRIPT_TEMPLATE` for the buggy `/#\~/"$HOME"` pattern. Catches the regression earlier than the live test and the failure message explains *why* (referencing bash 3.2 / macOS `/bin/bash`) so a future contributor doesn't have to rediscover the constraint.

Local `cargo test -p remote_server` is blocked on this Mac by an unrelated missing Metal Toolchain in `crates/warpui/build.rs` (`xcodebuild -downloadComponent MetalToolchain` would unblock it). The bash logic itself was verified manually against `/bin/bash 3.2.57` — see the repro above. CI will run the new tests cleanly.

I also confirmed the fix end-to-end on a real Mac→Mac SSH connection by pre-placing the binary at the correct path (so `binary_check_command` passes and the broken install is skipped); the agent launched and the Warp wrapper UI worked as expected. After this PR lands, that workaround should no longer be needed.

  (Verified via the equivalent end-to-end scenario described above; the `xcodebuild -downloadComponent MetalToolchain` step prevents me from running `./script/run` itself on this machine, but the bash-side change is the entire payload of the patch and was reproduced directly under `/bin/bash` 3.2.57.)

### Screenshots / Videos

Not applicable — no UI surface.

## Agent Mode

(Diagnosis was done with a Claude Code session against the open-source repo and live remote.)

CHANGELOG-BUG-FIX: Fix remote-server SSH wrapper failing to install the Oz agent on macOS remotes whose `bash -s` resolves to Apple bash 3.2 (the default), which surfaced as "Initialize: Response channel closed before receiving a reply" and a misplaced binary under `~/"/Users/.../"`.

### DIFF
--- a/crates/remote_server/src/install_remote_server.sh
+++ b/crates/remote_server/src/install_remote_server.sh
@@ -27,12 +27,20 @@ case "$os_kernel" in
 esac
 
 install_dir="{install_dir}"
-# Inner quotes around `$HOME` would be safe on bash 4+, but on macOS's
-# stock /bin/bash 3.2 they are taken as literal characters in the
-# replacement, producing a path like `"/Users/<user>"/.warp/...` and
-# silently steering the install into a directory tree literally named
-# `"`. Pipe the script through bash 3.2 before changing this line.
-install_dir="${install_dir/#\~/$HOME}"
+# Avoid `${var/pattern/replacement}` for tilde expansion. Two
+# interpreter quirks make it dangerous in this script:
+#   1. bash 3.2 (macOS /bin/bash) keeps inner double-quotes around the
+#      replacement literal, so `"$HOME"` ends up as 6 literal
+#      characters and the install lands under a directory tree
+#      literally named `"`.
+#   2. bash 5.2+ enables `patsub_replacement` by default, which makes
+#      `&` in the replacement expand to the matched pattern, so a
+#      `$HOME` containing `&` resolves to a `~`-substituted path.
+# Use `case` + `${var#\~}` instead — works on bash 3.2 and bash 5.2+
+# without surprises.
+case "$install_dir" in
+  "~"|"~/"*) install_dir="${HOME}${install_dir#\~}" ;;
+esac
 mkdir -p "$install_dir"
 
 tmpdir=$(mktemp -d "$install_dir/.install.XXXXXX")
@@ -50,8 +58,10 @@ trap cleanup EXIT
 staging_tarball_path="{staging_tarball_path}"
 if [ -n "$staging_tarball_path" ]; then
   # SCP fallback: tarball already uploaded by the client.
-  # Same bash 3.2 caveat as the install_dir tilde expansion above.
-  staging_tarball_path="${staging_tarball_path/#\~/$HOME}"
+  # Same tilde-expansion caveat as install_dir above.
+  case "$staging_tarball_path" in
+    "~"|"~/"*) staging_tarball_path="${HOME}${staging_tarball_path#\~}" ;;
+  esac
   mv "$staging_tarball_path" "$tmpdir/oz.tar.gz"
 else
   # Normal path: download via curl or wget.
--- a/crates/remote_server/src/install_remote_server.sh
+++ b/crates/remote_server/src/install_remote_server.sh
@@ -27,7 +27,12 @@ case "$os_kernel" in
 esac
 
 install_dir="{install_dir}"
-install_dir="${install_dir/#\~/"$HOME"}"
+# Inner quotes around `$HOME` would be safe on bash 4+, but on macOS's
+# stock /bin/bash 3.2 they are taken as literal characters in the
+# replacement, producing a path like `"/Users/<user>"/.warp/...` and
+# silently steering the install into a directory tree literally named
+# `"`. Pipe the script through bash 3.2 before changing this line.
+install_dir="${install_dir/#\~/$HOME}"
 mkdir -p "$install_dir"
 
 tmpdir=$(mktemp -d "$install_dir/.install.XXXXXX")
--- a/crates/remote_server/src/install_remote_server.sh
+++ b/crates/remote_server/src/install_remote_server.sh
@@ -50,7 +50,8 @@ trap cleanup EXIT
 staging_tarball_path="{staging_tarball_path}"
 if [ -n "$staging_tarball_path" ]; then
   # SCP fallback: tarball already uploaded by the client.
-  staging_tarball_path="${staging_tarball_path/#\~/"$HOME"}"
+  # Same bash 3.2 caveat as the install_dir tilde expansion above.
+  staging_tarball_path="${staging_tarball_path/#\~/$HOME}"
   mv "$staging_tarball_path" "$tmpdir/oz.tar.gz"
 else
   # Normal path: download via curl or wget.
--- a/crates/remote_server/src/setup_tests.rs
+++ b/crates/remote_server/src/setup_tests.rs
@@ -170,26 +170,31 @@ fn parse_preinstall_unsupported_non_glibc() {
     assert!(!result.is_supported());
 }
 
-/// Regression: the install script's tilde-expansion line must work in
-/// macOS's stock `/bin/bash` 3.2.57. In bash 3.2, `"$HOME"` inside the
-/// replacement of `${var/pattern/replacement}` is treated as 6 literal
-/// characters (the quotes are not stripped), which turned
-/// `~/.warp/remote-server` into the *relative* path
-/// `"/Users/<user>"/.warp/remote-server` and silently steered the
-/// install into a directory tree literally named `"`. The launch step
-/// then looked at the real `$HOME/.warp/remote-server/oz-...`, found
-/// nothing, and reported "no such file or directory" → "Response
-/// channel closed".
+/// Regression: the install script's tilde-expansion logic must work
+/// across the bash versions we actually invoke at install time
+/// (`run_ssh_script` pipes the script into `bash -s` on the remote).
+/// Two interpreter quirks have to be avoided simultaneously:
+///
+///   1. bash 3.2 (macOS `/bin/bash`) keeps inner double-quotes around
+///      the replacement of `${var/pattern/replacement}` literal, so
+///      `"$HOME"` ends up as 6 literal characters and the install
+///      lands under a directory tree literally named `"`.
+///   2. bash 5.2+ with `patsub_replacement` (default-on) treats `&`
+///      in the replacement as the matched pattern, so a `$HOME`
+///      containing `&` resolves to a `~`-substituted path.
+///
+/// Both bugs surface as the install binary landing somewhere Warp's
+/// launch step doesn't look, producing a misleading "Response channel
+/// closed before receiving a reply".
 ///
 /// This test drives the *actual* production script (via
-/// [`install_script`]) rather than a hand-copied snippet so a future
-/// change that reintroduces the quoted form can't slip past by being
-/// dissimilar enough to dodge the static-grep check below. We
-/// truncate just before `mkdir -p` so no filesystem side effects leak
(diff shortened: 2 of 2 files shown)


lucasdoell:
@cla-bot check

lucasdoell:
Good catch — the SCP fallback (`staging_tarball_path`, only triggered when the remote has neither curl nor wget) had the same buggy pattern. Pushed 98e8ab2 dropping the inner quotes there too. The static-scan test (`install_script_does_not_quote_home_in_tilde_substitution`) now covers both occurrences.

/oz-review

lucasdoell:
Addressed in d2bdcab:

- **Drift risk** — the live test now drives `install_script(None)` directly (truncated at the `mkdir -p "$install_dir"` checkpoint and capped with a `printf`) so it exercises the actual production substitution, not a hand-copied snippet. Verified end-to-end against `/bin/bash` 3.2.57 with `HOME=/Users/test`: resolves to `/Users/test/.warp/remote-server`, no literal quotes.
- **Broader static rule** — replaced the single-substring grep with a scan that requires every `/#\~/` substitution's replacement to start with `$`. That rejects `"$HOME"`, `"${HOME}"`, `'$HOME'`, and any other quoted variant bash 3.2 would preserve, while still allowing the correct `$HOME` and `${HOME}` forms. Mock-tested against the three regression shapes — all flagged with a clear message pointing back to the bash-3.2 root cause.

/oz-review

lucasdoell:
Good catch on the bash 5.2 patsub trap — applied your suggestion in 8914a6a. Both occurrences (install_dir and the SCP-fallback staging path) now use:

```bash
case "$var" in
  "~"|"~/"*) var="${HOME}${var#\~}" ;;
esac
```

That dodges both interpreter bugs cleanly and works identically on bash 3.2 and bash 5.2+.

Verified end-to-end against `/bin/bash` 3.2.57 with a matrix of HOME values:

```
HOME='/Users/test'        ->  '/Users/test/.warp/remote-server'
HOME='/Users/A&B'         ->  '/Users/A&B/.warp/remote-server'
HOME='/some/normal/home'  ->  '/some/normal/home/.warp/remote-server'
```

The live test (`install_script_tilde_expansion_resolves_correctly`) now iterates the same HOME matrix so the patsub case is exercised on any bash version CI runs against. The static check (`install_script_avoids_pattern_substitution_for_tilde_expansion`) is broadened to forbid `${var/#\~/...}` entirely, with the assertion message spelling out both bugs.

/oz-review

kevinyang372 (approved):
This is awesome! Thanks for fixing and adding the regression tests

kevinyang372:
@lucasdoell Mind fixing the clippy errors?

lucasdoell:
Sure, let me give it a try! I ran into some issues running the codebase locally before (something related to missing metal drivers) but I'll give it another shot and see if I can get the clippy errors fixed.

lucasdoell:
@kevinyang372 Should be all set — let me know if there's anything else that needs to be changed!

kevinyang372:
@lucasdoell Ah we should gate this test to be unix only. Can you fix that?

lucasdoell:
@kevinyang372 It has been done  🫡 Gated the test to unix and updated the docstring for clarity.

PR #10322, opened 2026-05-07
Fix Resizable panel drag lagging behind cursor

### LINKED ISSUE #10321: Resizable panel drag lags behind cursor (especially on Wayland)
## Description

When dragging a resizable panel edge (e.g. the right/left sidebar), the panel visually lags behind the cursor and can jump when reversing direction near the min/max size boundary. Most noticeable on **Wayland**.

## Root Cause

In `ResizableState::resize()` (`crates/warpui_core/src/elements/resizable.rs`), `last_position` was set to the **computed dragbar coordinate** (`origin ± new_size`) instead of the **actual cursor position**.

Each drag frame computed delta as:
```
delta = dragbar_position → cursor_now   // wrong
```
instead of:
```
delta = cursor_prev → cursor_now        // correct
```

This introduces systematic error that accumulates — especially when size hits min/max clamp, causing a visible jump on direction reversal.

`DragResizeElement` already does this correctly via `consume_delta()` which sets `last_y = y`.

## Fix

Set `last_position = new_position` (cursor position) in `resize()`. PR: leozeli/fix-resizable-drag-lag-wayland

leozeli:
Fixes #10321

## Summary

When dragging a resizable panel edge (right/left sidebar), the panel lags behind the cursor and jumps when reversing near the min/max boundary — most noticeable on Wayland.

**Root cause**: `ResizableState::resize()` was setting `last_position` to the computed dragbar coordinate (`origin ± new_size`) instead of the actual cursor position. Each frame's delta was measured from the dragbar location rather than the previous cursor location — causing systematic drift. At a clamped boundary the error accumulates and reverses into a jump.

**Fix**: Set `last_position = new_position` (cursor position), matching `DragResizeElement::consume_delta()` which already does this correctly. Also removed the now-unused `origin` parameter from the private `resize()` / `check_for_resize()` methods.

## Test plan

- [ ] Drag the right sidebar edge left/right — should track cursor exactly
- [ ] Drag to min/max width boundary, then reverse — no jump
- [ ] Verify on Wayland (most visible regression case)
- [ ] Verify on X11 / macOS — no regression

### DIFF
--- a/crates/warpui_core/src/elements/resizable.rs
+++ b/crates/warpui_core/src/elements/resizable.rs
@@ -89,11 +89,10 @@ impl ResizableState {
     fn check_for_resize(
         &mut self,
         position: Vector2F,
-        origin: Option<Vector2F>,
         dragbar_side: DragBarSide,
     ) -> Option<Vector2F> {
         if let ResizableMode::Dragging { last_position } = self.mode {
-            self.resize(last_position, position, origin, dragbar_side)
+            self.resize(last_position, position, dragbar_side)
         } else {
             None
         }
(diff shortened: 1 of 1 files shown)


app/src/workspace/view/vertical_tabs.rs
```diff
@@ -1449,23 +1449,11 @@ fn render_vertical_tabs_panel(
...
-                PositionedElementOffsetBounds::WindowByPosition,
-                PositionedElementAnchor::BottomLeft,
-                ChildAnchor::TopLeft,
-            ),
-        );
-        stack.finish()
-    } else {
-        panel_content
-    };
+    // The settings popup is rendered at the workspace level (with Dismiss for click-outside-
```
zachbai, line 1520: this change looks unrelated to the bug fix?

zachbai: would prefer to break this out into a separate branch (and eliminate the unnecessary variable assignment altogether)

zachbai (changes requested):
Thanks for the contribution - actual resize change looks good, though would like to get rid of the unrelated vertical tabs change prior to approving + merging.

leozeli:
Thanks for the catch! The branch was previously based on another fix branch which included the `vertical_tabs.rs` change. I've rebased onto `master` — the PR now contains only the `resizable.rs` fix.

PR #10327, opened 2026-05-07
Refactor orchestration pill bar swapping (fixes cancellation of child agents)

advait-m:
Loom: https://www.loom.com/share/77e59376fd5d4c51a15b1e3211b917ea
Fixes https://linear.app/warpdotdev/issue/QUALITY-627/remote-children-are-showing-blanks-agent-conversation-when-selected-in
Fixes https://linear.app/warpdotdev/issue/QUALITY-633/fix-child-agents-being-cancelled-when-switching-pills

------
## Problem

Fixes the cancelling of conversations due to exiting agent views.

The orchestration pill bar's swap mechanism kept two representations in sync: a hidden-pane visibility flag on the child's view, plus the layout tree itself. Steady state worked, but anything that took the child outside the in-place context — split-off into its own slot, open-in-new-tab, close, re-adopt, ESC back to the parent, cross-tab focus, app restart — accumulated bugs as the two drifted. The worst cases were the ones where a child was simultaneously the swap target of one orchestrator and a split-off sibling visible elsewhere, or where the parent was preserved on close but later resurrected the just-closed child via a stale entry.

## Design

Swaps now use one mechanism: a temporary replacement in the layout tree, with the original restored on revert. There is exactly one `TerminalView` per child agent for its entire lifetime. The child pane lives off-tree by default and only enters the layout when the user puts it there (swap, split, or open-in-new-tab); every path that changes the in-place context — close, split, re-adopt, exit-and-reenter agent view, cross-tab focus, snapshot persistence — reverts any active swap first and reuses the live view. The visibility flags are gone, so they can't drift. The persistence layer also substitutes the orchestrator for any swapped-in child when snapshotting a tab, so app restart restores the canonical layout rather than leaving the child stuck as the visible leaf.

A few smaller pieces fall out of this. A new history-model API updates the active-conversation pointer without ripping the conversation off other views (the existing transferring API is preserved for paths that genuinely move ownership). Guards on the agent-view exit path skip SSE teardown when the exit is part of an in-place swap. And the helper that reverts a swap also clears the orchestration-split-off marker, so a single call site handles both the in-place and split-off-then-swapped-over cases.

## UI fix

Breadcrumb row in split-off panes is now horizontally centered.

## Testing

- `cargo fmt && cargo check -p warp` and `cargo clippy -p warp --all-targets --tests -- -D warnings` both pass clean.
- Manually validated:
  - Pill-click swap between orchestrator and any child, in any order; child→child swap; orchestrator pill no-ops on self.
  - Open in new pane / Open in new tab on a child whether currently swapped in or not; live conversation survives across the move; Open in new tab close path re-adopts the live view back to the source pane group.
  - Close on a swapped-in child reverts cleanly and leaves the pill clickable.
  - Close on a split-off child preserves it off-tree; subsequent pill clicks swap as expected.
  - Close on a split-off-then-swapped-over child does not resurrect the just-closed pane via a future revert of the surviving sibling.
  - ESC + back button from a child navigates to the parent conversation, including across tabs.
  - Cross-tab focus to a pane currently occluded by a swap reveals it and refreshes back-button labels.
  - Tab restore on restart preserves the orchestrator (not the child) as the canonical leaf; active session marker matches what the user last saw.
  - Breadcrumbs centered when fitting; overflow scroll preserved.

### DIFF
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -537,11 +537,8 @@ impl BlocklistAIController {
                 return;
             };
 
-            // When the controller is just swapping the active conversation in place
-            // (e.g. clicking a pill in the orchestration pill bar), the synthetic
-            // ExitedAgentView is part of a switch — not a real exit. Cancelling the
-            // previously-active conversation here would kill its in-flight response
-            // stream / pending actions every time the user navigates between agents.
+            // Skip if this exit is part of an in-place switch — cancelling here
+            // would kill an in-flight stream every time the user navigates.
             if *is_exit_before_new_entrance {
                 return;
             }
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -530,11 +530,21 @@ impl BlocklistAIController {
             let AgentViewControllerEvent::ExitedAgentView {
                 conversation_id,
                 final_exchange_count,
+                is_exit_before_new_entrance,
                 ..
             } = event
             else {
                 return;
             };
+
+            // When the controller is just swapping the active conversation in place
+            // (e.g. clicking a pill in the orchestration pill bar), the synthetic
+            // ExitedAgentView is part of a switch — not a real exit. Cancelling the
+            // previously-active conversation here would kill its in-flight response
+            // stream / pending actions every time the user navigates between agents.
+            if *is_exit_before_new_entrance {
+                return;
+            }
 
             // If we exited a brand-new empty conversation, there's nothing meaningful to cancel.
             if *final_exchange_count == 0 {
--- a/app/src/pane_group/mod.rs
+++ b/app/src/pane_group/mod.rs
@@ -4446,7 +4446,14 @@ impl PaneGroup {
         // (e.g. via the undo stack expiring) also needs to relinquish any
         // child agent conversations back to their parents so the
         // orchestrator's pill bar keeps working in-place.
-        self.transfer_child_agent_conversations_to_parents_on_close(pane_id, ctx);
+        //
+        // Skip the transfer when the pane being discarded is itself a
+        // child agent pane: the child's `TerminalView` canonically owns
+        // its conversation, and transferring ownership away would orphan
+        // the conversation while the pane is being torn down.
+        if !self.is_child_agent_pane(pane_id) {
+            self.transfer_child_agent_conversations_to_parents_on_close(pane_id, ctx);
+        }
 
         if let Some(terminal_view) = self.terminal_view_from_pane_id(pane_id, ctx) {
             let terminal_view_id = terminal_view.id();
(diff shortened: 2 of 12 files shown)


app/src/workspace/view.rs
```diff
@@ -10242,9 +10301,17 @@ impl Workspace {
...
         }
 
-        if detach_panes_for_close {
+        // If this tab is a split-off child agent tab (created via
+        // `OpenChildAgentInNewTab`), re-adopt its lone pane back into the
+        // source pane group so the live `TerminalView` survives the tab
+        // close. When this succeeds, the tab's pane group becomes empty
+        // and we skip the regular detach-for-close path — there are no
+        // panes left to detach.
+        let re_adopted = self.try_re_adopt_split_off_child_agent_tab(index, ctx);
```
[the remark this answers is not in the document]
advait-m: fixed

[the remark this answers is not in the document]
advait-m: fixed

app/src/pane_group/mod.rs
```diff
@@ -6447,20 +6666,528 @@ impl PaneGroup {
...
+            if let Some(replacement_pane_id) =
+                self.panes.replacement_pane_for_original(parent_pane_id)
+            {
+                self.revert_swap_clearing_split_off(replacement_pane_id, ctx);
+            }
+            parent_pane_id
+        } else {
+            self.focused_pane_id(ctx)
+        };
+        let split_ok = self
```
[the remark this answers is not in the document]
advait-m: fixed

app/src/pane_group/tree.rs
```diff
@@ -324,6 +329,13 @@ impl PaneData {
         }
     }
 
+    /// Returns true if `id` is currently hidden because it is a child
+    /// agent pane. Mirrors `is_pane_hidden_for_orchestration_swap` so
```
cephalonaut: Comment stale? is_pane_hidden_for_orchestration_swap doesn't exist?

cephalonaut (approved):
Looks great! I tried it out and couldn't repro errors. I also think this might be fixing:

https://linear.app/warpdotdev/issue/QUALITY-609/subagents-become-blank-after-restarting-warp

...because I couldn't repro with you change and maybe:

https://linear.app/warpdotdev/issue/QUALITY-618/switching-between-agents-via-pill-bar-is-sluggish

because I'm wondering if the sluggishness Suraj saw was because we had to rebuild the conversation and UI when switching?

One note: seems like a lot of comments. I'm partly down with that because its complicated code and reading the comments made it more clear how things work, but I do worry about accuracy. Still we have the AI to maintain them. Up to you whether to tone them down or not. The main thing that worries me is comments that refer to behaviors of distant callers or other participants where the likelihood of change being missed is high.

advait-m:
> https://linear.app/warpdotdev/issue/QUALITY-609/subagents-become-blank-after-restarting-warp

Sweet - thanks for testing!

Yeah, I'd expect this to make switching faster

Yeah, totally agree re comments being overkill - I'll tone it down!

app/src/pane_group/tree.rs
advait-m: yep - thanks for catching will fix and tone down others!

app/src/ai/blocklist/controller.rs
```diff
@@ -530,12 +530,19 @@ impl BlocklistAIController {
...
+                is_exit_before_new_entrance,
                 ..
             } = event
             else {
                 return;
             };
 
+            // Skip if this exit is part of an in-place switch — cancelling here
+            // would kill an in-flight stream every time the user navigates.
+            if *is_exit_before_new_entrance {
```
seemeroland, line 542: With orchestration, are we still expecting to have multiple streams in the same terminal view? The pills are not backed by different agent views in the **same** terminal view right?

It feels to me like if we're exiting an agent view, we should be canceling the active stream for its conversation

advait-m: yep pills aren't different agent views in the same terminal view - each child (and the orchestrator) has its own `TerminalView` for its whole lifetime, and the pill bar just swaps panes in/out of the layout tree. a pill click shouldn't actually fire `ExitedAgentView`, so this handler isn't even hit on swap I believe - the swapped-out controller stays `Active` and keeps streaming.

and we do cancel on a real exit (`is_exit_before_new_entrance == false`). that flag is only true for the one case where the same view repoints from convo A→B in place (cmd+K new convo, restored convo taking over the pane). cancelling there would kill A's stream on every navigate, which is the bug this fixes.

re multiple streams: I think that's expected + predates orchestration — `PendingResponseStreams` is keyed per stream/convo so one controller can hold a few at once (fire A, cmd+K to B while A streams). Hence why we cancel per-convo, not blanket.

PR #10332, opened 2026-05-07
load files for the command palette async so it opens faster

### LINKED ISSUE #10256: Command Palette opens slowly in large projects before any query is typed
### Pre-submit Checks

- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). Not applicable.
- [x] I have technical issue and have included the logs (optional, but helps expedite the bug fix). Local instrumentation details are included below.

### Describe the bug

Opening the Command Palette can be noticeably slow in large projects before the user has typed a query or selected a filter.

In the default zero-state view, the palette is showing recent and suggested items. Despite that, opening the palette can still start an empty, unfiltered search. When file search is enabled, the files data source treats the empty query as a zero-state file query and scans/ranks files in the current repository. In a large repo this work happens during palette open even though the file results are not visible yet.

In local instrumentation, opening the Command Palette in this repository showed the slow path running `file_data_source.zero_state_query` over about 21,088 files. The first measured open took about 455 ms, and the next measured opens were about 107 ms. After avoiding the hidden empty query while the Command Palette zero-state is visible, repeated opens measured about 1.4-2.2 ms and no longer logged the file zero-state query.

### To reproduce

1. Open Warp in a large repository or directory with many searchable files.
2. Ensure Command Palette file search is enabled.
3. Open the Command Palette with no initial query, for example via `Cmd-P`.
4. Observe that the palette can take a noticeable amount of time to open even though it initially displays the zero-state recents/suggested UI rather than file search results.
5. Add local timing around Command Palette open and file-data-source zero-state queries. The slow opens correlate with the empty file zero-state query running during palette open.

### Expected behavior

Opening the Command Palette with an empty input should show the zero-state UI promptly. Expensive search providers should not run until their results are visible or the user types a query/selects a filter.

Selecting a visible filter such as `files:` or typing a non-empty query should still run the appropriate search immediately.

No screenshot is useful because this is a latency issue with no visible UI change.

Local instrumentation summary:

- Before fix:
  - `workspace.open_palette_handler`: about 455 ms on the first measured open.
  - Subsequent measured opens: about 107 ms.
  - `file_data_source.zero_state_query` ran over about 21,088 files while the Command Palette zero-state was visible.
- After fix:
  - `workspace.open_palette_handler`: about 1.4-2.2 ms across repeated opens.
  - No `file_data_source.zero_state_query` entries were emitted while opening the empty zero-state Command Palette.

### Operating system (OS)

macOS

macOS 26.3.1 (25D2128)

### Shell Version

### Current Warp version

v0.2026.04.29.08.57.stable_01

### Regression

### Recent working Warp date

No response

### Additional context

The implementation issue appears to be in the shared search bar behavior: changing a visible query filter to `None` can run an empty query even when the owning view intends to show a dedicated zero-state UI. Some search surfaces intentionally want empty-buffer queries and already opt into that behavior via `run_query_on_buffer_empty()`. The Command Palette does not need that empty query while its zero-state UI is visible.

Potential fix shape:

- In `SearchBar::set_visible_query_filter`, skip `run_query` when the buffer is empty, the filter state is zero-state, and the search bar has not opted into `run_query_on_buffer_empty()`.
- Keep running queries for non-empty buffers and visible filters.
- Keep fixed-filter palettes running their initial query.
- Avoid session-source-triggered query reruns while the Command Palette zero-state is visible.

No

Ignore

acarl005:
## Description

This PR refactors the `FileDataSource` to load the files asynchronously. This avoids blocking the main thread and solves the momentary pause when opening the command palette.

## Linked Issue

Fixes #10256

## Testing

No perceived changes in local testing. On slower machines, the results may not actually appear for a couple of seconds while the data source is still computing the results.

I sent the user a custom built from this branch and they have confirmed it solves the issue

https://github.com/warpdotdev/warp/issues/10256#issuecomment-4394618726

CHANGELOG-IMPROVEMENT: Performance improvement for opening the command palette when in a project with a lot of files.

### DIFF
--- a/app/src/search/command_palette/files/data_source.rs
+++ b/app/src/search/command_palette/files/data_source.rs
@@ -127,47 +127,56 @@ impl FileDataSource {
 
         let (contents, git_changed_files) = self.contents_with_git_changes(app);
 
-        let mut results = Vec::new();
-
         let opened_files = OpenedFilesModel::as_ref(app);
 
         let repo_root = file_search_model.repo_root(app);
-        let opened_files =
-            repo_root.and_then(|repo_root| opened_files.opened_files_for_repo(&repo_root));
+        let opened_files = repo_root
+            .and_then(|repo_root| opened_files.opened_files_for_repo(&repo_root))
+            .cloned();
 
-        for item in contents.iter() {
-            let mut file_ranking = if git_changed_files.contains(&item.path) {
-                FileRanking::ChangedInGit
-            } else {
-                FileRanking::None
-            };
+        const CHUNK_SIZE: usize = 50;
 
-            if let Some(last_opened_timestamp) =
-                opened_files.and_then(|opened_files| opened_files.get(&PathBuf::from(&item.path)))
-            {
-                file_ranking = FileRanking::OpenedInWarp {
-                    timestamp: *last_opened_timestamp,
-                };
+        Box::pin(async move {
+            let mut results = Vec::new();
+
+            for chunk in contents.chunks(CHUNK_SIZE) {
+                for item in chunk {
+                    let mut file_ranking = if git_changed_files.contains(&item.path) {
+                        FileRanking::ChangedInGit
+                    } else {
+                        FileRanking::None
+                    };
+
+                    if let Some(last_opened_timestamp) = opened_files
+                        .as_ref()
+                        .and_then(|opened_files| opened_files.get(&PathBuf::from(&item.path)))
+                    {
+                        file_ranking = FileRanking::OpenedInWarp {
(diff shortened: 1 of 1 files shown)


kevinchevalier (approved):
I'm approving since it fixed the issue for the user, but it's unclear to me which part of this is fixing the issue. Which part was slow before? 

It seems like computing the contents is still sync, and then the other work were just a couple of hashmap lookups.

Is it the sort?

acarl005:
I suspect it's the heap allocations happening inside the for-loop. There could be 100's of thousands if not millions of files.

PR #10333, opened 2026-05-07
Tab configs: optional profile field on agent panes (#10171)

### LINKED ISSUE #10171: Feature request: Allow tab configs to specify an agent profile for agent-mode panes

rgatx:
Hi @moirahuang, I'm Roam. Let me know if you see any of the design choices here different from what you intended in the issue thread. The duplicate-name abort and the missing-name Default fallback are my read of "error instead of trying to choose." Happy to iterate.

## Description

Adds support for specifying an AI execution profile by display name on agent-typed panes in tab configurations. A new optional `profile` field on `[[panes]]` of `type = "agent"` resolves at launch against the user's configured profiles. Honors @moirahuang's binding direction in the issue thread: name-based reference, error on duplicate names rather than silently choosing.

Per @moirahuang in the issue thread:

> Yeah I like this too! I think we don't expose the profile ID in the UI. So for now, let's use the profile name but if there are duplicate profile names, we should error instead of trying to choose one of the profiles. I'm going to move this to `ready-to-implement`!

## Linked Issue

Closes #10171
Refs #10229 (spec exploration; this PR ships against the binding direction in the issue thread)

## Design rationale

Five UX cases, each picked deliberately:

**Duplicate names → abort agent-mode entry.** The pane stays as a terminal with an error toast naming the duplicate count and pointing the user at Settings → Agents → Profiles. Defended against the obvious alternative of fall-back-to-default: a fall-back still picks a profile after ambiguity, which is exactly what moirahuang's "error instead of trying to choose" rules out. Aborting agent entry preserves the "no silent profile selection" principle while keeping the pane usable as a terminal.

**Missing names → fall back to Default + warning toast.** Defended against abort-agent-mode-entry: scripted or shared tab configs may reference profiles that don't exist on this machine due to typos, profile renames, or unsynced profiles. Hard-aborting on missing makes typos catastrophic; falling back to Default with a toast is graceful and recoverable. The active profile is visible in the pane chrome so the user sees they're not on the requested profile.

**Terminal pane with `profile` field set → log::warn and ignore.** Defended against reject-at-render. The existing `tab_config.rs:227-238` pattern falls back to a benign default-template on resolution errors rather than wholesale-failing the tab config; this matches that precedent. A reject-at-render approach would fail an entire tab config for one benign typo on a non-agent pane.

**Omitted `profile` field → preserve current behavior.** Verified in code (`profiles.rs:355-364`): `active_profile(Some(view_id))` falls back to `default_profile(ctx)` when no per-session profile exists. Today's behavior for new agent panes is "use the default profile," not "use the last-used profile." Preserving that exactly.

**Empty string after trim → treated as omitted.** Backward-compat for users who set `profile = ""`.

## Trust / consent

Profile selection is permission-affecting because it can apply an existing high-autonomy profile from the user's profile set. Verified via `BlocklistAIPermissions::active_permissions_profile` at `permissions.rs:215-223`, which derives the active permission set from the view's currently selected profile. The mitigation in this PR is visible disclosure: the active profile is shown in the agent input footer chrome, so the user can see whether the requested profile was applied. Tab configs are user-edited and user-launched, so the TOML serves as primary user consent.

The profile is applied **before** `enter_agent_view_for_new_conversation` at both ordering anchors:

- Immediate launch path: `pane_group/mod.rs:~1352` (`set_active_profile` runs in the `view.update` closure before the enter call at ~1353)
- Deferred launch path: `terminal/view.rs:10894-10901` (a stashed profile name on the new `pending_agent_profile_name` field is applied before the deferred enter call)

This ensures the first agent input runs under the configured profile at every entry path, including tab configs that combine `commands = [...]` with `type = "agent"`.

## Testing

**Automated**:

7 unit tests in `profiles_tests.rs` covering the lookup helper:

- `find_profile_by_display_name_unique_match`
- `find_profile_by_display_name_not_found`
- `find_profile_by_display_name_ambiguous_two_custom_same_name`
- `find_profile_by_display_name_default_resolves`
- `find_profile_by_display_name_two_untitled_collide`
- `find_profile_by_display_name_trims_input`
- `find_profile_by_display_name_empty_input_returns_not_found`

5 unit tests in `tab_config_tests.rs` covering schema:

- `test_render_tab_config_propagates_profile_field`
- `test_render_tab_config_trims_profile_whitespace`
- `test_render_tab_config_empty_profile_treated_as_omitted`
- `test_render_tab_config_profile_on_terminal_pane_passes`
- `test_render_tab_config_roundtrip_profile`

**Integration test gap**: `crates/integration/` has no existing harness for seeding multiple `AIExecutionProfile` fixtures, since profiles are cloud-synced and the integration harness doesn't currently mock the sync layer. The closest viable automated coverage is unit-level: the lookup helper's tests use the actual `CloudModel` reconciliation path with a real `AIExecutionProfilesModel`, exercising the sync flow inside a test app. Manual validation covers end-to-end.

**Manual validation**:

Six scenarios captured below.

### Screenshots / Videos

**1. Success, named profile applied**: tab config with `profile = "Personal"` opens the agent pane with Personal active in the input footer.

Screenshot 2026-05-06 at 10 49 49 PM

**2. Missing, named profile not found**: tab config with `profile = "Typo"` falls back to Default with a warning toast.

Screenshot 2026-05-06 at 10 52 12 PM

**3. Duplicate, two profiles same name**: tab config with `profile = "Work"` (where two profiles named "Work" exist) keeps the pane as a terminal and shows an error toast with the duplicate count.

Screenshot 2026-05-06 at 10 53 34 PM

**4a. Terminal pane with `profile` field**: terminal opens normally with no toast.

Screenshot 2026-05-06 at 10 54 48 PM

**4b. Terminal pane warn (log evidence)**: the matching `log::warn!` from `warp::tab_configs::tab_config` fires, naming the offending pane and the field that's being ignored.

Screenshot 2026-05-06 at 11 51 04 PM

**5. Deferred launch, commands then agent**: tab config with `commands = ["sleep 2"]` plus `type = "agent"` plus `profile = "Personal"` runs the command first in terminal mode, then transitions to agent mode with Personal active.

https://www.loom.com/share/d2ce9bc1db744da28aa5dca73c393330

**6. Per-launch toast dedup**: tab config with two agent panes both referencing missing profile "Typo" produces a single toast, not two, despite both panes being affected.

Screenshot 2026-05-07 at 12 03 38 AM

CHANGELOG-NEW-FEATURE: Tab configs: optional `profile` field on agent panes selects an AI execution profile by name. Errors clearly on duplicate names; falls back to Default with a warning toast on missing names.


app/src/terminal/view.rs
```diff
@@ -14740,6 +14760,62 @@ impl TerminalView {
...
+            .as_ref(ctx)
+            .find_profile_by_display_name(profile_name, ctx);
+        match lookup {
+            Ok(profile_id) => {
+                profiles_model.update(ctx, |model, ctx| {
+                    model.set_active_profile(view_id, profile_id, ctx);
+                });
+                true
+            }
+            Err(ProfileLookupError::NotFound(name)) => {
```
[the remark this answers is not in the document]
rgatx: just updated to address nit

rgatx:
R2 addressed: dedup is now per-launch-scoped. 

  On master at f37cc46d, `cargo test --lib` shows 10 unrelated pre-existing failures in server::telemetry, settings::init, terminal::bootstrap,  terminal::view, and workspace::view. All pass when run in isolation. None reference pane_group, execution_profiles, or tab-config code. The new test passes; the broader pane_group test run is clean (52 passed, 0 failed).

app/src/ai/execution_profiles/profiles.rs
```diff
@@ -456,6 +464,55 @@ impl AIExecutionProfilesModel {
...
+    /// Record that a missing tab-config agent profile name has been surfaced
+    /// to the user via a "not found; using Default" toast. Returns `true` if
+    /// this is the first time this name has been warned about in the current
+    /// app session (caller should emit the toast); returns `false` if a toast
+    /// has already been emitted for this name (caller should suppress to
+    /// avoid duplicate notices across the immediate and deferred tab-config
+    /// launch paths).
+    ///
+    /// Dedup scope is per app session; resets on restart.
+    pub fn note_missing_tab_config_profile(&mut self, name: String) -> bool {
```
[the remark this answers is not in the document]
rgatx: addressed in subsequent commit

rgatx:
@moirahuang rebased onto main;  CI re-running on 1f33a5bb.

app/src/ai/execution_profiles/profiles.rs
```diff
@@ -456,6 +472,42 @@ impl AIExecutionProfilesModel {
         })
     }
 
+    /// Look up a profile by its display name (the same string shown in the
+    /// agent input footer and Settings → Agents → Profiles). Trims input.
+    ///
+    /// Per moirahuang's binding direction in #10171, duplicate display names
```
moirahuang, line 478: nit: we don't need to reference my comment or the issue here, we can just leave it as the comment w rationale

app/src/launch_configs/launch_config.rs
```diff
@@ -105,6 +105,11 @@ pub enum PaneTemplateType {
         /// Sourced from the `shell` field of a tab config pane node.
         #[serde(skip_serializing_if = "Option::is_none", default)]
         shell: Option<String>,
+        /// Optional AI execution profile to apply to an agent pane, matched
+        /// by display name. Sourced from the `profile` field of a tab config
+        /// pane node. Trimmed; empty after trim is treated as `None`.
+        #[serde(skip_serializing_if = "Option::is_none", default)]
+        agent_profile_name: Option<String>,
```
moirahuang, line 112: let's name this `profile` to be consistent w the tab config field name

app/src/pane_group/mod.rs
```diff
@@ -1454,6 +1490,66 @@ impl PaneGroup {
         }
     }
 
+    /// Resolve a tab-config `agent_profile_name` against the user's profiles
+    /// and apply it to the given terminal view, emitting a toast on the
+    /// no-match / ambiguous-match cases.
+    ///
+    /// Returns `true` when agent-mode entry should proceed and `false` when
+    /// it should be aborted (i.e. ambiguous profile name — the pane stays as
+    /// a terminal). On a missing profile we fall through to the default
```
moirahuang, line 1499: i think we can still resolve to an agent pane, even if the profile is unknown. we can fall back to the default profile in that case. this should also simplify our logic so we generally are falling back to the default profile

moirahuang (commented):
a few higher level comments:

1. we should update the tab config skills to account for the new profile filed
2. we should update docs to reflect the new field fyi: @hongyi-chen 
3. i see we have logic in the PR to handle duplicate toasts but i'm still getting duplicate toasts here. i will say, seeing the multiple toasts actually feels useful to me. it tells me that multiple tabs attempted to open the duplicate profile. i'd propose we intentionally don't de-dupe, it'll also simplify some of the PR's logic as well

Image

nwp:
@rgatx, @moirahuang this seemed close to the finish line then was closed. Did the direction change or can this be revived?

PR #10334, opened 2026-05-07
Add "Reveal in Finder" to code pane overflow menu

### LINKED ISSUE #10257: Add 'Reveal in Finder' to code pane overflow menu

wzc520pyfm:
Close https://github.com/warpdotdev/warp/issues/10257 .

## Description

Adds a `Reveal in Finder` item to the per-code-pane overflow (`⋮`) menu, with a `Cmd+Option+R` shortcut on macOS. Selecting it opens the active code tab's file in the platform's file manager (Finder on macOS, Explorer on Windows, or the system file manager on Linux) via `AppContext::open_file_path_in_explorer`.

The label is platform-aware (`Reveal in Finder` / `Reveal in Explorer` / `Reveal in file manager`), and the keystroke chip is driven by `keybinding_name_to_keystroke`, so user overrides set in Settings → Keybindings are reflected in the menu chip.

The work is split across two focused commits:

1. **Add Reveal in Finder action to code pane overflow menu** — new `CodeViewAction::RevealInFinder` variant + menu item + handler. The item is only emitted when the active tab has a resolvable local path (mirroring the existing `Copy file path` guard), and the variant is gated by the `local_fs` feature for the same reason.
2. **Add Cmd+Option+R keybinding for Reveal in Finder** — registers `code_view:reveal_in_finder` as an `EditableBinding` scoped to `CodeEditorView` with a default macOS keystroke; the menu item label switches to `MenuItemFields::new_with_label` so the chip displays the bound shortcut. Linux/Windows ship without a default to avoid clashing with common shortcuts there; users can still bind it themselves.

## Linked Issue

Implements #10257.

## Testing

Manual verification on macOS:
- Open a file in the code pane → click the pane's `⋮` menu → `Reveal in Finder  ⌥⌘R` appears beneath `Copy file path`. Selecting it opens Finder at the file's location.
- Pressing `Cmd+Option+R` while the editor is focused does the same.
- The chip updates if the binding is overridden in Settings → Keybindings.
- The entry is hidden when the active tab has no resolvable local path (e.g. unsaved buffers).

**Why no integration test**: the user-visible side effect is the platform's `open_file_path_in_explorer` call, which is implemented as a no-op in the test platform delegate (`crates/warpui_core/src/platform/test/delegate.rs`), so there is no observable post-condition for an integration assertion. In addition, the integration crate cannot directly reference `CodeView` / `CodeViewAction` (the `code` module is private to the `warp` crate) or `MenuItem::fields()` (`#[cfg(test)]`-gated) without expanding the public API surface just for the test. I'm happy to revisit this in a follow-up if reviewers would like a UI-driven smoke test or want to expose those internals behind a `feature = "integration_tests"` gate.

### Screenshots / Videos

_Screenshot showing the new menu entry will be added shortly — captured locally from `./script/run`._

Snipaste_menu_action_item

Snipaste_menu_action_open_finder

## Agent Mode

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/src/code/view.rs
+++ b/app/src/code/view.rs
@@ -162,6 +162,11 @@ pub enum CodeViewAction {
     ToggleMaximized,
     #[cfg(feature = "local_fs")]
     CopyFilePath,
+    /// Open the active code tab's file in the platform's file manager
+    /// (Finder on macOS, Explorer on Windows). No-op when the active tab has
+    /// no resolvable local path.
+    #[cfg(feature = "local_fs")]
+    RevealInFinder,
     #[cfg(feature = "local_fs")]
     RenderMarkdown,
     DragOverIndex {
@@ -1953,10 +1958,20 @@ impl CodeView {
 
         #[cfg(feature = "local_fs")]
         if let Some(path) = self.local_path(ctx) {
+            let reveal_label = if cfg!(target_os = "macos") {
+                "Reveal in Finder"
+            } else if cfg!(target_os = "windows") {
+                "Reveal in Explorer"
+            } else {
+                "Reveal in file manager"
+            };
             items.extend([
                 MenuItem::Separator,
                 MenuItemFields::new("Copy file path")
                     .with_on_select_action(CodeViewAction::CopyFilePath)
+                    .into_item(),
+                MenuItemFields::new(reveal_label)
+                    .with_on_select_action(CodeViewAction::RevealInFinder)
                     .into_item(),
             ]);
 
@@ -2117,6 +2132,16 @@ impl TypedActionView for CodeView {
                 if let Some(path) = self.local_path(ctx) {
                     ctx.clipboard()
                         .write(ClipboardContent::plain_text(path.display().to_string()));
+                }
+            }
+            #[cfg(feature = "local_fs")]
+            CodeViewAction::RevealInFinder => {
+                if let Some(path) = self.local_path(ctx) {
+                    ctx.open_file_path_in_explorer(&path);
+                } else {
+                    log::warn!(
+                        "Reveal in Finder requested, but the active code tab has no local file path"
+                    );
                 }
             }
             #[cfg(feature = "local_fs")]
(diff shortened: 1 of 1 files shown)


wzc520pyfm:
@cla-bot check

app/src/code/view.rs
```diff
@@ -128,6 +132,14 @@ pub fn init(app: &mut AppContext) {
         )
         .with_context_predicate(id!("CodeEditorView"))
         .with_key_binding("cmdorctrl-r u"),
+        #[cfg(feature = "local_fs")]
+        EditableBinding::new(
```
kevinyang372: I don't think we need an editable keybinding for this action

kevinyang372 (commented):
This looks good! Can you get rid of the editable keybinding? I could approve and merge after that

app/src/code/view.rs
wzc520pyfm: I have removed editable keybinding. And the local verification function is normal.
image

kevinyang372 (approved):
Thanks for contributing!

PR #10067, opened 2026-05-04
Implement revised summary mode

zachbai:
## Description
[Demo](https://www.loom.com/share/d0847201dfb54d53ac21053b36861494)

Pretty much all explained in loom.

Issue #9897, opened 2026-05-02
Warp vertical tabs icons are not the same size


EdenBendheim:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

In the current state of the vertical tabs for warp, tab types like terminal/setting tabs (as seen below) have different sizing than agent tabs, independent of weather they are agentCLI or Oz.

### Is your feature request related to a problem? Please describe.

Visual

### Additional context

<img width="245" height="198" alt="Image" src="https://github.com/user-attachments/assets/0f61a7a8-b674-44aa-bf87-f28975b97962" />

### Operating system (OS)

macOS

### How important is this feature to you?

1 (Not too important)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

vkodithala:
Hey @EdenBendheim - Varoon from the Warp team here. Thanks for filing this issue. Is the issue that terminal/settings vtabs are actually a different size than agent tabs, or just that text on the Settings tab is off-center? Not really seeing a noticeable size difference locally, but lemme know!:
<img width="271" height="135" alt="Image" src="https://github.com/user-attachments/assets/7c8b1a9e-a0a3-41b5-9d01-b7105e6f3ab5" />

EdenBendheim:
Hi @vkodithala, it's great to meet you! 

It seems that when you prompt the Oz agent, and the little status icon shows up, the sizing pushes the text further right making it aligned again. This just does not happen for CLI agents.

<img width="237" height="139" alt="Image" src="https://github.com/user-attachments/assets/44b993da-5d64-4161-a5f6-96cb38901c0e" />
^(Easier to see in tab view but happens in pane as well)


I'd be happy to open a PR for this I am just unsure what the intended solution would look like exactly, between shrinking the icon when a status is added to it, or adjusting the spacing depending on the size so the text is always aligned.

Thanks for taking a look, and let me know what you think!

(Edited to correct cause of sizing issue)

EdenBendheim:
Additionally, for cases where two different icons are present in summary view, the different icons look different weather it is an agent as the first icon with a terminal as the second, or vice versa.

[...]

PR #10349, opened 2026-05-07
Align pane header and vertical tabs agent status with #10067 pattern

EdenBendheim:
## Description
This extends #10067 which moved conversation status out of the agent icon overlay for the tab view in the vertical tab summary mode. We now fix the same design issues which were present like the agent/CLI icons being too small be default (As mentioned in #9897).

*OLD*
image

*NEW*
image

## Linked Issue

## Testing

## Agent Mode

### DIFF
--- a/app/src/ai/conversation_status_ui.rs
+++ b/app/src/ai/conversation_status_ui.rs
@@ -2,7 +2,10 @@ use warp_core::ui::appearance::Appearance;
 use warp_core::ui::color::coloru_with_opacity;
 use warp_core::ui::theme::{Fill, WarpTheme};
 use warpui::color::ColorU;
-use warpui::elements::{ConstrainedBox, Container, CornerRadius, Radius};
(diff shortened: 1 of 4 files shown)


EdenBendheim:
Hi! I saw your change was merged yesterday so I figured I would update the pane view as well at least for the status icons. I did not move over the git view since that seems like a design choice. If you think it's a good idea I would be happy to implement it, or if I have made any errors please let me know :).

Thanks!

zachbai:
@EdenBendheim thanks for the effort here. That said, the icon design for the pane view is deliberate and we'd like to keep it that way

EdenBendheim:
@zachbai Okay I see. Would it be alright to refactor this PR just to address the issues pointed out in #9897?

This was fixed in your implementation for the tabs but the pane view still has smaller icons for agents, even when they don't have a status.

Is this intentional?

Issue #7436, opened 2025-09-12
# Bug Report: Sidebar Cannot Display Chinese Filenames Correctly


bookandlover:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [x] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID intructions](https://docs.warp.dev/agents/ai-faqs#gathering-ai-debugging-id)
- [x] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log intructions](https://docs.warp.dev/support-and-billing/sending-us-feedback#gathering-warp-logs)

## Related Issues
- #6891 (IME Support Issues for Non-Latin Languages) - Related to CJK character handling
- #7314 (Project Explorer: Limited visibility) - Similar sidebar display issues

## Environment
- macOS Version: [Your Version]
- Warp Version: [Your Version from Settings > About]
- Locale: zh_CN.UTF-8 (correctly configured)

## Problem Description
The Warp sidebar (file explorer/project explorer) cannot display Chinese characters in filenames. This is a rendering issue specific to the sidebar UI component, while the terminal itself handles Chinese perfectly.

## Current Behavior
- Chinese filenames are invisible or show as blank in the sidebar
- Terminal commands (`ls`, `cat`, `echo`) display Chinese correctly
- The issue only affects the sidebar/file explorer UI component

## Expected Behavior
Chinese filenames should be displayed correctly in the sidebar, matching the terminal's capability to display Chinese characters.

## Steps to Reproduce
1. Create files with Chinese names:
```bash
   touch 中文测试.txt
   touch 测试文档.md
   mkdir 中文目录

Open the sidebar/file explorer
Observe that Chinese filenames are not displayed

Attempted Solutions

Changed fonts to SF Mono, PingFang SC, and other Chinese-supporting fonts
Configured locale settings to zh_CN.UTF-8
Created user_preferences.json with font fallback settings
Cleared cache and reinstalled Warp
All solutions failed to resolve the sidebar display issue

### To reproduce

<img width="2113" height="976" alt="Image" src="https://github.com/user-attachments/assets/132af37d-0a79-4d41-91d8-a9c49ebfeb96" />

### Expected behavior

[...]

Issue #4972, opened 2024-05-31
Underline on double-width (e.g. CJK) characters is not properly drawn


wzxu:
### Dupe Check

- [X] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates

### Describe the bug

Underline only runs half-width for double-width characters leaving holes in the entire hyperlink
<img width="284" alt="76117" src="https://github.com/warpdotdev/Warp/assets/5619348/215bad19-d465-4430-a12d-e42f9a1e5b93">

### To reproduce

1. Go to folder where files with CJK characters in the filename exist, e.g. 000測試檔案-123-中文-456字.txt
2. Run ls
3. Hover over the filename and observe the underline

### Expected behavior

Underline should extend all the way across double-width characters

### Operating system

MacOS

### Operating system and version

Sonoma 14.5

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this a Warp specific issue? (i.e. does it happen in Terminal, iTerm, Kitty, etc.)

Yes, this I confirmed this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

dannyneira:
Hi @wzxu Thanks for letting us know. It's not clear but this may be related to [ligature](https://github.com/warpdotdev/Warp/issues/239) or [IME](https://github.com/warpdotdev/Warp/issues/341) support.

We'll post any updates on those threads.

To anyone else facing this issue, please add a 👍 to the original post at the top or comment with your details, and subscribe if you'd like to be notified.

Issue #4700, opened 2024-04-05
There is a bug when entering Chinese on the terminal, and the content is not displayed


h3lp404:
### Describe the bug

There is a bug when entering Chinese on the terminal, and the content is not displayed

### To reproduce

When you enter Chinese content in the terminal or edit content, the content will not be displayed

### Operating system

MacOS

### Operating system and version

14.4.1

### Current Warp version

v0.2024.04.02.08.02.stable_01

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this a Warp specific issue? (i.e. does it happen in Terminal, iTerm, Kitty, etc.)

Yes, this I confirmed this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

dannyneira:
Hi @h3lp404 thanks for letting us know. This may be due to the font which you're using, try make sure you have a font with Chinese characters set in your settings > appearance > fonts.

h3lp404:
> Hi @h3lp404 thanks for letting us know. This may be due to the font which you're using, try to make sure you have a font with Chinese characters set in your settings > appearance > fonts. This may be due to the font you are using, please try to make sure you have a font with Chinese characters set in your Settings> Appearance> Fonts.


> Hi @h3lp404 thanks for letting us know. This may be due to the font which you're using, try make sure you have a font with Chinese characters set in your settings > appearance > fonts.


No, not fonts setting ,I use hack font, the content can be displayed, but when editing(ex:vim test.txt), the front text content will not be displayed, and the front content will not be displayed until the content appears later.

h3lp404:
> Hi @h3lp404 thanks for letting us know. This may be due to the font which you're using, try make sure you have a font with Chinese characters set in your settings > appearance > fonts.

like this:
![Longshot-20240408104235](https://github.com/warpdotdev/Warp/assets/112472696/9e8bf268-612e-490f-b2a9-7a62f3a2cb80)

eliefly:
I encountered a similar issue on Windows.

  Environment:
  - OS: Windows 11 Pro, build 26200
  - IME: Microsoft Pinyin
  - Warp version: v0.2026.04.08.08.36.stable_05
  - Shell: PowerShell

  Behavior:
  - Chinese IME preedit/composition text is not visible while typing
[...]

Issue #1823, opened 2022-09-09
chinese support 


Sunbalcony:
### Describe the solution you'd like?

```text
The warp is very good, but all the Chinese are garbled. I hope to add support for Chinese in the future version, I am very optimistic about the warp, thank you
```

### Is your feature request related to a problem? Please describe.

```text
The warp is very good, but all the Chinese are garbled. I hope to add support for Chinese in the future version, I am very optimistic about the warp, thank you
```

### How important is this feature to you?

4

dannyneira:
@Sunbalcony Hi, i'm Danny Support Engineer at Warp. Thanks for this feature request. We also have a similar request out [#1194](https://github.com/warpdotdev/Warp/issues/1194). Please +1 that one and yours as it helps us gauge interest.

Sunbalcony:
OK

wdkGitHub:
+1

vorporeal:
Would you mind sharing a picture of what you're experiencing?  In what way does Chinese text get garbled, and what should it look like/what does it look like in other terminals?

JohnWangggg:
+1

MartinDai:
+1

foreverzmy:
您好，您的来信我已经收到，请等待回复！

orange-juzipi:
+1

foreverzmy:
您好，您的来信我已经收到，请等待回复！

icezohu:
It's fine when listing files on Mac. But it will show encoding or something unreadable on Linux. Please fix this. Thank you!

**Command "ls" on Linux:**
<img width="537" alt="Snipaste_2023-05-26_08-45-35" src="https://github.com/warpdotdev/Warp/assets/3282625/2ca8d15d-d54a-4964-8c25-68b4e34b57ec">

**Command "lnav" on Linux:**
<img width="235" alt="Snipaste_2023-05-26_08-53-49" src="https://github.com/warpdotdev/Warp/assets/3282625/f03ce6a0-e0e0-4c11-8747-921777d7e291">

icezohu:
The command for testing Chinese `touch 中文.txt`

BusyHenryMa:
+1

chowhwei:
+1

MaoMiMao:
+1

pglc1026:
+1

sharockys:
+1

ideban:
+1

zy-kkk:
+1

[...]

PR #10352, opened 2026-05-07
fix(terminal): paint full-width background for wide-char SGR attrs

spalagu:
## Summary

When a wide character (CJK / emoji) is emitted with a non-default background — most visibly via SGR `\e[7m...\e[27m` (reverse video) — only the leading half of the character receives the attribute background. The right half (the WIDE_CHAR_SPACER cell) stays at the default / previous background. This breaks any TUI that draws an inline pseudo-cursor by emitting `chalk.inverse(<wide-char>)` instead of using the terminal's native block cursor — most notably **Ink's `<TextInput>`**, used by Claude Code, opencode, and many other Ink-based CLIs.

## Repro

Inside Warp, run a TUI that uses Ink (e.g., Claude Code), type any CJK string into the input box, and arrow-key the cursor onto a wide character. Expected: the cursor block fully covers the wide char. Actual: only the left half is highlighted; the right half stays normal.

The same path is exercised by simply emitting:

    printf '\e[7m遮\e[27m\n'

inside Warp — the right half of `遮` is not reversed.

## Root cause

Ink's `<TextInput>` emits the cursor as inline reverse-video bytes ([source](https://github.com/vadimdemedes/ink-text-input/blob/master/source/index.tsx)):

\`\`\`js
// Fake mouse cursor, because it's too inconvenient to deal with
// actual cursor and ansi escapes
renderedValue += i >= cursorOffset - cursorActualWidth && i <= cursorOffset
    ? chalk.inverse(char)
    : char;
\`\`\`

Warp's grid handler correctly inherits the INVERSE flag from the SGR template onto **both** the leading WIDE_CHAR cell and the WIDE_CHAR_SPACER cell at the model layer (\`app/src/terminal/model/grid/ansi_handler.rs::write_at_cursor\`). The bug is purely in the renderer.

In \`app/src/terminal/grid_renderer.rs\`, the cell iteration loop skips WIDE_CHAR_SPACER cells with a hard \`continue;\` — they are treated as pure placeholders for the wide glyph drawn by the leading cell. That skip cuts off background painting too. The leading WIDE_CHAR cell only contributes 1 column to the batched-bg cache, the spacer contributes 0, so the bg is painted as 1 cell wide instead of 2.

## Fix

In \`render_cell()\`, after \`maybe_draw_background()\` returns, if the current cell is a leading WIDE_CHAR, extend the cached bg's end column by one (to the spacer's position). This makes the batched paint span both halves of the wide char — which is what every other terminal emulator does. The spacer's \`continue;\` skip elsewhere remains correct: the leading cell now paints both columns of bg, and the wide glyph itself was already drawn spanning both columns.

12 lines added, no other changes.

## Verification

- \`cargo check -p warp --bin warp-oss --features gui\` clean (0 errors).
- Manual repro on a \`cargo bundle --release\` build of this branch:
  - Before: \`printf '\e[7m遮\e[27m'\` shows half-reversed \`遮\`. Claude Code input with cursor on CJK shows half-cursor.
  - After: full-character reverse. Cursor block fully covers wide char.

## Related issues

- #1823 (Chinese support master, 2022)
- #4700 (Chinese input not displaying, \`ready-to-implement\`)
- #7436 (Chinese display anomalies, \`ready-to-implement\`)
- #4972 (double-wide underline half-width, closed \`not-planned\` — same class of bug, may want to revisit)

## Notes

LEADING_WIDE_CHAR_SPACER (the wrap-around spacer at the previous row's last column when a wide char wraps to the next line) is a separate case and unaffected by this change.

This change is independent and safe to ship in isolation.

### DIFF
--- a/app/src/terminal/grid_renderer.rs
+++ b/app/src/terminal/grid_renderer.rs
@@ -926,6 +926,18 @@ fn render_cell(
         ctx,
     );
 
+    // Extend the cached background to also cover the WIDE_CHAR_SPACER cell on the right.
+    // Spacer cells are skipped during cell iteration (they're placeholders for the wide
+    // glyph drawn by the leading cell), so without this extension the spacer's column
+    // would never paint a background. That breaks any background attribute applied to a
+    // wide character via SGR — most visibly the reverse-video pseudo-cursor that
+    // Ink-based TUIs (Claude Code, ink-text-input, etc.) emit via `chalk.inverse(<cjk>)`,
+    // which would otherwise show only the left half of the wide character highlighted.
+    if cell.flags().intersects(Flags::WIDE_CHAR) && col + 1 < grid.columns() {
+        cached_background_color =
+            cached_background_color.map(|cached| cached.with_end(col + 1, offset_row));
+    }
+
     let glyph_offset = cell_size * vec2f(col as f32, offset_row as f32);
 
     render_cell_glyph(
@@ -1350,6 +1362,16 @@ fn render_grid_with_ligatures<'a>(
                 cell_colors.background_color,
                 ctx,
             );
+
+            // Same wide-char background extension as in `render_cell` above. The ligature
+            // rendering path also skips WIDE_CHAR_SPACER cells via `continue;` earlier in
+            // this loop, so without this extension the spacer's column would not receive
+            // the leading cell's background — manifesting as half-reversed wide chars when
+            // a TUI emits e.g. `chalk.inverse(<cjk>)`.
+            if cell.flags().intersects(Flags::WIDE_CHAR) && col + 1 < grid.columns() {
+                cached_background_color =
+                    cached_background_color.map(|cached| cached.with_end(col + 1, offset_row));
+            }
 
             let glyph_offset = cell_size * vec2f(col as f32, offset_row as f32);
             if first_cell_in_link {


spalagu:
Force-pushed a clean rebase that addresses the issues from the prior round:

**1. Scope cleanup**
The previous push had a stale base that pulled in unrelated commits from a sibling branch (a separate opt-in mouse-selection setting, plus a now-reverted cursor-snap experiment). The PR is now a single 22-line commit on top of `master`, touching only `app/src/terminal/grid_renderer.rs`.

**2. Ligature renderer path also fixed (responding to oz-for-oss review)**
The first push only patched `render_cell`. As pointed out, the ligature rendering loop further down in the same file has equivalent `WIDE_CHAR_SPACER` skip + background-batching logic and would have kept the bug under ligature rendering. The same wide-char extension is now applied at both call sites, so the fix covers both code paths.

The other concern from the oz review (modifier state propagation in `should_intercept_mouse`) belonged to the unrelated mouse-selection diff that was accidentally swept in — that work is being tracked separately under #10353 and is no longer part of this PR.

**3. CLA**
Re-authored the commit with `m.spalagu@gmail.com`, which is verified on the GitHub account that opened this PR. The cla-bot should re-validate on the new commit.

Repro and verification unchanged from the PR description. `cargo check -p warp --bin warp-oss --features gui` is clean.

spalagu:
@cla-bot check

spalagu:
Sorry — the source branch was accidentally deleted during a downstream cleanup, which auto-closed this PR. Branch is restored to the same head SHA `5aaaa1c1` (the post-rebase state from the prior round) and the PR is reopened. No content has changed; please continue review where it left off.

Issue #9830, opened 2026-05-01
Idle Warp tabs drain GitHub GraphQL API rate limit (~2.4 calls/sec)


mickdarling:
## Summary

Warp appears to make continuous GitHub GraphQL API calls from idle/inactive tabs, rapidly consuming the user's GitHub API rate limit (5,000 calls/hour). With many windows and tabs open, this can exhaust the quota in under 30 minutes even with no active sessions.

## Environment

- **Warp version**: Latest (as of 2026-05-01)
- **OS**: macOS (Apple Silicon)
- **GitHub auth**: `gh` CLI authenticated via OAuth (`gho_` token)
- **Setup**: ~12 Warp windows, ~12 tabs each. Tabs had previously been used for Claude Code, Codex CLI, and Gemini CLI sessions. Only 1 tab had an active session at the time of the drain.

## Observed behavior

1. With Warp running on a second computer (~144 total tabs, mostly idle), GitHub GraphQL API rate limit was draining at **~2.4 calls/second** (measured via `gh api graphql -f query='{ rateLimit { used remaining } }'` at 60-second intervals).

2. No other applications were making GitHub API calls — verified by:
   - Killing Chrome on both machines (no change in drain rate)
   - Confirming zero GitHub network connections from the primary machine via `lsof`
   - Confirming no `gh` CLI processes running
   - Confirming Docker containers were not the source (restarted Docker with no change)

3. **Quitting Warp on the second computer immediately stopped the drain** (rate went to 0.0/sec).

4. **Reopening Warp fresh** (with restored windows/tabs from previous sessions) caused an immediate burst of ~100+ GraphQL calls, then continued sustained polling at ~2/sec.

5. **Important control**: Warp was also running on the primary computer during the entire investigation with ~12 inactive tabs and 1 active tab — this instance did NOT cause any measurable drain. The difference was likely the number of tabs/windows or something about the restored session state on the second computer.

6. **Reopening Warp a second time** on the second computer (after previously quitting and clearing the session state) showed near-zero drain — suggesting the issue is tied to restored/stale session state rather than Warp itself.

## Reproduction

Difficult to reproduce precisely, but the conditions were:
- Many Warp windows/tabs (~144 total)
- Tabs previously used with CLI tools that interact with GitHub (`gh` CLI, Claude Code, Codex CLI)
- Most tabs idle with no running process
- `gh` CLI authenticated to GitHub

## Impact

[...]

PR #10361, opened 2026-05-07
Stop re-firing gh pr view on every fs event in Code Review

croaky:
`DiffStateModel.handle_updated_metadata_for_repo` runs on every filesystem event the repo watcher emits. After the branch-change path, a fallback called `refresh_pr_info` whenever `pr_info` was `None`. Branches that legitimately have no PR (and lookups that returned `Ok(None)` from `unwrap_or(None)` on transient failures) keep `pr_info` at `None`, so this fallback re-shelled `gh pr view` on every fs event for the lifetime of the session.

Track whether `refresh_pr_info` has been called for the current branch in a new `pr_info_attempted_for_branch` field. The fallback now fires at most once per branch. Branch change still triggers a fresh refresh because the branch-change path always calls `refresh_pr_info`, which overwrites the attempted branch. After-push refreshes from `refresh_after_git_operation` continue to work for the same reason.

## Description

The chip-refresh side of #9830 was addressed in #10307. This PR addresses the second `gh pr view` call site identified in that issue — the Code Review header / `DiffStateModel`. Same bug shape: a network call gated only by per-fs-event metadata refresh, with no caching of the attempted state.

The new `pr_info_attempted_for_branch` field is set inside `refresh_pr_info` itself, so any caller (branch change, after-push, fallback) marks the branch as attempted. The fallback's existence is preserved (initial-load case where metadata arrives without a successful PR lookup yet); only the spam is removed.

Other code paths are unaffected:
- The branch-change refresh in `handle_updated_metadata_for_repo` always fires `refresh_pr_info`.
- The explicit `refresh_pr_info` from `refresh_after_git_operation` (post-commit, post-push) always fires.
- `metadata_refresh_enabled` and the `GitOperationsInCodeReview` feature flag continue to gate the entry points as before.

## Linked Issue

Related to #9830 (complements #10307).

## Testing

No new automated tests. The change is small and the existing `gh`-shelling path is hard to test without subprocess mocks. Verified by:

- Reading the call graph: `handle_updated_metadata_for_repo` is called from `set_active_repository`, `refresh_diff_metadata_for_current_repo` (throttled fs-event handler), and post-git-operation refreshes. The fallback at the modified site is the only path that previously re-fired without a branch change.
- `pr_info_attempted_for_branch` is set inside `refresh_pr_info` before the spawn, so it's recorded even if the lookup is aborted or fails.

## Agent Mode

CHANGELOG-BUG-FIX: Stop the Code Review pane from re-running `gh pr view` on every filesystem event when the current branch has no PR.

### DIFF
--- a/app/src/code_review/diff_state.rs
+++ b/app/src/code_review/diff_state.rs
@@ -374,6 +374,12 @@ pub struct DiffStateModel {
     computing_diffs_abort_handle: Option<SpawnedFutureHandle>,
     computing_metadata_abort_handle: Option<SpawnedFutureHandle>,
     refreshing_pr_info_handle: Option<SpawnedFutureHandle>,
+    /// Branch name for which `refresh_pr_info` has been called at least once.
+    /// Gates the metadata-refresh fallback so it only retries `gh pr view` once
+    /// per branch when `pr_info` is `None` (e.g. branch has no PR, lookup
+    /// failed). Without this, every filesystem event on the repo would re-fire
+    /// `gh pr view` for branches that legitimately have no PR.
+    pr_info_attempted_for_branch: Option<String>,
     /// Controls whether periodic throttled metadata refresh is active.
     /// Refresh is suppressed when the code review pane is not open.
     metadata_refresh_enabled: bool,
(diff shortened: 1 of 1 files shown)


kevinchevalier:
Great fix. I'm going to check this out to test locally and then review in full.

kevinchevalier (approved):
Thanks for this fix!

kevinchevalier:
@croaky Thanks again. I'm going to cherry-pick this into today's release to get this out ASAP.

croaky:
Thanks, @kevinchevalier!

PR #10363, opened 2026-05-07
Resolve agent skills during CLI startup

bnavetta:
## Description

Named agents will have attached skills, which Oz ensures are available on every cloud run. To implement this, we're:

1. Fetching the list of agent skills from the server (via the `User.globalSkills` property, which is only populated for agents)
2. Cloning the repos for each skill, using the same approach as for GitHub repos in an environment
3. Scanning each cloned repository (environment or agent-attached) for skills

As a side effect of this, I fixed a race condition where we would not properly wait for repository indexing to complete before scanning for skills. This was masked by waiting on codebase indexing, which typically (but not always) took long enough that skills would be indexed too. Fixing this depends on #25393, the parent to this PR.

There are a couple obvious followups, which are not in scope here:

1. Supporting global skills for users
2. Supporting global skills in the desktop app, such as by cloning to a managed `~/.warp/skills` directory and making them available regardless of a session's working directory
3. Periodically refreshing skill repositories

Unlike with skills from repositories in the run's environment, *only* the explicitly-requested skills are loaded. This lets the user attach marketplace-like repos with many skills, without bloating the agent's skillset. The resolution behavior is:
1. All skills from repositories in the environment are loaded
2. All skills in the current and home directories are loaded
3. Any explicitly-attached agent skills are loaded

## Testing

* Updated unit tests for resolution logic
* Manually tested e2e that the agent had access to attached skills, but not other skills in the same repo

## Agent Mode

### DIFF
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -29,6 +29,7 @@ use ai::skills::ParsedSkill;
 use warp_cli::agent::{Harness, OutputFormat};
 use warp_cli::mcp::MCPSpec;
 use warp_cli::share::ShareRequest;
+use warp_cli::skill::SkillSpec;
 use warp_core::{features::FeatureFlag, report_error, report_if_error, safe_debug, safe_info};
 use warp_graphql::ai::AgentTaskState;
 use warp_managed_secrets::ManagedSecretValue;
@@ -41,7 +42,9 @@ use crate::ai::blocklist::task_status_sync_model::TaskStatusSyncModel;
 use crate::ai::document::ai_document_model::{AIDocumentModel, AIDocumentModelEvent};
 use crate::ai::llms::{LLMId, LLMPreferences};
 use crate::ai::mcp::{JSONMCPServer, MCPServerState};
-use crate::ai::skills::{resolve_skill_repos, SkillManager, SkillWatcher};
+use crate::ai::skills::{
+    filter_explicit_global_skills, resolve_skill_repos, SkillManager, SkillWatcher,
+};
 use crate::ai::{
     agent::conversation::AIConversationId,
     agent_sdk::driver::harness::{
@@ -358,6 +361,22 @@ pub struct Task {
     pub mcp_specs: Vec<MCPSpec>,
     /// Which harness to use for executing the agent run.
     pub harness: HarnessKind,
+}
+
+struct GlobalSkillResolution {
+    specs: Vec<SkillSpec>,
+    repos: Vec<GithubRepo>,
+}
+
+/// Configuration for which skills in a repository should be loaded.
+enum SkillRepoLoadMode {
+    All,
+    ExplicitGlobal(Vec<SkillSpec>),
+}
+
+struct SkillRepoLoadRequest {
+    repo: GithubRepo,
+    load_mode: SkillRepoLoadMode,
 }
 
 /// Prompt that we initialize an agent driver with. Can represent either a local prompt or
@@ -1229,23 +1248,28 @@ impl AgentDriver {
         })
     }
 
-    /// Resolve additional GitHub repositories that should be cloned to provide
-    /// agent skills.
-    async fn resolve_global_skill_repos(
+    /// Resolve global skill specs and the GitHub repositories that should be cloned for them.
+    async fn resolve_global_skills(
         foreground: &ModelSpawner<Self>,
-    ) -> Result<Vec<GithubRepo>, AgentDriverError> {
+    ) -> Result<GlobalSkillResolution, AgentDriverError> {
         if !FeatureFlag::OzPlatformSkills.is_enabled() {
-            return Ok(Vec::new());
+            return Ok(GlobalSkillResolution {
+                specs: Vec::new(),
+                repos: Vec::new(),
+            });
         }
 
-        let global_specs = foreground
+        let raw_global_specs = foreground
             .spawn(|_, ctx| AuthStateProvider::as_ref(ctx).get().global_skills())
             .await?;
-        let global_repos = resolve_skill_repos(&global_specs);
+        let (global_specs, global_repos) = resolve_skill_repos(&raw_global_specs);
         if !global_repos.is_empty() {
             log::info!("Resolving {} global skill repo(s)", global_repos.len());
         }
-        Ok(global_repos)
+        Ok(GlobalSkillResolution {
+            specs: global_specs,
+            repos: global_repos,
+        })
     }
 
     /// Clone all passed-in global skill repositories.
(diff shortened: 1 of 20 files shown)


app/src/ai/agent_sdk/driver.rs
```diff
@@ -1177,6 +1200,253 @@ impl AgentDriver {
...
+        for repo in global_skill_repos {
+            let repo = repo.clone();
+            let repo_for_log = repo.clone();
+            let clone_future = foreground
+                .spawn(move |me, ctx| {
+                    let working_dir = me.working_dir.clone();
+                    me.terminal_driver.update(ctx, |_, ctx| {
+                        let spawner = ctx.spawner();
+                        async move {
+                            environment::ensure_repo_cloned(
```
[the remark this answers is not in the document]
bnavetta: I did some research with Agent Mode:

> To address the SkillWatcher bypass (global-only repos having ALL skills loaded instead of just the requested ones):
> 
> **Option 1: Skip `DetectedRepositories` for global-only repos.** Clone the repo but don't register it. Read skills directly from disk via `read_skills` + `SKILL_PROVIDER_DEFINITIONS` paths, filter, and add to `SkillManager`. Environment repos keep the current full registration path. Cleanest separation, but means global-only repos don't get file watching or codebase indexing (likely fine — they're just skill sources).
> 
> **Option 2: Remove the `CloudEnvironmentPrep` subscription in `SkillWatcher` and filter at query time.** Keep `DetectedRepositories` registration (needed for environment repos), but remove the redundant `CloudEnvironmentPrep`-specific subscription. Then filter out non-requested global skills in `get_skills_for_working_directory`. Requires `SkillManager` to know which repos are global-only and which skills were explicitly requested.
> 
> **Option 3: Have `load_skills_from_repos` prune excess skills.** After the SkillWatcher auto-scan loads everything, have the driver's filtered load explicitly remove skills from global-only repos that weren't requested. Fragile — races with the async SkillWatcher scan.
> 
> **Option 4: Accept the current behavior.** In cloud environments, `is_cloud_environment = true` makes all directory skills available regardless. The extra skills from global-only repos would be visible anyway. The filtering is effectively cosmetic in this context. Least work, but doesn't match the apparent intent of `filter_explicit_global_skills`.
> 
> Option 1 is the most robust. Option 4 is pragmatic if the filtering isn't actually important.

I'm inclined to go with Option 1, where we find+register global skills without relying on scanning the full repo. This should also be a lot faster when using marketplace-like repos, since we don't have to wait to build the index.

@peicodes @ianhodge does that sound reasonable to you?

bnavetta:
Adding @jasonkeung as a secondary reviewer!

ianhodge (approved):
took another pass - looks good to me!

app/src/ai/agent_sdk/driver.rs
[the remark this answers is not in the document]
bnavetta: Updated to use Option 1. Here's a conversation where the skill context shows that, out of all the `warpdotdev/docs` skills, only `afdocs-audit` is available. We also load all skills from the other repos in the environment: https://staging.warp.dev/debug/maa/79a36cb0-bca2-490c-8aee-6019978dabf7?request=57269baa-66df-427d-9f00-d46e97236102

app/src/ai/skills/file_watchers/skill_watcher.rs
```diff
@@ -191,27 +187,6 @@ impl SkillWatcher {
             }
         });
 
-        // Subscribe to DetectedRepositories to watch repos registered via CloudEnvironmentPrep.
```
bnavetta, line 194: AFAICT this isn't needed anymore, since registering the repo with `RepoMetadataModel` will lead to us scanning it for skills after indexing completes in the `RepoMetadataEvent::RepositoryUpdated` case above.

I confirmed via a MAA request that we _do_ successfully scan for all skills in environment repos without this.

specs/REMOTE-1453/TECH.md
```diff
@@ -0,0 +1,65 @@
...
+- `app/src/ai/skills/skill_manager.rs`: stores the loaded skills for the active run.
+- `app/src/ai/agent_sdk/driver/environment.rs`: environment repository clone helper used by environment prep.
+- `app/src/ai/cloud_environments/mod.rs`: `GithubRepo { owner, repo }` and `AmbientAgentEnvironment.github_repos`.
+- `app/src/ai/agent_sdk/driver.rs`: agent startup sequence, environment prep, and Oz-only repo skill loading.
+## Implemented approach
+### 1. Global skill parsing helpers
+`app/src/ai/skills/global_skills.rs` owns the reusable parsing and filtering helpers and is re-exported through `app/src/ai/skills/mod.rs`.
+- `resolve_skill_repos(raw_specs: &[String]) -> (Vec<SkillSpec>, Vec<GithubRepo>)` parses raw server strings with the existing CLI parser, logs warn-level parse failures, skips invalid entries, and extracts org-qualified `(org, repo)` pairs from the parsed specs while preserving first-seen order.
+- `filter_explicit_global_skills(repo_path, skills, specs)` takes the full scan result for one global-only repo and returns only the parsed skills explicitly requested for that repo.
```
vkodithala: Don't see this in `global_skills.rs` anymore, assuming this was replaced by `filter_skills_by_spec`? Not sure if there was a behavior change (functionally looks similar to what this promises), but would be nice to update specs and enumerate anything that changed!

specs/REMOTE-1453/TECH.md
```diff
@@ -0,0 +1,65 @@
...
+Filtering behavior:
+- Full-path specs match `repo_path.join(spec.skill_identifier)` exactly.
+- Simple-name specs match parsed `ParsedSkill::name` values and pick the first match by provider-directory precedence.
+- Duplicate selected paths are collapsed.
+### 2. Driver source classification
+`app/src/ai/agent_sdk/driver.rs` separates global skill resolution from repository skill loading.
+- `GlobalSkillResolution` carries parsed global specs and cloneable global skill repos.
+- `SkillRepoLoadRequest` carries one repo plus a `SkillRepoLoadMode`, so a repo either loads all discovered skills or filters to explicit global specs without representing contradictory states.
```
vkodithala: This looks deprecated, asked AM and seems like we replaced this with explicit `load_environment_skills` and `load_global_skills` functions? Functionally I think the current implementation makes sense, except maybe that we aren't explicitly deduping repos added for both global specs and to the current environment anymore. 

Seems fine since `handle_skills_added` is idempotent, but maybe we could add a quick test here? i.e. add a repo with skills to the environment, add a subset of those skills to the user's global skills config, and make sure we aren't keeping dups? OOS and completely up to you.

vkodithala (approved):
LGTM, added a few comments but mostly OOC and nothing blocking! Excited for this to go out!

app/src/ai/agent_sdk/driver.rs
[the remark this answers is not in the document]
vkodithala: Yeah, agree that Option 1 makes the most sense.

We probably won't need file watching for globally-added skills anyways, since we expect these to be static after initial load (a user manually adding/deleting/editing those skills shouldn't change their contents in any case IMO, and is probably a massive edge case anyways).

specs/REMOTE-1453/TECH.md
bnavetta: Right, will update the spec! The main difference, aside from some renames, is removing the unified loading path in favor of separate cases for environment skills and global skills - once we weren't using the in-memory index for global skills, the unified path was pretty hard to follow

specs/REMOTE-1453/TECH.md
bnavetta: Yeah, updating the spec and adding a test!

peicodes (approved):
OOC - how do global skills get loaded into `skill_manager`?

bnavetta:
> OOC - how do global skills get loaded into skill_manager?

They're added directly via `handle_skills_added`, as a one-time startup thing. I think it's worth exploring whether or not this generalizes to a skill marketplace construct (or per-user skills that are synced in all cases), but didn't want to expand the scope _too_ much for now.

PR #10384, opened 2026-05-07
Add custom warping verbs (spinner verbs) feature

### LINKED ISSUE #7342: Feature Request: Allow customization of startup 'warping...' message
## Summary
I would like to be able to customize the 'warping...' message that appears when Warp is starting up or initializing.

## Proposed Solution
Add a configuration option in the user preferences (similar to how themes and other UI elements are customizable) that allows users to:
1. Set a custom startup/loading message
2. Optionally disable the startup message entirely
3. Perhaps include some preset fun alternatives alongside the custom option

## Current Behavior
The 'warping...' message is hardcoded and cannot be changed through any configuration options.

## Expected Behavior
Users should be able to customize this message through:
- Settings UI in Warp
- Configuration file editing
- CLI parameter (optional)

## Additional Context
This would be a nice quality-of-life improvement that adds to Warp's already excellent customization options. Many users enjoy personalizing their development environment, and this would be another way to make Warp feel uniquely theirs.

Platform: Linux (Ubuntu)
Warp Version: Preview build

eluzzi5:
## Description
Adds custom spinner verbs for the generic in-progress text shown while Warp Agent or Oz is working. Users can keep the default `Warping...` text, choose a built-in preset pack, or enter custom comma-separated phrases in AI settings. Tool-specific status messages, such as searching files or generating diffs, remain unchanged.

Changes:
- Added the `CustomWarpingVerbs` feature flag.
- Added `agents.warp_agent.spinner_verbs` and `agents.warp_agent.custom_spinner_verbs` settings with normalized setting types.
- Added built-in preset packs: Medieval, Conspiracy, Cooking, and Warpy.
- Added the Settings UI mode selector and custom phrase editor for spinner verbs.
- Updated the status bar loading indicator to resolve the selected spinner verb and keep the displayed phrase stable for each response stream.
- Added grapheme-aware custom phrase truncation so emoji, ZWJ sequences, and combining marks are not split.
- Cached normalized custom phrases and avoided repeating the previous verb when alternatives are available.
- Hardened custom spinner settings persistence so empty editor input does not wipe a saved custom list unexpectedly.
- Updated bundled skill/test coverage for spinner verb settings.
- Updated custom inference copy to clarify that API keys are stored locally and never synced to the cloud.

## Linked Issue
Closes #7342
- [x] The linked issue is labeled ready-to-spec or ready-to-implement.

## Testing
- [x] `./script/format`

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp AI Agent Mode

## Screenshots
Screenshot 2026-05-18 at 4 53 54 PM
Screenshot 2026-05-18 at 4 56 32 PM
Screenshot 2026-05-18 at 4 56 53 PM


peicodes:
I'm probably not the right person to review this PR - I haven't touched any of the related code. Let's go through Oz's automated review process until it auto-assigns a reviewer based on STAKEHOLDERS.

Which repo was this copied from? How much of it was modified?

The original issue was marked ready to spec but not ready to implement, which means we're expecting a `PRODUCT.md` and a `TECH.md`. For internal contributors, it's probably fine to have the specs alongside the code, but we do expect to see those especially if the PR is agent-generated.

We should also include screenshots and videos for visual changes in PR descriptions.

app/src/ai/blocklist/block/status_bar.rs
```diff
@@ -1324,6 +1366,52 @@ impl View for BlocklistAIStatusBar {
     }
 }
 
+#[cfg(test)]
```
lucieleblanc, line 1348: Let's move these tests to a test file, `status_bar_tests.rs`, instead of leaving them inline in the main `status_bar.rs` source code.

specs/GH7342/product.md
```diff
@@ -0,0 +1,73 @@
...
+8. Each generic warping session chooses one phrase from the normalized list at random. The phrase stays stable for that session. When alternatives exist, the next session avoids immediately repeating the previous raw phrase.
+9. A "warping session" is keyed by the active response stream when available, falling back to the exchange key. One backend response that appends multiple exchanges should keep the same displayed phrase.
+10. Rendering appends `...` to a selected phrase unless it already ends in `.`, `!`, `?`, or `…`.
+11. Custom values are normalized before persistence and again at the renderer boundary so direct settings-file edits and synced values cannot bypass validation:
+    - Trim leading and trailing whitespace.
+    - Drop empty entries.
+    - Strip trailing `.` and `…` characters before display formatting.
+    - Drop entries that become empty after trimming/stripping, including dots-only entries.
+    - Sentence-capitalize the first character without title-casing the whole phrase.
```
lucieleblanc: nit: do users always want this? I'm not convinced changing the capitalization is necessary

app/src/ai/loading/warping_verb.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// Fallback display text shown when no custom verbs are configured.
+pub const DEFAULT_WARPING_VERB: &str = "Warping...";
+
+/// Maximum number of custom verbs allowed in the persisted list.
+pub const MAX_CUSTOM_WARPING_VERBS: usize = 50;
+
+/// Maximum display length (in chars, not bytes) of a single verb before the
+/// trailing ellipsis is appended. Longer entries are truncated.
+pub const MAX_WARPING_VERB_CHARS: usize = 40;
```
lucieleblanc, line 27: Out of curiosity, how was this number chosen? Are we trying to avoid the spinner text pushing other items out of the status bar?

specs/GH7342/product.md
```diff
@@ -0,0 +1,73 @@
...
+   - `Warpy` — writes the built-in warpy pack into the custom list.
+   - `Custom` — shows an editor for a comma-separated custom list.
+5. Built-in packs are curated, read-only source-code lists. Applying a pack copies that pack's verb list into `agents.warp_agent.custom_warping_verbs`; it does not create a separate pack reference in settings.
+6. Custom values are entered as comma-separated phrases, for example `Cooking, chopping, slicing`.
+7. The UI saves custom editor content on blur or Enter, not on every keystroke. While the user has unsaved edits, the UI keeps showing the Custom editor and does not overwrite the in-progress text with external setting changes. Once the user saves or selects another mode, later settings changes can resync the editor.
+8. Each generic warping session chooses one phrase from the normalized list at random. The phrase stays stable for that session. When alternatives exist, the next session avoids immediately repeating the previous raw phrase.
+9. A "warping session" is keyed by the active response stream when available, falling back to the exchange key. One backend response that appends multiple exchanges should keep the same displayed phrase.
```
lucieleblanc: To confirm my own understanding: the term "session" is a little overloaded, but if I understand correctly, a "warping session" is a single response from the LLM, not a full terminal session or conversation. This means we might see multiple spinner verbs within a single conversation.

specs/GH7342/tech.md
```diff
@@ -0,0 +1,119 @@
...
+- External settings-file edits while the custom editor has and does not have unsaved changes.
+- Generic warping text in local Warp Agent and Oz/cloud-agent flows.
+- Specific status messages and fallback-model messages overriding custom verbs.
+Suggested targeted commands:
+- `cargo test --manifest-path /Users/erica/warp/Cargo.toml -p app warping_verb`
+- `cargo test --manifest-path /Users/erica/warp/Cargo.toml -p app warping_verb_pack`
+Before pushing PR updates, run formatting and clippy using the repository-standard commands.
+## 10. Risks and Mitigations
+- Risk: custom verbs replace useful status labels. Mitigation: keep custom text only at the final generic fallback branch.
+- Risk: raw synced/settings-file values bypass UI normalization. Mitigation: normalize again in `WarpingVerbSelector::resolve_from_verbs`.
```
lucieleblanc: Ideally, we should only normalize once the editor text is synced or when the settings file has changed. Otherwise, we'll constantly be doing repeated normalization work on each render.

specs/GH7342/tech.md
```diff
@@ -0,0 +1,119 @@
...
+- Risk: the Settings UI shows stale custom text after external updates. Mitigation: track unsaved user edits and sync editor text only when safe.
+- Risk: phrase changes reset shimmer every render. Mitigation: cache one display phrase per session key.
+- Risk: natural-language agents modify third-party spinner config instead of Warp's setting. Mitigation: document `agents.warp_agent.custom_warping_verbs` as the canonical handler in the setting description and bundled settings skill.
+## 11. Follow-ups
+- Product/Design can revise built-in pack contents before broader rollout.
+- A future feature can add a separate hidden/suppress mode if users request it.
+- A future feature can address app-launch/startup splash copy if Product wants that surface to be configurable.
```
lucieleblanc, line 135: Customizing this at startup could be very fun!

app/src/ai/loading/warping_verb.rs
```diff
@@ -0,0 +1,210 @@
...
+        if let Some(cached) = self.cached.borrow().as_ref() {
+            if cached.session_key == session_key {
+                return cached.display.clone();
+            }
+        }
+
+        // Settings file edits and synced values can bypass
+        // AISettings::set_custom_warping_verbs, so normalize again at the
+        // renderer boundary before picking a display value.
+        let verbs = normalize_warping_verbs(verbs.to_vec());
```
lucieleblanc: I think this should happen earlier in the pipeline, ideally when the custom warping verbs are read into `AISettings`. Otherwise we're effectively repeating the normalization work on every render!

app/src/settings/ai.rs
```diff
@@ -2020,6 +2068,41 @@ impl AISettings {
             .plugin_update_chip_dismissed_for_version_map
             .set_value(map, ctx));
     }
+
+    /// Replaces the custom warping verbs list with a normalized version of `verbs`.
+    /// Normalization trims whitespace, sentence-capitalizes entries, drops
+    /// empty entries, truncates over-long entries, and caps the list length. See
+    /// [`crate::ai::loading::normalize_warping_verbs`].
+    pub fn set_custom_warping_verbs(&mut self, verbs: Vec<String>, ctx: &mut ModelContext<Self>) {
+        let normalized = crate::ai::loading::normalize_warping_verbs(verbs);
```
lucieleblanc: style nit: instead of qualifying inline, import `crate::ai::loading` at the top of the file and use `loading::normalize_warping_verbs`, or import `crate::ai::loading::normalize_warping_verbs` and use it without a qualifier.

Same applies to all the `crate::ai::loading::...` in this and other files.

app/src/settings/ai.rs
```diff
@@ -1478,6 +1478,54 @@ define_settings_group!(AISettings, settings: [
...
+For ANY natural-language request like 'change my spinner verbs', 'set my warping verbs', or 'use the <pack> pack', ALWAYS replace the entire list in this setting with exactly what the user asked for. \
+If a request like 'update spinner verbs' does not include a custom list or built-in pack name, ask the user which verbs or pack to use instead of searching. \
+NEVER search the codebase, read source files, or create/modify built-in packs in source code. Built-in packs are read-only. \
+Do NOT modify Claude Code, the Claude Code Warp Plugin, ~/.claude/, ~/.claude-buddy/, or any other plugin's spinner config unless the user explicitly names 'Claude', 'Claude Code', or 'buddy'.",
```
lucieleblanc: This description is very, very long. Let's shorten it to one phrase or sentence, to match the other settings.

app/src/settings/ai.rs
```diff
@@ -1478,6 +1478,54 @@ define_settings_group!(AISettings, settings: [
...
+    //
+    // ## Canonical handler for natural-language "spinner verbs" requests
+    //
+    // This setting is the ONE canonical handler for user prompts like:
+    //   - "change my spinner verbs to A, B, C"
+    //   - "change my spinner verbs to the cooking pack"
+    //   - "set my warping verbs to ..."
+    //   - "use the <pack name> spinner/warping/flavor pack"
+    //   - "customize the warping animation text"
+    //
```
lucieleblanc: I don't think this section of the comment is super necessary, but open to discussion!

app/src/ai/loading/warping_verb.rs
```diff
@@ -0,0 +1,205 @@
...
+    ///
+    /// `session_key` should identify the current warping session. A new verb is
+    /// picked whenever the key changes.
+    pub fn resolve(&self, session_key: &str, app: &AppContext) -> String {
+        if !FeatureFlag::CustomWarpingVerbs.is_enabled() {
+            self.cached.replace(None);
+            return DEFAULT_WARPING_VERB.to_owned();
+        }
+
+        let verbs = AISettings::as_ref(app).custom_warping_verbs.clone();
```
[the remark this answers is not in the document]
lucieleblanc: I agree with this; we should normalize and filter the list before saving it in AISettings. It seems we already do that in `set_custom_warping_verbs`; are there other cases in which we wouldn't call it by then?

app/src/settings_view/ai_page.rs
```diff
@@ -3094,6 +3236,63 @@ impl TypedActionView for AISettingsPageView {
...
+                    }
+                }
+                ctx.notify();
+            }
+            AISettingsPageAction::ApplyWarpingVerbPack(pack) => {
+                self.spinner_verb_mode = SpinnerVerbMode::Pack(*pack);
+                self.custom_warping_verb_editor_has_user_edits = false;
+                self.sync_custom_warping_verb_editor("", ctx);
+                AISettings::handle(ctx).update(ctx, |settings, ctx| {
+                    settings.set_custom_warping_verbs(pack.verbs_as_vec(), ctx);
```
lucieleblanc: From what I understand, this line means that if a user enters a custom verb list in settings, then selects a pack instead, their custom verb list will be gone. Is that the desired behavior?

specs/GH7342/tech.md
```diff
@@ -0,0 +1,119 @@
...
+The implementation customizes only the generic `Warping...` loading label for Warp Agent and Oz. App-launch/startup splash screens and application boot loading surfaces are explicitly out of scope, even if they use similar copy.
+Relevant code:
+- `app/src/settings/ai.rs` defines the public `custom_warping_verbs` setting at `agents.warp_agent.custom_warping_verbs`.
+- `app/src/ai/loading/warping_verb.rs` owns `DEFAULT_WARPING_VERB`, normalization constants, normalization helpers, render-time formatting, and `WarpingVerbSelector`.
+- `app/src/ai/loading/warping_verb_pack.rs` owns the read-only built-in packs.
+- `app/src/ai/blocklist/block/status_bar.rs` owns a `WarpingVerbSelector` and passes the resolved generic fallback string into `WarpingProps`.
+- `app/src/ai/blocklist/block/view_impl/common.rs` keeps action-specific loading labels ahead of `props.default_warping_text`.
+- `app/src/settings_view/ai_page.rs` renders the Settings UI mode buttons and custom editor.
+## 2. Settings Contract
+The single persisted user-facing setting is:
```
lucieleblanc: I'm wondering if a single setting with the custom verb list is the right way to persist this. I see the three options as
```rust
enum SpinnerVerbs {
    Default,
    PresetPack(WarpingVerbPack),
    CustomList(SpinnerVerbList),
}
```

If we want Warp to remember the setting when the app restarts, it'd be nice to know whether the list of spinner verbs came from a preset pack or a custom list. The settings file entries could look something like:

Default:
```toml
[agents.warp_agent]
spinner_verbs = "default"
```

Preset pack:
```toml
[agents.warp_agent]
spinner_verbs = "cooking"
```

Custom:
```toml
[agents.warp_agent]
spinner_verbs = "custom"
custom_spinner_verbs = ["Sautéing", "Braising", "Fermenting"]
```

Alternatively, we could stick with one setting, but with more information in it in the custom case:

Custom:
```toml
[agents.warp_agent]
custom_spinner_verbs = { custom_list = ["Sautéing", "Braising", "Fermenting"] }
```

I could see either of these working well.

app/src/settings/ai.rs
```diff
@@ -1478,6 +1478,54 @@ define_settings_group!(AISettings, settings: [
...
+    // Built-in preset packs (medieval, conspiracy, cooking, warpy) are read-only
+    // and defined in source code. Agents MUST NOT create, modify, or add packs
+    // in source. To apply a pack, write its verb list into this setting.
+    // To handle a custom list, replace the entire setting value — never append.
+    // If no list or pack name is provided, ask the user for values instead of searching.
+    //
+    // An empty list falls back to the built-in "Warping..." display.
+    custom_warping_verbs: CustomWarpingVerbs {
+        type: Vec<String>,
```
lucieleblanc: I think it'd be great to have a custom type for this setting to enforce normalization as soon as the settings file is read.

The inner type should guarantee “this one verb is normalized”:
```rust
pub struct NormalizedSpinnerVerb(String);

impl TryFrom<String> for NormalizedSpinnerVerb {
    type Error = ();

    fn try_from(value: String) -> Result<Self, Self::Error> {
        normalize_warping_verb(&value)
            .map(Self)
            .ok_or(())
    }
}
```

List-level behavior should probably be represented as an outer type:
```rust
#[serde(try_from = "Vec<String>", into = "Vec<String>")]
pub struct SpinnerVerbList(Vec<NormalizedSpinnerVerb>);

impl From<Vec<String>> for SpinnerVerbList {
    fn from(values: Vec<String>) -> Self {
        Self(
            values
                .into_iter()
                .filter_map(|value| NormalizedSpinnerVerb::try_from(value).ok())
                .take(MAX_CUSTOM_WARPING_VERBS)
                .collect(),
        )
    }
}
```
This lets us drop empty entries, cap to MAX_CUSTOM_WARPING_VERBS, dedupe, etc. This is the type of the overall setting.

lucieleblanc (changes requested):
I love this feature! Very excited to use it daily 😄 

There's a few product/behavior questions in here for you. Otherwise, most of my comments are about restructuring the Rust types to avoid repeatedly normalizing/reformatting custom verbs; those changes can likely be automated.

specs/GH7342/tech.md
eluzzi5: yes agreed that could be fun! I think it would make a good follow up in the future

specs/GH7342/product.md
eluzzi5: agreed, updated

app/src/ai/loading/warping_verb.rs
eluzzi5: We are trying to avoid the text getting cut off in the status bar, so we will show an ellipses instead of the text being truncated jarringly. The character size was chosen as a somewhat conservative limit to avoid very long phrases and keep a uniform maximum.

specs/GH7342/product.md
eluzzi5: Yes, correct

app/src/settings_view/ai_page.rs
eluzzi5: No it is not, good catch. Updated!

app/src/settings_view/ai_page.rs
```diff
@@ -1558,6 +1647,40 @@ impl AISettingsPageView {
         ctx.notify();
     }
 
+    fn save_custom_spinner_verbs_from_editor(&mut self, ctx: &mut ViewContext<Self>) {
+        let verbs = self.custom_spinner_verb_editor.as_ref(ctx).buffer_text(ctx);
+        self.custom_spinner_verb_editor_has_user_edits = false;
+        AISettings::handle(ctx).update(ctx, |settings, ctx| {
+            settings.set_custom_spinner_verbs(verbs.split(',').map(str::to_owned).collect(), ctx);
```
lucieleblanc: When a user clears the custom editor and blurs, I think the empty input gets split and normalized to an empty list, which ends up wiping the previously saved custom phrases. That seems to work against the spec's promise that selecting Default or a pack preserves the saved custom list. Could we treat empty input as a no-op (skip persisting when the field is cleared), or switch to Default mode when the normalized list is empty rather than staying in Custom with nothing? Either should keep the preservation guarantee and avoid accidental data loss.

app/src/settings_view/ai_page.rs
```diff
@@ -3094,6 +3231,61 @@ impl TypedActionView for AISettingsPageView {
                     settings.set_cli_agent_for_command(pattern, *agent, ctx);
                 });
             }
+            AISettingsPageAction::SelectSpinnerVerbMode(mode) => {
```
lucieleblanc, line 4238: Clicking a mode button while the custom editor is focused fires `EditorEvent::Blurred` first (which saves and sets mode to Custom), then this handler overwrites the mode. The net result is two persisted writes and two cloud-sync pushes per switch. It's functionally correct, but since these settings sync globally it might be worth consolidating. Could we combine the two into one update?

app/src/ai/loading/warping_verb.rs
```diff
@@ -0,0 +1,224 @@
...
+        None
+    } else {
+        Some(normalized)
+    }
+}
+
+fn truncate_warping_verb(verb: &str) -> String {
+    // Truncate to MAX_WARPING_VERB_CHARS (chars, not bytes, to avoid splitting
+    // multi-byte codepoints).
+    if verb.chars().count() > MAX_WARPING_VERB_CHARS {
```
lucieleblanc: One edge case worth considering: truncating with `chars().take(MAX_WARPING_VERB_CHARS)` can split a multi-codepoint grapheme (emoji, ZWJ sequences, combining marks) right at the 40-char boundary. Since the spec calls out emoji as a validation case, it might be nice to truncate on grapheme boundaries (e.g. via `unicode-segmentation`) or at least add a unit test confirming emoji survive truncation intact.

app/src/ai/loading/warping_verb.rs
```diff
@@ -0,0 +1,224 @@
...
+    /// `session_key` should identify the current warping session. A new verb is
+    /// picked whenever the key changes.
+    pub fn resolve(&self, session_key: &str, app: &AppContext) -> String {
+        if !FeatureFlag::CustomWarpingVerbs.is_enabled() {
+            self.cached.replace(None);
+            self.normalized_cache.borrow_mut().clear();
+            return DEFAULT_WARPING_VERB.to_owned();
+        }
+
+        let verbs = AISettings::as_ref(app).effective_custom_spinner_verbs();
```
lucieleblanc: Some efficiency things: `effective_custom_spinner_verbs()` runs on every render frame while the spinner animates, allocating a fresh `Vec<String>` from the pack verbs each time, even though `NormalizedVerbsCache` already avoids re-normalizing. The allocation is small, so this is low priority, but since it's on a hot animation path you might consider returning borrowed `&'static [&'static str]` for packs, or keying the cache on the `SpinnerVerbsMode` enum plus a generation counter, to skip both the allocation and the per-frame vec comparison.

app/src/ai/loading/warping_verb_pack.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub fn verbs_as_vec(self) -> Vec<String> {
+        self.verbs().iter().map(|v| (*v).to_string()).collect()
+    }
+
+    /// Looks up a pack by its identifier (case-insensitive).
+    pub fn from_identifier(identifier: &str) -> Option<Self> {
+        let needle = identifier.trim().to_ascii_lowercase();
+        Self::all().iter().copied().find(|pack| {
+            pack.identifier() == needle
+                || matches!(*pack, WarpingVerbPack::ConspiracyTheorist) && needle == "conpsiracy"
```
lucieleblanc: Just flagging that `from_identifier` looks unreachable from production code right now (only the tests call it), which is also why it carries `#[allow(dead_code)]`, and it special-cases the misspelling `"conpsiracy"`. If it's meant for the natural-language/skill routing described in the tech spec, it'd be great to wire up a real caller; otherwise we could drop it (and the one-off typo allowance). Happy to be wrong if there's a planned use I'm missing!

app/src/settings/ai.rs
```diff
@@ -399,6 +401,224 @@ impl ThinkingDisplayMode {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for NormalizedSpinnerVerb {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+        Self::try_from(value).map_err(|_| de::Error::custom("invalid spinner verb"))
```
lucieleblanc: A consistency note: `SpinnerVerbList` deserializes via `Vec<String>` and leniently drops invalid entries with `filter_map`, so this strict `Deserialize` for `NormalizedSpinnerVerb` (which errors on an invalid value) never actually runs through the list path. That leaves two paths with slightly different contracts. Could we either remove the strict impl and rely solely on the lenient `SpinnerVerbList` path, or add a comment explaining why both are needed?

app/src/ai/loading/mod.rs
```diff
@@ -2,3 +2,15 @@
...
 pub use shimmering_warp_loading_text::shimmering_warp_loading_text;
+
+mod warping_verb;
+pub use warping_verb::{normalize_warping_verb, WarpingVerbSelector};
+// Re-exported for tests and a planned Settings UI follow-up.
+#[allow(unused_imports)]
+pub use warping_verb::{DEFAULT_WARPING_VERB, MAX_CUSTOM_WARPING_VERBS, MAX_WARPING_VERB_CHARS};
+
+mod warping_verb_pack;
+// Re-exported for use by a planned Settings UI follow-up and by natural-language
```
lucieleblanc: These "planned Settings UI follow-up" comments (and the related `#[allow(unused_imports)]` / `#[allow(dead_code)]`) read as if the UI is still upcoming, but it actually ships in this same PR via `ai_page.rs`. Could we refresh the comments and double-check whether the `allow` attributes are still needed, so they don't end up masking genuinely unused items down the line? Thanks!

eluzzi5:
Hi @moirahuang! I addressed Lucie's comments, but I need another reviewer since she left. This is a fun feature, not an urgent PR :) ty!

app/src/settings/ai.rs
```diff
@@ -398,6 +399,215 @@ impl ThinkingDisplayMode {
...
+    #[default]
+    #[schemars(description = "Use the built-in default.")]
+    Default,
+    /// Use the medieval preset pack.
+    #[schemars(description = "Use the medieval preset pack.")]
+    Medieval,
+    /// Use the conspiracy preset pack.
+    #[schemars(description = "Use the conspiracy preset pack.")]
+    Conspiracy,
+    /// Use the cooking preset pack.
```
moirahuang, line 431: these are so fun

resources/bundled/skills/modify-spinner-verbs/SKILL.md
moirahuang, line 1: ideally we can gate this skill behind dogfood while the feature is in dogfood

app/src/ai/loading/warping_verb.rs
```diff
@@ -0,0 +1,289 @@
...
+            return candidates
+                .choose(&mut rng)
+                .map(|v| (*v).clone())
+                .unwrap_or_else(|| verbs[0].clone());
+        }
+    }
+    verbs
+        .choose(&mut rng)
+        .cloned()
+        .unwrap_or_else(|| verbs[0].clone())
```
moirahuang, line 261: although the `debug_assert` above is helpful for non release builds, we might want a more graceful fallback here is verbs is empty. same think for `pick_static_verb`

app/src/ai/blocklist/block/status_bar.rs
```diff
@@ -1087,11 +1114,27 @@ fn render_fallback_explanation<V: View>(
...
+    }
+    format!("exchange:{exchange_id}")
+}
+
 /// If the current exchange is using a fallback model, returns the warping message to display
 /// (e.g. "Warping with Claude 3.5 Haiku."). When the current exchange's output doesn't have
 /// model info yet (the ModelUsed message hasn't arrived), we check the most recent previous
 /// exchange as a best guess — if the conversation already fell back, the next exchange likely
-/// will too. This avoids a flicker from "Warping..." to "Warping with {name}." on follow-ups.
+/// will too. This avoids a flicker from "warping" to "Warping with {name}." on follow-ups.
```
moirahuang, line 1137: nit: not sure why we're chagning this to `"warping"` since i think we aren't changing that actual behavior?

crates/warp_features/src/lib.rs
```diff
@@ -980,6 +983,7 @@ pub const DOGFOOD_FLAGS: &[FeatureFlag] = &[
     FeatureFlag::CloudRunners,
     FeatureFlag::WaitForEventsParentRegistration,
     FeatureFlag::McpJsonTreeView,
+    FeatureFlag::CustomWarpingVerbs,
```
moirahuang, line 986: we'll also need to add `custom_warping_verbs` to `app/src/features.rs` so we can pass it in on demand to be run. i wanted to do this to test if some of the slowness issues i was running into was because i wasn't on a release build

moirahuang (commented):
[Screen Recording 2026-07-22 at 5.48.24 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/9b864a81-0ec3-4279-81ca-b4add5abfe04.mov)1\. i'm noticing that there's sometimes a lag when switching between packs



1. would be nice to be able to scroll + show a scrollbar if the list gets long
2. if i had a super long custom word and then switch away and then switch back, it's now cropped. i'm not opposed to us cropping but maybe we can surface a toast or something to say the input text was too long? or if we want to keep the long custom text i'm also ok with that!

PR #10391, opened 2026-05-07
introduce PR and repository info to input context

yuanben-warp:
## Description
I introduced PR and repository information to the client-sent agent request input context so that server is able to obtain this information through input context of the agent query for various usages, such as enterprise analytics. Specifically, the information below is included:
- PR related fields: they are sourced from `GitRepoStatusModel::pr_info()`. 
  - PR number
  - state (enum of `open_draft`, `open`, `closed`, `merged`)
  - base_branch
- repository related fields: they are sourced from terminal’s current working directory using local git discovery and origin remote parsing.
  - name (e.g. warp for repo `github.com/warpdotdev/warp`)
  - owner  (e.g. warpdotdev for repo `github.com/warpdotdev/warp`)

**_Note_**:  Since PR prompt chip values are used mainly for UI display, I did not source PR related fields from that. Instead, I used  `GitRepoStatusModel::pr_info()` directly, which powers the prompt chip values. Because `GitRepoStatusModel` only fetches PR info when something registers as a pr_info_consumer, I updated `TerminalView` to register the agent context path as an additional PR-info consumer, separate from the existing chip UI needs.

**_Note_**: the `state` field we included in the context is a computed field from the original `state` and `draft` fields of the `gh pr view` command. We made this computation to avoid introducing invalid combination of the two fields in our proto (e.g. draft is true while state is merged).

## Linked Issue

## Testing

### Recording on local testing

https://github.com/user-attachments/assets/4471a5e4-5d63-42f7-8610-aa182a1e7f6d

Added unit tests covering:
- Repository URL parsing (HTTPS, scp-style SSH, `ssh://` SSH, fallback, multi-segment defense).
- Conversion of `Repository` / `PullRequest` AI context to the proto `Git` message, including skipping invalid PR numbers and legacy string-encoded numbers.
- Structured PR chip parsing from JSON, legacy URL fallback, and rejection of invalid numbers without URL fallback.
- Stale chip clearing on chip disable, cached failure suppression, and pre-refresh.

## Agent Mode


app/src/context_chips/mod.rs
```diff
@@ -85,16 +95,99 @@ impl std::fmt::Display for ChipValue {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GithubPullRequestChipValue {
+    #[serde(default)]
```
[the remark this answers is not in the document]
yuanben-warp: Resolved

app/src/terminal/view.rs
```diff
@@ -4914,15 +4926,23 @@ impl TerminalView {
                     .contains(&ContextChipKind::GithubPullRequest))
     }
 
+    #[cfg(feature = "local_fs")]
+    fn needs_pr_info_for_agent_context(&self) -> bool {
+        self.current_local_repo_path().is_some()
```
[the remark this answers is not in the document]
yuanben-warp: Resolved by introducing a check to see if the AI input mode is enabled.

app/src/ai/agent/api/convert_to.rs
```diff
@@ -803,12 +803,35 @@ fn convert_context(context: &[AIAgentContext]) -> api::InputContext {
...
             AIAgentContext::Git { head, branch } => {
-                api_context.git = Some(api::input_context::Git {
-                    head,
-                    branch: branch.unwrap_or_default(),
-                    repository: None,   // TODO: populate?
-                    pull_request: None, // TODO: populate?
-                });
+                let git_context = api_git_context(&mut api_context);
+                git_context.head = head;
+                git_context.branch = branch.unwrap_or_default();
```
peicodes: I feel like it makes more sense for `api_git_context` to just accept all the args it needs to create a full git context

nit: rename to `create_api_git_context`

app/src/ai/agent/api/convert_to.rs
```diff
@@ -803,12 +803,35 @@ fn convert_context(context: &[AIAgentContext]) -> api::InputContext {
...
+                draft,
+                base_branch,
+            } => {
+                if number <= 0 {
+                    continue;
+                }
+                let Some(state) = api_pull_request_state(&state, draft) else {
+                    continue;
+                };
+                api_git_context(&mut api_context).pull_request =
```
peicodes: Does this assignment actually still return the object? If it does, I still think we should separate assignment + return to make it clear to the reader what's happening

app/src/ai/agent/api/convert_to.rs
```diff
@@ -829,6 +852,29 @@ fn convert_context(context: &[AIAgentContext]) -> api::InputContext {
...
+/// Returns `None` for unknown states so the caller can skip emitting a
+/// `pull_request` sub-message rather than sending `STATE_UNSPECIFIED` to the
+/// server.
+fn api_pull_request_state(
+    state: &str,
+    draft: bool,
+) -> Option<api::input_context::git::pull_request::State> {
+    use api::input_context::git::pull_request::State;
+    match state.to_ascii_uppercase().as_str() {
+        "OPEN" if draft => Some(State::OpenDraft),
```
peicodes: nit: I think this woudl be more readable if we did:
```
"OPEN" => if draft { Some(State::OpenDraft) } else { Some(State::Open) }
```
instead of having a conditional match + fallthrough

app/src/ai/agent/mod.rs
```diff
@@ -2047,6 +2071,37 @@ pub enum AIAgentContext {
     Block(Box<BlockContext>),
 }
 
+fn deserialize_pull_request_number<'de, D>(deserializer: D) -> Result<i32, D::Error>
```
peicodes, line 2074: This is a utility with parsing logic, so it should probably have tests

app/src/ai/blocklist/context_model.rs
```diff
@@ -282,6 +290,8 @@ impl BlocklistAIContextModel {
         Self {
             terminal_model,
             directory_context: Default::default(),
+            #[cfg(feature = "local_fs")]
```
peicodes: question: do we strictly need these flags to pass CI? 

nit: I ask because while it's technically true that no fs = no git, I usually prefer to use it only if the match arm makes a meaningful impact on bundle size or compute because there's a maintenance/complexity cost to include these

app/src/code_review/diff_state/local.rs
```diff
@@ -1418,7 +1418,7 @@ impl LocalDiffStateModel {
             self.load_diffs_for_current_repo(false, false, ctx);
         }
         if let Some(metadata) = self.metadata.clone() {
-            ctx.emit(DiffStateModelEvent::MetadataRefreshed(metadata));
+            ctx.emit(DiffStateModelEvent::MetadataRefreshed(Box::new(metadata)));
```
peicodes, line 1421: Why the box?

peicodes (approved):
Left some comments and nits but nothing strictly blocking. If you haven't already, let's make sure we test any existing functionality like the git chips

app/src/code_review/diff_state/local.rs
yuanben-warp: I believe the main reason is without it DiffStateModelEvent would become as large as DiffMetadata, even for tiny variants like CurrentBranchChanged. Additionally, this may trigger clippy::large_enum_variant, and with -D warnings that can fail presubmit/CI.
I think it's not required to have it, but it's more of a guard for enum-layout/performance/clippy concern. What's your thought on this?

app/src/ai/blocklist/context_model.rs
yuanben-warp: I removed this one alongside with some other ones, but let me know if I overdid that 👍

app/src/ai/blocklist/context_model.rs
```diff
@@ -992,6 +1007,45 @@ impl BlocklistAIContextModel {
...
     }
 
+    pub fn set_git_repo_status(&mut self, handle: Option<WeakModelHandle<GitRepoStatusModel>>) {
+        self.git_repo_status = handle;
+    }
+
+    /// Builds an `AIAgentContext::Repository` from the current working directory's git remote
+    /// metadata, if available. Only available on non-WASM targets (git2 is not available on WASM).
+    #[cfg(not(target_family = "wasm"))]
+    fn repository_context(&self) -> Option<AIAgentContext> {
```
kevinyang372: Are these supposed to be invoked on the main thread? This seems like it should be an async operation

app/src/ai/blocklist/context_model.rs
```diff
@@ -1005,6 +1059,70 @@ impl BlocklistAIContextModel {
     }
 }
 
+/// Parses a repository name and optional owner from a git remote URL.
+#[cfg(not(target_family = "wasm"))]
+fn parse_repo_name_and_owner(url: &str) -> Option<(String, String)> {
+    let url = url.trim().trim_end_matches(".git");
+    parse_scp_like_ssh_remote_path(url)
```
kevinyang372: Does this handle the possible repo URLs comprehensively? https://git-scm.com/docs/git-clone/2.44.0#_git_urls

kevinyang372: Also do we have to rely on parsing git remote URLs? Is there any alternative here?

app/src/context_chips/mod.rs
```diff
@@ -44,13 +44,15 @@ use crate::ui_components::icons::Icon;
 pub enum ChipValue {
     Text(String),
     GitDiffStats(display_chip::GitLineChanges),
+    GithubPullRequest(GithubPullRequestChipValue),
```
kevinyang372: I don't quite understand why we need to have a new chip value here for context chips for the purpose of this PR?

app/src/terminal/view.rs
```diff
@@ -4819,6 +4820,9 @@ impl TerminalView {
                 });
             }
         });
+        self.ai_context_model.update(ctx, |context_model, _| {
```
kevinyang372, line 4823: I also don't quite understand the logic changes here

app/src/remote_server/diff_state_proto.rs
```diff
@@ -120,6 +120,9 @@ impl From<&proto::PrInfo> for PrInfo {
         PrInfo {
             number: pr.number,
             url: pr.url.clone(),
+            state: String::new(),
```
kevinyang372: Why are we just leaving these to be empty for the remote case?

app/src/ai/blocklist/context_model.rs
yuanben-warp: Yes that makes sense, let me refactor this into `GitRepoStatusModel`

app/src/ai/blocklist/context_model.rs
yuanben-warp: So git naturally does not provide `repo name` or `owner` as first-class metadata. It only has remotes which are URLs. From local git data, parsing seems to be the only concrete sourcing available.
There are some other options with `gh repo view --json name,owner` but I think that would lock in with GitHub (would not work with Gitlab for example). What's your thought on us assuming this more specifically for github rather than including others like gitlab or bitbucket? If that's the case we can probably switch to the `gh` command to obtain this information

app/src/context_chips/mod.rs
yuanben-warp: I think this was originally needed because we were using the PR chip as structured data for AI context. With the refactor to `GitRepoStatusModel`, I think we no longer need PR chip to be structured and I can revert this change.

app/src/terminal/view.rs
yuanben-warp: So this is to give `BlocklistAIContextModel` access to the existing per-repo `GitRepoStatusModel`, so AI requests can include repository / PR context without doing sync git work and without depending on visible context chips. Does this overall make sense?

app/src/remote_server/diff_state_proto.rs
yuanben-warp: For remote case, right now we are sourcing the PR info from GitRepoStatusModel which is local-only and shells out to gh pr view. For remote repos, I believe we don’t currently have a remote PR-info pipeline 🤔
In the remote_server case, is it basically the cloud case? Do you have suggestion here on whether we should implement something to capture that?

app/src/ai/blocklist/context_model.rs
kevinyang372: We already has dependency on gh with our current pr_info implementation right? I don't think we need to separate out repo name or owner to be git-provider agnostic

app/src/remote_server/diff_state_proto.rs
```diff
@@ -115,19 +115,12 @@ impl From<&proto::Commit> for Commit {
     }
 }
 
-impl From<&proto::PrInfo> for PrInfo {
```
kevinyang372, line 118: I don't think we want to modify this part. This is needed for @MaggieShan's work

app/src/terminal/view.rs
kevinyang372: Ah gotcha

kevinyang372 (approved):

app/src/ai/blocklist/context_model.rs
yuanben-warp: Sounds good, let me then switch to use gh repo view to obtain the name and owner information then 👍

app/src/util/git.rs
```diff
@@ -682,6 +682,56 @@ pub async fn run_push(_repo_path: &Path, _branch: &str, _path_env: Option<&str>)
...
+        owner: Some(owner),
+    })
+}
+
+#[cfg(feature = "local_fs")]
+pub async fn get_repository_info(
+    repo_path: &Path,
+    path_env: Option<&str>,
+) -> Result<Option<RepositoryInfo>> {
+    let stdout = run_gh_command(
```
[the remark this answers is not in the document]
yuanben-warp: Discussed on this and we would like to use gh repo view to better capture the repo information than URL parsing

PR #10404, opened 2026-05-07
Add `DiffStateModel` local and remote wrapper

MaggieShan:
## Description
* As part of the work for supporting the code review and git states on a remote environment 
* Adds a diff state wrapper model which stores an enum to either a local or remote `DiffStateModel`
  * Adds template for `RemoteDiffStateModel` (to be filled later) 
  * Updates all callers to use the `DiffStateModel` wrapper instead of the local one
  * Updates `LocalDiffStateModel` to wrap `GitDiffWithBaseContent` with an `Arc` since the `NewDiffsComputed` event is now getting passed from local > wrapper > consumer 
* Removes a bunch of existing dead code in the `DiffStateModel` - a lot of these seem to have been unused/dead for some time now
  * unused debug logs for git operations  
  * get_metadata(), get_main_branch_stats(), get_diff_stats(), diff_mode_for_base_branch(), get_diff_data_for_mode(), label_text(), changes_vs_main_branch_label(), changes_vs_head_label(), UNCOMMITTED_CHANGES, GitDiffInfo
* Updates wiring such that WorkingDirectoriesModel stores a hash map of both local and remote diff states keyed by their local/remote path counter parts
  * Adds `debug_assert` to ensure local path is matched to local diff state model 

## Linked Issue
[APP-4351](https://linear.app/warpdotdev/issue/APP-4351/update-diffstatemodel-api)

## Testing
Should be a no-op, verified existing code review view locally/manually

## Agent Mode

Relevant plan: https://staging.warp.dev/drive/notebook/DiffStateModel-LocalDiffStateModel-rename-wrapper-wznqiCi3HlbQjbDBvoDbqE


app/src/code_review/diff_state/local.rs
```diff
@@ -837,10 +784,6 @@ impl DiffStateModel {
             stash_args.push(path.as_str());
         }
 
-        log::debug!(
```
kevinyang372, line 840: Could we keep these around for now? I know KC uses these debug logs

app/src/code_review/diff_state/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+    Remote(ModelHandle<RemoteDiffStateModel>),
+}
+
+// ── Wrapper ──────────────────────────────────────────────────────────────
+
+/// Wrapper that provides a unified API over local and remote diff state models.
+///
+/// All consumers should interact with this type rather than accessing
+/// [`LocalDiffStateModel`] or [`RemoteDiffStateModel`] directly.
+pub struct DiffStateModel {
```
kevinyang372: Do we actually need an inner here? We could just have the DiffStateModel itself be an enum right?

app/src/code_review/diff_state/mod.rs
```diff
@@ -0,0 +1,374 @@
...
+                local.update(ctx, |local, ctx| {
+                    local.stop_active_watcher(ctx);
+                });
+            }
+            DiffStateBackend::Remote(_) => {}
+        }
+    }
+
+    #[cfg(feature = "local_fs")]
+    pub async fn load_diff_data_for_mode(
```
kevinyang372: Do we need wrapper methods for these static methods? I think it might be better to pull them to app/src/util/git.rs?

app/src/pane_group/working_directories.rs
```diff
@@ -81,10 +82,10 @@ pub struct WorkingDirectoriesModel {
     /// Note, a single root path can be associated with multiple terminals.
     /// we're just storing an arbitrary terminal ID for each root path.
     directory_to_terminal: HashMap<EntityId, HashMap<PathBuf, EntityId>>,
-    /// Global mapping from repository root paths to their DiffStateModel.
+    /// Global mapping from repository keys to their DiffStateModel.
     /// Since git state is inherently tied to a repository (not a pane group),
     /// this is stored globally and shared across all pane groups viewing the same repo.
-    diff_state_models: HashMap<PathBuf, ModelHandle<DiffStateModel>>,
+    diff_state_models: HashMap<BufferLocation, ModelHandle<DiffStateModel>>,
```
kevinyang372: Let's create a wrapper type here?

It can expose an insert_local and insert_remote function that enforces the right combination of key and model

kevinyang372 (approved):

app/src/code_review/diff_state/mod.rs
MaggieShan: Hmm fair - the initial thinking was to keep the local and remote model handles themselves private and encourage only the pub apis but I don't think it's that big of deal

app/src/pane_group/working_directories.rs
MaggieShan: Done!

app/src/code_review/diff_state/mod.rs
MaggieShan: Moved! I think there are a few other local diff state fns that could also be moved to git.rs - can address in a follow up

PR #10409, opened 2026-05-07
feat(harness-support): add report-shutdown command

captainsafia:
## Description

Add a new `report-shutdown` subcommand to `oz harness-support` that reports agent process shutdown to the Oz platform.

This supports the agent reporting shutdown and execution completion discretely at the end of an Oz cloud agent turn. `finish-task` remains the user-facing task outcome signal, while `report-shutdown` lets the runtime close out the active execution after the agent process has finished its post-completion idle period.

This supports both clean shutdown (no error payload) and abnormal shutdown (with `--error-category` and `--error-message` flags).

**Command syntax:**
- Clean shutdown: `oz harness-support --run-id "$OZ_RUN_ID" report-shutdown`
- Abnormal shutdown: `oz harness-support --run-id "$OZ_RUN_ID" report-shutdown --error-category oom --error-message "out of memory"`

**Payload shape:**
- Clean: `{}` (empty JSON body, `error` field is omitted via `skip_serializing_if`)
- Abnormal: `{ "error": { "category": "oom", "message": "out of memory" } }`

**Server endpoint:** `POST /api/v1/harness-support/report-shutdown`

### Files changed:
- `crates/warp_cli/src/harness_support.rs` — CLI argument definitions (`ReportShutdownArgs`, `ReportShutdown` variant)
- `app/src/server/server_api/harness_support.rs` — `HarnessSupportClient::report_shutdown()` trait method and `ServerApi` impl
- `app/src/ai/agent_sdk/harness_support.rs` — Command dispatch
- `app/src/ai/agent_sdk/telemetry.rs` — Telemetry event
- `app/src/ai/agent_sdk/mod.rs` — Telemetry mapping
- `app/src/ai/agent_sdk/driver/snapshot_tests.rs` — TestClient impl
- `crates/warp_cli/src/lib_tests.rs` — CLI parsing tests
- `app/src/server/server_api/harness_support_tests.rs` — Serialization tests

## Linked Issue

N/A — This is part of the Docker sandbox report-shutdown feature.

## Testing

- CLI parsing tests pass for both clean and abnormal shutdown variants
- Serialization tests verify the wire format for both clean (`{}`) and abnormal (`{ "error": { ... } }`) payloads
- `cargo fmt` and `cargo clippy` pass cleanly

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/de2bb31c-8c5f-440d-8373-175529dc7cf2_
_Run: https://oz.staging.warp.dev/runs/019e044c-c561-7931-a402-1fa2ff60bb28_

### DIFF
--- a/app/src/ai/agent_sdk/telemetry.rs
+++ b/app/src/ai/agent_sdk/telemetry.rs
@@ -417,7 +417,9 @@ impl TelemetryEventDesc for CliTelemetryEventDiscriminants {
             | Self::HarnessSupportReportArtifact
             | Self::HarnessSupportNotifyUser
             | Self::HarnessSupportFinishTask
-            | Self::HarnessSupportReportShutdown => EnablementState::Flag(FeatureFlag::AgentHarness),
+            | Self::HarnessSupportReportShutdown => {
+                EnablementState::Flag(FeatureFlag::AgentHarness)
+            }
             Self::ArtifactUpload | Self::ArtifactGet | Self::ArtifactDownload => {
                 EnablementState::Flag(FeatureFlag::ArtifactCommand)
             }
--- a/app/src/ai/agent_sdk/telemetry.rs
+++ b/app/src/ai/agent_sdk/telemetry.rs
@@ -112,6 +112,8 @@ pub(super) enum CliTelemetryEvent {
     HarnessSupportNotifyUser,
     /// Executing `warp harness-support finish-task`
     HarnessSupportFinishTask { success: bool },
+    /// Executing `warp harness-support report-shutdown`
+    HarnessSupportReportShutdown,
 }
 
 impl TelemetryEvent for CliTelemetryEvent {
@@ -188,6 +190,7 @@ impl TelemetryEvent for CliTelemetryEvent {
             CliTelemetryEvent::HarnessSupportFinishTask { success } => {
                 Some(json!({ "success": success }))
             }
+            CliTelemetryEvent::HarnessSupportReportShutdown => None,
         }
     }
 
@@ -273,6 +276,9 @@ impl TelemetryEventDesc for CliTelemetryEventDiscriminants {
             }
             CliTelemetryEventDiscriminants::HarnessSupportFinishTask => {
                 "CLI.Execute.HarnessSupport.FinishTask"
+            }
+            CliTelemetryEventDiscriminants::HarnessSupportReportShutdown => {
+                "CLI.Execute.HarnessSupport.ReportShutdown"
             }
         }
     }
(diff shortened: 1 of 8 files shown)


app/src/ai/agent_sdk/telemetry.rs
```diff
@@ -407,7 +416,10 @@ impl TelemetryEventDesc for CliTelemetryEventDiscriminants {
             Self::HarnessSupportPing
             | Self::HarnessSupportReportArtifact
             | Self::HarnessSupportNotifyUser
-            | Self::HarnessSupportFinishTask => EnablementState::Flag(FeatureFlag::AgentHarness),
+            | Self::HarnessSupportFinishTask
+            | Self::HarnessSupportReportShutdown => {
```
bnavetta: Should this one be gated on `FeatureFlag::AgentHarness`? It doesn't seem tied to multi-harness

app/src/server/server_api/harness_support.rs
```diff
@@ -357,6 +393,22 @@ impl HarnessSupportClient for ServerApi {
...
 
+    async fn report_shutdown(
+        &self,
+        error_category: Option<String>,
+        error_message: Option<String>,
+    ) -> Result<()> {
+        let request = match (error_category, error_message) {
+            (Some(category), Some(message)) => ReportShutdownRequest::abnormal(category, message),
+            (None, None) => ReportShutdownRequest::clean(),
+            (Some(_), None) | (None, Some(_)) => {
```
bnavetta: What do you think of having `report_shutdown` take a `ReportShutdownRequest` as its argument, so this case is unreachable?

bnavetta (approved):
Nice! Implementing this as a separate command so we can handle crashes is a good call.

We might _also_ need to call `/harness-support/report-shutdown` in-process for the local CLI case, so those executions are also marked as ended. We could similarly call this as part of local->cloud handoff flows, so the client can declare to the server that it's safe to hand off its conversation.

app/src/server/server_api/harness_support.rs
captainsafia: That would make this neater, although I don't see this being used for other APIs. I wonder if the right thing to do here is to just expose two distinct methods: `report_clean_shutdown` and `report_errored_shutdown` that handle this.

app/src/ai/agent_sdk/telemetry.rs
captainsafia: Good call!

app/src/server/server_api/harness_support.rs
captainsafia: Not sure I like this more but at least we through earlier if you call the command with the wrong arguments now. 🤷🏽‍♀️

PR #10418, opened 2026-05-07
Scope upward menu positioning to confirmation card only

cephalonaut:
## What
Move upward dropdown positioning (TopLeft/BottomLeft anchors, FilterableDropdownOrientation::Up) out of the shared picker helpers and apply it only in RunAgentsCardView::ensure_pickers. This prevents the plan config card's pickers from also opening upward.

## Why
Follow-up to #10399 — the upward positioning was baked into the shared helpers, affecting both the confirmation card (where it's needed to avoid input overlap) and the plan config card (where it's not needed and looks wrong).

## Demo
https://www.loom.com/share/6fbd59a6a7364368b0eef0922710e3b4

## Agent Mode


cephalonaut:
Missed this on last PR

advait-m (approved):
ah, yep agree downward is better on plan

PR #10423, opened 2026-05-08
Fix a bunch of convo restoration issues, and show shell command run duration

seemeroland:
## Description

This PR refactors the conversation restoration flow and fixes several issues:
- Viewing ambient agent conversation transcripts was broken due to interpreting restored blocks as setup commands
- On quit/restart warp, we relied on blocks to be persisted in sqlite locally. But if we were viewing a transcript before quit/restart, those blocks were restored from the transcript and don't exist locally, so on restart the command blocks would be missing. As a fix, we fall back to restoring blocks from the transcript on startup if blocks don't exist locally in sqlite
- When creating a conversation, closing, and then restoring, we restore from the exchanges in memory rather than derive the exchanges from the task proto. When there's a CLI subagent, there is no run shell command tool call result in an exchange input. Actions are restored from exchange inputs, which means in this case we never restore the action model with a completed action result. The requested command view relied on an action status existing to be expandable. To fix, we no longer require an action result to expand as long as there is an associated block in the blocklist.
- Missing filter block button was due to missing mouse states for restored blocks

### Restoration Refactor

Previously, there were separate flows for restoring **on** view creation and **after** view creation. On view creation, we would extract command blocks and pass them to the terminal model constructor, and compute AI block insertion indices relative to those command blocks. After view creation, we would iterate to create AI blocks and command blocks together (this flow was broken, and did not include block attachments).

We now consolidate so that after view creation, we use the same command block extraction logic and compute AI block insertion indices. The only difference is that the terminal model has already been constructed and may have existing blocks, so we append to the terminal model instead of pass to constructor.
- Note that this means blocks in the terminal model are NOT necessarily sorted by increasing `start_ts`. If we restore a conversation inline later, the restored blocks that are appended could have earlier `start_ts`, like `[t3, t4, t5, t1, t2]` where `t1` and `t2` are restored. There is a unit test covering this case

### Command run duration

We pass the block start_ts and finish_ts as part of the run shell command tool call result, and use this during restoration to set timestamps on restored blocks and compute run duration.

## Testing

Also added some unit tests

### Screenshots / Videos

Showing all the problems that existed before: https://www.loom.com/share/421aade37aac41b5877ed358d1162afd
After this PR, problems are fixed: https://www.loom.com/share/f035540e312c4925a3148f15b10eb268

CHANGELOG-IMPROVEMENT: When viewing conversation transcripts, you can now see the run duration on shell commands and start/end timestamp on hover
CHANGELOG-BUG-FIX: Fixed several conversation restoration issues that prevented you from viewing shell command outputs


app/src/terminal/view.rs
```diff
@@ -21965,10 +21989,10 @@ impl TerminalView {
         };
 
         let start = block.start_ts().map_or_else(String::new, |b| {
-            format!("Started at: {}", b.format("%a %b %-d at %-I:%M %p"))
+            format!("Started at: {}", b.format("%a %b %-d at %-I:%M:%S %p"))
```
seemeroland, line 22002: Show second precision and fix tooltip so it doesn't get clipped by bounds

app/src/terminal/view/load_ai_conversation.rs
```diff
@@ -1285,28 +1201,55 @@ fn command_block_indices_for_exchanges<'a>(
...
+/// timestamp — within the sorted tail, everything earlier is also <=.
+fn find_block_indices_for_exchange_timestamps(
+    command_blocks: &[(BlockIndex, DateTime<Local>)],
+    exchange_timestamps: &[DateTime<Local>],
+) -> Vec<Option<BlockIndex>> {
+    let mut result = Vec::with_capacity(exchange_timestamps.len());
+
+    for &exchange_timestamp in exchange_timestamps {
+        let mut best: Option<(BlockIndex, DateTime<Local>)> = None;
+        for &(idx, ts) in command_blocks.iter().rev() {
```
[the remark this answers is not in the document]
seemeroland: This is true, but I think it's fine. The agent view filters out blocks not for the conversation, so it only really matters that the AI blocks are in the correct position relative to command blocks that will be shown in the agent view

app/src/pane_group/mod.rs
```diff
@@ -5450,6 +5456,16 @@ impl PaneGroup {
     ) {
         let mut conversation_id = None;
         terminal_view.update(ctx, |view, ctx| {
+            // The cloud-mode terminal model starts with
```
harryalbert, line 5457: is there any way we could be a bit smarter about this (at least for the initial setup commands)? IIRC the way that we know to clear this state in a regular run is when the first agent request comes in. maybe we could do the same thing here?

app/src/terminal/view/load_ai_conversation.rs
```diff
@@ -1285,28 +1201,55 @@ fn command_block_indices_for_exchanges<'a>(
...
+/// [`command_block_indices_for_exchanges`].
+///
+/// For each exchange timestamp, finds the command block with the smallest timestamp
+/// that is >= the exchange timestamp. Iterates the command block list backwards
+/// because `insert_restored_block` appends blocks at the end of the blocklist, so
+/// the tail is a sorted group of conversation blocks while the prefix contains
+/// pre-existing terminal blocks we don't want to match against. The backwards scan
+/// tracks the best candidate and stops as soon as it sees a block before the exchange
+/// timestamp — within the sorted tail, everything earlier is also <=.
+fn find_block_indices_for_exchange_timestamps(
```
harryalbert, line 1219: how often/how many times do we run this fn? Just wondering how much work this is on each restoration (especially for very long conversations)

harryalbert (approved):
nice! Just had a few questions

zachbai:
@seemeroland I already fixed the set up command conflict issue with restoring ambient agent conversations causing duplicate commands

seemeroland:
> I already fixed the set up command conflict issue with restoring ambient agent conversations causing duplicate commands

Which PR was that? This was my fix, is it not needed anymore? https://github.com/warpdotdev/warp/pull/10423/changes#r3230158975

app/src/terminal/view/load_ai_conversation.rs
seemeroland: We run it once when restoring any conversation

app/src/pane_group/mod.rs
seemeroland: I think this is right though. The is setup command thing only makes sense for live sessions as blocks come in, but this codepath is just restoring directly into a pane. It should not be set to true

zachbai:
> > I already fixed the set up command conflict issue with restoring ambient agent conversations causing duplicate commands
> 
> Which PR was that? This was my fix, is it not needed anymore? https://github.com/warpdotdev/warp/pull/10423/changes#r3230158975

It was this one https://github.com/warpdotdev/warp/pull/10776

PR #10425, opened 2026-05-08
[APP-3792] set up sqlite for remote codebase indexing caching

moirahuang:
## Description
Set up sqlite for remote server codebase indexing caching based off of the specs in https://github.com/warpdotdev/warp/pull/9508

Mirroring local setup by having per user sqlite in the data dir

## Testing
Unit tests + manually tested by remote deploying and checking for the existence of the tables

### Screenshots / Videos
Screenshot 2026-05-08 at 11 55 31 AM

## Agent Mode


app/src/persistence/remote_codebase_indexing.rs
```diff
@@ -0,0 +1,255 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result};
+
+const REMOTE_CODEBASE_INDEXING_SQLITE_FILE_NAME: &str = "index.sqlite";
+const REMOTE_CODEBASE_INDEXING_DIR_NAME: &str = "codebase-indexes";
+const SHARED_CACHE_DIR_NAME: &str = "shared";
+const SNAPSHOTS_DIR_NAME: &str = "snapshots";
+
+pub fn initialize_remote_codebase_indexing_storage() {
```
moirahuang: in a future PR i might make this be a singleton to manage all of the persistence related fields but i'm keeping this as a function for now to minimize unnecessary code in this PR

kevinyang372 (commented):
Let's chat about this tmrw morning -- two main thoughts:
1. I don't think we need to create a fully new RemoteCodebaseIndexCache. We should be able to re-use our existing tables to do this
2. We should think about if we want our sqlite file to be identity scoped

app/src/persistence/sqlite.rs
```diff
@@ -414,11 +430,61 @@ fn setup_database(database_path: &Path) -> Result<SqliteConnection> {
 /// Integration tests that initialize the database with known data should use
 /// this function to determine where to create the database file.
 pub fn database_file_path() -> PathBuf {
```
kevinyang372: Can we rename this `database_file_path_for_app` or remote this to force every caller to pass in the scope? Right now this seems to be a footgun

kevinyang372 (approved):

app/src/persistence/sqlite.rs
moirahuang: yeah i'll just remove it, i think it's good to force call sites to be explicit

PR #10429, opened 2026-05-08
Strongly typed error handling for remote server SSH installation flow

alokedesai:
## Description
Makes a bunch of improvements to the error UX for the remote code so that:

1) We have strongly typed errors, which allows us to separate out the error message we _log_ from the error we message to the user
2) Improves how we message the error to the user so that it's easier to understand and that the core details actually goes into the big red error box (which is the actual thing anyone is going to read)

https://www.loom.com/share/82f13736fc7741c1b18152b92ddd6047?from_recorder=1&focus_title=1

## Before
image

## After
image

## Agent Mode

---
[Oz conversation](https://staging.warp.dev/conversation/f40d630e-6c90-4f04-92fb-bdcd8dbc73be)

### DIFF
--- a/app/src/remote_server/ssh_transport.rs
+++ b/app/src/remote_server/ssh_transport.rs
@@ -18,8 +18,11 @@ use remote_server::manager::RemoteServerExitStatus;
 use remote_server::setup::{
     parse_uname_output, remote_server_daemon_dir, PreinstallCheckResult, RemotePlatform,
 };
-use remote_server::ssh::ssh_args;
-use remote_server::transport::{Connection, RemoteTransport};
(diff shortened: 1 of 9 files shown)


moirahuang (approved):
i know this isn't the scope of your PR but looking at this is making me realize, do we need some sort of a CTA for when it does fail? i get that often there's not a clear action but in the cases that there are, if i was the use and saw our banner i'd wonder "ok so i have this error, now what?"

alokedesai:
Yup, already planning on doing that as a followup.

It depends on the error type (which requires strongly typed errors)

PR #10443, opened 2026-05-08
fix: refresh IME cursor area after redraw

bet4it:
## Description

This change refreshes the IME cursor area after a successful redraw when IME is enabled.

That keeps the IME preedit/candidate UI aligned with the terminal cursor during redraw-heavy terminal apps.

## Linked Issue

## Testing

### Screenshots / Videos

Before this PR:
Screenshot1
After this PR:
Screenshot2

## Agent Mode

### DIFF
--- a/crates/warpui/src/windowing/winit/event_loop/mod.rs
+++ b/crates/warpui/src/windowing/winit/event_loop/mod.rs
@@ -1020,7 +1020,12 @@ impl EventLoop {
         })();
 
         match render_result {
-            Ok(_) => self.callbacks.for_window(window).frame_drawn(),
+            Ok(_) => {
+                self.callbacks.for_window(window).frame_drawn();
+                if self.ime_enabled {
+                    self.update_ime_position();
+                }
+            }
             Err(err) => {
                 log::warn!("Failed to render frame: {err:#}");
                 self.callbacks.for_window(window).frame_failed_to_draw();
--- a/crates/warpui/src/windowing/winit/event_loop/mod.rs
+++ b/crates/warpui/src/windowing/winit/event_loop/mod.rs
@@ -1020,12 +1020,7 @@ impl EventLoop {
         })();
 
         match render_result {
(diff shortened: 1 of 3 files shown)


bet4it:
@cla-bot check

vorporeal (commented):
my main question is whether this is the correct place to be doing this.  feels a bit odd that we're unconditionally attempting to update the IME position after every frame we render, as opposed to this being reactive to a change in cursor position?

we do seem to have this reactive logic (the `ActiveCursorPositionUpdated` event) - do you have a sense as to why that event isn't sufficient?

vorporeal:
i guess in some ways it ends up being a question of whether this is a flow that should be reactive (i.e.: focus and cursor position changes trigger a call into platform code to update the IME position) or whether we want to "poll" for an updated position on every frame when the IME is open.  i'm doing some quick testing on macOS to see what the update behavior/frequency is like there, to get a better sense of what might be reasonable to do on linux.

bet4it:
I’ve updated the patch to remove the redraw-side IME update entirely. The `winit` event loop now only reacts to `ActiveCursorPositionUpdated`; the actual cursor-position change detection lives in `AppContext::build_scene()`, where the rebuilt layout has the final active cursor rect. That path compares the newly computed cursor position against the last observed one and only emits the event when it actually changes.

The old event path wasn’t sufficient for terminal-style cursor movement because that movement doesn’t currently surface as a separate signal before layout/rendering. `build_scene()` is the first place where the layout-derived cursor position is reliably available, so that’s where I moved the comparison. On Linux/FreeBSD, the platform IME call is now strictly event-driven again.

vorporeal (approved):
lgtm; thanks for the fix!

bet4it:
Hope this PR can be merged @vorporeal @abhishekp106

vkodithala:
Hey @bet4it, I've enabled CI workflows. We'll get this merged in after they run!

Issue #10353, opened 2026-05-07
Opt-in: bare left-drag selects in TUI without holding Shift on macOS


spalagu:
## User pain point

On macOS, `Cmd+C` is intercepted by the terminal app and never reaches the running TUI as a stdin byte sequence. With Warp's current default, a bare left-button drag inside a full-screen TUI (Claude Code, vim, htop, tmux, etc.) is forwarded to the application via the mouse-reporting protocol. The application owns the selection, Warp itself has no native selection, and pressing `Cmd+C` appears to do nothing — Warp's clipboard handler has nothing to copy.

The existing workaround is `Shift`+drag, which bypasses mouse-reporting and routes the drag to Warp's native selection. But this is non-obvious: macOS users have decades of muscle memory expecting bare-drag to always select text. Discovering the `Shift` modifier requires reading docs or asking — which is why this question keeps coming up across community channels.

iTerm2 has solved this since forever: left-button drag is **always** a native selection gesture, and you hold `Option` to forward it to the TUI when needed. Other mouse events (right-click, scroll wheel, modifier+click) still follow mouse-reporting normally, so TUI features like scrolling, context menus, and click-to-position keep working.

## Proposed solution

Add an opt-in setting `terminal.native_left_drag_select_enabled` in the AltScreenReporting group, default `false` (no behavior change for existing users). When the setting is enabled:

- Left-button events (down, up, drag) are intercepted by Warp's native selection logic — same effect as if `Shift` was held.
- Bare left-drag in a TUI selects text natively in Warp; `Cmd+C` copies the resulting selection.
- All other mouse events (right-click, scroll, middle-click, modifier+click) continue to follow mouse-reporting, so TUI mouse functionality is preserved.

UI surface mirrors the existing `mouse_reporting_enabled` / `scroll_reporting_enabled` / `focus_reporting_enabled` toggles:

- Settings → Features → Terminal → "Native Left-Drag Selection" switch
- Command palette: `native left drag select` toggle action
- Context flag for keybinding integration

## Why opt-in (not default-on)

Some users / TUI workflows rely on bare-drag being forwarded to the application (e.g., custom drag-to-resize panes in tmux). Making this default-on would break their setup. Opt-in is safer and lets each user decide whether to follow Warp's literal-protocol default or iTerm2's selection-first default.

## Working code

The implementation is already done and verified in a fork — happy to open it as a code PR after this issue gets `ready-to-implement` (or `ready-to-spec` if a spec is preferred first):

[...]

PR #10455, opened 2026-05-08
spec: opt-in native left-drag selection in TUIs (#10353)

lonexreb:
## Summary

Spec for #10353. Adds an opt-in setting `terminal.native_left_drag_select_enabled` (default `false`, in the `AltScreenReporting` group) that makes bare left-drag perform Warp's native text selection inside alt-screen TUIs, so `Cmd+C` works without holding `Shift`. Mirrors iTerm2 semantics: when enabled, `Option+left-drag` is the inverse-modifier opt-out that forwards drag to the TUI.

The issue author has a working code branch at https://github.com/spalagu/warp/tree/feat/left-drag-select-default. This PR formalizes the behavior contract before merge.

## What's specified

- Setting key, group, default, and UI placement.
- Behavior contract for setting OFF (current behavior preserved) and ON (intercept bare left-button events for native selection).
- Other inputs (right-click, scroll, middle-click, `Cmd+left-click`) explicitly unchanged in both modes.
- `Option+left-drag` as the documented forward-to-TUI escape hatch when ON.
- Live toggle without app restart; in-flight gestures not interrupted.
- Acceptance criteria (A1–A8), test plan (T1–T8), and implementation pointers referencing the existing working branch.

## Test plan

Spec-only PR. Tests T1–T8 will be implemented alongside the code change.

- [ ] T1: default-off behavior unchanged
- [ ] T2: default-on bare-drag selects natively
- [ ] T3: Cmd+C copies after bare-drag selection
- [ ] T4: Option+left-drag forwards to TUI when ON
- [ ] T5: right-click/scroll/middle-click/Cmd+click unchanged
- [ ] T6: mid-session toggle takes effect immediately
- [ ] T7: setting persists across restart
- [ ] T8: keybinding context flag exposed

/label ready-to-spec

### DIFF
--- a/specs/GH10353/SPEC.md
+++ b/specs/GH10353/SPEC.md
@@ -21,7 +21,7 @@ The issue author has a working code branch at https://github.com/spalagu/warp/tr
 - Add a single opt-in setting that, when enabled, makes bare left-drag a native selection gesture inside alt-screen TUIs.
 - Preserve all other mouse-reporting behavior unchanged (right-click, scroll, middle-click, modifier-click variants, non-alt-screen apps).
 - Match iTerm2's inverse-modifier semantics: when the setting is on, `Option+left-drag` is the explicit "forward to TUI" gesture.
-- Take effect immediately for new mouse events after toggling — no app restart.
+- Take effect immediately for new mouse events after toggling — no app restart, but in-flight gestures remain consistent (see B6 routing latch).
 
 ## Non-Goals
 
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) defined click semantics in ON mode for single, double, triple, and quad+ clicks — all bare-left interactions are intercepted by Warp's native selection; (2) added Routing Latch subsection: routing decision is determined at mouse-down based on current setting + modifier state and LATCHED for the gesture lifetime, releasing only on mouse-up; (3) Implementation Pointers now require a per-gesture routing map keyed on `(button, gesture_id)` at the mouse-event-dispatch layer; (4) added tests T9–T14 for click count semantics, mid-gesture latch invariants, and new-gesture-after-toggle behavior.

lonexreb:
Round-2 fixes: (1) `should_intercept_mouse()` now takes a `MouseEventContext` struct with full modifier state, click_count, gesture_id, and event kind; modifier-keyed left clicks (Cmd+left-click, etc.) bypass the new routing entirely and follow existing handlers; (2) defined gesture_id source — references existing platform-level pointer-interaction IDs OR adds a new `GestureSession` in the event-dispatch layer (counter per mouse-down, timeout-on-200ms-idle); (3) educational toast resolved as YES in V1 — one-time per install via local sidecar; auto-dismisses 8s; T_first_enable_toast added.

lonexreb:
Addressed round-N oz-for-oss review:

**1. Modifier-keyed left-clicks no longer routed to ForwardToTui.** Introduced a tetra-state `MouseRoutingDecision` adding `DeferToLocalHandler`. `Cmd+left-click` (link follow, OS open) and `Ctrl+left-click` (macOS right-click stand-in) now route via `DeferToLocalHandler` so Warp's existing local handlers run — never forwarded to the TUI, never intercepted as native selection. Updated B3 table row and added explicit invariant that future modifier handlers MUST use `DeferToLocalHandler`, not `ForwardToTui`.

**2. Callsite update guidance now requires the FULL MouseEventContext.** The 7 callsites must populate every field — `is_left_button`, `event_kind`, the FULL `ModifierState` (shift/ctrl/alt_or_option/cmd_or_meta — partial passes like only `option_held` are explicitly forbidden because they would silently break Shift+drag, Cmd+click, and future modifier handlers), `click_count`, and the `gesture_id` from `GestureSession`. Also locked in latch behavior: MouseDown computes & inserts; MouseMove/MouseUp read latched decision and never recompute.

Commit: 10ca1909

peicodes:
I'm selecting https://github.com/warpdotdev/warp/pull/10552 as the canonical spec for this issue, and will request modifications there.

Issue #10340, opened 2026-05-07
Code Review: group changed files by folder (VS Code-style tree view)


johnweng1002:
## Summary
Request folder-based grouping for the Change Files list in Warp Code Review, similar to VS Code's Explorer-style tree view.

## Problem
When many files are changed, the current flat file list in Code Review becomes harder to scan and navigate. Grouping by folder would improve readability and make it faster to locate related changes.

## Reproduction steps or desired workflow
1. Open Warp Code Review in a repository with multiple changed files across several directories.
2. Open the changed files sidebar/list.
3. Observe that files are shown in a flat list.
4. Desired: provide a folder-grouped/tree view (expand/collapse by directory) like VS Code, ideally with an option to switch between flat and grouped views.

## Artifacts
None attached

## Warp version
Unknown

## Operating system
macOS

peicodes:
@alokedesai is actively working on this

pmarjanovic:
Hey @peicodes , any updates on this one? Thanks!

PR #10456, opened 2026-05-08
spec: folder tree view for Code Review file list (#10340)

lonexreb:
## Summary

Spec for #10340 — adds a VS Code-style folder-grouped tree view to the Code Review changed-files list, with a header toggle between Flat and Tree modes.

This is a **spec-only** PR. No implementation in this change.

## Highlights

- Header toggle "Flat | Tree", default Flat (preserves current behavior).
- Tree mode collapses single-child folder chains by default ("Compact folders" toggle, matches VS Code Explorer).
- Folder rows show file count and aggregate `+N -M` line counts.
- Selection preserved across mode toggle; matching files auto-expand their parent folders during search.
- Keyboard navigation: arrows, Enter, Cmd-click toggles all descendants.

## Settings

- `code.review.file_list_mode` (`"flat"` | `"tree"`, default `"flat"`)
- `code.review.tree_compact_folders` (bool, default `true`)

## Telemetry

Extends the existing `code_review.opened` event with a `file_list_mode` field.

## Open Question

Should folder expansion state persist across review sessions for the same PR? Suggested: yes, in V1.5.

Tag: `ready-to-spec`

Closes part of #10340 (spec only — implementation will follow once the spec is approved).

### DIFF
--- a/specs/GH10340/SPEC.md
+++ b/specs/GH10340/SPEC.md
@@ -1,0 +1,163 @@
+# Folder Tree View for Code Review File List (GH-10340)
+
+## Summary
+
+Add a VS Code-style folder-grouped tree view to the Code Review changed-files
+list, with a header toggle between Flat and Tree modes. Tree mode supports
+collapsible folders, compact single-child folder chains, folder-level change
+counts and aggregate `+/-` line counts, and preserves selection across mode
+switches.
+
+## Problem
+
+The Code Review changed-files list is currently flat. For PRs that touch many
+files spread across nested directories, this flat layout makes it hard to
+scan changes by area, hard to scope review per-package, and hides the shape of
+the change. VS Code's Explorer-style tree view is the expected mental model
+for a hierarchical changed-files list.
+
+## Goals
+
+- Tree view of changed files grouped by directory.
+- Collapsible folders with expand/collapse affordance.
+- User-visible toggle to switch between Flat and Tree views.
+- Preserve current selection when switching between modes.
+- Folder-level summaries: count of changed files and aggregate `+N -M` line
+  counts where available.
+- Compact-folder behavior matching VS Code (collapse single-child chains).
+
+## Non-Goals
+
+- Not a full file explorer; this is scoped to the Code Review changed-files
+  list only.
+- No new git operations exposed from the tree (stage/discard/etc.).
+- Not synchronizing folder expansion state across sessions in V1; tracked as
+  V1.5 (see Open Questions).
+- No symbol-level / hunk-level grouping; entries remain at file granularity.
+
+## Behavior Contract
+
+### B1. Mode toggle
+
+A header button group "Flat | Tree" switches the file list between modes.
+Default is Flat to preserve current behavior. The user's choice persists per
+user (across restarts).
+
+### B2. Tree rendering and compact folders
+
+In Tree mode the file list renders as a hierarchy of folders rooted at the
+review's repo root. Single-child folder chains collapse into a compact label
+matching VS Code's Explorer (e.g. `app/src/foo/bar/baz.rs` displays under a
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) added Renames & Cross-Directory Placement — renamed files default-group under NEW path with old-path subtitle; opt-in toggle to group by old path; never double-counted; (2) defined File-List Filter Contract — distinct new filter (filename-substring) at file-list header, separate from existing diff-content find bar; survives Flat↔Tree toggle; cleared per-session; (3) defined Selected File precisely — the file whose diff is open in the main pane; preserved across mode toggles; window-local; (4) telemetry extends existing `CodeReview.PaneOpened` with `file_list_mode` field, adds new `CodeReview.FileListModeToggled` and `CodeReview.FileListFiltered` events.

lonexreb:
Round-2 fixes: (1) UI placement reconciled — Flat/Tree toggle, filter input, kebab actions ALL in the file-list panel header (verified path); not in the Code Review header; (2) defined `code.review.file_list.hide_non_matches_when_filtering` (bool, default false) — false = dim non-matches, true = hide; placed in panel header kebab menu; (3) rename-grouping toggle now exposed as `code.review.file_list.group_renames_by_old_path` setting + kebab menu entry, with full acceptance + test coverage.

lonexreb:
Addressed round-N oz-for-oss review:

**Important — Rename-grouping toggle persistence contradiction.** B9 previously said the toggle was 'window-local and not persisted in V1' while Settings/API, A8, and T8 listed it as a persisted setting. Resolved in favor of the persisted-setting contract (consistent with the other kebab-menu toggles like Compact folders and Hide non-matching). Added an explicit 'Authoritative persistence contract' paragraph in B9 stating that `code.review.file_list.group_renames_by_old_path` is the single source of truth and that the prior 'window-local' wording is superseded.

**Suggestion — Telemetry not covered by acceptance/tests.** Added four new acceptance criteria (`A_telemetry_pane_opened`, `A_telemetry_mode_toggle`, `A_telemetry_filter_debounce`, plus the empty-input non-emission rule) and five new tests (`T_telemetry_pane_opened_flat`/`tree`, `T_telemetry_mode_toggle`, `T_telemetry_filter_debounce`, `T_telemetry_filter_no_event_on_empty`). Coverage now includes: per-event payload shape, the once-per-toggle invariant, the debounce-per-filter-session contract for both clear and focus-loss terminators, and the no-emit-on-empty invariant.

Commit: d240aa8e

lonexreb:
Addressed the latest oz-for-oss review.

**Filter match scope (concern 1)** — Filter is now defined as case-insensitive plain-text substring against the **basename only** (final path segment). It does NOT match full paths, intermediate directories, folder labels (compact or otherwise), or the non-attributed side of renames. For renamed files, matching uses the basename of the attribution path (new-path basename by default; old-path basename when `group_renames_by_old_path = true`). Tests T_filter_match_basename_only, T_filter_no_match_on_folder_label, T_filter_match_rename_attribution_path encode this.

**Filter-driven auto-expansion transience (concern 2)** — Spec now defines a two-layer expansion model: a persisted base map (mutated only by user clicks/keyboard) plus a transient filter-expansion overlay (cleared on filter empty/blur). Visible state is the union. Folders the user manually toggles while the filter is active update the base map and persist after clear. Tests T_filter_auto_expansion_transient and T_filter_manual_expansion_persists encode both halves.

**Implementation pointer clarification (suggestion)** — Implementation Pointers now identify `code_review_view.rs` as the existing flat-list renderer location, and describe the carve-out: a new `app/src/code_review/file_list/` module (mod.rs, tree.rs, renderer.rs, header.rs) takes over both Flat and Tree rendering. `code_review_view.rs` is updated to delegate sidebar rendering and contains no file-row code after the change. T_filelist_module_split asserts this.

Issue #10348, opened 2026-05-07
Configurable word delimiters for Delete Word Left shortcut


edwin-686:
## Summary
Add a setting to configure which characters are treated as word boundaries for word-deletion shortcuts such as Delete Word Left.

## Problem
Delete Word Left currently treats filesystem paths and similar tokens as a single word, deleting backward only to whitespace.

## Reproduction steps or desired workflow
1. Type text containing path-like delimiters, for example `/var/www/example.com/logs`.
2. Trigger Delete Word Left.
3. Observe that deletion jumps to the previous space instead of stopping at path delimiters.
4. Desired workflow: let users configure delimiter characters (for example `/`, `.`, `-`, `_`) so word-based deletion shortcuts stop at those boundaries.

## Artifacts
None attached.

## Warp version
Unknown

## Operating system
macOS

edwin-686:
Thanks for the response.

I’m mainly referring to normal Warp command input before pressing Enter, not an already-running interactive program.

Example:

cd /var/www/customer.example.com/public_html

If my cursor is at the end and I press Option + Backspace / Delete Word Left, Warp deletes the whole path back to the previous space:

cd 

What I’m hoping for is behaviour closer to iTerm2/readline-style editing, where delimiters such as /, ., -, and _ can be treated as word boundaries. So repeated Delete Word Left would behave more like:

cd /var/www/customer.example.com/public_
then
cd /var/www/customer.example.com/
then
cd /var/www/customer.
then
cd /var/www/

A few common examples where this matters:

https://portal.example.com/customer/login
192.168.1.254
customer-name-prod-server
/etc/nginx/sites-enabled/example.com.conf
git checkout feature/customer-api-fix

In all of these cases, Warp currently treats the full string as one word because there are no spaces. I’d like a setting where users can define characters considered word delimiters for command editing actions like Delete Word Left, Move Word Left, and ideally related word actions.

Regarding your second question: I have not intentionally changed the smart selection setting. I can check, but I’m not mainly referring to mouse double-click selection. This is specifically about keyboard word deletion/editing behaviour in the command input.

The closest reference would be configurable word delimiter behaviour like iTerm2/terminal/readline-style editing, where /, ., -, _, etc. can act as boundaries instead of only spaces.

PR #10459, opened 2026-05-08
spec: configurable word delimiters for word-deletion shortcuts (#10348)

lonexreb:
## Summary

Spec for #10348. Makes the set of characters that act as word boundaries configurable for **all four** word-cursor shortcuts: Delete Word Left, Delete Word Right, Move Cursor Word Left, Move Cursor Word Right. Today these shortcuts treat path-like strings (e.g. `/var/www/example.com/logs`) as a single word; with a configurable delimiter set including `/`, `.`, `-`, `_`, `:`, `=`, the shortcuts split per segment, matching VS Code, JetBrains, and Sublime conventions.

## What's specified

- New `editor.word_delimiters` setting (string of literal characters) with a sensible default: `"/.-_:= \t\n\r\x0b\x0c"`.
- Per-input-context overrides: `terminal.word_delimiters`, `agent.word_delimiters` (both fall back to `editor.word_delimiters`, then to the built-in default).
- Whitespace is **always** a delimiter regardless of configuration (B5) — prevents broken configs.
- Delimiter run collapse semantics matching modern editors (B7).
- Acceptance criteria (A1–A9), test plan (T1–T10), implementation pointers, and a V1.5 deferral for regex-based boundaries.

This spec affects 4 shortcuts (Delete Word Left/Right, Move Cursor Word Left/Right) so they share one boundary definition rather than diverging.

## Test plan

Spec-only PR. Tests T1–T10 will be implemented alongside the code change.

- [ ] T1: default behavior on path string
- [ ] T2: custom delimiter set
- [ ] T3: context override precedence
- [ ] T4: empty setting fallback
- [ ] T5: whitespace forced as delimiter
- [ ] T6: delimiter run collapse on `foo//bar`
- [ ] T7: word-navigation parity with deletion
- [ ] T8: live effect without restart
- [ ] T9: setting persists across restart
- [ ] T10: UTF-8 multi-byte safety

/label ready-to-spec

### DIFF
--- a/specs/GH10348/SPEC.md
+++ b/specs/GH10348/SPEC.md
@@ -40,13 +40,54 @@ A new setting `editor.word_delimiters` (string of characters) defines the global
 
 The default includes path/punct characters (`/`, `.`, `-`, `_`, `:`, `=`) plus all standard whitespace (space, tab, newline, carriage return, vertical tab, form feed). Whitespace is included for back-compat with current behavior.
 
-### B2. Word-deletion semantics
+### B2. Canonical word-deletion algorithm
 
-For Delete Word Left and Delete Word Right, characters in the delimiter set are word boundaries. Delete Word Left removes characters from the cursor leftward until (and including) the next delimiter run, OR start-of-line, whichever comes first. Delete Word Right is the mirror: rightward until a delimiter run or end-of-line.
+Let `D` be the resolved **delimiter character set** for the current input context (per B4). Define:
+
+- A **non-delimiter run** is a maximal contiguous sequence of characters NOT in `D`.
+- A **delimiter run** is a maximal contiguous sequence of characters IN `D`.
+
+The buffer is therefore an alternating sequence of non-delimiter runs and delimiter runs.
+
+#### Delete Word Left
+
+From the cursor position, walk LEFT and apply the following rules to a single Delete Word Left invocation:
+
+1. If the cursor is at start-of-line, do nothing.
+2. If the character immediately LEFT of the cursor is in `D`: consume the entire delimiter run that ends just left of the cursor, **then** consume the non-delimiter run that ends just left of that (if any). Stop at start-of-line.
+3. Otherwise (the character immediately LEFT of the cursor is NOT in `D`): consume only the non-delimiter run from the cursor leftward to its start. Do **not** consume any preceding delimiter run on this invocation.
+
+In all cases, do not cross start-of-line.
+
+#### Delete Word Right
+
+Mirror of Delete Word Left:
+
+1. If the cursor is at end-of-line, do nothing.
+2. If the character immediately RIGHT of the cursor is NOT in `D`: consume only the non-delimiter run from the cursor rightward to its end. Do **not** consume any following delimiter run on this invocation.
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) **CRITICAL fixed** — wrote ONE canonical Delete Word Left/Right algorithm (B2) with worked example for `/var/www/example.com|` and rewrote all examples + acceptance criteria + tests to match (no more contradictions); (2) added Delimiter Setting Representation subsection (B6) covering TOML escape sequences, Settings UI text-field with Show-Escapes toggle, validation rules; (3) resolved fallback semantics — empty string = missing (default); whitespace-only = invalid (default + warning); whitespace always forced as delimiter regardless; per-context override invalid → falls through to editor (B6.5); (4) Tests now cover Delete Word Right (T8a-d) and Move Cursor Word Right (T7a), plus fallback behaviors (T9a-g).

lonexreb:
Round-2 fixes: (1) **CRITICAL resolved** — every Delete Word Right acceptance criterion now matches the canonical mirror algorithm (cursor-mid-non-delim consumes to end-of-run; otherwise consumes delim+non-delim); two worked examples added; (2) TOML escape guidance corrected — only standard escapes (`\b\t\n\f\r"\` and `\uXXXX`/`\UXXXXXXXX`) — non-standard `\xXX` removed; default value now uses ``/``; (3) empty value behavior reconciled — empty string in TOML or UI -> fall back to default; UI shows default in field with reset indicator; no error on empty; (4) left-delete test sequence now walks all 6 states for `/var/www/example.com|`.

lonexreb:
Addressed all 3 round-N concerns:

**1. TOML escape for VT/FF is now unambiguous and copy-pasteable.** Replaced the placeholder backticks with explicit text: VT (U+000B) is encoded as the 4-hex-digit Unicode escape `\u000B` (TOML has no single-character VT escape, so this is the only valid form); FF (U+000C) is encoded as `\f` (preferred) or `\u000C`. Added a copy-pasteable TOML literal with the full whitespace floor: `editor.word_delimiters = "/.-_:= \t\n\r\u000B\f"` and noted that the simpler `"/.-_:= \t\n\r"` also works because B5's whitespace floor adds VT/FF back at runtime.

**2. A6 no longer flattens whitespace-only fallback to the built-in default.** Split A6 into A6.a (global `editor.word_delimiters` whitespace-only → built-in default per B1) and A6.b (per-context `terminal.` / `agent.` whitespace-only → fall THROUGH to the resolved `editor.word_delimiters`, NOT directly to the B1 default). Aligns A6 with the B6.5 per-context fallthrough rule and with T9g.

**3. Implementation pointers no longer make `editor.word_delimiters` a literal schema default.** Reworded the Settings schema bullet: all three keys are `Option<String>` with schema default `None`. The B1 character set is applied at RESOLUTION TIME by `resolve_word_delimiters(InputContext::Editor)`, not at schema-load time. Documented why a literal schema default would break the B6.5 contract that 'absent' and 'empty string' are indistinguishable, and would also corrupt the Settings UI hint/'Reset to default' indicator state. The B1 character set lives as a `const DEFAULT_WORD_DELIMITERS` next to `WordBoundaryClassifier`, not in the settings schema.

Commit: 6b439d5b

Issue #10320, opened 2026-05-07
Add search within the File Tree


bholmesdev:
## Summary
Add a way to search for files within Warp's File Tree / Project Explorer.

## Problem
The File Tree is useful for browsing project files, but there does not appear to be an in-tree way to quickly find a file by name from the tree itself. Users who are focused on the tree may expect the familiar editor behavior where search is scoped to the file explorer rather than the terminal output or another surface.

## Reproduction steps or desired workflow
1. Open the File Tree / Project Explorer.
2. Focus the tree panel.
3. Press `Cmd+F`, or click a visible search button in the tree header.
4. Type part of a file or folder name.
5. The tree filters, highlights, or otherwise navigates to matching files within the current project tree.

The desired behavior is similar to VS Code: when the explorer/tree has focus, `Cmd+F` should search within that tree, and the tree should also expose a visible search button for discoverability.

## Artifacts
None attached.

## Warp version
Unknown.

## Operating system
macOS

captainsafia:
Thanks for filing this issue, Ben! We think this is a valid request to take through but would definitely need some mocks as far as how an integrated search experience would look in the UI. After that, we'd like to move it to ready-to-spec based on the mocks that are created.

PR #10460, opened 2026-05-08
spec: file tree search (#10320)

lonexreb:
## Summary

Spec for #10320 — adds in-tree search to the File Tree / Project Explorer, openable via `Cmd+F` (`Ctrl+F` on Windows/Linux) when the tree has focus and via a discoverable magnifier button in the tree header.

This is a **spec-only** PR. No implementation in this change.

## Highlights

- `Cmd+F` is scoped to a new `KeyContext::FileTreeFocus` so it does not collide with editor / terminal `Cmd+F`.
- Live filter on every keystroke; auto-expands folders containing matches.
- Substring match against filename and path components, case-insensitive, NFC-normalized.
- Next / previous match navigation via `F3` / `Shift+F3` (and `Enter` / `Shift+Enter`).
- Dim-non-matches by default; `file_tree.search.show_all_on_filter` toggles to hide instead.
- `Esc` and clear restore prior selection and tree expand state.

## Scope

- V1: filename + path-component substring search.
- V1.5 (deferred): regex / glob, and an optional git-status filter row ("show only modified files").

## Telemetry

Extends the existing `file_tree.opened` event with an optional `search_used` boolean per session.

Tag: `ready-to-spec`

Closes part of #10320 (spec only — implementation will follow once the spec is approved).

### DIFF
--- a/specs/GH10320/SPEC.md
+++ b/specs/GH10320/SPEC.md
@@ -57,10 +57,98 @@ behaviors.
 - Case-insensitive substring match.
 - Match is checked against each path component (folder names and the file
   basename) and against the full relative path string.
-- A folder match reveals all of the folder's children as in-context matches
-  (folder-as-context).
-- Matching is locale-insensitive for ASCII; Unicode normalized using NFC
-  before matching.
+- A folder match reveals all of the folder's descendants as in-context
+  matches (folder-as-context). See B2a for the precise descendant
+  semantics.
+- Matching is performed after NFC normalization, then full Unicode default
+  case folding. See B2b for the precise Unicode rules.
+
+### B2a. Folder-match descendant semantics
+
+When a folder name matches the query, the folder is itself a match and its
+**descendants are included** in the match set as "in-context" matches. The
+following rules govern visibility, counting, highlighting, and navigation
+of folder-match descendants. They apply consistently in both visibility
+modes (B3 dim mode and `dim_non_matches = false` hide mode).
+
+#### B2a.1. Visibility
+
+- **Dim mode** (`file_tree.search.dim_non_matches = true`, default): the
+  matched folder and ALL its descendants render at FULL emphasis (NOT
+  dimmed) — they are part of the match set. Other (non-matching, non-
+  descendant) tree nodes follow the normal dim rules in B3.
+- **Hide mode** (`file_tree.search.dim_non_matches = false`): the matched
+  folder and ALL its descendants render normally (visible). They are NOT
+  hidden, regardless of whether each individual descendant's name matches
+  the query.
+
+#### B2a.2. Match count (no double-count)
+
+A descendant counted via folder-match is NOT additionally counted if its
+own name also matches the query. Counting is the **union** of the direct-
+match set and the folder-match-descendant set, not their sum:
+
+- A direct file match counts once.
+- A direct folder match counts once for the folder itself; its descendants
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) renamed setting to `file_tree.search.dim_non_matches` (bool, default `true`) — all references in B3, B6, Settings/API surface, Acceptance (A5), and Tests (T6) now use this single name with dimmed-by-default semantics (no more inverted variants); (2) Implementation Pointers audited against the codebase at commit `86940541` — corrected file tree path to `app/src/code/file_tree/` (not `app/src/file_tree/`), removed references to non-existent `header.rs`/`tree_view.rs`/`editor_page.rs`, pointed magnifier-button work at `view.rs` (where the header row is inlined), settings to `code_page.rs`, and keymap context registration to the existing `context.set.insert("FileTreeFocused")` pattern (mirrors `"EditorFocused"` at `app/src/terminal/view.rs:26640`); (3) telemetry uses the real existing event `TelemetryEvent::FileTreeToggled` (label "File Tree Toggled", `events.rs:5658`) — extended with `search_used: bool`, with a fallback option to add a net-new `FileTreeSearchUsed` event if extending is awkward.

lonexreb:
Round-2 fixes: (1) folder-match descendant semantics now explicit — descendants visible (dim mode) or visible (hidden mode), counted via union (no double-count), only the folder name itself highlighted, navigation cycles depth-first/lex through file matches and folder-match descendants; acceptance + tests added; (2) Unicode case folding now explicitly Unicode Default Case Folding per TR#21 (full folding, status C+F), applied AFTER NFC normalization; locale-independent (no Turkish-specific I/i); test coverage for `ß`/`SS`, Greek final-sigma, and Turkish default-folding.

lonexreb:
Addressed round-N review:

**Critical — Unicode case-folding contradiction.** The previous wording said the spec required full Unicode Default Case Folding (TR#21 C+F) but ALSO that `str::to_lowercase()` was an acceptable equivalent. Those two statements are not consistent: `to_lowercase()` applies `SpecialCasing.txt` lowercasing (which preserves `ß` as `ß` and does not collapse final-form sigma `ς` and middle sigma `σ` to a single form), while case folding applies `CaseFolding.txt` C+F (which folds `ß` → `ss` and collapses sigma forms).

Resolved by deleting the `to_lowercase()` permission entirely. The spec now requires a Unicode case-folding crate that implements C+F default folding (e.g. `caseless::default_caseless_match_str` / `caseless::Caseless::default_case_fold`) on the NFC-normalized strings, and explicitly forbids substituting `str::to_lowercase()` even as a fallback. Added a test gate: implementations MUST pass T_unicode_german_ss and T_unicode_greek_sigma — passing those is the authoritative check that the right folding is being applied.

**Important — Hide-mode descendant visibility inconsistency.** B2a.1 says all descendants of a matched folder are visible; B3's hide-mode rule said 'only matching files plus their ancestor folders are rendered'. Reconciled in B3 by stating that the rendered set in hide mode is (1) every node in the **union match set** (direct matches + folder-match descendants per B2a) plus (2) ancestor folders of those nodes. Folder-match descendants qualify as 'matching' for the hide rule because the match set is the union per B2a, so a matched folder DOES keep all descendants visible in hide mode. Wording in B3 explicitly supersedes the earlier 'filename matches plus ancestors' phrasing.

Commit: 713003d3

Issue #10292, opened 2026-05-06
Show start timestamp on collapsed Agent thinking/reasoning phases


MScottAdams:
## Summary
Add a start timestamp to each collapsible "thinking" / reasoning phase in an Agent conversation, visible on the collapsed header so users can see when each phase began without expanding it.

## Problem
During an Agent conversation, reasoning/thinking phases compress to a single expandable line. The collapsed header shows no time information. For longer or multi-turn conversations — especially ones the user steps away from, leaves running with `Run until completion`, or wants to correlate against external events (CI, Slack, logs, cloud agent runs) — there is no way to tell at a glance:

- when the agent started thinking on a given turn
- how long ago a particular phase began relative to neighboring blocks
- which collapsed phase corresponds to a specific point in time the user remembers

Today the only way to recover any per-phase timing context is to expand the phase or cross-reference Command History / the cloud agent management view, neither of which surfaces per-thinking-phase start time inline in the conversation.

## Reproduction steps or desired workflow
Desired workflow:

1. Start any Agent conversation.
2. Send a prompt that triggers a thinking/reasoning phase.
3. The thinking phase collapses to a single expandable line.
4. The collapsed header displays the start time of that phase, for example `▸ Thinking… (started 11:42:07)` or a relative form like `▸ Thinking… (2m ago)`.
5. When expanded, the start time (and ideally the duration) is also shown at the top of the expanded view.

Nice-to-haves:

- A Settings toggle (e.g. under **Settings** > **Agents** or **Appearance**) to turn timestamps on/off and choose absolute vs. relative format.
- Match Warp's existing locale / 12h-24h conventions.
- Apply the same treatment to other collapsible agent sub-blocks (tool calls, plan steps) for consistency.

Out of scope:

- Per-token timing.
- Editing or reordering thinking phases.

## Artifacts
None attached

## Warp version
v0.2026.04.29.08.57.preview_02

## Operating system
Windows 10.0.26200

PR #10461, opened 2026-05-08
spec: start timestamp on collapsed agent reasoning phases (#10292)

lonexreb:
Spec-only PR. References #10292.

## Summary
Adds a start timestamp to collapsed agent reasoning/thinking phase headers (and the same treatment to other agent sub-blocks: tool calls, plan steps), with start time + duration when expanded. Format is configurable (`off` / `absolute` / `relative` / `auto`, default `auto` — relative ≤60min, absolute beyond), locale + 12h/24h aware, and accessible via `aria-label`.

## Why
Long agent runs accumulate many collapsed sub-block rows with no time information. Users cannot correlate phases with terminal actions, judge how long an in-progress phase has been running, or audit a transcript after the fact without expanding each row.

## Spec
See `specs/GH10292/SPEC.md` for the full behavior contract, settings surface, acceptance criteria, implementation pointers (`reasoning_block.rs`, `tool_call_block.rs`, `plan_step.rs`, plus a coalesced ticker), tests, and open questions.

## Related
Builds on per-message timestamp work in PR #10128. That PR delivers message-level timing; this spec extends timestamp coverage *into* sub-blocks within a single agent message. The two share the same coalesced timer infrastructure (single ticker per conversation list, fan-out to subscribed rows — no per-row timers).

## Labels
`ready-to-spec`

### DIFF
--- a/specs/GH10292/SPEC.md
+++ b/specs/GH10292/SPEC.md
@@ -1,0 +1,210 @@
+# Start Timestamp on Collapsed Agent Reasoning Phases (GH-10292)
+
+## Summary
+
+During Agent conversations, reasoning/thinking phases (and other agent sub-blocks
+such as tool calls and plan steps) collapse into a single expandable header line
+that currently carries no time information. This spec adds a start timestamp to
+each collapsed sub-block header (e.g. `▸ Thinking… (2m ago)` or
+`▸ Thinking… (11:42:07)`), and shows start time + duration when the block is
+expanded. Format is configurable per user preference and locale-aware.
+
+## Problem
+
+- Long agent runs accumulate many collapsed reasoning / tool-call / plan-step
+  rows. Users cannot tell when each phase started without expanding it, which
+  makes it hard to:
+  - Correlate a phase with a terminal action they remember triggering.
+  - Judge whether a still-running phase has been pending too long.
+  - Audit a transcript after the fact (e.g. "what was the model doing at 11:42?").
+- The conversation already records start events for these sub-blocks server-side;
+  they just are not surfaced in the collapsed UI.
+- Per-message timestamps (PR #10128) cover the message-level granularity but do
+  not extend into the sub-blocks within a single agent message.
+
+## Goals
+
+- Collapsed reasoning / tool-call / plan-step rows show a start timestamp inline
+  in the header.
+- Expanded headers show both an absolute start time and a duration (or
+  `running…` for in-progress phases).
+- Format respects the user's locale and existing 12h/24h convention.
+- A user-facing setting selects between absolute, relative, automatic, or off.
+- Coverage extends consistently across all collapsible "agent sub-blocks"
+  (reasoning, tool calls, plan steps).
+
+## Non-Goals
+
+- Not per-token / per-chunk timing within a phase.
+- Not editing or reordering phases in the conversation.
+- Not new server-side analytics on per-phase durations beyond what already
+  exists.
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) absolute-time formatting unified — `HH:MM:SS` (24h) and `h:MM:SS AM/PM` (12h), always with seconds; older than 24h prefixed with locale date; all examples rewritten consistently; (2) forced `relative` mode now defines `Nm ago / Nh ago / Nd ago / Nw ago` for starts older than 60 min; auto mode unchanged; (3) Implementation Pointers verified against the worktree — fictional paths replaced or marked as net-new; (4) renamed setting to `agent.subblock_timestamp_format` and UI label to 'Agent sub-block timestamps' to reflect broader scope (reasoning + tool calls + plan steps); all references updated consistently.

lonexreb:
Round-2 fixes: (1) in-progress expanded view now uses a live elapsed counter driven by the existing 1s ticker (e.g., `Started 11:42:07 · 4s elapsed`), then becomes final duration on completion — no 'running…' placeholder; (2) `subblock_timestamp_show_in_expanded` precedence defined: `format` gates everything; `show_in_expanded` controls expanded-only visibility; full 3×2 matrix in Acceptance + tests; (3) `format_absolute(start, now, prefer_24h)` helper signature now takes injected clock; date-prefix triggers when `now-start > 24h`; testable via clock injection; (4) ONE shared helper across all sub-block types; date format uses existing locale + `time.use_24h` setting; ISO-8601 fallback if no locale infra.

lonexreb:
Addressed all 4 round-N concerns:

**1. Plan-step row definition.** Added B1.1 as the single source of truth: a plan-step row is a single rendered row in `plan_and_todo_list.rs` keyed by `(message_id, plan_step_index)`. Start event is `PlanStepStarted` from `orchestration_events.rs` (the same emitter as reasoning/tool-call events); `started_at` is authoritative even if `PlanStepCreated` came first. Completion event is `PlanStepCompleted` (with outcome `completed`/`failed`/`cancelled` — all three freeze the elapsed counter). Documented edge cases: `skipped` rows have no affordance; re-emitted `PlanStepStarted` uses latest `started_at`; pause/resume uses cumulative active time but the `Started` label stays at first `started_at`.

**2. Absolute date locale-vs-ISO conflict.** Resolved in B2.1 in favor of ISO-8601: V1 always uses `YYYY-MM-DD` for the date prefix (locale-independent) so transcripts/screenshots are unambiguous. Locale governs ONLY the AM/PM marker. Updated the implementation pointer's `format_date_prefix` accordingly — drops the `locale` parameter, always returns ISO-8601, V1.5 may revisit.

**3. Ticker cadence: 5 s vs 5 Hz.** Resolved in favor of periods (1 s and 5 s). B4 now states cadences as periods explicitly, with a note that '5 Hz' was wrong and supersedes the earlier wording. Updated the implementation pointer (1 s + 5 s ticker pair, frequencies 1 Hz and 0.2 Hz) and T11 (subscriber count is for 1-second and 5-second tickers, not 5 Hz).

**4. Settings matrix is 4 × 2, not 3 × 2.** Updated B6.1 header text to '4 × 2 matrix' and noted that the format-`off` row collapses both `show_in_expanded` cases to identical hidden output but the matrix shape is still 4 × 2. T_show_in_expanded_matrix updated to '4 × 2 matrix … 8 cells total'.

Commit: 8c2c24c7

PR #10466, opened 2026-05-08
[REMOTE-1544] Apply slash command enablement checks on execution

bnavetta:
## Description

This fixes a bug where executing `/agent <some query>` in the Cloud Mode composer would result in a broken session.

The root issue was that we check slash command availability when constructing the menu, but _not_ when actually executing a slash command. To fix this, we adjust `SlashCommandDataSource` to support checking if a specific command is available, and use this to gate execution in the `Input` view. This protects against a user typing or pasting `/invalid-command` and bypassing the menu.

For this fix to be effective, we also need to check against the proper set of slash commands - namely, the v2 Cloud Mode composer uses a separate data source, and we have to check against _that_ data source rather than the generic one. The `Input` view now receives handles to both `SlashCommandDataSource`s, and chooses the appropriate one based on its current state.

## Linked Issue

https://linear.app/warpdotdev/issue/REMOTE-1544/sending-a-slash-command-skill-in-the-old-cloud-mode-ui-and-current-ui

## Testing

- [x] Added unit tests for slash command enablement

### Screenshots / Videos

https://www.loom.com/share/db71b808547149f8a2f0b2921d75c87d

## Agent Mode

### DIFF
--- a/app/src/search/slash_command_menu/static_commands/commands.rs
+++ b/app/src/search/slash_command_menu/static_commands/commands.rs
@@ -177,7 +177,8 @@ pub static MOVE_TO_CLOUD: LazyLock<StaticCommand> = LazyLock::new(|| StaticComma
     icon_path: "bundled/svg/upload-cloud-01.svg",
     availability: Availability::AGENT_VIEW
         | Availability::ACTIVE_CONVERSATION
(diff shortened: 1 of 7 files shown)


abhishekp106 (approved):
perfect. no notes.

PR #10428, opened 2026-05-08
Add proto messages for remote `DiffStateModel`

MaggieShan:
## Description
* Adds tech spec for how to support `DiffStateModel` and git states over the remote server / on a remote environment 
* Adds relevant proto messages 

## Linked Issue
[APP-4354](https://linear.app/warpdotdev/issue/APP-4354/proto-messages-for-the-diffstatemodel)

## Testing
No-op

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

Relevant plan: https://staging.warp.dev/drive/notebook/Remote-DiffStateModel-Proto-Messages-Server-side-Global-Model-and-Client-Wrapper-wqMeKuOHZxWPnTNZiQqbAf

PR #10470, opened 2026-05-08
Hook `DiffStateModel` to remote server

MaggieShan:
## Description
* Built on https://github.com/warpdotdev/warp/pull/10428 
* Adds the server-side changes to support `DiffStateModel` and git states on remote environments  
  * Adds type conversions
  * Adds a `GlobalDiffStateModel` to track each `LocalDiffStateModel` per `repo_path`, `diff_mode`
  * Adds subscription to repositories to send server messages
  * Adds handlers for client messages

## Linked Issue
[APP-4355](https://linear.app/warpdotdev/issue/APP-4355/hook-up-diffstatemodel-to-remote-server)

## Testing
No-op 

## Agent Mode

### DIFF
--- a/app/src/remote_server/diff_state_proto.rs
+++ b/app/src/remote_server/diff_state_proto.rs
@@ -19,6 +19,21 @@ use crate::util::git::{Commit, PrInfo};
 
 // ── Proto → Rust (for incoming client messages) ────────────────────
 
+/// Rejects empty, absolute, and parent-traversal (`..`) paths.
+fn validate_relative_path(path: &str) -> Result<(), String> {
+    let p = Path::new(path);
+    if p.as_os_str().is_empty()
+        || p.is_absolute()
+        || p.components().any(|c| c == std::path::Component::ParentDir)
+    {
+        return Err(format!(
+            "rejecting path with traversal or absolute component: {}",
+            p.display()
+        ));
+    }
+    Ok(())
+}
+
 impl From<&proto::DiffMode> for DiffMode {
     fn from(proto_mode: &proto::DiffMode) -> Self {
         match &proto_mode.mode {
@@ -55,27 +70,26 @@ impl TryFrom<&proto::FileStatusInfo> for FileStatusInfo {
     type Error = String;
 
     fn try_from(proto_info: &proto::FileStatusInfo) -> Result<Self, Self::Error> {
-        let path = PathBuf::from(&proto_info.path);
-        // Reject empty, absolute paths and path traversal (".." components)
-        // to prevent destructive operations outside the repo root.
-        if path.as_os_str().is_empty()
-            || path.is_absolute()
-            || path
-                .components()
-                .any(|c| c == std::path::Component::ParentDir)
-        {
-            return Err(format!(
-                "rejecting path with traversal or absolute component: {}",
-                path.display()
-            ));
+        validate_relative_path(&proto_info.path)?;
+
+        let status: GitFileStatus = proto_info
+            .status
+            .as_ref()
+            .map(GitFileStatus::from)
+            .unwrap_or(GitFileStatus::Modified);
+
+        // Validate old_path in Renamed/Copied variants — these also flow
+        // into git restore/checkout commands during discard.
+        match &status {
+            GitFileStatus::Renamed { old_path } | GitFileStatus::Copied { old_path } => {
+                validate_relative_path(old_path)?;
+            }
+            _ => {}
         }
+
         Ok(FileStatusInfo {
-            path,
-            status: proto_info
-                .status
-                .as_ref()
-                .map(GitFileStatus::from)
-                .unwrap_or(GitFileStatus::Modified),
+            path: PathBuf::from(&proto_info.path),
+            status,
         })
     }
 }
--- a/app/src/remote_server/diff_state_proto.rs
+++ b/app/src/remote_server/diff_state_proto.rs
@@ -1,0 +1,342 @@
+//! Conversion between diff state Rust types and proto-generated types.
+//!
+//! Rust types are canonical, proto types are the wire format.
+//! Only the directions needed by the server are implemented here.
+//!
+//! This module lives in `app/` (rather than in the `remote_server` crate alongside
+//! `repo_metadata_proto`) because it depends on app-level types
+//! (`code_review::diff_state`, `util::git`) that are not available in the crate.
+use std::path::{Path, PathBuf};
+
+use super::proto;
+
+use crate::code_review::diff_size_limits::DiffSize;
+use crate::code_review::diff_state::{
+    DiffHunk, DiffLine, DiffLineType, DiffMetadata, DiffMetadataAgainstBase, DiffMode, DiffState,
+    DiffStats, FileDiff, FileDiffAndContent, FileStatusInfo, GitDiffWithBaseContent, GitFileStatus,
+};
+use crate::util::git::{Commit, PrInfo};
+
+// ── Proto → Rust (for incoming client messages) ────────────────────
+
+impl From<&proto::DiffMode> for DiffMode {
+    fn from(proto_mode: &proto::DiffMode) -> Self {
+        match &proto_mode.mode {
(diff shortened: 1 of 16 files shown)


app/src/remote_server/server_model.rs
```diff
@@ -1585,6 +1603,421 @@ impl ServerModel {
...
+    /// connection, and responds with the current state (sync) or defers the
+    /// response until the model finishes loading (async).
+    fn handle_get_diff_state(
+        &mut self,
+        msg: super::proto::GetDiffState,
+        request_id: &RequestId,
+        conn_id: ConnectionId,
+        ctx: &mut ModelContext<Self>,
+    ) -> HandlerOutcome {
+        let Some(mode_proto) = &msg.mode else {
```
kevinyang372, line 1697: Mode should never be empty no? Should we enforce that in the proto schema?

app/src/remote_server/server_model.rs
```diff
@@ -1585,6 +1603,421 @@ impl ServerModel {
...
+        );
+
+        // Subscribe this connection.
+        self.diff_states.subscribe_connection(key.clone(), conn_id);
+
+        // Look up or create the model.
+        if let Some(model) = self.diff_states.get_model(&key) {
+            let model_ref = model.as_ref(ctx);
+            let state = model_ref.get();
+            match state {
```
kevinyang372: This logic should belong to the GlobalDiffStatesModel

app/src/remote_server/server_model.rs
```diff
@@ -1543,6 +1563,417 @@ impl ServerModel {
...
+                    })),
+                },
+            ));
+        }
+
+        model.update(ctx, |m, ctx| {
+            m.discard_files(file_infos, msg.should_stash, msg.branch_name, ctx);
+        });
+
+        HandlerOutcome::Sync(server_message::Message::DiscardFilesResponse(
```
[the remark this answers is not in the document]
MaggieShan: The current request and response set up is closer to an ack than an actual success - this follows the local model which currently also doesn't handle success or errors. Alternatively, can update this to a fire and forget client message - don't have a strong opinion for this

app/src/remote_server/diff_state_proto.rs
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+impl From<&proto::GitFileStatus> for GitFileStatus {
+    fn from(proto_status: &proto::GitFileStatus) -> Self {
+        match &proto_status.status {
+            Some(proto::git_file_status::Status::NewFile(_)) => GitFileStatus::New,
+            Some(proto::git_file_status::Status::Modified(_)) => GitFileStatus::Modified,
+            Some(proto::git_file_status::Status::Deleted(_)) => GitFileStatus::Deleted,
+            Some(proto::git_file_status::Status::Renamed(r)) => GitFileStatus::Renamed {
```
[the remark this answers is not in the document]
MaggieShan: Updated

app/src/remote_server/server_model.rs
```diff
@@ -1543,6 +1563,417 @@ impl ServerModel {
...
+                DiscardFilesResponse {
+                    result: Some(discard_files_response::Result::Error(DiscardFilesError {
+                        message: "No valid files after path validation".to_string(),
+                    })),
+                },
+            ));
+        }
+
+        model.update(ctx, |m, ctx| {
+            m.discard_files(file_infos, msg.should_stash, msg.branch_name, ctx);
```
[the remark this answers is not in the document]
MaggieShan: I don't think this is a real concern

app/src/remote_server/diff_state_proto.rs
```diff
@@ -0,0 +1,366 @@
...
+};
+use crate::util::git::{Commit, PrInfo};
+
+// ── Proto → Rust (for incoming client messages) ────────────────────
+
+/// Rejects empty, absolute, and parent-traversal (`..`) paths.
+///
+/// Uses `typed_path` for platform-agnostic detection so that Unix-style
+/// absolute paths (e.g. `/etc/passwd`) are rejected even on Windows.
+fn validate_relative_path(path: &str) -> Result<(), String> {
```
kevinyang372: Why do we need to validate relative path here? I think it's clearer for us to enforce path communicated via protocol has to be absolute

app/src/remote_server/diff_state_proto.rs
```diff
@@ -0,0 +1,366 @@
...
+    type Error = String;
+
+    fn try_from(proto_info: &proto::FileStatusInfo) -> Result<Self, Self::Error> {
+        validate_relative_path(&proto_info.path)?;
+
+        let status: GitFileStatus = proto_info
+            .status
+            .as_ref()
+            .map(GitFileStatus::from)
+            .unwrap_or(GitFileStatus::Modified);
```
kevinyang372: This should be an error instead of fallback?

app/src/remote_server/diff_state_tracker.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// ── GlobalDiffStateModel ────────────────────────────────────────────
+
+/// Manages the lifecycle of server-side `LocalDiffStateModel` instances and
+/// per-connection subscription tracking.
+///
+/// A model is created when the first `GetDiffState` arrives for a given key
+/// and dropped when the last connection unsubscribes (or disconnects).
+pub(super) struct GlobalDiffStateModel {
```
kevinyang372: Can we make this naming clearer it is only intended for remote servers?

app/src/remote_server/diff_state_tracker.rs
```diff
@@ -0,0 +1,175 @@
...
+        fallback_state: DiffState,
+        app: &AppContext,
+    ) -> (DiffState, Option<DiffMetadata>) {
+        self.states
+            .get(key)
+            .map(|model| {
+                let m = model.as_ref(app);
+                (m.get(), m.metadata().cloned())
+            })
+            .unwrap_or((fallback_state, None))
```
kevinyang372: Why do we return fallback here? We should return an error if the state doesn't exist?

app/src/remote_server/diff_state_tracker.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    /// Removes all subscriptions for a disconnected connection.
+    /// Orphaned models (no remaining subscribers) are dropped inline.
+    pub fn remove_connection(&mut self, conn_id: ConnectionId) {
+        let keys = self
+            .key_to_connections
+            .iter()
+            .filter(|(_, conns)| conns.contains(&conn_id))
+            .map(|(key, _)| key.clone())
+            .collect_vec();
```
kevinyang372, line 190: Do we need to collect here?

app/src/remote_server/server_model.rs
```diff
@@ -1543,6 +1563,417 @@ impl ServerModel {
...
+            let model = ctx.add_model(|ctx| {
+                let mut m = LocalDiffStateModel::new(Some(repo_path_str), ctx);
+                m.set_diff_mode(key.mode.clone(), false, ctx);
+                m.set_code_review_metadata_refresh_enabled(true, ctx);
+                m
+            });
+
+            // Subscribe to events from this model.
+            let key_for_sub = key.clone();
+            ctx.subscribe_to_model(&model, move |me, event, ctx| {
```
kevinyang372: Instead of subscribing to every single model, should this be the responsibility of the global diff state model?

app/src/remote_server/server_model.rs
```diff
@@ -1543,6 +1563,417 @@ impl ServerModel {
...
+                    server_message::Message::DiffStateFileDelta(delta),
+                );
+            }
+        }
+    }
+
+    /// Spawns an async diff reload with `content_at_base` for late-joining
+    /// subscribers. Used when the model is already `Loaded` but the expensive
+    /// content data was shed after the initial `NewDiffsComputed` event.
+    fn spawn_content_reload_for_subscriber(
```
kevinyang372: This also feels like this should be in the global diff state model

kevinyang372 (approved):

PR #10478, opened 2026-05-08
fix(terminal): render startup inline images

### LINKED ISSUE #10020: Inline image protocols (Kitty graphics, iTerm2) are silently dropped during shell initialization, but render correctly post-init
### Describe the bug

Inline image-protocol escape sequences (Kitty graphics protocol and iTerm2 inline image protocol) emitted by a command running during shell initialization (e.g. from `~/.zshrc`) are silently discarded by Warp. The exact same command, typed manually in the same tab after the shell prompt is shown, renders the image correctly.

This makes it impossible to use tools like `fastfetch`, `neofetch`, `viu`, `imgcat`, `chafa` (in image-protocol mode), etc. as part of a shell welcome banner — a common use case in other terminal emulators that support these protocols.

### To reproduce

1. Install fastfetch: `brew install fastfetch`
2. Save any PNG to `~/test.png`.
3. Add this single line to `~/.zshrc`:
   ```sh
   fastfetch --logo-type kitty-direct --logo "$HOME/test.png" --logo-width 22 --logo-height 22 --pipe false
   ```
4. Open a **new** Warp tab.
   → Fastfetch's system info renders. **No image is shown** — fastfetch's built-in ASCII Apple logo appears in its place (this is fastfetch's silent fallback when an image protocol is unsupported).
5. In the same tab (after the prompt is visible), manually type:
   ```sh
   fastfetch --logo-type kitty-direct --logo ~/test.png --logo-width 22 --logo-height 22 --pipe false
   ```
   → Fastfetch's system info **and the PNG image** render correctly via the Kitty graphics protocol.

The same contrast occurs with `--logo-type iterm` (iTerm2 inline image protocol) and `--logo-type kitty`: works post-init, fails during init.

### Expected behavior

The image renders identically in both cases — during shell init **and** when typed manually. Other terminals supporting these protocols (iTerm2, WezTerm, Kitty, Ghostty) render the image at shell startup as documented in their respective specifications. The protocols themselves do not require the parser to be in any "post-prompt" or "interactive" state.

Capturing fastfetch's stdout via `> /tmp/init-output.bin` directly inside `~/.zshrc` reveals that during real shell initialization, fastfetch (with `--logo-type kitty-direct ... --pipe false`) emits a well-formed Kitty graphics escape sequence:

```
\e_Ga=T,f=100,t=f,c=22,r=22;<base64 of absolute image path>\e\
```

per the [Kitty graphics protocol spec](https://sw.kovidgoyal.net/kitty/graphics-protocol/). The same invocation typed manually after the prompt renders the image correctly. Since fastfetch's stdout is the PTY connected to Warp, the byte stream reaching Warp is by construction identical in both contexts. The rendering difference therefore originates in Warp's processing of the same input depending on session state.

### Operating system (OS)

macOS

macOS Tahoe 26.2 (25C56) — arm64

### Shell Version

zsh 5.9

### Current Warp version

v0.2026.04.29.08.57.01

### Regression

No — this has been the behavior for as long as I've been using Warp (months). I cannot identify a working version.

### Additional context

Approaches tried that all fail to render the image at shell startup:

- Switching protocol: `iterm`, `kitty`, `kitty-direct` — all silently fall back to the ASCII Apple logo at startup, all render correctly post-init.
- Reordering: placing `fastfetch` **before** any prompt-manager initialization in `.zshrc` (a known fix for Powerlevel10k instant-prompt collisions in Kitty terminal — see [fastfetch issue #485](https://github.com/fastfetch-cli/fastfetch/issues/485)).
- `fastfetch --pipe false` — forces emission regardless of TTY auto-detection. The escape sequences are emitted (verified via stdout capture, see logs section) but still not rendered by Warp.
- Deferring via `precmd` zsh hook (post-init): output is suppressed by Warp's block UI.
- Deferring via background subshell: `( sleep 0.2 && fastfetch ... ) &!`. Even when output arrives well after the first prompt, Warp does not render the image.
- Various `sleep` durations before fastfetch.

The only fallback that works at shell startup is pre-rendering the image with `chafa --format symbols` to a text file (Unicode block characters with truecolor ANSI escapes), which is plain text + color escapes — i.e. content that does **not** rely on the inline image-protocol parser.

**Hypothesis**: Warp's image-protocol parser is gated on a session state that is only entered after the first interactive command-output block is created. Output emitted during the shell-init phase is processed for normal text + ANSI color escapes, but the inline image protocols are silently dropped. The `( sleep N && fastfetch ) &!` test (output arrives after first prompt yet still does not render) suggests the gating is stronger than simply "post-prompt".

This affects every common shell-startup banner tool (neofetch, fastfetch, macchina, viu-based banners, custom OSC scripts).

No

Yes — the same setup renders correctly at shell startup in iTerm2, WezTerm, and Kitty per their documented protocol behavior.

---

If this gets triaged and labeled `ready-to-implement`, I'd be happy to have an Oz cloud agent take the implementation. I can verify the fix in my setup.

KimHyeongRae0:
## Description

Fixes inline image protocol routing during shell startup and early output.

Completed iTerm2 and Kitty image actions previously bypassed the early-output path, and block-level routing could send images received before preexec into the prompt/header grid. That made startup or background image output parse successfully but not render like normal command output.

This change:

- Routes image completions received before preexec to the output grid, while preserving the hidden `WarpInput` bootstrap behavior.
- Starts the `ScriptExecution` block only after its first image completion creates visible output.
- Sends image completions through `EarlyOutputHandler` so background image output creates a renderable background block.
- Starts early-output background blocks only after image handlers create visible output.
- Avoids creating empty startup/background blocks for invalid iTerm images, Kitty store-only actions, and zero-sized Kitty placements.
- Treats image placements as visible grid content so image-only startup output is not hidden as an empty bootstrap block.
- Keeps non-moving Kitty image placements visible even when they leave the cursor at the origin.
- Keeps normal parser-path active block height refresh centralized in `on_finish_byte_processing()` after image completions.
- Adds unit coverage for the startup, hidden-bootstrap, early-output, and non-renderable image routing paths.

## Linked Issue

Closes #10020

## Testing

- [x] `cargo fmt -- --check`
- [x] `git diff --check`
- [x] `cargo test -p warp test_non_moving_kitty_image_keeps_finished_grid_visible`
- [x] `cargo test -p warp test_image_completion`
- [x] `cargo test -p warp test_iterm_image`
- [x] `cargo test -p warp test_kitty_image`
- [x] `cargo test -p warp --lib image`
- [x] `cargo test -p warp --lib kitty_store_only`
- [x] `cargo test -p warp --lib zero_sized_kitty`
- [x] Manual GUI validation with the `target/debug/warp-oss` binary built by `./script/run --dont-open`

Post-review follow-up validation:

- [x] `cargo fmt -- --check`
- [x] `git diff --check`
- [ ] `cargo test -p warp script_execution_block`
  - Attempted locally after the post-review cleanup, but compilation stopped with `No space left on device (os error 28)` before test execution.

Manual GUI validation passed on macOS with a temporary zsh startup script running `fastfetch --logo-type kitty-direct`. Warp rendered the startup image output as a normal output block, including the `WARP_INLINE_IMAGE_STARTUP_TEST_BEGIN` / `WARP_INLINE_IMAGE_STARTUP_TEST_END` markers and the emitted image.

### Screenshots / Videos

No screenshot or video is attached because this update is being made from the GitHub CLI/API environment, which does not provide GitHub's browser-based PR attachment upload flow. The manual GUI validation above is the end-to-end visual check: a local `warp-oss` build launched with the temporary startup script, and the terminal rendered the fastfetch Kitty-direct image between the startup markers instead of dropping it.

## Agent Mode

- [ ] Warp Agent Mode - This PR was created via Warp AI Agent Mode

CHANGELOG-BUG-FIX: Render inline image protocols emitted during shell startup instead of dropping them.


KimHyeongRae0:
@cla-bot check

KimHyeongRae0:
@cla-bot check

app/src/terminal/model/blockgrid.rs
```diff
@@ -249,8 +249,10 @@ impl BlockGrid {
     /// grid-finish time (precmd or preexec), leaving the grid contentless. If the grid_cursor
     /// is not at (0,0), we can assume the grid contains content that will persist through finish.
     pub fn should_show_as_empty_when_finished(&self) -> bool {
-        self.finished_len() == 0
-            || !self.has_visible_chars()
+        let has_visible_images = self.grid_handler().has_visible_images();
+
+        (self.finished_len() == 0 && !has_visible_images)
+            || !self.has_visible_content()
```
vorporeal, line 254: `self.has_visible_content()` already checks `self.grid_handler.has_visible_images()`, right?  do we need to check the latter again here?

app/src/terminal/model/blocks.rs
```diff
@@ -3815,15 +3815,30 @@ impl ansi::Handler for BlockList {
...
 
     fn handle_completed_iterm_image(&mut self, image: ITermImage) {
-        delegate_to_block!(self.handle_completed_iterm_image(image))
+        let active_block_was_started = self.active_block().started();
+        delegate!(self.handle_completed_iterm_image(image));
+        if self.bootstrap_stage == BootstrapStage::ScriptExecution
+            && !active_block_was_started
+            && self.active_block().started()
+        {
+            self.update_active_block_height();
```
vorporeal: can you help me understand why we need the extra logic here?  i would have expected that the call to `self.update_active_block_height()` in the `on_finish_byte_processing()` function would have been sufficient.

app/src/terminal/model/blockgrid.rs
KimHyeongRae0: Right, has_visible_content() includes images. This branch handles finished_len()==0: non-moving Kitty placements leave the cursor at the origin but still render. Added a clarifying comment.

app/src/terminal/model/blocks.rs
KimHyeongRae0: Agreed. For the normal parser path, on_finish_byte_processing() updates the height after the image starts the block. The immediate update was leftover from the earlier pre-delegation version, so I removed it.

KimHyeongRae0:
Quick status check: the post-review follow-up is in place, `verification/cla-signed` is green on `bbc4da99`, and this should be ready for maintainer review.

vorporeal:
hey @KimHyeongRae0 - i tried to fix up your PR to merge in the latest commits from `master` + align this with the changes made in #10927, but `master` has some LFS object changes and so i don't appear able to update your PR.

can you pull in the latest master changes, and then also apply this patch?  it'll adjust your implementation to match the #10927 behavior to simplify starting bootstrap blocks:

```diff
diff --git a/app/src/terminal/model/block.rs b/app/src/terminal/model/block.rs
index 425d31b0a..398691fc9 100644
--- a/app/src/terminal/model/block.rs
+++ b/app/src/terminal/model/block.rs
@@ -2981,10 +2981,6 @@ macro_rules! delegate_image_completion {
                     if !$self.output_grid.started() {
                         $self.output_grid.start();
                     }
-                    if $self.bootstrap_stage == BootstrapStage::ScriptExecution && !$self.started()
-                    {
-                        $self.start();
-                    }
                 }
                 retval
             }
diff --git a/app/src/terminal/model/blocks_tests.rs b/app/src/terminal/model/blocks_tests.rs
index 83215372a..e1041fa43 100644
--- a/app/src/terminal/model/blocks_tests.rs
+++ b/app/src/terminal/model/blocks_tests.rs
@@ -165,13 +165,13 @@ fn advance_to_bootstrapped(block_list: &mut BlockList, data: BootstrappedValue)
 }
 
 #[test]
-fn test_iterm_image_starts_script_execution_block() {
+fn test_iterm_image_renders_in_script_execution_block() {
     let _iterm_images = FeatureFlag::ITermImages.override_enabled(true);
     let mut block_list = TestBlockListBuilder::new().build();
     advance_to_script_execution(&mut block_list);
 
     assert_eq!(block_list.bootstrap_stage, BootstrapStage::ScriptExecution);
-    assert!(!block_list.active_block().started());
+    assert!(block_list.active_block().started());
 
     block_list.handle_completed_iterm_image(test_utils::test_iterm_image(1));
     block_list.on_finish_byte_processing(&ansi::ProcessorInput::new(&[]));
@@ -184,7 +184,7 @@ fn test_iterm_image_starts_script_execution_block() {
 }
 
 #[test]
-fn test_invalid_iterm_image_does_not_start_script_execution_block() {
+fn test_invalid_iterm_image_does_not_render_in_script_execution_block() {
     let _iterm_images = FeatureFlag::ITermImages.override_enabled(true);
     let mut block_list = TestBlockListBuilder::new().build();
     let mut image = test_utils::test_iterm_image(1);
@@ -195,24 +195,23 @@ fn test_invalid_iterm_image_does_not_start_script_execution_block() {
     advance_to_script_execution(&mut block_list);
 
     assert_eq!(block_list.bootstrap_stage, BootstrapStage::ScriptExecution);
-    assert!(!block_list.active_block().started());
+    assert!(block_list.active_block().started());
 
     block_list.handle_completed_iterm_image(image);
     block_list.on_finish_byte_processing(&ansi::ProcessorInput::new(&[]));
-
-    assert!(!block_list.active_block().started());
+    assert!(block_list.active_block().started());
     assert!(block_list.active_block().output_grid().is_empty());
 }
 
 #[test]
-fn test_kitty_image_starts_script_execution_block() {
+fn test_kitty_image_renders_in_script_execution_block() {
     let _kitty_images = FeatureFlag::KittyImages.override_enabled(true);
     let mut block_list = TestBlockListBuilder::new().build();
     let mut metadata = test_utils::test_kitty_image_metadata_map(1);
     advance_to_script_execution(&mut block_list);
 
     assert_eq!(block_list.bootstrap_stage, BootstrapStage::ScriptExecution);
-    assert!(!block_list.active_block().started());
+    assert!(block_list.active_block().started());
 
     block_list
         .handle_completed_kitty_action(
@@ -231,22 +230,21 @@ fn test_kitty_image_starts_script_execution_block() {
 }
 
 #[test]
-fn test_kitty_store_only_does_not_start_script_execution_block() {
+fn test_kitty_store_only_does_not_render_in_script_execution_block() {
     let _kitty_images = FeatureFlag::KittyImages.override_enabled(true);
     let mut block_list = TestBlockListBuilder::new().build();
     let mut metadata = test_utils::test_kitty_image_metadata_map(1);
     advance_to_script_execution(&mut block_list);
 
     assert_eq!(block_list.bootstrap_stage, BootstrapStage::ScriptExecution);
-    assert!(!block_list.active_block().started());
+    assert!(block_list.active_block().started());
 
     block_list
         .handle_completed_kitty_action(test_utils::test_kitty_store_only_action(1), &mut metadata)
         .expect("kitty action should be handled")
         .expect("kitty action should store");
     block_list.on_finish_byte_processing(&ansi::ProcessorInput::new(&[]));
-
-    assert!(!block_list.active_block().started());
+    assert!(block_list.active_block().started());
     assert!(block_list.active_block().output_grid().is_empty());
 }
 
```

vkodithala (dismissed):
Hey @KimHyeongRae0, gentle nudge to respond to @vorporeal's request here!

KimHyeongRae0:
@vorporeal done — pulled master in and applied the patch. The only conflicts were the new split-import formatting in `block_tests.rs`/`test_utils.rs`, nothing functional. Agreed the `ScriptExecution` branch is redundant now that the bootstrap transition already starts the block, so dropping it is cleaner. Renamed the four tests to match.

vorporeal (approved):
looks good!  kicked off CI :)

PR #10480, opened 2026-05-08
Hide "sharer's" cursor in cloud agent sessions

harryalbert:
## Summary

There was a bug in cloud mode for 3rd party agents where the viewer's cursor was being rendered at the beginning of the input. See the tech spec below for why this never surfaced before (the agent did some gnarly debugging to figure this out), but regardless, the headless cloud-agent runner should not be sending these kinds of events (the viewer doesn't/shouldn't care about where the "sharer's" cursor is for example when the sharer is a cloud agent runner).

## demo

https://www.loom.com/share/b687e1e5d661438d8f400c0130e94acc

_This PR was created by [Oz](https://warp.dev/oz) (running Codex)._

### DIFF
--- a/app/src/terminal/local_tty/terminal_manager.rs
+++ b/app/src/terminal/local_tty/terminal_manager.rs
@@ -1476,19 +1476,23 @@ impl TerminalManager {
 
                 // Flush the initial input operations that the sharer performed
                 // in the latest buffer before the share was started.
-                let init_input_ops: Vec<CrdtOperation> = terminal_view
-                    .as_ref(ctx)
-                    .input()
-                    .as_ref(ctx)
-                    .latest_buffer_operations()
-                    .cloned()
-                    .collect();
-                network.update(ctx, |network, _ctx| {
-                    network.send_input_update(
-                        model.lock().block_list().active_block_id(),
-                        init_input_ops.iter(),
-                    );
-                });
+                // Skip for ambient agent sessions — the sharer is a headless
+                // worker whose input ops would cause a phantom cursor.
+                if !model.lock().is_shared_ambient_agent_session() {
+                    let init_input_ops: Vec<CrdtOperation> = terminal_view
+                        .as_ref(ctx)
+                        .input()
+                        .as_ref(ctx)
+                        .latest_buffer_operations()
+                        .cloned()
+                        .collect();
+                    network.update(ctx, |network, _ctx| {
+                        network.send_input_update(
+                            model.lock().block_list().active_block_id(),
+                            init_input_ops.iter(),
+                        );
+                    });
+                }
 
                 // Stream historical agent conversations so viewers have conversation and task context.
                 if FeatureFlag::AgentSharedSessions.is_enabled() {
@@ -2316,6 +2320,14 @@ impl TerminalManager {
                 // If the block ID has become stale by the time we get here,
                 // we don't need to send this update to the server.
                 if model.lock().block_list().active_block_id() != block_id {
+                    return;
+                }
+
+                // In cloud agent sessions the sharer is a headless worker
+                // that never interactively types in the rich input. Sending
+                // its CRDT selection ops would cause a phantom cursor to
+                // render on the viewer side.
+                if model.lock().is_shared_ambient_agent_session() {
                     return;
                 }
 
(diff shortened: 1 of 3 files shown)


harryalbert:
* **#10480** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10480?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10480?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

specs/REMOTE-1590/TECH.md
```diff
@@ -0,0 +1,50 @@
...
+2. The buffer has **remote selection data** for that peer's replica ID, populated by `UpdateSelections` CRDT operations received from the sharer.
+
+Condition 1 is always met: `handle_presence_manager_event` unconditionally registers the sharer as a remote peer when a viewer joins any shared session (`view_impl.rs:1132-1158`).
+
+Condition 2 was historically never met for cloud agent sessions because the worker's input buffer was inert — no code path on the worker ever called `edit()` on the rich input buffer.
+
+### What changed
+The CLI agent rich input feature introduced lifecycle events (`InputSessionChanged::Open` / `Closed`) that call `clear_buffer_and_reset_undo_stack` on the worker's input (`input.rs:2477`, `2492`). Unlike `reinitialize_buffer` (which creates a fresh buffer without emitting CRDT operations), `clear_buffer_and_reset_undo_stack` goes through the full `edit()` → `end_batch()` → `UpdatePeers` path, emitting CRDT selection operations. These operations are transmitted to the viewer via session sharing (`local_tty/terminal_manager.rs:2312-2326`), satisfying condition 2.
```
bnavetta, line 16: Think the answer is no, but would it make sense to avoid `clear_buffer_and_reset_undo_stack` in the CLI rich agent input and use other editor operations that don't emit CRDT updates?

bnavetta (approved):

specs/REMOTE-1590/TECH.md
harryalbert: I think we probably could? I asked the agent about this and it mentioned that there might be an edge case involved with using `reinitialize_buffer` instead, but we might just be able to replace `clear_buffer_and_reset_undo_stack` with `reinitialize_buffer` and avoid the emission.

That being said, I don't think we _should_ necessarily. I don't think it should be wrong/to-be-avoided for the ambient sharer to emit CRDT updates — I think the viewer should just ignore them.

harryalbert: Like, this solution feels more extensible and robust than fixing this specific case to just use a non-CRDT-emitting operation, in case in the future some other operation is triggered that also emits some CRDT op, and we hit the same issue (which seems likely)

harryalbert: does that make sense to you, or do you disagree/am I missing something here

bnavetta: No, I think fixing the general case seems right. I'm a little worried about skipping all sharer CRDT operations for cloud agents - if the sharer needs to write to the input for any reason, the buffer state could get out of sync. If we can, it'd be safer to _only_ skip selection operations.

I do think there's an argument that we should treat each rich input session as a new "command", which is what `reinitialize_buffer` does, but if that introduces other issues it seems fine not to change.

app/src/terminal/local_tty/terminal_manager.rs
```diff
@@ -1476,19 +1476,23 @@ impl TerminalManager {
...
-                    .collect();
-                network.update(ctx, |network, _ctx| {
-                    network.send_input_update(
-                        model.lock().block_list().active_block_id(),
-                        init_input_ops.iter(),
-                    );
-                });
+                // Skip for ambient agent sessions — the sharer is a headless
+                // worker whose input ops would cause a phantom cursor.
+                if !model.lock().is_shared_ambient_agent_session() {
```
zachbai: should maybe consider a higher-level helper `should_broadcast_input_events_as_sharer()` or something to put this behind

zachbai (approved):

specs/REMOTE-1590/TECH.md
harryalbert: yeah that makes sense, I'll give that a shot and see if things still WAI

PR #10496, opened 2026-05-08
spec: option to clamp/disable truecolor BCE bg painting (#10278)

### LINKED ISSUE #10278: Add option to disable per-line truecolor background painting (BCE-style flood) for diff-style ANSI output
## Summary

Warp currently paints `\x1b[48;2;R;G;Bm` (truecolor background) ANSI codes edge-to-edge across the full block width for every emitting line. When a tool (e.g. Claude Code's diff renderer, `delta`, `git diff` with custom config) emits contiguous lines all carrying the same bg code, the visible viewport is filled with a solid wall of color — perceived as the entire screen being tinted.

Please add a setting to either:
1. **Clamp** truecolor bg painting to the printed text width (don't extend to end of line / block), OR
2. **Disable** Background Color Erase (BCE) / per-line bg painting globally, OR
3. **Dim or desaturate** truecolor bg fills past a configurable threshold.

## Repro

1. Run any tool that emits truecolor bg codes on entire `+`/`-` diff lines (e.g. Claude Code with `theme: "dark-daltonized"` viewing a large patch).
2. Open a hunk where the visible viewport is 100% `+` or `-` lines (no context).
3. Observe: the entire viewport is flooded teal (`+`) or red (`-`).
4. Compare against pure `git diff --color=always` (foreground-only codes — no flood).
5. Compare against the same content in iTerm2 / Ghostty / Alacritty — visible flood still occurs in all of them, but with structural chrome (sidebars, padding) the perception is less aggressive in some.

## Why this matters

- **Eye strain / accessibility** — sustained large-area saturated bg fills are uncomfortable and can be migraine-triggering. Users who rely on colorblind-accessible themes (e.g. teal/orange) face worse contrast trade-offs.
- **No current workaround at the Warp layer** — verified via:
  - `~/.warp/settings.toml` — no per-line bg / BCE / clamp keys
  - `dev.warp.Warp-Stable` plist — no relevant render keys
  - Custom themes (`~/.warp/themes/*.yaml`) — palette swap doesn't change painting behavior
  - Window opacity / blur — affects whole window, not just bg fills
  - Vertical tabs / sidebar — adds chrome but doesn't reduce the painted area
  - Hard `⌘Q` + reopen — no effect (not a render-state staleness issue)

## Proposed UX

```toml
[appearance.rendering]
truecolor_bg_clamp = "text_width"   # or "off" / "block" (current behavior, default)
truecolor_bg_max_saturation = 0.6   # optional 0.0–1.0 desaturation cap
```

Either a single binary toggle or the more granular options above would solve it.

## System info

- Warp Stable on macOS (Darwin 25.4.0)
- Reproduced across multiple themes (built-in Phenomenon, custom Cmux palette, dark-daltonized-equivalent)

Happy to provide a screen recording or specific repro stream if useful.

lonexreb:
Closes #10278

Spec for #10278.

## Summary

Adds a setting `terminal.truecolor_background_mode` to control how Warp renders truecolor (24-bit) ANSI background codes (`\x1b[48;2;R;G;Bm`). Three modes:

- `flood` (default) — current behavior, paints to end of line/block.
- `clamp_to_text` — paints only behind printed glyphs; cells past the last write get default bg.
- `disabled` — truecolor bg ANSI is parsed but not rendered as bg fill; fg colors preserved.

Per-pane override via tab config. 256-color and 16-color BCE behavior is unchanged.

## Why

Modern diff/highlight tools (Claude Code's diff renderer, `delta`, `git diff` with custom config) emit contiguous bg-colored lines via truecolor codes. Warp's BCE flood paints those across the full block width on every line, so the cumulative effect reads as a full-screen tint that overwhelms the rendered content. The screenshots in #10278 show this directly.

## Scope

Spec-only — no code changes in this PR. See `specs/GH10278/SPEC.md` for behavior contract, acceptance criteria, implementation pointers, tests, and the one open question (whether explicit space writes should count as bg-painted in `clamp_to_text`).

Tag: `ready-to-spec`.

### DIFF
--- a/specs/GH10278/SPEC.md
+++ b/specs/GH10278/SPEC.md
@@ -44,7 +44,10 @@ entirely (`disabled`).
 
 ### B1. Setting enum
 
-`terminal.truecolor_background_mode` enum: `"flood"` (default), `"clamp_to_text"`,
+The user-level setting key is **`appearance.terminal.truecolor_background_mode`**
+(matches the verified `appearance.*` namespace in `app/src/settings/pane.rs`
+and is the canonical key used everywhere in this spec — Settings/API, UI,
+acceptance, and tests). Enum values: `"flood"` (default), `"clamp_to_text"`,
 `"disabled"`.
 
 ### B2. Mode `flood` (default)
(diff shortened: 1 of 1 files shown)


lonexreb:
Round-1 fixes for oz-for-oss review: (1) all bg-fill paths enumerated and aligned with the mode — cursor-to-EOL, erase ops (`\x1b[K`/`\x1b[J`), insert/delete line, scroll regions, block finalization padding all follow the same `terminal.truecolor_background_mode`; uniform principle stated; (2) verified `terminal.*` settings and `pane.*` tab-config override match existing schemas; module paths updated to actual codebase locations; (3) resolved the explicit-spaces conflict — `clamp_to_text` paints explicitly-written spaces; only suppresses IMPLICIT fill (EOL flood, erase ops, scroll, padding); acceptance + tests aligned.

lonexreb:
Addressed both important concerns from the latest oz-for-oss review:

1. **Setting key consistency (B1 vs Settings/API)**: Renamed B1 to use the canonical `appearance.terminal.truecolor_background_mode` key everywhere. The spec now uses one key — the one matching the verified `appearance.*` namespace in `app/src/settings/pane.rs`.
2. **Per-pane override TOML syntax (B5, A7, T_per_pane_override_toml)**: Verified `TabConfigPaneNode` in `app/src/tab_configs/tab_config.rs` — fields are flat keys on `[[panes]]` array entries (no `[pane]` table; struct is `#[serde(deny_unknown_fields)]`). Updated the spec to use a flat `truecolor_background_mode` key inside `[[panes]]` and added a complete TOML example with mixed-mode panes. Updated A7 and T_per_pane_override_toml to match.

Commit: b0d508ba

PR #10514, opened 2026-05-08
Fix remote file tree not updating with local file changes

kevinyang372:
## Problem
The file tree in remote SSH sessions does not update when files change on the server. Cd-ing in and out of the directory restores the correct state.

## Root Cause
The daemon process creates its `RepoMetadataModel` with `emit_incremental_updates = false`.

`initialize_app()` calls `RepoMetadataModel::new(ctx)` for **all** launch modes, including `RemoteServerDaemon`. The `new()` constructor leaves `emit_incremental_updates` disabled on the local sub-model. When the file system watcher detects changes and `apply_file_tree_mutations` runs, it returns `None` instead of a `RepoMetadataUpdate`, so `IncrementalUpdateReady` is never emitted.

The `ServerModel` subscribes to `IncrementalUpdateReady` events to push them to clients, but they never arrive.

Cd-ing fixes the issue because `NavigatedToDirectory` triggers a full snapshot push (via `RepositoryUpdated` or directly in the handler), which replaces the entire client-side tree.

This has been broken since the initial commit — the `new_with_incremental_updates` constructor existed but was never wired into the daemon initialization path.

## Fix
In `initialize_app()`, use `RepoMetadataModel::new_with_incremental_updates(ctx)` when the launch mode is `RemoteServerDaemon`.

---
[Oz conversation](https://staging.warp.dev/conversation/c0f4ca93-e310-48aa-afd1-0ccc6b8e384d) | [Plan](https://staging.warp.dev/drive/notebook/ehhdaoqpILmQhzyNRukMbl)

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -1464,8 +1464,7 @@ pub(crate) fn initialize_app(
             });
         }
 
-        let emit_incremental_updates =
-            matches!(launch_mode, LaunchMode::RemoteServerDaemon { .. });
+        let emit_incremental_updates = matches!(launch_mode, LaunchMode::RemoteServerDaemon { .. });
         ctx.add_singleton_model(|ctx| {
             let model = if emit_incremental_updates {
                 RepoMetadataModel::new_with_incremental_updates(ctx)
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -1464,8 +1464,14 @@ pub(crate) fn initialize_app(
             });
         }
 
+        let emit_incremental_updates =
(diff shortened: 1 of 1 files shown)


kevinyang372:
* **#10514** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

moirahuang (approved):

PR #10520, opened 2026-05-08
Migrate editor to use remote backed buffer

kevinyang372:
## Description
Wire editor to use remote backed buffer. A few noteable things:
1. LocalCodeEditor (I will change its name in a follow-up) now takes a FileLocation, which could be a local / remote file
2. Existing LocalCodeEditor concepts like conflict resolution banner is now hooked up with the remote logic

Also fixed a bug where server detected conflicts aren't being properly forwarded to the client

This also removes the remote file tree gating on not opening files

## Testing
Tested locally

### DIFF
--- a/app/src/code/file_tree/view.rs
+++ b/app/src/code/file_tree/view.rs
@@ -42,6 +42,7 @@ use warpui::{
 use warpui::{BlurContext, ModelHandle};
 
 use crate::code::active_file::{ActiveFileEvent, ActiveFileModel};
+use crate::code::buffer_location::FileLocation;
 use crate::coding_panel_enablement_state::CodingPanelEnablementState;
 use crate::editor::{EditorOptions, EditorView, TextOptions};
 #[cfg(feature = "local_fs")]
@@ -2224,7 +2225,7 @@ impl FileTreeView {
         );
 
         ctx.emit(FileTreeEvent::OpenFile {
-            path: path.to_path_buf(),
+            path: FileLocation::Local(path.to_path_buf()),
             target,
             line_col: None,
         });
@@ -2253,7 +2254,11 @@ impl FileTreeView {
                             host_id.clone(),
                             (*metadata.path).clone(),
                         );
-                        ctx.emit(FileTreeEvent::OpenRemoteFile { remote_path });
+                        ctx.emit(FileTreeEvent::OpenFile {
+                            path: FileLocation::Remote(remote_path),
+                            target: FileTarget::CodeEditor(EditorLayout::SplitPane),
+                            line_col: None,
+                        });
                     }
                 } else {
                     let path = metadata.path.to_local_path_lossy();
@@ -2879,14 +2884,9 @@ pub enum FileTreeEvent {
     AttachAsContext { path: PathBuf },
     #[cfg_attr(not(feature = "local_fs"), allow(dead_code))]
     OpenFile {
-        path: PathBuf,
+        path: FileLocation,
         target: FileTarget,
         line_col: Option<LineAndColumnArg>,
-    },
-    /// Open a remote file in the code editor.
-    #[allow(dead_code)]
-    OpenRemoteFile {
-        remote_path: warp_util::remote_path::RemotePath,
     },
     #[cfg_attr(not(feature = "local_fs"), allow(dead_code))]
     FileRenamed {
--- a/app/src/code/file_tree/view.rs
+++ b/app/src/code/file_tree/view.rs
@@ -67,7 +67,6 @@ use crate::{
 use warp_core::features::FeatureFlag;
 use warp_core::ui::theme::{color::internal_colors, Fill};
 use warp_core::HostId;
-use warpui::ui_components::components::UiComponent;
 
 mod editing;
 mod render;
@@ -1978,7 +1977,6 @@ impl FileTreeView {
         let is_selected = self.selected_item.as_ref() == Some(id);
         let is_expanded = self.is_item_expanded(&id.root, item);
         let render_state = item.to_render_state(is_expanded, appearance);
-        let is_remote_file = root_dir.is_remote() && matches!(item, FileTreeItem::File { .. });
 
         let item_display_name = render_state.display_name.clone();
         let item_position_id = format!("file_tree_item:{item_display_name}");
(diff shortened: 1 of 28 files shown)


kevinyang372:
* **#10520** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10520?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10520?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/code/global_buffer_model.rs
```diff
@@ -1745,7 +1827,7 @@ impl GlobalBufferModel {
             return Err(FileSaveError::RemoteError("Buffer deallocated".to_string()));
         };
         let content = buffer.as_ref(ctx).text().into_string();
-        let version = ContentVersion::new();
+        let version = buffer.as_ref(ctx).version();
```
kevinyang372: This was wrong previously. When saving, we should ensure the FileModel version is consistently with the internal buffer version

app/src/code/global_buffer_model.rs
```diff
@@ -1606,57 +1664,81 @@ impl GlobalBufferModel {
...
+                if let BufferSource::Remote { sync_clock, .. } = &mut state.source {
+                    *sync_clock = Some(SyncClock::from_wire(response.server_version, 0));
+                }
+                let Some(buffer) = state.buffer.upgrade(ctx) else {
+                    log::warn!("[remote-buffer] Buffer handle deallocated for file_id={file_id:?}");
+                    return;
+                };
+                let version = ContentVersion::new();
+                buffer.update(ctx, |buffer, ctx| {
+                    buffer.replace_all(&response.content, ctx);
```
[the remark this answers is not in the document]
kevinyang372: This is the existing behavior for local editors too. I am not too worried about it

app/src/code/global_buffer_model.rs
```diff
@@ -720,6 +728,36 @@ impl GlobalBufferModel {
...
+                };
+                ctx.spawn(
+                    async move { client.save_buffer(path).await.map_err(|e| format!("{e}")) },
+                    move |_me, result, ctx| match result {
+                        Ok(()) => {
+                            ctx.emit(GlobalBufferModelEvent::FileSaved { file_id });
+                        }
+                        Err(error) => {
+                            log::warn!("Remote save failed: {error}");
+                            ctx.emit(GlobalBufferModelEvent::FailedToSave {
```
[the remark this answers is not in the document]
kevinyang372: I will follow-up with a fix for the unsaved changes flow. This is not working fully right currently

app/src/code/global_buffer_model.rs
```diff
@@ -1823,6 +1905,75 @@ impl GlobalBufferModel {
...
+            });
+            return;
+        };
+
+        log::info!("[remote-buffer] Re-opening buffer: path={path_str}");
+        // Close first so the server deallocates the old in-memory buffer
+        // (which contains stale client edits). The subsequent OpenBuffer
+        // will create a fresh buffer and read from disk.
+        let path_for_close = path_str.clone();
+        client.close_buffer(path_for_close);
```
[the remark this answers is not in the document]
kevinyang372: This is a very good call. Updated to have a force_reload flag for opening buffers

app/src/code/file_tree/view.rs
```diff
@@ -2273,8 +2247,20 @@ impl FileTreeView {
 
         match item {
             FileTreeItem::File { metadata, .. } => {
-                // Remote file trees don't support opening files in the editor.
```
moirahuang, line 2276: very niceeeee

app/src/code/global_buffer_model.rs
```diff
@@ -1526,9 +1574,13 @@ impl GlobalBufferModel {
 
         // Subscribe to buffer content changes so edits are sent back to the daemon.
         let client_for_sub = {
-            let manager = remote_server::manager::RemoteServerManager::handle(ctx);
+            let manager = RemoteServerManager::handle(ctx);
             manager.as_ref(ctx).client_for_host(&host_id).cloned()
         };
+        log::debug!(
```
moirahuang, line 1580: nit: do we need the log debugs here and below?

app/src/code/global_buffer_model.rs
```diff
@@ -1606,57 +1664,81 @@ impl GlobalBufferModel {
...
+                    "[remote-buffer] OpenBuffer response: content_len={} server_version={}",
+                    response.content.len(),
+                    response.server_version,
+                );
+                let Some(state) = self.buffers.get_mut(&file_id) else {
+                    log::warn!("[remote-buffer] Buffer state missing for file_id={file_id:?}");
+                    return;
+                };
+                if let BufferSource::Remote { sync_clock, .. } = &mut state.source {
+                    *sync_clock = Some(SyncClock::from_wire(response.server_version, 0));
```
moirahuang: why do we save this version as 0 vs. the latest version?

moirahuang (commented):
1. the delay is kinda long when overwriting, can we auto dismiss and then bring back the banner if it fails to save
https://github.com/user-attachments/assets/56e11d37-251f-4329-8c02-0692982b22e6

2. when we discard, we aren't correctly writing the first time. it only correctly writes the 2nd time
https://github.com/user-attachments/assets/221b8d00-f826-48da-8e03-ca39d395637e

app/src/code/global_buffer_model.rs
```diff
@@ -1884,14 +2171,33 @@ impl GlobalBufferModel {
...
+                     push_sv={new_server_version} <= local_sv={:?}",
+                    sync_clock.server_version
+                );
+                return;
+            }
             // Conflict — local edits diverged from server.
-            log::info!(
-                "Remote buffer conflict for {path}: expected C={expected_client_version}, \
-                 local C={:?}",
+            log::warn!(
```
moirahuang: does this need to be `warn`?

crates/warp_files/src/lib.rs
```diff
@@ -638,8 +647,14 @@ impl FileModel {
...
                         WatcherType::Individual => {
+                            // Unwatch the parent directory (matching the register
+                            // in open() which watches the parent, not the file).
+                            let watch_path = path
+                                .parent()
+                                .map(|p| p.to_path_buf())
+                                .unwrap_or_else(|| path.clone());
                             self.watcher.update(ctx, |watcher, _ctx| {
-                                std::mem::drop(watcher.unregister_path(path.as_path()));
+                                std::mem::drop(watcher.unregister_path(&watch_path));
```
moirahuang: should we reference count here so if there's other paths to the parent directory we don't unconditionally drop?

app/src/code/global_buffer_model.rs
```diff
@@ -1823,6 +1920,190 @@ impl GlobalBufferModel {
...
+                    state.set_base_content_version(new_version);
+                    FileModel::handle(ctx).update(ctx, |file_model, _ctx| {
+                        file_model.set_version(file_id, new_version);
+                    });
+
+                    // Bump the server version in the sync clock.
+                    let new_server_version =
+                        if let BufferSource::ServerLocal { sync_clock, .. } = &mut state.source {
+                            let sv = sync_clock.bump_server();
+                            // Reset client version since the buffer is fresh from disk.
```
moirahuang: i see you added this comment but i'm sorry :woman-bowing: i still don't understand why it's reset vs. being the latest disk version

moirahuang (commented):
higher level comment, i think there's some `log::debug`s that can be cleaned up

moirahuang (approved):

app/src/code/global_buffer_model.rs
kevinyang372: I will keep this for now in case of future debugging

app/src/code/global_buffer_model.rs
kevinyang372: This is necessary. The client to server contract is a newly opened buffer starts with client version 0 (which makes sense because it's fresh). If we don't reset here, server and client will disagree about their tracked client version on the next sync

crates/warp_files/src/lib.rs
kevinyang372: Good catch

PR #10537, opened 2026-05-09
Do less work in command palette zero state

acarl005:
## Description

This is another optimization for #10256

In this PR, we remove unnecessary work when showing the command palette zero state.

## Linked Issue

Context for this is in Slack.

## Testing

I sent a custom build to the reporter in Slack and they confirmed this is much better.

### DIFF
--- a/app/src/search/command_palette/view.rs
+++ b/app/src/search/command_palette/view.rs
@@ -521,6 +521,10 @@ impl View {
     fn on_session_source_changed(&mut self, ctx: &mut ViewContext<Self>) {
         self.compute_recent_items_for_zero_state(ctx);
 
+        if self.search_bar_state.as_ref(ctx).should_show_zero_state() {
+            return;
+        }
+
         self.search_bar.update(ctx, |search_bar, ctx| {
             search_bar.run_query(ctx);
         });
--- a/app/src/search/command_palette/view.rs
+++ b/app/src/search/command_palette/view.rs
@@ -179,7 +179,7 @@ impl warpui::View for View {
         let appearance = Appearance::as_ref(app);
         let theme = appearance.theme();
 
-        let body = if self.search_bar_state.as_ref(app).should_show_zero_state() {
+        let body = if self.search_bar.as_ref(app).should_show_zero_state(app) {
             ChildView::new(&self.zero_state_handle).finish()
         } else {
             self.render_palette_list(theme, app)
@@ -364,15 +364,15 @@ impl View {
     }
 
     pub fn select_next_item(&mut self, ctx: &mut ViewContext<Self>) {
-        self.search_bar_state.update(ctx, |state, ctx| {
-            state.handle_selection_update(SelectionUpdate::Down, ctx);
+        self.search_bar.update(ctx, |search_bar, ctx| {
+            search_bar.handle_selection_update(SelectionUpdate::Down, ctx);
         });
         ctx.notify();
     }
 
     pub fn select_prev_item(&mut self, ctx: &mut ViewContext<Self>) {
-        self.search_bar_state.update(ctx, |state, ctx| {
-            state.handle_selection_update(SelectionUpdate::Up, ctx);
+        self.search_bar.update(ctx, |search_bar, ctx| {
+            search_bar.handle_selection_update(SelectionUpdate::Up, ctx);
         });
         ctx.notify();
     }
--- a/app/src/search/command_palette/view.rs
+++ b/app/src/search/command_palette/view.rs
@@ -521,11 +521,9 @@ impl View {
     fn on_session_source_changed(&mut self, ctx: &mut ViewContext<Self>) {
         self.compute_recent_items_for_zero_state(ctx);
 
-        if self.search_bar_state.as_ref(ctx).should_run_query() {
-            self.search_bar.update(ctx, |search_bar, ctx| {
-                search_bar.run_query(ctx);
-            });
-        }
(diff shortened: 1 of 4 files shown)


app/src/search/command_palette/view.rs
```diff
@@ -521,6 +521,10 @@ impl View {
     fn on_session_source_changed(&mut self, ctx: &mut ViewContext<Self>) {
         self.compute_recent_items_for_zero_state(ctx);
 
+        if self.search_bar_state.as_ref(ctx).should_show_zero_state() {
+            return;
+        }
+
```
acarl005: short-circuit this in the zero state. we don't need to do this work as we won't be showing these without a query

app/src/search/search_bar.rs
```diff
@@ -110,9 +100,8 @@ pub struct SearchBarState<T: Action + Clone> {
     /// The filter atom text to be rendered in the Search input, representing the currently applied
     /// filter.
     filter_atom_text: Option<&'static str>,
-    /// Whether the search bar is in a state where a zero state should be shown. This is true if
-    /// no search query has been run and no filters have been applied.
-    should_show_zero_state: bool,
+    /// Whether the search bar's editor buffer is currently empty.
+    buffer_is_empty: bool,
```
acarl005: `should_show_zero_state` is derived state, i.e. it is computed from other fields and needs to be updated together, e.g. `fn set_fixed_filters` was setting both `state.query_filter` *and* `state.should_show_zero_state` when just setting `state.query_filter` should be sufficient. This was actually _more_ of an actual problem inside `fn set_visible_query_filter`. Although both fields were updated, they were updated in different places in the function. That means if anything wanted to read the state in between those 2 updates, they might see an inconsistent view of the state. I got rid of the redundant field and new `fn should_show_zero_state` reads from the base fields that actually determine this

app/src/search/search_bar.rs
```diff
@@ -92,16 +92,6 @@ pub enum FilterState {
...
-impl FilterState {
-    fn zero_state(&self) -> bool {
-        match self {
-            FilterState::Unfiltered => true,
-            FilterState::Visible(_) => false,
-            FilterState::Fixed { .. } => true,
-        }
-    }
-}
-
```
acarl005, line 104: This is yet another fragmentation of this "should zero state show" concept. I would even argue that it's an error that `FilterState::Fixed` returns `true` here. It turns out that the entrypoint for adding a fixed filter, `fn set_fixed_filters`, was overriding this by setting `state.should_show_zero_state = false` anyways!

https://github.com/warpdotdev/warp/pull/10537/changes#diff-6d6577976d0f93f339110370c6e4e18621994c9bf3f3f051fa2754d873118852L672

FYI "fixed filters" are programmatically applied filters that the user cannot see. This is useful for more specific palette configurations. It was added to support a palette for a "recent conversations and projects" palette which was added as part of code mode. This is of course now dead code since we moved to the agent view.

app/src/search/search_bar.rs
```diff
@@ -751,6 +731,12 @@ impl<T: Action + Clone> SearchBar<T> {
         });
     }
 
+    pub fn run_query(&mut self, ctx: &mut ViewContext<Self>) {
+        if self.state.as_ref(ctx).should_run_query() {
+            self.run_query_internal(ctx);
+        }
+    }
+
```
acarl005, line 727: We have an internal and external version of this method where the external one actually enforces `should_run_query`. I'm worried about future callers violating this and not checking `should_run_query` so I force the pub one to check it.

There was *one particular caller* which was the culprit of the lag:

https://github.com/warpdotdev/warp/blob/7231efc6ced916f8292a154d8d44e02cbdbbd161/app/src/search/command_palette/view.rs#L525-L525

This `fn on_session_source_changed` is called when the command palette is opened in the zero-state. Before this change, `fn run_query` would cause every single mixer to run the query with an empty query, including the project files. After this change, that work is skipped in the zero state.

app/src/search/search_bar.rs
alokedesai: not urgent but can we remove the code mode dead code in a followup

app/src/search/search_bar.rs
```diff
@@ -833,22 +819,20 @@ impl<T: Action + Clone> SearchBar<T> {
...
             .editor_handle
             .read(ctx, |editor, ctx| editor.buffer_text(ctx));
-        if current_buffer_text.is_empty() && self.state.as_ref(ctx).query_filter.zero_state() {
-            self.state.update(ctx, |state, ctx| {
-                state.should_show_zero_state = true;
-                ctx.notify();
-            });
+
+        let buffer_is_empty = current_buffer_text.is_empty();
+        self.state.update(ctx, |state, ctx| {
```
alokedesai: not blocking but should the `state` hold onto a weakview handle and directly read whether the buffer is empty? It's a little brittle that we have to ensure `buffer_is_empty` is correctly set for all of this to work

alokedesai (approved):
This looks good.

I think we can do even more to optimize the zero state query for the file filter though. We're doing way too much eager work for the zero state where we only really need to show ~50 results at most.

app/src/search/search_bar.rs
acarl005: way ahead of ya
https://oz.staging.warp.dev/runs/019e1fae-0650-7549-b179-1661fe85cc11

app/src/search/search_bar.rs
acarl005: good point. fixed

Issue #10003, opened 2026-05-04
Add a Command Palette action to reload Wildtree


bholmesdev:
## Summary
Add a Command Palette action to reload Wildtree / the File Tree so users have a deliberate way to reconcile the explorer with disk when automatic refresh misses a filesystem change.

## Problem
A user reported that the file tree does not always refresh after terminal commands that change the filesystem, especially when the opened directory does not have a `.git` directory. One example they noticed is running a CLI deletion command on a directory and not seeing the file explorer update afterward.

This has not been reproduced locally yet, so the underlying automatic-refresh failure is still uncertain. Regardless, when the tree appears stale there is currently no obvious Command Palette action for the user to force Wildtree to reload. A visible toolbar refresh button would likely feel too heavy or janky for the file tree, but a Command Palette action would provide an explicit recovery path without adding persistent UI chrome.

Warp's file tree documentation says the tree automatically reflects project structure as files are added, removed, or changed, so this action should be a fallback rather than the primary refresh mechanism.

## Expected behavior
- The file tree continues to refresh proactively when terminal commands or filesystem watchers detect added, removed, or changed files.
- If the tree appears stale, users can open the Command Palette and run an action such as `Reload File Tree`, `Refresh File Tree`, or `Reload Wildtree`.
- The action reloads Wildtree from disk for the current project/session, including directories that are not Git repositories.

## Actual behavior
- A user reported that the file tree can remain stale after some terminal filesystem changes, including a CLI directory deletion.
- The report specifically calls out cases where there is no `.git` directory.
- There does not appear to be an obvious Command Palette action to reload Wildtree / the File Tree when this happens.

## Reproduction steps or desired workflow
Reported scenario:
1. Open Warp in a directory with the file tree visible.
2. Use a directory that may not contain a `.git` directory.
3. Run a terminal command that changes the filesystem, such as deleting a visible child directory from the CLI.
4. Observe that the file tree may not update to reflect the deletion.

Desired workflow:
1. Open the Command Palette.
2. Search for a file-tree reload action, for example `Reload File Tree` or `Reload Wildtree`.
3. Run the action.
4. Wildtree reloads from disk and the file tree reflects the current filesystem state.

## Artifacts
None attached.

## Warp version
v0.2026.05.01.15.43.dev_00

## Operating system
macOS 26.3.1 (build 25D771280a)

moirahuang:
[...]

Issue #9890, opened 2026-05-02
File Explorer refresh rate is too slow


inventumamet:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

The File Explorer doesn't refresh when new files are ccreated or file are renamed. I have click away and click back to force refresh.

### To reproduce

Rename a file form the commandline and the file explorer still shows teh old file

### Expected behavior

immediate refresh file name (<1s )

### Operating system (OS)

macOS

### Operating system and version

Latest 

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

zachlloyd:
Possible to see a screenrecording of your experience @inventumamet ? this doesn't repro for me

isleofgreg:
Here's my reproduction of the issue, it definitely happens and it's even worse than what OP said. It happens regardless of whether you did anything in the terminal commands or not. Here in the "before.png" you can see how there are bunch of files on my desktop that I've outlined in red that were only there for a split second and disappeared because they were temporary. But you can see that Warp still reports them as being there until I click to Agent Conversations or Global Search and then click back to Project Explorer.

<img width="512" height="1184" alt="Image" src="https://github.com/user-attachments/assets/137bbdf7-0159-403f-8d7a-3d1b304e9916" />
<img width="480" height="464" alt="Image" src="https://github.com/user-attachments/assets/3f357bb6-e271-4217-866f-3d160de306c1" />

leesa-ui-w:
[...]

Issue #9190, opened 2026-04-25
Warp's Project Explorer does not Refresh


LaniusSenatorShrike:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

I updated Warp with the latest version hoping the bug reported in https://github.com/warpdotdev/Warp/issues/7900 is resolved but still the project explorer doesn't get refreshed and there's also any refresh button there too. 

### To reproduce

edit the rep by adding removing a directory and simply nothing changes on the project explorer tab

### Operating system (OS)

Linux

### Operating system and version

Ubuntu

### Regression

Yes, this bug started recently or with an X Warp version

### Does this block you from using Warp daily?

Yes, this issue prevents me from using Warp daily.

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

alokedesai:
Hey @LaniusSenatorShrike thanks for filing and sorry for the issues here.

Is this reproducing for you consistently? Is it specifically inside or outside of a git repository? Any more details would help significantly in figuring out where the issue is. Thanks!

LaniusSenatorShrike:
Hey @alokedesai , yes it's consistently there and project explorer only updates when I quit Warp and enter again. This is inside the git directory on my local and haven't checked for other directories.

bettercbd:
Can I refresh it manually?
Seeing an incorrect file list prevents me from continuing to use this app.

michabbb:
> Can I refresh it manually?
> Seeing an incorrect file list prevents me from continuing to use this app.

Do you see any refresh option? Exactly, that's the next issue here. That makes it unusable for me as well.

[...]

PR #10551, opened 2026-05-09
fix(file-tree): add Refresh to directory right-click menu

### LINKED ISSUE #10205: Project Explorer does not refresh to show newly created files and folders
## Summary
Project Explorer does not auto-update when new folders or files are created in the current project.

## Problem
When working in a project, newly created filesystem entries are not reflected in Project Explorer in real time, making the tree view stale and unreliable.

## Expected behavior
Project Explorer refreshes automatically and shows newly created files/folders shortly after they are created.

## Actual behavior
Newly created files/folders do not appear in Project Explorer without additional manual action (unknown whether a manual refresh, tab reload, or restart is required).

1. Open a project in Warp with Project Explorer visible.
2. Create a new folder or file in that project (for example from the shell or an external editor).
3. Observe Project Explorer.
4. New entries are not shown automatically.

## Artifacts
None attached

## Warp version
v0.2026.04.29.08.57.stable_02

## Operating system
Ubuntu 24.04.1 LTS (Linux kernel 5.15.167.4-microsoft-standard-WSL2)

spalagu:
## Summary

Adds a `Refresh` menu item under directory headers in the file tree right-click context menu. Re-reads the clicked directory's contents from disk (one level deep, matching lazy-load behavior) via the existing `LocalRepoMetadataModel::load_directory` path.

## Why

Long-standing open issues report Project Explorer / file tree failing to reflect filesystem changes after the initial scan:

- #10205 — newly created files/folders don't appear
- #9190 — only quitting and reopening Warp updates the tree
- #9890 — refresh too slow on macOS

Source-level: in `apply_file_tree_mutations` (`crates/repo_metadata/src/local_model.rs`), three branches do `if lazy_load && !is_parent_loaded_in_entry { continue }`, silently dropping watcher mutations when a parent directory is not loaded. Combined with FSEvents itself being best-effort, the in-memory tree drifts from disk and there is currently no API or UI to reconcile.

## What this PR does

- New `FileTreeAction::RefreshDirectory { id }` variant
- Right-click menu on any directory header now shows `Refresh` after `Open in new tab`
- Dispatch handler calls a new `refresh_directory` helper
- Helper resolves the target directory and calls the existing wrapper `RepoMetadataModel::load_directory(repo_root, dir_path, ctx)`, which delegates to `LocalRepoMetadataModel::load_directory` → `Entry::load` → `Entry::build_tree` (force-reads from disk)

Surgical scope: only the targeted subtree is re-scanned. No watcher churn, no full root re-index. `FileTreeItem::File` does not get the option since files don't have contents to refresh.

## What this PR does NOT do (intentionally scoped out)

- **Does not fix the underlying watcher event drop** — the lazy_load `continue` paths in `apply_file_tree_mutations` still silently skip mutations. That is a separate, deeper architectural fix (would need observability + force-rescan + reconciliation hooks layered on top, see analysis in #10205).
- **Does not implement the Command Palette action requested in #10003** — that is tracked separately by spec PR #10025. This PR adds the right-click entry only; a Command Palette / keybinding entry can be a follow-up that calls the same `refresh_directory` helper.
- **Single-level rescan only** — recursive deep refresh is out of scope for v1.

## Files

- `app/src/code/file_tree/view.rs` — +42 lines, 0 deletions

## Verification

- `cargo check -p warp --bin warp-oss` clean
- `cargo clippy -p warp --bin warp-oss --no-deps` no new warnings
- Manually verified on a self-built `warp-oss` bundle: right-click any directory header → `Refresh` appears after `Open in new tab`; selecting it reconciles a drifted subtree

Closes #10205

### DIFF
--- a/app/src/code/file_tree/view.rs
+++ b/app/src/code/file_tree/view.rs
@@ -136,6 +136,9 @@ pub enum FileTreeAction {
     CDToDirectory {
         id: FileTreeIdentifier,
     },
+    RefreshDirectory {
+        id: FileTreeIdentifier,
+    },
     DismissEditor,
     ItemDroppedOnInput {
         id: FileTreeIdentifier,
@@ -2370,6 +2373,13 @@ impl FileTreeView {
                             .with_on_select_action(FileTreeAction::OpenInNewTab { id: id.clone() })
                             .into_item(),
                     );
+                    items.push(
+                        MenuItemFields::new("Refresh")
+                            .with_on_select_action(FileTreeAction::RefreshDirectory {
+                                id: id.clone(),
+                            })
+                            .into_item(),
+                    );
                 }
             };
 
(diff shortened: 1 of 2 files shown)


spalagu:
Addressed concern #1 (RepoNotFound in cwd-inside-repo case): `refresh_directory` now resolves the indexed repository root via the existing `RepoMetadataModel::find_repository_for_path` before calling `load_directory`, so the displayed file-tree root being a subdirectory of an indexed repo no longer fails the call. Falls back to the displayed root for standalone lazy-loaded paths where the two coincide.

Force-pushed a squashed single commit (`397a5a80`) capturing the original menu wiring + the indexed-root resolution fix.

Concern #2 (visual evidence): I'll attach a screenshot/recording in a follow-up comment shortly.

`/oz-review`

spalagu:
Addressed the stale-entry concern. Added one line to `FileTreeState::load_at_path` (`crates/repo_metadata/src/file_tree_store/file_tree_state.rs`):

```rust
entry.load(gitignores)?;
self.remove(path);   // ← NEW: clear stale entries before re-inserting
self.insert_entry_at_path(child_path, entry);
```

The bot is correct that `insert_entry_at_path` only `extend`s the maps without removing prior entries — a refresh of a directory whose contents shrank on disk would leave deleted files visible. `FileTreeState::remove` (line 83) recursively removes the path's entries plus all descendants, which is exactly what we need before laying down the freshly-built subtree.

Verified safe for the three existing callers of `load_at_path` (`view.rs:1375` / `view.rs:1413` / `server_model.rs:1167`) — they all hit the path on first-time-expand when the prior entry is just an unloaded placeholder (`loaded: false`, no descendants in the maps), so the new `remove` is a no-op for them.

Force-pushed a single squashed commit (`1042ccfd`).

Visual evidence still pending — will follow up.

`/oz-review`

PR #10560, opened 2026-05-09
Support undercurl SGR styling

### LINKED ISSUE #5633: Support for undercurl
### Dupe Check

- [X] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates

Add support for undercurl so it can have different colors as texts.

References:
https://github.com/wez/wezterm/issues/415
https://github.com/microsoft/terminal/issues/7228

Most modern terminals (Wezterm, Kitty, Ghostty) support this. It should be implemented in Warp as well. It's especially important for people writing code with lsp in terminal.

### Additional context

_No response_

4

None

jbetala7:
Fixes #5633.

## Summary
- Parse SGR underline style subparameters for single, double, and curly underline, plus underline color (58) and reset (59).
- Store underline color in terminal cells and flat scrollback style data.
- Render curly underline decorations and use the SGR underline color when present.
- Preserve undercurl/underline color in grid string serialization and underline ref-test data.

## Verification
- `cargo check -p warp --lib`
- `cargo test -p warp_terminal test_contains_cell_decorations`
- `cargo test -p warp_terminal test_push_rows_with_color`

## Visual proof
- Final screen recording: https://raw.githubusercontent.com/jbetala7/warp/proof-10560-undercurl-final/warp-undercurl-screen-recording.mov
- Captured from the locally built PR branch (`./target/debug/warp-oss`, commit `38e2f1a3`) in WarpOss. The video shows normal terminal output and alternate-screen output with curly underline plus explicit underline color.

## Notes
- Attempted `cargo test -p warp test_parsing_undercurl_and_underline_color`, but the local machine ran out of disk while compiling unrelated app test dependencies (lsp, ai, tantivy, warp_graphql) before the test could run.


jbetala7:
@cla-bot check

jbetala7:
Addressed the requested undercurl positioning change in `38e2f1a3` by using `cell_size.y()` as the curly underline baseline, so the shared decoration origin applies the decoration-height offset once.

Verification run on the current head:

- `cargo fmt --check`
- `git diff --check`
- `cargo check -p warp --lib`

I do not have a reliable app-level visual screenshot from this local environment. The broader app-level verification was previously blocked locally by resource/environment constraints while compiling unrelated app test dependencies, so I am noting that limitation here rather than attaching a misleading screenshot. The renderer-level positioning concern has been corrected in code and the current head now passes the checks above.

/oz-review

crates/warp_terminal/src/model/grid/cell.rs
```diff
@@ -115,8 +116,12 @@ struct CellExtra {
     /// Zerowidth characters stored in this cell WITH the base character at the start. This helps
     /// optimize reads on this data structure (we don't need to allocate a new string to join the
     /// base character and zerowidth characters).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
lucieleblanc, line 119: Why did these `serde` annotations change? This is the kind of change that would be great to explain with an inline GitHub comment.

lucieleblanc (changes requested):
Hi @jbetala7, thanks for taking on this issue. This is a big user-facing change so it needs a screen capture. Please attach a screen recording to the PR description in which the curly underline and underline color are shown working correctly, both in standard terminal blocks and in alt screen programs (full-pane terminal interfaces, like Neovim or another text editor with LSP). On MacOS, you can start a screen recording using `cmd`-`shift`-`5`.

Please remove the linter steps from the PR description. They are presubmit checks, not proof of correctness.

Feel free to comment or re-request review when ready and we'll take another look!

crates/warp_terminal/src/model/grid/cell.rs
jbetala7: They changed so `CellExtra` stays sparse now that the struct has three independent optional fields. `underline_color` can exist without zero-width text or an end-of-prompt marker, and existing `CellExtra` values can exist without an underline color. Without `skip_serializing_if`, those cases would serialize unrelated `null` keys and churn the stored/ref-test JSON; without `default`, omitted optional keys in older or sparse serialized payloads would not be handled explicitly. No behavior change was intended for the existing fields.

jbetala7:
Uploaded a cleaner final screen recording from the locally built PR branch (`./target/debug/warp-oss`, commit `38e2f1a3`):

https://raw.githubusercontent.com/jbetala7/warp/proof-10560-undercurl-final/warp-undercurl-screen-recording.mov

I sampled the capture before posting it. It shows regular terminal output with curly underline plus underline color, then a full-screen/alternate-screen sequence using DEC 1049h/1049l with colored curly underline and underline color.

lucieleblanc:
> I sampled the capture before posting it

The point of the screen capture is to show an end-to-end flow (seeing the cursor, typing, etc.). but there is no difference between this "sampled" clip and a series of static screenshots. I would encourage you to run this manually and record the full interaction 🙂 it's fine to have an LLM generate a helper script or other material that you use for the demo, but ultimately the screen recording should show a human using the feature, with typing/clicking, opening and closing programs, etc.

I'll be happy to review this again with an actual recording! Otherwise this will be on the back burner.

PR #10584, opened 2026-05-10
fix(warpify): use target-OS path separator for rc file paths (#10474)

### LINKED ISSUE #10474: Auto-Warpify zsh setup command uses ~\.zshrc path and fails with 'no such user or named directory'
﻿
## Summary
Auto-Warpify generated zsh bootstrap command uses a Windows-style backslash in the rc-file path (`~\.zshrc`), causing zsh to fail.

## Problem
When copying/running the Auto-Warpify command for zsh, the command appends to `~\.zshrc` instead of `~/.zshrc`.
zsh interprets `~\.zshrc` as a user-home expansion target and errors out.

## Expected behavior
The generated command should use a POSIX path for zsh rc files (for example `~/.zshrc`) and run without path-expansion errors.

## Actual behavior
Running the generated command fails with:
`zsh: no such user or named directory: .zshrc`

1. In Warp, trigger the Auto-Warpify output for zsh.
2. Copy and run the generated command that ends with `>> ~\.zshrc`.
3. Observe zsh error: `no such user or named directory: .zshrc`.

## Artifacts
None attached

## Warp version
v0.2026.04.29.08.57.stable_02

## Operating system
Microsoft Windows 11 专业版 (10.0.26200, build 26200)

## Possible source references
- app/assets/bundled/bootstrap/bash_zsh_subshell_bootstrap_block_output.txt: bundled template for bash/zsh Auto-Warpify bootstrap output.
- app/src/terminal/warpify/success_block.rs: renders the Auto-Warpify success snippet shown to users.

maxmilian:
## Description

`ShellType::rc_file_paths` (`crates/warp_terminal/src/shell/mod.rs`) is rendered into the Auto-Warpify shell snippet that runs on the **target** (e.g. an SSH remote, or a subshell during local Auto-Warpify). It used `PathBuf::join`, whose separator follows the **host** OS that built Warp. On a Windows host targeting a Unix shell this produced `~\.zshrc`, which the remote zsh rejected with `no such user or named directory`.

This change builds the path string explicitly with a separator chosen from the `TargetOS` argument rather than relying on `PathBuf::join`'s host-dependent behavior. The internal forward slashes inside the literal relative paths (e.g. `.config/fish/config.fish`) are preserved as-is — only the leading separator that joins `~` / `$HOME` to the relative path now follows the target OS.

## Linked Issue

Fixes #10474.

## Screenshots / Videos

n/a — the rendered string is covered by a new unit test.

## Testing

- New `test_rc_file_paths_use_target_os_separator` in `crates/warp_terminal/src/shell/mod_tests.rs` locks in the expected output for every (shell, target OS) combination — `~/.zshrc`, `~/.bashrc`, `~/.config/fish/config.fish` and the two PowerShell paths on `TargetOS::Linux`/`MacOS`; `\$HOME\.config/powershell/...` on `TargetOS::Windows`; and the deliberately-empty Unix-shell vectors on `TargetOS::Windows`.
- Full `cargo test -p warp_terminal --lib shell::` passes (16/16).

## Agent Mode

### DIFF
--- a/crates/warp_terminal/src/shell/mod.rs
+++ b/crates/warp_terminal/src/shell/mod.rs
@@ -333,31 +333,39 @@ impl ShellType {
     }
 
     /// Returns the potential paths to the RC file relative to the `home` directory.
+    ///
+    /// The path separator is chosen based on the target OS rather than the host OS,
+    /// because the resulting path is rendered into a shell command executed on the
+    /// target (e.g. via SSH or Auto-Warpify). Using `PathBuf::join` here would pick
+    /// the host's separator and produce strings like `~\.zshrc` on a Windows host
+    /// when targeting a Unix shell, which the remote shell cannot resolve.
     pub fn rc_file_paths(&self, os: TargetOS) -> Vec<PathBuf> {
-        let home_dir = Path::new(match os {
+        let home_dir = match os {
             TargetOS::Windows => "$HOME",
             _ => "~",
-        });
-        let relative_paths = match (self, os) {
+        };
+        let separator = match os {
+            TargetOS::Windows => '\\',
+            _ => '/',
+        };
+        let relative_paths: Vec<&str> = match (self, os) {
             (ShellType::PowerShell, TargetOS::Windows) => {
-                vec![Path::new(
-                    ".config/powershell/Microsoft.PowerShell_profile.ps1",
-                )]
+                vec![".config/powershell/Microsoft.PowerShell_profile.ps1"]
             }
             // We need to make sure this works for either editor of PowerShell (PowerShell Core or
             // Windows PowerShell) so just write the file to both.
             (ShellType::PowerShell, _) => vec![
-                Path::new("Documents/PowerShell/Microsoft.PowerShell_profile.ps1"),
-                Path::new("Documents/WindowsPowerShell/Microsoft.PowerShell_profile.ps1"),
+                "Documents/PowerShell/Microsoft.PowerShell_profile.ps1",
+                "Documents/WindowsPowerShell/Microsoft.PowerShell_profile.ps1",
             ],
             (_, TargetOS::Windows) => vec![],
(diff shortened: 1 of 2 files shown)


zachbai:
@MaggieShan mind taking a look here?

crates/warp_terminal/src/shell/mod.rs
```diff
@@ -333,31 +333,39 @@ impl ShellType {
...
+    /// when targeting a Unix shell, which the remote shell cannot resolve.
     pub fn rc_file_paths(&self, os: TargetOS) -> Vec<PathBuf> {
-        let home_dir = Path::new(match os {
+        let home_dir = match os {
             TargetOS::Windows => "$HOME",
             _ => "~",
-        });
-        let relative_paths = match (self, os) {
+        };
+        let separator = match os {
```
MaggieShan: Can we update this fn to return a `TypedPathBuf` instead of a `PathBuf`? This should help enforce using the right target's separator

MaggieShan (commented):
One small comment but this looks good - thanks for fixing!

crates/warp_terminal/src/shell/mod.rs
maxmilian: Done in `e982a21c` — `rc_file_paths` now returns `Vec<TypedPathBuf>` encoded for the target OS. The manual `'/'` vs `'\\'` separator match is gone; `TypedPathBuf::from_unix`/`from_windows` plus `push` derive the separator from the encoding, so it's enforced by the type rather than hand-written code.

Verified `cargo check`/`clippy`/`test` on `warp_terminal` and `cargo check` on the `warp` app crate — the caller just does `.to_str()`, which is signature-identical to `PathBuf`, so no caller change was needed. Output is byte-identical to before (interior `/` in the PowerShell paths is preserved), and the #10474 regression test still passes. Thanks for the review!

MaggieShan (approved):
Nice!

PR #10594, opened 2026-05-10
fix(macos): use CStr::from_ptr for NSString to Rust string conversion

### LINKED ISSUE #10587: Chinese characters render as garbled text in input panel and session titles on macOS
## Summary
Chinese text is rendered as garbled characters while composing in Warp input surfaces (including the feedback input panel), even though submitted command text is correct.

## Problem
On macOS, typing Chinese in Warp's input panel shows corrupted characters before submission. Similar corruption appears in auto-generated session titles. The user reports this worked correctly a few days ago, suggesting a regression.

## Expected behavior
Chinese characters should render correctly while typing in all input surfaces, including the feedback/input panel and session titles.

## Actual behavior
While composing text, Chinese characters display as corrupted or garbled text. After pressing Enter, the command executes with correct Chinese text, but session title text remains garbled.

1. Open Warp on macOS and switch to a Chinese IME.
2. Type Chinese text in an input surface (feedback/input panel).
3. Observe that pre-submit text appears corrupted.
4. Press Enter; submitted command text appears correct, but the related session title text can remain corrupted.

## Artifacts
**Screenshot 1:** Warp terminal and input panel showing Chinese command content is correct after Enter, while the input panel text appears corrupted during composition.

Image

**Screenshot 2:** Warp conversation/session list showing a session title rendered with garbled Chinese characters.

Image

## Warp version
v0.2026.05.06.15.42.stable_03

## Operating system
macOS 26.4.1 (build 25E253)

Lemonawa:
## Description
Fix garbled non-ASCII text (Chinese, Japanese, emoji, etc.) in macOS input panels and session titles.

Root cause: the `to_string()` function in `crates/warpui/src/platform/mac/window.rs` used `-[NSString length]` (UTF-16 code unit count) as the byte length for `UTF8String` (UTF-8 bytes). For CJK characters (1 UTF-16 unit = 3 UTF-8 bytes), the string was sliced mid-character, producing garbled text.

Fix: use `CStr::from_ptr()` which correctly reads the null-terminated UTF-8 C string.

Same bug pattern fixed in `app.rs` (2 file-path conversions) and `clipboard.rs` (1 clipboard path conversion).

## Linked Issue
Fixes #10587
- [x] The linked issue is a triaged bug (implicitly `ready-to-implement` per CONTRIBUTING.md).

## Testing
- [x] I have manually tested my changes locally with `./script/run` — Chinese text input now renders correctly.
- [x] Added 5 regression tests (`to_string_ascii`, `to_string_chinese`, `to_string_japanese`, `to_string_emoji`, `to_string_mixed_cjk`) that verify `to_string()` produces correct output for non-ASCII text.
- [x] `./script/presubmit` passes: fmt, clippy, clang-format, wgslfmt, 6206/6212 tests pass (6 failures are pre-existing SSH integration tests).

## Agent Mode

CHANGELOG-BUG-FIX: Fixed Chinese/Japanese/non-ASCII text rendering as garbled characters in input panel and session titles on macOS

https://github.com/user-attachments/assets/76bba7bd-da81-4de3-9f50-02f1abb330ed


Lemonawa:
@cla-bot check

crates/warpui/src/platform/mac/window.rs
```diff
@@ -1681,8 +1680,23 @@ fn transform_origin_from_rect_coord_to_frame_coord(origin: Vector2F, size: Vecto
...
+/// Converts an Objective-C `Object` (NSString) into a Rust `String`.
+///
+/// Uses `CStr::from_ptr` to read the null-terminated UTF-8 output of
+/// `-[NSString UTF8String]`, because `-[NSString length]` returns the
+/// number of UTF-16 code units, not the number of UTF-8 bytes.
+///
+/// Returns an empty string if `UTF8String` returns null, which may occur
+/// when the receiver is `nil`.
 unsafe fn to_string(value: *mut Object) -> String {
-    let slice = slice::from_raw_parts(value.UTF8String() as *const c_uchar, value.len());
```
vorporeal, line 1686: what i'm confused about here is that, if i take a look at the implementation of the `len()` function invoked here, it appears to be getting the number of bytes in the UTF-8-encoded string: https://docs.rs/cocoa-foundation/latest/src/cocoa_foundation/foundation.rs.html#667

what am i missing?

crates/warpui/src/platform/mac/window_tests.rs
vorporeal, line 1: did you confirm that these tests failed before you changed the implementation of `to_string()`?

Lemonawa:
Hey @vorporeal, good catch on the `len()` implementation — you're right, `cocoa-foundation 0.2.1`'s `len()` calls `lengthOfBytesUsingEncoding: NSUTF8StringEncoding`, so it returns UTF-8 byte count, not UTF-16 code units. My root cause analysis in the PR description was incorrect.

Also to your second question — no, I did not confirm the tests failed before changing `to_string()`. I wrote the tests after the fix, which was a mistake.

I've just tested again on the latest stable build (`v0.2026.05.06.15.42.stable_04` or maybe the one just before it) and Chinese text now renders correctly. It appears the original issue has already been fixed in a recent update.

Given that the root cause was wrong and the issue is already resolved, I think the right call is to close this PR. Sorry for the noise and thanks for the thorough review!

PR #10614, opened 2026-05-10
Add one-time migration to add handoff chip to custom toolbar layouts

harryalbert:
## Description

When we release local -> cloud handoff, we'll want to make it so that the handoff chip is added to folks' toolbars for visibility into the new feature. For people who just use the default chip  set, this will happen automatically. However, for people with custom toolbar arrangements, the chip won't be added by default because it's not part of their custom set.

To fix this, we can add a one-time migration that functions similarly to the code we have around showing launch modals. On-app-start, we check if we have already tried to add the chip to the user's toolbar arrangement, and if we have we immediately return. If we haven't, we first update the persistent setting to record that we now have done the update (so in the future we early return), and we then add it to their toolbar arrangement (also returning early if they just have the default arrangement, because there's nothing to do in this case). We also gate this entire fn on the handoff feature flag being enabled so that this process doesn't run until the feature is released.

The end-result here is that, when we release the handoff feature, the cloud chip will be added to everyone's toolbelts _once_ (if they remove it it will stay removed).

Implements APP-4424

## Demo

https://www.loom.com/share/ca4b8a8289b8444e9d187bc0a307861a

## Testing

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/d0938e82-581e-45a9-a7ed-bc41bfdba1b0_
_Run: https://oz.staging.warp.dev/runs/019e1357-8372-772d-824d-35285a5466b6_

### DIFF
--- a/app/src/ai/blocklist/handoff/mod.rs
+++ b/app/src/ai/blocklist/handoff/mod.rs
@@ -14,9 +14,15 @@
 //! `AmbientAgentViewModel::submit_handoff`, which reads the cached
 //! `forked_conversation_id` and `snapshot_upload` off `PendingHandoff`.
 
-use super::PendingAttachment;
+use crate::features::FeatureFlag;
 use crate::server::server_api::ai::AttachmentInput;
 
+use super::PendingAttachment;
+
+#[cfg(feature = "local_fs")]
+pub(crate) mod touched_repos;
+
+#[cfg_attr(target_family = "wasm", allow(dead_code))]
 #[derive(Debug, Clone, Default)]
 pub struct HandoffLaunchAttachments {
     pub(crate) request_attachments: Vec<AttachmentInput>,
@@ -26,11 +32,15 @@ pub struct HandoffLaunchAttachments {
 /// Carries the auto-submit payload for `& query` and `/handoff query`.
 /// `request_attachments` feed the spawn request while `display_attachments`
 /// are restored into the source input on failure.
+#[cfg_attr(target_family = "wasm", allow(dead_code))]
 #[derive(Debug, Clone)]
 pub struct PendingCloudLaunch {
     pub(crate) prompt: String,
     pub(crate) attachments: HandoffLaunchAttachments,
 }
 
-#[cfg(feature = "local_fs")]
-pub(crate) mod touched_repos;
+pub(crate) fn is_local_to_cloud_handoff_available() -> bool {
+    FeatureFlag::OzHandoff.is_enabled()
+        && FeatureFlag::HandoffLocalCloud.is_enabled()
+        && cfg!(all(feature = "local_fs", not(target_family = "wasm")))
+}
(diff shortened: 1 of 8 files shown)


harryalbert:
* **#10614** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/blocklist/mod.rs
```diff
@@ -5,15 +5,8 @@ pub mod block;
 pub mod code_block;
 mod context_model;
 mod controller;
-#[cfg(feature = "local_fs")]
 pub(crate) mod handoff;
 
-pub(crate) fn is_local_to_cloud_handoff_available() -> bool {
```
harryalbert, line 11: this is a drive-by refactor, but I noticed this fn was in a weird place

app/src/ai/blocklist/handoff/mod.rs
```diff
@@ -26,11 +32,15 @@ pub struct HandoffLaunchAttachments {
...
+#[cfg_attr(target_family = "wasm", allow(dead_code))]
 #[derive(Debug, Clone)]
 pub struct PendingCloudLaunch {
     pub(crate) prompt: String,
     pub(crate) attachments: HandoffLaunchAttachments,
 }
 
-#[cfg(feature = "local_fs")]
-pub(crate) mod touched_repos;
+pub(crate) fn is_local_to_cloud_handoff_available() -> bool {
```
zachbai: we might just wanna inline this because at some point soon the feature flags will get graduated and what will remain is just the cfg! check

zachbai (approved):

app/src/terminal/session_settings.rs
```diff
@@ -406,6 +406,17 @@ define_settings_group!(SessionSettings, settings: [
...
     },
+    // One-time flag: whether we've already migrated the handoff-to-cloud chip
+    // into a user's custom agent toolbar layout. When `Default`, the chip is
+    // already present via `AgentToolbarItemKind::default_right()`, so this
+    // only matters for `Custom` layouts that were saved before the chip existed.
+    did_add_handoff_chip_to_toolbar: DidAddHandoffChipToToolbar {
+        type: bool,
+        default: false,
+        supported_platforms: SupportedPlatforms::ALL,
+        sync_to_cloud: SyncToCloud::Globally(RespectUserSyncSetting::No),
```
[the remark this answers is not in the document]
harryalbert: This is fine

app/src/ai/blocklist/handoff/mod.rs
harryalbert: makes sense

PR #10615, opened 2026-05-10
Fix bash HISTSIZE startup sentinel

### LINKED ISSUE #10565: bash: in-memory history truncated to ~500 entries despite HISTSIZE=20000 (regression of #1262)
### Dupe check

Searched issues — closest is #1262 ("Truncating bash history to 500 items", closed as fixed).
The on-disk truncation that #1262 fixed is no longer happening (file is preserved correctly via `WARP_INITIAL_HISTFILESIZE`). However, the **in-memory** history list is still capped at ~500 entries, even though `HISTSIZE=20000` is set in `~/.bashrc`.

### Describe the bug

In a Warp bash tab, `history | wc -l` returns ~500 even though:
- `~/.bash_history` contains 30 000 lines
- `~/.bashrc` sets `HISTSIZE=20000` and `HISTFILESIZE=30000`
- `echo $HISTSIZE` in the same shell returns `20000`

### Comparison with another terminal (same machine, same `~/.bashrc`)

| Terminal | `echo $HISTSIZE` | `history \| wc -l` | `$TERM_PROGRAM` |
|----------|------------------|--------------------|------------------|
| **Warp** v0.2026.04.15.08.45.stable_02 | `20000` | **`500`** :x: | `WarpTerminal` |
| **GNOME Terminal** 3.52.0 (VTE 0.76.0) | `20000` | `20000` :white_check_mark: | *(empty)* |

Identical user, identical `~/.bashrc`, identical `~/.bash_history` — only the terminal emulator differs.

### To reproduce

1. In `~/.bashrc`, set `HISTSIZE=20000` and `HISTFILESIZE=30000`.
2. Have a `~/.bash_history` with more than 500 lines.
3. Open a new Warp tab.
4. Run:
   ```bash
   echo "HISTSIZE=$HISTSIZE"      # -> HISTSIZE=20000
   history | wc -l                # -> ~500  (expected: 20000)
   wc -l ~/.bash_history          # -> 30000+
   history -r ; history | wc -l   # -> 20000 (force reload works)
   ```

### Root cause hypothesis

The bootstrap script (`bundled/bootstrap/bash_body.sh`) sets a sentinel for `HISTFILESIZE` to protect the on-disk file from truncation, but does **not** apply the same sentinel mechanism for `HISTSIZE`. When bash reaches its first prompt and loads `~/.bash_history`, the in-memory list is capped at the default 500 — even though the rcfile has by then set `HISTSIZE=20000`.

The comment in the bootstrap script explicitly mentions HISTSIZE:
> "We need to set the initial value of HISTSIZE to ensure that the user's history file doesn't get truncated"

…but the actual code only handles `HISTFILESIZE`.

### Workaround

Append to `~/.bashrc`:
```bash
if [ "$TERM_PROGRAM" = "WarpTerminal" ]; then
  history -c
  history -r
fi
```

### Operating system

Linux — Ubuntu 24.04.4 LTS

### Shell Version

GNU bash 5.2.21(1)-release

### Current Warp version

v0.2026.04.15.08.45.stable_02

### Regression

Yes — the 2023 fix for #1262 protected the on-disk file but not the in-memory list.

Yes. Confirmed: same user, same `~/.bashrc`, GNOME Terminal 3.52.0 loads all 20 000 entries while Warp caps at 500.

jbetala7:
## Summary
- set a large initial HISTSIZE sentinel alongside HISTFILESIZE for bash startup
- unset HISTSIZE after rcfile loading when the user did not override it
- cover host bash, non-bash, and docker sandbox command environments with regression tests

Fixes #10565

## Tests
- cargo fmt --check
- git diff --check
- cargo test -p warp history_size_sentinels

### DIFF
--- a/app/assets/bundled/bootstrap/bash_body.sh
+++ b/app/assets/bundled/bootstrap/bash_body.sh
@@ -1133,17 +1133,22 @@ esac
         rcfiles_end_time="$(LC_ALL="C"; echo $EPOCHREALTIME)"
     fi
 
-    # Unset HISTFILESIZE if the user rcfiles didn't change it away from our
-    # very large sentinel value.  We need to set the initial value of HISTSIZE
-    # to ensure that the user's history file doesn't get truncated when we spawn
-    # the shell, but once bootstrap has completes, we want the value to be what
-    # it would have been if we hadn't set an initial value.
+    # Unset HISTFILESIZE and HISTSIZE if the user rcfiles didn't change them
+    # away from our very large sentinel values.  We need to set the initial
+    # values to ensure that the user's history file and in-memory history list
+    # don't get truncated when we spawn the shell, but once bootstrap has
+    # completed, we want the values to be what they would have been if we hadn't
+    # set initial values.
     #
     # For more context, see: https://github.com/warpdotdev/Warp/issues/1262
     if [[ $HISTFILESIZE == $WARP_INITIAL_HISTFILESIZE ]]; then
         unset HISTFILESIZE
     fi
     unset WARP_INITIAL_HISTFILESIZE
+    if [[ $HISTSIZE == $WARP_INITIAL_HISTSIZE ]]; then
+        unset HISTSIZE
+    fi
+    unset WARP_INITIAL_HISTSIZE
 
     # Save the value of HISTCONTROL as it existed just after reading the user's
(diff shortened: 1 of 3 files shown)


zachbai (approved):
nice, thanks for the fix!

PR #10624, opened 2026-05-11
Show close button on narrow code-pane tabs (#10592)

### LINKED ISSUE #10592: Show close button in place of file icon on hover for narrow code-pane tabs
## Summary
When many files are open in the code pane, individual editor tabs become too narrow to interact with. Hovering a tab should reveal a close button (×) in place of the file icon, similar to Chrome's tab behavior, so users can still close a specific file even when the tab bar is crowded.

## Problem
Opening many files in the code pane causes each editor tab to shrink. Currently the tab layout always shows the file-type icon and only reveals the close button when there is enough horizontal space. Once tabs become narrow, it becomes difficult or impossible to close a specific file from the tab because the close affordance is not visible or easily clickable.

## Expected behavior
When hovering over a narrow code-pane tab, the file icon should be temporarily replaced by a close (×) button so the user can close that file without needing keyboard shortcuts or first activating the tab.

I believe that there should be different display methods for tabs occupying varying degrees of overall space:

1. When there is sufficient space, a single file tab has ample room to display "X" normally. (Current behavior)

Image

2. When the space becomes narrower and the tab space for a single file shrinks, the "X" should only be displayed for the currently opened file. For other file tabs, the "X" should not be displayed on hover. (About https://github.com/warpdotdev/warp/issues/3870)

Image

3. As the space continues to narrow, the "X" should be removed from the currently open text tab, and only displayed when hovering over it.（like chrome）

Image

Image

## Actual behavior
The file icon remains visible on hover, and the close button is either hidden or clipped when tabs are narrow, making it hard to close individual files from the tab bar.

1. Open the code pane in Warp (e.g., via Command-P or clicking a file in the project explorer).
2. Open several files so the tab bar contains 6–10+ tabs.
3. Resize the pane or window so the tabs become narrow.
4. Hover over an inactive tab and try to close it.
5. Observe that the file icon is still shown and the close button is not easily accessible.

Desired workflow:
- On hover over any code-pane tab, the file-type icon is replaced by a close button.
- Clicking the close button closes that specific file tab.
- This mirrors the behavior in Chrome and Safari browser tabs.

## Artifacts
None attached.

## Warp version
Unknown.

## Operating system
macOS.

## Possible source references
- `app/src/code/view.rs` (`render_tab_internal`): tab rendering logic that shows the file icon and conditionally shows the close button based on hover and active state.

wzc520pyfm:
## Description

Address [#10592](https://github.com/warpdotdev/warp/issues/10592): when many files are open and the code-pane tab strip shrinks each tab below its preferred width, the close button (`X`) becomes unreachable. This change keeps the close affordance reachable across the full tab-width range while preventing the button from being squeezed or overflowing into neighboring tabs.

### Approach

Introduce a width-driven `TabDisplayMode` (`Wide` / `Medium` / `Narrow` / `VeryNarrow`) computed from the tab strip's measured width divided by tab count, captured with a `SavePosition` on the strip so the calculation is stable (independent of any single tab's rendered width — avoids layout oscillation between modes). Each mode picks a different visibility rule for the close button:

| Mode | Per-tab width | Close button visible when… |
| --- | --- | --- |
| Wide | ≥ 160px | tab is active OR hovered (historical behavior) |
| Medium | 100 – 160px | tab is active (inactive tabs no longer reveal X on hover) |
| Narrow | 64 – 100px | tab is **active AND hovered** (prevents accidental closes when tabs are tiny) |
| VeryNarrow | < 64px | same as Narrow, **plus** the file icon is hidden while the overlay is shown so the user sees a clean X instead of an X stacked on top of the file icon |

The close button is rendered as a `Stack` positioned overlay (anchored to the tab's right edge with `ParentOffsetBounds::ParentByPosition`) rather than as a regular Flex child. This guarantees the 24×24 button is never squeezed *and* never overflows past the tab boundary, even when the tab is too narrow for icon + button + filename to fit side by side.

The tab's hover state OR's in `close_handle.is_hovered()` so the overlay does not flicker when the mouse is over the X itself — the outer tab's `Hoverable` would otherwise treat itself as "covered" by the higher-z overlay and drop hover, hiding the X, which would un-cover the cursor and re-fire hover (a flicker loop).

## Linked Issue

Closes #10592.

- [ ] The linked issue is labeled \`ready-to-implement\`.
- [x] Screenshots of each mode are included below.

## Testing

- [x] I have manually tested my changes locally with \`./script/run\` across all four modes (Wide, Medium, Narrow, VeryNarrow).
- 6 unit tests covering the threshold boundaries in \`tab_display_mode_tests\` (\`cargo test -p warp --lib tab_display_mode\`).
- \`cargo fmt --check\`: clean.
- \`cargo clippy -p warp --lib --tests -- -D warnings\`: clean.

### Screenshots / Videos

Snipaste_mul_tab_1
Snipaste_mul_tab_2
Snipaste_mul_tab_3
Snipaste_mul_tab_4
Snipaste_mul_tab_5
Snipaste_mul_tab_6

## Agent Mode

CHANGELOG-IMPROVEMENT: Code pane tabs now progressively hide the close button as space tightens, replacing the file icon with the close button on hover when tabs are very narrow (#10592).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/src/code/view.rs
+++ b/app/src/code/view.rs
@@ -53,8 +53,8 @@ use warpui::{
     id,
     keymap::EditableBinding,
     ui_components::{button::ButtonVariant, components::UiComponent},
-    AppContext, Element, Entity, ModelHandle, SingletonEntity, TypedActionView, View, ViewContext,
-    ViewHandle, WindowId,
+    AppContext, Element, Entity, EntityId, ModelHandle, SingletonEntity, TypedActionView, View,
+    ViewContext, ViewHandle, WindowId,
 };
 
 use crate::{
@@ -90,9 +90,16 @@ const TAB_INTERNAL_MARGIN: f32 = 4.;
 const TAB_HORIZONTAL_MARGIN: f32 = 8.;
 const TAB_PADDING: f32 = 2.;
 
-/// Position id used to track the rendered width of the editor tab strip so
-/// [`TabDisplayMode`] can be derived from the previous-frame layout.
-const TABS_ROW_POSITION_ID: &str = "code_view_tabs_row";
+/// Builds a position id used to track the rendered width of one editor tab
+/// strip so [`TabDisplayMode`] can be derived from the previous-frame layout.
+///
+/// The id is scoped by the owning pane's entity id so that split panes with
+/// multiple `CodeView`s in the same window don't overwrite each other's
+/// cached width (which would otherwise cause one pane to render with the
+/// other pane's display mode).
+fn tabs_row_position_id(pane_id: EntityId) -> String {
+    format!("code_view_tabs_row_{pane_id}")
+}
 
 /// Estimated per-tab width at which we keep the close button visible for every
 /// tab (the historical "wide" behavior). Above this we have enough room for an
@@ -142,6 +149,13 @@ impl TabDisplayMode {
     /// number of tabs currently in the strip. Falls back to [`Self::Wide`] when
     /// the width is unknown (e.g. on the very first frame) so the initial
     /// render keeps the existing behavior.
+    ///
+    /// The tab strip is laid out as N `Shrinkable(flex=1)` tabs *plus* a
+    /// trailing `Expanded(flex=1)` spacer that fills the gap between the last
+    /// tab and the right-hand buttons, so the Flex layout allocates one of
+    /// `tab_count + 1` equal slices to each tab. Dividing by just `tab_count`
+    /// over-estimates the per-tab width and would pick a wider mode than the
+    /// user actually sees (5 tabs is a 20% overestimate, 10 tabs is 10%).
     fn from_tabs_row_width(tabs_row_width: Option<f32>, tab_count: usize) -> Self {
         let Some(total) = tabs_row_width else {
             return Self::Wide;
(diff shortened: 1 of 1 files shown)


wzc520pyfm:
Addressed all 3 IMPORTANT review comments in 295fd3f:

1. **Scoped `SavePosition` id by pane** — `TABS_ROW_POSITION_ID` constant replaced with `tabs_row_position_id(pane_id)`, keyed by `self.pane_configuration.id()`. Split panes no longer overwrite each other's cached strip width.
2. **Close button now respects the clip rect** — switched from `Stack::add_positioned_overlay_child` to `Stack::add_positioned_child`. The button now paints in the same clipped layer as the rest of the tab strip; off-screen / partially-clipped tabs no longer leak X buttons into neighboring UI.
3. **Spacer accounted for in per-tab width** — `from_tabs_row_width` now divides by `tab_count + 1` to include the trailing `Expanded(flex=1)` spacer. Threshold boundary tests updated to the corrected formula, plus a regression test (`spacer_aware_split_at_wide_boundary`) pinning the 5-tab Wide-threshold case to Medium.

Verified locally with `./script/run` across Wide / Medium / Narrow / VeryNarrow modes and split-pane scenarios. All 7 unit tests in `tab_display_mode_tests` pass.

/oz-review

PR #10625, opened 2026-05-11
Focus code review panel on open

### LINKED ISSUE #10406: Code review find bar persists (visible but broken) after panel close/reopen or repo switch
## Summary
The find bar in the code review panel persists across panel close/reopen and repo switches, leaving it visible but non-functional.

## Problem
When the code review panel's find bar is open, closing and reopening the panel (or switching repositories via the repo switcher dropdown) leaves the find bar visible with its previous query text. However, advancing to the next/previous match no longer works because the underlying editor handles have changed while the stale find state remains.

## Expected behavior
Closing the code review panel or switching repositories should dismiss the find bar, clearing its query and any match highlights. Reopening the panel or completing a repo switch should start with the find bar closed.

## Actual behavior
1. The find bar remains visually open after panel close/reopen, with the previous query still populated.
2. Clicking next/previous match buttons does nothing — the find model's results reference editors from the previous session.
3. The same stale state occurs when switching repos via the repo switcher dropdown within the panel.

## Reproduction steps
1. Open the code review panel.
2. Open find (Cmd+F) and enter a search query — matches highlight as expected.
3. Close the code review panel.
4. Reopen the code review panel.
5. Observe: the find bar is still visible with the previous query, but clicking "next match" does not navigate.

Alternate path (repo switch):
1. Open the code review panel with find active and a query entered.
2. Use the repo switcher dropdown to change repositories.
3. Observe: the find bar remains open with the old query, but match navigation is broken.

## Artifacts
None attached.

## Warp version
0.2026.05.07.09.05.01 (WarpDev)

## Operating system
macOS

## Possible source references
- `app/src/code_review/code_review_view.rs`: `on_close()` (line ~873) does not call `close_find_bar()`, so find state survives panel close/reopen.
- `app/src/code_review/code_review_view.rs`: `update_current_repo()` (line ~780), called on repo switch, also does not dismiss the find bar.
- `app/src/code_review/code_review_view.rs`: `close_find_bar()` (line ~2395) clears the find model and editor decorations — this is the method that should be invoked in both paths.

EdenBendheim:
## Description
Focus the Code Review panel immediately after it is opened. Shortcuts like Command-F now open in the newly opened diff editor instead of the old one.

Also updates find bars to be dismissed when they are already focused and Command-F is pressed.

## Linked Issue
Related & Closes: #10406

## Testing
- [x] `cargo check -p warp --features local_fs`

No new integration test: the changes are focused on existing focus/find plumbing and are small enough to validate with manual shortcut testing.

### Screenshots / Videos
https://github.com/user-attachments/assets/fe66e6cf-94f2-4361-854c-c770bf16f10c

## Agent Mode


kevinyang372 (changes requested):
I don't think focusing by default on open is the right product behavior, especially since we have the auto-open code review behavior on agent mode changes. This could cause user's input to be stolen for example when they are typing

EdenBendheim:
Hi @kevinyang372, that makes sense, thanks for calling it out. I updated this so the code review panel only takes focus for explicit user actions and the passive auto open should leave focus where it was.

Sorry for taking some time on this.

kevinyang372 (changes requested):
@EdenBendheim I would advocate for putting this behind an opt-in settings. This is a big change to existing product behavior and I am not fully convinced this should be the default behavior

Issue #9522, opened 2026-04-30
Warp violates Alacritty's licenses


chrisduerr:
I've looked through some of Warp's terminal emulation code after it was made public and have found several locations where Alacritty's code was copied verbatim without attribution. To make matters worse, warp uses AGPL for (all?) its source code, making it look like this code is publicly available under the AGPL license (which is incorrect).

I don't care if people profit from Alacritty's code by using the terminal, or reusing some of the code to build their own things. That's why me (and the rest of the Alacritty contributiors) give it away for free. It's unfortunate to see that other people seem to enjoy taking advantage of that generosity, by completely ignoring the very minimal restrictions we put on reusing our code. Especially for-profit companies with the means to fund open source development should be held to a higher standard.

alokedesai:
Hey, thanks for filing. Nathan (@nixpulvis) reached out to us yesterday about this. We include an Alacritty license in our bundled binary because we pull in the vte crate, but I recognize we didn't include a license for the Alacritty code that we forked and this was purely an oversight on our part. 

@vorporeal sent a PR (#9513) earlier today, now merged, based on the discussion with Nathan yesterday. We are very thankful for all of Alacritty's work and for your and Nathan's help early on in at Warp.

chrisduerr:
While @nixpulvis is a valued Alacritty contributor, he does not speak for Alacritty in an official manner. I'm not sure what you've discussed, but I certainly would have appreciated reaching out to me or @kchibisov directly.

While the PR to mitigate this is a step in the right direction, the modules I've looked at so far do not have notices.
Just as an example:
 - This is an almost verbatim copy of VTE's ansi.rs module:
   https://github.com/warpdotdev/warp/blob/master/crates/warp_terminal/src/model/ansi/control_sequence_parameters.rs
 - This function (and several others) are also verbatim copies:
   https://github.com/warpdotdev/warp/blob/157f35868b3ea66ef883a5b620913067033197a4/app/src/terminal/grid_renderer/cell_type.rs#L160

> We are very thankful for all of Alacritty's work

Giving appropriate attribution is how this thanks should be expressed, which should have happened long before Warp was ever open sourced. At this point you've contributed net negative value back to Alacritty, making me deal with annoying issues like these.

alokedesai:
Thanks for spotting those, we'll fix those asap.

We are truly grateful for all of your work and in no way are we trying to minimize, or take credit for, the amount of effort you and the rest of the Alacritty maintainers have put into your product.

[...]

PR #10631, opened 2026-05-11
docs: attribute Alacritty VTE parser integration

jbetala7:
## Summary
- add an explicit Alacritty/vte attribution notice to the ANSI terminal processing module
- point the module at the existing bundled Alacritty Apache license file

## Validation
- git diff --check
- cargo test -p warp_editor table --lib

Refs #9522

### DIFF
--- a/app/src/terminal/model/ansi/mod.rs
+++ b/app/src/terminal/model/ansi/mod.rs
@@ -1,3 +1,6 @@
+// This module integrates with the vte crate (an Alacritty project) under the
+// Apache license; see: crates/warp_terminal/src/model/LICENSE-ALACRITTY.
+
 //! This module implements processing logic for pty output.
 //!
 //! The top-level export is [`Processor`], which is a lightweight wrapper


vorporeal:
did you find code in here that was adapted from alacritty code, or is it simply a matter of this using `vte`?  we already redistribute the `vte` license in our releases and our online documentation.

vorporeal (commented):
(marking this as reviewed to get it out of my inbox)

jbetala7:
Thanks for checking. I did not find a separate copied block beyond Warp using the `vte` integration path; this was a conservative attribution note because `vte` is an Alacritty project and this module is where Warp wires ANSI processing through that path.

If the existing redistributed `vte` license already satisfies the requirement, I am fine with this PR being closed as unnecessary rather than adding redundant source comments.

vorporeal:
hey, sorry i missed your reply here.  the attribution notices at the top of files are used to indicate code that was copied/adapted from alacritty sources; i believe that our existing THIRD_PARTY_LICENSES compilation (via `cargo about`) is sufficient for standard dependencies.

PR #10670, opened 2026-05-11
fix bug where the agent view was empty when joining shared sessions

harryalbert:
## Description

There was a bug where, when opening an in-progress cloud conversation from the shared session link, the conversation would sometimes be completely empty.

The root cause here is was a race condition in the event loop, triggered by a change from last week to optimistically create the shared conversation's task Id in `on_shared_init`. The race is as follows:
1. `AgentResponseEvent::Init` → `on_shared_init` creates an exchange with the optimistic root task ID → `AppendedExchange` event emitted (deferred)
2. `AgentResponseEvent::ClientActions` → `apply_client_actions` processes `CreateTask` → replaces the optimistic root task with a server-assigned task that has a different ID (`conversation.rs:2253-2280`)
3. Deferred events dispatch → `AppendedExchange` handler runs → `conversation.get_task(optimistic_id)` returns None → AI block is never inserted → pane is blank

## Testing

### Screenshots / Videos

https://www.loom.com/share/5ae0c49e08ef4d79989a58409e1ecae9

## Agent Mode

CHANGELOG-BUG-FIX: Fixed a bug where cloud conversations were sometimes empty on-join.

### DIFF
--- a/app/src/ai/agent/task_store.rs
+++ b/app/src/ai/agent/task_store.rs
@@ -27,6 +27,11 @@ pub struct TaskStore {
     root_task_id: TaskId,
     tasks: HashMap<TaskId, Task>,
     linearized_refs: Vec<ExchangeRef>,
+    /// If the root task was upgraded from an optimistic (client-generated) ID
+    /// to a server-assigned ID, stores `(old_id, new_id)` so that deferred
+    /// event handlers referencing the stale optimistic ID can still resolve
+    /// the task.
+    upgraded_task_id: Option<(TaskId, TaskId)>,
 }
 
 impl TaskStore {
@@ -36,6 +41,7 @@ impl TaskStore {
             tasks: HashMap::new(),
             linearized_refs: Vec::new(),
             root_task_id: root_task_id.clone(),
+            upgraded_task_id: None,
         };
         store.tasks.insert(root_task_id, root_task);
         store.rebuild_linearized_refs_index();
@@ -49,6 +55,7 @@ impl TaskStore {
             tasks,
             linearized_refs: Vec::new(),
             root_task_id,
+            upgraded_task_id: None,
         };
         store.rebuild_linearized_refs_index();
         store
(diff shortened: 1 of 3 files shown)


harryalbert:
* **#10670** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10670?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10670?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/agent/task_store.rs
```diff
@@ -27,6 +27,11 @@ pub struct TaskStore {
     root_task_id: TaskId,
     tasks: HashMap<TaskId, Task>,
     linearized_refs: Vec<ExchangeRef>,
+    /// If the root task was upgraded from an optimistic (client-generated) ID
+    /// to a server-assigned ID, stores `(old_id, new_id)` so that deferred
+    /// event handlers referencing the stale optimistic ID can still resolve
+    /// the task.
+    upgraded_task_id: Option<(TaskId, TaskId)>,
```
zachbai: This could just be the optimistically generated root task id, we double set this and `root_task_id` at construction, then only change root_task_id when we upgrade the task

zachbai (approved):

app/src/ai/agent/task_store.rs
liliwilson: Can we just store `optimistic_root_task_id: Option<TaskId>` instead of the tuple?

liliwilson (approved):

app/src/ai/agent/task_store.rs
```diff
@@ -59,7 +66,10 @@ impl TaskStore {
     }
 
     pub fn get(&self, task_id: &TaskId) -> Option<&Task> {
-        self.tasks.get(task_id)
+        self.tasks.get(task_id).or_else(|| {
```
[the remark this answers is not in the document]
harryalbert: I don't think this is a concern

PR #10671, opened 2026-05-11
Clean up nld flags

evelyn-with-warp:
Remove the NLD-related feature flags and fasttext classifier 
Now that the ONNX/BertTiny classifier is the production default:

## Linked Issue

## Testing

https://www.loom.com/share/2d0bc3c4834e43928356266039da1b40

## Agent Mode

## Reviewers
@vorporeal @szgupta


app/src/ai/blocklist/input_model.rs
```diff
@@ -712,7 +712,6 @@ impl BlocklistAIInputModel {
 
                     // If we have history entries (i.e., a live session), check for
                     // close matches to short-circuit as shell input.
-                    // TODO(vorporeal): decide if we still want to do this with NldImprovements.
```
szgupta, line 715: Are we tracking this somewhere? Linear?

app/src/terminal/input.rs
```diff
@@ -12514,7 +12514,7 @@ impl Input {
                 || self.is_cloud_mode_input_v2_composing(ctx))
         {
             // If we're submitting an AI query, we want to send telemetry for the input type.
-            if FeatureFlag::NldImprovements.is_enabled() {
+            {
```
szgupta: nit: remove unnecessary scoping

script/linux/bundle
```diff
@@ -195,7 +195,7 @@ fi
 if [[ "$ARTIFACT" == "cli" ]]; then
   FEATURES="$FEATURES,standalone"
 elif [[ "$ARTIFACT" == "app" ]]; then
-  FEATURES="$FEATURES,gui,nld_improvements"
```
szgupta, line 198: Per https://warpdev.slack.com/archives/C08QY6ZBK2A/p1778531062702359?thread_ts=1778529278.316949&cid=C08QY6ZBK2A, how are we ensuring that CLI builds don't bundle the model?

evelyn-with-warp: CLI build will bundle the model, as a price to pay for a cleaner state (with as minimum feature flags as possible); 
Here's the change https://github.com/warpdotdev/warp/pull/10671/changes/592c7b65618bb997cc2a9ee5e05a21a9a0e1d98b I was thinking of for long term: gate all app ml models together and let CLI builds ignore them; The benefit is we could edit the ml_model gate freely without changing bundle scripts, and meanwhile protected cli/oz builds from them

evelyn-with-warp: it looks like we are replacing nld_improvment flag with bundled_all_ml_models flag, hopefully this will have long term benefit for readability and maintenance.

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: This should be related from internal report: https://linear.app/warpdotdev/issue/QUALITY-577/improve-nld-on-dependency-for-shell-command-history
side note: looks like we are doing it today with nld improvement, but we might hope to do a better job in short-circuit with user command history cc @vorporeal

app/Cargo.toml
```diff
@@ -126,7 +126,7 @@ jemalloc_pprof = { version = "0.8.1", optional = true, features = [
 ] }
 lsp-types = "0.97.0"
 indexmap = { version = "2.0.2", features = ["serde"] }
-input_classifier = { workspace = true, features = ["onnx_candle"] }
+input_classifier.workspace = true
```
evelyn-with-warp, line 129: features will be captured in ml_model_input_classifier_onnx

script/linux/bundle
vorporeal: > CLI build will bundle the model, as a price to pay for a cleaner state

i don't think we're ok with paying this price.  100% of the time we would take slightly more complex code in the bundling side of things over bloating the release size by 17MB with no functional benefit.

evelyn-with-warp: @vorporeal i agree for this. So current solution is we use bundled_app_ml_model flag, and the thought here is going forward this flag could handle all ml models (not only for nld work) that cli/oz should be excluded from

app/Cargo.toml
```diff
@@ -659,6 +659,16 @@ grep_tool = []
...
 # for things that pull in external dependencies.
 gui = ["voice_input"]
+# Bundled ML models (ONNX / etc.).
+nld_onnx_classifier = ["input_classifier/onnx_candle"]
+
+nld_classifier_v1 = [
+    "nld_onnx_classifier",
+]
+nld_classifier_v2 = [
+    "nld_classifier_v1",
```
evelyn-with-warp: cc @szgupta  this will be modified in the following PR when we add the new classifier

crates/input_classifier/src/onnx/mod.rs
```diff
@@ -38,13 +38,13 @@ impl Model {
...
         match self {
-            Model::BertTiny => "bert_tiny.onnx",
+            Model::BertTinyV1 => "bert_tiny_v1.onnx",
         }
     }
 
     fn tokenizer_path(&self) -> &'static str {
         match self {
-            Model::BertTiny => "bert_tiny_tokenizer.json",
+            Model::BertTinyV1 => "bert_tiny_tokenizer.json",
```
szgupta, line 47: will both v1 and v2 use the same tokenizer?

szgupta (approved):

PR #10672, opened 2026-05-11
Include platform_cost in cloud agent credit totals

jefflloyd:
## Description

The conversation details panel often rendered no `Credits used` field at all, even when the Oz web app showed a non-zero total for the same run. Two compounding bugs were responsible:

1. `ConversationDetailsData::from_task` used the `?` operator inside an `and_then` closure, so if either `inference_cost` or `compute_cost` was `null` on `RequestUsage`, the entire `credits` value became `None` and the row was dropped from render. Oz web handled the same data with `?? 0` and still rendered a number.
2. The client's `RequestUsage` struct was missing the server's `platform_cost` field entirely, so even after the Oz-web `fix/oz-credits-include-platform-cost` change, the local panel would still under-count by whatever was sitting in the platform bucket.

### What changed
- Added `platform_cost: Option<f64>` (with `#[serde(default)]` for forward compat with older server responses) to the client's `RequestUsage` in `app/src/ai/ambient_agents/task.rs`.
- Updated `AmbientAgentTask::credits_used()` to sum `inference + compute + platform`, each defaulting to `0.0` when null. This is the single helper every cloud-agent surface (details panel, tombstone, conversation list entries) now goes through.
- Collapsed the `CreditsInfo` enum in `conversation_details_panel.rs` down to `credits: Option<f32>` and deleted the inference/compute split renderer (`render_credits_with_split` / `render_cost_sub_row` and the two info-tooltip mouse states). Per product direction we now want a single `Credits used: X.X` line everywhere; no split UI.
- Both code paths land on that single total:
  - Local conversations: `conversation.credits_spent()` (already sums platform credits at the controller).
  - Cloud agent runs: `task.credits_used()` (now also sums platform).

## Linked Issue

## Testing
- Updated `conversation_details_panel_tests::test_from_conversation_populates_local_conversation_fields` to assert `data.credits.is_some()` after dropping the `CreditsInfo` enum.
- Updated `conversation_ended_tombstone_view_tests` to populate `platform_cost` on the test `RequestUsage` and to expect `inference + compute + platform` in the formatted credits string (catches future regressions if a bucket gets dropped from the sum again).
- `cargo clippy -p warp --all-targets --tests --no-deps -- -D warnings` is clean.
- `cargo nextest run -p warp -E 'test(conversation_details_panel) or test(conversation_ended_tombstone_view)'` — 15 passed.

### Screenshots / Videos
N/A — purely additive to an existing row in the details panel; the row now appears (with the correct total) for runs that previously rendered no credits at all.

## Agent Mode

CHANGELOG-BUG-FIX: Cloud agent run details now include platform cost in the "Credits used" total (and no longer drop the field entirely when one cost bucket is null).


app/src/ai/conversation_details_panel.rs
```diff
@@ -188,13 +186,6 @@ impl From<&TaskPrincipalInfo> for PrincipalInfo {
     }
 }
 
-/// Credit usage information for a conversation or task.
-#[derive(Debug, Clone)]
-enum CreditsInfo {
-    LocalConversation(f32),
-    AmbientConversation { inference: f32, compute: f32 },
-}
```
jefflloyd, line 196: Instead of splitting everything out by cost bucket, we're just going to sum up to one overall number. The API always exists for more detailed breakdowns.

app/src/ai/ambient_agents/task.rs
```diff
@@ -521,6 +523,8 @@ pub struct TaskStatusMessage {
 pub struct RequestUsage {
     pub inference_cost: Option<f64>,
     pub compute_cost: Option<f64>,
+    #[serde(default)]
+    pub platform_cost: Option<f64>,
```
tylerlam-warp: nit: Do we need the serde default here? If the struct is already an option, if there is no platform cost, shouldn't it default to None?

tylerlam-warp (approved):

app/src/ai/ambient_agents/task.rs
jefflloyd: We don't need it but it doesn't harm anything I don't think.

jefflloyd: I'll get rid of it

PR #10681, opened 2026-05-11
Fix daemon message too big error

kevinyang372:
## Description
This PR includes two parts:
1. In our run command executor, we do not put a cap on the response sizing. This could cause us to return an output that exceeds the max allowed size for the proto stream
2. On the remote server, we were dropping the stream when we encounter a write error, even when the error is recoverable (e.g. message too big). Instead we should send an error response and allow the stream to continue.

### DIFF
--- a/app/src/remote_server/server_model.rs
+++ b/app/src/remote_server/server_model.rs
@@ -1022,16 +1022,35 @@ impl ServerModel {
             move |me, result, _ctx| {
                 let result_oneof = match result {
                     Ok(output) => {
+                        let mut stdout = output.stdout.clone();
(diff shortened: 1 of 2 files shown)


kevinyang372:
* **#10681** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10681?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

MaggieShan (approved):

PR #10685, opened 2026-05-11
[Remote codebase indexing] wire up daemon indexing / drop indexing

moirahuang:
## Description
Wire daemon side indexing per https://github.com/warpdotdev/warp/pull/9508. I'm gating the `LaunchMode::RemoteServerDaemon` bit of logic behind a feature flag so we don't think we have codebase indexing remotely while I finish out the implementation.

I also moved some status stuff from the PR above this in the stack down to here just to even out the number of lines per PR, even though it's not strictly relevant for this PR

## Testing
Ran this w the branch on top of this to see if indexing works when indexing requests are actually kicked off
Screenshot 2026-05-11 at 3 21 59 PM
image

## Agent Mode

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -460,14 +460,16 @@ impl LaunchMode {
         }
     }
 
-    /// Returns `true` if running in app mode or via `agent run` to permit codebase indexing.
+    /// Returns `true` if this process can build and sync codebase indices.
     fn supports_indexing(&self) -> bool {
         match self {
             LaunchMode::CommandLine { command, .. } => {
                 matches!(command, CliCommand::Agent(AgentCommand::Run { .. }))
             }
-            LaunchMode::App { .. } | LaunchMode::Test { .. } => true,
-            LaunchMode::RemoteServerProxy | LaunchMode::RemoteServerDaemon { .. } => false,
+            LaunchMode::App { .. }
+            | LaunchMode::Test { .. }
+            | LaunchMode::RemoteServerDaemon { .. } => true,
+            LaunchMode::RemoteServerProxy => false,
         }
     }
 
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -466,9 +466,10 @@ impl LaunchMode {
             LaunchMode::CommandLine { command, .. } => {
                 matches!(command, CliCommand::Agent(AgentCommand::Run { .. }))
             }
-            LaunchMode::App { .. }
-            | LaunchMode::Test { .. }
-            | LaunchMode::RemoteServerDaemon { .. } => true,
+            LaunchMode::RemoteServerDaemon { .. } => {
+                FeatureFlag::RemoteCodebaseIndexing.is_enabled()
+            }
+            LaunchMode::App { .. } | LaunchMode::Test { .. } => true,
             LaunchMode::RemoteServerProxy => false,
         }
     }
@@ -2582,6 +2583,8 @@ pub fn enabled_features() -> HashSet<FeatureFlag> {
         FeatureFlag::PredictAMQueries,
         #[cfg(feature = "full_source_code_embedding")]
         FeatureFlag::FullSourceCodeEmbedding,
+        #[cfg(feature = "remote_codebase_indexing")]
+        FeatureFlag::RemoteCodebaseIndexing,
         #[cfg(feature = "use_tantivy_search")]
         FeatureFlag::UseTantivySearch,
         #[cfg(feature = "grep_tool")]
--- a/app/src/remote_server/server_model.rs
+++ b/app/src/remote_server/server_model.rs
@@ -6,21 +6,22 @@ use std::collections::{HashMap, HashSet};
 use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
+use warp_core::SessionId;
 use warp_core::channel::ChannelState;
 use warp_core::safe_error;
-use warp_core::SessionId;
 use warp_util::standardized_path::StandardizedPath;
 use warpui::r#async::{Spawnable, SpawnableOutput, SpawnedFutureHandle};
 use warpui::platform::TerminationMode;
 use warpui::{Entity, ModelContext, ModelHandle, SingletonEntity};
 
+use super::codebase_index_status::{
+    codebase_index_status_to_proto, disabled_codebase_index_status,
+    not_enabled_codebase_index_status, queued_codebase_index_status,
+};
 use crate::code::global_buffer_model::{GlobalBufferModel, GlobalBufferModelEvent};
 use ::ai::index::full_source_code_embedding::manager::{
-    CodebaseIndexFinishedStatus, CodebaseIndexManager, CodebaseIndexManagerEvent,
-    CodebaseIndexStatus as LocalCodebaseIndexStatus,
+    CodebaseIndexManager, CodebaseIndexManagerEvent,
 };
-use ::ai::index::full_source_code_embedding::SyncProgress;
 use warp_files::{FileModel, FileModelEvent};
 use warp_util::content_version::ContentVersion;
 use warp_util::file::FileId;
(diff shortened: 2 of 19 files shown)


app/src/lib.rs
```diff
@@ -461,14 +461,17 @@ impl LaunchMode {
...
     }
 
-    /// Returns `true` if running in app mode or via `agent run` to permit codebase indexing.
+    /// Returns `true` if this process can build and sync codebase indices.
     fn supports_indexing(&self) -> bool {
         match self {
             LaunchMode::CommandLine { command, .. } => {
                 matches!(command, CliCommand::Agent(AgentCommand::Run { .. }))
             }
+            LaunchMode::RemoteServerDaemon { .. } => {
```
moirahuang, line 469: gating specifically this behind a feature flag so we don't think we can support indexing even as i build out the implementation

crates/remote_server/src/manager.rs
```diff
@@ -95,6 +95,35 @@ pub enum RemoteServerInitPhase {
...
+    DropCodebaseIndex,
+}
+
+#[derive(Clone, Copy, Debug)]
+enum RemoteCodebaseIndexMutation {
+    Index,
+    Drop,
+}
+
+impl RemoteCodebaseIndexMutation {
```
moirahuang, line 109: added this to make the signature of `mutate_codebase_index` simpler. otherwise i had something like this:
Image

app/src/remote_server/server_model.rs
```diff
@@ -672,15 +680,110 @@ impl ServerModel {
...
+                    code: ErrorCode::InvalidRequest.into(),
+                    message: error,
+                }));
+            }
+        };
+        log::info!(
+            "[Remote codebase indexing] Daemon handling IndexCodebase: request_id={request_id} conn_id={conn_id} repo_path={repo_path_for_log}"
+        );
+
+        CodebaseIndexManager::handle(ctx).update(ctx, |manager, ctx| {
```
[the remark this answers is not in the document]
moirahuang: coming in subsequent PR, going to return 1NotEnabled` for now

app/src/remote_server/server_model.rs
```diff
@@ -672,15 +680,110 @@ impl ServerModel {
     }
 
     fn codebase_index_statuses_snapshot(&self) -> CodebaseIndexStatusesSnapshot {
-        // PR1 has no canonical daemon-side codebase-indexing state yet, so
-        // the bootstrap snapshot is empty. Later PRs will populate this from
-        // the remote indexing manager rather than deriving status from
-        // navigation events.
         CodebaseIndexStatusesSnapshot {
             statuses: Vec::new(),
```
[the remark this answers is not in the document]
moirahuang: in this PR i'll no longer push an empty snapshot but in subsequent PR it'll have to implement the correct snapshotting

crates/remote_server/src/manager.rs
```diff
@@ -1107,6 +1110,161 @@ impl RemoteServerManager {
...
+        operation: RemoteServerOperation,
+        request: F,
+        ctx: &mut ModelContext<Self>,
+    ) where
+        F: FnOnce(Arc<RemoteServerClient>, String, String) -> Fut + Send + 'static,
+        Fut: Future<Output = Result<RemoteCodebaseIndexStatus, crate::client::ClientError>>
+            + Send
+            + 'static,
+    {
+        let Some((session_id, client, remote_identity_key)) =
```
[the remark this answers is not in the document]
moirahuang: good point, updating to get current auth identify first and then selects connected session where key matches

script/deploy_remote_server
```diff
@@ -109,7 +109,7 @@ case "$PROFILE_MODE" in
     ;;
 esac
 
-FEATURES="release_bundle,crash_reporting,standalone,agent_mode_debug"
+FEATURES="release_bundle,crash_reporting,standalone,agent_mode_debug,remote_codebase_indexing"
```
moirahuang, line 112: adding the flag to our scripts as well

app/src/remote_server/server_model.rs
```diff
@@ -657,28 +675,219 @@ impl ServerModel {
...
+        ctx: &mut ModelContext<Self>,
+    ) {
+        if !FeatureFlag::RemoteCodebaseIndexing.is_enabled() {
+            log::info!(
+                "[Remote codebase indexing] Daemon skipping bootstrap codebase index statuses snapshot because remote indexing is disabled: conn_id={conn_id}"
+            );
+            return;
+        }
+        let snapshot = self.codebase_index_statuses_snapshot(ctx);
+        if snapshot.statuses.is_empty() {
```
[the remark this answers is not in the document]
kevinyang372: This seems valid?

app/src/remote_server/server_model.rs
```diff
@@ -657,28 +675,219 @@ impl ServerModel {
...
+        }
+        if auth_token.is_empty() {
+            return HandlerOutcome::Sync(server_message::Message::Error(ErrorResponse {
+                code: ErrorCode::InvalidRequest.into(),
+                message: "Missing authentication credentials for remote codebase indexing"
+                    .to_string(),
+            }));
+        }
+
+        self.auth_state.set_remote_server_bearer_token(auth_token);
```
kevinyang372: Hmmm I thought the logic we want is we verify if the auth_token that is passed in this request matches what we have cached on the server before starting the request

app/src/remote_server/server_model.rs
```diff
@@ -657,28 +675,219 @@ impl ServerModel {
...
+
+        let repo_path = match canonicalize_repo_path(&repo_path) {
```
kevinyang372: We should be using StandardizedPath::from_local_canonicalized here

Maybe in a follow-up but we should migrate CodebaseIndexManager to use StandardizedPath as well

app/src/remote_server/server_model.rs
```diff
@@ -657,28 +675,219 @@ impl ServerModel {
...
+                }));
+            }
+        };
+        log::info!(
+            "[Remote codebase indexing] Daemon handling IndexCodebase: request_id={request_id} conn_id={conn_id} repo_path={repo_path_for_log}"
+        );
+
+        CodebaseIndexManager::handle(ctx).update(ctx, |manager, ctx| {
+            if manager.root_path_for_codebase(&repo_path).is_some() {
+                manager.try_manual_resync_codebase(&repo_path, ctx);
```
kevinyang372: Do we need a manual resync here if the codebase is already indexed / indexing?

app/src/remote_server/server_model.rs
```diff
@@ -657,28 +675,219 @@ impl ServerModel {
...
+            return HandlerOutcome::Sync(server_message::Message::Error(ErrorResponse {
+                code: ErrorCode::InvalidRequest.into(),
+                message: "Missing authentication credentials for remote codebase index removal"
+                    .to_string(),
+            }));
+        }
+
+        self.auth_state.set_remote_server_bearer_token(auth_token);
+
+        let repo_path = match canonicalize_repo_path(&repo_path) {
```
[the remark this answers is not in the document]
kevinyang372: This seems reasonable

app/src/remote_server/server_model.rs
```diff
@@ -1545,6 +1754,118 @@ impl ServerModel {
...
+fn canonicalize_repo_path(repo_path: &str) -> Result<PathBuf, String> {
+    if repo_path.is_empty() {
+        return Err("repo_path is required".to_string());
+    }
+
+    dunce::canonicalize(Path::new(repo_path))
+        .map_err(|error| format!("Invalid repo_path {repo_path}: {error}"))
+}
+
+fn current_epoch_millis() -> u64 {
```
kevinyang372: Could we move these to a separate file? server_model is getting pretty long...

kevinyang372 (approved):
Approving to unblock

app/src/remote_server/server_model.rs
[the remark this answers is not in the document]
moirahuang: will instead require request token to match cached token and reject mismatched tokens

app/src/remote_server/server_model.rs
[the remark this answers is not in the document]
moirahuang: updating to send even when empty

app/src/remote_server/server_model.rs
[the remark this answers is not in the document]
moirahuang: will implement

app/src/remote_server/server_model.rs
moirahuang: will implement this

app/src/remote_server/server_model.rs
moirahuang: tracking followup in https://linear.app/warpdotdev/issue/APP-4447/migrate-codebaseindexmanager-to-use-standardizedpath

app/src/remote_server/server_model.rs
moirahuang: will remove the manual resync

app/src/remote_server/server_model.rs
moirahuang: yup will do

PR #10691, opened 2026-05-12
Trim orchestration config block UI

cephalonaut:
## What
- Shorten the orchestration config block description to a single sentence
- Left-align the View details control (was right-aligned)

Screenshot 2026-05-11 at 8 28 32 PM

## Why
Feedback from Peter: the config block takes up too much space and the View details control is better positioned on the left.

## Agent Mode


advait-m (approved):
nice, yeah I agree it's nice to trim a bit here!

PR #10692, opened 2026-05-12
Fix orchestration picker theming in light themes

cephalonaut:
## What
Remove custom font_color and border_color overrides from picker_styles(), letting ButtonVariant::Secondary theme-aware defaults handle contrast in all themes.

Screenshot 2026-05-11 at 8 32 05 PM

Screenshot 2026-05-11 at 8 32 26 PM

## Why
The previous custom colors used blended_colors::text_main with a composited background, which produced washed-out text and borders in light themes. The dropdown's built-in Secondary styling uses theme.outline() and theme.main_text_color() which are contrast-aware and adapt correctly.

Affects both the plan config block and the run_agents confirmation card (shared picker_styles function).

## Agent Mode


advait-m (approved):
ah, that makes sense, nice

PR #10698, opened 2026-05-12
[Tab configs] make tab configs run commands sequentially

moirahuang:
## Description
It was flagged in our preview Slack that tab configs that run a worktree + a long running command like `opencode` will make the Warp UI such as project explorer and code review panel show the original directory, not the worktree directory: https://warpcommunity.slack.com/archives/C092WE7HDH9/p1778148593003159

This is because we execute all the `commands` in the tab config in one go with `&&` between commands so the UI doesn't refresh the current working directory. I considered 2 options: 1) overwriting `cd` 2) making the commands fire sequentially. I went with 2 since it felt less brittle but if users really dislike the UX, we can consider going back

Note: launch configs were also built to run commands in one go with `&&` appending the commands. I decided to not change that existing behavior since launch configs are not in active development anymore but I'm not entirely opposed to having consistent behavior between the 2

## Testing
Locally ran and confirmed that w sequentially run commands from a tab confg 

### Screenshots / Videos
https://github.com/user-attachments/assets/632126d5-d5d4-410e-8350-0e966b8177fc

## Agent Mode


app/src/launch_configs/launch_config.rs
```diff
@@ -89,6 +92,22 @@ pub enum PaneMode {
...
 }
 
+/// How a pane template's saved commands should be submitted.
+///
+/// Launch configs default to the legacy `&&` chain. Tab configs opt into
+/// separate blocks so shell hooks like `precmd` can report cwd changes between
+/// setup commands and long-running commands.
+#[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum CommandExecutionMode {
```
moirahuang: i'd considered updating launch configs to also use sequential blocks. it would clean up the code by not needing to support both code paths. but because we've said we're not actively developing on launch configs anymore, i figured it'd be better to preserve the original behavior. but i'm honestly not strongly opinionated here because i also see the benefit of not having 2 code paths

app/src/terminal/view.rs
```diff
@@ -2518,6 +2510,9 @@ pub struct TerminalView {
     /// next `AfterBlockCompleted`, at which point `Event::PendingCommandCompleted`
     /// is emitted so subscribers know the command has finished.
     awaiting_pending_command_completion: bool,
+    /// Commands that should run as separate blocks after the active pending
+    /// command finishes successfully.
+    pending_command_sequence: VecDeque<String>,
```
alokedesai: maybe `pending_command_queue`?

app/src/launch_configs/launch_config.rs
alokedesai: i'd just do it for both honestly, the new behavior is just UX better for both in my opinion and it simplifies the code

alokedesai (approved):

PR #10704, opened 2026-05-12
Third-party harness session sharing fixes.

liliwilson:
## Description

This PR addresses three session sharing issues for third-party harnesses (each landed in its own commit):
1. We increase the width of the default PTY size for a headless Warp instance so that for a terminal session that never has a viewer, the saved block transcript is a reasonable width and not super skinny. Tested by running locally and never opening the session before it closes; the width is much more reasonable now.

2. We don't ever reset to the sharer size when a viewer joins and exits an ambient session. It doesn't make sense to revert to the default headless value once we've gotten a real value from a viewer.

3. We dedupe the viewer checks by firebase ID. There was a bug where, if you join a session, leave it, and rejoin it, your resize events don't get respected on the second rejoin. This happened because the presence manager holds onto stale visitors for some time before dropping them, and it incorrectly decided we were ineligible for resizing because it saw "another person" in the session, even though it was a phantom you.
    - This does mean that if you are viewing an ambient agent session from two different locations simultaneously, we'll allow resizing from both. I think this is actually correct given that you can't really be resizing both at the same time if you are one person.

## Testing

Behavior before: https://www.loom.com/share/0443fc35324d4b0a9e207bbb8f0a9fa9
Behavior after: https://www.loom.com/share/6ff55284504e4d37b6c0b6d98198fe0f

New default width example (ran without ever opening from a viewer):
image
 

## Agent Mode

### DIFF
--- a/app/src/terminal/terminal_manager.rs
+++ b/app/src/terminal/terminal_manager.rs
@@ -54,9 +54,9 @@ pub(super) fn compute_block_size(initial_size: Vector2F, ctx: &mut AppContext) -
         TerminalSettings::as_ref(ctx).terminal_spacing(appearance.line_height_ratio(), ctx);
     let size_info = if ctx.is_headless() {
         // In headless mode, we don't actually have a font since we aren't rendering anything.
-        // We skip the font-based size computation and hardcode a standard 80x24 terminal, so that
+        // We skip the font-based size computation and hardcode a terminal size, so that
         // viewers of the shared session see a reasonable terminal width.
-        SizeInfo::new_without_font_metrics(24, 80)
+        SizeInfo::new_without_font_metrics(24, 120)
     } else {
         let font_cache = ctx.font_cache();
         create_size_info_for_blocklist(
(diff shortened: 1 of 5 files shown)


app/src/terminal/terminal_manager.rs
```diff
@@ -54,9 +54,9 @@ pub(super) fn compute_block_size(initial_size: Vector2F, ctx: &mut AppContext) -
         TerminalSettings::as_ref(ctx).terminal_spacing(appearance.line_height_ratio(), ctx);
     let size_info = if ctx.is_headless() {
         // In headless mode, we don't actually have a font since we aren't rendering anything.
-        // We skip the font-based size computation and hardcode a standard 80x24 terminal, so that
+        // We skip the font-based size computation and hardcode a 80x120 terminal, so that
         // viewers of the shared session see a reasonable terminal width.
-        SizeInfo::new_without_font_metrics(24, 80)
+        SizeInfo::new_without_font_metrics(40, 120)
```
liliwilson: This is somewhat arbitrary—the height doesn't really matter, given that PTY _height_ shouldn't really impact how CLI agents render, but the width is what matters. When testing, 120 was pretty reasonable—not too wide but doesn't look weird on a wider terminal view

harryalbert (approved):
nice

PR #10030, opened 2026-05-04
fix: close cli rich input with ctrl-g from editor

nihalxkumar:
## Description
Fixes #9286 - Ctrl-G now closes CLI agent rich input when dispatched from the focused editor context. Previously the keybinding only matched the terminal context, so rich input couldn't be closed with Ctrl-G when the editor was focused.

This issue was reproduced on Arch Linux based system. The issue wasn't reproducible on MacOS.

## Linked Issue

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [x] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Screenshots / Videos

Before:

https://github.com/user-attachments/assets/0fa76723-1a6d-4122-ae63-c639d1164b24

After:

[warp-ctrlg-fix.webm](https://github.com/user-attachments/assets/76cca669-17e2-43eb-847f-cedee89da09a)

## Testing

Added regression test `ctrl_g_closes_cli_agent_rich_input_when_editor_is_focused()` that verifies Ctrl-G closes rich input when dispatched from the focused editor context.

## Agent Mode
- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: Fixed Ctrl-G not closing CLI agent rich input on linux when editor is focused (fixes #9286)

Issue #9286, opened 2026-04-28
Can't exit Rich text prompt with Ctrl + g


nihalxkumar:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

You can enter the rich text prompt with the shortcut but can not exit it with the shortcut

### To reproduce

1. Start opencode session
2. Press `Ctrl`+`g`
3. When you enter the rich text prompt, press `Ctrl`+`g` again to exit it

### Screenshots, videos, and logs

https://github.com/user-attachments/assets/0fa76723-1a6d-4122-ae63-c639d1164b24

### Operating system (OS)

Linux

### Operating system and version

EndeavourOS(Arch based)

### Current Warp version

v0.2026.04.22.08.46.stable_03

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

advait-m:
Thanks for the report @nihalxkumar 

Wondering if you could help verify that the bug is fixed (and exists on `master`) in the two PRs linked above (branches in the forks), by running Warp locally? Just since I don't see a repro of this bug myself (Loom where I'm testing with OpenCode: https://www.loom.com/share/c0cf690290ca417a87e54ead1a8e5ef8) so I'm wondering what's going on here.

Or would be helpful to add a video here showing what you're seeing (ideally including keypresses if possible, but helpful either way)

Curious if you have an IME keyboard for another language open too? We disable the shortcut in that case, so wondering if that's relevant

nihalxkumar:
[...]

PR #10706, opened 2026-05-12
fix(terminal-input): toggle Rich input panel on Ctrl+G (#9916)

### LINKED ISSUE #9916: Rich input toggle shortcut (Ctrl+G) only works to open, not to hide
## Description

The Rich input feature has a toggle shortcut `Ctrl+G` that should work bidirectionally — open and hide. However, the shortcut only works to **open** Rich input. Once the Rich input panel is visible, pressing `Ctrl+G` again does **not** hide it. The only way to dismiss it is by manually clicking the "Hide Rich input" button.

## Steps to Reproduce

1. Open Warp terminal
2. Press `Ctrl+G` to open Rich input ✓ (works)
3. With Rich input visible, press `Ctrl+G` again ✗ (does not hide)
4. Manually click "Hide Rich input" button ✓ (works)

## Expected Behavior

`Ctrl+G` should be a toggle — pressing it when Rich input is open should hide the panel, just like pressing it when closed opens it.

## Actual Behavior

`Ctrl+G` only works in one direction (open). To hide Rich input, you must use the mouse to click the "Hide Rich input" button.

## Environment

- **OS:** Windows 10
- **Warp version:** latest
- **Feature:** Rich input

## Additional Context

This is a minor UX friction point. For keyboard-driven workflows, being able to toggle Rich input without reaching for the mouse would be more consistent with the toggle behavior users expect from a shortcut.

lonexreb:
Closes #9916

## Summary

Resolves #9916. Ctrl+G now toggles the CLI agent Rich input panel in both directions instead of only opening it.

## Root cause

The Ctrl+G binding in `app/src/terminal/view/init.rs` had two context-predicate branches:

1. `Terminal & (LongRunningCommand | AltScreen) & CLI_AGENT_FOOTER_ENABLED & CLI_AGENT_RICH_INPUT_CHIP_ENABLED` — the open path.
2. `EditorView & !IMEOpen & CLI_AGENT_RICH_INPUT_OPEN` — added in PR #10030 to close from the focused editor (#9286).

The `CLI_AGENT_RICH_INPUT_OPEN` flag was set only on the terminal input's `EditorView` keymap context (in `app/src/terminal/input.rs`). Predicates are evaluated per-view against that view's own context — they are not merged across the responder chain (see `Matcher::push_keystroke` in `crates/warpui_core/src/keymap/matcher.rs`).

That left a gap: when the CLI agent block transitions out of `LongRunningCommand` (e.g., the agent paused waiting for user input) AND focus is not on the embedded editor (e.g., the user clicked into the block list, or a child view of the terminal is focused), neither branch matched and the Ctrl+G keystroke was silently dropped. The "Hide Rich input" footer button kept working because it emits `UseAgentToolbarEvent::HideRichInput` directly without going through the keybinding.

## Fix

A two-part predicate-coverage fix that reuses the existing toggle action and close path (no behavior change in the action handler):

1. **`app/src/terminal/view.rs`** — mirror `flags::CLI_AGENT_RICH_INPUT_OPEN` onto `TerminalView`'s `keymap_context` whenever `CLIAgentSessionsModel::is_input_open(view_id)` is true. The flag was previously only set on the editor's context modifier.
2. **`app/src/terminal/view/init.rs`** — add a third branch to the binding predicate:

   ```
   | (id!("Terminal") & !id!("IMEOpen") & id!(flags::CLI_AGENT_RICH_INPUT_OPEN))
   ```

   With (1), this matches anywhere in the terminal subtree once rich input is open, independent of which descendant has focus and independent of `LongRunningCommand` / `AltScreen` / chip-visibility state.

The existing `TerminalAction::ToggleCLIAgentRichInput` handler at `app/src/terminal/view.rs:26164` already routes to `close_cli_agent_rich_input_and_disable_auto_toggle` when a session is active, so this is purely a predicate-coverage fix that reuses the same close path used by the "Hide Rich input" footer button.

## Test plan

- [x] Added unit test `ctrl_g_closes_cli_agent_rich_input_from_terminal_context` — dispatches Ctrl+G with only the terminal view in the responder chain and asserts the session closes. This case is exactly what previously failed.
- [x] Added unit test `ctrl_g_toggles_cli_agent_rich_input_from_terminal_context` — exercises open → close → open → close from the terminal context, verifying true toggle behavior.
- [x] Existing regression test `ctrl_g_closes_cli_agent_rich_input_when_editor_is_focused` (for #9286) still covers the EditorView path.
- [ ] Manual on Linux/macOS: start a CLI agent (Claude/OpenCode/Gemini), press Ctrl+G to open Rich input, click into the block list to move focus off the editor, press Ctrl+G — Rich input should hide.
- [ ] Manual on Windows: same as above (the issue reporter is on Windows 10).

## Notes

- Local `cargo check -p warp` requires the Xcode Metal toolchain (not just Command Line Tools) on macOS, so I could not run the test suite in this environment. The changes are deliberately small, idiomatic, and pattern-match the existing predicate / `keymap_context` style in the file. CI will exercise full compilation and the new tests.
- The new third branch deliberately overlaps with Case 2 (EditorView) and Case 1 (open-path) in some states; this is fine because all three dispatch the same `TerminalAction::ToggleCLIAgentRichInput`, which is idempotent w.r.t. session state and bails early if the rich-input session is already in the target state.
- No `Cargo.lock`, CI, or unrelated files were touched.

## Manual testing note

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -26792,6 +26792,15 @@ impl View for TerminalView {
                     context.set.insert(flags::CLI_AGENT_RICH_INPUT_CHIP_ENABLED);
                 }
             }
+
+            // Mirror the rich-input-open flag onto the terminal context so the
+            // Ctrl+G toggle binding can close rich input regardless of which
+            // descendant view currently holds focus, and even when the
+            // active block has transitioned out of `LongRunningCommand`
+            // (e.g., the CLI agent has paused waiting for user input). See #9916.
+            if CLIAgentSessionsModel::as_ref(app).is_input_open(self.view_id) {
+                context.set.insert(flags::CLI_AGENT_RICH_INPUT_OPEN);
+            }
         }
 
         if FeatureFlag::AgentView.is_enabled() {
--- a/app/src/terminal/view/init.rs
+++ b/app/src/terminal/view/init.rs
@@ -317,9 +317,11 @@ pub fn init(app: &mut AppContext) {
 
     app.register_editable_bindings([
         // Ctrl-G: toggle CLI agent rich input.
-        // Two contexts match this binding:
+        // Three contexts match this binding:
         // 1. Terminal context when CLI agent footer is visible (opens rich input)
         // 2. EditorView context when rich input is already open (closes rich input, fix for #9286)
+        // 3. Terminal context when rich input is open (closes rich input regardless
+        //    of focus location or active-block state; fix for #9916)
         EditableBinding::new(
             OPEN_CLI_AGENT_RICH_INPUT_KEYBINDING,
             "Toggle CLI Agent Rich Input",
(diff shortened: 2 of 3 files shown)


kevinchevalier (approved):
Thanks for the fix!

PR #10720, opened 2026-05-12
Add CLI agent IPC bridge

KRRT7:
Adds a per-terminal CLI agent IPC listener and exports WARP_CLI_AGENT_IPC so Warp no longer depends on /dev/tty for Claude Code hook notifications. The existing OSC 777 path stays intact as the fallback.

### DIFF
--- a/app/src/terminal/cli_agent_sessions/ipc.rs
+++ b/app/src/terminal/cli_agent_sessions/ipc.rs
@@ -1,0 +1,176 @@
+use std::{
+    fs,
+    io::Read,
+    os::unix::{fs::DirBuilderExt, net::UnixListener},
+    path::PathBuf,
+    sync::{
+        atomic::{AtomicBool, AtomicU64, Ordering},
+        Arc,
+    },
+    thread::{self, JoinHandle},
+    time::Duration,
+};
+
+use async_channel::Sender;
+
+use crate::terminal::{cli_agent_sessions::event::CLI_AGENT_NOTIFICATION_SENTINEL, event::Event};
+
+static NEXT_SOCKET_ID: AtomicU64 = AtomicU64::new(1);
+const MAX_IPC_MESSAGE_BYTES: u64 = 1024 * 1024;
+
+pub struct CLIAgentIpcListener {
+    socket_path: PathBuf,
+    shutdown: Arc<AtomicBool>,
+    join_handle: Option<JoinHandle<()>>,
+}
+
+impl CLIAgentIpcListener {
+    pub fn start(event_sender: Sender<Event>) -> std::io::Result<Self> {
+        let socket_path = socket_path();
+        if let Some(parent) = socket_path.parent() {
+            fs::DirBuilder::new()
+                .mode(0o700)
+                .recursive(true)
+                .create(parent)?;
+        }
+        let _ = fs::remove_file(&socket_path);
+
+        let listener = UnixListener::bind(&socket_path)?;
+        listener.set_nonblocking(true)?;
+
+        let shutdown = Arc::new(AtomicBool::new(false));
+        let thread_shutdown = shutdown.clone();
+        let thread_socket_path = socket_path.clone();
+        let join_handle = thread::Builder::new()
+            .name("cli-agent-ipc-listener".to_owned())
+            .spawn(move || {
+                run_listener(listener, event_sender, thread_shutdown);
+                let _ = fs::remove_file(thread_socket_path);
+            })?;
+
+        Ok(Self {
+            socket_path,
+            shutdown,
+            join_handle: Some(join_handle),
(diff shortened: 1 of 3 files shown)


app/src/terminal/cli_agent_sessions/ipc.rs
```diff
@@ -0,0 +1,176 @@
...
+        let _ = fs::remove_file(&self.socket_path);
+        if let Some(join_handle) = self.join_handle.take() {
+            if let Err(err) = join_handle.join() {
+                log::warn!("Failed to join CLI agent IPC listener thread: {err:?}");
+            }
+        }
+    }
+}
+
+fn run_listener(listener: UnixListener, event_sender: Sender<Event>, shutdown: Arc<AtomicBool>) {
```
harryalbert, line 75: Instead of polling every 100ms, could we follow the existing pattern from the remote server daemon (`remote_server/unix/mod.rs:78-104`)? That uses `async_io::Async::new(listener)` with `.accept().await`, meaning it has less latency/wasted cycles IIUC.

app/src/terminal/cli_agent_sessions/ipc.rs
```diff
@@ -0,0 +1,176 @@
...
+            join_handle: Some(join_handle),
+        })
+    }
+
+    pub fn socket_path(&self) -> &PathBuf {
+        &self.socket_path
+    }
+}
+
+impl Drop for CLIAgentIpcListener {
```
harryalbert, line 63: Drop calls `fs::remove_file` on line 66 before `join_handle.join()` on line 67, but the thread also calls `fs::remove_file` when it exits. The socket disappears while the listener's final loop iteration may still be running. I think it would be cleaner to let the thread be the sole owner of cleanup — just signal shutdown and join here, let the thread remove the file on exit.

app/src/terminal/cli_agent_sessions/ipc.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+}
+
+fn run_listener(listener: UnixListener, event_sender: Sender<Event>, shutdown: Arc<AtomicBool>) {
+    while !shutdown.load(Ordering::Relaxed) {
+        match listener.accept() {
+            Ok((mut stream, _)) => {
+                let mut bytes = Vec::new();
+                match stream
```
harryalbert, line 80: `stream.by_ref().take(MAX_IPC_MESSAGE_BYTES).read_to_end(&mut bytes)` will block the listener thread indefinitely if a client connects but never sends data or hangs mid-write. While blocked, no other hook notifications can get through. Could you add a timeout to prevent this?

app/src/terminal/cli_agent_sessions/ipc.rs
```diff
@@ -0,0 +1,176 @@
...
+};
+
+use async_channel::Sender;
+
+use crate::terminal::{cli_agent_sessions::event::CLI_AGENT_NOTIFICATION_SENTINEL, event::Event};
+
+static NEXT_SOCKET_ID: AtomicU64 = AtomicU64::new(1);
+const MAX_IPC_MESSAGE_BYTES: u64 = 1024 * 1024;
+
+pub struct CLIAgentIpcListener {
```
harryalbert, line 21: nit: this could be `pub(crate)`

harryalbert (changes requested):

KRRT7:
@cla-bot check

app/src/terminal/cli_agent_sessions/ipc.rs
```diff
@@ -0,0 +1,176 @@
...
+    let socket_id = NEXT_SOCKET_ID.fetch_add(1, Ordering::Relaxed);
+    let uid = unsafe { libc::getuid() };
+    PathBuf::from(format!(
+        "/tmp/warp-cli-agent-{uid}/terminal-{}-{socket_id}.sock",
+        std::process::id()
+    ))
+}
+
+#[cfg(test)]
+mod tests {
```
harryalbert, line 150: nit: can we move these tests into a separate tests file?

harryalbert:
@KRRT7 I'm a little concerned that this solution doesn't work on windows — I think we might need a more generalizable fix

KRRT7:
Could we treat this as a Warp-owned IPC bridge instead of writing OSC to /dev/tty from the hook?\n\nThe plugin already builds the structured payload, so Warp could export a local endpoint + token into the shell, and the plugin could hand the JSON to a tiny helper that forwards it to Warp. That avoids the current Claude Code hook TTY issue, keeps the fallback for older Warp builds, and avoids nc -U / python3 portability problems.

KRRT7:
I believe this has been fixed upstream w support from anthropic, closing due to that

PR #10726, opened 2026-05-12
Stop old remote daemon when autoupdating

kevinyang372:
## Summary

Fix an infinite version-mismatch loop during remote server initialization after a client update (e.g. `stable_03` → `stable_04`).

## Root cause

The remote server uses a daemon architecture: the `remote-server-proxy` connects to a long-lived daemon via a Unix socket. When the client updates, the new binary is installed but the **old daemon process keeps running** from the previous binary. The proxy finds it alive (via PID check) and connects to it, so the daemon reports the old version → mismatch → client removes the new binary → reinstalls → same old daemon → loop.

The APP-3805 version-skew fix added binary removal on mismatch, but missed the daemon — removing the binary does not stop the running process.

## Fix

Added `stop_remote_server_daemon()` to the `RemoteTransport` trait and `SshTransport`. On version mismatch, after removing the stale binary, we now also kill the old daemon process and clean up its PID/socket files. The next connection attempt starts a fresh daemon from the correct binary.

### Files changed
- `crates/remote_server/src/transport.rs` — new trait method
- `app/src/remote_server/ssh_transport.rs` — SSH implementation (single round-trip: read PID, kill, rm files)
- `crates/remote_server/src/manager.rs` — call `stop_remote_server_daemon()` alongside `remove_remote_server_binary()` in the version mismatch handler

---

[Oz conversation](https://staging.warp.dev/conversation/ed895867-f6c2-418f-b8a9-bac814ddd93b) | [Plan](https://staging.warp.dev/drive/notebook/SfDfpJHCRtgWfnPANoVfL2)

## Testing
Tested locally by spawning a named version client and confirmed the mismatched version client is successfully killed

### DIFF
--- a/app/src/remote_server/ssh_transport.rs
+++ b/app/src/remote_server/ssh_transport.rs
@@ -337,6 +337,44 @@ impl RemoteTransport for SshTransport {
         })
     }
 
+    fn stop_remote_server_daemon(
+        &self,
+    ) -> Pin<Box<dyn Future<Output = anyhow::Result<()>> + Send>> {
+        let ssh_socket_path = self.socket_path.clone();
+        let pid_path = self.remote_daemon_pid_path();
+        let daemon_socket_path = self.remote_daemon_socket_path();
+        Box::pin(async move {
+            // Read the PID, kill the daemon, and clean up its files in a
+            // single SSH round-trip. The tilde-safe expansion mirrors
+            // install_remote_server.sh: use `case` + `${var#~}` instead
+            // of `${var/~/...}` to avoid bash version quirks.
+            let cmd = format!(
+                "pid_path={pid_path}; \
+                 case \"$pid_path\" in \"~\"|\"~/\"*) pid_path=\"${{HOME}}${{pid_path#\\~}}\";; esac; \
+                 sock_path={daemon_socket_path}; \
+                 case \"$sock_path\" in \"~\"|\"~/\"*) sock_path=\"${{HOME}}${{sock_path#\\~}}\";; esac; \
+                 if [ -f \"$pid_path\" ]; then \
+                   pid=$(cat \"$pid_path\" 2>/dev/null) && \
+                   [ -n \"$pid\" ] && kill \"$pid\" 2>/dev/null; \
+                 fi; \
+                 rm -f \"$pid_path\" \"$sock_path\""
+            );
+            log::info!("Stopping stale remote server daemon: {cmd}");
+            let output = remote_server::ssh::run_ssh_command(
+                &ssh_socket_path,
+                &cmd,
+                remote_server::setup::CHECK_TIMEOUT,
+            )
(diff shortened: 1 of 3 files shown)


kevinyang372:
* **#10726** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10726?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10726?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/remote_server/ssh_transport.rs
```diff
@@ -337,6 +337,44 @@ impl RemoteTransport for SshTransport {
...
+                 case \"$pid_path\" in \"~\"|\"~/\"*) pid_path=\"${{HOME}}${{pid_path#\\~}}\";; esac; \
+                 sock_path={daemon_socket_path}; \
+                 case \"$sock_path\" in \"~\"|\"~/\"*) sock_path=\"${{HOME}}${{sock_path#\\~}}\";; esac; \
+                 if [ -f \"$pid_path\" ]; then \
+                   pid=$(cat \"$pid_path\" 2>/dev/null) && \
+                   case \"$pid\" in ''|*[!0-9]*|0) ;; *) kill \"$pid\" 2>/dev/null;; esac; \
+                 fi; \
+                 rm -f \"$pid_path\" \"$sock_path\""
+            );
+            log::info!("Stopping stale remote server daemon: {cmd}");
```
alokedesai, line 362: logging the entire cmd doesn't seem valuable

crates/remote_server/src/manager.rs
```diff
@@ -975,6 +975,24 @@ impl RemoteServerManager {
...
                 );
             }
+
+            // Also stop the running daemon process. The daemon is
+            // long-lived and was likely started from the old binary,
+            // so it will keep reporting the old version. Without
+            // stopping it, the next proxy invocation finds the old
+            // daemon still alive (via PID file) and connects to it,
+            // producing the same version mismatch in a loop.
+            if let Err(e) = transport
```
alokedesai, line 985: So a few thoughts on this:
1) Before `remove_remote_server_binary` we should actually have a `kill` or `close` function that sends a message to the daemon to kill itself. Advantage of this is that we can also call this when the app terminates (it's questionable if we want this but i think it's better because it gives the user another opportunity to bail out of the 10min remote server idle period if they're hitting bugs with the client <> server connection)

2) This is arguably a reason for the proxy to not actually be a true proxy. The ideal `kill` implementation on the proxy would be to _not_ send it to the daemon but actually have the proxy intercept the message and send `SIGTERM` to the process so it terminates. Not something we need to fix this second, but we should discuss how we want to fix that

PR #10668, opened 2026-05-11
Enrich initialization error telemetry

kevinyang372:
## Description

Enrich remote server initialization telemetry with SSH subprocess exit status data and classify initialization failures as user-initiated cancellations vs real errors.

### Problem
The `RemoteServerInitialization` telemetry error field only ever contains `"initialize: Response channel closed before receiving a reply"`, providing no diagnostic insight. With ~91% success rate and only 2 Sentry server panics, most of the ~9% failures are NOT server crashes — they are user cancellations (exiting SSH during init) that were indistinguishable from real failures.

### Changes

**Cancellation detection** (`manager.rs`):
- Capture SSH subprocess exit status from `Initializing._child` before `mark_session_disconnected` runs
- Classify failures as cancelled when: session was already deregistered (user exited) OR transport reports unrecoverable disconnect (e.g. SSH exit 255) and process was not signal-killed
- Add `is_cancelled: bool` to `SessionConnectionFailed` event — the manager computes this with full context, subscribers just read the flag
- Always emit `SessionConnectionFailed` so `RemoteServerController` can flush stashed bootstrap state (fixes Oz review feedback)

**Telemetry enrichment** (`events.rs`, `terminal/view.rs`):
- Add `exit_code: Option<i32>` and `signal_killed: Option<bool>` to `RemoteServerInitialization` telemetry payload
- Terminal view skips telemetry and failed banner when `is_cancelled` is true, but always forwards `RemoteServerFailed` terminal event for downstream subscribers

**Error string cleanup** (`manager.rs`):
- Remove redundant `"connect: "` / `"initialize: "` prefix from `ConnectAndHandshakeError` Display — `phase` is already a separate telemetry field
- Delegate transport-specific exit code classification to `transport.is_reconnectable()` instead of hardcoding SSH exit code 255

[Conversation](https://staging.warp.dev/conversation/eeb2663e-6cce-404c-b17d-76e296354a85) | [Plan](https://staging.warp.dev/drive/notebook/SIBjqIbPCmnE4ELohM8JTM)

## Linked Issue
- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Testing
- Telemetry-only change with no user-visible UI impact
- Verified clippy passes clean on `remote_server` and `warp` crates

- [x] I have manually tested my changes locally with `./script/run`

## Agent Mode
[...]

PR #10728, opened 2026-05-12
Wait on child process to exit before showing error

kevinyang372:
## Description
In my previous fix #10668 , we were trying to capture exit status with `async_process::Child::try_status`. Note that this is a synchronous method that will return None if the Child has not dropped yet -- this is very likely to happen in a killed stream

Instead, we should try to call `async_process::Child::status` which returns an async future that only resolves when the child is definitely exited. We also added a small timeout in case the child takes a while to exit so we don't block on it forever.

## Testing
Confirmed now exit status is consistently captured
```
16:47:48.226 [WARN] [remote_server::manager] DEBUG: Sleep complete, proceeding with initialize handshake
16:47:48.227 [WARN] [remote_server::manager] Remote server connection failed: session=SessionId(17786188531007) error=Connection was dropped
16:47:48.227 [INFO] [remote_server::manager] Remote server process exited (async): session=SessionId(17786188531007) code=Some(255) signal_killed=false
16:47:48.227 [INFO] [remote_server::manager] Remote server connection failure exit status captured: session=SessionId(17786188531007) exit_code=Some(255) signal_killed=Some(false)
```

### DIFF
--- a/crates/remote_server/src/manager.rs
+++ b/crates/remote_server/src/manager.rs
@@ -38,6 +38,11 @@ const MAX_RECONNECT_ATTEMPTS: u32 = 2;
 /// Delay between reconnection attempts.
 #[cfg(not(target_family = "wasm"))]
 const RECONNECT_DELAY: Duration = Duration::from_secs(2);
+/// Brief timeout for awaiting a child process's exit status after a
+/// connection failure. Gives the SSH subprocess time to report its exit
+/// code and signal status before we give up and report `None`.
+#[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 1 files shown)


kevinyang372:
* **#10728** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10728?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10728?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

MaggieShan (approved):

PR #10729, opened 2026-05-12
fix(terminal-input): keep pinned input anchored during inline interactive prompts (#10144)

### LINKED ISSUE #10144: Interactive input for aws configure jumps off the pinned bottom command line
Image

Image
Image

## Summary
Bug: interactive input for `aws configure` renders in the wrong place in Warp, leaving a large blank area and moving the active input away from the pinned bottom command line.

## Problem
When running `aws configure` in Warp, the terminal view shifts and shows a large blank area. The active input location moves upward so the user has to scroll or look above the visible command line to continue entering values. In some cases the input is effectively off-screen while the command is running.

This appears to be specific to Warp. The same shell setup, including Starship, works normally in the macOS Terminal app.

## Expected behavior
While `aws configure` is prompting for input, the active input area should remain visible and usable, with the command line anchored at the bottom when the terminal is configured for pinned-to-bottom input.

## Actual behavior
Running `aws configure` causes the visible input area to jump upward and leaves a large empty "page" of space in the terminal view. The line where input is expected is no longer where the user would expect it, and typing can happen outside the normal visible bottom command area.

Pressing `Ctrl-C` to cancel `aws configure` immediately restores normal behavior and returns the shell to the expected bottom-aligned prompt.

An additional observation: after running `exec zsh` and then running `aws configure`, the input still jumps upward, but only to a position a little above the middle of the screen and it remains visible. That suggests a fresh shell partially changes the rendering state, but does not fix the underlying issue.

1. Open Warp on macOS.
2. Use zsh with a custom or default (warp) prompt.
3. Run `aws configure` (or any aws cli command).
4. Observe that the visible input area jumps upward and a large blank area appears in the terminal.
5. Try typing the prompted values and note that the active input area is no longer aligned with the normal pinned-bottom command line.
6. Press `Ctrl-C` to cancel the command.
7. Observe that the terminal immediately returns to the normal bottom-aligned prompt/input behavior.

Additional variation:
1. Run `exec zsh`.
2. Run `aws configure` again.
3. Observe that the input still jumps upward, but less severely than before.

## Artifacts
Screenshots attached

## Warp version
v0.2026.04.29.08.57.01

## Operating system
macOS 26.4.1

lonexreb:
Closes #10144

## Summary

Fixes #10144. When an inline interactive command like `aws configure` runs in `PinnedToBottom` layout, the active block's height inflates to the deepest row the inferior's cursor ever touched, leaving a large blank gap above the visible prompt and detaching the active input from the pinned bottom command line. `Ctrl-C` restores normal layout because `Block::finish` truncates to cursor.

The fix turns on the existing `trim_trailing_blank_rows` mechanism for every block at `preexec`, gated on the same `TrimTrailingBlankLines` feature flag used for CLI-agent sessions, so the rollout matches the existing path.

## Root cause

`app/src/terminal/model/blockgrid.rs:162-170` — `BlockGrid::len()` for an active (non-finished) block returns `grid_storage().max_cursor_point.row + history_size + 1`. `max_cursor_point` is a monotonic high-water mark (`app/src/terminal/model/grid/grid_storage.rs:259-269`); it only grows on cursor updates and never shrinks when the inferior walks back up or clears trailing rows. `Block::output_grid_displayed_height` (`app/src/terminal/model/block.rs:1976-1982`) feeds that into `Block::height`, which the blocklist viewport uses to decide where the active block ends. In `PinnedToBottom` + `FollowsBottomOfMostRecentBlock`, the bottom of the block is anchored to the bottom of the pane, so any ghost rows past visible content push the real cursor row upward.

Readline-style prompts (Python `input()`, `aws configure`, `node`/`python` REPL) routinely produce this pattern via `\r\x1b[K` redraws and cursor probes that move past the current line. Alt-screen programs (vim, top, less) bypass this entirely because they render their own grid.

The trim mechanism that fixes it already exists — `GridHandler::set_track_content_length` and `BlockGrid::set_trim_trailing_blank_rows` cap `len_displayed()` at the bottommost row with visible content via a cached backward scan refreshed in `on_finish_byte_processing`. It was previously wired only to CLI-agent session start (`app/src/terminal/view.rs:12253-12256`).

## Fix

`app/src/terminal/model/block.rs::preexec` (~line 3315): after `state = BlockState::Executing`, call `self.output_grid.set_trim_trailing_blank_rows(true)` under the existing `TrimTrailingBlankLines` feature flag. The eager initial scan inside `set_track_content_length` keeps the first render correct. Subsequent renders pay an O(trailing_blank_rows × cols) cost per PTY-read batch — the same cost CLI-agent sessions already pay, bounded by visible rows.

The CLI-agent session handler still calls `set_trim_trailing_blank_rows` explicitly; the redundant call from `preexec` is idempotent.

## Test plan

Unit tests (in `app/src/terminal/model/blockgrid_tests.rs`):
- `test_trim_trailing_blank_rows_handles_inline_interactive_prompt_pattern` — reproduces the `aws configure` cursor-walk pattern (write two prompt rows, jump cursor to row 20, return cursor to row 1) and asserts `len()` reflects the high-water mark (21) while `len_displayed()` stays bounded by real content (≤3 rows).
- `test_without_trim_inline_interactive_prompt_pattern_balloons_height` — pins the without-fix behavior so the trim contract can't silently regress: with trim off, `len_displayed() == len() == 21` even though only one row has visible content.

Build status:
- [ ] `cargo check -p app` — could not run locally: this environment is missing the macOS `metal` shader toolchain that `warpui/build.rs` requires. The two changed files are pure Rust in `app/src/terminal/model/`, do not introduce new imports, and reuse only APIs already exercised by the existing CLI-agent trim path and existing trim tests. Please run CI to confirm.
- [x] Tests authored to use the same fixtures and helpers as the existing trim tests in the same file (`set_trim_trailing_blank_rows`, `on_finish_byte_processing`, `goto`, `input`, `linefeed`, `carriage_return`).

Manual repro (please verify on macOS):
1. Launch Warp with default `PinnedToBottom` input mode.
2. In a fresh shell, run `aws configure`.
3. Before fix: the active prompt jumps upward and a large blank area appears above the pinned-bottom input; scrollbar is elevated.
4. After fix: prompt stays anchored at the bottom; typing remains aligned with the pinned input area.
5. Regression sanity: run `top`, `vim`, `less` (alt-screen) — behavior unchanged. Run a CLI-agent session — behavior unchanged.

## Files

- `app/src/terminal/model/block.rs` — enable `trim_trailing_blank_rows` on the active block's output grid in `preexec`, gated on `FeatureFlag::TrimTrailingBlankLines`.
- `app/src/terminal/model/blockgrid_tests.rs` — two new tests covering the regression and the without-fix baseline.

## Manual testing note

### DIFF
--- a/app/src/terminal/model/block.rs
+++ b/app/src/terminal/model/block.rs
@@ -3296,6 +3296,26 @@ impl ansi::Handler for Block {
         self.state = BlockState::Executing;
         self.is_for_in_band_command = is_for_in_band_command;
 
+        // Enable trailing-blank-row trimming on the output grid so the block's
+        // displayed height tracks actual rendered content rather than the
+        // monotonic `max_cursor_point` watermark. Without this, inline
+        // interactive readline-style prompts (e.g. `aws configure`, REPLs that
+        // use `\r\x1b[K` to redraw the input line) can grow the block height
+        // past their visible content. In `PinnedToBottom` layouts this leaves
+        // a large blank area above the input and detaches the visible cursor
+        // from the pinned bottom command line — see issue #10144.
+        //
+        // Note: alt-screen programs (vim, top, less, etc.) take over their
+        // own grid and do not display the output grid in the blocklist, so
+        // turning trim on here is harmless for them. CLI-agent sessions
+        // already enable this from their session-start handler in
+        // `view.rs::handle_cli_agent_sessions_event`; this call is idempotent
+        // and the second `set_track_content_length(true)` is a no-op (the
+        // eager backward scan recomputes the already-correct value).
+        if FeatureFlag::TrimTrailingBlankLines.is_enabled() {
+            self.output_grid.set_trim_trailing_blank_rows(true);
+        }
+
         self.wakeup_after_delay();
     }
 
--- a/app/src/terminal/model/blockgrid_tests.rs
+++ b/app/src/terminal/model/blockgrid_tests.rs
@@ -244,3 +244,117 @@ pub fn test_cursor_display_point_not_clipped_when_trimming_disabled() {
         Some(CursorDisplayPoint::HiddenCache(Point::new(0, 1)))
     );
 }
+
+/// Regression test for #10144: when an inline interactive command (e.g.
+/// `aws configure`) writes a series of prompts and the inferior moves the
+/// cursor back up — for example to redraw the current input line after a
+/// terminal resize, or simply because the program walked its cursor down for
+/// some intermediate operation — the block's `max_cursor_point` watermark
+/// stays at the deepest row ever reached. Without trimming, `len_displayed`
+/// returns that watermark and the block height balloons past its actual
+/// content, leaving a large blank gap above the pinned-bottom input.
+///
+/// With trim enabled, `len_displayed` must track the last row with visible
+/// content, which is what the user actually sees.
+#[test]
+pub fn test_trim_trailing_blank_rows_handles_inline_interactive_prompt_pattern() {
+    let size = SizeInfo::new_without_font_metrics(40, 80);
+    let mut block_grid = BlockGrid::new(
+        size,
+        1000,
+        ChannelEventListener::new_for_test(),
+        ObfuscateSecrets::No,
+        PerformResetGridChecks::default(),
+    );
+
+    block_grid.start();
+    block_grid.set_trim_trailing_blank_rows(true);
+
+    // Simulate the inferior writing four `aws configure`-style prompts on
+    // successive rows, advancing the cursor as it goes.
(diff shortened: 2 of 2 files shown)


vorporeal:
for non-trivial changes, especially UI ones, we require some amount of manual testing by the author to validate the correctness of the change.

from our CONTRIBUTING.md file:

> Manual testing is required for changes that can be manually tested, and almost all changes can be manually tested. For small, isolated, and visual changes, you should include before and after screenshots. For larger, broad, or interactive changes, you should also include a narrated screen recording.

---

that said, i'm not convinced we want this behavior for _all_ block types.  that's a substantial change in order to fix these rare cases.  we should discuss the implications before moving any further here.

vorporeal (commented):
(marking as reviewed)

PR #10730, opened 2026-05-12
fix(settings-keybindings): honor user-defined keybindings over default conflicts (#9128)

### LINKED ISSUE #9128: Split Pane Down mapped to Ctrl Shift C not working (except in bash subshell) [Issue #8731]
### Pre-submit Checks

### Describe the bug

I was on vacation and didn't respond in time before this issue was closed as "not planned"
The most recent version of Warp (just updated) did not fix the issue.

Original bug report:

I have my split pane shortcuts to ctrl-shift-D (horizontal) and C (vertical).
This is the same setup that I have in Tabby.

Image

This key combination does not work. You can see from the Screenshot there are no conflicts.

In Tabby it works, so this is not because of some system override catching the key combination. 

NOTE: I just discovered, if I run `bash` directly in the Warp session, then the key combination works in Warp as well, but in the default prompt it does not work:

Image

### To reproduce

Open Warp
Map split pane to ctrl-shift-C
Press ctrl-shift-C -> nothing happens
run `bash`
Press ctrl-shift-C again, now it works

### Expected behavior

Warp should perform the actions bound in keyboard shortcuts.

_No response_

### Operating system (OS)

Windows

WSL on Win11

### Shell Version

GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)

### Current Warp version

v0.2026.04.15.08.45.stable_02

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

lonexreb:
Closes #9128

## Summary

Fixes #9128 — user keybindings (e.g. `Split Pane Down` → `Ctrl+Shift+C`) were silently shadowed by a default binding that used the same keystroke (`Copy` on Linux/Windows), so the user override only fired in narrow contexts (the bash-subshell case from the bug report).

### Root cause

`Keymap::bindings()` iterates editable bindings in reverse-registration order, then fixed bindings. The matcher takes the first binding whose keystroke + context predicate both match.

`pane_group::init` (registers `pane_group:add_down` / SplitPaneDown) runs **before** `terminal::init` (registers `terminal:copy`). In reverse-registration order, `terminal:copy` is iterated first. Its trigger resolves to `ctrl-shift-C` on Linux/Windows (via `cmd_or_ctrl_shift("c")`), and its context predicate `Terminal & !IMEOpen` matches the main terminal — so `Copy` wins, and the user's explicit Ctrl+Shift+C → SplitPaneDown override never fires.

Inside a bash subshell, the user happens to have block-selection state and altscreen flags that change which `Copy`-flavored binding matches, which is why the override appears to work there — pure coincidence, not by design.

### Fix

Partition editable bindings in `Keymap::editable_bindings()` / `editable_custom_action_bindings()` so user-customized ones (`custom_trigger.is_some()`) iterate **before** defaults. Within each partition, reverse-registration order is preserved, so default-vs-default precedence is unchanged. This makes an explicit user override always beat a default binding that happens to share the same keystroke, regardless of which crate registered first.

### Files

- `crates/warpui_core/src/keymap.rs` — partition the editable-binding iterators
- `crates/warpui_core/src/keymap/matcher_tests.rs` — new regression test exercising the #9128 scenario (SplitPaneDown rebound to ctrl-shift-C while Copy default also uses ctrl-shift-C)
- `crates/warpui_core/src/keymap_tests.rs` — update the existing custom-trigger ordering test to reflect the new precedence semantics

## Test plan

- [x] cargo test -p warpui_core --lib keymap (19/19 passing including new regression test)
- [x] cargo test -p warpui_core --lib (275/275 passing)
- [x] cargo clippy -p warpui_core --tests --no-deps -- -D warnings (clean)
- [x] cargo fmt --check -p warpui_core (clean)
- [ ] Manual: on Linux/WSL, bind Split Pane Down to Ctrl+Shift+C in Settings, press in main terminal, confirm split fires
- [ ] Manual: confirm default Ctrl+Shift+C still copies when no override set
- [ ] Manual: confirm other defaults (Ctrl+Shift+D, etc.) unchanged

## Manual testing note

### DIFF
--- a/crates/warpui_core/src/keymap.rs
+++ b/crates/warpui_core/src/keymap.rs
@@ -439,9 +439,23 @@ impl Keymap {
     /// Items will be returned in the reverse order they were registered, the most recently
     /// registered editable binding will have the highest precedence
     fn editable_bindings(&self) -> impl Iterator<Item = EditableBindingLens<'_>> {
-        self.editable_bindings
+        // Iterate user-customized bindings before defaults, so an explicit user
+        // override always wins precedence over a default binding that happens to
+        // use the same keystroke. Within each partition, retain reverse-registration
+        // order (most recently registered first), matching the prior behavior.
+        // See https://github.com/warpdotdev/warp/issues/9128.
+        let user_overridden = self
+            .editable_bindings
             .iter()
             .rev()
+            .filter(|binding| binding.custom_trigger.is_some());
+        let defaults = self
+            .editable_bindings
+            .iter()
+            .rev()
+            .filter(|binding| binding.custom_trigger.is_none());
+        user_overridden
+            .chain(defaults)
             .map(|binding| binding.as_lens())
             .filter(|binding| binding.is_enabled())
     }
@@ -450,7 +464,9 @@ impl Keymap {
     /// modified by the custom bindings, where appropriate.
     ///
     /// Editable bindings will be returned first, followed by any fixed bindings in the reverse
-    /// order they were added.
+    /// order they were added. Among editable bindings, user-customized ones take precedence
+    /// over defaults so an explicit user override is not shadowed by a default binding that
+    /// shares the same keystroke (#9128).
     fn bindings(&self) -> impl Iterator<Item = BindingLens<'_>> {
         self.editable_bindings()
             .map(|lens| lens.as_binding())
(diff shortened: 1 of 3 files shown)


vorporeal (commented):
for non-trivial changes, especially UI ones, we require some amount of manual testing by the author to validate the correctness of the change.

from our CONTRIBUTING.md file:

> Manual testing is required for changes that can be manually tested, and almost all changes can be manually tested. For small, isolated, and visual changes, you should include before and after screenshots. For larger, broad, or interactive changes, you should also include a narrated screen recording.

PR #10731, opened 2026-05-12
feat(terminal-input): add command_input_min_lines setting (#9197)

### LINKED ISSUE #9197: Ability to increase the height/vertical space of the command input field
### Pre-submit Checks

I would like to have the ability to increase the height/vertical space of the terminal input field.

I work on macOS and often in fullscreen or in a full window. To conserve space, I always have my dock auto-hide, but when I need to select an option at the bottom of the Warp, my dock pops up, and it won't let me select the said function. 

### Additional context

_No response_

### Operating system (OS)

macOS

3

None

lonexreb:
## Summary

Adds a `terminal.input.command_input_min_lines` setting (`u8`, default `1`, clamped to `1..=20`) that controls the minimum number of lines reserved for the bottom-pinned command input field.

Closes #9197.

## UX / Motivation

Users on macOS with an auto-hide dock report that the dock can overlap the command input area, leaving very little vertical space to type. Issue #9197 asks for a way to increase the height of the input field. A drag-handle was also floated but is deferred — this PR ships the requested settings-level control first.

There is a related but distinct angle: an OS-level safe-area padding setting could address the dock-collision case directly. That is mentioned here for the maintainers' awareness; the input-height setting is still the requested feature and ships now.

## Settings

| Field | Value |
| --- | --- |
| Name | `command_input_min_lines` (`CommandInputMinLines`) |
| TOML path | `terminal.input.command_input_min_lines` |
| Type | `u8` |
| Default | `1` (preserves current rendering) |
| Validation | clamped to `[1, 20]` via `Setting::validate` and `effective()` |
| Sync | `Globally(RespectUserSyncSetting::Yes)` |
| Platforms | `ALL` |

The default of `1` means the rendered behaviour is unchanged unless the user opts in.

## Rendering

`render_input_box` in `app/src/terminal/input.rs` now reads the setting and applies a `with_min_height(min_lines * line_height)` floor on the `ConstrainedBox`, while keeping the existing `with_max_height(editor_height_rounded_down)` cap. The floor is also `.min(max_height_px)` so the setting can never force the input larger than the half-pane upper bound that already constrains it.

Auto-grow on multiline content continues to work because only a minimum is set, not a fixed height.

## Backwards Compatibility

- Default of `1` line matches existing behaviour exactly.
- Existing settings files / persisted state are unaffected.
- Out-of-range values from any source are clamped silently (with a default-safe floor of `1`).

## Test Plan

Automated:

- [x] `cargo check -p warp`
- [x] `cargo test -p warp --lib command_input_min_lines_tests` — 5 unit tests for clamping and height math (default range, sub-range, oversize, zero/negative line height).
- [x] `cargo test -p warp --lib settings::schema_validation_tests` — confirms the new file-default validates against the generated JSON schema.
- [x] `cargo test -p warp --lib terminal::input` — existing 143 input tests still pass.
- [x] `cargo fmt -- --check`

Manual verification steps:

- [ ] Default install: input field renders at one line of height (no visible change).
- [ ] Set `terminal.input.command_input_min_lines = 4` in the user settings TOML; relaunch; input field now reserves four lines at the bottom of the terminal pane.
- [ ] Type multi-line input above the configured floor; the input grows up to the existing half-pane cap.
- [ ] Set the value to `0`, `99`, or `255` directly in the TOML; verify the runtime clamps to `1` and `20` respectively without panicking.
- [ ] Confirm both Classic and Universal Developer Input paths honour the floor.

## Notes for Reviewers

- The pure helper `command_input_min_height_px(min_lines, line_height_px)` is exposed so the layout math is unit-testable without constructing an `AppContext`. It is also defensively guarded against non-positive line heights.
- Validation is centralised in `clamp_command_input_min_lines` and used by both `Setting::validate` (write-time) and `CommandInputMinLines::effective()` (read-time) so callers never see an out-of-range value even if persisted state ever drifts.

### DIFF
--- a/app/src/settings/input.rs
+++ b/app/src/settings/input.rs
@@ -181,8 +181,59 @@ define_settings_group!(InputSettings,
             sync_to_cloud: SyncToCloud::Never,
             private: true,
         },
+        command_input_min_lines: CommandInputMinLines {
+            type: u8,
+            default: 1,
+            supported_platforms: SupportedPlatforms::ALL,
+            sync_to_cloud: SyncToCloud::Globally(RespectUserSyncSetting::Yes),
+            private: false,
+            toml_path: "terminal.input.command_input_min_lines",
+            description: "Minimum number of lines reserved for the command input field. Values are clamped to 1..=20.",
+        },
     ]
 );
+
+impl CommandInputMinLines {
+    /// Minimum allowed value for `command_input_min_lines`.
+    pub const MIN: u8 = 1;
+    /// Maximum allowed value for `command_input_min_lines`.
+    ///
+    /// Capped at 20 lines so that overly large values cannot crowd out the
+    /// terminal block list. The render path also constrains the field to at
+    /// most half the visible terminal height regardless of this setting.
+    pub const MAX: u8 = 20;
+
+    fn validate(&self, new_value: u8) -> u8 {
+        clamp_command_input_min_lines(new_value)
+    }
+
+    /// Returns the validated current value of the setting.
+    ///
+    /// The stored value is always validated on load and on update, but reading
+    /// through this accessor guarantees callers never see an out-of-range value
+    /// even if persisted state ever drifts.
+    pub fn effective(&self) -> u8 {
+        clamp_command_input_min_lines(**self)
+    }
+}
+
+/// Clamp a raw `command_input_min_lines` value into the supported range.
+///
+/// Exposed at module scope so it can be exercised in unit tests without
+/// constructing a full settings group.
+pub fn clamp_command_input_min_lines(value: u8) -> u8 {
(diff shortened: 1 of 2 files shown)


lucieleblanc:
Hi @lonexreb, the discussion in PR #9197 is still waiting for design input. Without any screenshots or screen recordings, it's not immediately clear whether the code works as expected; in any case, I'm not sure this is the right way to get around the dock issue. If you have ideas or a proposal for how to solve this, please post mocks/screenshots in that thread instead and get approval there instead.

I'm going to close this for now. Feel free to re-open a PR once the discussion in the #9197 issue thread has settled and you've tested + screen-captured your solution. On MacOs, you can use `cmd`-`shift`-`5` to record a specific window.

PR #10735, opened 2026-05-12
fix(window-tabs-panes): preserve terminal scroll position across tab switches (#9171)

### LINKED ISSUE #9171: Scroll position resets to the top when switching tabs (especially with long outputs)
### Describe the bug
When switching between tabs, the scroll position frequently resets to the top of the buffer instead of staying at the bottom or the last viewed position. This is particularly disruptive when working with tools that generate long outputs, such as Gemini CLI, Claude, or when running 'cat' on large log files.

### To reproduce
1. Generate a long output in Tab A (e.g., cat a large log or interact with an AI CLI).
2. Switch to Tab B.
3. Switch back to Tab A.
4. Observe that the buffer has scrolled to the top/beginning of the history instead of remaining at the latest output.

### Expected behavior
The scroll position should be strictly preserved when switching between tabs.

### Details
- Warp Version: v0.2026.04.15.08.45.stable_04
- OS Version: Linux (Kernel 6.x/Ubuntu/etc. as per user environment)

lonexreb:
Closes #9171

## Summary

Fixes #9171.

When a user scrolls up to read long output (`cat` on a large log, an AI CLI session, etc.) and switches to another tab, returning yanks the scroll to the top of the buffer — disrupting the read in progress. Inactive tabs aren't in the layout tree, but their `TerminalView` is still live and `scroll_position` keeps getting mutated by streaming output and rich-content insertions, so by the time the tab is reactivated, the user's last-viewed position has drifted away.

## What changed

- `ScrollState` (in `app/src/terminal/block_list_viewport.rs`) now carries `saved_for_tab_deactivation: Option<ScrollPosition>` with two new methods:
  - `save_for_tab_deactivation()` — snapshots `position` into the slot; idempotent so the first deactivation signal wins (matters because window-blur + tab-switch can fire in sequence and we want the pre-drift value).
  - `restore_after_tab_activation()` — writes the snapshot back to `position` and clears the slot.
- `TerminalView` exposes `save_scroll_position_for_tab_deactivation` and `restore_scroll_position_after_tab_activation` wrappers.
- `Workspace::set_active_tab_index` snapshots every terminal view in the outgoing tab's pane group before flipping the index, then restores every terminal view in the incoming tab's pane group after notifying focus change.

The contract is exactly what the issue asks for: on tab activation, the user returns to the scroll position recorded at the moment of last deactivation. Render-time clamping (`ViewportState::scroll_top_in_lines`'s existing `.min(max_scroll_top)`) handles the content-shrink edge case for free.

The fix is deliberately surgical — we save/restore the opaque `ScrollPosition` value without redesigning anything around block-list virtualization, scroll anchoring, or input-mode (`PinnedToBottom` / `PinnedToTop` / `Waterfall`) semantics.

## Test plan

- [x] New unit tests in `scroll_state_tab_switch_tests` (`block_list_viewport.rs`) — all 5 pass:
  - `save_then_restore_returns_to_recorded_position` — round-trip with simulated mid-deactivation drift.
  - `restore_without_prior_save_is_a_noop` — first activation / never-deactivated case.
  - `save_is_idempotent_within_one_deactivation` — multiple save signals don't overwrite the pre-drift snapshot.
  - `restore_clears_snapshot_so_next_cycle_is_fresh` — back-to-back deactivate/activate cycles work.
  - `restore_is_a_noop_when_position_did_not_drift` — clean exit when nothing moved.
- [x] `cargo check -p warp --lib` — clean.
- [x] `cargo clippy -p warp --lib --no-deps` — clean.
- [x] `cargo fmt -p warp -- --check` — clean.
- [x] `cargo test -p warp --lib scroll` — 65 scroll-related tests pass (no regressions).
- [ ] Manual repro on Linux per the original bug report (Classic / Start-at-top mode, shell PS1 input, long-running output): expected fix.

## Manual testing note

### DIFF
--- a/app/src/terminal/block_list_viewport.rs
+++ b/app/src/terminal/block_list_viewport.rs
@@ -38,11 +38,26 @@ use super::{
 /// Wraps a scroll position for the purposes of centralizing update logic.
 pub struct ScrollState {
     position: ScrollPosition,
+
+    /// The scroll position recorded at the moment of last tab deactivation.
+    ///
+    /// While a tab is inactive, the blocklist is not rendered and `after_terminal_view_layout`
+    /// doesn't fire — but streaming output, rich-block insertion, and other model updates
+    /// can still mutate `position` (e.g. switching `FixedAtPosition` to
+    /// `FollowsBottomOfMostRecentBlock`). On reactivation we restore the snapshot so the
+    /// user returns to exactly where they left off, instead of getting yanked to the top
+    /// or bottom of a buffer that grew while they were away.
+    ///
+    /// This is `None` whenever the owning tab is active or has never been deactivated.
+    saved_for_tab_deactivation: Option<ScrollPosition>,
 }
 
 impl ScrollState {
     pub fn new(position: ScrollPosition) -> Self {
-        Self { position }
+        Self {
+            position,
+            saved_for_tab_deactivation: None,
+        }
     }
 
     pub fn position(&self) -> ScrollPosition {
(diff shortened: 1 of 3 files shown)


captainsafia:
Closing this PR out as it lacks visual verifications of the fix.

Issue #9418, opened 2026-04-29
Typing synch error on slow systems.


raydelto:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

While running warp on Mint Linux on a Macbook pro 2012, when typing letters were consistently entered to warp in a different order that they were pressed. To have control over what I really wanted to type I had to type very slowly.

### To reproduce

Run warp on a resource limited environment and start typing on WARP. In my case was a Macbook pro 2012 with 8GB of RAM. The bug was consistently showing up.

### Expected behavior

The expected behaviour is that text is shown on WARP in the same order as typed. What it is currently happenning is that letters are not been introduced to WARP in the same order they were typed, seems like a thread race condition.

### Operating system (OS)

Linux

### Operating system and version

Linux Mint 21.3 XFCE

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

Yes, this issue prevents me from using Warp daily.

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

PR #10737, opened 2026-05-12
draft: diagnosis for #9418 (typing order on slow systems)

lonexreb:
## Summary

**This is a draft diagnosis PR, not a fix.** Issue #9418 reports printable characters appearing out of order while typing on a resource-constrained Linux Mint 21.3 XFCE machine. I traced the full input pipeline and could not find an actual race or reorder; opening a draft so a maintainer can weigh in before any speculative code change lands.

Closes: nothing yet — needs more info from reporter to make this actionable.

## What this PR changes

A single comment in `crates/warpui/src/windowing/winit/event_loop/mod.rs` in the `ConvertedEvent::KeyDownWithTypedCharacters` arm, documenting the ordering invariant of the printable-character hot path. Zero behavior change.

## Investigation trace

End-to-end pipeline for a printable keypress on Linux:

1. `Event::WindowEvent { WindowEvent::KeyboardInput }` arrives on the winit event loop (single-threaded). `event_loop/mod.rs:1270`
2. `convert_keyboard_input_event` (sync, pure). `event_loop/key_events.rs:60`
3. `handle_converted_warpui_event` dispatches `KeyDown`, then `TypedCharacters` if unhandled. `event_loop/mod.rs:1081-1100`
4. `dispatch_event` is fully synchronous; any side effects emitted via `ctx.emit` go onto the `pending_effects: VecDeque<Effect>` and are drained `pop_front` in FIFO order. `core/app.rs:3136`
5. `TerminalView::typed_characters_on_terminal` -> `system_insert` (editor) or `write_user_bytes_to_pty` -> `ctx.emit(Event::WriteBytesToPty)`. `terminal/view.rs:7979-8009`
6. Subscription -> `PtyController::write_bytes` -> `send_message_to_event_loop(Message::Input)`. `writeable_pty/pty_controller.rs:636-735`
7. mio channel -> `EventLoop::drain_recv_channel` -> `state.write_list.push_back(input)` (a `VecDeque<Cow<[u8]>>`). `local_tty/event_loop.rs:164-176`
8. Write loop consumes `write_list` front-to-back to the PTY fd.

Every hop is FIFO; the entire flow runs on the UI thread; there is no background-thread hop for printable characters; `pending_flushes`/`flushing_effects` guards in `core/app.rs` prevent re-entrant flush reorder. No coalescing, no batching, no debounce in this path.

## Why I am not shipping a code fix

The reporter's symptom — letters arriving in a different order than typed on a Linux Mint machine with 8GB RAM — is the kind of bug where a speculative fix is more likely to regress good behavior than to fix the real cause. Probable real causes that the current report cannot disambiguate:

- **Display latency, not insertion order.** Under heavy load, render frames could lag per-character, making characters appear briefly out of order even though they were processed in order. The reporter has not provided a screen recording confirming insertion order vs render order.
- **Xorg/winit-level reordering on this specific hardware/driver combo.** A 2012 MacBook Pro running Linux Mint XFCE has well-known graphics-stack quirks. winit's X11 backend, the xkb composition state, or even a low-level input method (ibus/fcitx) could be involved. The reporter did not say whether an IME is active.
- **IME `Ime::Preedit` interleaving with `KeyboardInput`.** On Linux with IME enabled, winit emits both keyboard events and `Ime::Preedit`/`Ime::Commit`. These are processed in arrival order, but if the user is typing fast and the IME is committing pre-edits asynchronously at the OS level, characters can land non-monotonically. This is OS/IME behavior, not Warp.
- **Stale Warp version.** The reporter never gave a version or channel, and `oz-for-oss` already requested both.

## Required follow-up before a fix can be designed

1. Screen recording showing the symptom and roughly what was typed vs what appeared.
2. Exact Warp version and channel.
3. Whether an IME (ibus, fcitx, scim) is active on the reporter's system.
4. Warp logs from a session in which this reproduces.

oz-for-oss already requested 1 and 2 on the issue; this PR documents the diagnostic ground truth so a maintainer doesn't have to redo the trace.

## Test plan

- [ ] N/A — comment-only change, no behavior modified.
- [ ] When the reporter supplies a recording + version + IME info, revisit and design a targeted fix (most likely an IME interleave guard or render-side per-frame coalescing, scoped to Linux).

### DIFF
--- a/crates/warpui/src/windowing/winit/event_loop/mod.rs
+++ b/crates/warpui/src/windowing/winit/event_loop/mod.rs
@@ -1085,6 +1085,15 @@ impl EventLoop {
                 // that include the cmd key because they are assumed to be
                 // intended as OS-level or application-level shortcuts. This matches the behavior
                 // on macOS.)
+                //
+                // Ordering invariant (see issue #9418): each winit `KeyboardInput` is processed
+                // synchronously on the UI thread, in the same order winit delivered it. Both
+                // `dispatch_event` calls below are synchronous; any side effects they emit (e.g.
+                // `WriteBytesToPty`) are appended to `pending_effects` and flushed in FIFO order.
+                // Downstream, the PTY event loop's mio channel and `State::write_list` (a
+                // `VecDeque`) also preserve order. Do NOT introduce any background-thread hop,
+                // batching, or coalescing for printable characters here without an equivalent
+                // ordering guarantee.
                 let cmd_pressed = match &event {
                     crate::event::Event::KeyDown { keystroke, .. } => keystroke.cmd,
                     _ => false,


lonexreb:
Closing stale draft to reduce review burden. Will reopen with a working demo if/when the underlying issue regains priority.

PR #10742, opened 2026-05-12
fix(workspace): suppress stale "still running" toast when input box is hidden for non-command reasons

### LINKED ISSUE #9100: A command in this session is still running.
### Pre-submit Checks

### Describe the bug

Why can't I automatically open a new tab anymore? How should I set it up

### To reproduce

click Warp Drive

### Expected behavior

_No response_

_No response_

### Operating system (OS)

macOS

A command in this session is still running.

### Shell Version

_No response_

### Current Warp version

_No response_

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

lonexreb:
Closes #9100.

## Summary

The toast **"A command in this session is still running."** fires falsely when the user clicks something in Warp Drive (or other call sites of `focus_terminal_input`) and the active terminal's input box happens to be hidden for reasons unrelated to a running command. This restructures `focus_terminal_input` to (a) only create a new pane in the explicit "open new" branch and (b) only fire the busy toast when the active block is actually executing.

## Root cause

`app/src/workspace/view.rs:focus_terminal_input` (lines ~15165–15237 before this change):

1. It called `active_session_view(...).unwrap_or_else(|| { add_terminal_pane(...); ...unwrap() })` to obtain a terminal view handle. Even for the `RequireExisting` and `OpenIfNone` paths, this **always** created a new pane when none existed.
2. It then called `self.is_input_box_visible(ctx)`. If false, and the fallback behavior was anything other than `OpenIfNeeded`, it fired the stale-command toast and returned `None` — leaving the user with a surprise new pane and a misleading "still running" error.
3. `TerminalView::is_input_box_visible` returns `false` for many reasons that have nothing to do with a running command:
   - `shared_session_status().is_view_pending()` (`app/src/terminal/view.rs:7227`) — true for freshly-created panes still establishing the session.
   - `is_alt_screen_active()` without agent control/tag-in (`view.rs:7220`).
   - Active SSH remote-server choice block / remote-server setup in progress (`view.rs:7261–7278`).
   - Active AI block blocked on user confirmation or expanded running commands (`view.rs:7280–7287`).
   - Active env-var collection block, init project last step, cloud-agent pre-first-exchange, etc.

None of those are "a command is running." Mapping all of them to the `still running` toast is the bug reported in #9100.

## Fix

`app/src/workspace/view.rs`
- `focus_terminal_input` now resolves the existing active session **without** creating a new pane up front. The new pane is only added in the explicit "open new" branch at the bottom, eliminating the surprise-pane behavior for `RequireExisting`.
- When there *is* an existing session whose input box is hidden, we delegate to a new `should_show_busy_toast` helper that returns `true` only when the active block is actually executing.

`app/src/terminal/view.rs`
- New `TerminalView::is_active_block_running` predicate that composes the existing `is_writing_or_executing_in_band_command`, `Block::is_executing`, and `Block::is_active_and_long_running` predicates. We intentionally exclude blocks under agent control / tagged-in agents so the toast doesn't surface for agent flows that legitimately hide the input.

## Test plan

`app/src/terminal/model/block_tests.rs` — new regression test `test_idle_and_finished_blocks_are_not_executing` that asserts the underlying predicates we now use to gate the toast (`Block::is_executing`, `Block::is_active_and_long_running`) return `false` for:
- a freshly-built (never-started) block, and
- a block that ran briefly and finished with exit 0.

These are exactly the states that previously surfaced the false toast in #9100.

## Manual testing note

End-to-end repro requires building the app, which depends on `crates/warpui` compiling Metal shaders via Xcode's `metal` toolchain (`xcrun metal`). The Xcode toolchain is not available in this environment — only the Command Line Tools are — so `cargo check -p warp` and `cargo test -p warp` both fail at the `crates/warpui/build.rs:92` shader-compile step:

```
thread 'main' panicked at crates/warpui/build.rs:92:5:
error compiling metal shaders to .air; xcrun: error: unable to find utility "metal", not a developer tool or in PATH
```

For local verification I therefore relied on:
- Reading the surrounding state-machine code in `terminal/view.rs::is_input_box_visible`, `pane_group/mod.rs::active_session_view`, and `terminal/model/block.rs` to confirm the predicates compose correctly.
- The new unit test covering the predicate-level invariant.
- A crate-scoped `cargo check -p command` to confirm the shared workspace deps still build.

A reviewer with the full Xcode toolchain can finish the end-to-end repro by:
1. Triggering any code path that hides the input box without running a command (e.g., enter `vim`, switch to alt-screen, or open Warp Drive on a fresh tab while shared-session is still resolving).
2. Clicking a Warp Drive workflow / hitting a `focus_terminal_input` call site with `RequireExisting` or `OpenIfNone`.
3. Confirming the "A command in this session is still running." toast no longer appears and the workflow either runs (if the session is actually free) or fails silently (`RequireExisting`) / opens a new pane (`OpenIfNeeded`) as documented on `TerminalSessionFallbackBehavior`.

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -7312,6 +7312,24 @@ impl TerminalView {
         true
     }
 
+    /// Returns true if the active block of this terminal is currently running a command
+    /// (writing to the PTY or actively executing). Used to gate the "still running" toast
+    /// — see #9100. We intentionally do NOT treat "input box hidden" as equivalent to
+    /// "command running": the input can be hidden for many reasons (pending shared
+    /// session, alt-screen application, SSH choice block, AI confirmation block, etc.)
+    /// that have nothing to do with a running command.
+    pub fn is_active_block_running(&self, _app: &AppContext) -> bool {
+        let model = self.model.lock();
+        if model.block_list().is_writing_or_executing_in_band_command() {
+            return true;
+        }
+        let active_block = model.block_list().active_block();
+        active_block.is_executing()
+            || (active_block.is_active_and_long_running()
+                && !active_block.is_agent_in_control()
+                && !active_block.is_agent_tagged_in())
+    }
+
     /// Give the agent control of the active long running command
     /// (which was started outside of a conversation).
     fn tag_agent_in(&mut self, ctx: &mut ViewContext<Self>) {
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -7320,14 +7320,19 @@ impl TerminalView {
     /// that have nothing to do with a running command.
     pub fn is_active_block_running(&self, _app: &AppContext) -> bool {
         let model = self.model.lock();
+        let active_block = model.block_list().active_block();
+        // Agent-controlled / tagged-in blocks must NOT trigger the "still running"
+        // toast, even when they are technically executing or writing in-band — the
+        // user is intentionally interacting with the agent and the input box is
+        // hidden by design. This exclusion must gate the executing / in-band checks
+        // below, otherwise `is_executing()` short-circuits to `true` first.
+        if active_block.is_agent_in_control() || active_block.is_agent_tagged_in() {
+            return false;
+        }
         if model.block_list().is_writing_or_executing_in_band_command() {
             return true;
         }
-        let active_block = model.block_list().active_block();
-        active_block.is_executing()
-            || (active_block.is_active_and_long_running()
-                && !active_block.is_agent_in_control()
-                && !active_block.is_agent_tagged_in())
+        active_block.is_executing() || active_block.is_active_and_long_running()
     }
 
     /// Give the agent control of the active long running command
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -15239,15 +15239,26 @@ impl Workspace {
         //      session as if it were focusable. Returning the busy session would
         //      reintroduce #9100 by letting callers act on a session they can't
         //      actually type into.
+        //
+        // For `OpenIfNone` specifically: if an existing session was found but its
+        // input box is not visible (e.g. alt-screen application, SSH choice block,
+        // AI confirmation block, init-project step, cloud-agent pre-first-exchange),
+        // we do NOT create a new pane — `OpenIfNone` only creates when there is no
+        // active session at all. We also must NOT return the unfocusable existing
+        // session: callers would proceed as if they had a focusable input and the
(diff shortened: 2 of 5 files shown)


zachbai:
@MaggieShan can you take a look at this?

app/src/terminal/view.rs
```diff
@@ -7312,6 +7312,29 @@ impl TerminalView {
...
+    /// that have nothing to do with a running command.
+    pub fn is_active_block_running(&self, _app: &AppContext) -> bool {
+        let model = self.model.lock();
+        let active_block = model.block_list().active_block();
+        // Agent-controlled / tagged-in blocks must NOT trigger the "still running"
+        // toast, even when they are technically executing or writing in-band — the
+        // user is intentionally interacting with the agent and the input box is
+        // hidden by design. This exclusion must gate the executing / in-band checks
+        // below, otherwise `is_executing()` short-circuits to `true` first.
+        if active_block.is_agent_in_control() || active_block.is_agent_tagged_in() {
```
MaggieShan, line 7329: I don't think it makes sense to be checking if the agent is in control or tagged in - this is only possible when there is a command actively running and it shouldn't be possible to run any other commands

MaggieShan: I believe there are other cases that should also return true/count towards an active running block like an alt-screen application and the ssh choice block (which is only shown while a user runs ssh <host>)

app/src/workspace/view.rs
```diff
@@ -15170,48 +15170,43 @@ impl Workspace {
...
+                    .update(ctx, |pane_group, ctx| pane_group.focus_active_session(ctx));
+                return Some(terminal_view_handle);
+            } else if is_env_var_block {
+                terminal_view_handle.update(ctx, |terminal_view, ctx| {
+                    terminal_view.cancel_env_var_block(ctx);
+                });
+                active_pane_group
+                    .update(ctx, |pane_group, ctx| pane_group.focus_active_session(ctx));
+                return Some(terminal_view_handle);
+            } else if Self::should_show_busy_toast(fallback_behavior, &terminal_view_handle, ctx) {
```
MaggieShan, line 15196: I agree it would be good to have an explicit check for if there is an active command running before showing the toast - but I think just checking `active_block.is_active_and_long_running()` is enough

MaggieShan (changes requested):
Thanks for taking a pass at fixing this! 

We'd like to see some tests/proof of manual testing to confirm this fixes the bug - personally I haven't been able to repro it so if you have any specific steps and screenshots that'd be great

PR #9877, opened 2026-05-02
Implement full-frame clear for active block for CLI Agents.

zachbai:
## Description
Fixes #9838 and fixes REMOTE-1423

Add a block-level flag for conditionally clearing output grid contents in-place on ansi::ClearMode::All (`esc[2j`) rather than clearing by 'scrolling' + appending to block-level scrollback.

In a traditional terminal the sequence clears the entire screen - in ghostty we see that frame-redrawing resizes for claude code, for instance, clear the entire terminal display (not appending to scrollback). This implements that behavior at a block-level for CLI agent TUIs that Warp is 'aware' of.

Preserves existing behavior for other cases to limit blast radius of the change - I think there are cases where the current block-level clear-by-scrolling behavior might be preferable/intended

CHANGELOG-BUG-FIX: Fixes issue with repeated TUI redraws for CLI agents on terminal pane resize.

PR #10743, opened 2026-05-12
fix: clear in place on CSI 2J at home position (#9181)

### LINKED ISSUE #9181: Terminal Screen Replicates Throughout Session
### Pre-submit Checks

### Describe the bug

When working in Claude Code CLI,terminal window will begin to fill with a visual replication of the session input and output, as the session goes on more and more of the content will be repeated and longer. WHen copying the full output for archive it is quite a nuisance. I thought this was isolated to one version of Warp but the last two updates have not fixed it. I believe this came with the move to the horizontal tabs, at least it was noticed after that update. 

### To reproduce

Honestly I don't need to do any steps to reproduce it has happened daily in every session short, medium and long.

### Expected behavior

One clean session, not partial replications one after the other growing as the session grows. 

I will be happy to paste the output, my concern is privacy and would like to know if what I present is accessible to anyone who access the bug list submissions. 

### Operating system (OS)

macOS

26.5.0

### Shell Version

I have no idea

### Current Warp version

v0.2026.04.23.19.42.preview_02, also occurs in the regular warp. I installed the preview hoping it was fixed

### Regression

### Recent working Warp date

Can't remember it has been at least two weeks like this.

### Additional context

_No response_

None

lonexreb:
Closes #9181

## Summary
- The "Terminal Screen Replicates Throughout Session" symptom (reproduced daily during Claude Code CLI sessions) is caused by Warp's primary-screen `CSI 2J` handler taking the `clear_viewport()` path, which scrolls the prior visible region into block-level scrollback before re-rendering.
- Inline-rendering CLI agents that do not opt into the alt screen (Claude Code in its default mode is the canonical example) issue `CSI [H` (cursor home) followed by `CSI 2J` many times per second to redraw their frame. Each redraw cycle duplicates the previous frame into the block's scrollback, producing the unbounded "replication" the user reports.
- This change adds a small heuristic in `ansi_handler::clear_screen` for `ansi::ClearMode::All`: when the cursor is already at the home position (`VisibleRow(0)`, col `0`) **and a prior at-home `CSI 2J` fired within the recent cadence window** (`TUI_REDRAW_CADENCE_WINDOW = 500ms`), and the alt-screen / `FullGridClearBehavior::Clear` paths are not engaged, clear the visible rows in place via the existing `clear_visible_rows_in_place` helper instead of `clear_viewport()`.
- The legacy scroll-into-scrollback behavior is preserved for both non-home `CSI 2J` and for the *first* at-home `CSI 2J` in a long while — the latter case is what the shell `clear` command produces under `xterm-256color` terminfo, which emits `CSI [H` followed by `CSI [2J`.

## Why this is the right shape of fix
- It is symmetric with PR #9877 (`Implement full-frame clear for active block for CLI Agents.`), which solved the same family of bug for the **resize** code path once Warp had already detected the CLI agent. This change covers the **redraw** code path *before* detection registers (or when the agent is not detected at all — e.g. third-party agents or `claude` invoked without the Warp plugin), which matches both the symptom timeline and the "Yes, this issue prevents me from using Warp daily" severity in the report.
- It matches `xterm` / `ghostty` / `iTerm2` semantics for `CSI 2J`: those terminals do not push the prior frame into scrollback on full-screen erase. Warp's behavior of conflating "user clear screen" with "TUI redraw" is the non-standard interpretation that caused the regression in the first place.
- Cursor-at-home alone is **not** a sufficient signal — `xterm-256color` terminfo's `clear` capability is `CSI [H` then `CSI [2J`, so the cursor is at home when `2J` fires for the shell `clear` command too. The cadence check is what cleanly disambiguates the two cases: inline-rendering CLI agents emit many at-home `CSI 2J` per second, while an interactive user runs `clear` at human speed. The first at-home `2J` in a long while is therefore treated as a one-off `clear` and preserves the visible region in scrollback; subsequent rapid ones are treated as TUI redraws and clear in place.

## Test plan
Added four regression tests in `app/src/terminal/model/grid/grid_handler_tests.rs`:
- [x] `test_clear_screen_all_primary_tui_redraw_at_home_clears_in_place` — the core regression: home + `CSI 2J` on the default primary-screen path (with active TUI redraw cadence) must leave `history_size() == 0`.
- [x] `test_clear_screen_all_primary_repeated_tui_redraws_do_not_duplicate` — drives 10 redraw cycles and asserts the scrollback does not grow.
- [x] `test_clear_screen_all_primary_non_home_cursor_preserves_scrollback` — the non-home `CSI 2J` path must continue to push the visible region into scrollback.
- [x] `test_shell_clear_command_preserves_scrollback_when_not_in_cli_agent_redraw_loop` — **new**: directly addresses the bot review concern. Reproduces the xterm-256color shell `clear` byte sequence (`CSI [H` followed by `CSI [2J`) with no active TUI redraw cadence and asserts that scrollback is preserved (legacy behavior).

The existing tests `test_clear_screen_all_primary_preserves_visible_rows_in_history_by_default`, `test_clear_screen_all_primary_with_full_grid_clear_behavior_clears_in_place`, and `test_clear_screen_all_alt_screen_clears_in_place` continue to cover the unchanged paths and were not modified.

## Manual testing note
This worktree environment lacks the Xcode `metal` toolchain (`xcrun: error: unable to find utility "metal"`), so `cargo check -p warp --lib` and `cargo check -p warp_terminal` could not complete locally — both transitively depend on `warpui`'s Metal-shader build step. The change was validated by direct review of the surrounding `clear_screen` / `clear_visible_rows_in_place` / `clear_viewport` paths and the new unit tests above (which exercise only the `GridHandler` and have no GPU dependency). CI on the upstream repo will exercise the full crate-scoped build and test suite.

The cadence window of 500ms is chosen to be a generous upper bound on inline-CLI-agent frame intervals (Claude Code emits multiple redraws per second under any non-trivial streaming load) while leaving a wide margin above any plausible human-interactive `clear` cadence (a user mashing `clear` at the keyboard typically achieves ~3-5 invocations per second at most, and crucially does not chain them with no intervening prompt — but even back-to-back `clear` invocations within 500ms would simply behave like an in-place redraw, which is visually indistinguishable from the legacy scroll-into-scrollback path when the prior frame is already blank from the first `clear`).

### DIFF
--- a/app/src/terminal/model/grid/ansi_handler.rs
+++ b/app/src/terminal/model/grid/ansi_handler.rs
@@ -11,6 +11,7 @@ use std::collections::HashMap;
 use std::io;
 use std::ops::Range;
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 
 use base64::Engine as _;
 use bounded_vec_deque::BoundedVecDeque;
@@ -106,7 +107,25 @@ pub(super) struct State {
     /// `push` appends the new mode; `pop` truncates and restores
     /// the previous entry as the active mode.
     pub keyboard_mode_stack: BoundedVecDeque<KeyboardModes>,
+
+    /// Timestamp of the most recent at-home CSI 2J on the primary screen
+    /// (GH #9181). Used to disambiguate TUI redraw loops (rapid back-to-back
+    /// at-home full clears emitted by inline-rendering CLI agents such as
+    /// Claude Code) from one-off shell `clear` invocations. The shell
+    /// `clear` command also emits CSI [H followed by CSI [2J — meaning the
+    /// cursor is at home when 2J fires — so cursor-at-home alone is not
+    /// sufficient to tell them apart. The TUI redraw cadence (many frames
+    /// per second) is, however, easy to distinguish from a one-off clear.
+    pub last_full_clear_at_home: Option<Instant>,
 }
+
+/// Maximum gap between two at-home CSI 2J events on the primary screen
+/// that still qualifies as a "TUI redraw loop" rather than a one-off shell
+/// `clear`. Claude Code and similar inline CLI agents redraw multiple
+/// frames per second, so a few hundred milliseconds is a generous upper
+/// bound; an interactive user running `clear` repeatedly by hand will
+/// almost always exceed it.
+pub(super) const TUI_REDRAW_CADENCE_WINDOW: Duration = Duration::from_millis(500);
 
 impl State {
     pub fn new(
@@ -143,6 +162,7 @@ impl State {
             pane_size: size_info.pane_size_px(),
             keyboard_mode: KeyboardModes::NO_MODE,
             keyboard_mode_stack: BoundedVecDeque::new(super::KEYBOARD_MODE_STACK_MAX_DEPTH),
+            last_full_clear_at_home: None,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


peicodes:
Hi @lonexreb, please include a narrated screen recording of this fix

peicodes:
Closed since this is stale and no proof of manual testing was provided

PR #10756, opened 2026-05-12
[Quality-547] add bert_tiny_v2 model and update input classifier

evelyn-with-warp:
## Description

Add [retrained nld classifier](https://github.com/warpdotdev/ml_models/pull/2) to dogfood; 
Meanwhile changing model file storage from raw blobs to lfs for optimizing git efficiency

It could resolve the misfires reported in [quality-547](https://linear.app/warpdotdev/issue/QUALITY-547/laundry-list-of-misfires); For quality 543 on file paths misfires, some are still pending to resolve by improving heuristics.

## About switching to git lfs to store model binary files
| Scenario | Potential Impact | Failure Risk | Mitigation |
|---|---|---|---|
| Fresh clone with `git-lfs` installed | `git clone` (and subsequent `git pull`) automatically invokes the smudge filter, which fetches the real ONNX blobs from LFS in place of the pointer files. `rust-embed` embeds the actual model bytes at build time. | None — classifier loads normally. | No mitigation needed; this is the happy path. |
| Fresh clone without `git-lfs` installed | ONNX paths land as ~130-byte pointer stubs. `rust-embed` embeds the pointer file at build time; runtime tries to load it as ONNX. | `candle::InferenceRunner::new` fails and the classifier silently falls back to `HeuristicClassifier`. App still builds and runs — no compile-time break, but a behavioral regression. | `script/bootstrap` now refuses to proceed if `git-lfs` is missing and runs `git lfs install --local && git lfs pull` when present; `WARP.md` documents it. |
| Existing clone, user pulls this branch with `git-lfs` installed but never enabled in the repo | `git lfs install --local` hasn't been run, so the smudge filter is inactive. | The v1 file may stay correct (already a blob locally) but the new v2 file appears as a pointer. | Same bootstrap step covers it; users who skip bootstrap can run `git lfs install && git lfs pull` once. |
| Existing clone, user pulls without ever installing `git-lfs` | The v2 model lands as a pointer (silent classifier fallback). | Build succeeds, but classifier silently degrades to heuristic fallback. | Bootstrap requirement + `WARP.md` note + the runtime `log::warn!("Failed to load onnx classifier (v2): ...")` makes the failure greppable in logs. |
| Existing clone with `git-lfs` installed and enabled | `git pull` invokes the smudge filter automatically, fetching the v2 blob from LFS to replace the pointer. `rust-embed` picks up the real bytes on the next build. | None — classifier loads normally. | No mitigation needed; this is the happy path for existing clones. |
| Forks / branches based on pre-LFS commits | Cherry-picks or merges crossing the LFS boundary need `git-lfs` locally to render the binaries correctly. | Low — no history rewrite is needed since we're only renormalizing on this commit. | Old branches keep working with raw blobs because their `.gitattributes` doesn't yet have the LFS rule. |

### TL;DR — Risk Mitigation Summary

**1. Fresh clones (hard enforcement at bootstrap)**
The actual gate lives in `script/bootstrap`, which now (a) refuses to proceed if `git-lfs` is not installed, and (b) runs `git lfs install --local && git lfs pull` automatically when it is. The `WARP.md` note complements this by explaining the requirement and providing recovery commands for users who bypass bootstrap.

**2. Existing clones (tiered recovery)**
   - **Hard failure path:** If a build or runtime error surfaces (e.g. `candle::InferenceRunner::new` panics or the agent reports a missing model), the agent's remediation is to run `git lfs install --local && git lfs pull` once and retry.
   - **Silent fallback path:** If the build succeeds but the classifier silently regresses to `HeuristicClassifier`, we accept it as a non-blocking degradation. The `log::warn!("Failed to load onnx classifier (v2): ...")` line makes the condition greppable in logs, and the next bootstrap run will self-heal.

**3. CI (requires explicit opt-in to LFS)**
Update CI `with: lfs: true` to any job that compiles the app, runs classifier tests, or bundles release artifacts. Please see https://github.com/warpdotdev/warp/pull/10756/changes/3bf92118b8135ce817e0d90949e0a3a739c7d073

## Linked Issue

## Testing

For the CI changes on LFS, we see below logs in CI logs to verify CI does fetch raw binary model files
```
28217b297a * app/assets/windows/arm64/OpenConsole.pdb (30 MB)
6eca0f5ba1 * app/assets/windows/arm64/conpty.pdb (6.4 MB)
3cf69e7c9b * app/assets/windows/x64/OpenConsole.pdb (34 MB)
e9347e3fa6 * app/assets/windows/x64/conpty.pdb (6.5 MB)
05b2aaba0e * crates/input_classifier/models/onnx/bert_tiny_v1.onnx (18 MB)
1fd8afc92b * crates/input_classifier/models/onnx/bert_tiny_v2.onnx (18 MB)
```
[macos CI](https://github.com/warpdotdev/warp/actions/runs/25768466066/job/75686351916)
[Linux CI](https://github.com/warpdotdev/warp/actions/runs/25768466066/job/75686351926)
[Windows CI](https://github.com/warpdotdev/warp/actions/runs/25768466066/job/75686351937)
### Screenshots / Videos

https://www.loom.com/share/94f99078f8614b2bbc05eab7f8d1a7d1
in this video we tested on the misfires listed in quality-547
- left: warp dev
- right: warp local build on this change

Video for dogfood feature
https://www.loom.com/share/196ffbc863aa4b39894f4e821ab54d85
## Agent Mode


evelyn-with-warp:
The change we see in crates/input_classifier/models/onnx/bert_tiny_v1.onnx in github refers to updating raw blob to lfs pointer

.gitattributes
```diff
@@ -1,3 +1,3 @@
 *.pdb filter=lfs diff=lfs merge=lfs -text
-input_classifier/models/* filter=lfs diff=lfs merge=lfs -text
-input_classifier/models/**/*tokenizer.json linguist-generated=true
+crates/input_classifier/models/** filter=lfs diff=lfs merge=lfs -text
+crates/input_classifier/models/**/*tokenizer.json -filter -diff -merge text linguist-generated=true
```
evelyn-with-warp, line 3: this is probably the reason why previous onnx model was checked in as raw blob not lfs: the paths did not include `crates/` so those lines matched nothing

vorporeal: i'm pretty sure it originally did use LFS; this probably broke when we moved the crate from `./input_classifier` to `./crates/input_classifier`.  (i would be surprised if i failed to set this up properly when i originally did it.)

evelyn-with-warp: @vorporeal  does that mean git lfs was already enabled in CI? 
CI didn't fail so far
pls lmk if other tests could cross-verify

evelyn-with-warp: anyway , i didn't see lfs:true at every checkout, so probably worth adding/re-adding it

vorporeal: i believe we never needed lfs in CI because we weren't using the classifier in unit or integration tests (because it wasn't a default feature)

WARP.md
```diff
@@ -37,12 +37,15 @@ Environment variables:
 - `find . -name "*.wgsl" -exec wgslfmt --check {} +` - Check WGSL shader formatting
 
 ### Platform Setup
-- `./script/bootstrap` - Platform-specific setup (calls platform-specific bootstrap scripts)
+- `./script/bootstrap` - Platform-specific setup (calls platform-specific bootstrap scripts). Also verifies `git-lfs` is installed and runs `git lfs install --local && git lfs pull` so model binaries under `crates/input_classifier/models/**` are materialized in the working tree.
```
vorporeal: this is not a useful update to our top-level public documentation.  agents like to add extra context here - in the extreme, this lists every single thing that the bootstrap script does, which does not provide value to the end user reading our readme.

WARP.md
```diff
@@ -37,12 +37,15 @@ Environment variables:
...
 - `./script/bootstrap --install-common-skills` - Platform setup plus common agent skill installation from `skills-lock.json`.
 - `./script/install_common_skills --if-needed` - Install or refresh shared agent skills from the standard `npx skills` project lock.
 - `./script/install_cargo_build_deps` - Install Cargo build dependencies
 - `./script/install_cargo_test_deps` - Install Cargo test dependencies
 
+#### Git LFS
```
vorporeal: i also don't think we need this here - the bootstrap script will print out a notice with this exact same information, right?

.gitattributes
```diff
@@ -1,3 +1,3 @@
 *.pdb filter=lfs diff=lfs merge=lfs -text
-input_classifier/models/* filter=lfs diff=lfs merge=lfs -text
-input_classifier/models/**/*tokenizer.json linguist-generated=true
+crates/input_classifier/models/** filter=lfs diff=lfs merge=lfs -text
```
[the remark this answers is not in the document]
evelyn-with-warp: see the mitigation solution in PR description; for CI, we have updated to fetch from LFS as well (as seen in Tests section in PR description)

crates/input_classifier/src/onnx/mod.rs
```diff
@@ -22,9 +22,10 @@ use crate::{
 #[folder = "models/onnx"]
 struct Models;
```
vorporeal, line 26: we should be changing this so that we're only bundling one model into the binary instead of both.

you probably want to use `#[cfg_attr(feature = "...", include = "...")]`, to switch from unconditionally including everything in the models folder to using an allowlist

.github/actions/prepare_environment/action.yml
```diff
@@ -53,6 +53,24 @@ runs:
...
+
+    # Print a per-file listing of the LFS payloads so the CI log shows that
+    # the ONNX/fasttext model binaries (and Windows .pdb files) are real
+    # binaries rather than 133-byte pointer stubs. The `*` marker before
+    # each path means "data is present in the working tree".
+    - name: Verify LFS payloads
+      shell: bash
+      run: |
+        cd "${{ steps.repo-root.outputs.path }}"
+        git lfs ls-files --size
```
vorporeal: now that we've verified that things are working properly, we can remove this, yeah?  if you want to keep it, we should at least condition it on the action being executed in debug mode.

WARP.md
evelyn-with-warp: removed in https://github.com/warpdotdev/warp/pull/10756/changes/d9608a34b1f10c4c2918ed5ec31c1bf769e4f68d
the initiative was giving agent some hint when there's lfs related failure, but it's a fair point that agent might able to get this from session context as well

.github/actions/prepare_environment/action.yml
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10756/changes/617cceb22ddab14274522e858fbcc69b6fdc5864

crates/input_classifier/src/onnx/mod.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10756/changes/e849231dff359260472906f5e61312dd3570b281

crates/input_classifier/Cargo.toml
```diff
@@ -24,6 +24,9 @@ onnx_ort = [
     "dep:parking_lot",
 ]
 onnx = ["dep:rust-embed", "dep:tokenizers"]
+# Selects which BERT-tiny ONNX model bytes are embedded via rust-embed; enable exactly one.
+nld_classifier_v1 = ["onnx"]
+nld_classifier_v2 = ["onnx"]
```
szgupta: Is this actually necessary?

szgupta (approved):
LGTM but would like @vorporeal's stamp on the LFS changes

crates/input_classifier/Cargo.toml
szgupta: ignore me

app/Cargo.toml
```diff
@@ -659,8 +659,14 @@ grep_tool = []
 # For the most part, including GUI code in headless builds is fine; the main use case is
 # for things that pull in external dependencies.
 gui = ["voice_input"]
-nld_classifier_v1 = ["input_classifier/onnx_candle"]
-nld_classifier_v2 = ["input_classifier/onnx_candle"]
+nld_classifier_v1 = [
+    "input_classifier/onnx_candle",
```
szgupta: nit: can we have the `input_classifier/*` flags transitively enable the relevant onnx flags? It's really an implementation detail of what `vN` does. So in the overall app cargo.toml, it would just be
```
nld_classifier_v1 = [input_classifier/nld_classifier_v1]
```

script/bootstrap
```diff
@@ -80,6 +80,29 @@ maybe_install_common_skills() {
...
 }
 
+# Ensure Git LFS is installed and initialized for this checkout. The
+# `crates/input_classifier/models/**` binaries (ONNX/fasttext models) are
+# tracked via LFS; without `git-lfs` those paths will be ~130-byte pointer
+# stubs and the app will silently fall back to the heuristic classifier.
+ensure_git_lfs() {
+  if ! command -v git-lfs >/dev/null 2>&1; then
+    echo
+    echo "warning: git-lfs is not installed. Install it before continuing:"
```
vorporeal: do we actually want to force all users (including external ones) to install git-lfs?  if so, we might not want this to say "warning:" - warnings imply execution can continue without addressing them.

script/bootstrap
```diff
@@ -80,6 +80,29 @@ maybe_install_common_skills() {
   fi
 }
 
+# Ensure Git LFS is installed and initialized for this checkout. The
+# `crates/input_classifier/models/**` binaries (ONNX/fasttext models) are
+# tracked via LFS; without `git-lfs` those paths will be ~130-byte pointer
+# stubs and the app will silently fall back to the heuristic classifier.
```
vorporeal: i think it's probably enough to say that the repository requires use of LFS - the extra details here will either 1) get stale or 2) encourage agents to turn this into an exhaustive list of things that use LFS, neither of which are ideal.

vorporeal (approved):
lgtm for lfs stuff

app/Cargo.toml
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10756/changes/e00f383f7741003aabec41d2ebd76db019ec655b

script/bootstrap
evelyn-with-warp: @vorporeal my thought was not forcing external user for this change , if their interests/development WERE independent with the lfs files. In other words, if users get silent fallback without interrupting their work, they don't have to spend time on LFS set up. 
I'm okay with forcing user to set up LFS, if this is most beneficial for oss repo going forward 
WDYT?

script/bootstrap
evelyn-with-warp: agreed! 
https://github.com/warpdotdev/warp/pull/10756/changes/e3bbd81bd9fd52875d9fd0e2367eed3cd7666455

Issue #10767, opened 2026-05-12
[Windows] Auto-update fails with "the system cannot find the file specified"


acarl005:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When clicking the update button in Warp on Windows to initiate the auto-update flow, sometimes it fails with a message "". It often succeeds on retry.

### To reproduce

No definitive repro.

### Expected behavior

Clean auto-update with no errors

### Screenshots, videos, and logs

<img width="300" height="146" alt="Image" src="https://github.com/user-attachments/assets/1f7fce4a-8e49-4615-8eb9-c8c4ef66250a" />

### Operating system (OS)

Windows

### Operating system and version

11

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

Issue #10044, opened 2026-05-04
[Windows] Auto-update failures


peicodes:
Related GitHub issue [https://github.com/warpdotdev/Warp/issues/8339](https://github.com/warpdotdev/Warp/issues/8339)

Have seen a few reports about this, including from a VIP at Microsoft.

---
Linear: [APP-3082](https://linear.app/warpdotdev/issue/APP-3082/windows-auto-update-failures)

acarl005:
Related: #10767

PR #10768, opened 2026-05-12
avoid dropping installer file handle to prevent race

acarl005:
## Description

This PR avoids deletion of the inno setup installer during auto-update.

The linked issue shows that in some cases, auto-update fails with a "file not found" error dialog. This is because of a race condition between dropping the `TempPath`, an RAII wrapper around the installer exe, and the installer itself actually starting up. It may have seemed like the installer should have started up sequentially first by calling a blocking call to `spawn`. This is not so b/c the exe actually spawns another process and exits immediately. To resolve this race, we simply do not delete the installer. There is no need, as the installer lives in `%TEMP%` anyways. 

## Linked Issue

This PR addresses #10767. Master issue is #10044.

## Testing

Untested due to the lack of a robust reproduction case.

### DIFF
--- a/app/src/autoupdate/windows.rs
+++ b/app/src/autoupdate/windows.rs
@@ -259,14 +259,6 @@ pub(super) fn relaunch() -> Result<()> {
         ])
         .spawn()?;
 
-    // DEV ONLY: Sleep after spawning the installer so this process is still alive
-    // when Inno Setup tries to overwrite files. This reliably reproduces the
-    // auto-update race condition (APP-3702) for testing.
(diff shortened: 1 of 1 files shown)


app/src/autoupdate/windows.rs
lucieleblanc: Is this code at all related to the issue? Should we clean this up?

lucieleblanc (approved):
This fix makes sense to me!

Left a question about another race condition cited in the code.

app/src/autoupdate/windows.rs
acarl005: oh yes! ready to be deleted

PR #10774, opened 2026-05-12
feat: default environment/host selection and deferred auto-launch for orchestration config UI

cephalonaut:
## Description

Two changes to orchestration config handling:

**1. Default environment and host selection** — When the orchestration config UI opens in Remote mode with no environment selected, auto-fill from the user's last-selected or most recently used environment (matching `/cloud-agent` behavior). Worker host defaults to "warp". Selections are persisted to settings for cross-flow consistency.

**2. Deferred auto-launch with unconditional config override** — Auto-launch evaluation is deferred to stream completion so the full request (including plan_id and agent_run_configs) is available. When an approved plan config exists, its model/harness/execution_mode unconditionally override the LLM's run_agents call (preserving only computer_use_enabled). This also applies to the autonomous executor path for CLI-driver agents.

[Agent conversation](https://staging.warp.dev/conversation/ba65e2ae-16d1-458e-8de7-c599e0b83c24) | [Plan](https://staging.warp.dev/drive/notebook/qS8CS7tIdHkdE5EKkHFeqi)

## Testing

- [x] New unit tests for `override_from_approved_config` (6 tests covering all override/preserve scenarios)

## Agent Mode

### DIFF
--- a/app/src/ai/document/orchestration_config_block.rs
+++ b/app/src/ai/document/orchestration_config_block.rs
@@ -118,6 +118,10 @@ pub struct OrchestrationConfigBlockView {
     /// UI-only per-harness model memory so switching harnesses preserves
     /// the user's previous model selection for each harness.
     saved_model_per_harness: HashMap<String, String>,
+    /// Suppresses self-triggered refresh when `apply_field_change`
+    /// saves the config and the resulting event re-enters
+    /// `refresh_from_model`.
+    suppress_refresh: bool,
 }
 
 impl OrchestrationConfigBlockView {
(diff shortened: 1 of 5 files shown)


app/src/ai/document/orchestration_config_block.rs
```diff
@@ -263,6 +272,26 @@ impl OrchestrationConfigBlockView {
...
+            let needs_host = worker_host.is_empty();
+            let needs_env = environment_id.is_empty();
+            if needs_host {
+                self.edit_state
+                    .set_worker_host(oc::ORCHESTRATION_WARP_WORKER_HOST.to_string());
+            }
+            if needs_env {
+                if let Some(default_env) = oc::resolve_default_environment_id(ctx) {
+                    self.edit_state.set_environment_id(default_env);
+                }
```
advait-m: ^ following Oz comment, I think adding something like this should address it 
```
        if filled_defaults && self.is_approved {
            self.apply_field_change(ctx);
        }
```

advait-m (approved):

app/src/ai/document/orchestration_config_block.rs
cephalonaut: Done

PR #10777, opened 2026-05-12
Add pinning to orchestration pill bar

advait-m:
Loom: https://www.loom.com/share/240fef8cf25041f8ae312088d79e29ba
Fixes https://linear.app/warpdotdev/issue/QUALITY-672/add-pill-pinning-from-orchestration-pill-bar-mocks

## Description

Adds pinning to the orchestration pill bar so frequently-used child agents stay anchored to the leading section of the bar instead of getting lost in a long row of pills. Pin state is shared across panes and persists across app restarts.

- New `Icon::Pin` (outline) + `Icon::PinFilled` (solid) variants, bundled from Figma SVGs under `app/assets/bundled/svg/`.
- Each child pill shows an outline pin on hover when unpinned (click to pin) and a solid pin on hover when pinned (click to unpin). At rest the avatar disc is shown — pin state is communicated by the pill's position to the left of a vertical divider.
- Bar partition: orchestrator → pinned children (spawn order) → divider (only when both sides are non-empty) → unpinned children (spawn order).
- Pin button has its own dedicated `MouseStateHandle` per pill, independent of the pill body and the 3-dot overflow button.

## Technical Design

**Cross-pane state via a singleton model.** Pin state lives in a new `OrchestrationPinModel` singleton, not on `OrchestrationPillBar` (which is per-`TerminalView`). Each pill bar reads from the singleton and subscribes to `OrchestrationPinEvent::PinSetChanged`, so toggling a pin in one pane immediately re-renders every other pane's bar. The singleton also subscribes to `BlocklistAIHistoryModel` events (`RemoveConversation` / `DeletedConversation`) so deleted conversations are pruned from the pin set globally — keeping that cleanup in one place avoids each pill bar racing to clobber sibling panes' pin sets. A per-view `HashSet<AIConversationId>` was considered first but rejected because it would have required broadcasting toggles across panes manually and would leak across `TerminalView` lifecycle boundaries.

**Persistence piggybacks on the existing conversation write path.** Rather than introducing a separate pin table or settings entry, pinned state is added as `pinned: bool` on the existing `AgentConversationData` (which is the JSON blob already persisted per-conversation in SQLite). Toggling a pin calls `BlocklistAIHistoryModel::set_conversation_pinned`, which updates the in-memory `AIConversation.pinned` flag and reuses `write_updated_conversation_state` to emit an `UpdateMultiAgentConversation` event. The singleton is seeded at app boot from the same `multi_agent_conversations` vec already loaded for `BlocklistAIHistoryModel`, so there's no extra DB read. The field uses `#[serde(default, skip_serializing_if = "is_false")]` so existing rows deserialize cleanly and unpinned conversations don't bloat the persisted JSON.

**Click-handler scoping to avoid hover-delay races.** The first pass wrapped the pill's leading element (avatar / pin glyph) in a `Hoverable` that always carried the `TogglePin` click handler. That stole clicks during the 300ms `with_hover_in_delay` window — a user clicking the avatar to navigate could land on the toggle instead. The current code only wraps the inner element in `Hoverable` (and only attaches the toggle handler) when `show_pin_glyph` is true; otherwise clicks bubble to the outer pill's navigate handler. This keeps the navigate-vs-toggle contract tied 1:1 to what the user actually sees rendered.

**Logout cleanup.** `OrchestrationPinModel::reset()` is invoked from `log_out` alongside the other `*::reset()` calls so a subsequent user doesn't inherit the previous account's pins. The persisted per-conversation flags are wiped by the existing sqlite reset that runs alongside logout, so this just clears the in-memory mirror. `set_conversation_pinned` also `log::warn!`s when the conversation isn't loaded so silently-dropped writes are visible in logs rather than disappearing.

Plan: https://staging.warp.dev/drive/notebook/DavRjsSq2TUxXeF8PBQsFc

## Testing

Unit tests:
- `orchestration_pin_model_tests.rs` covers `toggle_pin_in_set` (flip-on-each-call, only-affects-target-id).
- E2E persistence test `toggle_pin_persists_pinned_state_to_sqlite_event` wires up settings + a mock `GlobalResourceHandles` channel, restores a conversation, toggles a pin, and asserts (a) the in-memory `AIConversation.pinned` flips, (b) an `UpdateMultiAgentConversation` event arrives with `pinned: true`, and (c) a follow-up unpin emits a second event with `pinned: false`.
- `model.rs` round-trip tests cover `AgentConversationData.pinned` serde defaulting and skip-if-false behavior.
- Existing `orchestration_pill_bar_tests.rs` descendant-walking + navigation-action tests still pass.

`cargo fmt`, `cargo build -p warp`, `cargo nextest run -p warp orchestration_pin_model`, and `cargo clippy -p warp --tests --all-features` all clean on this branch.


app/assets/bundled/svg/pin-01.svg
```diff
@@ -0,0 +1,4 @@
...
...43719 10.6752C4.45019 10.6297 4.45769 10.0282 4.45769 9.04873V7.49473L5.74269 7.48823C6.99269 7.48123 7.03219 7.47973 7.18419 7.43723C7.39719 7.37773 7.52119 7.30623 7.64819 7.16873C7.87269 6.92623 7.96619 6.58673 7.88869 6.29523C7.82069 6.03923 7.74069 5.92223 6.99519 4.99023L6.45769 4.31773V3.70573V3.09423L6.77519 2.29773C7.13219 1.40373 7.19219 1.22923 7.20969 1.03673C7.23419 0.76223 7.12019 0.47423 6.90869 0.27723C6.77469 0.15223 6.67069 0.0937297 6.48769 0.0392298C6.37619 0.00622975 6.20569 0.00322971 4.03769 0.000229715C2.75619 -0.00127029 1.65069 0.00472975 1.58119 0.0137297ZM6.18769 1.01273C6.19469 1.02423 6.04519 1.42223 5.85519 1.89773C5.66569 2.37373 5.49869 2.81223 5.48419 2.87273C5.45069 3.01523 5.44669 4.35873 5.47969 4.51673C5.52319 4.72573 5.60469 4.85023 6.11069 5.48323C6.75119 6.28423 6.87969 6.45323 6.86619 6.47473C6.85169 6.49873 1.06369 6.49873 1.04869 6.47473C1.03619 6.45373 1.19219 6.24973 1.84469 5.43323C2.30969 4.85173 2.39269 4.72273 2.43569 4.51673C2.46869 4.35873 2.46469 3.01523 2.43119 2.87273C2.41669 2.81223 2.24969 2.37373 2.06019 1.89773C1.87019 1.42223 1.72069 1.02423 1.72769 1.01273C1.74469 0.98473 6.17069 0.98473 6.18769 1.01273Z" fill="#121212"/>
```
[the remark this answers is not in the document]
advait-m: hmmm, light theme looks fine

image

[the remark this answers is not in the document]
advait-m: we tint it at paint time so this isn't relevant

app/src/ai/agent/conversation.rs
```diff
@@ -237,6 +237,10 @@ pub struct AIConversation {
     /// `OrchestrationConfigSnapshot` messages in the conversation's task list.
     /// Keyed by `plan_id`; snapshots with empty `plan_id` are ignored.
     orchestration_configs: HashMap<String, (OrchestrationConfig, OrchestrationConfigStatus)>,
+
+    /// Whether the user has pinned this child agent in the orchestration
+    /// pill bar. Persisted via `AgentConversationData.pinned`.
+    pinned: bool,
```
cephalonaut, line 261: There might be some structural grouping we could do of child agent specific properties: parent_agent_id, pinned, is_remote_child, parent_conversation_id, orchestration_harness_type evidently.

cephalonaut (approved):
Looking good!

app/src/ai/agent/conversation.rs
advait-m: agree, maybe smth like `ChildAgentState` or similar - tho will be a big surface area refactor, so can follow up on that

advait-m: added TODO

PR #10779, opened 2026-05-12
Hide third-party transcript vehicle conversations from the management view.

liliwilson:
## Description

When restoring third-party conversation transcripts, we create a vehicle `AIConversation` that we use to enter the agent view.

We used to be rendering these `AIConversation` entries within management surfaces throughout the app, resulting in ghost entries in the management view that got created every time you tried to restore a CLI transcript.

We fix this here by tracking when an `AIConversation` is being used as a CLI vehicle (which we set based on the `AgentEntryOrigin`) and add these to the existing `should_exclude_from_navigation`.

We didn't see this for live sessions, even though they also use vehicle conversations, because we already check in `should_exclude_from_navigation` whether we are viewing a shared session.

## Testing

Tested locally: https://www.loom.com/share/20b9e025ab9c49f49c21a929ab702d25

## Agent Mode

### DIFF
--- a/app/src/ai/agent/conversation.rs
+++ b/app/src/ai/agent/conversation.rs
@@ -259,7 +259,7 @@ pub(crate) fn artifact_from_fork_proto(
 }
 
 impl AIConversation {
-    pub fn new(is_viewing_shared_session: bool) -> Self {
+    pub fn new(is_viewing_shared_session: bool, is_cli_agent_transcript: bool) -> Self {
         let root_task = Task::new_optimistic_root();
         Self {
             id: AIConversationId::new(),
@@ -267,7 +267,7 @@ impl AIConversation {
             optimistic_cli_subagent_subtask_id: None,
             code_review: None,
             is_viewing_shared_session,
-            is_cli_agent_transcript: false,
+            is_cli_agent_transcript,
             todo_lists: vec![],
             status: ConversationStatus::InProgress,
             status_error_message: None,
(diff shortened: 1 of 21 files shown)


app/src/terminal/view/agent_view.rs
```diff
@@ -93,19 +93,17 @@ impl TerminalView {
         ctx: &mut ViewContext<Self>,
     ) -> Option<AIConversationId> {
         let origin = AgentViewEntryOrigin::ThirdPartyCloudAgent;
-        let conversation_id = BlocklistAIHistoryModel::handle(ctx).update(ctx, |history, ctx| {
```
liliwilson, line 96: We can create the convo within the agent view, this movement is supposed to be a noop change

app/src/ai/agent/conversation.rs
```diff
@@ -511,6 +518,14 @@ impl AIConversation {
         self.is_viewing_shared_session = is_viewing_shared_session;
     }
 
+    pub fn is_cli_agent_transcript(&self) -> bool {
+        self.is_cli_agent_transcript
+    }
```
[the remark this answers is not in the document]
liliwilson: fixed

app/src/ai/blocklist/agent_view/controller.rs
```diff
@@ -767,12 +767,18 @@ impl AgentViewController {
...
+                let id = history_model.start_new_conversation(
                     self.terminal_view_id,
                     false,
                     matches!(origin, AgentViewEntryOrigin::CloudAgent),
                     ctx,
-                )
+                );
+                if matches!(origin, AgentViewEntryOrigin::ThirdPartyCloudAgent) {
+                    if let Some(conversation) = history_model.conversation_mut(&id) {
+                        conversation.mark_as_cli_agent_transcript();
```
zachbai: lets just make this a constructor arg so u never need to reason about this state being mutable (which the API implies)

zachbai (approved):

app/src/ai/blocklist/agent_view/controller.rs
liliwilson: Good call

PR #10782, opened 2026-05-13
Version aware daemon socket

kevinyang372:
## Description
This creates a version aware daemon socket. There are two main components:
1. Now socket connections are versioned. So given proxy version X, we will connect to server-X.pid and server-X.socket. This allows proxy to enforce it's connecting to a daemon with the SAME version
2. After connection, proxy also cleans up any old socket and pid file that doesn't match with its version

## Testing
I tested the versioned socket creation and clean up separately

Versioned socket:
1. Deploy the server with dev channel and rename it to a versioned binary `v0.2026.05.12.09.10.dev_00`
2. Run a local binary with `GIT_RELEASE_TAG=v0.2026.05.12.09.10.dev_00 cargo run --bin dev`
3. Confirm the connection is successful and the right socket / pid file is created
Screenshot 2026-05-13 at 11 02 59 AM

Clean up:
1. Rename the existing socket/pid file from the last step to `server-old-X.pid` and `server-old-X.socket`
2. Run a local binary with `GIT_RELEASE_TAG=v0.2026.05.12.09.10.dev_00 cargo run --bin dev`
3. Confirm the connection is successful, old socket/pid is removed, new socket/pid file is created

### DIFF
--- a/app/src/remote_server/unix/proxy.rs
+++ b/app/src/remote_server/unix/proxy.rs
@@ -41,12 +41,16 @@ fn daemon_dir(identity_key: &str) -> PathBuf {
     PathBuf::from(expanded)
 }
 
-/// Scans the identity-key daemon directory and cleans up socket/PID files
+/// Scans the identity-key daemon directory and removes socket/PID files
 /// from previous daemon versions.
 ///
-/// For each old PID file found, sends `SIGTERM` to the old daemon (if still
-/// alive) so it shuts down gracefully, then removes both the PID and socket
-/// files. Errors are logged but do not prevent the proxy from proceeding.
+/// Old daemons are **not** killed — they may still be serving active
+/// connections from an older Warp client. Removing their socket file
+/// prevents new proxies from accidentally connecting to them, and the
+/// daemon's built-in grace timer (10 min with no connections) will shut
+/// it down naturally after the last client disconnects.
+///
+/// Errors are logged but do not prevent the proxy from proceeding.
 fn cleanup_old_versions(identity_key: &str) {
     let dir = daemon_dir(identity_key);
     let current_socket = setup::daemon_socket_name();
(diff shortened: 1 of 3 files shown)


kevinyang372:
* **#10782** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10782?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10782?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/remote_server/src/setup.rs
```diff
@@ -343,6 +343,32 @@ pub fn remote_server_daemon_data_dir(identity_key: &str) -> String {
     format!("{}/data", remote_server_daemon_dir(identity_key))
 }
 
+/// Returns the daemon socket filename, versioned on release channels.
+///
+/// - Release channels: `server-{version}.sock`
+/// - Local / Oss:      `server.sock`
+pub fn daemon_socket_name() -> String {
+    match ChannelState::channel() {
+        Channel::Local | Channel::Oss => "server.sock".to_string(),
```
alokedesai: If the daemon has a version set for Local / Oss,  I think we should use it. Otherwise you can't really test the versioned flow locally

alokedesai (approved):
noice

app/src/remote_server/unix/proxy.rs
```diff
@@ -19,18 +19,80 @@ use std::time::Duration;
...
+        // Match old PID files: server*.pid that aren't the current version.
+        if name_str.ends_with(".pid") && name_str.starts_with("server") && name_str != current_pid {
+            let old_pid_path = entry.path();
+            // Try to SIGTERM the old daemon.
+            if let Ok(contents) = std::fs::read_to_string(&old_pid_path) {
+                if let Ok(pid) = contents.trim().parse::<libc::pid_t>() {
+                    // SAFETY: sending SIGTERM is safe — it requests graceful
+                    // shutdown. If the process doesn't exist, kill returns -1
+                    // with ESRCH which we ignore.
+                    let ret = unsafe { libc::kill(pid, libc::SIGTERM) };
```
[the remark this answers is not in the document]
kevinyang372: This is good feedback. I think we could avoid eagerly SIGTERMing the connection and just remove the socket / pid file to be safe. This will prevent new connections to that daemon but not interrupt its active connections. The daemon will self-shutdown when its connection drops

crates/remote_server/src/setup.rs
kevinyang372: Yeah agreed. Done

PR #10794, opened 2026-05-13
Fix QUALITY-668 orchestration child restore persistence

kjankov:
## Description
Persist orchestration child conversation metadata as soon as the child is created or assigned server identity so Warp can restore child conversations, participant names, and the pill UI after restart. This also adds a shared test persistence setup helper for unit tests that now exercise those early writes.

## Linked Issue
- Linear: https://linear.app/warpdotdev/issue/QUALITY-668/cloud-subagents-lost-maybe-after-warp-restart

## Testing
- Targeted nextest rerun of the 11 presubmit regressions
- `cargo fmt --all`
- `./script/presubmit`
- Not manually tested; validated through automated coverage and full presubmit

## Agent Mode

## Artifacts
- Conversation: https://app.warp.dev/conversation/4853ad4c-5fff-4a6c-a63c-ef592385b52d
- Plan: https://app.warp.dev/drive/notebook/aOGYEB6jISDh8Q5WNXBquj

CHANGELOG-BUG-FIX: Preserve orchestration child agents across restart so agent names and child pills restore correctly.


app/src/ai/agent/task.rs
```diff
@@ -568,6 +568,21 @@ impl Task {
         self.try_get_source().ok()
     }
 
+    pub(super) fn source_for_persistence(&self) -> Option<api::Task> {
```
cephalonaut, line 589: Hey - not positive whether it's important, but agent mode came up with this suggestion:

Task::source_for_persistence changes the meaning of TaskImpl::Server

The new branch for Optimistic::Root synthesizes an empty api::Task so that the conversation has a root task to persist. That fixes the restore path, but on the way back in, Task::new_restored_root constructs the task as TaskImpl::Server — so a task that was Optimistic before the restart comes back as Server even though the server never confirmed it.

The user-visible consequence is that into_server_created_task becomes unreachable for that task: it returns UpgradeOptimisticTaskError::UnexpectedUpgrade on any TaskImpl::Server. If the server later replies for a conversation that was persisted in the optimistic window and then restored, the local UUID stays installed forever and the server's real task id never gets adopted locally.

For QUALITY-668 the conversation is mostly identified by run_id / server_conversation_token on AgentConversationData, so the shell still restores correctly. But the invariant "TaskImpl::Server ⇒ server-confirmed" is no longer strictly true, and that's surprising for anyone reading the enum or into_server_created_task later.

Two ways to address:

1. Preserve the variant on round-trip. Add root_task_is_optimistic: Option<bool> to AgentConversationData (forward/backward compatible since it's a serde-json blob, not a schema column), plumb it through write_updated_conversation_state and AIConversation::new_restored, and add a Task::new_restored_optimistic_root constructor that rebuilds the original TaskImpl::Optimistic(Task::Root) with the persisted local UUID. ~Tens of LOC + a round-trip test.
2. Document the invariant change. If the trade-off is acceptable, at minimum add a doc comment on source_for_persistence explaining that Optimistic::Root is intentionally synthesized as a placeholder and that the restored task will be Server-typed (no longer upgradable), plus a note on why Optimistic::CLIAgent returns None.

My preference is (1) — it cleanly removes the landmine — but (2) is acceptable if you want to keep this PR small.

cephalonaut (approved):

app/src/ai/agent/task.rs
kjankov: Preserved the optimistic root variant on round-trip instead of restoring it as `TaskImpl::Server`. The follow-up adds `root_task_is_optimistic` to `AgentConversationData`, restores via `Task::new_restored_optimistic_root`, and covers both the restore path and the later upgrade to the server task id in tests.

Issue #10269, opened 2026-05-06
New worktree config menu item is clipped


moirahuang:
## Summary
The unified new-session/tab config menu is too narrow for the `New worktree config` entry, causing the label to be clipped in the menu.

## Problem
The horizontal width of the tab config/new-session menu does not accommodate the full `New worktree config` label. This makes the item look visually truncated even though it is a primary creation action in the menu.

## Expected behavior
The menu should be wide enough to show `New worktree config` without clipping or truncating the label.

## Actual behavior
`New worktree config` is clipped at the right edge of the menu in the visible UI.

## Reproduction steps or desired workflow
1. Open the new-session/tab config menu from the tab bar `+`/chevron menu.
2. Look at the `New worktree config` entry near the bottom of the menu.
3. Observe that the label is clipped because the menu width is too narrow.

## Artifacts

<img width="213" height="273" alt="Image" src="https://github.com/user-attachments/assets/84a133db-cd00-4e93-8a62-5e1ccb08a84f" />

## Warp version
v0.2026.05.05.21.36.dev_00

## Operating system
macOS 26.4.1 (25E253)

## Possible source references
- `app/src/workspace/view.rs:6057`: `unified_new_session_menu_items` builds the unified new-session menu used by the tab bar chevron and vertical tab bar `+` button.
- `app/src/workspace/view.rs:6215`: the `New worktree config` submenu item is added to the menu.
- `specs/APP-3886/PRODUCT.md:36`: product spec lists `New worktree config` as a top-level menu item in the tab config menu structure.

moirahuang:
My proposed implementation is to increase the width of the tab config menu just a bit, probably similar to the width of the vertical tab config menu

dhruvac29:
I can take this one. I’ll follow the proposed direction and update the unified tab config/new-session menu width to match the vertical tab config menu width so `New worktree config` is no longer clipped.

PR #10542, opened 2026-05-09
Fix New worktree config clipping in tab config menu

SagarSDagdu:
## Description
Fixes the clipping of `New worktree config` in the unified new-session/tab config menu opened from the horizontal tab-bar chevron.

The chevron variant used `MENU_DEFAULT_WIDTH` (186px) while the vertical-tab `+` variant already used 268px. Both branches render the same `unified_new_session_menu_items`, and 186px is too narrow for the longer label. Both variants now use 268px.

## Linked Issue
Fixes #10269
- [x] The linked issue is labeled `ready-to-implement`.
- [x] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

### Before
Screenshot 2026-05-09 at 8 40 24 AM

### After

Screenshot 2026-05-09 at 8 36 16 AM

## Testing
- `cargo check -p warp` passes.
- Local debug build (`cargo run -p warp --bin warp-oss`) launches; attached before/after screenshots of the chevron menu in horizontal-tab mode with `TabConfigs` enabled.

## Agent Mode
- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: Fix `New worktree config` clipping in the tab-bar new-session menu.

PR #10799, opened 2026-05-13
refactor(workspace): clarify new-session submenu width constant + apply consistently

lonexreb:
## Description

`Menu::with_width()` on `Menu<A>` in `app/src/menu.rs:2122` sets the **`submenu_width`** field, not the main menu's own rendered width. The constant `NEW_SESSION_MENU_WIDTH` in `app/src/workspace/view.rs:1805` was misleadingly named — setting it does not affect the visible width of the dropdown shown when clicking the tab bar `+` chevron; only the width of nested submenus (e.g. the `Search repos` panel that opens off of `New worktree config`).

This PR:

1. Renames `NEW_SESSION_MENU_WIDTH` → `NEW_SESSION_SUBMENU_WIDTH` to accurately describe what the constant drives.
2. Expands the surrounding comment to call out the gotcha (`Menu::with_width` does not control the main menu's own width) and references the architectural follow-up (adding a `min_width` to `Menu` and plumbing it through `SubMenu::render` / the wrapping `ConstrainedBox`).
3. Applies `.with_width(NEW_SESSION_SUBMENU_WIDTH)` to the non-`ShellSelector` else branch as well, so submenus are 300 px wide regardless of feature flag state. Previously the else branch fell back to `DEFAULT_WIDTH = 186` from `Menu::new()`, producing a noticeably narrower submenu in OSS builds.

## Linked Issue

Refs #10269 — this PR explicitly does **not** close that issue. The visible `New worktree config` label clipping reported there lives in the main menu's own (content-sized) width, not the submenu width. Fixing it requires the Menu API enhancement described in the updated comment. I started down that path and confirmed the layout pipeline (`ConstrainedBox::with_width` + `SubMenu::render`'s outer wrapping) doesn't currently expose a hook for the main menu to honor a minimum width — that's a larger architectural change worth its own PR.

## Testing

- `cargo check --bin warp-oss --features gui` — clean
- `cargo test -p warp --lib test_unified_new_session_menu` — passes (existing tests for menu item presence/order still green; this change touches only the width constant and its application, not menu items themselves)
- Manual GUI verification on macOS: built locally, opened the tab bar `+` chevron, opened the `Search repos` submenu off `New worktree config`. Submenu width is now consistent at 300 px in both the `ShellSelector`-on and `ShellSelector`-off paths.

### Why no new unit test

The rename is a no-op for tests (the constant's value is unchanged for the `ShellSelector`-on path). The else-branch width application is a UI dimension change with the same testability constraints as the surrounding constants — there's no clean unit-test surface for "submenu renders at expected width" without standing up the layout pipeline. Existing `test_unified_new_session_menu_uses_new_worktree_config_label_and_order` / `test_unified_new_session_menu_includes_reopen_closed_session` continue to cover the menu's structural invariants.

- [x] I have manually tested my changes locally with `./script/run` (via `cargo build --bin warp-oss --features gui`).

### Screenshots / Videos

No user-visible visual change in the screenshots already attached to #10269 — that bug is upstream of this change. The behavior change here is for users with `FeatureFlag::ShellSelector` disabled (the OSS path), where the submenu now appears 300 px wide instead of 186 px, matching the experience users with `ShellSelector` enabled already get.

## Agent Mode

CHANGELOG-IMPROVEMENT: New-session submenus render at a consistent 300 px width regardless of the `ShellSelector` feature flag state.

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -1797,17 +1797,22 @@ impl Workspace {
             me.handle_tab_right_click_menu_event(event, ctx);
         });
 
-        // Currently setting the width to 300 px as a middle ground that looks
-        // ok when the shells show the path to the executables, and when they
-        // don't. Going forward we may want to enhance the menu to allow for a
-        // `max_width` and `min_width` instead, so we can allow the menu to
-        // grow as needed.
-        const NEW_SESSION_MENU_WIDTH: f32 = 300.;
+        // Width applied to *submenus* of the new-session menu — `Menu::with_width`
+        // sets the `submenu_width` field rather than the main menu's own width
+        // (see `app/src/menu.rs` for the underlying field). The main menu's own
+        // width is content-sized, which is why the `New worktree config` submenu
+        // entry can clip — that requires a `min_width` plumbed through the Menu
+        // rendering pipeline. See #10269 for the original report and the comment
+        // below for the architectural follow-up.
+        //
+        // Going forward we may want to enhance the menu to allow for a `max_width`
+        // and `min_width` instead, so we can allow the menu to grow as needed.
+        const NEW_SESSION_SUBMENU_WIDTH: f32 = 300.;
         let new_session_menu = ctx.add_typed_action_view(|ctx| {
             if FeatureFlag::ShellSelector.is_enabled() {
                 let theme = Appearance::as_ref(ctx).theme();
                 Menu::new()
-                    .with_width(NEW_SESSION_MENU_WIDTH)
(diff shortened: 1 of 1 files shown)


lonexreb:
Closing this — I should have surveyed open PRs against #10269 before opening this. PR #10542 (already approved) is the real fix and lands the chevron menu width consistency at the right layer (`MENU_DEFAULT_WIDTH` vs `268` reconciliation across both menu variants, with a net code reduction).

The rename + comment-clarification in this PR was based on a misreading of which menu the user actually clicks from the tab bar; the architectural follow-up I referenced doesn't apply once #10542 lands. Apologies for the noise — thanks for the patience.

PR #10801, opened 2026-05-13
Canonicalize local conversation ID mapping to task ID for cloud agent tasks.

zachbai:
## Description

Fixes APP-4460.

There are a host of issues arising from a lack of consistent mapping between conversation ID and task ID in the client. It is currently possible to have multiple local conversations map to the same ambient agent task. In many cases task ID is treated with conversation-like semantics, and we assume that task ID is unique for a given coverastion object.

This PR implements a much stronger 1:1 association between task and coversation ID, though does not go so far as to claim it is an invariant that always holds.

The history model maintains a map of canonical local conversation to server conversation id, which is used in each place where we need to resolve a server conversation id to a local id. Previously we did not consistently use this map, and in certain cases could end up minting new local conversations unnecessarily.

In cases where we end up with multiple local conversation ids mapping to the same ambient task, we could easily end up in a state where the cloud mode pane for a given ambient task assumes conversation ID A is _the_ local conversation ID, while there also exists conversation B which maps to the same task. Then, since conversation ID drives AI block visibility in the agent view, we could end up with an empty looking agent view.

### DIFF
--- a/app/src/ai/agent/api/convert_conversation.rs
+++ b/app/src/ai/agent/api/convert_conversation.rs
@@ -101,11 +101,9 @@ pub fn convert_conversation_data_to_ai_conversation(
             orchestration_harness_type: None,
             parent_conversation_id: None,
             is_remote_child: false,
-            // TODO: Populate run_id from server metadata once it is exposed
-            // in ServerAIConversationMetadata. For cloud conversations that
-            // were spawned via the server API, the run_id is created at task
-            // dispatch time; adding it here would avoid a round-trip to StreamInit.
-            run_id: None,
+            run_id: metadata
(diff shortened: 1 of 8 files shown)


zachbai:
* **#10801** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10801?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10801?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

liliwilson (approved):

PR #10826, opened 2026-05-13
Don't set computer_use_enabled on config snapshot for third-party harnesses

liliwilson:
## Description

WISOTT. Noticed while testing cloud mode that we set this in the snapshot, which doesn't make sense:

image

This is a noop change, since we just ignored the flag before, even if it was set on the snapshot.

## Agent Mode


abhishekp106 (approved):
makes sense to me!

PR #10829, opened 2026-05-13
Fix link display text escaping for notebooks/plans

### LINKED ISSUE #10055: Text for links escaping improperly in notebooks/plans
Text for links is escaping improperly in notebooks/plans. The link itself is fine, but the display text is not.

Linear: https://linear.app/warpdotdev/issue/APP-2628/text-for-links-escaping-improperly-in-notebooksplans

johnturcoo:
## Description

Resolves #10055 

Each time a notebook or plan was saved and reopened, backslash escape characters accumulated in the display text of any link. 

The root cause: `parse_url` captured URLs as raw `&str` slices without unescaping backslash sequences. When the serializer wrote those URLs back as link display text it re-escaped them (. → \\. , \\. -> \\\\\\. , etc), so each load/save cycle added another layer of backslashes.

This PR modifies `parse_url` to process backslash escapes inline (e.g. \\. → .) and return an owned String.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-2628/text-for-links-escaping-improperly-in-notebooksplans

## Testing

- [x] I have manually tested my changes locally with `./script/run` -> [Demo of fix](https://www.loom.com/share/862e429d1ea64c2b9df9a06602a02fbb)
- [x] Added several tests to `markdown_parser_tests.rs` to verify correct URL parsing behavior
- [x] Added round trip test to `markdown_tests.rs` to verify link consistency after several save/load cycles

### Screenshots / Videos

[Demo of issue](https://www.loom.com/share/7051729fc9a1433d89bd004f47cde70d)
[Demo of fix](https://www.loom.com/share/862e429d1ea64c2b9df9a06602a02fbb)

CHANGELOG-BUG-FIX: Fixed escape characters accumulating in notebook and plan link display text on each save/load cycle.

### DIFF
--- a/crates/markdown_parser/src/markdown_parser.rs
+++ b/crates/markdown_parser/src/markdown_parser.rs
@@ -1851,7 +1851,7 @@ fn parse_url_prefix<'a, E: ContextError<&'a str> + ParseError<&'a str>>(
 // - starts with "https://" or "http://" or "www."
 // - has at least one alphanumeric char after the prefix
 // - does not include trailing formatting characters (*, _, ~)
-// - backslash escapes are processed (e.g., `\.` → `.`) 
+// - backslash escapes are processed (e.g., `\.` → `.`)
 fn parse_url<'a, E: ContextError<&'a str> + ParseError<&'a str>>(
     i: &'a str,
 ) -> IResult<&'a str, String, E> {
@@ -1880,7 +1880,7 @@ fn parse_url<'a, E: ContextError<&'a str> + ParseError<&'a str>>(
     let trimmed_raw_url = &i[..trimmed_len];
     let new_remaining = &i[trimmed_len..];
 
-    // Process backslash escapes within the URL using parse_escape (e.g., `\.` → `.`). 
+    // Process backslash escapes within the URL using parse_escape (e.g., `\.` → `.`).
     let mut trimmed_url = String::with_capacity(trimmed_len);
     let mut remaining = trimmed_raw_url;
     while !remaining.is_empty() {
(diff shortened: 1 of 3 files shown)


lucieleblanc (commented):
Nice. This is almost exactly what I had in warpdotdev/warp-internal#22999, with cleaner logic using `parse_escape`. There are some assertions and new test cases I'd add (see that old PR), otherwise LGTM.

peicodes:
@johnturcoo I see you've included a video of the bug - could you also include a video of the fix?

johnturcoo:
@peicodes The demo of the fix is in the testing section but I will add it to the "Screenshots / Videos" section for clarity

crates/markdown_parser/src/markdown_parser.rs
```diff
@@ -1850,34 +1851,48 @@ fn parse_url_prefix<'a, E: ContextError<&'a str> + ParseError<&'a str>>(
...
     if trimmed_len <= min_valid_len {
         return Err(nom::Err::Error(make_error(i, ErrorKind::TakeWhile1)));
     }
 
-    // Return the trimmed URL and adjust remaining
-    let trimmed_url = &i[..trimmed_len];
+    let trimmed_raw_url = &i[..trimmed_len];
     let new_remaining = &i[trimmed_len..];
+
+    // Process backslash escapes within the URL using parse_escape (e.g., `\.` → `.`).
```
peicodes, line 1883: This seems really low level (char by char). Are we able to use `serde_json`?
Not sure if we can but if we can abstract away low-level operations this code will be more maintainable

johnturcoo: `serde_json` does not work here because the escape sequences we may need to remove (\\. , \\# , etc) are not valid [JSON escape sequences](https://www.rfc-editor.org/rfc/rfc8259#section-7). The char by char approach mirrors the existing serializer at `markdown.rs:append_content()`

peicodes (approved):
Sorry I missed the video! Great fix :)

PR #10835, opened 2026-05-13
Remove problematic sentry error in the client

alokedesai:
## Description

The Sentry team reached out to me to let me know that we had a problematic message that was causing us to get rate limited by Sentry (see [here](https://warpdotdev.sentry.io/settings/stats/?project=5658526&statsPeriod=90d))

Even though we had dropped the message within Sentry so it wasn't consuming quota, we were exhausting a separate per-workspace rate limit enforced by Sentry for any event that hits their servers.

As far as I can tell, we're still getting a flood of errors because there are users on old versions of Warp that still exhibit this error. I don't see this log line on stable from my own usage and it's pretty telling that the rate at which we're getting rate limited has been going down over time since mid March.

That being said, I still think this is valuable:
1. In the off chance this is still occurring in newer versions of Warp at a lower rate.
2. To prevent any future problematic occurrences of this causing us to get rate limited by Sentry.

I acknowledge this approach is brittle because we're matching on a string that could easily change. I considered pulling this into a constant or adding something akin to `IFChange` / `ThenChange` but felt it wasn't worth it for now. Happy to add if you feel strongly.

## Agent Mode


alokedesai:
@vorporeal I'll ping you once this is ready for review

PR #10837, opened 2026-05-13
fix(agent): sanitize overlong tool call IDs before OpenAI Responses requests (#8997)

### LINKED ISSUE #8997: Tool call IDs from GLM-5/Kimi models exceed OpenAI's 64-character limit
### Pre-submit Checks

### Describe the bug

When using GLM-5 or Kimi models, tool call IDs are generated that exceed 
OpenAI's Responses API 64-character limit, causing HTTP 400 errors:

OpenAI API error: POST "https://api.openai.com/v1/responses": 400 Bad Request {
    "message": "Invalid 'input[24].call_id': string too long. 
    Expected a string with maximum length 64, but got a string with 
    length 210 instead.",
    "type": "invalid_request_error",
    "param": "input[24].call_id",
    "code": "string_above_max_length"
}

### Debug information
Conversation ID: 406ae4eb-feb5-4aa3-800d-dd0f59a552a0

Warp's backend needs to add ID sanitization:

// In the request handler before sending to OpenAI
const MAX_CALL_ID_LENGTH = 64;

function sanitizeCallId(id: string): string {
  if (id.length <= MAX_CALL_ID_LENGTH) {
    return id;
  }
  // Option 1: Truncate
  return id.slice(0, MAX_CALL_ID_LENGTH);
  
  // Option 2: Hash (prevents collisions)
  // const hash = crypto.createHash('sha256').update(id).digest('hex');
  // return `call_${hash.slice(0, 12)}`; // 17 chars total
}

### Does Not Reproduce in
Opencode or Claude code

### Reproduction
1. Start a conversation with GLM-5 or Kimi model
2. Make multiple tool calls (read files, run commands, etc.)
3. Continue conversation for several turns
4. Error appears and session becomes unrecoverable

### Expected behavior
Warp should sanitize/truncate tool call IDs to 64 characters before 
sending to OpenAI's Responses API, similar to how OpenClaw fixed this 
in https://github.com/openclaw/openclaw/issues/31934

_No response_

### Operating system (OS)

Select an OS

Kali Linux 2026.1, Windows 11 Pro, Windows Server 2022, Ubuntu 24.04 LTS, ,Ubuntu 22.04 LTS

### Shell Version

_No response_

### Current Warp version

v0.2026.03.25.08.24.stable_05 

### Regression

### Recent working Warp date

Noticed issue 2 weeks ago and has become more prevalent 

### Additional context

Our Linux Workstations of Ubuntu run GNOME

Company Kali Linux workstations run:
•  Desktop Environment: XFCE 4.20 (xfce4-session 4.20.4)
•  Display Server: X11
•  Display Manager: LightDM (/usr/sbin/lightdm)
•  Session: lightdm-xsession

None

david-engelmann:
## Description

When using GLM-5 or Kimi models via OpenAI-compatible BYOK endpoints, the upstream provider emits `tool_call_id` values that exceed OpenAI's Responses API 64-character limit. The result is an unrecoverable 400 from OpenAI when Warp sends the corresponding tool-call result downstream:

```text
OpenAI API error: POST "https://api.openai.com/v1/responses": 400 Bad Request {
    "message": "Invalid 'input[24].call_id': string too long. Expected a string
                with maximum length 64, but got a string with length 210 instead.",
    "type": "invalid_request_error",
    "param": "input[24].call_id",
    "code": "string_above_max_length"
}
```

Fix: at the request-building boundary (`AIAgentInput → api::request::input::ToolCallResult`), pass `tool_call_id` through a new `tool_call_id_for_request` helper that:

- passes through IDs `<= 64` chars unchanged (preserves OpenAI-native IDs)
- for longer IDs, SHA-256-hashes the original and emits `call_<hex>` truncated to 64 chars

The mapping is deterministic, preserving the OpenAI tool-call ↔ result pairing invariant: the same upstream ID always sanitizes to the same client-side ID.

## Testing

Three new unit tests cover the three behavior cases:

- `tool_call_result_preserves_openai_compatible_tool_call_id` — IDs at exactly 64 chars aren't rewritten.
- `tool_call_result_hashes_long_tool_call_id_for_openai_responses` — overlong IDs get a `"call_"`-prefixed deterministic hash `<= 64` chars.
- `tool_call_result_hashes_long_tool_call_ids_without_truncation_collisions` — two different overlong IDs that share a prefix produce distinct sanitized IDs (no birthday-prefix-truncation collision).

Verified locally:

```text
cargo nextest run -p warp -E 'test(tool_call_id) or test(hash_long_tool_call) or test(transfer_control_)'
8 tests run: 8 passed

cargo clippy -p warp --tests -- -D warnings
Finished (no warnings)

cargo fmt -p warp -- --check
clean
```

## Server API dependencies

None — this is client-only. The sanitization happens before the request leaves the Rust client.

## Agent Mode

Unchecked (external contribution).

## Changelog

`CHANGELOG-BUG-FIX: Sanitize tool call IDs longer than 64 chars before sending to OpenAI Responses API. Fixes recoverability when using GLM-5 / Kimi via BYOK.`

Fixes #8997

### DIFF
--- a/app/src/ai/agent/api/convert_to.rs
+++ b/app/src/ai/agent/api/convert_to.rs
@@ -3,6 +3,7 @@
 use ai::agent::convert::ConvertToAPITypeError;
 use anyhow::anyhow;
 use chrono::{DateTime, Local, Timelike};
+use sha2::Digest as _;
 use warp_multi_agent_api as api;
 
 use crate::ai::{
@@ -14,11 +15,28 @@ use crate::ai::{
     block_context::BlockContext,
 };
 
+const OPENAI_RESPONSES_MAX_TOOL_CALL_ID_LEN: usize = 64;
+const SANITIZED_TOOL_CALL_ID_PREFIX: &str = "call_";
+
 fn local_datetime_to_timestamp(timestamp: DateTime<Local>) -> prost_types::Timestamp {
     prost_types::Timestamp {
         seconds: timestamp.timestamp(),
         nanos: timestamp.timestamp_subsec_nanos() as i32,
     }
+}
+
+/// Some OpenAI-compatible providers emit tool call IDs that OpenAI's Responses
+/// API rejects when Warp sends the corresponding tool call result downstream.
+fn tool_call_id_for_request(id: impl Into<String>) -> String {
+    let id = id.into();
+    if id.len() <= OPENAI_RESPONSES_MAX_TOOL_CALL_ID_LEN {
+        return id;
+    }
+
+    let digest = sha2::Sha256::digest(id.as_bytes());
+    let mut sanitized = format!("{SANITIZED_TOOL_CALL_ID_PREFIX}{digest:x}");
+    sanitized.truncate(OPENAI_RESPONSES_MAX_TOOL_CALL_ID_LEN);
+    sanitized
 }
 
 impl TryFrom<StaticQueryType> for api::request::input::query_with_canned_response::Type {
@@ -712,7 +730,7 @@ impl TryFrom<AIAgentActionResult> for api::request::input::user_inputs::user_inp
         Ok(
             api::request::input::user_inputs::user_input::Input::ToolCallResult(
                 api::request::input::ToolCallResult {
-                    tool_call_id: action_result.id.into(),
+                    tool_call_id: tool_call_id_for_request(action_result.id),
                     result,
                 },
             ),
--- a/app/src/ai/agent/api/convert_to_tests.rs
+++ b/app/src/ai/agent/api/convert_to_tests.rs
@@ -8,6 +8,55 @@ use chrono::Utc;
 use warp_core::command::ExitCode;
 use warp_multi_agent_api as api;
 
+fn open_code_review_tool_call_result_id(action_id: &str) -> String {
+    let input =
+        api::request::input::user_inputs::user_input::Input::try_from(AIAgentActionResult {
+            id: action_id.to_string().into(),
+            task_id: TaskId::new("task".to_string()),
+            result: AIAgentActionResultType::OpenCodeReview,
+        })
+        .unwrap();
+
+    match input {
+        api::request::input::user_inputs::user_input::Input::ToolCallResult(result) => {
+            result.tool_call_id
+        }
+        other => panic!("Expected tool-call-result input, got {other:?}"),
+    }
+}
+
+#[test]
+fn tool_call_result_preserves_openai_compatible_tool_call_id() {
+    let id = "a".repeat(super::OPENAI_RESPONSES_MAX_TOOL_CALL_ID_LEN);
+
+    assert_eq!(open_code_review_tool_call_result_id(&id), id);
+}
+
+#[test]
+fn tool_call_result_hashes_long_tool_call_id_for_openai_responses() {
+    let id = "glm_tool_call_".repeat(16);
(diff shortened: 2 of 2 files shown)


zachbai:
@szgupta can you look at this - i'm not sure why toolcall ids from glm/kimi requests are making it to openai at all

david-engelmann:
@zachbai @szgupta Yeah, the narrow thing this PR fixes is just "tool_call_id over 64 chars makes `/v1/responses` 400." The thing you're pointing at, why a tool call result from a non-OpenAI BYOK backend (GLM/Kimi/etc. talking to their OpenAI-equivalent endpoint) ends up in a later `/v1/responses` call at all. Is a real concern about cross-provider conversation state getting re-bundled into OpenAI's shape on subsequent turns. This PR doesn't try to address that part.

Still figuring out the contribution flow for this repo, would it have been better to open a separate tracking issue for the cross-provider mixing first and treat this as a follow-up under it, or is shipping the narrow 64-char defense first (as an adapter at the wire boundary) the right path? Happy to do whichever fits how the team wants external contributions to land here.

zachbai:
Opening a github issue about the cross-provider bit would be welcomed, just so we can ensure we're fixing the right issue at the root!  
  
even if this is the right conceptual fix it might be more appropriate to do this type of enforcement/sanitization in our server

david-engelmann:
@zachbai That makes sense, thanks for the clarification!

On the cross‑provider side, I’ll go ahead and open a separate issue that describes the full end‑to‑end behavior I’m seeing (GLM/Kimi BYOK emitting long tool_call_ids, then those ending up in later /v1/responses calls against OpenAI) so you can decide what the “right” root‑level fix looks like in the broader architecture.

For this PR specifically, I’m happy to defer to you and the team on where this kind of enforcement belongs. If it’s cleaner to handle the 64‑char constraint entirely in the server / internal code paths, feel free to close this out in favor of that approach. If a thin client‑side adapter at the wire boundary still ends up being useful as part of the longer‑term cross‑provider story, I’m also fine keeping it open until that shakes out.

In the meantime I’ll leave the PR as‑is so you can reference the tests and helper behavior while you’re deciding where to land the final fix.

szgupta:
Hey @david-engelmann, thanks for bringing this to our attention. Indeed, this issue would be best-addressed with a fix in our harness (server code). The reason being that there are multiple places in the data model where we use the tool call ID as a reference key, so we'll need to make sure the sanitized key is consistent. Can you create an issue and link it here?

In terms of why we even made a request to OpenAI, I assume this is because of a fallback - GLM failed and so we retried with an OpenAI model. If you have a conversation ID, I can take a closer look to make sure that was the case. You can also turn that behaviour off if you prefer:
Screenshot 2026-05-14 at 11 08 39 AM

david-engelmann:
@szgupta I can raise a ticket for this so that its properly explained and points to the internal setup for the fix, the original issue listed the convo id of 406ae4eb-feb5-4aa3-800d-dd0f59a552a0 which might help you on the internal fix side

szgupta:
Thanks, will close this out for now

david-engelmann:
@szgupta filed the tracking issue at #11001 — covers the cross-provider `tool_call_id` reach into `/v1/responses`, the 64-char limit failure mode, conversation ID `406ae4eb-feb5-4aa3-800d-dd0f59a552a0`, and notes that the sanitization needs to be applied consistently across every consumer of the ID (not just the wire boundary) since it's used as a reference key in multiple places. Closing this loop.

PR #10852, opened 2026-05-13
Migrate code view to use FileLocation

kevinyang372:
### Problem
Clicking a remote file in the file tree twice opens two separate editor panes for the same file. Locally, the app correctly detects the file is already open and focuses the existing editor instead.

### Root Cause
All duplicate-detection paths relied on `CodeSource::path()`, which returns `None` for remote files (`FileLocation::Remote`). This caused every dedup check to be skipped entirely for remote files.

### Changes

**Core type migration (`TabData.path` → `TabData.location`)**
- Replaced `TabData.path: Option<PathBuf>` with `location: Option<FileLocation>`, so both local and remote files are tracked as the canonical tab identity
- Added `local_path()` convenience accessor for callers that are inherently local-only (save, rename, persistence, markdown preview)
- Migrated all callers: display (tab titles, tooltips), dedup (focus existing, merge tabs), and local-only operations

**Dedup fix (the actual bug fix)**
- `focus_existing_tab_if_present` now compares `FileLocation` instead of local path only
- Added `CodeManager::get_locator_for_location_in_tab` that matches on `CodeSource::location()` for both local and remote
- Updated `open_code` (both grouping-on and grouping-off paths) to use `source.location()` instead of `source.path()`
- Updated `CodePane::pre_attach` to use `source.location()` and the new location-based lookup

**ActiveFileModel migration**
- Changed `ActiveFileModel` to store `Option<FileLocation>` instead of `Option<PathBuf>`
- Updated `CodeViewEvent::TabChanged`/`FileOpened` to carry `FileLocation`
- File tree handler now works for both local (`try_from_local`) and remote (`remote.path` directly) — enables "reveal active file in tree" for remote files
- `OpenedFilesModel` tracking gated on `FileLocation::Local` since `DetectedRepositories` is local-only

### Bonus fixes
- Remote tab titles now show actual file name via `FileLocation::display_name()` instead of "Untitled"
- `merge_tabs` correctly deduplicates remote tabs (previously silently dropped them)
- Unsaved changes dialog shows correct file names for remote files

---
[Conversation](https://staging.warp.dev/conversation/18fcb46f-ab51-4429-833e-fb3a9d4ea203) | [Plan](https://staging.warp.dev/drive/notebook/Zr9BSvVisGLAPP8BXRyTJc)

## Demo
https://www.loom.com/share/6cc736c68bf24242b2919a3c2e746d7d

### DIFF
--- a/app/src/code/active_file.rs
+++ b/app/src/code/active_file.rs
@@ -1,22 +1,21 @@
 //! Module containing the definition of [`ActiveFileModel`],
 //! which tracks the currently focused file across an entire PaneGroup.
 
-use std::path::PathBuf;
-
+use super::buffer_location::FileLocation;
 use warpui::{Entity, ModelContext};
 
 /// Events emitted by the ActiveFileModel.
 #[derive(Debug, Clone)]
 pub enum ActiveFileEvent {
     /// A new file became focused.
-    ActiveFileChanged { file_info: PathBuf },
+    ActiveFileChanged { location: FileLocation },
 }
 
 /// Model that tracks the currently focused file.
 #[derive(Default)]
 pub struct ActiveFileModel {
     /// The currently focused file, if any.
-    active_file: Option<PathBuf>,
+    active_file: Option<FileLocation>,
 }
 
 impl Entity for ActiveFileModel {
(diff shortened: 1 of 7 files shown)


kevinyang372:
* **#10852** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10852?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10852?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

moirahuang (approved):
i'm not sure if this is caused by this PR but we're missing a bunch of stuff on the right click menu. here's remote
Image
vs local

Image

not being able to open the same file in multiple panes or tabs makes it hard to test the other case i flagged to you when one of the files has unsaved changes and the other file thinks it also has unsaved changes

kevinyang372:
@moirahuang This is not caused by this PR -- I will followup on this

PR #10870, opened 2026-05-14
Respect CODEX_HOME and CLAUDE_CONFIG_DIR for writing config.

liliwilson:
## Description

Update the implementations for claude and codex harnesses to consistently respect the environment variables `CLAUDE_CONFIG_DIR` and `CODEX_HOME` respectively.

While `.claude.json` gets written to `~` (not `~/.claude`) when the `CLAUDE_CONFIG_DIR` is unset, we need to make sure that we set it to `CLAUDE_CONFIG_DIR/.claude.json` when the environment variable is set.

This is important for the self-hosted worker direct backend, to make sure that we can seed per-task config in worker-scoped repos. Associated `oz-agent-worker` PR: https://github.com/warpdotdev/oz-agent-worker/pull/72

## Testing

Ran with `./script/oz-local` and passed in `-e CODEX_HOME` and `-e CLAUDE_CONFIG_DIR`. Nuked my local `~/.claude` and `~/.codex` directories and then confirmed agents could run to completion and seed the correct config files.

Also ran without setting the env vars and confirmed that we run to completion.

## Agent Mode

### DIFF
--- a/app/src/ai/agent_sdk/driver/harness/claude_code.rs
+++ b/app/src/ai/agent_sdk/driver/harness/claude_code.rs
@@ -27,8 +27,8 @@ use warp_managed_secrets::ManagedSecretValue;
 use super::super::terminal::{CommandHandle, TerminalDriver};
 use super::super::{AgentDriver, AgentDriverError};
 use super::claude_transcript::{
-    claude_config_dir, read_envelope, write_envelope, write_session_index_entry, ClaudeResumeInfo,
-    ClaudeTranscriptEnvelope,
+    claude_config_dir, home_dir_for_claude_config, read_envelope, write_envelope,
+    write_session_index_entry, ClaudeResumeInfo, ClaudeTranscriptEnvelope,
 };
 use super::json_utils::{read_json_file_or_default, write_json_file};
 use super::{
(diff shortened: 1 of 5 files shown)


app/src/ai/agent_sdk/driver/harness/claude_code.rs
```diff
@@ -578,24 +578,25 @@ pub(crate) fn prepare_claude_environment_config(
...
 }
 
-fn claude_home_dir() -> Result<PathBuf> {
-    #[cfg(test)]
-    if let Some(home_dir) = std::env::var_os("HOME") {
-        if !home_dir.as_os_str().is_empty() {
-            return Ok(PathBuf::from(home_dir));
+// This function is used specifically for determining where to land `.claude.json`.
+fn claude_global_config_path() -> Result<PathBuf> {
+    if let Ok(dir) = std::env::var("CLAUDE_CONFIG_DIR") {
```
bnavetta, line 591: Oh interesting, I thought `.claude.json` didn’t respect the env vars, only the directory did

bnavetta (approved):

app/src/ai/agent_sdk/driver/harness/claude_code.rs
liliwilson: Yeah, I originally did too... I did test this though and fortunately it does respect the env var

PR #10875, opened 2026-05-14
add nld decision telemetry event

evelyn-with-warp:
## Linked Issue

## Testing

./script/run --features warpui/log_named_telemetry_events

### Screenshots / Videos

**ls**
image

**ls with file attached**
image

** ls with file attached and ⌘ I**
image

**hello in terminal mode without enabling onnx classifier**
image

** approve after hello, should trigger AgentFollowUp whitelist**
image

**natural language without enabling onnx classifier**
image
**hello with enabling onnx classifier, should trigger whitelisting**
image
**!hello with enabling onnx classifier, should trigger shell prefix**
image
**what's the weather should trigger nldclassifier**
image
**/plan should trigger app level heuristics**
image

some video on latest commit
https://www.loom.com/share/480d640e565d4d28a40e1a5d15fc709c
## Agent Mode


evelyn-with-warp:
@szgupta I occasionally saw decision source is Null, but it did not reproduce robustly. 
```
09:48:39.634 [INFO] [warpui_core::telemetry::event_store] Recorded telemetry event: Event {
    payload: NamedEvent {
        user_id: Some(
            "paKnCQIb4NVDXhTQambFVn6dnej1",
        ),
        anonymous_id: "9a826897-8e46-4716-afa7-3fcfc7bc1c42",
        name: "AgentMode.NaturalLanguageDetection.InputBufferSubmitted",
        value: Some(
            Object {
                "decision_source": Null,
                "input": String("approve"),
                "input_type": String("Shell"),
                "is_locked": Bool(false),
                "was_lock_set_with_empty_buffer": Bool(false),
            },
        ),
    },
    session_created_at: 2026-05-18T16:34:43.407171Z,
    timestamp: 2026-05-18T16:48:39.633855Z,
    contains_ugc: true,
```
Despite our code does imply if no NLD decision applied when is_locked is False the decision source will be Null, it's not clear to me why there's no NLD decision for unlocked sessions. 
I'm okay with current implementation having coverage less than 100%, as long as the logged decision source is accurate

app/src/ai/blocklist/input_model.rs
```diff
@@ -139,6 +140,8 @@ pub struct BlocklistAIInputModel {
     /// The timestamp of the last time the input mode was switched, if the switch was to AI mode and
     /// it was autodetected. Else, `None`.
     last_ai_autodetection_ts: Option<Instant>,
+    /// The source of the final input decision that is currently reflected in `input_config`.
+    input_decision_source: Option<InputDecisionSource>,
```
szgupta: When is this `None`?

szgupta: "input_decision" is vague - can we be more clear here?

crates/input_classifier/src/onnx/mod.rs
```diff
@@ -161,20 +135,96 @@ impl InputClassifier for OnnxClassifier {
...
+                })
             }
         }
     }
 }
 
+#[cfg_attr(not(target_family = "wasm"), async_trait)]
+#[cfg_attr(target_family = "wasm", async_trait(?Send))]
+impl InputClassifier for OnnxClassifier {
+    async fn detect_input_decision(
```
szgupta: Does this mean checking the input decision happens separately from actually running NLD? Feel like the `detect_input_type` API ought to return the type and why that decision was made

szgupta (commented):
Initial thoughts

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: like the case i got https://github.com/warpdotdev/warp/pull/10875#issuecomment-4480106102, although not robustly reproducible; 
i'd rather leave it as none (trade coverage for accuracy) if there's some decision path that I am not aware of.

evelyn-with-warp: agreed on this. maybe `nld_decision_source` is better?

app/src/ai/blocklist/input_model.rs
```diff
@@ -139,6 +140,8 @@ pub struct BlocklistAIInputModel {
     /// The timestamp of the last time the input mode was switched, if the switch was to AI mode and
     /// it was autodetected. Else, `None`.
     last_ai_autodetection_ts: Option<Instant>,
+    /// The source of the final input decision that is currently reflected in `input_config`.
```
vorporeal: if this is tied to some data in `InputConfig`, why is it not part of `InputConfig`?

also seems like it might simplify some other code

evelyn-with-warp: @vorporeal  it was intended to say the source will be set in `set_input_config_internal`. Will update the comment to be less confusing; The decision source is dependent on user input text, and will be decided in runtime (e.g if there's ONNX panic, the source will be NLDHeuristicClassifier instead of NLDClassifier..etc), and therefore i think it's not appropriate to put it in input config, which aims to track the mode state (it's logically possible to have different decision source in one mode state)
i also didn't get why put nld_decision_source in inputConfig will simplify this implementation, as we will still need to update everywhere we call set_input_config. 
Pls lmk if i oversee some potential shortcut here

crates/input_classifier/src/onnx/mod.rs
evelyn-with-warp: it was mostly a rename. we did append source besides decision. Reverted it back to avoid confusion https://github.com/warpdotdev/warp/pull/10875/changes/557db91464ec17044c13eb5206fbea8ca3bd1db1

app/src/ai/blocklist/input_model.rs
```diff
@@ -139,6 +140,8 @@ pub struct BlocklistAIInputModel {
     /// The timestamp of the last time the input mode was switched, if the switch was to AI mode and
     /// it was autodetected. Else, `None`.
     last_ai_autodetection_ts: Option<Instant>,
+    /// The source of the final input decision upon submission
+    nld_decision_source: Option<NldDecisionSource>,
```
szgupta: nit: should we call this `last_ai_autodetection_decision_source` to be consistent with the property above? even better, wrap the two into a mini struct!

app/src/ai/blocklist/input_model.rs
```diff
@@ -394,21 +414,26 @@ impl BlocklistAIInputModel {
...
 
     /// Swaps between Agent/Shell input types while preserving lock state. Temporarily disables
     /// autodetection.
     pub fn set_input_type(&mut self, input_type: InputType, ctx: &mut ModelContext<Self>) {
         self.temporarily_disable_autodetection();
         let current_config = self.input_config();
-        self.set_input_config_internal(current_config.with_input_type(input_type), ctx);
+        self.set_input_config_internal(
+            current_config.with_input_type(input_type),
+            Some(NldDecisionSource::ManualToggle),
```
szgupta: hrm, how are we sure this is a manual toggle? this API is in a model, not a ui view

app/src/terminal/input.rs
```diff
@@ -476,6 +476,32 @@ fn get_agent_mode_new_conversation_hint_text() -> &'static str {
     let index = HINT_INDEX.fetch_add(1, Ordering::Relaxed) % AGENT_MODE_HINT_OPTIONS.len();
     AGENT_MODE_HINT_OPTIONS[index]
 }
+fn submitted_nld_decision_source(input_model: &BlocklistAIInputModel) -> Option<NldDecisionSource> {
```
szgupta: What is this doing?

app/src/terminal/input.rs
```diff
@@ -9754,12 +9821,13 @@ impl Input {
                                 });
 
                             self.ai_input_model.update(ctx, |ai_input_model, ctx| {
-                                ai_input_model.set_input_config(
+                                ai_input_model.set_input_config_with_source(
```
szgupta: nit: i'd leave this API as `set_input_config` and make `NldDecisionSource` required. There shouldn't be a single callsite that can't specify why we chose to change the input type, right?

crates/input_classifier/src/lib.rs
```diff
@@ -7,12 +7,55 @@ pub mod test_utils;
...
+use serde::{Deserialize, Serialize};
 
 pub use heuristic_classifier::HeuristicClassifier;
 pub use input_type::InputType;
 #[cfg(feature = "onnx")]
 pub use onnx::{Model as OnnxModel, OnnxClassifier};
 
+/// The source of the final input type decision applied to the user input.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum NldDecisionSource {
```
szgupta: I think we should decouple the decision source in the `input_classifier` crate vs the application level decision source. The one coming from this crate should only include the sources that it can actually produce (stuff in the NLD pipeline), whereas the one in the application should be like

```
enum AiAutodetectionSource {
    ManualToggle,
    ...,
    ClassifierSource(ClassifierSource)
}
```
(naming TBD)

crates/input_classifier/src/onnx/mod.rs
```diff
@@ -85,53 +91,21 @@ impl OnnxClassifier {
...
-
-        // Otherwise, defer all decision-making to the model.
-        self.classify_input(input, context)
-            .await
-            .map(|result| result.to_input_type())
-            .unwrap_or(context.current_input_type)
-    }
-
-    async fn classify_input(
+    async fn classify_input_with_source(
```
szgupta: nit: let's keep the name as `classify_input`

crates/input_classifier/src/onnx/mod.rs
```diff
@@ -85,53 +91,21 @@ impl OnnxClassifier {
...
         &self,
         input: warp_completer::ParsedTokensSnapshot,
-        _context: &Context,
-    ) -> anyhow::Result<ClassificationResult> {
+        context: &Context,
+    ) -> anyhow::Result<ClassificationWithSource> {
```
szgupta: Can we just add the source to the existing `ClassificationResult`?

app/src/terminal/input.rs
evelyn-with-warp: removed; it seems not necessary https://github.com/warpdotdev/warp/pull/10875/changes/b6f6fe2b53c6a4e2071c9bb06a401a4d05de8a01

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: also for some case if it's set to enable auto detection, then it's okay to be None for NLD to fill in later e.g. in [enable_auto_detection](https://github.com/warpdotdev/warp/blob/a67c7adece6dd8dffce9013d81777bada99a0b48/app/src/terminal/input.rs#L6118)

crates/input_classifier/src/lib.rs
```diff
@@ -7,12 +7,26 @@ pub mod test_utils;
...
+use serde::{Deserialize, Serialize};
 
 pub use heuristic_classifier::HeuristicClassifier;
 pub use input_type::InputType;
 #[cfg(feature = "onnx")]
 pub use onnx::{Model as OnnxModel, OnnxClassifier};
 
+/// Sources produced by the NLD pipeline.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum NldDecision {
```
szgupta: nit: `InputClassifierDecisionSource`

let's avoid "nld" since we don't really use that term in the codebase today

crates/input_classifier/src/lib.rs
```diff
@@ -7,12 +7,26 @@ pub mod test_utils;
...
+    NldClassifier,
+    NldClassifierFallbackHeuristic,
+    NldClassifierFallbackCurrentInput,
+    Denylist,
+    HistoryMatch,
```
szgupta: The `input_classifier` crate doesn't check for history

crates/input_classifier/src/lib.rs
```diff
@@ -7,12 +7,26 @@ pub mod test_utils;
...
+    OneOffWhitelist,
```
szgupta: Can you break this out into the two allowlists: "ShellCommandAllowlist" vs "NaturalLanguageAllowlist" vs "NaturalLanguageFollowupAllowlist"

crates/input_classifier/src/lib.rs
```diff
@@ -7,12 +7,26 @@ pub mod test_utils;
...
 #[cfg(feature = "onnx")]
 pub use onnx::{Model as OnnxModel, OnnxClassifier};
 
+/// Sources produced by the NLD pipeline.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum NldDecision {
+    NldClassifier,
+    NldClassifierFallbackHeuristic,
+    NldClassifierFallbackCurrentInput,
+    Denylist,
```
szgupta: Which denylist?

crates/input_classifier/src/lib.rs
szgupta: All of these variants should be more specific and documented please!

crates/input_classifier/src/heuristic_classifier/mod.rs
```diff
@@ -39,24 +39,31 @@ pub struct HeuristicClassifier;
 #[cfg_attr(not(target_family = "wasm"), async_trait)]
 #[cfg_attr(target_family = "wasm", async_trait(?Send))]
 impl InputClassifier for HeuristicClassifier {
-    async fn detect_input_type(&self, input: ParsedTokensSnapshot, context: &Context) -> InputType {
+    async fn detect_input_type(
+        &self,
+        input: ParsedTokensSnapshot,
+        context: &Context,
+    ) -> (InputType, NldDecision) {
```
szgupta: Instead of a tuple, can you create a new return type (struct) with these two properties?

app/src/ai/blocklist/input_model.rs
```diff
@@ -17,7 +17,92 @@ use settings::Setting as _;
...
+    AtContextMenuInsert,
+}
+
+/// The source of the final input type decision applied to the user input.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum InputTypeAutoDetectionSource {
+    /// App level settings from explicit user actions.
+    AppLevelOverride(AppLevelOverride),
+    /// App level implicit / side-effect transitions from other UI flows.
+    AppLevelHeuristic(AppLevelHeuristic),
```
szgupta: Hrm, I feel like we should collapse these into one. In fact, I would change this type to
```
enum InputTypeAutoDetectionSource {
    InputClassifier(InputClassifierDecisionSource),
    HistoryMatch,
    ... // all the other app types inlined here
}
```

crates/input_classifier/src/lib.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10875/changes/bdff07cae2dd27e111ba9990072b8e68ad442bb5

crates/input_classifier/src/lib.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10875/changes/9643ea55fb201b937b59f66e82631867246bd8e0

crates/input_classifier/src/lib.rs
evelyn-with-warp: [AISettings.autodetection_command_denylist](https://github.com/warpdotdev/warp/blob/dbca9ac43664135b07545ee5cc96fa8662b67417/app/src/ai/blocklist/input_model.rs#L634-L638) to block auto detection run ; also moved to app level tracking now

crates/input_classifier/src/lib.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10875/changes/d9c5dc437cb23d151e98bca494c7749eaf2a07a0

crates/input_classifier/src/lib.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10875/changes/66845db2290053bf7108b4f2779462f2a770df93

crates/input_classifier/src/heuristic_classifier/mod.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/10875/changes/8ad713375679282fce1eef3dcef090410ad01e24

app/src/ai/blocklist/input_model.rs
```diff
@@ -17,7 +17,74 @@ use settings::Setting as _;
 use warp_core::features::FeatureFlag;
 use warpui::{AppContext, Entity, EntityId, ModelContext, ModelHandle, SingletonEntity};
 
-pub use input_classifier::InputType;
+pub use input_classifier::{InputClassifierDecisionSource, InputType};
+
+/// The source of the final input type decision applied to the user input.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum InputTypeAutoDetectionSource {
```
szgupta, line 24: nit: this is more like `InputTypeDecisionSource` - because a lot of these don't go through "auto-detection" (e.g. voice input)

app/src/ai/blocklist/input_model.rs
```diff
@@ -17,7 +17,74 @@ use settings::Setting as _;
...
+    /// Decision produced by the input classifier pipeline.
+    InputClassifierDecisionSource(InputClassifierDecisionSource),
+    /// User explicitly toggled the input type via cmd + I.
+    ManualToggle,
+    /// `!` shell prefix force-locked the input to Shell mode.
+    ShellPrefix,
+    /// Image / file attachment in progress force-locked AI mode.
+    AttachmentForcedAi,
```
szgupta, line 32: nit: I would just call this "AttachedFile"

app/src/ai/blocklist/input_model.rs
```diff
@@ -17,7 +17,74 @@ use settings::Setting as _;
...
+    /// First token matched the autodetection command denylist.
+    Denylist,
```
szgupta, line 34: Can you make this more specific (i.e. which denylist)

app/src/ai/blocklist/input_model.rs
```diff
@@ -333,6 +410,7 @@ impl BlocklistAIInputModel {
         } else {
             AISettings::as_ref(ctx).is_ai_autodetection_enabled(ctx)
         };
+        let initial_decision_source = None;
```
szgupta, line 405: Can you just inline this? no need for variable declaration

app/src/ai/blocklist/input_model.rs
```diff
@@ -372,6 +451,9 @@ impl BlocklistAIInputModel {
     pub fn input_config(&self) -> InputConfig {
         self.input_config
     }
+    pub fn last_ai_autodetection_source(&self) -> Option<InputTypeAutoDetectionSource> {
```
szgupta, line 446: nit: newline above

app/src/ai/blocklist/input_model.rs
```diff
@@ -139,6 +206,8 @@ pub struct BlocklistAIInputModel {
     /// The timestamp of the last time the input mode was switched, if the switch was to AI mode and
     /// it was autodetected. Else, `None`.
     last_ai_autodetection_ts: Option<Instant>,
+    /// the latest input type classification decision source
+    last_ai_autodetection_source: Option<InputTypeAutoDetectionSource>,
```
szgupta, line 202: The more I think about it, I think this should be stored on `InputConfig` and called `decision_source`. it makes more sense there than on the InputModel itself.

app/src/ai/blocklist/input_model.rs
```diff
@@ -17,7 +17,74 @@ use settings::Setting as _;
 use warp_core::features::FeatureFlag;
 use warpui::{AppContext, Entity, EntityId, ModelContext, ModelHandle, SingletonEntity};
 
-pub use input_classifier::InputType;
+pub use input_classifier::{InputClassifierDecisionSource, InputType};
+
+/// The source of the final input type decision applied to the user input.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
szgupta, line 23: Can you move this down, past the imports

app/src/server/telemetry/events.rs
```diff
@@ -4753,6 +4759,7 @@ impl TelemetryEvent {
             TelemetryEvent::AgentExitedShellProcess { .. } => true,
             TelemetryEvent::CreateProjectPromptSubmitted { .. } => false,
             TelemetryEvent::CreateProjectPromptSubmittedContent { .. } => true,
+            TelemetryEvent::InputBufferSubmitted { .. } => false,
```
szgupta, line 4742: revert this? it was fine before

crates/input_classifier/src/onnx/mod.rs
```diff
@@ -101,36 +106,53 @@ impl InputClassifier for OnnxClassifier {
...
+                );
             }
 
             // If the first token is one of a specific set of shell command keywords (e.g.: echo or sudo),
             // we should classify it as shell.
             if is_one_off_shell_command_keyword(&first_word) {
-                return InputType::Shell;
+                return InputClassificationResult::new(
+                    InputType::Shell,
+                    InputClassifierDecisionSource::ShellHeuristic,
```
szgupta, line 120: Should we break this up into two types rather than use the same one here and in `is_likely_shell_command`?

crates/input_classifier/src/lib.rs
```diff
@@ -37,20 +70,24 @@ pub struct ClassificationResult {
     p_shell: f32,
     /// The probability that the input is a natural language query to AI.
     p_ai: f32,
+    /// The classifier source that produced this classification.
+    pub source: InputClassifierDecisionSource,
```
szgupta, line 74: nit: instead of making this pub, define a getter

crates/input_classifier/src/onnx/mod.rs
```diff
@@ -201,3 +223,70 @@ impl HasPanicked {
         self.inner.is_completed()
     }
 }
+
+#[cfg(test)]
+mod tests {
```
szgupta, line 228: nit: move to `mod_tests` file

szgupta (approved):
Awesome, thank you for iterating!

PR #10883, opened 2026-05-14
[Remote envs] codebase indexing persistence

moirahuang:
## Description
Adding persistence for codebase indexing based off of https://github.com/warpdotdev/warp/pull/9508/changes/7e20f47b3298526a21045c51d93f41df8c8848da

1. Added a resync proto to better mirror the local state where we have index + resync
2. Separating app vs. local restore behavior
3. Always expose the search codebase tool to better mirror local, instead of only exposing it once indexing is ready 

## Testing
Locally tested + unit tests

### Screenshots / Videos
https://www.loom.com/share/61589523592c4375bfeb911b25da617c

## Agent Mode


app/src/ai/agent/api/impl.rs
```diff
@@ -193,9 +193,7 @@ fn get_supported_tools(params: &RequestParams) -> Vec<api::ToolType> {
             // after a successful connection handshake, so its presence is a
             // sufficient proxy for client availability.
             supported_tools.extend(&[api::ToolType::ReadFiles, api::ToolType::ApplyFileDiffs]);
-            if FeatureFlag::RemoteCodebaseIndexing.is_enabled()
-                && params.remote_codebase_search_available
```
moirahuang, line 213: removing `remote_codebase_search_available` to more closely match local aka always have codebase search available as a tool but we fall back to other methods if indexing isn't done

crates/remote_server/proto/remote_server.proto
```diff
@@ -337,6 +338,10 @@ message IndexCodebase {
   string repo_path = 1;
   string auth_token = 2;
 }
+message ResyncCodebase {
```
moirahuang, line 343: adding this to mirror local where it distinguishes between index and retry

app/src/lib.rs
```diff
@@ -343,6 +350,102 @@ fn determine_agent_source(
...
+            }
+            Some(persistence::RestoredPersistenceData::RemoteCodebaseIndexing(
+                codebase_indices,
+            )) => {
+                let codebase_index_count = codebase_indices.len();
+                log::info!(
+                    "[Remote codebase indexing] Restored daemon codebase index metadata: metadata_count={codebase_index_count}"
+                );
+                Self {
+                    persisted_workspaces: codebase_indices,
```
moirahuang: this is what matters for the local vs. remote case

app/src/remote_server/codebase_index_model.rs
```diff
@@ -148,13 +148,16 @@ impl RemoteCodebaseIndexModel {
                 is_git,
             } => {
                 self.record_navigated_directory(remote_path);
-                if *is_git && should_auto_index_remote_codebase(ctx) {
+                if *is_git
+                    && should_auto_index_remote_codebase(ctx)
+                    && self.should_request_auto_index_for_navigated_git_repo(remote_path)
```
moirahuang, line 186: adding this so we don't just immediately reindex when we cd to a repo, similar to local

app/src/persistence/mod.rs
```diff
@@ -69,6 +69,11 @@ pub enum PersistenceScope {
     RemoteServerDaemon { identity_key: String },
 }
 
+pub enum RestoredPersistenceData {
```
kevinyang372: Why do we need different persistence scope for app and remote daemon? It should be fine if read app-scoped data right

app/src/remote_server/codebase_index_model.rs
```diff
@@ -183,12 +220,42 @@ impl RemoteCodebaseIndexModel {
...
+        }
+    }
 
     fn apply_statuses_snapshot(
         &mut self,
         host_id: &HostId,
         statuses: &[RemoteCodebaseIndexStatusWithPath],
     ) {
+        let status_count = statuses.len();
+        log::info!(
```
kevinyang372: Do we need all of these debugging logs

crates/ai/src/index/full_source_code_embedding/manager.rs
```diff
@@ -298,6 +369,8 @@ impl CodebaseIndexManager {
             max_files_repo_limit: 0,
             embedding_generation_batch_size: 100,
             indexing_enabled: true,
+            #[cfg(feature = "local_fs")]
+            snapshot_storage: SnapshotStorage::app_default(),
```
kevinyang372, line 389: I think we should have a tech spec on this. It's hard for me to grok the high-level approach. There are a lot of moving pieces here

skills-lock.json
```diff
@@ -3,30 +3,35 @@
   "skills": {
     "brandalf": {
       "source": "warpdotdev/common-skills",
+      "ref": "main",
```
kevinyang372: Remove?

crates/remote_server/proto/remote_server.proto
kevinyang372, line 343: Do this need to be a separate proto? Isn't the existing behavior of IndexCodebase to full resync already if it is already indexed?

moirahuang: i mainly added it for better parity with local. it is true that we can just have `IndexCodebase` resync

app/src/persistence/mod.rs
moirahuang: i didn't want to read all the other app scoped data. i can spec this out in more detail for us to discuss though. i had debated this for a while myself

moirahuang: updated so we have the same scope for app and remote and at initialization we filter what data is read for remote

app/src/lib.rs
```diff
@@ -1256,6 +1280,30 @@ pub(crate) fn initialize_app(
             )
         });
 
+    if matches!(launch_mode, LaunchMode::RemoteServerDaemon { .. }) {
```
moirahuang: here is where we explicitly set what data we're setting for remote

crates/ai/src/index/full_source_code_embedding/manager.rs
moirahuang: added a tech spec

app/src/remote_server/codebase_index_model.rs
moirahuang: cleaned up

app/src/remote_server/server_model.rs
```diff
@@ -1274,6 +1306,42 @@ impl ServerModel {
         }
     }
 
+    fn prepare_codebase_index_request(
```
moirahuang, line 1338: added this to reduce repeated shared logic

app/src/lib.rs
```diff
@@ -1258,6 +1279,30 @@ pub(crate) fn initialize_app(
...
+        let codebase_index_count = persisted_workspaces.len();
+        log::info!(
+            "[Remote codebase indexing] Restored daemon codebase index metadata: metadata_count={codebase_index_count}"
+        );
+        cloud_objects = Default::default();
```
kevinyang372: This seems like a potential footgun here. Someone could easily add a new field and forget to update it here

Could we make sure we initialize all of them to be default and branch off on remote / local on assigning values?

app/src/remote_server/codebase_index_model.rs
```diff
@@ -237,6 +281,59 @@ impl RemoteCodebaseIndexModel {
         search_availability_for_status(status, remote_path)
     }
 
+    fn resolve_remote_repo_path(
+        &self,
+        host_id: &HostId,
+        current_working_directory: Option<&str>,
+        explicit_repo_path: Option<&str>,
```
kevinyang372: What is explicit_repo_path here? Not sure I fully understand what it means

app/src/remote_server/server_model.rs
```diff
@@ -94,6 +95,16 @@ enum HandlerOutcome {
     Async(Option<SpawnedFutureHandle>),
 }
 
+struct CodebaseIndexRequest {
+    repo_path: PathBuf,
+}
+
+#[derive(Clone, Copy)]
+enum CodebaseIndexRequestPathKind {
```
kevinyang372: I am not sure if I would understand what this means as a reader -- can we just have an operation type enum (Drop / Sync / Index) that derives things like the canonicalization behavior / auth name / operation name / ?

kevinyang372 (approved):

app/src/remote_server/codebase_index_model.rs
moirahuang: renamed to `requested_codebase_path`

app/src/lib.rs
moirahuang: good call, updated

PR #10885, opened 2026-05-14
Add auth secret dropdown picker for non-Oz orchestration

advait-m:
Loom for run agents card: https://www.loom.com/share/10263029b97d4bada81f24b2081e3410
Loom for plan card: https://www.loom.com/share/f3e9b7dc44044b8187c723c8d59e4ae4
Fixes https://linear.app/warpdotdev/issue/QUALITY-700/api-key-configuration-for-orchestration-card-in-line-and-plan

## Description

Adds an "API key" picker to the orchestration plan card and the `run_agents` confirmation card. When the orchestration harness is set to a non-Oz harness (Claude Code or Codex) running on Cloud, the user can now pick a managed secret that will be injected into each cloud child agent as it spawns. This unblocks the multi-harness local → cloud orchestration flow that broke when the team-scoped Anthropic / OpenAI blanket env var was removed — Ian's screenshot of a Claude Code child failing to authenticate with no Oz-side fix was the immediate motivation.

Note that we haven't added the ability to add in a new secret via this flow yet - I'll follow up on that in another PR, but this should unblock the demo I believe. We can pop up a modal for new secret flow.

Let me know if I missed anything here - I have less context on these codepaths/product flows!

### High-level design

The selection lives in the same per-harness setting that cloud single-agent mode already uses, so picking a key from either the plan card, the confirmation card, or the existing single-agent cloud picker is unified — change it in one place and every surface reflects the choice. The picker is gated by the same condition cloud-mode applies: only shown when execution mode is Cloud and the harness has at least one supported managed-secret type. Local harness children continue to inherit credentials from the user's shell environment as they did before.

On dispatch, the selected secret flows through the existing orchestration pipeline as a Rust-only field that does not round-trip through the multi-agent proto. The orchestrate request carries it from the UI into the per-child Start-Agent dispatch, where it gets mapped into the harness-specific config shape the cloud worker already understands. No server or proto changes are needed because the cloud spawn API has supported per-harness auth secrets ever since the single-agent feature shipped — this PR just teaches the orchestration pathway to populate that field instead of leaving it empty. The persistence story is also reused: the secret name is stored client-side in the same settings entry as cloud mode, so it survives across sessions and across both UIs without introducing new state.

Two flow-control subtleties worth flagging for reviewers: the picker's action handler intentionally does not re-render its own dropdown (the user's click already updated the selection, and a self-update would trip the UI framework's circular-update guard), and the confirmation card's streaming/auto-launch paths re-seed the secret from settings after every proto-derived state rebuild so a plan-card pick survives the auto-launch handoff. Both came up during local testing and are documented inline.

## Testing

Manual:
- Created an API key for myself for both CC and Codex, tested with that - see Looms

Automated:
- Added unit tests covering: Claude/Codex auth-secret propagation through the orchestrate → start-agent conversion, whitespace-only names being normalized to None, Local mode ignoring the field.
- Updated the existing orchestration / start-agent / convert-from / block-tests test suites for the new struct field; all 131 orchestration/run_agents/start_agent tests pass.

### DIFF
--- a/app/src/ai/blocklist/inline_action/orchestration_controls.rs
+++ b/app/src/ai/blocklist/inline_action/orchestration_controls.rs
@@ -31,13 +31,15 @@ use warp_cli::agent::Harness;
 use warp_core::channel::{Channel, ChannelState};
 use warp_core::ui::theme::Fill;
 
+use crate::ai::auth_secret_types::auth_secret_types_for_harness;
 use crate::ai::cloud_agent_settings::CloudAgentSettings;
 use crate::ai::cloud_environments::CloudAmbientAgentEnvironment;
 use crate::ai::execution_profiles::model_menu_items::available_model_menu_items;
-use crate::ai::harness_availability::HarnessAvailabilityModel;
+use crate::ai::harness_availability::{AuthSecretFetchState, HarnessAvailabilityModel};
 use crate::ai::harness_display;
 use crate::appearance::Appearance;
 use crate::menu::{MenuItem, MenuItemFields};
+use crate::report_if_error;
 use crate::ui_components::blended_colors;
 use crate::view_components::dropdown::{Dropdown, DropdownAction, DropdownStyle};
 use crate::view_components::FilterableDropdown;
@@ -56,6 +58,12 @@ pub const ORCHESTRATION_PICKER_MAX_WIDTH: f32 = 205.;
 
 const DEFAULT_MODEL_LABEL: &str = "Default model";
 
+/// Label shown in the auth secret picker when no secret is selected
+/// (the child agent will inherit credentials from its environment).
+const AUTH_SECRET_INHERIT_LABEL: &str = "Inherit key from environment";
+/// Label for the auth secret column.
+pub const AUTH_SECRET_COLUMN_LABEL: &str = "API key";
+
 // ── Action trait ────────────────────────────────────────────────────
 
 /// Trait that both `RunAgentsCardViewAction` and
@@ -67,6 +75,9 @@ pub trait OrchestrationControlAction: Clone + Debug + Send + Sync + 'static {
     fn harness_changed(harness_type: String) -> Self;
     fn environment_changed(environment_id: String) -> Self;
     fn worker_host_changed(worker_host: String) -> Self;
+    /// Fires when the auth secret picker selects a managed secret.
+    /// `None` means "clear the selection / inherit from environment".
+    fn auth_secret_changed(name: Option<String>) -> Self;
 }
 
 // ── Shared edit state ───────────────────────────────────────────────
@@ -80,6 +91,12 @@ pub struct OrchestrationEditState {
     pub model_id: String,
     pub harness_type: String,
     pub execution_mode: RunAgentsExecutionMode,
+    /// Managed-secret name selected for the active harness, when the
+    /// harness is non-Oz and the execution mode is Cloud.
+    /// Persisted side-channel via `CloudAgentSettings.last_selected_auth_secret`
+    /// and propagated onto `RunAgentsRequest.harness_auth_secret_name` at
+    /// dispatch time. The proto does NOT carry this field.
+    pub auth_secret_name: Option<String>,
 }
 
 impl OrchestrationEditState {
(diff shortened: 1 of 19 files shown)


app/src/ai/blocklist/inline_action/run_agents_card_view.rs
```diff
@@ -531,6 +553,17 @@ impl RunAgentsCardView {
...
+                // harness; re-resolve the auth secret from settings so
+                // the auto-launched dispatch carries the user's
+                // plan-card pick instead of `None`.
+                if self.state.orch.auth_secret_name.is_none() {
+                    self.state.orch.auth_secret_name = oc::resolve_default_auth_secret_for_harness(
+                        &self.state.orch.harness_type,
+                        ctx,
+                    );
+                }
+
```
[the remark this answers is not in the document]
advait-m: taking a look at this + below

app/src/ai/blocklist/inline_action/orchestration_controls.rs
```diff
@@ -710,6 +735,170 @@ pub fn persist_environment_selection<V: View>(environment_id: &str, ctx: &mut Vi
...
+                        ),
+                    ));
+                }
+            }
+            AuthSecretFetchState::NotFetched | AuthSecretFetchState::Loading => {
+                items.push(MenuItem::Item(
+                    MenuItemFields::new("Loading…").with_disabled(true),
+                ));
+            }
+            AuthSecretFetchState::Failed(_) => {
```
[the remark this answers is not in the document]
advait-m: fixed - added `HarnessAvailabilityEvent::AuthSecretsFetchFailed`

app/src/ai/blocklist/inline_action/run_agents_card_view.rs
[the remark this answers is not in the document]
advait-m: fixed - removed gating

app/src/ai/harness_availability.rs
```diff
@@ -60,8 +60,24 @@ pub struct AuthSecretEntry {
...
+    /// Emitted when a lazy auth-secrets fetch fails. Subscribers should
+    /// re-render so any "Loading…" placeholders can transition to an
+    /// error state — without this signal the picker would otherwise be
+    /// stuck on the loading placeholder until the next refetch.
+    ///
+    /// `harness` is kept on the payload for future subscribers that
+    /// need to filter by harness; current subscribers ignore it.
+    AuthSecretsFetchFailed {
+        #[allow(dead_code)]
+        harness: Harness,
```
liliwilson: nit: can we remove this if we don't need it now? Can always add it back later

crates/ai/src/agent/action/mod.rs
```diff
@@ -193,6 +193,13 @@ pub struct RunAgentsRequest {
     pub execution_mode: RunAgentsExecutionMode,
     pub agent_run_configs: Vec<RunAgentsAgentRunConfig>,
     pub plan_id: String,
+    /// Name of a managed secret to forward as the authentication credential
+    /// for non-Oz cloud child agents (Claude / Codex). Resolved client-side
+    /// from `CloudAgentSettings.last_selected_auth_secret` at dispatch time;
+    /// not serialized to the wire — the field is populated by the UI on
+    /// Accept and consumed by `launch_remote_child` when building the
+    /// `AgentConfigSnapshot.harness_auth_secrets`.
```
liliwilson: mega nit but comment seems to deviate from precedent for this struct

liliwilson (approved):
Nice, this makes sense to me! Thanks for getting it out so quickly

One UI comment: 
Image
I think for the plan mode, API key should probably be just below the harness selector (or at least shouldn't be in between the model selector and the "Primary model..." subtext)

advait-m:
> Nice, this makes sense to me! Thanks for getting it out so quickly
> 
> One UI comment: Image I think for the plan mode, API key should probably be just below the harness selector (or at least shouldn't be in between the model selector and the "Primary model..." subtext)

good catch, fixed

image

crates/ai/src/agent/action/mod.rs
advait-m: agree, missed cleaning, thanks, fixing

app/src/ai/harness_availability.rs
advait-m: yep can do

PR #10889, opened 2026-05-14
[Remote envs] codebase indexing UI

moirahuang:
## Description
Add UI for remote codebase indexing repos

## Testing
Locally tested + added unit tests

### Screenshots / Videos
image (1)

## Agent Mode

### DIFF
--- a/app/src/remote_server/codebase_index_model.rs
+++ b/app/src/remote_server/codebase_index_model.rs
@@ -240,22 +240,6 @@ impl RemoteCodebaseIndexModel {
         host_id: &HostId,
         statuses: &[RemoteCodebaseIndexStatusWithPath],
     ) {
-        let status_count = statuses.len();
-        log::info!(
-            "[Remote codebase indexing] Client received bootstrap codebase index statuses snapshot: host_id={host_id} status_count={status_count}"
-        );
-        for status_with_path in statuses {
-            log::debug!(
-                "[Remote codebase indexing] Client received bootstrap codebase index status: repo_path={} state={:?} has_root_hash={}",
-                status_with_path.status.repo_path,
-                status_with_path.status.state,
-                status_with_path
-                    .status
-                    .root_hash
-                    .as_deref()
-                    .is_some_and(|root_hash| !root_hash.is_empty()),
-            );
-        }
         self.statuses.retain(|key, _| key.host_id != *host_id);
         for status_with_path in statuses {
             self.apply_status_update(
@@ -266,20 +250,6 @@ impl RemoteCodebaseIndexModel {
     }
 
     fn apply_status_update(&mut self, remote_path: RemotePath, status: RemoteCodebaseIndexStatus) {
-        log::info!(
-            "[Remote codebase indexing] Client applying codebase index status update: host_id={} state={:?} has_root_hash={}",
-            remote_path.host_id,
-            status.state,
-            status
-                .root_hash
-                .as_deref()
-                .is_some_and(|root_hash| !root_hash.is_empty()),
-        );
-        log::debug!(
-            "[Remote codebase indexing] Client applying codebase index status update: repo_path={} state={:?}",
-            status.repo_path,
-            status.state,
-        );
         self.statuses.insert(remote_path, status);
     }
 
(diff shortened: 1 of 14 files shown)


app/src/remote_server/codebase_index_model.rs
```diff
@@ -169,19 +214,22 @@ impl RemoteCodebaseIndexModel {
         match event {
             RemoteServerManagerEvent::CodebaseIndexStatusesSnapshot { host_id, statuses } => {
                 self.apply_statuses_snapshot(host_id, statuses);
+                ctx.emit(RemoteCodebaseIndexModelEvent::SettingsEntriesChanged);
```
kevinyang372: nit: for these branches, should we only emit the event when an entry is actually updated?

app/src/terminal/writeable_pty/remote_server_controller.rs
```diff
@@ -538,16 +583,52 @@ impl<T: EventLoopSender> RemoteServerController<T> {
...
+
+    connection_label_from_user_and_host(&session_info.user, host.as_deref())
+}
+
+fn connection_label_from_user_and_host(user: &str, host: Option<&str>) -> String {
+    match (user.is_empty(), host.filter(|host| !host.is_empty())) {
+        (false, Some(host)) => format!("{user}@{host}"),
+        (false, None) => user.to_string(),
+        (true, Some(host)) => host.to_string(),
+        (true, None) => String::new(),
```
kevinyang372: Should we have some better fallback if user is empty and host is None?

crates/remote_server/src/manager.rs
```diff
@@ -589,6 +589,11 @@ pub struct RemoteServerManager {
     sessions: HashMap<SessionId, RemoteSessionState>,
     /// Reverse index: host → sessions for O(1) lookup by `HostId`.
     host_to_sessions: HashMap<HostId, HashSet<SessionId>>,
+    /// User-facing connection labels by host, used by host-scoped UI.
+    host_labels: HashMap<HostId, String>,
```
kevinyang372: Why do we need to cache both host and session labels?

app/src/settings_view/code_page.rs
```diff
@@ -131,15 +139,39 @@ struct LspServerRowMouseStates {
...
+struct IndexingStatusPresentation {
+    text: Cow<'static, str>,
+    color: ColorU,
+    icon: Option<Icon>,
+    show_retry: bool,
+    show_delete: bool,
+}
+
+#[derive(Clone)]
+enum IndexingActionTarget {
```
kevinyang372: This is just LocalOrRemotePath -- we shouldn't need another enum

app/src/settings_view/code_page.rs
```diff
@@ -523,6 +583,10 @@ pub enum CodeSettingsPageAction {
     ToggleAutoIndexing,
     ManualResync(PathBuf),
```
kevinyang372, line 580: Can these methods be `ManualResync(LocalOrRemotePath)` and `DeleteIndex(LocalOrRemotePath)`

app/src/settings_view/code_page.rs
```diff
@@ -1398,6 +1509,61 @@ impl CodePageWidget {
             .finish()
     }
 
+    #[cfg(not(target_family = "wasm"))]
+    fn render_remote_workspace_row(
```
kevinyang372, line 1488: On a high-level, we should make sure as much of this logic between remote and local workspace row

The difference I could tell is:
1. Remote workspace has a different label
2. It doesn't render some actions (e.g. open project context)

kevinyang372 (approved):
Approving to unblock

app/src/settings_view/code_page.rs
moirahuang: refactored to share more logic

crates/remote_server/src/manager.rs
moirahuang: removed cache of host

moirahuang:
i'm going to wait to merge this because it's currently stacked on top of the persistence PR and i don't want to make that PR harder to review

PR #10890, opened 2026-05-14
Orchestration pill bar for shared session viewers

cephalonaut:
## Description

Adds an orchestration pill bar to shared-session viewers (native + web). A viewer joining a shared orchestrated cloud run can now see the orchestrator and each child agent as a pill in the agent view header and switch between their transcripts.

https://www.loom.com/share/d20056325a65450f8a257428f7d8926c

### Why

Before this PR, opening a shared session for an orchestrated cloud agent run rendered only the orchestrator's transcript with no way to discover or navigate to child agents. The pill bar already worked for local orchestration; this extends it to the shared-session viewer path. Spec: `specs/orch-pill-bar-web/`.

### Mechanism

**`OrchestrationViewerModel`** (new, gated by `FeatureFlag::OrchestrationViewerPillBar`) polls `GET /agent/runs?ancestor_run_id={task_id}` for the orchestrator's descendants and registers each one as a child conversation in `BlocklistAIHistoryModel`. The existing `OrchestrationPillBar` view already renders from this model, so the pill UI requires no new rendering code.

- Two-tier polling: ~5s while any child is non-terminal, ~30s once all are terminal. An `AppendedExchange` history event kicks the cadence back to fast so follow-up input that spawns new children surfaces promptly.
- `fetch_generation` counter drops stale responses when a fresh kick fetch races a timer-fired one. Each `schedule_next_poll` explicitly aborts the prior `SpawnedFutureHandle` (which doesn't abort on drop) so timer chains don't stack.

**Per-child hidden shared-session panes.** When a child first reports a `session_id`, the model emits `Event::EnsureSharedSessionViewerChildPane` exactly once, and `PaneGroup` materializes a hidden pane with its own `TerminalView`, `BlocklistAIController`, `TerminalManager`, and viewer-side `Network` joining the child's shared session. This mirrors local orchestration's remote-child architecture and avoids the single-stream `BlocklistAIController` cross-talk that a parent-routed approach would suffer.

**Click path.** Pill clicks dispatch `TerminalAction::SwitchAgentViewToConversation`, which `TerminalView` translates to `Event::SwapPaneToConversation`. `PaneGroup` swaps the visible pane to the corresponding hidden child pane — same mechanism the native pill bar already uses for local children. If the user clicks before the child has a `session_id`, a loading-placeholder pane renders (with the pill bar visible so they can navigate back); it's replaced by the real shared-session viewer pane once the next poll surfaces the `session_id`.

**Lifecycle.** The model lives on the parent viewer's `TerminalManager`, created in the `JoinedSuccessfully` arm of the network subscription. Gated by `enable_orchestration_polling: bool` so only root viewer panes poll — per-child viewer panes skip the model since the parent's transitive `ancestor_run_id` filter already covers grandchildren. Torn down via `stop_orchestration_polling` from `NetworkEvent::SessionEnded` (non-ambient or non-owner ambient), `ViewerRemoved`, and `FailedToReconnect`.

**WASM compatibility.** Pane-management helpers in `orchestration_pill_bar.rs` that depend on `WorkspaceRegistry`/`PaneGroup` are now `#[cfg(not(target_family = "wasm"))]`. The 3-dot overflow menu and its actions are suppressed on WASM (single-pane viewer has no use for them).

**Feature flag.** New `FeatureFlag::OrchestrationViewerPillBar` added to `DOGFOOD_FLAGS` and `PREVIEW_FLAGS`. Independent of the existing `OrchestrationPillBar` flag, which still gates only the local-orchestration pill bar.

## Testing

Adds `App::test` integration tests for `OrchestrationViewerModel::apply_children_fetch` covering new child registration, parent-task-id skip, no-active-parent retry, status transitions, the materialization gate, session_id transition, and multi-child fetches. Existing pure-function tests for `conversation_status_from_state` preserved.

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/04e324bc-1e26-4355-8b19-32349c540dc2_
_Run: https://oz.staging.warp.dev/runs/019e1fbc-66cf-7bb3-b5d6-ff532d376328_

### DIFF
--- a/app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs
+++ b/app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs
@@ -96,14 +96,23 @@ pub struct OrchestrationViewerModel {
     terminal_view: WeakViewHandle<TerminalView>,
     /// Known child agents indexed by server-side task id.
     children: HashMap<AmbientAgentTaskId, ChildAgentEntry>,
-    /// Handle for the next scheduled poll. Replaced (and implicitly
-    /// cancelled) by every [`Self::schedule_next_poll`] call, and cleared
-    /// while a kick-induced fetch is in flight — the fetch's response
-    /// callback schedules a fresh timer using the freshest child state.
-    /// The timer chain itself runs for the lifetime of this model; the
-    /// model is dropped when the parent pane's `TerminalManager` is torn
-    /// down (closing the orchestrator's shared session).
+    /// Handle for the next scheduled poll. Aborted and replaced by every
+    /// [`Self::schedule_next_poll`] call so we never have more than one
+    /// timer chain in flight, and cleared while a kick-induced fetch is
+    /// in flight — the fetch's response callback schedules a fresh timer
+    /// using the freshest child state. The timer chain itself runs for
+    /// the lifetime of this model; the model is dropped when the parent
+    /// pane's `TerminalManager` is torn down (closing the orchestrator's
+    /// shared session).
     polling_handle: Option<SpawnedFutureHandle>,
+    /// Monotonic counter incremented before every `fetch_children`
+    /// dispatch. The response callback compares the captured generation
+    /// against this field and drops stale responses, preventing two
+    /// concurrent fetches (typically a timer-fired fetch racing a kick-
+    /// induced fetch) from clobbering each other. Without this guard a
+    /// late-arriving stale snapshot could revert a child's status from
+    /// the freshly-applied state to an older one.
+    fetch_generation: u64,
 }
 
 impl Entity for OrchestrationViewerModel {
@@ -135,12 +144,13 @@ impl OrchestrationViewerModel {
             me.maybe_kick_polling(event, ctx);
         });
 
-        let model = Self {
+        let mut model = Self {
             parent_task_id,
             terminal_view_id,
             terminal_view,
             children: HashMap::new(),
             polling_handle: None,
+            fetch_generation: 0,
         };
 
         // Kick off the recurring poll. `fetch_children`'s response
@@ -158,6 +168,15 @@ impl OrchestrationViewerModel {
     /// terminal state. Polling never stops on its own — see the field doc
     /// on [`Self::polling_handle`] for the lifecycle.
     fn schedule_next_poll(&mut self, ctx: &mut ModelContext<Self>) {
+        // Abort any prior timer before replacing it. `SpawnedFutureHandle`
+        // does NOT abort on drop, so without this call the previous
+        // timer's continuation could still fire and dispatch an extra
+        // `fetch_children`, eventually multiplying into parallel timer
+        // chains.
+        if let Some(prior) = self.polling_handle.take() {
+            prior.abort();
+        }
+
         let all_terminal = !self.children.is_empty()
             && self
                 .children
@@ -236,11 +255,15 @@ impl OrchestrationViewerModel {
         log::info!(
             "[orch-viewer] kick polling: AppendedExchange for tracked conv={conversation_id:?}"
         );
-        // Cancel the pending timer; `fetch_children`'s response callback
+        // Abort the pending timer; `fetch_children`'s response callback
         // will queue a fresh one using the post-fetch state, so a kick
         // landing during the idle (30s) interval correctly tightens back
         // to the fast (5s) interval as soon as the new child surfaces.
-        self.polling_handle = None;
+        // Aborting (rather than just dropping the handle) is required
+        // because `SpawnedFutureHandle` does not cancel on drop.
+        if let Some(prior) = self.polling_handle.take() {
+            prior.abort();
+        }
         self.fetch_children(ctx);
     }
 
@@ -251,7 +274,15 @@ impl OrchestrationViewerModel {
     /// preferable to surfacing a failure modal because orchestrations that
     /// haven't spawned any children yet legitimately return an empty list,
     /// and transient REST errors should not break the viewer.
-    fn fetch_children(&self, ctx: &mut ModelContext<Self>) {
+    fn fetch_children(&mut self, ctx: &mut ModelContext<Self>) {
+        // Stamp the generation BEFORE dispatching so any in-flight stale
+        // fetch (a timer-fired fetch whose REST request hasn't returned
+        // yet, racing this kick) is invalidated when its response
+        // callback compares against the freshly-bumped
+        // `self.fetch_generation`.
+        self.fetch_generation = self.fetch_generation.wrapping_add(1);
+        let fetch_generation = self.fetch_generation;
+
         let ai_client = ServerApiProvider::as_ref(ctx).get_ai_client();
         let filter = TaskListFilter {
             ancestor_run_id: Some(self.parent_task_id.to_string()),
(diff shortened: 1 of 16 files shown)


app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs
```diff
@@ -0,0 +1,412 @@
...
+        AmbientAgentTaskState::Succeeded => ConversationStatus::Success,
+        AmbientAgentTaskState::Failed | AmbientAgentTaskState::Error => ConversationStatus::Error,
+        AmbientAgentTaskState::Blocked => ConversationStatus::Blocked {
+            blocked_action: String::new(),
+        },
+        AmbientAgentTaskState::Cancelled => ConversationStatus::Cancelled,
+        // The `Unknown` variant only exists for forward-compatibility with
+        // newer server states. Treat it as still working so the pill bar
+        // doesn't prematurely commit to a final badge.
+        AmbientAgentTaskState::Unknown => ConversationStatus::InProgress,
```
advait-m: optional nit: should we align this method in `app/src/ai/ambient_agents/task.rs (461-474)` (came up with agent review)? or different semantically?

```
pub fn is_terminal(&self) -> bool {
    match self {
        AmbientAgentTaskState::Succeeded
        | AmbientAgentTaskState::Failed
        | AmbientAgentTaskState::Error
        | AmbientAgentTaskState::Blocked
        | AmbientAgentTaskState::Cancelled
        | AmbientAgentTaskState::Unknown => true,   // ← Unknown is terminal
        AmbientAgentTaskState::Queued
        | AmbientAgentTaskState::Pending
        | AmbientAgentTaskState::Claimed
        | AmbientAgentTaskState::InProgress => false,
    }
}
```

specs/orch-pill-bar-web/PRODUCT.md
```diff
@@ -0,0 +1,90 @@
...
+    - **Blocked**: pause icon.
+    - **Cancelled**: cancelled icon.
+12. The orchestrator pill does not display a status badge.
+13. Status badges update as the sharer's session progresses (live sessions). A brief delay (a few seconds) between a child completing and its badge updating is acceptable. For completed sessions, badges show the final state immediately.
+
+### Pill click interaction
+14. Clicking a non-selected pill switches the agent view in-place to display that agent's conversation transcript. The clicked pill becomes the selected pill. The pill bar remains visible.
+15. Clicking the currently-selected pill is a no-op.
+16. Clicking the orchestrator pill while viewing a child conversation switches back to the orchestrator's conversation. This is the primary "back" navigation.
+17. On web, there is no 3-dot overflow menu on pills. Pane-management actions (open in new pane, open in new tab, focus pane) are not available since the web viewer is single-pane and read-only. On native, the overflow menu and pane-management actions behave the same as for local orchestration.
```
advait-m, line 45: spoke with daniel and the reason Warp on Web doesn't have pane splitting is to prevent the ability to ever spawn a new terminal session. In this case though, it doesn't really apply, since we're swapping to a hidden terminal view i.e. existing session.

i.e. it isn't a fundamental can't do multiple panes limitation

Anyhow, I think this isn't a blocker - we can merge as is, but perhaps something to file a Linear in backlog for, since I think this should actually be possible on web for things like "open in new pane" (if my understanding is right)?

specs/orch-pill-bar-web/PRODUCT.md
```diff
@@ -0,0 +1,90 @@
...
+28. When viewing the orchestrator conversation, inline orchestration cards (run_agents confirmation, lifecycle status) render as they appear in the orchestrator's transcript (these flow through the existing session WebSocket).
+29. Clicking a child pill joins that child's session. For a live child, new messages stream in after the initial replay completes. For a terminal child, the full transcript is replayed and the session completes.
+30. If the sharer's session ends while the viewer is watching, the pill bar remains visible with the last known statuses.
+
+### Completed session behavior
+31. When a viewer opens a completed orchestrated session, the full orchestration structure (orchestrator + all children with final statuses) is available immediately.
+32. The viewer can click through all child conversations to review their work.
+
+### Horizontal overflow
+33. When there are more child pills than fit in the available width, overflow pills are clipped at the pane boundary, matching the native pill bar's current behavior.
```
advait-m, line 77: yeah, this will be improved by horizontal scrolling PR

specs/orch-pill-bar-web/PRODUCT.md
advait-m: it is probably more involved since we might need to remove some WASM gating too

specs/orch-pill-bar-web/TECH.md
```diff
@@ -0,0 +1,300 @@
...
+
+**Cons:**
+- Repos touched: **4** (session-sharing-protocol, warp-proto-apis, warp-server, warp client).
+- Complex server plumbing: forwarding child events to parent session, lifecycle notifier integration.
+- Event volume scales with child count.
+- `session-sharing-protocol` is an external repo pinned by git rev.
+
+### Recommendation
+
+Option A for V1, with each child hosted in its own hidden shared-session pane (see Layer 2). The polling latency is acceptable for viewing sessions (status badges don't need sub-second updates), and touching 1-2 repos vs 4 significantly reduces implementation scope. Option B can be layered on later if real-time fidelity becomes important.
```
advait-m, line 123: yeah, that makes sense to me

crates/warp_features/src/lib.rs
```diff
@@ -942,17 +948,13 @@ pub const DOGFOOD_FLAGS: &[FeatureFlag] = &[
...
 ];
 
 /// Features enabled for feature preview build users (e.g.: Friends of Warp).
 /// All PREVIEW_FLAGS are also automatically added to dogfood builds (WarpDev).
 pub const PREVIEW_FLAGS: &[FeatureFlag] = &[
     FeatureFlag::Orchestration,
-    FeatureFlag::OrchestrationV2,
-    FeatureFlag::OrchestrationPillBar,
-    FeatureFlag::RunAgentsTool,
+    FeatureFlag::OrchestrationViewerPillBar,
```
advait-m, line 964: agree we can get this to preview asap :+1:

advait-m (approved):
this is so awesome!!! 🚀 

couple comments, no blockers

app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs
advait-m: specifically potential user-facing impact, but it's pretty niche case: 

`OrchestrationViewerModel::schedule_next_poll uses is_terminal()` to decide between 5s fast cadence and 30s idle cadence. Since `is_terminal()` treats `Unknown` as terminal but this mapping treats it as InProgress, a child stuck in `Unknown` would show the spinner while polling drops to 30s - so transitions out of Unknown would lag by up to ~30s in the UI. Could either align both to terminal (and pick a final badge here), or keep `Unknown → InProgress` here and use a local helper for the polling check.

but it's probably fine to ignore this for now, can tackle if folks actually run into it

app/src/terminal/view/pane_impl.rs
```diff
@@ -507,7 +507,15 @@ impl TerminalView {
...
         // list a second time alongside the orchestrator's own pill bar.
-        if FeatureFlag::OrchestrationPillBar.is_enabled()
+        //
+        // `OrchestrationViewerPillBar` is the parallel flag for shared
+        // session viewers (web + native). Children are registered via the
+        // REST data fetch in `OrchestrationViewerModel`; when none have
+        // arrived yet, `OrchestrationPillBar::pill_specs` returns `None`
+        // and the pill bar's `render` short-circuits to `Empty`, so the
+        // gate here is intentionally permissive.
+        if (FeatureFlag::OrchestrationPillBar.is_enabled()
```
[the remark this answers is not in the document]
advait-m: I think this is fine, I don't think we'll be unshipping pill bar to preview

app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs
cephalonaut: I aligned it for now - wasn't really a conscious choice, so better to have no surprises and address impact if it arises

specs/orch-pill-bar-web/PRODUCT.md
cephalonaut: Oh good to know - I'll file an issue

app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs
advait-m: sg!

specs/orch-pill-bar-web/PRODUCT.md
advait-m: :+1:

PR #10930, opened 2026-05-14
Update pr chip and pr button to have a single source of truth

MaggieShan:
## Description
* Adds `PrInfo` as a stored field in `GitRepoStatusModel` so that both pr chip and button in the code review can pull the data from one global singleton model 
  * Compared to the git chip policy that runs every 30 sec, this runs `gh pr view` every 60 sec
  * We similarly refresh pr info after a gh or gt command has run 
  * We store a set of the specific consumers for pr info to ensure we fetch pr info only when needed
  * Removes the shell builtin fallback
* Context: https://warpdev.slack.com/archives/C08KTPNQN65/p1778107435756519 
* Note: supporting the pr chip and git ops buttons remotely is tracked and will be a follow up

## Linked Issue
https://linear.app/warpdotdev/issue/APP-4397/fix-graphql-rate-limits

## Testing
https://www.loom.com/share/792bccfcefcf4439b404e9b1d56d774b

## Agent Mode


app/src/remote_server/diff_state_proto.rs
```diff
@@ -425,7 +416,9 @@ impl From<&DiffMetadata> for proto::DiffMetadata {
             has_head_commit: m.has_head_commit,
             unpushed_commits: m.unpushed_commits.iter().map(proto::Commit::from).collect(),
             upstream_ref: m.upstream_ref.clone(),
-            pr_info: m.pr_info.as_ref().map(proto::PrInfo::from),
+            // PR info is fetched independently per repo by GitRepoStatusModel
+            // and is not part of the diff state. Always None on the wire.
+            pr_info: None,
```
MaggieShan, line 421: This will be addressed in https://linear.app/warpdotdev/issue/APP-4512/wire-up-git-ops-with-remote-code-review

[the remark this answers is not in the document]
MaggieShan: KI and will be addressed later: https://github.com/warpdotdev/warp/pull/10930#discussion_r3276815545

app/src/code_review/git_status_update.rs
```diff
@@ -125,13 +140,25 @@ pub struct GitRepoStatusModel {
     subscriber_id: Option<SubscriberId>,
     metadata: Option<GitStatusMetadata>,
     computing_metadata_abort_handle: Option<SpawnedFutureHandle>,
+    computing_pr_info_abort_handle: Option<SpawnedFutureHandle>,
+    /// Branch name that the in-flight `refresh_pr_info` is fetching for.
+    /// Used to make `refresh_pr_info` idempotent: while a fetch for the
+    /// current branch is in flight, additional calls are no-ops.
+    refreshing_pr_info_branch: Option<String>,
```
kevinchevalier: `refreshing_pr_info` and `computing_pr_info_abort_handle` seem to be closely linked. We shouldn't have the handle set without the branch. Can we combine these into a new private struct?

app/src/code_review/git_status_update.rs
```diff
@@ -317,11 +561,32 @@ impl GitRepoStatusModel {
...
+            .map(|m| m.current_branch_name.clone());
+        let current_branch = metadata.as_ref().map(|m| m.current_branch_name.clone());
         self.metadata = metadata;
         ctx.emit(GitRepoStatusEvent::MetadataChanged);
+        if previous_branch != current_branch && self.pr_info.take().is_some() {
+            ctx.emit(GitRepoStatusEvent::PrInfoChanged);
+        }
+    }
+
+    pub(crate) fn set_pr_info_for_test(
```
kevinchevalier, line 589: Can we add a #cfg for test here?

kevinchevalier (approved):
This looks good. Did you consider creating a new mirror model just for this PR information?

I assume that you didn't go that route because the PR metadata depends on the diff metadata (branch)?

If we did that, we wouldn't need to register subscribers to just the PR info, which is a little brittle.

app/src/code_review/git_status_update.rs
MaggieShan: Done!

MaggieShan:
> This looks good. Did you consider creating a new mirror model just for this PR information?
> 
> I assume that you didn't go that route because the PR metadata depends on the diff metadata (branch)?
> 
> If we did that, we wouldn't need to register subscribers to just the PR info, which is a little brittle.

Ya my original thinking was to keep all relevant information in one model especially since PR info depends on the model's metadata, but I agree that the current logic to register subscribers isn't great - can refactor as a follow up!

PR #10832, opened 2026-05-13
Integrate Oz changelog skill into create_release workflow

vikvang:
## Description
Replace the `warpdotdev/generate-changelog` GitHub Action in the `generate_changelogs` job with the Oz changelog-draft skill (`oz-agent-action`). This makes the release changelog generation AI-powered, using the same skill that was landed in #10280.

### What changed
- **SKILL.md**: Step 8 now produces a third output file `changelog-release.json` with the `{newFeatures, improvements, bugFixes, images, oz_updates}` schema expected by the Slack payload builder and in-app changelog.json steps.
- **create_release.yml**: The `generate_changelogs` job now:
  1. Runs `oz-agent-action` with a prompt to follow the changelog-draft skill workflow
  2. A bridge step reads `changelog-release.json` and sets `outputs.changelog` so all downstream steps (Slack post, GCS upload) work unchanged
  3. Fixes `.image` to `.images` key in the changelog.json builder for schema consistency

### How it works
The Oz agent runs the full skill workflow (fetch PRs, classify contributors, extract feature flags, classify unmarked PRs, assemble draft) and writes three files:
- `changelog-draft.md` — human-reviewable markdown
- `changelog-draft.json` — machine-readable audit artifact
- `changelog-release.json` — release-pipeline-compatible JSON (new)

The bridge step loads `changelog-release.json` into `steps.generate_changelog.outputs.changelog`, maintaining the same interface the downstream Slack and GCS steps expect.

## Linked Issue
- N/A (continuation of #10280)

## Testing
- Validated YAML structure (1788 lines, all key references intact)
- Generated a real `changelog-release.json` from the latest stable release range (v0.2026.05.06 → v0.2026.05.13): 2 new features, 10 improvements, 27 bug fixes, 1 image
- Ran the exact downstream jq transforms (Slack payload builder + changelog.json builder) against the generated JSON — all produce valid output
- Full end-to-end test will occur on the next release cut

- [ ] I have manually tested my changes locally with `./script/run`

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

Co-Authored-By: Oz <oz-agent@warp.dev>

PR #10940, opened 2026-05-14
Use legacy changelog action for non-stable channels

vikvang:
## Summary

Fixes the dev in-app changelog regression from #10832 without re-running the Oz agent on every non-stable cut.

- **Stable**: unchanged — keeps the new Oz changelog-draft agent flow.
- **Dev / preview / beta**: restores the legacy `warpdotdev/generate-changelog` action, so a fresh `changelog.json` is uploaded to GCS for every cut.
- A small bridge step picks whichever path ran and re-emits the result as `steps.generate_changelog.outputs.changelog`, so the downstream Slack post + GCS upload steps work unchanged.

## Background

#10832 gated `generate_changelogs` to stable-only to save agent tokens. That broke the dev in-app changelog: the dev client fetches `changelog.json` from GCS first (`app/src/autoupdate/changelog.rs:84` — `should_fetch_changelog_json` is `true` only for `Channel::Dev`) and falls back to `channel_versions.json`'s `changelogs.dev.latest` on 404. Skipping the upload for dev releases caused the client to surface a 2021-era `[TEST]` fixture from `channel_versions.json` in the in-app \"What's new in Oz\" section.

Preview/stable read straight from `channel_versions.json` (no GCS), so they were never exposed to the bug. The follow-up cleanup of the stale `changelogs.dev.latest` fixture in `warpdotdev/channel-versions` is in flight separately and is good hygiene regardless.

## Why hybrid instead of \"new flow everywhere\"

- Saves ~27-min agent invocation × daily dev cuts + preview/beta RCs
- Dogfooding signal on the Oz flow is preserved on stable cuts (where output quality matters most) plus the existing `workflow_dispatch` `changelog_draft.yml` for ad-hoc runs
- Existing battle-tested generate-changelog action keeps running for the channels users consume daily

## Test plan

- [ ] Watch the next dev release cut and confirm `warp-releases/dev/<release_tag>/changelog.json` is uploaded
- [ ] Confirm dev client renders the new release notes (not the `[TEST]` fixtures)
- [ ] Confirm next stable cut still runs through Oz and produces a sensible Slack post + GCS changelog.json
- [ ] Confirm preview RC cut runs the legacy path and posts to Slack as before

### DIFF
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -1619,7 +1619,7 @@ jobs:
 
   generate_changelogs:
     name: Generate Changelogs
-    if: ${{ needs.prepare_release.outputs.should_publish == 'true' && inputs.channel == 'stable' }}
+    if: ${{ needs.prepare_release.outputs.should_publish == 'true' }}
     runs-on: ubuntu-latest
     needs:
       - prepare_release
--- a/.github/workflows/create_release.yml
+++ b/.github/workflows/create_release.yml
@@ -1646,7 +1646,12 @@ jobs:
           config_file: ${{ env.CONFIG_FILE }}
           channel: ${{ inputs.channel }}
 
-      - name: Generate changelog via Oz
+      # Stable releases use the Oz changelog-draft agent for higher-quality,
+      # human-reviewable output. Non-stable channels (dev/preview/beta) use the
+      # legacy generate-changelog action to avoid spending Oz agent tokens on
+      # daily dev cuts and preview RCs.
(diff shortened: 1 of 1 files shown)


danielpeng2 (commented):
Can we use the old (non-agent) changelog generation step for non-stable runs, instead of running the new agent flow?

vikvang:
Sure, I updated here as suggested. Non-stable channels use the old changelog action

danielpeng2 (approved):

PR #10960, opened 2026-05-14
Update orchestration auth secret picker copy to 'Skip setting an API key'

advait-m:
## Description
Changes the orchestration auth-secret picker copy from "Inherit key from environment" → "Skip setting an API key".

Per Slack discussion (#10885 follow-up): "Inherit key from environment" was unclear about the fact that no env var is actually injected by default — users were clicking through expecting credentials to be auto-resolved. "Skip setting an API key" makes the no-op nature of the choice obvious in the small dropdown.

Cloud-mode FTUX (`auth_secret_selector.rs`) keeps its existing copy since its richer layout has supporting sidecar text that explains the behavior. We can iterate from there if needed.

## Linked Issue
N/A — copy-only follow-up to #10885.

## Testing
String-only change; verified `cargo check -p warp --bin warp` succeeds. The label is also referenced as the selection-restore name in `sync_picker_selections`, and that path uses the same `AUTH_SECRET_INHERIT_LABEL` constant, so the picker still selects the right entry after the rename.

### Screenshots / Videos
_to be attached_

## Agent Mode

[Warp Conversation](https://staging.warp.dev/conversation/e9eeca41-8ef1-4128-a9d8-e1cdd248c5ce)

### DIFF
--- a/app/src/ai/blocklist/inline_action/orchestration_controls.rs
+++ b/app/src/ai/blocklist/inline_action/orchestration_controls.rs
@@ -64,7 +64,7 @@ const DEFAULT_MODEL_LABEL: &str = "Default model";
 
 /// Label shown in the auth secret picker when no secret is selected
 /// (the child agent will inherit credentials from its environment).
-const AUTH_SECRET_INHERIT_LABEL: &str = "Inherit key from environment";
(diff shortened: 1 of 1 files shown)


advait-m:
@liliwilson addressed this in another PR

PR #10969, opened 2026-05-14
Don't wait for file-based MCPs we never attempt to start

seemeroland:
## Description

In an oz run, we were timing out after waiting 60s for discovered file-based MCP servers to reach a terminal status.

We intentionally do not start project-scoped file-based MCP servers for security reasons in [this PR](https://github.com/warpdotdev/warp-internal/pull/24640)

But the agent runner was still discovering these and waiting for them to reach a terminal status, even though we never attempt to start them

Now `CloudEnvMcpScanComplete` contains all detected file-based mcp servers, but also notes which we will actually attempt to auto-start. We only wait for those. Did some refactoring to make clear we don't auto-start every file-based mcp server we detect.

Also add some logging and reduce the timeout to 20s.

### DIFF
--- a/app/src/ai/mcp/file_based_manager.rs
+++ b/app/src/ai/mcp/file_based_manager.rs
@@ -200,7 +200,7 @@ impl FileBasedMCPManager {
             scanned_servers.insert(hash);
         }
 
-        let auto_started_uuids = self.auto_start_file_based_servers(servers_to_spawn, ctx);
+        let auto_started_uuids = self.maybe_autostart_file_based_servers(servers_to_spawn, ctx);
         self.pending_scan_auto_started_servers_by_root
             .entry(root_path.clone())
             .or_default()
@@ -296,7 +296,8 @@ impl FileBasedMCPManager {
         }
     }
 
-    fn auto_start_file_based_servers(
+    /// Returns the UUIDs of servers that were actually auto-started.
+    fn maybe_autostart_file_based_servers(
         &mut self,
         servers_to_consider: Vec<TemplatableMCPServerInstallation>,
         ctx: &mut ModelContext<Self>,
--- a/app/src/ai/mcp/file_based_manager.rs
+++ b/app/src/ai/mcp/file_based_manager.rs
@@ -285,14 +285,22 @@ impl FileBasedMCPManager {
         warp_managed_mcp_config_path().is_some_and(|path| root_path == path.root_path.as_path())
     }
     fn auto_start_decision(&self, hash: u64, file_based_mcp_enabled: bool) -> AutoStartDecision {
-        let is_global_warp_server = self.is_global_warp_server(hash);
-        let is_global_server = self.is_global_server(hash);
-        if is_global_warp_server || (is_global_server && file_based_mcp_enabled) {
-            AutoStartDecision::AutoStart
-        } else if is_global_server {
-            AutoStartDecision::GlobalDisabled
+        let server_type = if self.is_global_warp_server(hash) {
+            FileBasedMCPServerType::GlobalWarp
+        } else if self.is_global_server(hash) {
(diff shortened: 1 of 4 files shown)


app/src/ai/mcp/templatable_manager/native.rs
```diff
@@ -1679,15 +1678,25 @@ impl TemplatableMCPServerManager {
...
+            }
+            if self.spawned_servers.contains_key(&uuid) {
+                log::info!(
+                    "Skipping file-based MCP server '{server_name}' ({uuid}); already starting"
+                );
+                continue;
+            }
+            log::info!("Spawning file-based MCP server '{server_name}' ({uuid})");
+            new_installations.push(installation.clone());
+        }
```
vkodithala, line 1699: This is a nice improvement! Thanks for making this more verbose.

app/src/ai/mcp/file_based_manager.rs
```diff
@@ -25,6 +25,10 @@ pub struct FileBasedMCPManager {
     file_based_servers: HashMap<u64, TemplatableMCPServerInstallation>,
     /// Reverse mapping: logical root path → provider → set of server hashes.
     file_based_servers_by_root: HashMap<PathBuf, HashMap<MCPProvider, HashSet<u64>>>,
+    /// UUIDs that were actually auto-start requested while parsing each `(root, provider)`.
+    /// They are temporarily stored here and removed to emit FileBasedMCPManagerEvent::CloudEnvMcpScanComplete
+    pending_scan_auto_started_servers_by_root:
+        HashMap<PathBuf, HashMap<MCPProvider, HashSet<Uuid>>>,
```
vkodithala, line 31: nit: This is spaced weirdly. Should the `HashMap` declaration be inline?

app/src/ai/mcp/file_based_manager.rs
```diff
@@ -293,47 +313,85 @@ impl FileBasedMCPManager {
...
             };
-            if self.is_global_warp_server(hash) || (self.is_global_server(hash) && mcp_enabled) {
-                to_spawn.push(installation);
+            let installation_uuid = installation.uuid();
+            let server_name = installation.templatable_mcp_server().name.clone();
+            match self.auto_start_decision(hash, mcp_enabled) {
+                AutoStartDecision::AutoStart => {
+                    log::info!(
+                        "Auto-spawning file-based MCP server '{server_name}' ({installation_uuid})"
+                    );
```
vkodithala: nit: This might get noisy, especially for local logs. I think we already have sufficient signal from logs you included in `driver.rs` to enumerate which servers were auto-started, is there any functional reason to keep this? Or maybe just make this a `log::debug`?

app/src/ai/mcp/file_based_manager.rs
vkodithala, line 421: Can we reuse `auto_start_decision` here? Think this is a nice utility and could enable some code reuse.

vkodithala (approved):
LGTM! Left a few comments, nothing blocking. Thanks for the fix!

app/src/ai/mcp/file_based_manager.rs
seemeroland: Are we autospawning servers that often? I thought it would be useful to know locally whenever mcp servers are being spawned, since we don't seem to have that now

app/src/ai/mcp/file_based_manager.rs
seemeroland: `cargo fmt` wants it separated like this :/

app/src/ai/mcp/file_based_manager.rs
seemeroland: Not as is, but I think I can refactor `auto_start_decision` to be more reusable and use it here. This path specifically looks for third party global servers to start

PR #10984, opened 2026-05-15
MCP Oauth Fix: Create authenticated client before refreshing token

peicodes:
On app restart, Warp seeds rmcp with persisted OAuth credentials and immediately calls get_access_token() before reconfiguring the OAuth client from the stored client_id/client_secret. 

If the cached access token is expired or near expiry, rmcp tries to refresh, but refresh requires a configured OAuth client, so refresh fails and Warp falls back to interactive reauth.

Code from rmcp at the start of `refresh_token()`:
```
let oauth_client = self
    .oauth_client
    .as_ref()
    .ok_or_else(|| AuthError::InternalError("OAuth client not configured".to_string()))?;
```

Instead, we should start by creating a configured OAuth client, so that rmcp can successfully refresh the token if needed.

To aid with review, **ignore the last commit**. That just moves tests to its own file.


app/src/ai/mcp/templatable_manager/oauth_tests.rs
```diff
@@ -0,0 +1,194 @@
...
+    let config = persisted_oauth_client_config(
+        "client-id".to_string(),
+        Some("client-secret".to_string()),
+        "warp://callback",
+    );
+
+    assert_eq!(config.client_id, "client-id");
+    assert_eq!(config.client_secret, Some("client-secret".to_string()));
+    assert_eq!(config.redirect_uri, "warp://callback");
+}
```
vorporeal: i'm not sure these tests provide much value, especially given that restructuring the logic in the other file makes things harder to understand vs. keeping it closer to the way it was.

app/src/ai/mcp/templatable_manager/oauth.rs
```diff
@@ -244,13 +271,18 @@ pub async fn make_authenticated_client(
 
     // If we have a valid access token (or successfully refreshed a valid refresh token),
     // we're already authorized and good to go.
-    if auth_manager.get_access_token().await.is_ok() {
-        if let (Some(client_id), Some(client_secret)) = (client_id, client_secret) {
-            auth_manager.configure_client(
-                OAuthClientConfig::new(client_id, redirect_uri.clone())
-                    .with_client_secret(client_secret),
-            )?;
+    if let Some(client_id) = persisted_client_id {
```
vorporeal: ah shit i think the issue here is that in one of my rewrites of this logic, i accidentally dropped the call to [`initialize_from_store()`](https://docs.rs/rmcp/latest/rmcp/transport/auth/struct.AuthorizationManager.html#method.initialize_from_store).

if it solves the issues here, i think a better fix would be to replace the explicit `configure_auth_manager...` logic with something like:

```rust
if auth_manager.initalize_from_store()? {
  if auth_manager.get_access_token().await.is_ok() {
    ...
  }
}
```

i'd rather us use the upstream helper functions where possible; they should better ensure everything we're doing is correct if the library internals change.

app/src/ai/mcp/templatable_manager/oauth_tests.rs
peicodes: Oh. I thought we specifically wanted tests to be in its own file and was going to start linting against it

app/src/ai/mcp/templatable_manager/oauth.rs
peicodes: That makes sense, I'll try that

app/src/ai/mcp/templatable_manager/oauth_tests.rs
vorporeal: yes definitely!  i meant just the two tests that were added in this PR

app/src/ai/mcp/templatable_manager/oauth.rs
vorporeal: ok, i did a little manual testing and this minimal patch seems to be sufficient to fix a reauth issue i was seeing on WarpLocal:

```diff
diff --git a/app/src/ai/mcp/templatable_manager/oauth.rs b/app/src/ai/mcp/templatable_manager/oauth.rs
index 63ed456b..d6c05314 100644
--- a/app/src/ai/mcp/templatable_manager/oauth.rs
+++ b/app/src/ai/mcp/templatable_manager/oauth.rs
@@ -242,9 +242,11 @@ pub async fn make_authenticated_client(
     )
     .await;
 
-    // If we have a valid access token (or successfully refreshed a valid refresh token),
-    // we're already authorized and good to go.
-    if auth_manager.get_access_token().await.is_ok() {
+    // If we loaded persisted credentials from the store, and we have a valid access token
+    // (or successfully refreshed a valid refresh token), we're already authorized and good
+    // to go.
+    if auth_manager.initialize_from_store().await? && auth_manager.get_access_token().await.is_ok()
+    {
         if let (Some(client_id), Some(client_secret)) = (client_id, client_secret) {
             auth_manager.configure_client(
                 OAuthClientConfig::new(client_id, redirect_uri.clone())
```

vorporeal (approved):
ship it

PR #10996, opened 2026-05-15
Enter the agent view for third-party harnesses started outside of cloud mode.

liliwilson:
## Description

This PR adds support for entering the agent view more consistently when viewing a session with a third-party harness cloud agent:
- When we get harness data from the server, enter the agent view
- When we detect a third-party CLI harness command running, enter the agent view

We also switch away from using `HarnessSelected` to try to enter the agent view to using a new event to track this. `HarnessSelected` was an overloaded event: it was called in `set_harness`, which is used to update the UI when a user changes the dropdown OR when we receive harness data for a task from an ambient task fetch on joining a new cloud session. This doesn't really make sense for two reasons:
- We shouldn't be trying to enter an agent view when toggling a harness in the cloud mode input
- `set_harness` early returns when the harness is unchanged, which means that `HarnessSelected` doesn't _consistently_ fire on session join. The harness setting doesn't just default/reset to Oz with every run, it retains prior selections, so it was not uncommon that you'd find a 3p harness run coming in and hitting the early return for `set_harness`. 

We fix this by handling entering the agent view using a new event, `ViewerHarnessUpdated`, which is triggered when we pull in task data on a session join. 

Loom: https://www.loom.com/share/bfa261c9392947b09442561792175d5a

## Testing

Tested locally, joining a session before the agent has begun running setup commands, joining while it is in progress, and joining after it's done, but idling.

## Agent Mode

### DIFF
--- a/app/src/terminal/shared_session/shared_handlers.rs
+++ b/app/src/terminal/shared_session/shared_handlers.rs
@@ -426,6 +426,10 @@ pub(crate) fn apply_cli_agent_state_update(
                     }
                 });
             }
+
+            view.update(ctx, |view, ctx| {
+                view.sync_agent_view_for_shared_third_party_viewer(ctx);
+            });
         }
         CLIAgentSessionState::Inactive => {
             // Session cleanup is handled by BlockCompleted events on the
--- a/app/src/terminal/view/ambient_agent/model.rs
+++ b/app/src/terminal/view/ambient_agent/model.rs
@@ -852,6 +852,7 @@ impl AmbientAgentViewModel {
 
                     me.set_environment_id(environment_id, ctx);
                     me.set_harness(harness, ctx);
+                    ctx.emit(AmbientAgentViewModelEvent::ViewerHarnessUpdated);
                 }
                 Err(err) => {
                     log::warn!("Failed to fetch ambient agent task for shared session: {err}");
@@ -1555,6 +1556,8 @@ pub enum AmbientAgentViewModelEvent {
     Cancelled,
     /// The selected execution harness (Oz / Claude Code) changed.
     HarnessSelected,
+    /// A shared-session viewer resolved the run harness from the server task.
+    ViewerHarnessUpdated,
     /// The selected worker host changed via the HostSelector.
     HostSelected,
     /// The selected third-party harness model id changed (e.g. user picked `"opus"` for Claude).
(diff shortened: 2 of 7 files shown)


app/src/terminal/shared_session/shared_handlers.rs
```diff
@@ -426,6 +426,10 @@ pub(crate) fn apply_cli_agent_state_update(
                     }
                 });
             }
+
+            view.update(ctx, |view, ctx| {
```
harryalbert, line 430: OOC, why do we need this both here and when handling `OrderedTerminalEventType::CommandExecutionStarted`? Is it just defensive/for safety's sake, or are these covering different entry cases?

app/src/terminal/view/ambient_agent/model.rs
```diff
@@ -1556,6 +1556,8 @@ pub enum AmbientAgentViewModelEvent {
     Cancelled,
     /// The selected execution harness (Oz / Claude Code) changed.
     HarnessSelected,
+    /// A shared-session viewer resolved the run harness from the server task.
+    ViewerHarnessUpdated,
```
harryalbert: nit: I feel like this event name makes it seem like the viewer selected a harness, not that the viewer's harness was resolved

harryalbert (approved):
nice! Sorry my fix from a while back didn't fully squash this issue

app/src/terminal/shared_session/shared_handlers.rs
liliwilson: Different entry cases. `CommandExecutionStarted` covers the block-start path (harness command block now exists). This one covers viewer join/reconnect when the CLI session is already active but no new block-start has fired yet. Added a comment clarifying.

liliwilson: Yeah, defensive. It may just be more confusing that way though... going to remove from `OrderedTerminalEventType::CommandExecutionStarted` because the other two cases _should_ be sufficient, and are both more directly tied to a CLIAgent (getting harness data from the server, or getting a CLIAgent session synced from the sharer).

app/src/terminal/view/ambient_agent/model.rs
liliwilson: Good call, fixed!

PR #10998, opened 2026-05-15
CLI commands for API key management

bnavetta:
## Description

This adds a set of CLI commands for managing API keys:
* `oz api-key list`: list all keys, with support for JSON output and `jq` filtering
* `oz api-key create`: create a new API key, either personal or for an agent. Choosing an expiry option is required
* `oz api-key expire`: immediately delete/expire an API key, prompting for confirmation first

As the number of uses for API keys grows, I figured this would be useful as a quick way to manage them, especially for creating agent-specific keys.

Plan: https://staging.warp.dev/drive/notebook/Add-Oz-CLI-API-key-management-commands-NmCZta9GAiXjQEcQJJRERl

## Testing

- [x] Added unit tests for sorting and for more complicated parsing logic

### Screenshots / Videos

https://www.loom.com/share/d5d5000e975a48699ade8290079f5f0f (used staging, so restricted to Warp team)

## Agent Mode

### DIFF
--- a/crates/warp_cli/src/api_key.rs
+++ b/crates/warp_cli/src/api_key.rs
@@ -1,0 +1,109 @@
+use crate::date_time::parse_rfc3339;
+use chrono::{DateTime, Utc};
+use clap::{Args, Subcommand, ValueEnum};
+
+use crate::json_filter::JsonOutput;
+
+/// API key-related subcommands.
+#[derive(Debug, Clone, Subcommand)]
+pub enum ApiKeyCommand {
+    /// List active API keys.
+    List(ListApiKeysArgs),
+    /// Create a new API key.
+    Create(CreateApiKeyArgs),
+    /// Immediately expire an API key.
+    #[command(alias = "delete")]
+    Expire(ExpireApiKeyArgs),
+}
+
+#[derive(Debug, Clone, Args)]
+pub struct ListApiKeysArgs {
+    /// Sort field.
+    #[arg(long = "sort-by", value_enum, value_name = "FIELD")]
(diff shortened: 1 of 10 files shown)


crates/warp_cli/src/lib.rs
```diff
@@ -243,6 +245,15 @@ impl Args {
                     }
                 }
 
+                if !FeatureFlag::APIKeyManagement.is_enabled() {
+                    let args: Vec<String> = env::args().collect();
+                    if args.len() > 1 && args[1] == "api-key" {
```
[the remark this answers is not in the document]
bnavetta: The API key management feature has been flagged on for a while, so this isn't really an issue. We should clean it up, but that's out of scope for this PR

crates/warp_cli/src/api_key.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    /// Create an API key with no expiration.
+    #[arg(long = "no-expiration")]
+    pub no_expiration: bool,
+}
+
+#[derive(Debug, Clone, Args)]
+pub struct ExpireApiKeyArgs {
+    /// UID of the API key to expire.
+    pub key_uid: String,
```
captainsafia, line 72: Nit: I feel like I would also like to be able to remove keys by name.

captainsafia (approved):

crates/warp_cli/src/api_key.rs
bnavetta: Good idea, added! https://www.loom.com/share/ccc2ba285d9c4f10a76253296f5fb424

PR #10975, opened 2026-05-15
fix(lsp): size-rotate per-server LSP logs (#10877)

david-engelmann:
Wires the LSP namespace into the size-based rotation primitives shipped for MCP in #10874 (#7723). Same cap as MCP: **10 MiB per file × 5 rotated copies = ~60 MiB per LSP server per workspace**.

Fixes #10877. Depends on #10874 landing first — `LogManager::register_with_rotation` is introduced there.

## What changes

- `crates/lsp/src/model.rs:272` — replaces the unbounded `manager.register("lsp", ...)` with `manager.register_with_rotation("lsp", ..., logs::lsp_log_rotation_config())`. Identical pattern to the MCP wire-up at `app/src/ai/mcp/templatable_manager/native.rs:810`.
- `crates/lsp/src/logs.rs` (new) — `lsp_log_rotation_config()` mirroring `app/src/ai/mcp/logs.rs::mcp_log_rotation_config()`.
- `crates/lsp/src/lib.rs` — registers the new `mod logs`.

Net diff: 49 lines added, 1 removed.

## Why mirror MCP's policy

`simple_logger`'s rotation primitives are namespace-agnostic — the cap constants are the only knobs. Choosing the same 10 MiB × 5 cap as MCP keeps both namespaces well below the multi-GB unbounded growth observed for verbose servers (rust-analyzer in #10877, MCP servers in #7723) while preserving a useful multi-day debugging window. If a future change wants per-namespace caps the constants are already isolated in `crates/lsp/src/logs.rs`.

## Tests

Two unit tests pin the cap constants and document the intentional sync with MCP:

```
$ cargo test -p lsp --lib logs::
test logs::tests::lsp_log_rotation_config_uses_expected_caps ... ok
test logs::tests::lsp_log_rotation_caps_match_mcp_namespace ... ok
```

Full lsp suite still green (18/18), rustfmt + clippy -D warnings clean.

## Why no integration / live-rotation evidence

The rotation logic itself lives in `simple_logger` and is fully covered by the existing `manager_tests.rs` cases shipped in #10874. This PR is strictly the **wire-up**: changing one `register` call site and adding the config helper. The unit tests pin that the helper produces the policy the wire-up depends on; runtime correctness of rotation is already proven for the MCP namespace and is identical here.

[...]

PR #10874, opened 2026-05-14
fix(mcp): size-rotate MCP server log files to cap disk usage (#7723)

david-engelmann:
## Description

Warp captures every MCP server's stderr/stdout to a per-server log file under the `mcp/` namespace. Once a logger was registered, the file was opened with `truncate(true)` and then grew without bound for the lifetime of the app session — a verbose server like iMCP could reach **20 GB in five days** (reported in #7723).

This PR adds size-based rotation to the `simple_logger` crate that backs MCP log capture, and wires the MCP path to use it.

### `crates/simple_logger`

- **`RotationConfig { max_file_size_bytes, max_rotation }`.** `RotationConfig::new` returns `Option<Self>`, mapping any zero to `None` so callers reading from config can statically express "rotation disabled".
- **`SimpleLogger::new`** now takes `rotation: Option<RotationConfig>`.
  - `None` → behavior is unchanged (one file per logger lifetime, truncate-on-create).
  - `Some` → the writer tracks bytes written and, after any write that brings the counter to or above the threshold, drops the file handle, shifts `.N` files up by one (discarding `.{max_rotation}` first), renames the active file to `.1`, and reopens a fresh active file. Log lines are never split across files; the active file may briefly overshoot the threshold by one line.
- **`LogManager::register_with_rotation`** is a sibling of `register` that threads the config through. The existing `register` becomes a thin wrapper that passes `None`, so all current callers other than MCP are unaffected.

### MCP wiring

- `app/src/ai/mcp/logs.rs` exposes `mcp_log_rotation_config()` with a policy of **10 MiB per file × 5 rotations** = 60 MiB cap per server (1 active + 5 rotated). Generous enough to keep a useful debugging window; small enough that even a chatty server can't fill a multi-day session.
- `templatable_manager/native.rs` uses `register_with_rotation` with that config when registering each MCP server's writer.

### Notable design choices

- **Suffix path math operates on the raw `OsString`**, not via `Path::set_extension` — so compound names like `server.stderr.log` get rotated to `server.stderr.log.1` rather than `server.stderr.1` (which is what `set_extension` would produce).
- **The active file handle is dropped before the rename.** Linux would tolerate renaming an open file, but Windows would not — and the subsequent reopen gets a fresh inode either way.
[...]

PR #11000, opened 2026-05-15
feat(logging): in-session size rotation for warp.log (#10879)

### LINKED ISSUE #10879: Size-based rotation for warp.log within a session (currently startup-only)
### Pre-submit Checks

`warp_logging::rotate_log_files()` rotates `warp.log` and its telemetry siblings once at app startup: the previous session's log becomes `.old.0`, older ones shift up to `.old.N`, and the oldest gets discarded. That model works well for users who restart Warp frequently — each session gets its own log file.

For long-running sessions (developers who keep Warp open for days or weeks), the active `warp.log` grows unboundedly during a single session, with no checkpoints in between. The startup rotation never fires until the next restart.

Proposal: add an in-session size-based rotation mode to `warp_logging`, complementing the existing startup rotation. When the active `warp.log` exceeds a configured size threshold mid-session, it's rotated to an in-session sibling (e.g. `warp.log.in_session.N`) and a fresh active file is opened. The startup `.old.N` semantics that log-bundle uploads and other UX depend on stay untouched.

Two related reports for *other* log files exist:

- #7723 / #10874 — MCP server logs not rotated (`simple_logger`)
- #10877 — LSP server logs not rotated (`simple_logger`)

Both of those operate on a different rotation engine than `warp.log`, but the underlying user complaint is the same: Warp's diagnostic data grows without bound during long sessions. Capping `warp.log` at e.g. 50 MiB × 5 rotated copies would bound it at ~300 MiB even across multi-day sessions, while preserving the per-session file boundary at startup that today's UX (e.g. log-bundle uploads) relies on.

### Additional context

This complements (does not replace) the MCP/LSP rotation work. Different file, different rotation engine, but the same principle: cap on-disk footprint during long sessions while preserving the current per-session semantics.

Sketch of the implementation surface:

- Extend `LogConfig` with optional `max_file_size_bytes` (or equivalent threshold)
- Plumb that to the log writer (e.g. tracing-appender's rolling appender, or a manual periodic size check)
- Rotation pattern: e.g. `warp.log.in_session.N` to distinguish from the existing startup `.old.N` files

### Operating system (OS)

macOS

3

Ignore

david-engelmann:
Fixes #10879. Independent of #10874 / #10975 — `warp_logging` is a different engine (env_logger + a pipe target), so this PR can land on its own.

## What

Adds an orthogonal **in-session** size bound to `warp.log`, complementing the existing **per-startup** rotation. When the active `warp.log` exceeds `LogConfig::max_file_size_bytes` mid-session it is rotated to `warp.log.in_session.0`, existing `.in_session.N` slots shift up, and the oldest beyond `max_rotation` is discarded. The startup `.old.N` slots — which log-bundle uploads and crash recovery depend on — are untouched.

Sketch matches the one in #10879's description: extend `LogConfig`, plumb to the writer, use `.in_session.N` to distinguish from `.old.N`.

## API

`max_file_size_bytes` defaults to `None`, preserving today's unbounded-within-session behavior for every existing caller. Adopting the bound is opt-in:

```rust
warp_logging::init(LogConfig {
    max_file_size_bytes: Some(50 * 1024 * 1024),
    ..Default::default()
})?;
```

Setting a concrete cap on the GUI startup path (e.g. 50 MiB × the existing `MAX_FILES_IN_GUI_ROTATION` = 5 → ~300 MiB bound per multi-day session) is intentionally **not part of this PR** — it's a follow-up. This change only ships the mechanism so the policy decision can be made separately, with feature-flag staging if desired.

## How

- New `crates/warp_logging/src/rotation.rs`: a `Write` adapter (`RotatingFileWriter`) that owns a `File`, tracks `bytes_written`, and rotates pre-write when the next chunk would cross `max_bytes`. Rotation order: delete oldest slot → shift remaining slots → rename active into slot 0 → open a fresh active. Matches the overflow semantics of the existing startup `rotate_files`.
- `crates/warp_logging/src/lib.rs`: `LogConfig.max_file_size_bytes: Option<u64>`; `LogConfig` now derives `Default`.
- `crates/warp_logging/src/native.rs`: `init_internal` threads the cap through; the file pipe target is wrapped via `rotation::wrap_for_rotation` for the main process. Crash recovery's `.recovery` log isn't wrapped — it has its own short-lived lifecycle.

Net diff: **+326 / -6 lines**, the bulk of which is `rotation.rs` and its tests.

## Tests

```
$ cargo test -p warp_logging --lib rotation::

running 6 tests
test rotation::tests::zero_max_rotation_truncates_in_place_without_sidecar ... ok
test rotation::tests::writes_below_threshold_do_not_rotate ... ok
test rotation::tests::crossing_threshold_rotates_to_in_session_zero ... ok
test rotation::tests::zero_max_bytes_disables_rotation_entirely ... ok
test rotation::tests::overflow_drops_the_oldest_slot ... ok
test rotation::tests::repeated_rotations_shift_slots_up ... ok

test result: ok. 6 passed; 0 failed
```

The tests exercise: sub-threshold writes don't rotate; crossing the threshold rotates to slot 0; repeated rotations shift older slots up; overflow drops the oldest; `max_rotation = 0` truncates in place without producing sidecars; `max_file_size_bytes = 0` disables rotation entirely.

`cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean.

## Companion PRs

This is the third in a coherent rotation trilogy across Warp's three log namespaces:

- **MCP server logs** — #10874 (depends on `simple_logger`'s rotation primitives)
- **LSP server logs** — #10975 (depends on #10874)
- **`warp.log` itself** — this PR (independent engine, no dependency)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14872,6 +14872,7 @@ dependencies = [
  "log-panics",
  "sentry",
  "sentry-log",
+ "tempfile",
  "warp_core",
  "warp_web_event_bus",
  "wasm-bindgen",
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -654,6 +654,7 @@ pub fn run() -> Result<()> {
                 warp_logging::init(warp_logging::LogConfig {
                     is_cli: true,
                     log_destination: launch_mode.log_destination(),
+                    ..Default::default()
                 })?;
                 return crate::remote_server::run_proxy(args.identity_key.clone());
             }
@@ -794,10 +795,18 @@ fn run_internal(mut launch_mode: LaunchMode) -> Result<()> {
             if crash_recovery::is_crash_recovery_process(launch_mode.args().as_ref()) {
                 warp_logging::init_for_crash_recovery_process()?;
             } else {
-                warp_logging::init(warp_logging::LogConfig { is_cli, log_destination })?;
+                warp_logging::init(warp_logging::LogConfig {
+                    is_cli,
+                    log_destination,
+                    ..Default::default()
+                })?;
             }
         } else {
-            warp_logging::init(warp_logging::LogConfig { is_cli, log_destination })?;
+            warp_logging::init(warp_logging::LogConfig {
+                    is_cli,
+                    log_destination,
+                    ..Default::default()
+                })?;
         }
     }
 
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -803,10 +803,10 @@ fn run_internal(mut launch_mode: LaunchMode) -> Result<()> {
             }
         } else {
             warp_logging::init(warp_logging::LogConfig {
-                    is_cli,
-                    log_destination,
-                    ..Default::default()
-                })?;
+                is_cli,
+                log_destination,
+                ..Default::default()
+            })?;
         }
     }
 
--- a/crates/onboarding/examples/callout_flow.rs
+++ b/crates/onboarding/examples/callout_flow.rs
@@ -175,6 +175,7 @@ fn main() -> Result<()> {
     warp_logging::init(warp_logging::LogConfig {
         is_cli: false,
         log_destination: None,
+        ..Default::default()
     })?;
 
     let app_builder =
--- a/crates/onboarding/src/bin/main.rs
+++ b/crates/onboarding/src/bin/main.rs
@@ -44,6 +44,7 @@ fn main() -> Result<()> {
     warp_logging::init(warp_logging::LogConfig {
         is_cli: false,
         log_destination: None,
+        ..Default::default()
     })?;
 
     let app_builder =
--- a/crates/warp_logging/Cargo.toml
+++ b/crates/warp_logging/Cargo.toml
@@ -29,6 +29,9 @@ wasm-bindgen.workspace = true
 web-sys.workspace = true
 warp_web_event_bus.workspace = true
 
+[target.'cfg(not(target_family = "wasm"))'.dev-dependencies]
+tempfile.workspace = true
+
 [features]
 crash_reporting = ["dep:sentry", "dep:sentry-log", "warp_core/crash_reporting"]
 agent_mode_evals = []
--- a/crates/warp_logging/src/lib.rs
+++ b/crates/warp_logging/src/lib.rs
@@ -8,18 +8,28 @@ pub enum LogDestination {
 }
 
 /// Configuration for initializing the logger.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct LogConfig {
     /// Whether the caller is the CLI. When true, logs are written to a separate subdirectory
     /// with a higher rotation limit so that CLI invocations don't evict GUI application logs.
     pub is_cli: bool,
     /// The destination for log output. If `None`, the destination is inferred from the environment.
     pub log_destination: Option<LogDestination>,
+    /// Optional in-session size threshold for `warp.log`. When `Some(n)` and the active
+    /// file accumulates more than `n` bytes during a single execution, it is rotated to
+    /// `warp.log.in_session.0` and a fresh active file is opened. Older `.in_session.N`
+    /// files shift up and the oldest is discarded, matching the per-startup
+    /// `rotate_log_files` behavior. `None` preserves the existing unbounded-within-session
+    /// growth (warpdotdev/warp#10879).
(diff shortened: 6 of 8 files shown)


david-engelmann:
Addressed in `adb00b1a`: extended `current_and_rotated_log_paths` to also collect `<name>.log.in_session.N` files, ordered newest-first (active → `in_session.0..N` → `old.0..M`). Pulled the directory scan into a pure `collect_log_paths_in` helper and added five unit tests covering the new ordering, channel-name gating, malformed suffixes, and the empty-directory error path.

/oz-review

david-engelmann:
Addressed the nit in `0cdd8bf0`: guarded the rotation predicate on `bytes_written > 0`, so an oversized first write no longer promotes an empty active file into `.in_session.0`. The payload stays in the active file and gets preserved on the next real rotation. Added `oversized_first_write_does_not_promote_empty_file_to_in_session_zero` as a regression test; all 12 `warp_logging` tests pass.

david-engelmann:
Addressed rustfmt nit in `0e6abec1`. Indentation on the non-crash-recovery `LogConfig` literal now matches its sibling at [app/src/lib.rs:798](app/src/lib.rs#L798). Also picked up a couple of `cargo fmt` cleanups in the new test file. `cargo fmt -- --check` is clean.

/oz-review

kevinchevalier (commented):
- Do we need to rotate the `in_session` logs on app start to be called `old.in_session`?
- I think the comment about including `.in_session` files in the zip file of aggregated logs makes sense.

david-engelmann:
@kevinchevalier — addressed in `d7fdd268`.

**Q1 (the `.in_session.*` lifetime across sessions)** — you're right that this was a gap. Until now `rotate_files` cycled the `.old.N` startup slots but never touched the previous session's `.in_session.N` files, so today's session opened with yesterday's mid-rotation chunks still occupying its `.in_session.*` namespace. The `max_rotation` cap was per-namespace, not per-session — the new session only ever got fresh slots after displacing every previous-session chunk one rotation at a time.

The fix nests each old session's chunks under its `.old.K` slot:

```text
on startup, in rotate_files:
  delete .old.{max-1}              # existing
  delete .old.{max-1}.in_session.M # NEW: drop the nested chunks of the slot we just discarded
  for K in (max-2 .. 0):
    rename .old.{K} → .old.{K+1}                  # existing
    rename .old.{K}.in_session.M → .old.{K+1}.in_session.M for all M   # NEW
  rename .in_session.M → .old.0.in_session.M for all M  # NEW: clear .in_session.* for new session
  rename .old.temp → .old.0        # existing
```

After this, each previous session's final-state log at `.old.K` is co-located with that session's mid-rotation chunks at `.old.K.in_session.0..N`, and the bare `.in_session.*` namespace always belongs to the currently-running session.

**Q2 (`.in_session.*` files in the log bundle)** — already addressed in `adb00b1a` (the earlier round of changes from Oz's first review). `collect_log_paths_in` now also surfaces the new `.old.K.in_session.M` files, interleaved with their parent `.old.K` so a session's final state is immediately followed by that session's mid-rotation chunks before the next-older session begins:

```
warp.log
warp.log.in_session.0        ← current session's mid-rotations
warp.log.in_session.1
warp.log.old.0               ← previous session's final state
warp.log.old.0.in_session.0  ← previous session's mid-rotations
warp.log.old.0.in_session.1
warp.log.old.1               ← session before that
warp.log.old.1.in_session.0
…
```

Orphan nested chunks (parent `.old.K` missing from disk) still surface rather than silently dropping.

**Tests** — 8 new unit tests in `crates/warp_logging/src/native.rs` cover:

- `migrate_previous_session_renames_in_session_to_old_0_in_session`
- `migrate_previous_session_is_a_noop_when_no_in_session_chunks_exist`
- `migrate_previous_session_ignores_malformed_in_session_filenames`
- `shift_nested_chunks_renames_old_n_in_session_to_old_n_plus_1`
- `shift_nested_chunks_only_touches_the_requested_slot`
- `remove_nested_chunks_deletes_every_chunk_of_the_target_slot`
- `interleaves_old_logs_with_their_nested_in_session_chunks`
- `nested_chunks_surface_even_when_parent_old_slot_is_missing`

```text
$ cargo test -p warp_logging --lib
20 passed; 0 failed
$ cargo clippy -p warp_logging --tests -- -D warnings
clean
$ cargo fmt -p warp_logging -- --check
clean
```

kevinchevalier (approved):
This looks good, thanks!

PR #11009, opened 2026-05-15
Fix remote server socket path exceeding sun_path limit

kevinyang372:
## Description

PR #10782 ("Version aware daemon socket") changed daemon socket filenames from `server.sock` (11 chars) to `server-v0.2026.05.13.09.15.stable_01.sock` (41 chars), adding ~30 bytes to the Unix domain socket path. This pushes users over the `sun_path` limit (107 bytes on Linux, 103 on macOS), causing `UnixListener::bind` to fail silently in the daemon. The proxy then times out after 10s, exits with code 1, and the client sees `ResponseChannelClosed`.

This accounts for the ~20% drop in remote server initialization success rate (90% → 70%). Sentry issue: WARP-CLIENT-BETA-STABLE-7M9M (11 occurrences across 9 users, all anonymous with 36-char UUID identity keys).

**Fix (two changes):**

1. **Hash the version string** in socket/PID filenames to an 8-hex-char suffix (`server-a1b2c3d4.sock`, 20 chars) instead of the full version string (41 chars).

2. **Hash the identity key** directory name to 8 hex chars instead of using the raw identity key (up to 36 chars for anonymous UUIDs). This is needed because longer channel base dirs like `.warp-preview` (+8 chars vs `.warp`) would otherwise narrow the headroom too much on macOS.

Both changes use `std::hash::DefaultHasher` for deterministic, fixed-length output. The existing `cleanup_old_versions()` logic automatically cleans up old long-form files.

Also adds an explicit `sun_path` length guard in the proxy that fails fast with a clear error message instead of silently timing out.

**Worst-case path after fix:** `/home/{32-char-user}/.warp-preview/remote-server/{8-char-hash}/server-{8-char-hash}.sock` = **97 bytes**, well under both limits (103 macOS, 107 Linux).

Fixes WARP-CLIENT-BETA-STABLE-7M9M

## Linked Issue

## Testing
- 10 unit tests: version hash determinism, identity key hash (length, determinism, uniqueness), socket/PID name length bounds, and worst-case path length regression test against the preview channel (longest base dir)
- Validated via Docker containers on linux/amd64, linux/arm64, and macOS/aarch64 with the actual 9 Sentry-reported identity keys
- Cross-referenced Sentry events to confirm all affected users are anonymous (36-char UUID identity keys)
- `cargo fmt` and `cargo clippy` pass

## Agent Mode

### DIFF
--- a/app/src/remote_server/unix/proxy.rs
+++ b/app/src/remote_server/unix/proxy.rs
@@ -91,6 +91,17 @@ pub(super) fn ensure_private_daemon_dir(path: &std::path::Path) -> anyhow::Resul
     Ok(())
 }
 
+/// Maximum usable `sun_path` length for Unix domain sockets.
+///
+/// The `sockaddr_un.sun_path` field is 108 bytes on Linux and 104 on
+/// macOS, but one byte is reserved for the null terminator.
+#[cfg(target_os = "linux")]
+const SUN_PATH_MAX: usize = 107;
+#[cfg(target_os = "macos")]
+const SUN_PATH_MAX: usize = 103;
+#[cfg(not(any(target_os = "linux", target_os = "macos")))]
+const SUN_PATH_MAX: usize = 103; // conservative default
+
 /// Entry point for `remote-server-proxy`.
 ///
 /// Ensures the daemon is running, then bridges stdin/stdout to the daemon's
@@ -98,6 +109,19 @@ pub(super) fn ensure_private_daemon_dir(path: &std::path::Path) -> anyhow::Resul
 pub fn run(identity_key: &str) -> anyhow::Result<()> {
     let socket_path = socket_path(identity_key);
     let pid_path = pid_path(identity_key);
+
+    // Guard against socket paths that exceed the platform's sun_path
+    // limit. This would cause UnixListener::bind to fail silently in
+    // the daemon, leaving the proxy to time out waiting for a socket
+    // that never appears.
+    let path_len = socket_path.as_os_str().len();
+    if path_len > SUN_PATH_MAX {
+        anyhow::bail!(
+            "daemon socket path is {path_len} bytes, which exceeds the \
+             platform sun_path limit of {SUN_PATH_MAX} bytes: {}",
+            socket_path.display()
+        );
+    }
 
     // Ensure the parent directory exists.
     if let Some(parent) = socket_path.parent() {
--- a/app/src/remote_server/unix/proxy.rs
+++ b/app/src/remote_server/unix/proxy.rs
@@ -93,14 +93,10 @@ pub(super) fn ensure_private_daemon_dir(path: &std::path::Path) -> anyhow::Resul
 
 /// Maximum usable `sun_path` length for Unix domain sockets.
 ///
-/// The `sockaddr_un.sun_path` field is 108 bytes on Linux and 104 on
-/// macOS, but one byte is reserved for the null terminator.
-#[cfg(target_os = "linux")]
-const SUN_PATH_MAX: usize = 107;
-#[cfg(target_os = "macos")]
+/// macOS has the strictest limit (104 bytes including null terminator,
+/// 103 usable). We use 103 on all platforms so a single binary works
+/// everywhere without per-target branching.
 const SUN_PATH_MAX: usize = 103;
-#[cfg(not(any(target_os = "linux", target_os = "macos")))]
-const SUN_PATH_MAX: usize = 103; // conservative default
 
 /// Entry point for `remote-server-proxy`.
 ///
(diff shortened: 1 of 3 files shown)


kevinyang372:
* **#11009** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11009?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11009?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/remote_server/src/setup.rs
```diff
@@ -394,26 +394,48 @@ pub fn remote_server_daemon_data_dir(identity_key: &str) -> String {
...
+/// The hash keeps the filename at a fixed 24 chars, well under the
+/// `sun_path` limit. The full version string used to be embedded
+/// directly (e.g. `server-v0.2026.05.13.09.15.stable_01.sock`, 43
+/// chars), which caused `UnixListener::bind` failures for users with
+/// longer paths.
 pub fn daemon_socket_name() -> String {
-    match ChannelState::app_version() {
-        Some(version) => format!("server-{version}.sock"),
+    match version_hash() {
+        Some(hash) => format!("server-{hash}.sock"),
```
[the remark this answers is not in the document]
kevinyang372: Added identity key hashing as well

crates/remote_server/src/setup.rs
```diff
@@ -356,26 +356,23 @@ pub fn remote_server_dir() -> String {
...
-            b'a'..=b'z' | b'A'..=b'Z' | b'0'..=b'9' | b'-' | b'_' => {
-                encoded.push(byte as char);
-            }
-            _ => encoded.push_str(&format!("%{byte:02X}")),
-        }
-    }
-    encoded
+    let mut hasher = std::collections::hash_map::DefaultHasher::new();
+    identity_key.hash(&mut hasher);
+    format!("{:016x}", hasher.finish())[..8].to_string()
```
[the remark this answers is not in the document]
kevinyang372: We don't need migration here. SQLite DB is unused right now

app/src/remote_server/unix/proxy.rs
```diff
@@ -91,6 +91,17 @@ pub(super) fn ensure_private_daemon_dir(path: &std::path::Path) -> anyhow::Resul
...
 }
 
+/// Maximum usable `sun_path` length for Unix domain sockets.
+///
+/// The `sockaddr_un.sun_path` field is 108 bytes on Linux and 104 on
+/// macOS, but one byte is reserved for the null terminator.
+#[cfg(target_os = "linux")]
+const SUN_PATH_MAX: usize = 107;
+#[cfg(target_os = "macos")]
+const SUN_PATH_MAX: usize = 103;
```
alokedesai, line 99: Let's just take 103 as the min? the per target branching is unnecessary

app/src/remote_server/unix/proxy.rs
```diff
@@ -99,6 +110,19 @@ pub fn run(identity_key: &str) -> anyhow::Result<()> {
...
+    // Guard against socket paths that exceed the platform's sun_path
+    // limit. This would cause UnixListener::bind to fail silently in
+    // the daemon, leaving the proxy to time out waiting for a socket
+    // that never appears.
+    let path_len = socket_path.as_os_str().len();
+    if path_len > SUN_PATH_MAX {
+        anyhow::bail!(
+            "daemon socket path is {path_len} bytes, which exceeds the \
+             platform sun_path limit of {SUN_PATH_MAX} bytes: {}",
+            socket_path.display()
```
alokedesai, line 121: In the case we hit this, what do we or the user do?

crates/remote_server/src/setup.rs
```diff
@@ -394,26 +391,48 @@ pub fn remote_server_daemon_data_dir(identity_key: &str) -> String {
...
+/// a release tag is baked in.
+///
+/// - With `GIT_RELEASE_TAG`:    `server-{hash8}.sock`  (e.g. `server-a1b2c3d4.sock`)
 /// - Without (plain cargo run): `server.sock`
+///
+/// The hash keeps the filename at a fixed 24 chars, well under the
+/// `sun_path` limit. The full version string used to be embedded
+/// directly (e.g. `server-v0.2026.05.13.09.15.stable_01.sock`, 43
+/// chars), which caused `UnixListener::bind` failures for users with
+/// longer paths.
```
alokedesai: not blocking at all but comments about the old impl aren't worth keeping in the code

crates/remote_server/src/setup.rs
```diff
@@ -356,26 +356,23 @@ pub fn remote_server_dir() -> String {
...
 ///
-/// The identity key is not secret, but it can contain bytes that are unsafe or
-/// ambiguous in paths. Keep ASCII alphanumeric characters plus `-` and `_`;
-/// percent-encode all other UTF-8 bytes.
+/// Hashes the identity key to an 8-hex-char string to keep the daemon
+/// socket path well within the `sun_path` limit across all channels.
+/// The previous implementation percent-encoded the raw key (up to 36+
+/// chars for a UUID), which, combined with longer channel base dirs
+/// like `.warp-preview`, could push the socket path over the limit.
 pub fn remote_server_identity_dir_name(identity_key: &str) -> String {
```
alokedesai, line 364: This is also where we store the sqlite directory, no? Can we decouple where we store the state from the daemom sock file to minimize the impact if we were to have a hash collision?

alokedesai (approved):

crates/remote_server/src/setup.rs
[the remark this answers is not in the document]
alokedesai: cool, as long as we use the actual right identity key for sqlite directory i'm good

app/src/remote_server/unix/proxy.rs
kevinyang372: This is realistically never going to happen after this fix. It is just a safe guard so we could return clearer errors rather than hitting generic timeout

crates/remote_server/src/setup.rs
kevinyang372: Yeah we could decouple the data dir. Done

PR #11044, opened 2026-05-15
disable dogfood custom endpoints + enable only for warp plan

dagmfactory:
## Description
Disabled custom inference for everyone on dogfood and enabled it for users on warp plan for testing
## Agent Mode

<!--
CHANGELOG-NONE

### DIFF
--- a/app/src/workspaces/user_workspaces.rs
+++ b/app/src/workspaces/user_workspaces.rs
@@ -512,9 +512,15 @@ impl UserWorkspaces {
 
         self.current_workspace()
             .map(|workspace| {
-                workspace.billing_metadata.customer_type != CustomerType::Enterprise
-                    || FeatureFlag::CustomInferenceEndpointsEnterprise.is_enabled()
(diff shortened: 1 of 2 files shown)


app/src/workspaces/user_workspaces.rs
```diff
@@ -512,9 +512,15 @@ impl UserWorkspaces {
 
         self.current_workspace()
             .map(|workspace| {
-                workspace.billing_metadata.customer_type != CustomerType::Enterprise
-                    || FeatureFlag::CustomInferenceEndpointsEnterprise.is_enabled()
-                    || ChannelState::channel().is_dogfood()
+                let is_enterprise_customer =
+                    workspace.billing_metadata.customer_type == CustomerType::Enterprise;
+                let is_warp_plan_billing_workspace = workspace.billing_metadata.is_warp_plan();
```
danielpeng2, line 517: i think `is_warp_plan` only works for our stable team. is there a way we can get an exception for our staging team as well?

PR #11049, opened 2026-05-15
Add sleep auto handoff to cloud

dplakon:
## Description
Adds a macOS-only **Auto handoff on sleep** setting under Warp Agent → Cloud Handoff.

When enabled, Warp listens for macOS sleep events and attempts to hand off the last-focused eligible local Warp Agent session to Cloud Mode using the existing local-to-cloud handoff flow. This also adds a `warplocal://action/auto_handoff_to_cloud` trigger for testing and hook-driven invocation; the trigger now respects the same **Auto handoff on sleep** opt-in setting.

The setting defaults to off so existing users do not silently enable automatic cloud handoff or workspace snapshot upload.

https://github.com/user-attachments/assets/38161805-065a-49b4-a4e6-18403de5b0a1

Screenshot 2026-05-15 at 5 06 56 PM

## Linked Issue
No linked issue.

## Testing
- [x] `cargo fmt`
- [x] `cargo nextest run -p warp auto_handoff --no-fail-fast` — 13 passed
- [x] `cargo nextest run -p warp last_focused_terminal_tracks_most_recent_globally conversation_switch_updates_last_focused_terminal_state --no-fail-fast` — 2 passed
- [x] `git --no-pager diff --check origin/master...HEAD`
- [x] Manual smoke test with `./script/run --open_with_launchd` and `warplocal://action/auto_handoff_to_cloud` in the original implementation checkout

### Screenshots / Videos
Not included in this update. The PR now notes the setting is opt-in and the URI trigger respects that opt-in; please attach UI media from a local app session if required for review.

## Agent Mode

CHANGELOG-OZ: Added a macOS setting to automatically hand off active local Warp Agent sessions to Cloud Mode before sleep.

### DIFF
--- a/app/src/settings/ai.rs
+++ b/app/src/settings/ai.rs
@@ -1751,7 +1751,11 @@ impl AISettings {
     }
 
     pub fn is_auto_handoff_on_sleep_enabled(&self, app: &warpui::AppContext) -> bool {
-        self.is_cloud_handoff_enabled(app) && *self.auto_handoff_on_sleep_enabled
+        self.is_cloud_handoff_enabled(app)
+            && self
+                .auto_handoff_on_sleep_enabled
+                .is_supported_on_current_platform()
+            && *self.auto_handoff_on_sleep_enabled
     }
 
     /// Determines whether a quota reset banner should be displayed to the user.
--- a/app/src/settings/ai.rs
+++ b/app/src/settings/ai.rs
@@ -1497,6 +1497,16 @@ define_settings_group!(AISettings, settings: [
         toml_path: "agents.warp_agent.other.should_force_disable_ampersand_handoff",
         description: "Whether to force-disable the & prefix for cloud handoff compose mode.",
     }
+
+    auto_handoff_on_sleep_enabled: AutoHandoffOnSleepEnabled {
+        type: bool,
+        default: false,
+        supported_platforms: SupportedPlatforms::MAC,
+        sync_to_cloud: SyncToCloud::Globally(RespectUserSyncSetting::Yes),
+        private: false,
+        toml_path: "agents.warp_agent.other.auto_handoff_on_sleep_enabled",
+        description: "Whether Warp automatically hands off local agent conversations to cloud when the computer is about to sleep.",
+    }
 ]);
 
 impl AISettings {
@@ -1738,6 +1748,10 @@ impl AISettings {
     ) -> bool {
         self.is_cloud_handoff_enabled_for_terminal_view(terminal_view_id, app)
             && !*self.should_force_disable_ampersand_handoff
+    }
+
+    pub fn is_auto_handoff_on_sleep_enabled(&self, app: &warpui::AppContext) -> bool {
+        self.is_cloud_handoff_enabled(app) && *self.auto_handoff_on_sleep_enabled
     }
 
     /// Determines whether a quota reset banner should be displayed to the user.
--- a/app/src/settings_view/ai_page.rs
+++ b/app/src/settings_view/ai_page.rs
@@ -2677,6 +2677,7 @@ pub enum AISettingsPageAction {
     SetConversationLayout(crate::util::file::external_editor::settings::OpenConversationPreference),
     ToggleCloudHandoff,
     ToggleAmpersandHandoff,
+    ToggleAutoHandoffOnSleep,
     ToggleShowConversationHistory,
     ToggleAutoToggleRichInput,
     ToggleAutoOpenRichInputOnCLIAgentStart,
@@ -3435,6 +3436,14 @@ impl TypedActionView for AISettingsPageView {
                 AISettings::handle(ctx).update(ctx, |settings, ctx| {
                     report_if_error!(settings
                         .should_force_disable_ampersand_handoff
+                        .toggle_and_save_value(ctx));
+                });
+                ctx.notify();
+            }
+            AISettingsPageAction::ToggleAutoHandoffOnSleep => {
+                AISettings::handle(ctx).update(ctx, |settings, ctx| {
+                    report_if_error!(settings
+                        .auto_handoff_on_sleep_enabled
                         .toggle_and_save_value(ctx));
                 });
                 ctx.notify();
@@ -6712,6 +6721,7 @@ impl SettingsWidget for CloudAgentComputerUseWidget {
 #[derive(Default)]
 struct CloudHandoffWidget {
     handoff_toggle: SwitchStateHandle,
+    auto_handoff_on_sleep_toggle: SwitchStateHandle,
     ampersand_toggle: SwitchStateHandle,
 }
 
@@ -6719,7 +6729,7 @@ impl SettingsWidget for CloudHandoffWidget {
     type View = AISettingsPageView;
 
     fn search_terms(&self) -> &str {
-        "cloud handoff ampersand & move to cloud local"
+        "cloud handoff auto sleep ampersand & move to cloud local"
     }
 
     fn should_render(&self, _app: &AppContext) -> bool {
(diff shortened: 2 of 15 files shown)


app/src/workspace/auto_handoff.rs
```diff
@@ -0,0 +1,246 @@
...
+        }
+    }
+
+    fn handle_system_stats_event(
+        &mut self,
+        event: &SystemStatsEvent,
+        ctx: &mut ModelContext<Self>,
+    ) {
+        match event {
+            SystemStatsEvent::CpuWillSleep => {
```
harryalbert, line 135: I'm a little concerned about using this event as the trigger for handoff, just because handoff involves some async processes that could take a while and might not reliably trigger before the computer shuts down.

I'm not sure if there's a great way to fix this reliably, but I'm curious if you ran into it when testing this feature w/ longer conversations that might have taken longer to hand off

app/src/workspace/auto_handoff.rs
```diff
@@ -0,0 +1,246 @@
...
+        };
+
+        if let Some(reason) = skip_reason {
+            log::debug!(
+                "Skipping auto handoff to cloud for conversation {conversation_id:?}: {reason:?}"
+            );
+            return;
+        }
+
+        self.attempted_conversation_ids.insert(conversation_id);
```
harryalbert: we should probably wait to insert the conversation id into this list until the handoff has happened successfully. Might make sense to switch this to a map that tracks handoff state (like in flight, succeeded, failed, etc), listening for handoff lifecycle events (which I think we'd have to add) and keeping track of state.

If this fix doesn't make sense or ends up being too heavy weight, we could also just remove conversations from `attempted_conversation_ids` on-handoff-failure, although this probably wouldn't cover the case where the handoff was stopped because the computer shut down too early (pertaining to the comment above this one)

app/src/workspace/view.rs
```diff
@@ -21074,6 +21198,31 @@ impl TypedActionView for Workspace {
...
+                trigger,
+            } => {
+                #[cfg(all(feature = "local_fs", not(target_family = "wasm")))]
+                {
+                    let launch = Some(PendingCloudLaunch {
+                        prompt: AUTO_CLOUD_HANDOFF_PROMPT.to_owned(),
+                        attachments: HandoffLaunchAttachments::default(),
+                    });
+                    if let Some(source_view) = self.terminal_view(*terminal_view_id, ctx) {
+                        self.start_local_to_cloud_handoff_from_source(
```
harryalbert, line 21357: would probably make sense to pass the conversation_id through to this event directly (instead of re-getting the active conversation from the source view), given that the view code above does some validation and bookkeeping on the active conversation before emitting this event (so we'd want to avoid a case where the conversation that we end up handing off somehow changes)

app/src/workspace/view.rs
```diff
@@ -13650,22 +13738,25 @@ impl Workspace {
         }
```
harryalbert, line 13885: <p class='graphite__hidden'><i>[Re: line +13733]</i></p>
Depending on why we entered this path (manual cancellation versus automatic handoff), we should probably provide a separate/new cancellation reason for automatic handoff
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11049?utm_source=unchanged-line-comment">Graphite</a>.</p>

app/src/settings/ai.rs
```diff
@@ -1507,6 +1507,16 @@ define_settings_group!(AISettings, settings: [
         toml_path: "agents.warp_agent.other.should_force_disable_ampersand_handoff",
         description: "Whether to force-disable the & prefix for cloud handoff compose mode.",
     }
+
+    auto_handoff_on_sleep_enabled: AutoHandoffOnSleepEnabled {
+        type: bool,
+        default: false,
+        supported_platforms: SupportedPlatforms::MAC,
```
harryalbert, line 1504: nice

app/src/settings_view/ai_page.rs
```diff
@@ -6833,6 +6843,39 @@ impl SettingsWidget for CloudHandoffWidget {
...
+                    .check(*ai_settings.auto_handoff_on_sleep_enabled)
+                    .build()
+                    .on_click(move |ctx, _, _| {
+                        ctx.dispatch_typed_action(AISettingsPageAction::ToggleAutoHandoffOnSleep);
+                    })
+                    .finish();
+
+                let auto_handoff_on_sleep_row = build_toggle_element(
+                    render_body_item_label::<AISettingsPageAction>(
+                        "Auto handoff on sleep".to_string(),
```
harryalbert: I wonder if we should make this text a little more detailed. idk if I would understand what this means as a newer user

harryalbert (commented):
Very cool change. Just had a few thoughts/comments.

Also, just to double-check, this doesn't auto-cancel cloud conversations if they're focused, right (from the code it looks like it doesn't but just wanted to make sure)?

app/src/workspace/auto_handoff.rs
dplakon: Yeah, it's a good call out. I'm hoping some of the work on improving speed for handoff would be a potential fix here. I will do some research on a better solution, but it does have a 30s buffer.

app/src/workspace/auto_handoff.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum AutoCloudHandoffAttemptState {
+    InFlight,
+    Succeeded,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct AutoCloudHandoffEligibility {
```
harryalbert, line 35: oh we should also skip handing off for conversations w/ orchestration (so either a parent orchestrator or a child that is orchestrated).

We generally gate handoff on this condition because moving an orchestrated conversation to the cloud has some weird side-effects, so we should do that here too. 

I made a change to the cloud handoff enabled fn to check that the given conversation isn't orchestrated, so you might have to rebase onto master to get that change

app/src/workspace/auto_handoff.rs
```diff
@@ -0,0 +1,264 @@
...
+        if !self.has_server_conversation_token {
+            return Some(AutoCloudHandoffSkipReason::MissingServerConversationToken);
+        }
+        None
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct AutoCloudHandoffTriggerSettings {
+    pub(crate) cloud_handoff_enabled: bool,
```
harryalbert: ooc, why are we holding this state here? I don't see where it's used/why it makes sense to keep track of as opposed to checking the handoff setting directly

app/src/workspace/auto_handoff.rs
```diff
@@ -0,0 +1,264 @@
...
+            SystemStatsEvent::CpuWillSleep => {
+                self.trigger(AutoCloudHandoffTrigger::MacOsSleep, ctx);
+            }
+            SystemStatsEvent::CpuWasAwakened => {}
+        }
+    }
+
+    fn trigger(&mut self, trigger: AutoCloudHandoffTrigger, ctx: &mut ModelContext<Self>) {
+        if !Self::trigger_settings(ctx).is_enabled_for(trigger) {
+            log::debug!("Skipping auto handoff to cloud via {trigger:?}: trigger is disabled");
```
harryalbert: I think we should clear out these log::debug calls.

app/src/workspace/view.rs
```diff
@@ -13587,17 +13652,94 @@ impl Workspace {
...
-            .cloned();
+        let source_conversation = {
+            let history_model = BlocklistAIHistoryModel::as_ref(ctx);
+            match intent.expected_conversation_id() {
+                Some(expected_conversation_id) => {
+                    let Some(active_conversation) =
+                        history_model.active_conversation(terminal_view_id)
+                    else {
+                        if let Some(trigger) = intent.automatic_trigger() {
+                            log::debug!(
```
harryalbert: same comment here about clearing out these debug logs (or shifting to log::info or warn if you think they're valuable)

harryalbert (approved):
Approving to unblock, but had a few more comments that would be good to address before merging

harryalbert:
@dplakon had the agent take a look, and it had one other comment that I also think is worth addressing:
```
app/src/workspace/auto_handoff.rs:198-207 and app/src/workspace/view.rs:13677-13685 — automatic handoff starts while AutoCloudHandoffController is already inside an update(...), but several early failure paths call record_automatic_handoff_failed(), which immediately does another AutoCloudHandoffController::handle(ctx).update(...). WarpUI treats same-model reentry as a circular update, so a race like “focused conversation changed/disappeared between selection and handoff start” can panic the client instead of cleanly bailing out.
```

PR #11061, opened 2026-05-15
Adding guard for async process

evelyn-with-warp:
## Description

Context: https://warpdev.slack.com/archives/C08QY6ZBK2A/p1778877145314179
We are seeing this as we allocate more traffic to onnx classifier, which widens an existing stale async callback race.
The proposed solution is adding a counter autodetection_generation to track for the latest/valid async runs, and we only take the classification result if the counter value matches with the latest async run.
## Linked Issue

## Testing

### Screenshots / Videos

autodetect_generation increments as expected as user types ,h, he, hel,, hell, hello
image

updated screenshot with is_current_input
image

Note: since local build cannot reproduce the issue, we might have to merge it to see if it resolves the misfire in warp dev 
## Agent Mode


app/src/ai/blocklist/input_model.rs
```diff
@@ -157,6 +157,8 @@ pub struct BlocklistAIInputModel {
     terminal_view_id: EntityId,
 
     autodetect_abort_handle: Option<AbortHandle>,
+    // Id (or counter) to track the latest valid async task
+    autodetect_generation: u64,
```
szgupta: We already do some stale-buffer checks by comparing the current buffer to the one that was in the input when we kicked off some async work. Can we do the same thing here?

https://staging.warp.dev/debug/maa/a34e95a9-cd8e-488e-ae10-3ea1a74ed609?request=30688d57-a673-4ffb-9898-9499c5cffee7

evelyn-with-warp: https://github.com/warpdotdev/warp/pull/11061/changes/ae649d1dcf0147015dda4b384252a85b65074e12

app/src/ai/blocklist/input_model.rs
```diff
@@ -603,13 +603,17 @@ impl BlocklistAIInputModel {
...
     /// `EmptyCompletionContext`).
-    pub fn detect_and_set_input_type<C: CompletionContext + Clone + Send + 'static>(
+    pub fn detect_and_set_input_type<C, F>(
         &mut self,
         input: ParsedTokensSnapshot,
         completion_context: C,
         session_id: Option<SessionId>,
         ctx: &mut ModelContext<Self>,
-    ) {
+        is_current_input_check: F,
```
szgupta: Hrm this seems like a roundabout way to check (and maybe susceptible to ownership cycles) - can't we just check if the current buffer is equal to the buffer we did the computation on?

evelyn-with-warp: @szgupta do you mean something like this https://github.com/warpdotdev/warp/pull/11061/changes/b6fb566847e6343eabfab85b56233f714d221ef0 
This does make each call site lighter. AFAIK we still have to create a new field to track the up-to-date buffer text (as it hasn't implement current buffer text tracking)

szgupta:
@evelyn-with-warp i meant something more like keeping a `current_buffer_text: String` property on the input model, which is updated whenever there is an editor event in input view. The issue I have with the callback is that we have to move the editor model into the closure, which makes me a little bit worried about reference cycles.

evelyn-with-warp:
close it as 1. normally not reproducible after v0.2026.05.20.09.21.dev_00; 2. for reproducible build, the root cause is history match not async guard

PR #11083, opened 2026-05-15
Escape vim insert mode before leaving ampersand handoff

liliwilson:
## Description
Fixes an issue with vim escaping in the ampersand handoff view so that you can escape into normal mode before you escape handoff setup. Copies the same logic we use for doing this when the inline history menu is open.

## Testing
Unit test and I tested locally!

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/2b3ae6bd-0e53-4ecb-bbc7-581dd80e84b6_
_Run: https://oz.staging.warp.dev/runs/019e2df1-c5c2-7394-a922-0324d06cd929_


harryalbert (approved):
💌 vim

PR #11092, opened 2026-05-16
[5/5] Handle remote codebase auto-index follow-ups

moirahuang:
## Description

1. Allow for going from no auto indexing of repo -> auto starting indexing when the setting gets changed. Calling out that even if indexing hasn't happened for the repo, the agent knows that it has the search codebase tool available to itself (this is true for local)
2. Pass the number of allowed indexed repos

## Testing
Added unit tests + local tests
- [x] I have manually tested my changes locally with `./script/run`

### Screenshots / Videos
https://www.loom.com/share/a2e572c9338e43d0b329ef241256b91d

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: Fix remote codebase indexing incremental sync and queued-status reporting for indexes that do not actually start.

_Conversation: https://staging.warp.dev/conversation/cdd3638d-f1c3-479f-ab50-f62018beee7e_
_Run: https://oz.staging.warp.dev/runs/019e2e95-b78f-79f5-892a-56aa7d09b25a_
_This PR was generated with [Oz](https://warp.dev/oz)._

PR #10985, opened 2026-05-15
[2/5] remote codebase indexing telemetry

moirahuang:
## Description
Add telemetry for remote codebase indexing

## Testing

- [ ] I have manually tested my changes locally with `./script/run`

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

PR #10981, opened 2026-05-15
[1/5] [Remote codebase indexing] pass embedding configs through

moirahuang:
## Description
Passing embedding configs through for remote codebase indexing

Also made a small update to the refresh icon for remote envs to distinguish when there's already an index vs. when we need to create an index

## Testing
Added unit tests + locally tested
- [x] I have manually tested my changes locally with `./script/run`

### Screenshots / Videos
Screenshot 2026-05-14 at 8 17 05 PM

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

PR #11093, opened 2026-05-16
[3/5] Reduce duplicate remote codebase index status pushes

moirahuang:
## Description
Reduces duplicate codebase indexing notifications by 1) updating `push_all_codebase_index_statuses` to instead only push for 1 repo as `push_codebase_index_status_update` 2) dedupe on the status sent

## Testing
Added unit tests + locally tested

### Screenshots / Videos
https://www.loom.com/share/d20fe6af4a2e40acbbf0fb8f2194aeaa

## Agent Mode

### DIFF
--- a/app/src/remote_server/server_model.rs
+++ b/app/src/remote_server/server_model.rs
@@ -40,10 +40,10 @@ use super::proto::{
     get_fragment_metadata_from_hash_response, resolve_conflict_response, run_command_response,
     save_buffer_response, server_message, write_file_response, Abort, Authenticate, BranchInfo,
     BufferEdit, BufferUpdatedPush, ClientMessage, CloseBuffer, CodebaseIndexStatus,
-    CodebaseIndexStatusState, CodebaseIndexStatusUpdated, CodebaseIndexStatusesSnapshot,
-    DeleteFile, DeleteFileResponse, DeleteFileSuccess, DiscardFilesError, DiscardFilesResponse,
-    DiscardFilesSuccess, DropCodebaseIndex, ErrorCode, ErrorResponse, FailedFileRead,
-    FileContextProto, FileOperationError, FragmentMetadata as ProtoFragmentMetadata,
+    CodebaseIndexStatusUpdated, CodebaseIndexStatusesSnapshot, DeleteFile, DeleteFileResponse,
+    DeleteFileSuccess, DiscardFilesError, DiscardFilesResponse, DiscardFilesSuccess,
+    DropCodebaseIndex, ErrorCode, ErrorResponse, FailedFileRead, FileContextProto,
+    FileOperationError, FragmentMetadata as ProtoFragmentMetadata,
     FragmentMetadataLookupError as ProtoFragmentMetadataLookupError,
     FragmentMetadataLookupErrorCode, GetBranchesError, GetBranchesResponse, GetBranchesSuccess,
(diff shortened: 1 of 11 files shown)


app/src/remote_server/server_model.rs
```diff
@@ -216,6 +238,15 @@ pub struct ServerModel {
...
     diff_states: ModelHandle<RemoteDiffStateManager>,
+    /// Last remote codebase index status sent through the server protocol.
+    ///
+    /// The local index manager can emit several internal events for one
+    /// externally-visible status change. For example, finishing an incremental
+    /// sync can update sync state, persist index metadata, and then update sync
+    /// state again, while all three events serialize to the same remote
+    /// `Stale` status. This cache keeps that internal event fan-out from
+    /// becoming duplicate remote pushes.
+    last_pushed_codebase_index_statuses: HashMap<String, CodebaseIndexStatusPushKey>,
```
moirahuang: here i thought about if we can avoid emitting multiple events that are triggering the logs but since each of the events are important, i landed on it being better to dedupe so we only log once per change

moirahuang:
* **#11092** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11092?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11094** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11094?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11093** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11093?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11093?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10985** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10985?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10981** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/10981?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

app/src/settings_view/code_page.rs
```diff
@@ -199,7 +199,7 @@ impl CodeSettingsPageView {
             .count();
 
         ctx.subscribe_to_model(&index_manager, |me, index, event, ctx| {
-            if let CodebaseIndexManagerEvent::SyncStateUpdated = event {
+            if let CodebaseIndexManagerEvent::SyncStateUpdated { .. } = event {
```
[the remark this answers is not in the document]
kevinyang372: This seems valid?

app/src/remote_server/server_model.rs
kevinyang372: The main thing I think that's worth thinking through here is whether this dedupe logic should actually live in the codebase index manager layer rather than the server model

Like should we instead check if the codebase index state has changed in the manager before emitting SyncStateUpdated? That will better solve this case but prevent other unnecessary re-rendering cases when the state actually hasn't changed too

kevinyang372 (approved):
Approving to unblock

app/src/remote_server/server_model.rs
moirahuang: yeah, moved to the manager now

PR #11097, opened 2026-05-16
When opening a cloud agent convo, attach to existing session if there is one.

zachbai:
Fixes a number of issues with continuing cloud agent conversations.

1. When opening a cloud agent conversation - if the conversation is still running and has an active shared session, proactvely attach to the session. Previously we (implicitly, uniformly) assumed that the cloud agent conversation is inactive, and for a submitted user query, would assume that it needs to start a new task execution instead of just being sent as input to the existing shared session.
2. There wa a race condition arising from a misaligned assumption about follow-up requests synchronously moving tasks to IN_PROGRESS - the server does _not_ do this - the status change to IN_PROGRESS is async. So after sending a follow-up request, dont assume the task is already marked in progress, and tolerate a stale terminal task state while we poll for updated task state (up to 10x, ~30s)

Separately, also includes a fix for the pending user query occasionally staying visible even after the session starts - we were previously relying on exact query equality between the pending query and the actual user query streamed in from the shared session, which is a bad assumption (the server may add additional prompting to the query, and this comparison may fail).


zachbai:
* **#11097** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11097?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11097?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

moirahuang (approved):

PR #11101, opened 2026-05-16
Add model reasoning level for codex.

liliwilson:
## Description

This PR adds a few improvements to the model selector for third-party cloud agents:
- Add reasoning levels for codex models—we set up the reasoning level by writing to `model_reasoning_effort` in the codex config
- Make sure we show the third-party model selector during environment setup, rather than an unrelated model choice
- Make sure the `default` option is always accessible via the dropdown

## Testing

Tested locally with different harnesses and model selection choices.

### Screenshots / Videos

Loom attached to the associated server PR: https://github.com/warpdotdev/warp-server/pull/11224

## Agent Mode

### DIFF
--- a/app/src/ai/ambient_agents/task.rs
+++ b/app/src/ai/ambient_agents/task.rs
@@ -79,6 +79,16 @@ pub struct HarnessConfig {
     /// The model to use with this harness. None means use the harness default.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub model_id: Option<String>,
+    /// Optional reasoning level for harnesses that support it.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub reasoning_level: Option<String>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct HarnessModelConfig {
+    pub model_id: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub reasoning_level: Option<String>,
 }
 
 impl HarnessConfig {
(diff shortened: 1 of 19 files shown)


app/src/ai/ambient_agents/task.rs
```diff
@@ -79,6 +77,16 @@ pub struct HarnessConfig {
     /// The model to use with this harness. None means use the harness default.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub model_id: Option<String>,
+    /// Optional reasoning level for harnesses that support it.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub reasoning_level: Option<String>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct HarnessModelConfig {
```
liliwilson, line 88: Added this struct, even though it's slightly diff than the serde shape we get from the API, because it makes it easier to reason about state in the driver code

app/src/ai/cloud_agent_settings.rs
```diff
@@ -40,7 +56,7 @@ define_settings_group!(CloudAgentSettings, settings: [
         private: true,
     },
     last_selected_harness_model: LastSelectedHarnessModel {
-        type: HashMap<String, String>,
+        type: HashMap<String, HarnessModelSelection>,
```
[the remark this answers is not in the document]
liliwilson: I thought about this, but seeing as it hasn't hit prod yet I don't think this is worth doing. The new HarnessModelSelection struct also adds a layer of indirection so we can more easily add fields to this later without needing a legacy handling story

app/src/ai/ambient_agents/task.rs
abhishek-pandya: nice!

abhishek-pandya (approved):
thanks for grabbing this!

PR #11123, opened 2026-05-17
REV-1595 [5/n] Render per-user/team/own rows in cycle usage section

IsaiahWitzke:
image

image


app/src/settings_view/billing_and_usage/billing_cycle_usage_rows.rs
```diff
@@ -0,0 +1,1019 @@
...
+    /// Sorted by [`COST_TYPE_ORDER`] then [`BUCKET_ORDER`]; zero-credit entries dropped.
+    pub segments: Vec<BarSegment>,
+}
+
+fn member_base_limit(member: &WorkspaceMember) -> Option<i64> {
+    if member.usage_info.is_unlimited {
+        None
+    } else {
+        Some(member.usage_info.request_limit as i64)
+    }
```
tylerlam-warp, line 95: I haven't read the whole code yet so I can be missing some context, but if the request limit is 0 (i think for enterprises who don't have base limits, or service accounts) we should also return None?

app/src/settings_view/billing_and_usage/billing_cycle_usage_rows.rs
```diff
@@ -0,0 +1,1019 @@
...
+        }
+    }
+
+    segments.retain(|s| s.credits > 0);
+    segments.sort_by_key(segment_sort_key);
+
+    let total_credits = segments.iter().map(|s| s.credits).sum();
+    let total_cost_cents = segments.iter().map(|s| s.cost_cents).sum();
+
+    (segments, total_credits, total_cost_cents)
```
tylerlam-warp: nit: Instead of doing a two-pass here to get totals after the fact, can we just calculate these in the first pass?

app/src/settings_view/billing_and_usage/billing_cycle_usage_rows.rs
```diff
@@ -0,0 +1,1019 @@
...
+        .iter()
+        .filter(|e| entry_is_renderable(e))
+        .filter(|e| source_filter.matches(&e.usage_source))
+        // Defensive: filter to viewer-only even though OwnOnly should already be redacted.
+        .filter(|e| match (viewer_uid, e.subject_uid.as_deref()) {
+            (Some(uid), Some(entry_uid)) => uid == entry_uid,
+            (_, None) => true,
+            (None, _) => true,
+        })
+        .collect();
```
tylerlam-warp: nit: I _might_ be wrong but I think `collect_vec()` instead of `collect` here could save you from strongly typing the type for `viewer_entries`

app/src/settings_view/billing_and_usage/billing_cycle_usage_rows.rs
```diff
@@ -0,0 +1,1019 @@
...
+            (
+                entry.subject_type.clone(),
+                entry
+                    .subject_display_name
+                    .clone()
+                    .unwrap_or_else(|| "Unknown".to_string()),
+                Vec::new(),
+            )
+        });
+        group.2.push(entry.clone());
```
tylerlam-warp: not a huge fan of needing doing `group.2`, could we make the value a struct?

app/src/settings_view/billing_and_usage/billing_cycle_usage_section.rs
```diff
@@ -144,8 +175,10 @@ impl TypedActionView for BillingCycleUsageSectionView {
...
             }
             BillingCycleUsageAction::Refresh => {
-                let _ = TeamUpdateManager::handle(ctx)
-                    .update(ctx, |mgr, ctx| mgr.refresh_workspace_metadata(ctx));
+                std::mem::drop(
+                    TeamUpdateManager::handle(ctx)
+                        .update(ctx, |mgr, ctx| mgr.refresh_workspace_metadata(ctx)),
+                );
                 AIRequestUsageModel::handle(ctx)
                     .update(ctx, |m, ctx| m.refresh_request_usage_async(ctx));
```
tylerlam-warp: What's the benefit of using `mem::drop` here? Why aren't you doing it for refresh request usage async as well

tylerlam-warp:
I think there's something wrong coming in from the server, seeing duplicates here:

image

tylerlam-warp:
Also curious if you've tested this on your end? I haven't investigated too hard but I'm getting empty entries for some teams on my end that have made AI requests:

image

I think it might be a server thing?

tylerlam-warp (approved):
approved to unblock, left some comments and questions tho

app/src/settings_view/billing_and_usage/billing_cycle_usage_rows.rs
IsaiahWitzke: good point. ive updated this in the rendering side

app/src/settings_view/billing_and_usage/billing_cycle_usage_rows.rs
IsaiahWitzke: very cool. done

app/src/settings_view/billing_and_usage/billing_cycle_usage_rows.rs
IsaiahWitzke: very fair this entire func is pretty ugly. lemme clean

app/src/settings_view/billing_and_usage/billing_cycle_usage_section.rs
IsaiahWitzke: i think TeamUpdateManager's update had a `#[must_use]` or something? clippy was yelling at me for doing something like `let _ = ...`. Anyways, its gone now

PR #11124, opened 2026-05-17
Inline create-API-key flow on orchestration cards

advait-m:
Loom: https://www.loom.com/share/53bb766f3f9a4ecc983d516716a88627
Loom 2 (updates for plan cards): https://www.loom.com/share/532d7a88ef084b5baee269f041ad1ccc
Slack thread: https://warpdev.slack.com/archives/C0AAMT5TKC2/p1778989617745909?thread_ts=1778872788.221739&cid=C0AAMT5TKC2
Resolves https://linear.app/warpdotdev/issue/QUALITY-702/add-flow-to-create-a-new-key-when-orchestrating-non-oz-harness-in-the

### Product flow

Today, when an orchestration card surfaces for a non-Oz harness (Claude Code, Codex, etc.) and the user hasn't set up an API key for that harness yet, the card is effectively a dead-end: the "API key" picker is empty, Accept either silently dispatches with inherited environment credentials or fails downstream, and there's no clear "create one now" affordance. The only path forward is to drop out of the conversation, go find the cloud-mode FTUX, create a key there, then come back. That's a poor first-run experience and a real blocker for users trying to orchestrate with a fresh harness.

This PR fixes that by making the orchestration card itself the place where the user can finish setup. If a card needs an API key and none exist, the workspace pops a blocking modal that hosts the same create-key form used by cloud mode, scoped to the card's current harness. The picker also grows a permanent "+ New API key…" entry so users can re-open the modal any time, even when other keys already exist. The Accept button stays disabled — with a tooltip explaining why — until the user has either picked a managed key or explicitly clicked "Inherit key from environment" as a deliberate choice. After a key is created the card auto-selects it and re-enables Accept, so the user can keep going without a single context switch. The whole experience now matches what cloud mode has had for a while, and works identically whether the orchestration card was raised by the user or by an agent mid-task.

### Technical design

The bulk of the work was decoupling the existing create-key view from cloud mode's ambient agent model so it could be reused inside a generic workspace modal. Previously the FTUX view and its harness-picker dropdown both reached straight into the cloud-mode view model for harness selection and for the side effects that happen on success (persisting the selected secret, marking FTUX completed, etc.). Both now take a plain harness value at construction time, expose a setter for harness changes, and instead of mutating model state directly they emit a small set of lifecycle events — Created, Cancelled, Skipped, Failed — that their host subscribes to. Cloud mode's input layer wires those events back to the same side effects it used to perform inline, so its UX is unchanged.

On the card side, the user's auth-secret choice used to be represented as an optional name plus a sibling boolean for "explicitly inherit". That two-field encoding had a handful of subtle edge cases around defaulting, persistence, and round-tripping through the wire format. It's now collapsed into a single three-state enum — Unset, Inherit, Named — which makes the picker label, the Accept gate, and the per-harness persistence logic each reduce to a small match. The "+ New API key…" entry is a new variant on the shared orchestration-control action trait, so both the confirmation card and the plan-card config block can bubble it up to their parents and ultimately to the workspace, which owns the modal. The workspace subscribes to the FTUX view's lifecycle events to persist the new key and close the modal; cards listen for a harness-availability event so they can adopt the freshly-created key as their selection. A one-shot guard on each card ensures the auto-pop fires at most once per harness/mode change, so a user who cancels the modal isn't immediately re-prompted, but switching harness gives them a fresh chance.

## Testing


app/src/ai/document/orchestration_config_block.rs
```diff
@@ -222,13 +265,72 @@ impl OrchestrationConfigBlockView {
             details_mouse_state: MouseStateHandle::default(),
             saved_model_per_harness: HashMap::new(),
             suppress_refresh: false,
+            has_auto_opened_create_modal: false,
+            user_has_interacted: false,
         };
         if view.is_approved {
             view.ensure_pickers(ctx);
+            // Skip auto-open here: construction is also the restore code
```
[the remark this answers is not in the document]
advait-m: hm, don't think this is true for paths I tested - checked the restoration + create new plan paths (I specifically disabled for restoration)

app/src/ai/blocklist/inline_action/run_agents_card_view.rs
```diff
@@ -581,6 +610,11 @@ impl RunAgentsCardView {
         }
         resolve_interactive_defaults(&mut self.state, &*self.block_model, ctx);
         oc::repopulate_all_pickers(&mut self.state.orch, &self.handles.pickers, ctx);
+        self.refresh_accept_button_state(ctx);
+        // Stream completion is the authoritative final snapshot — re-arm
+        // the auto-open in case earlier evaluations fired against stale state.
+        self.has_auto_opened_create_modal = false;
```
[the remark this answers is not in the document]
advait-m: removed - I needed this earlier but don't need anymore

app/src/ai/blocklist/inline_action/orchestration_controls.rs
```diff
@@ -857,24 +882,70 @@ pub fn resolve_default_auth_secret_for_harness(
...
+        return false;
+    }
+    let Some(harness) = Harness::parse_orchestration_harness(&state.harness_type) else {
+        return false;
+    };
+    if harness == Harness::Oz || auth_secret_types_for_harness(harness).is_empty() {
+        return false;
+    }
+    let availability = HarnessAvailabilityModel::as_ref(ctx);
+    matches!(
```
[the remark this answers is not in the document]
advait-m: fixed - made it blocking

advait-m:
@cephalonaut @liliwilson - adding both of you for whoever has a chance to take a look first, thanks!

cc @abhishekp106 since this refactors out some cloud mode secrets logic into a common piece (I checked the cloud mode flow to ensure no changes)

app/src/ai/blocklist/inline_action/orchestration_controls.rs
```diff
@@ -930,32 +1002,89 @@ pub fn populate_auth_secret_picker_for_harness<A: OrchestrationControlAction, V:
...
+///
+/// Deliberately does NOT clear the persisted setting — cancel must not
+/// destroy the user's prior cross-session selection. Successful creation
+/// goes through the workspace modal's own persist + the
+/// `AuthSecretCreated` subscription.
+pub fn apply_create_new_auth_secret_requested<V: View>(
+    state: &mut OrchestrationEditState,
+    _ctx: &mut ViewContext<V>,
+) {
+    state.auth_secret_selection = AuthSecretSelection::Unset;
```
cephalonaut: Opening “New API key…” overwrites the current selection with `Unset`, and modal cancel doesn’t restore it. If the user had an existing key selected, opens the modal, then cancels, Accept becomes disabled. Please preserve the prior selection until creation succeeds, or restore it on cancel.

app/src/ai/blocklist/inline_action/run_agents_card_view.rs
```diff
@@ -129,7 +132,7 @@ impl RunAgentsEditState {
             execution_mode: self.orch.execution_mode.clone(),
             agent_run_configs: self.agent_run_configs.clone(),
             plan_id: self.plan_id.clone(),
-            harness_auth_secret_name: self.orch.auth_secret_name.clone(),
+            harness_auth_secret_name: self.orch.auth_secret_name().map(str::to_string),
```
cephalonaut, line 135: Switching away from Cloud/non-Oz can leave a stale `Named(_)` auth secret in state, and `to_request()` will still serialize it. Please clear auth selection when the picker no longer applies, or make `auth_secret_name()` return `None` unless the current mode/harness supports auth secrets.

app/src/ai/document/orchestration_config_block.rs
```diff
@@ -627,13 +746,8 @@ impl TypedActionView for OrchestrationConfigBlockView {
                 ctx.notify();
             }
             OrchestrationConfigBlockAction::AuthSecretChanged { auth_secret_name } => {
```
cephalonaut, line 749: The plan-card “Inherit key from environment” choice is only in-memory. When the later `RunAgents` card is built, it reconstructs auth state from persisted named secrets only, so an approved plan can still block/re-prompt instead of honoring Inherit. Please carry explicit inherit through the plan/config handoff.

app/src/view_components/compactible_split_action_button.rs
```diff
@@ -107,6 +107,19 @@ impl CompactibleSplitActionButton {
     ) {
         self.primary_button.set_keybinding(keybinding, ctx);
     }
+
+    /// Sets the disabled state on both the primary and menu buttons.
+    pub fn set_disabled<T: View>(&mut self, disabled: bool, ctx: &mut ViewContext<T>) {
+        self.primary_button.set_disabled(disabled, ctx);
```
cephalonaut, line 113: Disabling the split button disables both Accept and the “Accept w/o orchestration” menu. Auth-secret validation only applies to launching orchestration, so the fallback should remain available. Please support disabling only the primary action for this gate.

app/src/terminal/view/ambient_agent/auth_secret_ftux_view.rs
```diff
@@ -396,25 +629,15 @@ impl AuthSecretFtuxView {
...
-        });
-        CloudAgentSettings::handle(ctx).update(ctx, |settings, ctx| {
-            settings.mark_harness_auth_ftux_completed(harness, ctx);
-            // Persist the newly created secret so it is sticky across sessions.
-            let mut map = settings.last_selected_auth_secret.value().clone();
-            map.insert(harness.config_name().to_string(), name);
-            let _ = settings.last_selected_auth_secret.set_value(map, ctx);
-        });
         self.clear_creation_state(ctx);
+        ctx.emit(AuthSecretFtuxViewEvent::Created { harness, name });
```
cephalonaut, line 689: Any global `AuthSecretCreated` event is treated as this view’s successful submit whenever `creation_state` exists. With this component reused in multiple hosts, another creation event could close/select the wrong modal. Please match against the pending harness/name/request before emitting `Created`.

cephalonaut (approved):
Looks good - comments are at your discretion

advait-m:
> Looks good - comments are at your discretion

Thanks - will address these!

app/src/ai/blocklist/inline_action/orchestration_controls.rs
advait-m: agree this would be better, fixed

app/src/view_components/compactible_split_action_button.rs
advait-m: hmm, wouldn't it apply to non-orchestration but multi-harness case (i.e. single agent claude code) too? 

but yeah, I think this piece is fine, leaving as is - can return if folks are confused

app/src/ai/blocklist/inline_action/run_agents_card_view.rs
advait-m: thanks - fixing

app/src/ai/document/orchestration_config_block.rs
advait-m: thanks - fixing

app/src/terminal/view/ambient_agent/auth_secret_ftux_view.rs
advait-m: thanks, fixing

PR #11133, opened 2026-05-17
Make agent run skill flag repeatable

bnavetta:
## Summary

This makes the `--skill` flag on `oz agent run` repeatable. We'll use this to fully support multiple run-attached skills. All requested skills are resolved and registered with `SkillsManager`. However, only the _first_ skill is treated as the invoked skill.

In combination with warpdotdev/warp-server#11252, we'll correctly prioritize all attached skills for a given run.

While testing this, I fixed local handling of skill invocation with a prompt. We'll now use an `InvokeSkill` message, instead concatenating the skill and prompt client-side.

## Testing

- Ran `cargo fmt --manifest-path /Users/ben/src/warp/Cargo.toml --all`.
- Ran focused tests:
  - `cargo nextest run --manifest-path /Users/ben/src/warp/Cargo.toml -p warp handle_skills_added_registers_multiple_explicit_skills cloud_environment_skills_always_included`
  - Result: 2 passed.
- Ran targeted clippy:
  - `cargo clippy --manifest-path /Users/ben/src/warp/Cargo.toml -p warp --lib -- -D warnings`
  - Result: passed.
- Manually tested repeatable `--skill` with `./script/run` using real sandboxed skills.
  - Confirmed same-repo skills only clone once.
  - Confirmed cross-repo skills clone both repos.
  - Confirmed that the agent reports both skills in context.

Example output from a test run:

```sh
$ ./script/run -- agent run --cwd work --skill warpdotdev/common-skills:resolve-merge-conflicts --skill warpdotdev/oz-for-oss:bootstrap-issue-config --skill warpdotdev/oz-for-oss:verify-pr --prompt "Please confirm that you have access to verify-pr, bootstrap-issue-config, and resolve-merge-conflicts skills" --sandboxed
...
Run ID: 019e418c-759e-7b29-b0d4-99e91fba4cad
Open in Oz: https://oz.staging.warp.dev/runs/019e418c-759e-7b29-b0d4-99e91fba4cad

New conversation started with debug ID: e79169fe-8a9b-4577-8fb2-6905e1c8b70d

I can confirm that I have access to all three skills you mentioned:

1. **resolve-merge-conflicts** - Available at multiple paths including `/Users/ben/src/warp-internal/.agents/skills/resolve-merge-conflicts/SKILL.md`, `/Users/ben/src/warp/work/common-skills/.agents/skills/resolve-merge-conflicts/SKILL.md`, and `/Users/ben/.agents/skills/resolve-merge-conflicts/SKILL.md`

2. **bootstrap-issue-config** - Available at `/Users/ben/src/warp/work/oz-for-oss/.agents/skills/bootstrap-issue-config/SKILL.md`

3. **verify-pr** - Available at `/Users/ben/src/warp/work/oz-for-oss/.agents/skills/verify-pr/SKILL.md`

All three skills are in the curated skills list and ready to use. How would you like me to proceed with these skills?
```

_Conversation: https://staging.warp.dev/conversation/859faac9-88c6-4cd1-b33e-a37f3efe44fd_
_Run: https://oz.staging.warp.dev/runs/019e34e0-eae3-7f3a-b66c-ca1e13e3c6f5_

### DIFF
--- a/app/src/ai/skills/file_watchers/utils.rs
+++ b/app/src/ai/skills/file_watchers/utils.rs
@@ -1,12 +1,9 @@
-use std::collections::HashSet;
 use std::path::{Path, PathBuf};
-use std::sync::LazyLock;
 
 use ai::skills::{
     home_skills_path, read_skills, ParsedSkill, SkillProvider, SKILL_PROVIDER_DEFINITIONS,
 };
 use anyhow::Error;
-use regex::Regex;
 use repo_metadata::local_model::GetContentsArgs;
 use repo_metadata::{RepoContent, RepoMetadataModel};
 use warpui::AppContext;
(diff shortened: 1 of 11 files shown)


app/src/ai/skills/file_watchers/utils.rs
```diff
@@ -69,52 +64,45 @@ pub fn is_skill_file(path: &Path) -> bool {
...
 
-static SKILL_PROVIDER_PATHS: LazyLock<HashSet<String>> = LazyLock::new(|| {
-    // Collect the skill provider paths from the definitions
-    SKILL_PROVIDER_DEFINITIONS
-        .iter()
-        .map(|p| p.skills_path.to_string_lossy().to_string())
-        .collect()
-});
-
-// Pattern: {prefix}/{provider_path}/{skill-name}/SKILL.md
```
bnavetta, line 79: I replaced these regexes with path parsing - this should be more robust in general, and makes it easier to handle mixed path separators on Windows

specs/repeatable-oz-agent-run-skill/PRODUCT.md
```diff
@@ -0,0 +1,45 @@
...
+- Preserve existing behavior for commands that provide zero or one `--skill`.
+## Non-goals
+- Changing `oz agent run-cloud --skill` behavior.
+- Changing skill resolution precedence or accepted skill spec formats.
+- Combining multiple skill instruction bodies into the initial prompt.
+- Changing the skill file format or the displayed skill-invoked output.
+## Figma / design references
+Figma: none provided.
+## User Experience
+1. A user can run `oz agent run --skill primary --skill helper --prompt "Do the task"`.
```
vkodithala, line 20: Is the idea here that the user can specify any number of `--skill <SPEC>` arguments, and only the first is considered "invoked"? Any reason why the signature here is `--skill <primary> --skill <helper>` vs. something like `--skill <primary> <helper> <helper>`? (not blocking, just curious on the design)

vkodithala: Also, is there a reason we treat the first skill in this sequence as 'invoked' vs. the rest as helpers? Wondering why this is the intended behavior vs. loading all specified skills into the environment w/o explicitly invoking any.

app/src/ai/skills/file_watchers/utils.rs
vkodithala: Think this is great and agree that it avoids some redundancy on Windows!

vkodithala (approved):
Specs LGTM, thanks for the fix Ben! Left a few comments but mostly around product thinking & for my own understanding!

specs/repeatable-oz-agent-run-skill/PRODUCT.md
vkodithala: If the answers here are reiterated in some doc/discussion, please link! I'm pretty out of the loop here

bnavetta: > Is the idea here that the user can specify any number of `--skill <SPEC>` arguments, and only the first is considered "invoked"?

> Also, is there a reason we treat the first skill in this sequence as 'invoked' vs. the rest as helpers? Wondering why this is the intended behavior vs. loading all specified skills into the environment w/o explicitly invoking any.

Ah there was some discussion offline before the Orchestrate launch on how to handle this. The context for invoking the first skill is that we're in a sort of awkward migration phase for agents/skills where invocation doesn't make as much sense once an agent has multiple skills, but we also need to not break existing user workflows that run a single skill. To that end, I'd prefer that a command like `oz agent run --skill <my-skill>` still invoke the skill (especially because this lets you specify a skill rather than a prompt).

The question is really then what to do with `oz agent run --skill <first-skill> --skill <second-skill>`. I had decided that invoking the first skill is the least surprising option, so that adding a second skill doesn't significantly change how the agent behaves. However, given that any existing invocation-based workflows will only have a single skill, it should be safe to invoke if and only if there's exactly one skill. That's how I'm leaning, but curious what you and @ianhodge think!

> Any reason why the signature here is `--skill <primary> --skill <helper>` vs. something like `--skill <primary> <helper> <helper>`? (not blocking, just curious on the design)

Yeah, repeating the option name is generally safer with multi-value arguments. The problem with `--skill <primary> <helper> <helper>` is that it's indistinguishable from `--skill <primary> <helper> <unrelated-positional-argument>`. While the `agent run` subcommand doesn't have any positional arguments today, we shouldn't rule out that possibility.

PR #11162, opened 2026-05-18
Add orchestration telemetry events

advait-m:
Loom: https://www.loom.com/share/9dc498c4adae4fd39bc9b0183b57575a
Resolves: https://linear.app/warpdotdev/issue/QUALITY-683/do-a-telemetry-pass-on-client-server

## Description

Adds client-side telemetry to instrument every orchestration entry point and the run-wide config / pill-bar interactions that follow. Prior to this change, the only orchestration telemetry on the client was `AgentMode.Orchestration.TeamAgentCommunicationFailed` (error path only); we had no visibility into how often users entered orchestration, which surface they entered from, how they edited the configs, or how they used the pill bar. Server-side `AgentMode.OrchestrateOutcome` told us *that* the launched config diverged from the tool call, but not *which* fields, and there was no analog of `is_planning_query` for `/orchestrate` queries.

This PR is the **client** half. The **server** half is up at warpdotdev/warp-server#11236 — it adds `is_orchestrate_query` everywhere, the launched config fingerprint on `OrchestrateOutcome`, and `parent_execution_location` on `StartAgentInvocation`. The pair gives us closed-enum, UGC-free coverage across every orchestration surface.

> Note: an earlier revision of this PR pair routed agent-driven config-creation telemetry through a server-side `CreateConfigInvocation` event. That event has been removed in favor of the client-side `AgentMode.Orchestration.AgentProposedConfig` event described below, so all plan-card telemetry is co-located client-side.

### Design principles

- **Metadata-only.** Every new payload is enums, IDs, booleans, and counts. No prompts, summaries, agent names, environment names, worker host slugs, or auth-secret labels are ever serialized.
- **Closed enums for free-form values.** `OrchestrationHarnessKind` and `OrchestrationExecutionModeKind` bucket harness strings and execution modes into a closed set so the analytics columns stay low-cardinality even if the server adds a new harness.
- **Idempotent at the surface level.** Each card instance emits at most one `Entered` and one `Decision` event (guarded by `entered_event_emitted` / `decision_event_emitted`) so re-renders don't double-count.
- **No UGC, period.** Free-form fields that are arguably "server-controlled identifiers" (e.g. environment_id) are still collapsed to booleans (`has_environment`, `had_previous`, `now_set`) to avoid leaking workspace topology through analytics.

### New events

All variants live on `BlocklistOrchestrationTelemetryEvent` in `app/src/ai/blocklist/telemetry.rs`.

#### `AgentMode.Orchestration.PlanConfigApprovalToggled`
Fires from `OrchestrationConfigBlockView::handle_action` when the user toggles the **Use orchestration** switch on a plan card. The user-driven counterpart to `AgentProposedConfig` (below), which captures agent-driven config snapshots.

| Field | Type | Notes |
| --- | --- | --- |
| `conversation_id` | id | |
| `plan_id` | optional string | omitted when empty |
| `status` | `approved \| disapproved` | state after the toggle |
| `execution_mode` | `local \| remote` | |
| `harness` | `oz \| claude_code \| codex \| open_code \| gemini \| unknown` | |
| `has_model`, `has_environment`, `has_worker_host`, `has_auth_secret` | bool | presence-only |

#### `AgentMode.Orchestration.Entered`
Fires once per orchestration entry, attributing the source. Two call sites:

- **`BlocklistAIController::send_query`** — when the extracted `UserQueryMode == Orchestrate` (slash command). Emits with `entry_source = slash_command_orchestrate`.
- **`RunAgentsCardView::try_auto_launch_on_stream_complete`** — once per card instance when the card is actually rendered (i.e. *not* in the auto-launched path). Emits with `entry_source = run_agents_card_shown`.

Plan-card entry is intentionally not a variant here: it's covered by `AgentProposedConfig` (agent surfaced a snapshot) plus `PlanConfigApprovalToggled` (user committed via the toggle), both already in this PR.

| Field | Type |
| --- | --- |
| `conversation_id` | id |
| `plan_id` | optional string |
| `entry_source` | `slash_command_orchestrate \| run_agents_card_shown` |

#### `AgentMode.Orchestration.RunAgentsCardDecision`
Fires once per `RunAgentsCardView` instance on **Accept**, **Accept without orchestration**, or **Reject** (guarded by `decision_event_emitted`).

| Field | Type | Notes |
| --- | --- | --- |
| `conversation_id`, `plan_id` | id / optional string | |
| `decision` | `accept \| accept_without_orchestration \| reject` | |
| `agent_count` | usize | |
| `harness`, `execution_mode` | enums above | |
| `modified_fields_from_tool_call` | `Vec<&'static str>` | names of run-wide fields that diverged from the LLM's original `RunAgentsRequest`. Values are stable identifiers in `orchestration_modified_field::{MODEL_ID, HARNESS, EXECUTION_MODE, ENVIRONMENT_ID, WORKER_HOST, AUTH_SECRET}` — the same strings the server's `RunAgentsOutcome.modified_fields` (`RunAgentsModifiedField*` in the server PR) uses so the two streams join 1:1. |
| `modified_fields_from_active_config` | `Vec<&'static str>` | same shape, compared against the approved orchestration config snapshot when one exists. Empty when no approved config exists or the dispatched config matches. |
| `had_active_config`, `active_config_status` | bool / optional `approved \| disapproved` | |

This event answers questions like:
- *How often do users accept orchestration as the LLM proposed it vs. tweak the config first?*
- *Which fields do users most commonly edit before accepting?*
- *How often do users hit "Accept without orchestration" to opt out of an approved plan?*

To compute the per-edit diff we snapshot `original_tool_call_request` on every `update_request` call (so it always reflects the final streamed request) and diff against `state.orch` at decision time. See `diverged_orch_fields` / `diverged_orch_fields_against_config` in `run_agents_card_view.rs`.

#### `AgentMode.Orchestration.AgentProposedConfig`
Fires once per `OrchestrationConfigBlockView` instance when an agent-authored `OrchestrationConfigSnapshot` first becomes visible to the user on a plan card. Replaces the server-side `CreateConfigInvocation` event so all plan-card telemetry sits on the client. Server-only fields from the earlier design (`validation_outcome`, `request_id`, `tool_call_id`, `task_id`, `orchestration_version`, `is_planning_query`, `is_orchestrate_query`) are dropped — they require plumbing the client doesn't naturally have, and invalid configs never surface to the client anyway (no snapshot is produced).

| Field | Type | Notes |
| --- | --- | --- |
| `conversation_id` | id | |
| `plan_id` | optional string | omitted when empty |
| `harness` | `oz \| claude_code \| codex \| open_code \| gemini \| unknown` | |
| `execution_mode` | `local \| remote` | |
| `has_model`, `has_environment`, `has_worker_host` | bool | presence-only |

#### `AgentMode.Orchestration.PillBarInteraction`
Fires from `OrchestrationPillBar::handle_action` via `emit_pill_bar_interaction`. Captures every meaningful pill-bar interaction; hover events are intentionally not tracked.

| Field | Type | Notes |
| --- | --- | --- |
| `action` | enum | `switch \| open_in_new_pane \| open_in_new_tab \| focus_opened_conversation \| stop \| kill \| toggle_pin_on \| toggle_pin_off \| open_menu`. `switch` covers all pill-body clicks (slice by `switch_outcome`); `focus_opened_conversation` is reserved for the menu-driven "Focus pane" gesture. |
| `pill_kind` | `orchestrator \| child` | |
| `total_pills`, `total_pinned` | usize | resolved from the current `pill_specs` at emit time |
| `source_conversation_id` | id | the orchestrator at the root of this bar |
| `target_conversation_id` | id | the pill being acted on |
| `switch_outcome` | optional `switched_in_place \| focused_existing_pane` | present only when `action = switch`; closed enum so future outcomes don't require a new action variant |

To make `Switch` (pill click) telemeterable, the body click handler now routes through a new `OrchestrationPillBarAction::PillClicked { conversation_id, pill_kind }` rather than dispatching the navigation `TerminalAction` directly from the click closure. `handle_action` emits the telemetry, then reproduces the original "focus opened pane vs. switch in place" logic. Net behavior is unchanged.

All body clicks always emit a single `switch` event, with `switch_outcome` capturing the resulting navigation — "how often do users click pills?" is `WHERE action = 'switch'`, no UNION needed. The menu-driven "Focus pane" gesture stays its own `focus_opened_conversation` event so the two user intents remain separable. The shared focus-existing-pane navigation lives in a private `navigate_to_owner_pane` helper reused by both code paths.

No labels (which can be user/agent-generated agent names) are ever sent.

### Pre-existing telemetry that stays as-is

- `AgentMode.Orchestration.TeamAgentCommunicationFailed` — unchanged. Still the only client-side error-path telemetry.

### Cross-side coverage map

| Surface | Action | Where it's tracked |
| --- | --- | --- |
| Plan card | Agent calls `create_orchestration_config` | **Client** (this PR) — `AgentProposedConfig` (fires when the snapshot first becomes visible) |
| Plan card | User edits / toggles approval | **Client** (this PR) — `PlanConfigApprovalToggled` |
| run_agents card | Tool call lands & is decided | **Server** — `OrchestrateOutcome` (this PR's paired server change adds launched fingerprint + `is_orchestrate_query`) |
| run_agents card | Card rendered | **Client** (this PR) — `Entered { entry_source: run_agents_card_shown }` |
| run_agents card | User accepts / rejects | **Client** (this PR) — `RunAgentsCardDecision` |
| /orchestrate slash | User submits | **Client** (this PR) — `Entered { entry_source: slash_command_orchestrate }` |
| /orchestrate downstream | Tool calls inherit attribution | **Server** — `is_orchestrate_query` now present on `OrchestrateOutcome` and `StartAgentInvocation` |
| Pill bar | User navigates between agents | **Client** (this PR) — `PillBarInteraction` |
| start_agent | Lead invokes single-agent | **Server** — `StartAgentInvocation` (paired PR adds `parent_execution_location` + `is_orchestrate_query`) |

### Companion server PR

warpdotdev/warp-server#11236 — additive, can ship independently in either order.


app/src/ai/document/orchestration_config_block.rs
```diff
@@ -560,10 +582,28 @@ impl TypedActionView for OrchestrationConfigBlockView {
...
+                } else {
+                    OrchestrationApprovalStatus::Disapproved
+                };
+                self.emit_plan_config_approval_toggled(previous_status, new_status, ctx);
+                // Treat the off→on transition as the moment the user
+                // committed to using orchestration via the plan card.
+                if matches!(previous_status, OrchestrationApprovalStatus::Disapproved)
+                    && matches!(new_status, OrchestrationApprovalStatus::Approved)
+                {
+                    self.emit_orchestration_entered(ctx);
```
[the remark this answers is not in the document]
advait-m: fixed

app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
```diff
@@ -706,11 +819,57 @@ impl TypedActionView for OrchestrationPillBar {
...
                 OrchestrationPillBarModel::handle(ctx).update(ctx, |model, ctx| {
                     model.toggle_pin(id, ctx);
                 });
             }
+            OrchestrationPillBarAction::PillClicked {
+                conversation_id,
+                pill_kind,
+            } => {
+                let id = *conversation_id;
+                self.emit_pill_bar_interaction(
```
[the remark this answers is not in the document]
advait-m: fixed

advait-m:
Spoke w Daniel on slack wrt these PRs - moving `AgentMode.Orchestration.CreateConfigInvocation` over to client-side to have it alongside other plan card telemetry :+1:

app/src/ai/blocklist/telemetry.rs
```diff
@@ -58,6 +63,190 @@ pub(crate) struct TeamAgentCommunicationFailedEvent {
...
+    pub const WORKER_HOST: &str = "worker_host";
+    pub const AUTH_SECRET: &str = "auth_secret";
+}
+
+#[derive(Debug, Serialize)]
+pub(crate) struct PlanConfigApprovalToggledEvent {
+    pub conversation_id: AIConversationId,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub plan_id: Option<String>,
+    pub previous_status: OrchestrationApprovalStatus,
```
danielpeng2: nit: is this redundant? unless you expect there will be more states to track other than `approved` / `disapproved` in the future.

app/src/ai/blocklist/telemetry.rs
```diff
@@ -58,6 +63,190 @@ pub(crate) struct TeamAgentCommunicationFailedEvent {
...
+}
+
+/// Surface that first introduced orchestration into a conversation.
+#[derive(Clone, Copy, Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum OrchestrationEntrySource {
+    /// `/orchestrate` slash-command mode on a user query.
+    SlashCommandOrchestrate,
+    /// Plan-card `Use orchestration` switch toggled to approved.
+    PlanCardApproved,
```
danielpeng2: Is `PlanCardApproved` the right event to track here? What if the user toggles the state to `disapproved` then back to `approved`, would we track an `Entered` event again?

Feels like here we'd want to track the plan card being shown (basically the `AgentProposedConfig` event we're also adding).

app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
```diff
@@ -706,11 +810,57 @@ impl TypedActionView for OrchestrationPillBar {
...
+                let is_open_elsewhere =
+                    is_conversation_open_in_other_visible_view(id, self_terminal_view_id, ctx);
+                if is_open_elsewhere {
+                    // FocusOpenedConversation's handler emits its own
+                    // telemetry; skip the Switch event so one click =
+                    // one event.
+                    ctx.dispatch_typed_action(
+                        &OrchestrationPillBarAction::FocusOpenedConversation(id),
+                    );
+                } else {
```
danielpeng2, line 965: thoughts on whether this makes the telemetry harder to work with? might not be clear that to track all switches, we have to join `FocusOpenedConversation` and the `Switch` events

danielpeng2 (approved):
approving to unblock. my main question is around the right places to track in the `Entered` event.

app/src/ai/blocklist/telemetry.rs
advait-m: yeah, this was mostly from that angle but agree we can just remove for now.

removing.

app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
advait-m: yeah, I originally did have it emitting both but Oz comment made me consider switching it to this 😅. I agree emitting both makes telem easier to query tho.

Let me actually simplify this and make `switch` have `switched_in_place`, `focused_existing_pane` as an enum field (only 2 for now, but in case we need to add more in the future - and enum is a bit easier to understand than a bool here). This better matches semantically too.

Note that this does require some logic to be extracted out to a helper function (for `navigate_to_owner_pane`) - doing that now.

app/src/ai/blocklist/telemetry.rs
advait-m: so I think what you're saying should be covered by the bool we're tracking already
```
                // First off→on is the plan-card entry point; later
                // re-toggles don't re-fire (matches RunAgentsCardView).
                if !was_approved && self.is_approved && !self.entered_event_emitted {
                    self.entered_event_emitted = true;
                    self.emit_orchestration_entered(ctx);
                }
```

re the second piece - hmm, yeah now that we moved that over to client there's a decent bit of overlap there. 

This enum is already a bit mixed wrt agent-driven vs user-driven (orchestrate slash command is user-driven, run agents card could be either kinda depending on how strongly the user prompted it)

Anyhow, I think it's reasonable to replace this with `AgentProposedConfig` (a bit different, and will have a lot more events, but has overlap). Working on that change.

advait-m: though yeah, not sure if putting that under `OrchestrationEntrySource` makes much sense 🤔 - perhaps just dropping this enum variant now that we have the other event client-side is fine. Did that.

PR #11172, opened 2026-05-18
Persist killed child agents and rename Kill→Delete after terminal state

advait-m:
Loom: https://www.loom.com/share/45edb8dbfdad48409c63e1dd4396916c
Fixes https://linear.app/warpdotdev/issue/QUALITY-669/ensure-killing-persist-across-app-restarts

## Description

Two small fixes to the orchestration pill bar's child-agent menu.

**1. Killing a child agent is now durable across app restarts.**

**2. "Kill agent" → "Delete agent" once the agent has reached a
terminal state.**

Stop is unchanged. Orchestrator pills are unaffected.

### Heads-up on widened semantics

- **Cloud delete is now unconditional for Kill.** This should match semantic intent + we don't want this in cloud convos.
  - Archive is a different action, we can follow up on adding this later (the current semantics map to removal/deletion only)

## Testing

- `cargo fmt` clean.
- `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings` clean.

### DIFF
--- a/app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
+++ b/app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
@@ -429,16 +429,17 @@ impl OrchestrationPillBar {
                 ),
             ]
         };
-        // Stop is shown only while in progress; Kill becomes Delete once
-        // the agent has reached a terminal state. Blocked is treated as
-        // non-terminal (still mid-flight, waiting on user input).
+        // Stop is shown only while the agent is in progress; Kill becomes
+        // Delete once the agent's run has finished (Success / Error /
+        // Cancelled). Blocked is treated as not-yet-finished (the agent
+        // is still mid-flight, waiting on user input).
         let conversation_status = BlocklistAIHistoryModel::as_ref(ctx)
             .conversation(&conversation_id)
             .map(|conversation| conversation.status().clone());
         let is_in_progress = conversation_status
             .as_ref()
             .is_some_and(|status| status.is_in_progress());
-        let is_terminal = conversation_status
+        let is_in_finished_state = conversation_status
             .as_ref()
             .is_some_and(|status| status.is_done());
         items.push(MenuItem::Separator);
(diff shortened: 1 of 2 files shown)


app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
```diff
@@ -429,9 +429,18 @@ impl OrchestrationPillBar {
...
+        // the agent has reached a terminal state. Blocked is treated as
+        // non-terminal (still mid-flight, waiting on user input).
+        let conversation_status = BlocklistAIHistoryModel::as_ref(ctx)
             .conversation(&conversation_id)
-            .is_some_and(|conversation| conversation.status().is_in_progress());
+            .map(|conversation| conversation.status().clone());
+        let is_in_progress = conversation_status
+            .as_ref()
+            .is_some_and(|status| status.is_in_progress());
+        let is_terminal = conversation_status
```
danielpeng2: nit: `is_terminal` is an overloaded term for us, might be clearer if we named this something else, maybe `is_done`?

danielpeng2 (approved):

app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
advait-m: haha yes I agree

hm - done is a bit weird since it implies an agent state too

maybe a lot more verbose and do `is_in_terminal_agent_state` or similar

advait-m: i.e. `Done` is a successfully done agent for us

advait-m: im doing `is_in_finished_state`

advait-m: oh yeah, the prior helper is a bit misleading with `is_done`, but that's fine

PR #11181, opened 2026-05-18
Fix stale dependency definitions causing unnecessary rebuilds.

vorporeal:
## Description

Noticed that `cargo check -p warp` wasn't taking advantage of the cache for all crates, even when I hadn't modified any files.

Asked agent mode, and it found two stale dependencies between crates in our workspace.  I cleaned these up, and it fixed the issue.  (One of them was a test-only dependency, but we didn't even really need the dependency at all, so I cleaned it up.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -562,7 +562,6 @@ dependencies = [
  "tower-http",
  "tracing",
  "tracing-subscriber",
- "warp_cli",
  "win32job",
 ]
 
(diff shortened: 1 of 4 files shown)


vorporeal:
* **#11181** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11181?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11181?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

bnavetta (approved):

PR #11187, opened 2026-05-18
Fix agent dropdown rendering in new API key modal

bnavetta:
## Description

This fixes a rendering bug in the API key creation modal. The modal content was showing through the dropdown to select which agent the key is for.

The problem was that the agent dropdown menu renders as an overlay, but the modal is already an overlay. As a result, other modal content wouldn't render fully below the dropdown - rendering instead depends on the order that the UI framework processes elements in. Here, it would paint the collapsed exipry selector dropdown on top of the opened agent dropdown menu.

This category of bugs has been a longstanding issue with the UI framework's z-index handling. As a workaround for now, we can:
1. Reserve space in the modal for the agent dropdown
2. Render the dropdown in a new `Stack` as an overlay that's painted after the other modal contents

## Testing

### Screenshots / Videos
Screenshot 2026-05-20 at 1 41 35 AM

## Agent Mode


jasonkeung (approved):
lgtm after changing the width

PR #11195, opened 2026-05-18
Fix asset-loading errors on WASM

bnavetta:
## Description

This fixes two bugs in our asset cache, which I noticed while debugging Warp-on-Web errors:
1. Caching asset data to disk was not correctly gated. This meant that loading the settings page on the web would panic, since we cache the user's profile picture. This may have been dev-only, since I couldn't reproduce on stable.
2. We didn't check the HTTP status code when fetching assets by URL. Instead, we'd try to parse whatever the error response was. For example, I got a 429 error back from Google, and we tried to parse it as an SVG.

## Testing

Tested by running the WASM app and server together, and opening the settings page.

## Agent Mode

### DIFF
--- a/crates/asset_cache/src/lib.rs
+++ b/crates/asset_cache/src/lib.rs
@@ -3,9 +3,7 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use anyhow::Result;
-use async_fs::{OpenOptions, create_dir_all};
 use bytes::Bytes;
-use futures::AsyncWriteExt;
 use reqwest::Url;
 use warpui_core::assets::asset_cache::{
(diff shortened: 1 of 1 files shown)


crates/asset_cache/src/lib.rs
```diff
@@ -109,7 +107,11 @@ fn get_file_path_for_asset(url: &Url, cache_dir: &Path) -> PathBuf {
     cache_dir.join(filename)
 }
 
+#[cfg(not(target_family = "wasm"))]
 async fn persist_bytes(bytes: &Bytes, file: &Path) {
```
bnavetta, line 111: We could probably implement a WASM version of asset caching with something like https://emnudge.dev/notes/cache-api/, but I don't think it's worth it right now - profile pictures are the main thing we seem to cache on-disk.

vorporeal (approved):

PR #11205, opened 2026-05-18
fix: Highlight focused AI block matches under async find

vkodithala:
## Description

Highlights focused AI block matches under async find. With the `AsyncFind` feature flag enabled (default-off for dogfood), pressing up/down on the find bar advances the `current/total` counter through AI matches but never (1) applies the focused-match highlight color to the AI match, (2) scrolls to the AI block containing the focused match, or (3) auto-expands a collapsed reasoning block containing it. Sync find handles all three correctly.

Root cause: `AsyncFindController`'s focused-match cache was terminal-only, so the AI-rendering call site and the reasoning auto-expand handler — both of which look up the focused `RichContentMatchId` via the sync `BlockListFindRun` — see `None` whenever async find is active.

Fix:

- Refactor the controller's focused-match cache to a single `Option<FocusedMatchResolution>` that resolves to either a terminal grid match or an AI match, replacing the previous terminal-only `Option<AsyncBlockGridMatch>`. `focused_terminal_match()` and `focused_ai_match()` are thin pattern-matching accessors over the unified cache.
- Plumb the AI variant through `TerminalFindModel::focused_block_list_match()` so `TerminalView::scroll_to_match` routes to `FindMatchScrollLocation::RichContent` for AI matches.
- Add a path-agnostic `TerminalFindModel::focused_rich_content_match_id()` helper used by both `get_highlight_ranges_for_find_matches` (highlight rendering) and `AIBlock::handle_find_match_focus_change` (reasoning auto-expand) in place of their previous sync-only lookups.
- AI matches inside one block are walked bottom-to-top for `MostRecentLast` so per-AI-block focus traversal matches sync find.

## Linked Issue

N/A; found while dogfooding.

## Testing

Added three unit tests in `async_find_tests.rs`:

- `test_focused_ai_match_resolves_only_ai_block` — seeds a single AI block with two matches and verifies `focused_ai_match()` resolves correctly under `MostRecentLast` (per-block iteration reversal).
- `test_focused_ai_match_most_recent_first_preserves_storage_order` — same fixture under `MostRecentFirst` to exercise un-reversed iteration.
- `test_focused_match_index_walks_across_terminal_and_ai_blocks` — mixed terminal + AI fixture asserting the unified cache holds at most one variant at a time and the focused index walks across both block types in TotalIndex-descending order.

Existing parity tests (`test_async_focused_order_matches_sync_*`) and all 18 prior async-find tests continue to pass (21 total).

Verified:

- `cargo check -p warp --tests` → clean
- `cargo clippy -p warp --tests -- -D warnings` → clean
- `cargo fmt -p warp` → clean
- `cargo test -p warp --lib async_find` → 21 passed

### Screenshots / Videos

Manually tested; demo here:

https://github.com/user-attachments/assets/1cad4b21-cbb2-4424-8478-2023698d84e5

First half is the existing implementation; notice how matches in AI blocks are highlighted, but not properly focused (aka given _orange_ highlights and scrolled to) when the match index changes. Second half (new implementation) fixes this.

There's a KI (that's around for the sync implementation, too) around match highlights for AI blocks not being dismissed when the find bar _is_. i.e., if I type a find query with matches in AI blocks, then dismiss that find bar with `esc`, matches in AI blocks aren't dismissed until I refocus the pane being searched. Matches in terminal blocks are handled fine. I plan on fixing this in a stacked PR.

## Agent Mode

### DIFF
--- a/app/src/terminal/find/model/async_find.rs
+++ b/app/src/terminal/find/model/async_find.rs
@@ -202,6 +202,27 @@ pub struct AsyncBlockGridMatch {
     pub block_index: BlockIndex,
 }
 
+/// A focused match in a rich content (AI) block.
+///
+/// Mirrors the data carried by `BlockListMatch::RichContent` in the sync path,
+/// so callers can synthesize a `BlockListMatch` from either path.
+#[derive(Debug, Clone)]
+pub struct AsyncFocusedAiMatch {
+    /// The view id of the rich content block.
+    pub view_id: EntityId,
+    /// The id of the focused match within the rich content block.
+    pub match_id: RichContentMatchId,
+    /// The total index of the rich content block in the blocklist sumtree.
+    pub total_index: TotalIndex,
+}
+
+/// Resolution of the global focused match index to either a terminal or an AI match.
+#[derive(Debug, Clone)]
+enum FocusedMatchResolution {
+    Terminal(AsyncBlockGridMatch),
+    Ai(AsyncFocusedAiMatch),
+}
+
 /// Per-block find results, keyed by block index.
 #[derive(Debug, Default)]
 pub(crate) struct BlockFindResults {
@@ -371,9 +392,10 @@ pub struct AsyncFindController {
     /// The FindOptions for the current find run, stored for `active_find_options()` access.
     current_find_options: Option<FindOptions>,
 
-    /// Cached result of `focused_terminal_match()`, updated when focus or matches change.
-    /// Avoids re-sorting HashMap keys and iterating on every call.
-    cached_focused_match: Option<AsyncBlockGridMatch>,
+    /// Cached resolution of `focused_match_index` to either a terminal match,
+    /// an AI match, or `None` (out of range). Updated when focus or matches
+    /// change
+    cached_focused_match: Option<FocusedMatchResolution>,
 
     /// Monotonically increasing generation counter, bumped each time new streams
     /// are spawned. The result stream callback captures the generation at spawn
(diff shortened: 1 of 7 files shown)


app/src/terminal/find/model/async_find.rs
```diff
@@ -204,6 +204,27 @@ pub struct AsyncBlockGridMatch {
...
+/// Mirrors the data carried by `BlockListMatch::RichContent` in the sync path,
+/// so callers can synthesize a `BlockListMatch` from either path.
+#[derive(Debug, Clone)]
+pub struct AsyncFocusedAiMatch {
+    /// The view id of the rich content block.
+    pub view_id: EntityId,
+    /// The id of the focused match within the rich content block.
+    pub match_id: RichContentMatchId,
+    /// The total index of the rich content block in the blocklist sumtree.
+    pub total_index: TotalIndex,
```
vorporeal, line 231: we might want to clarify in the doc comment for the struct what the expected lifetime is, as these two values can get stale as new output is streamed in/blocks get created.

as-is, nothing prevents an `AsyncFocusedAiMatch` from being stored somewhere and getting stale.  if it's intended to be a short-lived/transient state holder, we may want to consider removing `Clone`, so that we can return a reference to it, but nothing outside of this module can ever hold onto one.

app/src/terminal/find/model/async_find.rs
```diff
@@ -484,27 +506,42 @@ impl AsyncFindController {
         self.update_cached_focused_match();
     }
 
-    /// Returns the focused match as an AsyncBlockGridMatch if it's a terminal match.
+    /// Returns the focused match as an `AsyncBlockGridMatch` if it's a terminal match.
```
vorporeal: nit/tip: further wrapping this in brackets (e.g.: "[`" with a pair at the end) will linkify it, if the item within the brackets is a symbol in the current scope.

vorporeal (approved):

vkodithala:
* **#11205** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11205?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11205?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/terminal/find/model/async_find.rs
vkodithala: Learned something new! This is great, thanks and fixed.

app/src/terminal/find/model/async_find.rs
vkodithala: Did some digging, I don't actually think that this is new. `BlocklistMatch` (the corresponding type in the synchronous path) also inherits `Clone`, and stores ephemeral `RichContentMatchId` and `TotalIndex` values directly.

From my understanding, a `RichContentMatchId` is invalidated every time we re-run find on an AI block, whereas `TotalIndex` seems a little more stable (since sumtrees are mostly append-only, aside from things like block removal events).

Added some comments to both the synchronous and asynchronous AI block match types that reference the fact that they're short-lived. Removing `Clone` is a larger change with more deps that I think might be OOS since it exists for the sync implementation too, so I left it as a TODO for now.

PR #11217, opened 2026-05-18
fix(tmux): break out of ReadingCommandOutput on OpenSSH disconnect (#9900)

### LINKED ISSUE #9900: Command wrapper emits truncated ltin echo fragments and stray new-window token after SSH session drops, causing repeated command not found errors in zsh
### Pre-submit Checks

### Describe the bug

After an SSH session times out, Warp's command-instrumentation wrapper continues to inject commands into the now-local zsh prompt. The injected commands are malformed in two ways:

The leading bui of builtin echo … is sliced off, producing the literal token `ltin echo -n "^^^<id>|||"; ….`
The wrapper string contains an unbalanced closing ) and ends with a stray new-window token (apparently the tail of a tmux new-window '<payload>' call whose outer framing was stripped).

zsh tries to execute new-window as a standalone binary and prints zsh: command not found: new-window for every emission. The wrapper keeps firing with monotonically increasing internal IDs (observed eight times within ~50 ms).

### To reproduce

From Warp on macOS, open an SSH session to a warpified remote host.

Trigger Warp's directory-listing instrumentation by using cd dir.

Allow the connection to idle until the network drops it, or simulate it by turning off WiFi until SSH reports:
```
Read from remote host IP: Operation timed out
Connection to IP closed.
client_loop: send disconnect: Broken pipe
```

### Expected behavior

When SSH reports Broken pipe / Operation timed out, Warp should detect the dead transport and stop routing remote-targeted instrumentation into the local shell.

A single, clear notice should be shown to the user instead.

Any command Warp does emit must be syntactically complete and correctly quoted; truncating builtin to ltin and leaving an unmatched ) and a bare new-window token is never valid output.

```
Read from remote host IP: Operation timed out
Connection to IP closed.
client_loop: send disconnect: Broken pipe
ltin echo -n "^^^1777710953501548|||";  cd /Users/user/bench &&  find . -maxdepth 1 -type d -print0 && printf '"'"'%b'"'"' '"'"'\0'"'"' && find . -maxdepth 1 -not -type d -print0             ; builtin echo "|||$?\$\$\$")|command cat; command sleep 1'
zsh: command not found: new-window
ltin echo -n "^^^1777710953501568|||"; …
zsh: command not found: new-window
ltin echo -n "^^^1777710953501571|||"; …
zsh: command not found: new-window
ltin echo -n "^^^1777710953534481|||"; …
zsh: command not found: new-window
ltin echo -n "^^^1777710953544068|||"; …
zsh: command not found: new-window
ltin echo -n "^^^1777710953544191|||"; …
zsh: command not found: new-window
ltin echo -n "^^^1777710953544823|||"; …
zsh: command not found: new-window
ltin echo -n "^^^1777710953545060|||"; …
zsh: command not found: new-window

```

### Operating system (OS)

macOS

26.2

### Shell Version

### Current Warp version

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

maxmilian:
## Summary
Fixes #9900 — after an SSH session times out, Warp keeps injecting tmux `new-window -d '(builtin echo …)|cat; sleep 1'` wrapper bytes into the now-local zsh prompt, producing repeated `zsh: command not found: new-window` plus mangled `ltin echo` fragments.

## Root cause
At the tmux parser layer. After `%begin`, `ParserState::ReadingCommandOutput` accepts ANY byte as command-output content and only escapes when it sees `%end` / `%error`. If the multiplexed SSH transport dies first, the parser stays stuck in that state forever:

- The local OpenSSH client's disconnect diagnostics
- The local zsh re-prompt
- Warp's still-queued wrapper bytes

…are all silently swallowed as "more command output." `tmux_performer.exited` is never set, `ControlModeEvent::Exited` never fires, and `tmux_control_mode_context` is never cleared. Each new completion request keeps queueing more wrapper commands.

Empirical proof in `app/src/terminal/model/tmux/parser_tests.rs` (`test_begin_without_end` was the existing assertion that 0 messages are emitted in this scenario — that's exactly the latent bug).

## Fix
Inside `ReadingCommandOutput`, after each completed content line, sniff for non-localized OpenSSH disconnect diagnostics:

- `client_loop: send disconnect`
- `packet_write_wait`
- `ssh_exchange_identification`
- `kex_exchange_identification`
- `Read from remote host`
- `Connection to ` … ` closed` (both substrings on the same line)

When matched, emit `TmuxMessage::ParseError`. The existing parse-error recovery path (`ansi/mod.rs:471-481`) already (a) sets `tmux_performer.exited = true`, (b) emits `ControlModeEvent::Exited` upstream, and (c) resets the ansi state machine back to default. That reuse keeps the diff small.

Plain `Broken pipe` is deliberately NOT a needle (too generic; many tools print it). The 6 needles above are emitted directly by OpenSSH and unlikely to occur in legitimate tmux command responses.

Uses `memchr::memmem::find` (already a workspace dep).

## Tests
9 new parser tests, all pass (24 total in `parser_tests.rs`):
- Regression test mirroring the byte stream from the issue body
- 1 positive test per supported needle (6 tests)
- 2 negative tests:
  - lone `Broken pipe` does NOT trigger (deliberate exclusion)
  - `Connection to` / `closed` on different lines does NOT trigger

`cargo fmt` + `cargo clippy -p warp --lib --no-deps --tests` clean.

## Scope
`ParserState::ReadingPaneOutput` has the same structural escape hatch problem, but pane output is byte-escaped (`\NNN` octal for bytes < 32), so raw SSH disconnect text from the local PTY is unlikely to land there. Left for a follow-up to keep this PR focused.

## Follow-up worth flagging
Pattern-matching OpenSSH stderr is a band-aid. A timeout-based escape from `ReadingCommandOutput` (or tracking SSH process state) would catch all transport failures, not just OpenSSH's. The current approach was chosen for minimal blast radius; happy to follow up with the more general fix in a separate PR if maintainers want.

### DIFF
--- a/app/src/terminal/model/tmux/parser.rs
+++ b/app/src/terminal/model/tmux/parser.rs
@@ -295,6 +295,25 @@ impl TmuxControlModeParser {
                             output_lines: Err(std::mem::take(lines)),
                         });
                         self.state = ParserState::BeginningOfLine;
+                    } else if looks_like_ssh_disconnect_diagnostic(current_line) {
+                        // Issue #9900: if the SSH transport dies after `%begin`
+                        // but before `%end` arrives, this state would otherwise
+                        // swallow every subsequent byte (local zsh's prompt,
+                        // queued wrapper fragments, etc.) as command-output
+                        // content forever. OpenSSH's standard disconnect
+                        // diagnostics are highly recognizable, so seeing one of
+                        // them inside a `%begin/%end` block is a strong
+                        // signal that the multiplexed transport is gone and
+                        // the rest of this stream is local shell bytes, not
+                        // tmux output. Treat as a parse error so the existing
+                        // recovery path tears down control mode and
+                        // re-renders to the user.
+                        report_parse_error(
+                            handler,
+                            "SSH transport disconnect detected inside %begin/%end",
+                            b'\n',
+                        );
+                        self.state = ParserState::Error;
                     } else {
                         // Command output still ongoing -- append to list of lines.
                         lines.push(std::mem::take(current_line));
@@ -458,6 +477,41 @@ impl Default for TmuxControlModeParser {
 
 fn report_parse_error(handler: &mut impl TmuxControlModeHandler, message: &'static str, byte: u8) {
     handler.tmux_control_mode_message(TmuxMessage::ParseError { message, byte })
+}
+
+/// Returns `true` if `line` looks like one of OpenSSH's standard disconnect
+/// diagnostics. Used inside `ReadingCommandOutput` to detect that the
+/// multiplexed SSH transport has died mid-command, so we can break out of
+/// the begin/end block instead of buffering every subsequent byte forever.
+///
+/// We deliberately stick to needles that are emitted by OpenSSH itself (not
+/// translatable, present across versions):
+/// - `client_loop: send disconnect`     -- OpenSSH client's `client_loop`
+/// - `packet_write_wait`                -- OpenSSH stderr on write-side EPIPE
+/// - `ssh_exchange_identification`      -- handshake failed at version-string exchange
+/// - `kex_exchange_identification`      -- modern variant of the above for key exchange
+/// - `Read from remote host`            -- OpenSSH stderr on read error
+/// - `Connection to ` ... ` closed`     -- OpenSSH's final "Connection to <host> closed." line (both substrings must appear on the same line)
+///
+/// Plain `Broken pipe` is intentionally NOT included because that text is
+/// produced by many unrelated tools and would risk false positives. The
+/// needles above are specific enough that they should not occur in a
+/// legitimate tmux command response.
+fn looks_like_ssh_disconnect_diagnostic(line: &[u8]) -> bool {
+    const NEEDLES: &[&[u8]] = &[
+        b"client_loop: send disconnect",
+        b"packet_write_wait",
+        b"ssh_exchange_identification",
+        b"kex_exchange_identification",
+        b"Read from remote host",
+    ];
+    for needle in NEEDLES {
+        if memchr::memmem::find(line, needle).is_some() {
+            return true;
+        }
+    }
+    memchr::memmem::find(line, b"Connection to ").is_some()
+        && memchr::memmem::find(line, b" closed").is_some()
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


Xytronix:
Great fix for the #9900 mid-command timeout path — the `ReadingCommandOutput` escape hatch is solid and the test coverage is thorough.

I wanted to flag a related but distinct failure mode that this PR does **not** cover. After a **graceful** SSH logout (not a timeout), Warp's control-mode teardown still doesn't fire, and the wrapper keeps dispatching complete, un-garbled `new-window` commands into the now-local shell:

```
logout
logout
Connection to <host> closed.
new-window -d  -e PATH='...'  -PF "background window: #{window_id} #{pane_id}" '(builtin echo -n "^^^1781035783015785|||"; cd /Users/user/bench && find . -maxdepth 1 -type d -print0 && ...)'
zsh: command not found: new-window
new-window -d -c '/Users/user/project' -e PATH='...'  -PF "background window: #{window_id} #{pane_id}" '(builtin echo -n "^^^1781035783015832|||"; git --no-optional-locks branch ...)'
zsh: command not found: new-window
(repeated ~10 times with monotonically increasing IDs)
```

Key differences from #9900:

| | #9900 (fixed by this PR) | This variant |
|---|---|---|
| **Trigger** | SSH timeout mid-command | Graceful `logout` between commands |
| **Parser state at disconnect** | `ReadingCommandOutput` | Likely `BeginningOfLine` (idle between commands) |
| **Symptom** | Garbled `ltin echo` fragments, unbalanced `)` | Complete, correctly-quoted `new-window` commands |
| **Fixed by `looks_like_ssh_disconnect_diagnostic`?** | Yes — runs inside `ReadingCommandOutput` | No — that check doesn't run in other parser states |

The `Connection to <host> closed.` line arrives while the parser is between commands, so the new `looks_like_ssh_disconnect_diagnostic` guard never fires. `ControlModeEvent::Exited` is never emitted, `tmux_control_mode_context` is never cleared, and Warp's command queue keeps draining into the local shell.

A possible follow-up fix could be to also check for SSH disconnect diagnostics in `BeginningOfLine` (or other non-`ReadingCommandOutput` states), or to add a higher-level transport-death detection outside the tmux parser — e.g. watching for the SSH process exit or PTY EOF.

Happy to provide logs or help test a broader fix. Running macOS 27 Developer Beta 1, Warp v0.2026.04.27.15.33.preview_02, zsh 5.9.

MaggieShan:
Thanks for taking a pass on fixing this! The team is actually planning on sunsetting tmux warpification very soon and thus won't be making any further changes to the current logic. Would highly recommend using the [SSH extension](https://docs.warp.dev/terminal/warpify/ssh/#installing-the-ssh-extension) instead and let me know if the issue still persists there.

PR #11219, opened 2026-05-18
Remove custom models options from cloud runs

dagmfactory:
This change removes the ability to select custom models when running cloud runs via cloud run on terminal or via orchestrate.
Custom endpoints remain available for non-cloud agent conversations and local orchestration agents.

https://www.loom.com/share/2e2fa19213b9430fa7fbae763203b5c3


app/src/ai/blocklist/inline_action/orchestration_controls.rs
```diff
@@ -454,10 +454,24 @@ pub fn populate_model_picker_for_harness<A: OrchestrationControlAction, V: View>
...
+                let (auto_models, rest): (Vec<_>, Vec<_>) = all_choices
+                    .into_iter()
+                    .partition(|llm| llm.id.as_str().starts_with("auto"));
+                let (custom_models, other_models): (Vec<_>, Vec<_>) = rest
+                    .into_iter()
+                    .partition(|llm| llm_prefs.custom_llm_info_for_id(&llm.id).is_some());
+                let choices: Vec<_> = auto_models
+                    .into_iter()
+                    .chain(custom_models)
+                    .chain(other_models)
```
danielpeng2: The logic here might be clearer if we do something like this to exclude custom models. I think it's a little confusing to filter out the custom models at the `all_choices` level but still include `custom_models` in the chain.
```
let mut choices: Vec<_> = auto_models.into_iter();
if is_local {
  choices = choices.chain(custom_models);
}
choices = choices.chain(other_models);
```

app/src/ai/blocklist/inline_action/orchestration_controls.rs
```diff
@@ -454,10 +454,24 @@ pub fn populate_model_picker_for_harness<A: OrchestrationControlAction, V: View>
...
-                // Oz / unset: current behavior — Warp LLM catalog.
+                // Oz / unset: Warp LLM catalog. Custom models excluded for
+                // cloud runs (not supported by remote workers).
+                // Order: auto models first, then custom models, then other models.
                 let llm_prefs = LLMPreferences::as_ref(ctx_dropdown);
-                let choices: Vec<_> = llm_prefs
+                let all_choices: Vec<_> = llm_prefs
                     .get_base_llm_choices_for_agent_mode(ctx_dropdown)
+                    .filter(|llm| is_local || llm_prefs.custom_llm_info_for_id(&llm.id).is_none())
+                    .collect();
```
danielpeng2: Do we need to collect into a `Vec` here? That's a memory allocation that I don't think we need here, since the only time we use `all_choices` we immediately transform it back into an iterator using `into_iter()`.

app/src/ai/blocklist/inline_action/orchestration_controls.rs
```diff
@@ -548,6 +562,7 @@ pub fn is_model_in_filtered_choices<V: View>(
             let llm_prefs = LLMPreferences::as_ref(ctx);
             llm_prefs
                 .get_base_llm_choices_for_agent_mode(ctx)
+                .filter(|llm| is_local || llm_prefs.custom_llm_info_for_id(&llm.id).is_none())
```
danielpeng2: Does the logic here rely on having the same model list as `populate_model_picker_for_harness`? We might want to extract the logic into a common method to avoid drift.

I wonder if we should include a `include_custom_model` or `is_local` parameter on `get_base_llm_choices_for_agent_mode` so that callers are forced to make a decision on whether or not they want custom models.

dagmfactory: good point, added a common helper

app/src/ai/blocklist/inline_action/orchestration_controls.rs
dagmfactory: nope, not needed.
removed

app/src/terminal/view/ambient_agent/model_selector.rs
```diff
@@ -468,41 +468,36 @@ impl ModelSelector {
             .clone();
 
         let mut auto_choices = Vec::new();
-        let mut custom_choices = Vec::new();
         let mut other_choices = Vec::new();
         for llm in llm_preferences.get_base_llm_choices_for_agent_mode(ctx) {
```
danielpeng2, line 472: can we use the common `get_base_model_choices` method here as well?

danielpeng2 (approved):

PR #11223, opened 2026-05-18
Forward proxy errors for initialization error telemetry

kevinyang372:
### Motivation

When the remote server proxy exits during SSH initialization, the client reports `ResponseChannelClosed` in the `RemoteServerInitialization` telemetry event — but we have no visibility into *why* the proxy died. We triaged ~100 recent initialization failures and found that:

- 99/100 were `ResponseChannelClosed` with exit_code=1 (proxy crash)
- Sentry events from the daemon on the latest release showed no clear root cause — the dominant errors (SQLite, D-Bus) are non-fatal runtime noise
- The exit_code comes from the SSH/proxy subprocess, not the daemon, meaning the proxy itself is crashing or the SSH transport is dropping before the Initialize response arrives

The proxy already logs to stderr, and `spawn_stderr_forwarder` already pipes those lines into client-side logging — but none of it reaches telemetry. This means we cannot distinguish between different proxy failure modes without asking users to share logs.

### Approach

This PR captures the last 5 lines of the proxy's stderr in a shared tail buffer (`StderrTail`) and attaches them to the `RemoteServerInitialization` telemetry event when a connection failure occurs. This gives us the exact proxy-side error context (e.g. "failed to bind socket", "timed out waiting for daemon socket") directly in the telemetry warehouse.

**Why this over alternatives:**

- **vs. Sentry in the proxy**: The proxy is a thin, short-lived bridge process. A full Sentry SDK would add resource overhead and complexity disproportionate to its role. The proxy already has crash reporting via the daemon's Sentry (for errors that happen after Sentry init), but pre-Sentry crashes and proxy-specific failures are invisible.
- **vs. structured exit codes**: Would require maintaining a mapping between codes and error types across the proxy and client, becoming a maintenance burden as error paths evolve. The raw stderr string is self-describing and stays in sync automatically.
- **vs. crash file on remote host**: Writing a last-error file and having the daemon relay it adds file I/O and cross-process coordination for minimal benefit over the stderr approach.

The stderr tail buffer is entirely client-side — no new processes, no Sentry SDK, no network calls from the remote host. Truncation preserves the *end* of stderr (most diagnostic value) and caps at 2048 chars to keep payloads reasonable.

### Changes

- `StderrTail` type in `client/mod.rs`: shared `Arc<Mutex<VecDeque<String>>>` buffer (5 lines, 2048 char cap)
- `spawn_stderr_forwarder` now returns the tail buffer alongside existing log forwarding
- `Connection` struct carries the buffer through `SshTransport::connect`
- `RemoteSessionState::Initializing` stores the buffer; drained on failure
- `SessionConnectionFailed` event gains `proxy_stderr: Option<String>`
- `RemoteServerInitialization` telemetry event gains `proxy_stderr` in its JSON payload

## Linked Issue

## Testing

- Deployed a test build with fake proxy error logs to a remote host via `./script/deploy_remote_server`
- Verified the stderr lines appear in client-side logs via `spawn_stderr_forwarder`
- Verified the `proxy_stderr` field is populated in the `RemoteServerInitialization` telemetry event on connection failure

## Agent Mode

CHANGELOG-NONE


kevinyang372:
* **#11223** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11223?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11223?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/server/telemetry/events.rs
```diff
@@ -4213,13 +4216,15 @@ impl TelemetryEvent {
...
                 signal_killed,
+                proxy_stderr,
             } => Some(json!({
                 "phase": phase,
                 "error": error,
                 "remote_os": remote_os,
                 "remote_arch": remote_arch,
                 "exit_code": exit_code,
                 "signal_killed": signal_killed,
+                "proxy_stderr": proxy_stderr,
```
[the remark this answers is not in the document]
kevinyang372: We are not passing sensitive infos to the remote server proxy. I don't think there is anything here that would expose privacy issues

crates/remote_server/src/client/mod.rs
```diff
@@ -35,6 +36,55 @@ use warpui::r#async::TransportStream;
...
 const REQUEST_TIMEOUT: Duration = Duration::from_secs(120);
 
+/// Maximum number of stderr lines to retain in the tail buffer.
+const STDERR_TAIL_MAX_LINES: usize = 5;
+
+/// A shared buffer that retains the last [`STDERR_TAIL_MAX_LINES`] lines
+/// from the remote server proxy's stderr. Used to attach proxy-side
+/// context to telemetry when the connection fails.
+#[derive(Clone, Debug)]
+pub struct StderrTail(Arc<Mutex<VecDeque<String>>>);
```
alokedesai: Just use a `BoundedVecDeque` here

crates/remote_server/src/client/mod.rs
```diff
@@ -35,6 +36,55 @@ use warpui::r#async::TransportStream;
...
+        let joined = lines.join("\n");
+        if joined.len() > 2048 {
+            // Truncate from the start — the tail end of stderr is the
+            // most useful context for diagnosing why the proxy died.
+            let mut start = joined.len() - 2048;
+            // Snap forward to a char boundary to avoid panicking on
+            // multibyte UTF-8.
+            while !joined.is_char_boundary(start) {
+                start += 1;
+            }
```
alokedesai: not blocking but honeslty it's going to be simpler to just take a char iterator, reverse it, take the first 2048 chars, and reverse it again

crates/remote_server/src/client/mod.rs
```diff
@@ -35,6 +36,55 @@ use warpui::r#async::TransportStream;
...
+/// Maximum number of stderr lines to retain in the tail buffer.
+const STDERR_TAIL_MAX_LINES: usize = 5;
+
+/// A shared buffer that retains the last [`STDERR_TAIL_MAX_LINES`] lines
+/// from the remote server proxy's stderr. Used to attach proxy-side
+/// context to telemetry when the connection fails.
+#[derive(Clone, Debug)]
+pub struct StderrTail(Arc<Mutex<VecDeque<String>>>);
+
+impl StderrTail {
```
alokedesai: nit, not blocking and optional: Would recommend putting this in a separate module (the more focused this mod is with actual implementation details, the better)

crates/remote_server/src/client/mod.rs
alokedesai: not blocking, nit: Thoughts on just naming this `RemoteServerLog` or something more specific than stderrtail? The fact that it's using stderr is an implementation detail of how we populate imo.

alokedesai (approved):

PR #11240, opened 2026-05-18
Custom model TOS hyperlink coloring

dagmfactory:
### Issue
The warp terms of service hyperlink on custom inference settings was not readable because of coloring

### Before

Screenshot 2026-05-18 at 6 27 11 PM

Screenshot 2026-05-18 at 6 27 39 PM

### After

Screenshot 2026-05-18 at 6 27 59 PM

Screenshot 2026-05-18 at 6 28 47 PM

### DIFF
--- a/app/src/settings_view/ai_page.rs
+++ b/app/src/settings_view/ai_page.rs
@@ -53,6 +53,7 @@ use strum::IntoEnumIterator;
 use warp_core::channel::ChannelState;
 use warp_core::context_flag::ContextFlag;
 use warp_core::features::FeatureFlag;
+use warp_core::ui::color::blend::Blend;
 use warp_core::ui::theme::color::internal_colors;
 use warpui::elements::{
     Border, ChildAnchor, ChildView, ConstrainedBox, CornerRadius, CrossAxisAlignment, Dismiss,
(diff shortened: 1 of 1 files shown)


app/src/settings_view/ai_page.rs
```diff
@@ -7106,7 +7107,13 @@ impl ApiKeysWidget {
                                 appearance.theme().background().into_solid(),
                                 self.custom_inference_terms_index.clone(),
                             )
-                            .with_hyperlink_font_color(appearance.theme().accent().into_solid())
+                            .with_hyperlink_font_color(
+                                appearance
+                                    .theme()
+                                    .accent()
+                                    .blend(&warp_core::ui::theme::Fill::black().with_opacity(50))
+                                    .into_solid(),
```
danielpeng2: In general we try to avoid going rogue and specifying our own colours for UI in the app. The designs use an accent colour and the rest of our tooltips use an accent colour, so it'd be good to do the same here.

I'd try following the same code we use for font colour in other tooltips, to see if that makes a difference: https://github.com/warpdotdev/warp/blob/bfb4e980318bb85e21e7dcf7aa43ca101df991a0/crates/warp_core/src/ui/builder.rs#L613-L620. In this example, we use an `on_background` field which should adjust the colour to account for contrast.

danielpeng2 (approved):

PR #11286, opened 2026-05-19
CLI commands for CRUD on agents

bnavetta:
## Description

This adds a set of CLI commands for operating on named agents:
* `oz agent create [options]` to define a new agent
* `oz agent list` and `oz agent get` to query agents
* `oz agent update` to patch agent configs (setting a base model, adding/removing skills, etc.)
* `oz agent delete` to delete an agent

## Testing

- [x] Added CLI validation tests

### Screenshots / Videos

https://www.loom.com/share/0f7f3f142ebe46f79582c22a6027b34e

## Agent Mode

CHANGELOG-IMPROVEMENT: The Oz CLI now has commands for operating over reusable agents

### DIFF
--- a/crates/warp_cli/src/agent.rs
+++ b/crates/warp_cli/src/agent.rs
@@ -4,8 +4,10 @@ use std::path::PathBuf;
 use clap::{Args, Subcommand, ValueEnum};
 use serde::{Deserialize, Serialize};
 
+use crate::SortOrderArg;
 use crate::config_file::ConfigFileArgs;
 use crate::environment::EnvironmentCreateArgs;
+use crate::json_filter::JsonOutput;
 use crate::mcp::MCPSpec;
 use crate::model::ModelArgs;
 use crate::scope::ObjectScope;
@@ -283,8 +285,8 @@ pub struct RunAgentArgs {
     ///
     /// When used with --prompt, the skill provides the base context and the prompt is the task.
     ///
-    /// To automate a skill on a schedule, use `oz schedule create --skill <SPEC>`.
-    #[arg(long = "skill", value_name = "SPEC")]
+    /// To automate a skill on a schedule, use `oz schedule create --skill <SKILL>`.
+    #[arg(long = "skill", value_name = "SKILL")]
     pub skill: Option<SkillSpec>,
 
     /// Name for this agent task.
@@ -436,8 +438,8 @@ pub struct RunCloudArgs {
     ///
     /// When used with --prompt, the skill provides the base context and the prompt is the task.
     ///
-    /// To automate a skill on a schedule, use `oz schedule create --skill <SPEC>`.
-    #[arg(long = "skill", value_name = "SPEC")]
+    /// To automate a skill on a schedule, use `oz schedule create --skill <SKILL>`.
+    #[arg(long = "skill", value_name = "SKILL")]
     pub skill: Option<SkillSpec>,
 
     /// Name for this agent task.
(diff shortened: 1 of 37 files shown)


specs/REMOTE-1696/PRODUCT.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Summary
+Add first-class Oz CLI commands for managing named agents. Developers can list, inspect, create, update, and delete named agents from the terminal with output modes that match existing Oz run commands.
+
+## Problem
+Named agents are manageable through the public API and related Oz surfaces, but the Oz CLI does not expose the full CRUD surface. The existing `oz agent list` command lists repository-discovered agent skills, so the CLI needs a clear split between named agents and skills before adding CRUD commands.
+
+## Goals
+- Make `oz agent list` mean "list named agents".
+- Preserve skill discovery by renaming the existing command to `oz agent skills`.
```
captainsafia, line 11: I dunno if we have established guidance around breaking changes for CLI commands given usage but do we want to do something more mindful here around ensuring that `oz agent list` retains the skills-focused behavior for existing users?

specs/REMOTE-1696/PRODUCT.md
```diff
@@ -0,0 +1,45 @@
...
+- Changing `oz agent run` or `oz agent run-cloud` semantics beyond continuing to accept an agent UID where already supported.
+
+## Behavior
+1. `oz agent skills` lists available agent skills using the current `oz agent list` behavior, including the existing `--repo` option and repository authorization flow.
+2. `oz agent list` lists named agents accessible to the authenticated user's team.
+3. `oz agent list` supports `--sort-by name` and `--sort-by created-at`.
+4. `oz agent list` supports `--sort-order asc` and `--sort-order desc`. If omitted, name sorting defaults to ascending, and timestamp sorting defaults to descending.
+5. Sorting is performed client-side after the API response is fetched and before rendering pretty, text, or ndjson output.
+6. `oz agent get <uid>` retrieves a single named agent by UID.
+7. `oz agent create` creates a named agent. `--name` is required. Optional fields are accepted for the fields supported by the API: description, secrets, skills, base model, and default environment.
```
captainsafia, line 29: Totally optional: I wonder if it would be good for programmability if these create/update commands also accepted a file (or JSON payload via stdin) as the input?

crates/warp_cli/src/agent.rs
```diff
@@ -502,9 +513,178 @@ pub struct RunCloudArgs {
...
+    /// Description of the agent.
+    #[arg(long = "description")]
+    pub description: Option<String>,
+
+    /// Attach a secret to the agent. Repeat the flag for multiple secrets.
+    #[arg(long = "secret", value_name = "NAME")]
+    pub secrets: Vec<String>,
+
+    /// Attach a skill to the agent. Repeat the flag for multiple skills.
+    #[arg(long = "skill", value_name = "SPEC")]
```
captainsafia: Is thee a reason we refer to it as SPEC and not SKILL here or elsewhere? I can see this being potentially confusing since it surfaces in help docs.

captainsafia (approved):

crates/warp_cli/src/agent.rs
bnavetta: Force of habit from places we refer to skill specs - will update!

specs/REMOTE-1696/PRODUCT.md
bnavetta: We don't really have existing guidance, no - this is the first breaking change I can think of

The `oz agent list` command does print out a message that if you're looking for skills, they're now in `oz agent skills`! It won't help with programmatic usage, but the skill-based version of this command wasn't really set up for scripting anyways.

specs/REMOTE-1696/PRODUCT.md
bnavetta: Yes, completely agree! I'm hesitant to do it in this PR though, since I think we'll need to be really intentional about what the file format is. It'd also be useful to think through how this could work for combined definitions of an agent and skills/scripts it should have access to.

PR #11295, opened 2026-05-19
Enable cmd-O and @ context on remote SSH session

kevinyang372:
## Summary
Enable **Cmd-O file search palette** and **@ context menu** file/code categories in remote SSH sessions.

Both features were broken in SSH because the client-side gating assumed local-only paths. The underlying repo metadata layer already supported remote repos — this PR threads `LocalOrRemotePath` through the remaining client-side consumers.

### Core: Workspace `ActiveSession` stores `LocalOrRemotePath`
- `WindowActiveSession.path_if_local: Option<PathBuf>` → `working_directory: Option<LocalOrRemotePath>`
- New `working_directory()` accessor returns `Option<&LocalOrRemotePath>` (local or remote)
- `path_if_local()` remains as a derived convenience for callers that only need local paths
- `update_active_session` now calls `terminal.pwd_as_local_or_remote()` to populate it

### `FileSearchModel` — remote repo contents + cache
- New `repo_root_location()` method uses `ActiveSession::working_directory()` → `DetectedRepositories::get_root_for_path()` (handles both local/remote)
- `get_contents_from_repo()` refactored to accept `&LocalOrRemotePath` and handle both local and remote repos
- Cache key changed from `HashMap<PathBuf, ...>` to `HashMap<LocalOrRemotePath, ...>`
- Cache invalidation handles `RepositoryIdentifier::Remote` via new `to_local_or_remote_path()` helper

### `RepositoryIdentifier` — new conversion helper
- Added `to_local_or_remote_path()` method for clean `RepositoryIdentifier` → `LocalOrRemotePath` conversion

### `OpenedFilesModel` — remote path support
- Repo root key: `HashMap<PathBuf, ...>` → `HashMap<LocalOrRemotePath, ...>`
- Inner file map: `HashMap<PathBuf, Instant>` → `HashMap<String, Instant>` (repo-relative paths as strings)
- `file_opened()` uses `strip_repo_prefix()` to compute relative paths (works for both local/remote)
- `code_pane.rs` `FileOpened` handler now tracks files for all repo types, not just local

### UI gating updates
- **Command palette** (`data_sources.rs`): uses `repo_root_location()` to detect remote repos
- **@ context menu** (`view.rs`): `is_active_dir_in_git_repo` uses `ActiveSession::working_directory()` directly
- **@ context menu file data source**: `snapshot_last_opened` uses `working_directory()` → `get_root_for_path()`

## Plan
[Enable File Search Palette (Cmd-O) and @ Context Menu in Remote SSH](https://staging.warp.dev/drive/notebook/A5Jg3SkafG1RO5zKQCWtCF)

## Conversation
https://staging.warp.dev/conversation/700915cd-6778-4a7e-9686-3a25eeb476ce


kevinyang372:
* **#11295** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11295?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11295?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/search/files/model.rs
```diff
@@ -198,71 +194,91 @@ impl FileSearchModel {
...
-                                        .to_string(),
-                                    is_directory: false,
-                                })
-                            } else {
-                                None
-                            }
+                            let relative_path =
+                                file_local.strip_prefix(&canonical_repo_path).ok()?;
+                            Some(FileSearchResult {
+                                path: relative_path.to_string_lossy().to_string(),
```
moirahuang, line 270: should we store the host id as well?

app/src/terminal/view.rs
```diff
@@ -11766,6 +11766,31 @@ impl TerminalView {
                                                 );
                                             },
                                         );
+
+                                        // Force the workspace to re-evaluate
+                                        // ActiveSession::working_directory. The
+                                        // AppStateChanged at line 11671 already ran
+                                        // update_active_session once, but for remote
```
moirahuang, line 11773: should we make this the definitive place for us to evaluate `ActiveSession::working_directory`?

moirahuang (approved):
1. called this out over slack but we should fix the at context menu outside of repos, this can be in a follow-up PR

https://github.com/user-attachments/assets/44ce60ad-29d3-4d56-84c9-8ce34b248e0e

2. also in the above video, i'm seeing some flicker from "Loading" to "No results" which might be unavoidable but maybe we can debounce or skip "Loading" if we quickly are able to know we have no results

3. not related to this PR but we're missing path on hover of the tooltip when we open a remote code file

Image

app/src/search/files/model.rs
kevinyang372: Erm why do we need the host ID here?

app/src/terminal/view.rs
kevinyang372: There are multiple sources for ActiveSession's working directory to change. I think semantically what we want to communicate here is remote repo path updating is a valid state change which should cause us to re-evaluate active session data

kevinyang372:
I addressed (1)

I will look into (2)(3) as a followup

Issue #11262, opened 2026-05-19
SECRETS_REGEX silently empty at startup despite persisted custom_secret_regex_list — local MCP-save redaction check never fires


david-engelmann:
## Summary

The custom secret regex set (`SECRETS_REGEX`) read by `find_secrets_in_text` can be empty at runtime even when:

- `~/.warp-oss/settings.toml` has `[privacy.secret_redaction] enabled = true`
- The same TOML has `custom_secret_regex_list = [...20 default patterns...]`
- The Settings → Privacy UI correctly shows the toggle ON and lists all 20 default patterns

As a result, the local redaction check that PR #10839 fixes (`should_block_save_for_secrets(safe_mode, enterprise, contains_secrets)`) returns `contains_secrets=false` regardless of payload, because `find_secrets_in_text` runs against an empty `SECRETS_REGEX`. Saves go through, cloud sync receives the secret, and the *server* rejects with `Secrets detected while creating generic string object` — the cloud catches what the local check should have caught.

## Why it happens

`app/src/terminal/secret_regex_updater.rs:14` carries the comment `Initialize with current custom regexes (will be empty until safe mode is enabled)`. The intended flow:

1. App starts with `safe_mode=false`, `user_secret_regex_list=[]`
2. User toggles safe mode ON
3. `initialize_default_regexes_once` pushes the 20 defaults into `user_secret_regex_list`
4. That mutation fires `PrivacySettingsChangedEvent::CustomSecretRegexList`
5. The updater subscribes to that event (`secret_regex_updater.rs:18-22`) and recompiles `SECRETS_REGEX`

The flow assumes a transition from empty list → populated list. It does NOT handle the case where TOML already has the defaults persisted from a prior session: `PrivacySettings::new` loads the list, but no `CustomSecretRegexList` change event fires, so `SECRETS_REGEX` remains the initial empty `Regex::new_many(&[])` from `app/src/terminal/model/secrets.rs:39-42`.

The updater's startup call to `update_custom_secret_regex_list` at `secret_regex_updater.rs:15` either runs before `PrivacySettings` exposes the populated list, or runs after but is racing some other init step — needs confirmation during the fix.

The empty-regex condition also persists across a user toggling safe mode OFF→ON at runtime, because the updater only subscribes to `CustomSecretRegexList` changes, not `SafeModeEnabled` changes.

## Repro (verified 2026-05-19)

1. Quit warp-oss
2. Ensure `~/.warp-oss/settings.toml` has:
[...]

PR #11297, opened 2026-05-19
Fix MCP +Add path bypassing secret redaction check

### LINKED ISSUE #11265: MCP server save: '+ Add' (new server) path bypasses redaction predicate entirely
## Summary

The "+ Add" (new MCP server) save path bypasses `parse_templatable_json` and therefore never reaches `detect_secrets_in_templatable_mcp_server`. PR #10839's `should_block_save_for_secrets` predicate is correct, but it is only wired into the **edit-existing-server** path, not the **add-new-server** path.

Net effect: secrets in a freshly added MCP server config are never blocked locally, regardless of toggle state. The cloud-side check (`warp::server::sync_queue` rejecting with `Secrets detected while creating generic string object`) is the only thing protecting the wire today — and the user sees a generic `Failed to create mcp server` toast instead of the actionable `This MCP server contains secrets. Visit Settings > Privacy ...` toast.

## Why it happens

In `app/src/settings_view/mcp_servers/edit_page.rs`, the `MCPServersEditPageViewAction::Save` handler (around line 892) branches on `server_card_item_id`:

- `Some(TemplatableMCP)` / `Some(TemplatableMCPInstallation)` → `handle_save_templatable_mcp_server` → `build_templatable_mcp_server_result_from_json` → `parse_templatable_json` → `detect_secrets_in_templatable_mcp_server` ✅
- `None` (new server via "+ Add") → inline block at lines 917-982 that calls `ParsedTemplatableMCPServerResult::from_user_json` directly, skipping every redaction-related step ❌

```rust
None => {
    // This is a new MCP server, we should treat it like a legacy MCP server
    let json = self.json_editor.as_ref(ctx).text(ctx).into_string();

    let parsed_servers =
        match ParsedTemplatableMCPServerResult::from_user_json(&json) {
            // ... no redaction check anywhere here ...
        };
    // ... directly creates the server ...
}
```

## Repro (verified 2026-05-19)

With #11262 either fixed or worked around (toggle OFF→ON to trigger `SafeModeSettings` event → `SECRETS_REGEX` recompile):

1. Settings → Privacy → confirm Secret redaction is ON
2. Settings → MCP Servers → + Add (fresh dialog; `server_card_item_id` is `None`)
3. Paste:
   ```json
   {
     "demo-new-server-bypass": {
       "command": "/usr/bin/true",
       "args": ["--api-key=sk-FAKE0000000000FAKEdemoWARP8761"]
     }
   }
   ```
   (secret in `args` — survives templatization, so we know the regex *would* match it if the path were exercised)
4. Click Save.

**Expected:** "This MCP server contains secrets. Visit Settings > Privacy ..." toast; save blocked.

**Actual:** local save succeeds; the cloud sync rejects with a generic `Failed to create mcp server` toast. Confirmed via `~/Library/Logs/warp-oss.log`:
```
[WARN] [warp::server::sync_queue] Failed to create GenericStringObject(Json(TemplatableMCPServer)) ... Secrets detected
```

Editing an *existing* server (the `Some(TemplatableMCP)` branch) does fire the local block correctly — only the `None` branch is leaky.

## Proposed fix (small)

Either:

**(a)** Route the new-server save through `parse_templatable_json` (the same function the edit path uses) so the existing redaction check applies uniformly.

**(b)** Inline the predicate check at the top of the `None` branch:

```rust
None => {
    let json = self.json_editor.as_ref(ctx).text(ctx).into_string();

    let safe_mode_enabled = *SafeModeSettings::as_ref(ctx).safe_mode_enabled.value();
    let enterprise_enforced =
        UserWorkspaces::as_ref(ctx).is_enterprise_secret_redaction_enabled();
    let contains_secrets = !find_secrets_in_text(&json).is_empty();
    if should_block_save_for_secrets(safe_mode_enabled, enterprise_enforced, contains_secrets) {
        ToastStack::handle(ctx).update(ctx, |toast_stack, ctx| {
            toast_stack.add_ephemeral_toast(
                DismissibleToast::error(
                    "This MCP server contains secrets. Visit Settings > Privacy to modify your secret redaction settings.".to_string(),
                ),
                ctx.window_id(),
                ctx,
            );
        });
        return;
    }

    // ... existing parse + create logic ...
}
```

(a) is structurally cleaner. (b) is the local minimum.

## Related

- Companion to #11262 (SECRETS_REGEX-empty-at-startup) and #11263 (env/headers templatized away before scan). All three were discovered while producing evidence for #10839 (the underlying `should_block_save_for_secrets` predicate). The predicate is correct; this issue is about the predicate not being called on every save path.

SagarSDagdu:
## Description

The "+ Add" (new MCP server) save path in `app/src/settings_view/mcp_servers/edit_page.rs` was calling `ParsedTemplatableMCPServerResult::from_user_json` directly and skipping the `detect_secrets_in_templatable_mcp_server` check that the edit-existing path runs via `parse_templatable_json`.

Net effect today: secrets pasted into a freshly added MCP server config are not blocked locally — the cloud-side `sync_queue` reject is the only guard, and the user sees a generic `Failed to create mcp server` toast instead of the actionable `This MCP server contains secrets. Visit Settings > Privacy ...` toast.

This PR reuses the existing per-server helper after parsing so both save paths get the same local guard.

## Linked Issue

Fixes #11265.

## Screenshots / Videos

Video showing error toast after the fix:

https://github.com/user-attachments/assets/4dc9ba93-9ebd-49c8-b959-1c5f60d1b8de

## Testing

Manual repro per #11265:

1. Settings → Privacy → confirm Secret redaction is ON and the default recommended regexes are loaded (click "Add all recommended" if the list is empty — see #11262)
2. Settings → MCP Servers → + Add
3. Paste:
   ```json
   {
     "demo-new-server-bypass": {
       "command": "/usr/bin/true",
       "args": ["--api-key=sk-FAKE0000000000FAKEdemoWARP8761"]
     }
   }
   ```
4. Click Save → expect the actionable redaction toast, save blocked.

No automated test added; the secret-detection helper is already exercised via the edit-existing path and the change is a 12-line reuse of that helper. Happy to add an integration test on request.

## Agent Mode

### DIFF
--- a/app/src/settings_view/mcp_servers/edit_page.rs
+++ b/app/src/settings_view/mcp_servers/edit_page.rs
@@ -903,6 +903,18 @@ impl TypedActionView for MCPServersEditPageView {
                         return;
                     }
 
+                    for parsed_server in &parsed_servers {
+                        if self
+                            .detect_secrets_in_templatable_mcp_server(
+                                ctx,
+                                &parsed_server.templatable_mcp_server,
+                            )
+                            .is_err()
+                        {
+                            return;
+                        }
+                    }
+
                     for parsed_server in parsed_servers {
                         TemplatableMCPServerManager::handle(ctx).update(
                             ctx,
--- a/app/src/settings_view/mcp_servers/edit_page.rs
+++ b/app/src/settings_view/mcp_servers/edit_page.rs
@@ -903,16 +903,17 @@ impl TypedActionView for MCPServersEditPageView {
                         return;
                     }
 
-                    for parsed_server in &parsed_servers {
-                        if self
-                            .detect_secrets_in_templatable_mcp_server(
+                    if parsed_servers
+                        .iter()
+                        .try_for_each(|parsed_server| {
+                            self.detect_secrets_in_templatable_mcp_server(
                                 ctx,
                                 &parsed_server.templatable_mcp_server,
                             )
-                            .is_err()
-                        {
-                            return;
-                        }
+                        })
+                        .is_err()
+                    {
+                        return;
                     }
 
                     for parsed_server in parsed_servers {


SagarSDagdu:
/oz-review I added video of the fix.

SagarSDagdu:
@peicodes can you take a look at this? A very small but important change in the MCP addition.

vkodithala:
Hey @SagarSDagdu, I'm taking over here. Will take a look later today!

app/src/settings_view/mcp_servers/edit_page.rs
```diff
@@ -903,6 +903,18 @@ impl TypedActionView for MCPServersEditPageView {
...
+                            .detect_secrets_in_templatable_mcp_server(
+                                ctx,
+                                &parsed_server.templatable_mcp_server,
+                            )
+                            .is_err()
+                        {
+                            return;
+                        }
+                    }
+
```
vkodithala: nit: Can we use `try_for_each` here? Something like:
```
if parsed_servers.iter().try_for_each(|s| self.detect...)
```

vkodithala (approved):
This is a great, well-scoped PR. Great job!

Added a small comment that's non-blocking. I like that you didn't make an extra test here. Approved, will merge once CI passes.

app/src/settings_view/mcp_servers/edit_page.rs
SagarSDagdu: Done in 581f6fda, reads cleaner. Thanks!

PR #10944, opened 2026-05-14
fix: hide handoff details panel by default

captainsafia:
## Description
Hide the conversation details panel by default when a local-to-cloud handoff joins the cloud session share. Normal shared ambient agent session joins still auto-open the details panel.

## Linked Issue
- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

Slack request: D0AU96L8CDV / 1778785222.326509

## Testing
- `CARGO_BUILD_JOBS=1 cargo nextest run --manifest-path /workspace/warp/Cargo.toml -p warp -E 'test(test_ambient_session_join_auto_opens_details_panel) | test(test_local_to_cloud_handoff_session_join_keeps_details_panel_hidden)' -j 1`
- `cargo fmt --manifest-path /workspace/warp/Cargo.toml --all -- --check`

- [ ] I have manually tested my changes locally with `./script/run`

### Screenshots / Videos
Demo: https://www.loom.com/share/c555f624fcea4937b3611144dd5ee8a5

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: Hide the conversation details panel by default when moving a local Warp Agent conversation to cloud.

Co-Authored-By: Oz <oz-agent@warp.dev>

_Conversation: https://staging.warp.dev/conversation/8f301dfc-0625-4230-b7aa-b49c362c78f9_
_Run: https://oz.staging.warp.dev/runs/019e27dc-84f5-79ff-8778-e89699d2ef96_
_This PR was generated with [Oz](https://warp.dev/oz)._

PR #11353, opened 2026-05-19
fix redirects going haywire

dmichelin:
A bug was noted that the conversation link was not being preserved when going to conversations. This change prevents faulty history changes to `/app` when the URL is a shared conversation link.

https://linear.app/warpdotdev/issue/REMOTE-1732/conversation-link-rewrite-strips-slug-and-routes-to-app

Loom: https://www.loom.com/share/cb2e52bf9316494490b1c1b16f9265c6

https://github.com/warpdotdev/warp/pull/11142/changes and https://github.com/warpdotdev/warp/pull/10944
introduced issues into the WASM build of warp where:
1. the url redirects to app
2. the sidebar auto hides in web

The URL rewrite comes from this interaction:

•  TerminalPane::shareable_link() only preserves /conversation/:id if the terminal model is still marked as a conversation transcript viewer.
•  But PR #11142 intentionally diverts ambient-task-backed conversation deep links away from the transcript-viewer path into restored cloud-mode.
•  For owned restored cloud-mode panes, load_data_into_restored_ambient_cloud_mode_view sets SharedSessionStatus::NotShared.
•  TerminalPane::shareable_link() maps SharedSessionStatus::NotShared to ShareableLink::Base.
•  PaneGroup::update_browser_url() maps ShareableLink::Base to update_browser_url(None, false).
•  browser_url_handler.rs:get_base_app_url() turns the current URL into /app.

So the bug is that #11142 made deep-linked ambient/non-quick conversations restore as cloud-mode continuation panes. Those panes are not transcript viewers, and their initial/owned restored status can be NotShared, so dynamic browser URL sync treats them as “base app” panes and replaces the deep link with /app.

### DIFF
--- a/app/src/uri/browser_url_handler.rs
+++ b/app/src/uri/browser_url_handler.rs
@@ -1,5 +1,7 @@
 use url::Url;
 
+use super::web_intent_parser::WebIntent;
+
 const DEFAULT_TITLE: &str = "Warp";
 const BASE_APP_PATH: &str = "/app";
 
@@ -54,6 +56,9 @@ pub fn parse_current_url() -> Option<Url> {
 
 fn get_base_app_url() -> Option<Url> {
     if let Some(current_url) = parse_current_url() {
+        if is_conversation_url(&current_url) {
+            return Some(current_url);
+        }
         let mut new_url = current_url.clone();
         new_url.set_path(BASE_APP_PATH);
         new_url.set_query(None);
@@ -61,4 +66,11 @@ fn get_base_app_url() -> Option<Url> {
     }
     log::error!("Failed to get the base url");
     None
+}
+
+fn is_conversation_url(url: &Url) -> bool {
+    matches!(
+        WebIntent::try_from_url(url),
+        Ok(WebIntent::ConversationView(_))
+    )
 }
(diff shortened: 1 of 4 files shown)


jasonkeung:
could you add a quick loom of the fix please?

dmichelin:
> could you add a quick loom of the fix please?

just added!

jasonkeung (approved):

app/src/uri/browser_url_handler.rs
```diff
@@ -54,6 +56,9 @@ pub fn parse_current_url() -> Option<Url> {
 
 fn get_base_app_url() -> Option<Url> {
     if let Some(current_url) = parse_current_url() {
+        if is_conversation_url(&current_url) {
```
jasonkeung: curious why the default is to reroute to app, and the special case is to preserve the current url (as opposed to reroute to app if the url is malformed / whatever the original fix was supposed to handle)

dmichelin: Allegedly https://docs.google.com/document/d/1tRatMnKrXjSKf4Ai9KSXUuaeHaHeTqEqtor8--iSTP8/edit?tab=t.0#heading=h.dxa3021c8p6n

jasonkeung: hmmm okay sounds good to me

Legoben (approved):

PR #11358, opened 2026-05-19
Bedrock UI Model Picker Updates

JaidenRatti:
## Description

Adds AWS Bedrock specific indicators to the model selector UI. Previously, if a user enabled AWS Bedrock for inference, no icon appeared next to the applicable models in the model picker. However, if a user provides an API key, we display a key next to the model (to indicate that the model is going through some other inference). This PR adds AWS imaging for Bedrock supported models when enabled. 

Changes include:
•  Add a bundled AWS SVG icon and expose it through Icon::Aws.
•  Add Bedrock credential detection for models with enabled AwsBedrock host config. We prioritize Bedrock > API Key (for routing and visually).
•  Update cost/details rows to show Billed to Bedrock when applicable.
•  Route the Manage action to Bedrock settings for Bedrock-backed models.

## Linked Issue

[Linear issue](https://linear.app/warpdotdev/issue/REV-1489/add-client-ui-for-this-model-is-backed-by-bedrock)

## Testing

### Screenshots / Videos

~[Loom](https://www.loom.com/share/e68b54da30af49e1a84f307939210a18). Note the cloud svg has been changed to an AWS svg, all other functionality is the same.~

~[Updated Loom](https://www.loom.com/share/06248f2078d64b55b378b60ee73b4a84)~

Updated version: Replace the model provider icon with aws logo where applicable. Keep the key icon for api the same as before.

## Agent Mode


app/src/terminal/input/models/data_source.rs
```diff
@@ -263,6 +265,7 @@ impl ModelSearchItem {
         let is_custom_endpoint = LLMPreferences::as_ref(app)
             .custom_llm_info_for_id(&llm.id)
             .is_some();
+        let is_using_bedrock = !is_auto(llm) && should_show_bedrock_icon_for_model(llm, app);
```
IsaiahWitzke: can you do a code dive and double check that `!is_auto` is correct here?

i seem to remember there was some switch we made on the server side recently where we allow you to give aws credentials for auto models...

app/src/ai/llms.rs
```diff
@@ -41,6 +41,16 @@ pub fn is_using_api_key_for_provider(provider: &LLMProvider, app: &AppContext) -
     }
 }
 
+/// Returns `true` if the model is configured to route through AWS Bedrock and
+/// the current workspace/user settings enable sending AWS Bedrock credentials.
```
IsaiahWitzke: nit: i think AM likes to make a bunch of comments, but if the code is pretty obvious (like here) then you can delete. Comments can get stale... source code does not (by defn?)

app/src/terminal/input/models/data_source.rs
```diff
@@ -478,15 +494,19 @@ impl SearchItem for ModelSearchItem {
...
                     });
                 })
                 .finish();
-
-            CostRow::BilledToApi {
+            CostRow::BilledToProvider {
+                label: if self.is_using_bedrock {
+                    "Billed to Bedrock"
+                } else {
+                    "Billed to API"
```
IsaiahWitzke: do an iteration with chatgpt on Copy here, but technically, this isnt true anymore, and this is a good op to clean it up.

In the enterprise / business case we will be billing you some $s for agent hours... maybe something like "Inference charged to Bedrock" or somethiing? (idk i dont love that, but i dont love lying to people saying that "all costs here are going to your api key" when in reality, you're gonna get a small amount of platform credits charged to your warp credits)

JaidenRatti: that's a good point, will workshop it.

app/src/terminal/input/models/data_source.rs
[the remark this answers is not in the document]
IsaiahWitzke: uh... i guess technically it's correct? but i dont know if i care too much... we do this RequiresUpgrade when you're on a free tier trying to use Opus or Sonnet, but the bedrock case is only going to be relevant to enterprises on the total oposite side of the spectrum from free... feel free to ignore

IsaiahWitzke: if it is the case that the server can use your bedrock creds for auto models though, maybe we just change the text... instead of "billed to API" or "billed to Bedrock", we need to probably communicate something kinda like "warp may use your aws creds if the auto model that we resolve to is one of your bedrock supported ones"

once again, needs some copy workshopping

JaidenRatti: you are right, the server can use bedrock creds for auto. will think of a better way to communicate that we *may* use your aws creds

app/src/terminal/input/models/data_source.rs
```diff
@@ -40,10 +40,13 @@ use warpui::keymap::Keystroke;
...
 
 use super::model_spec_scores::{
-    render_model_spec_header, render_model_spec_scores, CostRow, ModelSpecScoresLayout,
-    MODEL_SPECS_DESCRIPTION, MODEL_SPECS_TITLE, REASONING_LEVEL_DESCRIPTION, REASONING_LEVEL_TITLE,
+    render_model_spec_header, render_model_spec_scores, CostRow, CostRowTooltip,
+    ModelSpecScoresLayout, MODEL_SPECS_DESCRIPTION, MODEL_SPECS_TITLE, REASONING_LEVEL_DESCRIPTION,
+    REASONING_LEVEL_TITLE,
 };
 
+const AUTO_BEDROCK_TOOLTIP: &str = "Warp uses Bedrock when the resolved model supports it; otherwise it may use Warp-hosted inference.";
```
IsaiahWitzke:
```suggestion
const AUTO_BEDROCK_TOOLTIP: &str = "Warp uses Bedrock when the model Auto selects supports it; otherwise it may use Warp-hosted inference.";
```

nit... could talk about how the "Auto model is the one doing the selecting" to better match the [existing verbiage](https://github.com/warpdotdev/warp/blob/0d0b697d21ec38b49d956bcf3ac7aeeadc5091e9/app/src/terminal/profile_model_selector.rs#L2055) of auto in-app.

IsaiahWitzke (approved):

PR #11363, opened 2026-05-19
Order orchestration pills by status (attention first, done by recency)

advait-m:
Loom: https://www.loom.com/share/e011f21848864c4282c699fa426780a4
Fixes https://linear.app/warpdotdev/issue/QUALITY-739/move-done-agents-to-the-end-of-pill-bar

## Description
The orchestration pill bar should surface attention-needing agents and push inert agents out of the way, rather than rendering everything in spawn order.

This PR sorts both the pinned and unpinned sections of the pill bar by a status-priority key. The orchestrator pill is still first, the pinned/unpinned divider still appears between the two sections — only the order *within* each section changes.

### Ordering rules
Lower = closer to the start of each section (= more important to the user):

1. **Blocked** — waiting on the user, highest urgency
2. **Error** — needs the user to look
3. **InProgress** — running
4. **Done** (Cancelled + Success combined into a single bucket)

Within the non-done buckets the tiebreaker is **spawn order**, so freshly-spawned siblings stay where they were and don't shuffle while they run.

Within the **done** bucket the tiebreaker is **recency**: the most recently finished agent renders leftmost (based on the conversation's `last_modified_at` — latest exchange's `finish_time`). Pills with no known finish time sort to the trailing edge of the done section. Cancelled and Success share the same bucket key on purpose — once an agent is done, the only signal that matters for ordering is how recently it finished, not *how* it finished.

### Why these specific rules
- **Blocked > Error** — pills that explicitly need user input outrank pills that merely need attention.
- **Done bucket merged + sorted by recency** — a 5-minute-old Cancelled should sink behind a fresh Success. The user wants "left = more important", which for completed agents maps to "most recently active".
- **Re-sort on every render** — the pill bar already re-renders on `BlocklistAIHistoryEvent::UpdatedConversationStatus`, so status transitions (Blocked → InProgress when permission is granted, InProgress → Success when an agent finishes) reflect in the bar immediately. No extra plumbing needed.
- **Pinned section uses the same sort** — kept consistent with the unpinned section so users see the same ordering principle everywhere; pinned still wins the section-level ordering.

### Implementation
- New `pill_status_sort_key(Option<&ConversationStatus>) -> u8` helper (Blocked=0, Error=1, InProgress=2, Done=3).
- New `pill_done_recency_key(Option<i64>) -> i64` helper that returns the negation of the last-modified epoch-ms so larger timestamps (newer) sort first; missing timestamps sort last.
- `PillSpec` gains a `last_modified_ms: Option<i64>` field, populated via `AIConversation::last_modified_at()`.
- Render loop now collects `(status_key, secondary_key, spawn_index, Box<dyn Element>)` tuples per bucket and `sort_by_key`s each independently. `secondary_key` is the recency key for done pills and `0` everywhere else, so spawn order is the natural tiebreaker outside the done bucket.

## Testing
Five unit tests in `orchestration_pill_bar_tests.rs`:
- `pill_status_sort_key_orders_attention_then_in_progress_then_done` — Blocked < Error < InProgress < Done, with Cancelled == Success.
- `pill_status_sort_key_treats_none_as_in_progress` — orchestrator's `None` status safety default.
- `pill_done_recency_key_puts_most_recent_first_and_unknown_last` — direction of the recency key + None handling.
- `sort_pills_bubbles_attention_in_progress_keeps_spawn_done_uses_recency` — end-to-end mixed-status sort with recency tiebreaking inside done.
- `sort_pills_is_stable_within_in_progress_bucket` — InProgress pills keep spawn order regardless of input order.
- `sort_pills_done_bucket_orders_by_recency_regardless_of_completion_type` — an old Cancelled sinks behind a fresh Success.

### DIFF
--- a/app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
+++ b/app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs
@@ -102,6 +102,44 @@ pub(crate) fn pill_initial(name: &str) -> char {
         .map(|c| c.to_ascii_uppercase())
         .unwrap_or('A')
 }
+
+/// Sort priority for a pill given its conversation status. Lower numbers
+/// sort to the start of their section, so attention-needing pills
+/// (Blocked, Error) lead and finished pills (Cancelled, Success) trail.
+/// Applied to both the pinned and the unpinned sections.
+///
+/// Cancelled and Success share the same key so the trailing "done"
+/// section is sorted by recency (most recently finished first — see
+/// `pill_done_recency_key`) rather than by completion type. Within the
+/// non-done buckets (Blocked / Error / InProgress) spawn order remains
+/// the tiebreaker so freshly-spawned siblings don't shuffle around
+/// while they run.
+///
+/// Orchestrator pills carry `status = None` but are never routed through
+/// this sort (they always render first), so the `None` arm just picks
+/// the same key as `InProgress` for safety.
+fn pill_status_sort_key(status: Option<&ConversationStatus>) -> u8 {
+    match status {
+        Some(ConversationStatus::Blocked { .. }) => 0,
+        Some(ConversationStatus::Error) => 1,
+        Some(ConversationStatus::InProgress) => 2,
+        Some(ConversationStatus::Cancelled) | Some(ConversationStatus::Success) => 3,
+        None => 2,
+    }
+}
+
+/// Tiebreaker for pills that fall in the trailing "done" bucket
+/// (`pill_status_sort_key` == 3). Returns a value that sorts ascending
+/// so the most recently finished pill ends up leftmost. Pills without a
+/// known finish time (`None`) sort to the end of the done section.
+///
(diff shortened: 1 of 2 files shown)


cephalonaut:
Awesome! Seems helpful. I'm a little unsure whether the auto sorting will break expectations of stability, but might be just what people need too. We could always add some sort of option if so.

advait-m:
> Awesome! Seems helpful. I'm a little unsure whether the auto sorting will break expectations of stability, but might be just what people need too. We could always add some sort of option if so.

Ah, yeah if folks raise complaints we could add an option in settings for this perhaps

advait-m:
@cephalonaut did you mean to approve (not seeing it)? if not, lemme know if I should change anything here

cephalonaut (approved):

cephalonaut:
Sorry - forgot to approve!

advait-m:
> Sorry - forgot to approve!

Np, thanks!

PR #11366, opened 2026-05-19
Add reload credit confirmations for team actions

tylerlam-warp:
## Description
Adds reload-credit warning confirmation copy for team leave and member removal flows behind `BillingAndUsagePageV2`. Leave uses the reload-credit copy only when the current user has remaining user-level interactive bonus credits; member removal shows the warning whenever the feature flag is enabled. Delete team continues to use the existing delete confirmation.

## Testing
Removing a user from a team:
image

Leaving a team:
image

Deleting a team:
image

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/src/settings_view/teams_page.rs
+++ b/app/src/settings_view/teams_page.rs
@@ -790,12 +790,9 @@ impl TeamsPageView {
 
         let team_action_confirmation_dialog =
             ctx.add_typed_action_view(|_| CloudActionConfirmationDialog::new());
-        ctx.subscribe_to_view(
-            &team_action_confirmation_dialog,
-            |me, _, event, ctx| {
-                me.handle_cloud_action_confirmation_dialog_event(event, ctx);
-            },
-        );
+        ctx.subscribe_to_view(&team_action_confirmation_dialog, |me, _, event, ctx| {
+            me.handle_cloud_action_confirmation_dialog_event(event, ctx);
+        });
 
         let transfer_ownership_modal_body =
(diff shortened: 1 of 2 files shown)


app/src/settings_view/teams_page.rs
```diff
@@ -501,7 +511,18 @@ impl TypedActionView for TeamsPageView {
             TeamsPageAction::LeaveTeam => self.leave_team(ctx),
             TeamsPageAction::CreateTeam => self.create_team(ctx),
             TeamsPageAction::RemoveUserFromTeam { user_uid, team_uid } => {
-                self.remove_user_from_team(*user_uid, *team_uid, ctx)
+                if FeatureFlag::BillingAndUsagePageV2.is_enabled() {
```
IsaiahWitzke, line 514: idk if possible, but can we query the user's remaining personal addon credits at this point? if they're 0 already, we probably don't need to show the alert

im guessing "no... this requires peronsal addon credits to exist on the workspace member graph" or something... if that's the case, do you mind just addint a ticket to linear?

IsaiahWitzke (approved):

app/src/settings_view/teams_page.rs
tylerlam-warp: ah I missed your comment.

Yeah, but I don't believe we store the workspace member bonus grant data, and we don't necessarily expose that information in the usage table either.

I can create a linear ticket for this, although I'm not entirely sure if there's a **amazing** use case for this right now other than for this

Issue #6945, opened 2025-07-31
MCP changes string value to float


PaulDHaes:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [x] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/support-and-billing/sending-us-feedback#gathering-warp-logs)

### Describe the bug

Using the MCP server it looks like a string value is getting changed to a float value.
I'm using a tool called burpsuite with a extension called mcp-server.
Using this it should be able to make new web request but the problem is that the web port number is getting changed to a float.
normally a request should look like this 127.0.0.1:80 but the target port (80) is changed to 80.0 instead of \"80\" this ofcourse does not work and therefore the request fails.
All other tools of the program work except making a new web request

### To reproduce

Install burpsuite commonity eddition.
install the mcp extention.
<img width="1343" height="1021" alt="Image" src="https://github.com/user-attachments/assets/b986468b-b683-46c4-a1d7-420a078f1406" />
configure warp mcp server like this and change the <me> to the path of the create jar file that can be done by clicking on the extension bottom right. 

```
{
  "burp": {
    "command": "/Applications/Burp Suite Professional.app/Contents/Resources/jre.bundle/Contents/Home/bin/java",
    "args": [
      "-jar",
      "/Users/<me>/.BurpSuite/mcp-proxy/mcp-proxy-all.jar",
      "--sse-url",
      "http://127.0.0.1:9876"
    ],
    "env": {},
    "working_directory": null
  }
```

ask the AI agent to make a request using burp to a ip address like 127.0.0.1

### Expected behavior

Using Claude or open WebUI this worked but only using warp it failes in claude dekstop it looked like this.

<img width="748" height="297" alt="Image" src="https://github.com/user-attachments/assets/b528d2a6-290e-4801-8c1f-d710aee01a3e" />

`send_http1_request`
```
{
  `content`: `GET /vulnerabilities/exec/ HTTP/1.1
Host: 127.0.0.1:8080
Accept-Language: en-GB,en;q=0.9
Upgrade-Insecure-Requests: 1
[...]

PR #11407, opened 2026-05-20
fix(mcp): recursively coerce integer args nested in tool input schemas

### LINKED ISSUE #10596: Warp sends whole-number floats (5.0) to MCP tools instead of integers (5), violating tool input schemas
### Pre-submit Checks

### Describe the bug

similar to https://github.com/warpdotdev/warp/issues/6945 which was solved, but with nested integers.
When the agent calls an MCP tool with a numeric argument that the tool's `inputSchema` declares as `"type": "integer"`, Warp sends a JSON float (e.g. `5.0`, `1730419200000.0`) instead of a JSON integer (e.g. `5`, `1730419200000`).

Many OpenAPI-backed servers like Cortex/Snyk-style auditors) reject these calls with `HTTP 500` or `Bad Request`, because their schema validation distinguishes integers from floats.

There **is** existing coercion code for exactly this — `coerce_integer_args` in `app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs` — but it only handles the simplest case: a top-level property whose schema is literally `{"type": "integer"}`. Every other common JSON Schema shape silently leaks the float through.

### To reproduce

1. Configure an MCP tool whose `inputSchema` declares an integer field nested inside an object, an array item, a `oneOf` branch, a `$ref`, or as a nullable `["integer", "null"]`.

   Minimal repro schema (matches a real PANW mcp server tool):

   ```yaml
   AuditManagementLogRequest:
     type: object
     properties:
       request_data:
         type: object
         properties:
           search_from: { type: integer }
           search_to:   { type: integer }
           filters:
             type: array
             items:
               oneOf:
                 - { properties: { value: { type: array, items: { type: string } } } }
                 - { properties: { value: { type: integer } } }   # ← timestamp ms
   ```

2. Ask the agent to call the tool with values like `search_from: 0`, `search_to: 100`, and a timestamp filter `value: 1730419200000`.
3. Observe (server logs / network capture) that Warp sends:

   ```json
   {
     "request_data": {
       "search_from": 0.0,
       "search_to": 100.0,
       "filters": [{ "field": "timestamp", "operator": "gte", "value": 1730419200000.0 }]
     }
   }
   ```

   The server rejects the call.

### Expected behavior

Warp should send:

```json
{
  "request_data": {
    "search_from": 0,
    "search_to": 100,
    "filters": [{ "field": "timestamp", "operator": "gte", "value": 1730419200000 }]
  }
}
```

Image

### Operating system (OS)

macOS

macOS 26.4.1 (build 25E253), Apple Silicon (arm64)

### Shell Version

zsh 5.9

### Current Warp version

v0.2026.05.06.15.42.03.stable_03

### Regression

### Recent working Warp date

_No response_

### Additional context

MCP tool arguments round-trip through `google.protobuf.Struct` between the agent backend and the Warp client. `structpb.NumberValue` stores **all** numbers as `f64`, erasing the integer/float distinction. By the time the args land in the client as a `serde_json::Value`, an integer `5` has become `5.0`, and serde_json (via ryu) serializes it back out as `"5.0"` — which strict MCP servers reject for integer-typed fields.

The mitigation lives in `app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs`:

```rust
pub(crate) fn coerce_integer_args(
    args: &mut serde_json::Map<String, serde_json::Value>,
    input_schema: &serde_json::Map<String, serde_json::Value>,
) {
    let Some(properties) = input_schema.get("properties").and_then(|p| p.as_object()) else {
        return;
    };

    for (key, prop_def) in properties {
        let is_integer = prop_def.get("type").and_then(|t| t.as_str()) == Some("integer");
        if !is_integer { continue; }
        // ...coerce f64 -> i64
    }
}
```

It walks **only depth-1 properties** and **only** schemas whose `type` is the exact string `"integer"`. It does not recurse

None

david-engelmann:
Fixes #10596.

## Why

MCP tool args round-trip through `google.protobuf.Struct` between the agent backend and the Warp client. `structpb.NumberValue` stores all numbers as `f64`, erasing the integer/float distinction. By the time args arrive client-side as `serde_json::Value`, an integer `5` has become `5.0`, and serde_json serializes it back as `\"5.0\"` — which strict MCP servers reject for `\"type\": \"integer\"` fields.

The previous mitigation (`coerce_integer_args`, originally from #6945) only walked depth-1 `properties` and matched the literal string `\"integer\"`, so every nested or composed integer field leaked the float through.

## What

Walks schema and args in parallel covering:

- nested `properties` (objects)
- array `items` (single-schema and positional/tuple form)
- `additionalProperties` (for keys outside `properties`)
- `allOf` / `oneOf` / `anyOf` (every branch applied — coercion is conservative and a non-matching branch is a no-op, so this sidesteps the ambiguity of \"picking the right branch\")
- internal `$ref` pointers (`#/$defs/Foo`, `#/definitions/Foo`) with a cycle guard
- nullable type-arrays like `[\"integer\", \"null\"]` and singleton `[\"integer\"]`

Intentionally skipped (no-op, preserves existing wire behavior): external `$ref` URLs, `not`, `if`/`then`/`else`, `patternProperties`. These are vanishingly rare in real MCP tool schemas, and skipping them never makes a value *worse* — at worst the float survives, which is today's behavior.

The public signature `coerce_integer_args(&mut Map, &Map)` is unchanged, so the existing call site at [app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs](app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs) is untouched.

## Before / after (the issue's PANW repro)

Schema (excerpt):

\`\`\`yaml
AuditManagementLogRequest:
  type: object
  properties:
    request_data:
      type: object
      properties:
        search_from: { type: integer }
        search_to:   { type: integer }
        filters:
          type: array
          items:
            oneOf:
              - { properties: { value: { type: array, items: { type: string } } } }
              - { properties: { value: { type: integer } } }
\`\`\`

Before — wire payload:

\`\`\`json
{ \"request_data\": { \"search_from\": 0.0, \"search_to\": 100.0,
    \"filters\": [{ \"field\": \"timestamp\", \"operator\": \"gte\", \"value\": 1730419200000.0 }] } }
\`\`\`

After — wire payload:

\`\`\`json
{ \"request_data\": { \"search_from\": 0, \"search_to\": 100,
    \"filters\": [{ \"field\": \"timestamp\", \"operator\": \"gte\", \"value\": 1730419200000 }] } }
\`\`\`

## Test plan

- [x] \`cargo build -p warp --lib\` — clean
- [x] \`cargo test -p warp --lib ai::blocklist::action_model::execute::call_mcp_tool\` — 21/21 pass (2 existing + 19 new)
- [x] \`cargo clippy -p warp --lib --tests -- -D warnings\` — clean
- [x] \`cargo fmt -p warp -- --check\` — clean

New test coverage by shape:

- nested object: \`nested_object_integer_is_coerced\`
- array items (single schema): \`array_items_integer_is_coerced\`
- tuple items: \`tuple_items_coerced_positionally\`
- \`oneOf\` branch: \`one_of_branch_integer_is_coerced\`
- \`anyOf\` branch: \`any_of_branch_integer_is_coerced\`
- \`allOf\` branches: \`all_of_branches_apply_in_order\`
- nullable `[\"integer\", \"null\"]`: \`nullable_integer_is_coerced\` + \`null_value_for_nullable_integer_passes_through\`
- singleton `[\"integer\"]`: \`singleton_type_array_integer_is_coerced\`
- internal \`\$ref\`: \`internal_ref_resolves\`
- self-referencing \`\$ref\` (cycle terminates): \`internal_ref_cycle_terminates\`
- pure A→B→A schema cycle: \`pure_ref_cycle_does_not_loop\`
- external \`\$ref\`: \`external_ref_is_skipped\`
- \`additionalProperties\`: \`additional_properties_integer_is_coerced\` + \`additional_properties_does_not_overwrite_known_properties\`
- conservative: \`non_whole_float_is_not_truncated\`, \`out_of_i64_range_is_not_coerced\`, \`large_timestamp_is_coerced\`
- end-to-end PANW repro: \`panw_audit_management_repro_coerces_all_integers\`

### DIFF
--- a/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
+++ b/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
@@ -250,10 +250,31 @@ fn coerce_recursive(
                     .and_then(|v| v.as_object())
                     .map(|p| p.keys().map(String::as_str).collect())
                     .unwrap_or_default();
+                // Keys covered by `patternProperties` are governed by their
+                // pattern's schema, not by `additionalProperties` — so they
+                // must be excluded here even though we don't (yet) coerce
+                // through `patternProperties` itself. Without this carve-out
+                // a pattern-governed value could be coerced by the wrong
+                // schema. Patterns that fail to compile are ignored, which
+                // matches the existing "unsupported shapes are skipped"
+                // policy elsewhere in this walker.
+                let pattern_regexes: Vec<regex::Regex> = schema
+                    .get("patternProperties")
+                    .and_then(|v| v.as_object())
+                    .map(|p| {
+                        p.keys()
+                            .filter_map(|pat| regex::Regex::new(pat).ok())
+                            .collect()
+                    })
+                    .unwrap_or_default();
                 for (k, v) in map.iter_mut() {
-                    if !known.contains(k.as_str()) {
-                        coerce_recursive(v, additional, root);
+                    if known.contains(k.as_str()) {
+                        continue;
                     }
+                    if pattern_regexes.iter().any(|re| re.is_match(k)) {
+                        continue;
+                    }
+                    coerce_recursive(v, additional, root);
                 }
             }
         }
--- a/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
+++ b/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
@@ -243,39 +243,52 @@ fn coerce_recursive(
                 }
             }
         }
-        if let Some(additional) = schema.get("additionalProperties") {
-            if additional.is_object() {
-                let known: HashSet<&str> = schema
-                    .get("properties")
-                    .and_then(|v| v.as_object())
-                    .map(|p| p.keys().map(String::as_str).collect())
-                    .unwrap_or_default();
-                // Keys covered by `patternProperties` are governed by their
-                // pattern's schema, not by `additionalProperties` — so they
-                // must be excluded here even though we don't (yet) coerce
-                // through `patternProperties` itself. Without this carve-out
-                // a pattern-governed value could be coerced by the wrong
-                // schema. Patterns that fail to compile are ignored, which
-                // matches the existing "unsupported shapes are skipped"
-                // policy elsewhere in this walker.
-                let pattern_regexes: Vec<regex::Regex> = schema
-                    .get("patternProperties")
-                    .and_then(|v| v.as_object())
-                    .map(|p| {
-                        p.keys()
-                            .filter_map(|pat| regex::Regex::new(pat).ok())
-                            .collect()
-                    })
-                    .unwrap_or_default();
-                for (k, v) in map.iter_mut() {
-                    if known.contains(k.as_str()) {
-                        continue;
-                    }
-                    if pattern_regexes.iter().any(|re| re.is_match(k)) {
-                        continue;
-                    }
-                    coerce_recursive(v, additional, root);
+        'additional: {
+            let Some(additional) = schema.get("additionalProperties") else {
+                break 'additional;
+            };
+            if !additional.is_object() {
+                break 'additional;
+            }
+            let known: HashSet<&str> = schema
+                .get("properties")
+                .and_then(|v| v.as_object())
+                .map(|p| p.keys().map(String::as_str).collect())
+                .unwrap_or_default();
+            // Keys covered by `patternProperties` are governed by their
+            // pattern's schema, not by `additionalProperties` — so they
+            // must be excluded here even though we don't (yet) coerce
+            // through `patternProperties` itself.
+            //
+            // JSON Schema's pattern grammar is ECMA-262; Rust's `regex`
+            // crate is a strict subset and rejects valid patterns such as
+            // those with lookaheads or certain Unicode constructs. We
+            // can't safely tell those apart from "this key doesn't match
+            // the pattern", so if any pattern fails to compile we fall
+            // back to skipping `additionalProperties` for this schema —
+            // a key that should have been governed by the pattern won't
+            // get coerced with the wrong schema.
+            let pattern_compile = schema
+                .get("patternProperties")
+                .and_then(|v| v.as_object())
+                .map(|p| {
+                    p.keys()
+                        .map(|pat| regex::Regex::new(pat))
+                        .collect::<Result<Vec<_>, _>>()
+                });
+            let pattern_regexes = match pattern_compile {
+                Some(Ok(res)) => res,
+                Some(Err(_)) => break 'additional,
+                None => Vec::new(),
+            };
+            for (k, v) in map.iter_mut() {
+                if known.contains(k.as_str()) {
+                    continue;
                 }
+                if pattern_regexes.iter().any(|re| re.is_match(k)) {
+                    continue;
+                }
+                coerce_recursive(v, additional, root);
             }
         }
     }
--- a/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
+++ b/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
@@ -168,6 +168,24 @@ impl Entity for CallMCPToolExecutor {
     type Event = ();
 }
 
+/// Maximum recursion depth for [`coerce_recursive`]. Real MCP schemas top out
+/// well under this — the limit only matters for schemas that intentionally
+/// or accidentally compose into a cycle (e.g. `allOf: [{$ref: "#/Self"}]`
+/// where `Self.allOf = [{$ref: "#/Self"}]`). [`resolve_refs`] already breaks
+/// pure `$ref` chains; this guards every other recursive edge so a malicious
+/// MCP server can't hang the client by advertising a recursive composed
+/// schema (#10596 review).
+const MAX_COERCE_DEPTH: usize = 64;
+
+/// Maximum total recursive calls allowed across a single
+/// [`coerce_integer_args`] invocation. The depth cap alone doesn't bound work:
+/// a schema with branching factor B and depth D produces up to `B^D` calls,
+/// so an adversarial schema with `allOf: [s, s]` where each `s` references
+/// the same shape can still fan out exponentially before any single chain
+/// reaches [`MAX_COERCE_DEPTH`]. Real MCP schemas use a few hundred calls at
+/// most; 10_000 leaves ~20× headroom for unusual but legitimate shapes.
+const MAX_COERCE_OPS: usize = 10_000;
+
 /// Coerces whole-number floats in `args` to integers wherever the tool's JSON
 /// Schema `input_schema` declares an [integer type], at any depth.
 ///
@@ -184,15 +202,6 @@ impl Entity for CallMCPToolExecutor {
 /// behavior.
 ///
 /// [integer type]: https://json-schema.org/understanding-json-schema/reference/type
-/// Maximum recursion depth for [`coerce_recursive`]. Real MCP schemas top out
-/// well under this — the limit only matters for schemas that intentionally
-/// or accidentally compose into a cycle (e.g. `allOf: [{$ref: "#/Self"}]`
-/// where `Self.allOf = [{$ref: "#/Self"}]`). [`resolve_refs`] already breaks
-/// pure `$ref` chains; this guards every other recursive edge so a malicious
-/// MCP server can't hang the client by advertising a recursive composed
-/// schema (#10596 review).
-const MAX_COERCE_DEPTH: usize = 64;
-
 pub(crate) fn coerce_integer_args(
     args: &mut serde_json::Map<String, serde_json::Value>,
     input_schema: &serde_json::Map<String, serde_json::Value>,
@@ -201,7 +210,8 @@ pub(crate) fn coerce_integer_args(
     // `$ref` resolution against the same root). Schemas are small in practice.
     let root = serde_json::Value::Object(input_schema.clone());
     let mut value = serde_json::Value::Object(std::mem::take(args));
-    coerce_recursive(&mut value, &root, &root, 0);
+    let mut budget = MAX_COERCE_OPS;
+    coerce_recursive(&mut value, &root, &root, 0, &mut budget);
     if let serde_json::Value::Object(map) = value {
         *args = map;
     }
@@ -214,16 +224,18 @@ fn coerce_recursive(
     schema: &serde_json::Value,
     root: &serde_json::Value,
     depth: usize,
+    budget: &mut usize,
 ) {
-    if depth >= MAX_COERCE_DEPTH {
+    if depth >= MAX_COERCE_DEPTH || *budget == 0 {
         return;
     }
+    *budget -= 1;
     let schema = resolve_refs(schema, root);
 
     // `allOf` — apply every branch; can stack with sibling keywords.
     if let Some(branches) = schema.get("allOf").and_then(|v| v.as_array()) {
         for b in branches {
-            coerce_recursive(value, b, root, depth + 1);
+            coerce_recursive(value, b, root, depth + 1, budget);
         }
     }
 
@@ -235,7 +247,7 @@ fn coerce_recursive(
     for key in ["oneOf", "anyOf"] {
         if let Some(branches) = schema.get(key).and_then(|v| v.as_array()) {
             for b in branches {
-                coerce_recursive(value, b, root, depth + 1);
+                coerce_recursive(value, b, root, depth + 1, budget);
             }
         }
     }
@@ -252,7 +264,7 @@ fn coerce_recursive(
         if let Some(props) = schema.get("properties").and_then(|v| v.as_object()) {
             for (k, child_schema) in props {
                 if let Some(child_value) = map.get_mut(k) {
-                    coerce_recursive(child_value, child_schema, root, depth + 1);
+                    coerce_recursive(child_value, child_schema, root, depth + 1, budget);
                 }
             }
         }
@@ -301,7 +313,7 @@ fn coerce_recursive(
                 if pattern_regexes.iter().any(|re| re.is_match(k)) {
                     continue;
                 }
-                coerce_recursive(v, additional, root, depth + 1);
+                coerce_recursive(v, additional, root, depth + 1, budget);
             }
         }
     }
@@ -312,12 +324,12 @@ fn coerce_recursive(
         match schema.get("items") {
             Some(item_schema @ serde_json::Value::Object(_)) => {
                 for v in arr.iter_mut() {
-                    coerce_recursive(v, item_schema, root, depth + 1);
+                    coerce_recursive(v, item_schema, root, depth + 1, budget);
                 }
             }
             Some(serde_json::Value::Array(item_schemas)) => {
                 for (v, s) in arr.iter_mut().zip(item_schemas.iter()) {
-                    coerce_recursive(v, s, root, depth + 1);
+                    coerce_recursive(v, s, root, depth + 1, budget);
                 }
             }
             _ => {}
--- a/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
+++ b/app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use futures::future::BoxFuture;
 use futures::FutureExt;
 #[cfg(not(target_family = "wasm"))]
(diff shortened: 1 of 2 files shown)


david-engelmann:
Thanks for the catch — you're right, a key matching a `patternProperties` regex is governed by its pattern's schema, not `additionalProperties`, and my walker was leaking through.

Addressed in `f46924c0`:

- When iterating `additionalProperties`, compile each `patternProperties` key as a regex and skip args matching any of them. The pattern's own schema still isn't followed (we keep treating `patternProperties` as out-of-scope), but those keys no longer fall back to `additionalProperties` incorrectly.
- Patterns that fail to compile are ignored — matching the "unsupported shapes are skipped" policy used everywhere else in the walker.

Added 2 regression tests in [call_mcp_tool_tests.rs](app/src/ai/blocklist/action_model/execute/call_mcp_tool_tests.rs):
- `pattern_properties_keys_are_excluded_from_additional_properties` — pattern-matched key keeps its float (governed by `patternProperties: { type: number }`); unrelated key is still coerced via `additionalProperties: { type: integer }`.
- `invalid_pattern_property_regex_is_ignored` — an uncompilable regex (`"("`) doesn't panic and doesn't block sibling coercion.

23/23 in `call_mcp_tool` pass; clippy + fmt clean.

Re: visual evidence — this PR is a behind-the-scenes serialization fix (whole-number f64 → i64 on the wire), so there's no UI surface to capture. The PANW-style end-to-end repro lives in [`panw_audit_management_repro_coerces_all_integers`](app/src/ai/blocklist/action_model/execute/call_mcp_tool_tests.rs) — it pins the exact schema + payload shape from the issue body.

/oz-review

david-engelmann:
Good catch — Rust's regex crate is a strict ECMA-262 subset and silently rejecting unparseable patterns let a real pattern-governed key slip through to `additionalProperties`.

Addressed in `266845ed`:

- If any `patternProperties` entry fails to compile, break out of the `additionalProperties` branch entirely (labeled break, scoped to just that block). `properties` recursion at the same level is unaffected.
- A key that should have been governed by the unparseable pattern won't be coerced with the wrong schema — it just keeps its float on the wire.

Replaced `invalid_pattern_property_regex_is_ignored` (which encoded the old, unsafe behavior) with two new tests:
- `uncompilable_pattern_skips_additional_properties_conservatively` — would-be-coerced key keeps its float when the pattern is unparseable.
- `uncompilable_pattern_still_allows_properties_recursion` — `properties` coercion still runs alongside the unparseable pattern.

24/24 in `call_mcp_tool` pass; clippy + fmt clean.

/oz-review

david-engelmann:
Right — `resolve_refs` only breaks pure `$ref` chains; the other recursive edges had no cross-cutting guard.

Addressed in `726c8535`:

- Threaded a `depth: usize` counter through `coerce_recursive`. Bails at `MAX_COERCE_DEPTH = 64`.
- Real MCP schemas top out well under this; the limit only matters for cycles. Termination is the guarantee — correctness on an intentionally-recursive schema isn't a goal (the server would have to be cooperating to send valid args anyway).

Added 3 regression tests:
- `all_of_self_reference_terminates` — `$defs.Loop = { allOf: [{ "$ref": "#/$defs/Loop" }], properties: {…} }`. Without the limit this looped indefinitely through `allOf → resolve_refs → allOf → …`.
- `one_of_self_reference_terminates` — same shape via `oneOf`.
- `deep_finite_nesting_under_limit_still_coerces` — 10 levels of nested `properties` still coerces (pins that the limit doesn't block legitimate schemas).

27/27 in `call_mcp_tool` pass; clippy + fmt clean.

/oz-review

david-engelmann:
Right — the depth cap bounds the longest single chain, but `allOf: [Self, Self]` doubles work at each level, so a B=2 schema can fan out to ~2^64 calls before any single chain hits the depth cap.

Addressed in \`2f784f01\`:

- Added \`MAX_COERCE_OPS = 10_000\` and threaded a \`budget: &mut usize\` through \`coerce_recursive\`. Each recursive call decrements the budget; the function bails on either depth ≥ 64 OR budget == 0, whichever fires first.
- 10_000 ops bounds the worst case to a few milliseconds while leaving ~20× headroom for real MCP schemas (which use a few hundred calls in practice).

Added 1 regression test:
- \`fan_out_via_all_of_self_reference_terminates\` — \`allOf: [{ $ref: "#/Self" }, { $ref: "#/Self" }]\` doubling schema. Without the op budget this generated 2^D calls; with the budget it bails at 10_000.

(Also reordered the const declarations so the \`coerce_integer_args\` docstring attaches to the function rather than to \`MAX_COERCE_DEPTH\`.)

28/28 in \`call_mcp_tool\` pass. clippy + fmt clean.

/oz-review

david-engelmann:
Visual evidence for the nested integer coercion, recorded end-to-end against a `warp-oss` binary built from this PR's branch (`/tmp/warp-10596` @ `2f784f01`).

#11407 nested integer coercion — end-to-end recording

(MP4: https://github.com/david-engelmann/warp-taper/raw/01d9cab83251127f83082b9596cf6aa144e83177/docs/evidence/11407-mcp-integer-coercion/master-captioned.mp4)

## What the recording shows

| Time | Caption | On-screen content |
|---|---|---|
| 0-4s | "warpdotdev/warp PR #11407 - nested integer coercion" | warp-oss terminal at `/tmp/warp-10596` on branch `david/10596-mcp-integer-coercion` |
| 4-10s | "Schema declares search_from / search_to / filters[].value as integer" | `cat` output of the PANW test schema — `object → object → array → oneOf → integer` (the exact shape from the issue body) |
| 10-16s | "Without the fix the agent would send 0.0 / 100.0 / 1730419200000.0" | schema continues |
| 16-22s | "Walker recurses through object - array - oneOf - integer leaves" | `cargo test ... -- --nocapture` issued |
| 22-32s | "Test passes - all three nested integers reach the wire as integers" | `test result: ok. 1 passed; 0 failed` — the three nested integer fields all coerced from f64 → i64 |

## The test the recording exercises

From [call_mcp_tool_tests.rs:591](app/src/ai/blocklist/action_model/execute/call_mcp_tool_tests.rs#L591):

```rust
fn panw_audit_management_repro_coerces_all_integers() {
    let mut args = obj(json!({
        "request_data": {
            "search_from": 0.0,
            "search_to": 100.0,
            "filters": [
                { "field": "timestamp", "operator": "gte", "value": 1730419200000.0 }
            ]
        }
    }));
    let schema = obj(json!({
        "type": "object",
        "properties": {
            "request_data": {
                "type": "object",
                "properties": {
                    "search_from": { "type": "integer" },
                    "search_to":   { "type": "integer" },
                    "filters": {
                        "type": "array",
                        "items": {
                            "oneOf": [
                                { "properties": { "value": { "type": "array", "items": { "type": "string" } } } },
                                { "properties": { "value": { "type": "integer" } } }
                            ]
                        }
                    }
                }
            }
        }
    }));

    coerce_integer_args(&mut args, &schema);

    assert_serialized_as(&args, "request_data.search_from", "0");
    assert_serialized_as(&args, "request_data.search_to", "100");
    assert_serialized_as(&args, "request_data.filters.0.value", "1730419200000");
}
```

The three assertions pin the wire format: each integer field that the agent supplies as `5.0` / `100.0` / `1730419200000.0` reaches `git ls-files` / the MCP server as the string `"5"` / `"100"` / `"1730419200000"` after `coerce_integer_args` runs — i.e. the receiving MCP server sees JSON integers, not whole-number floats.

## Reproducing

The scenario + recipe + the new `warp-taper caption` CLI subcommand all live in [warp-taper @ 01d9cab](https://github.com/david-engelmann/warp-taper/tree/01d9cab83251127f83082b9596cf6aa144e83177):

- [Scenario](https://github.com/david-engelmann/warp-taper/blob/01d9cab/crates/warp-taper-core/src/scenarios/mcp_integer_coercion.rs) (5 timed captions, metadata)
- [Recipe](https://github.com/david-engelmann/warp-taper/blob/01d9cab/scripts/recipes/11407-mcp-integer-coercion.json) (warp-driver steps)
- Run with `RECIPE=11407-mcp-integer-coercion DURATION_S=35 EMIT_GIF=1 ./scripts/record-agent-evidence.sh` against a warp-oss built from this PR's branch.

Full call-mcp-tool suite (all 28 tests covering the walker's recursion shapes — composition, `$ref`, `additionalProperties`, depth guard, op budget):

```
test result: ok. 28 passed; 0 failed; 0 ignored; 0 measured; 3613 filtered out; finished in 0.02s
```

david-engelmann:
Rebased onto upstream/master (`885c5406`) to clear the merge conflict in `call_mcp_tool.rs` (import collision from #11426's wasm gate). All 28 integer-coercion tests still pass post-rebase. Branch tip: `c876721d`.

/oz-review

david-engelmann:
Addressed both concerns in `f08cec79` (rebased onto latest master).

**1. `oneOf`/`anyOf` coercing through non-matching branches.** Added `branch_excluded`, which skips a composition branch when the value *provably* violates its `const`, `enum`, `type`, `required`, or a `const`/`enum` discriminator on a property present in the value. So a tagged union like

```json
{ "oneOf": [
  { "properties": { "kind": {"const":"a"}, "val": {"type":"integer"} } },
  { "properties": { "kind": {"const":"b"}, "val": {"type":"number"} } }
]}
```

with `{ "kind": "b", "val": 5.0 }` now leaves `val` as `5.0` — branch "a" is excluded by the `kind` discriminator. It's a deliberately partial check: it only ever *excludes* provable non-matches, so untagged unions behave exactly as before, and a whole-number float still counts as a valid `integer` candidate (the case this walker exists for).

**2. `patternProperties` regex compilation on untrusted input.** Rust's `regex` is linear-time at match (finite automaton, no catastrophic backtracking), so the real surface is pathological *compile* cost. `compile_pattern_properties` now bounds pattern count (64), per-pattern length (1024), and compiled program size (64 KiB via `RegexBuilder::size_limit` + `dfa_size_limit`). Any rejection falls back to the existing conservative skip of `additionalProperties` coercion.

8 new tests covering tagged-union branch selection (const / enum / required / type, both matching and non-matching) and the three pattern-compile bounds. All 36 coercion tests pass; clippy + fmt clean.

/oz-review

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
```diff
@@ -166,36 +168,381 @@ impl Entity for CallMCPToolExecutor {
...
-/// Coerces whole-number floats in `args` to integers for fields declared as
-/// [`"type": "integer"`](https://json-schema.org/understanding-json-schema/reference/type)
-/// in the tool's JSON Schema `input_schema`.
+/// Maximum recursion depth for [`coerce_recursive`]. Real MCP schemas top out
+/// well under this — the limit only matters for schemas that intentionally
+/// or accidentally compose into a cycle (e.g. `allOf: [{$ref: "#/Self"}]`
+/// where `Self.allOf = [{$ref: "#/Self"}]`). [`resolve_refs`] already breaks
+/// pure `$ref` chains; this guards every other recursive edge so a malicious
+/// MCP server can't hang the client by advertising a recursive composed
+/// schema (#10596 review).
```
vkodithala: nit: We shouldn't directly reference "#10596 review" here. PR-specific context doesn't belong in code comments; contributors can reference the original PR/git history as needed to determine intent.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
```diff
@@ -166,36 +168,381 @@ impl Entity for CallMCPToolExecutor {
...
+const MAX_COERCE_DEPTH: usize = 64;
+
+/// Maximum total recursive calls allowed across a single
+/// [`coerce_integer_args`] invocation. The depth cap alone doesn't bound work:
+/// a schema with branching factor B and depth D produces up to `B^D` calls,
+/// so an adversarial schema with `allOf: [s, s]` where each `s` references
+/// the same shape can still fan out exponentially before any single chain
+/// reaches [`MAX_COERCE_DEPTH`]. Real MCP schemas use a few hundred calls at
+/// most; 10_000 leaves ~20× headroom for unusual but legitimate shapes.
+const MAX_COERCE_OPS: usize = 10_000;
```
vkodithala: nit: Marking this as a 'nit' since in practice, we probably won't see schema recursion of this magnitude unless we have a malformed server. What's the rationale behind the upper bound being 10k here? I'd imagine we have something much smaller i.e. on the scale of 100 - 1,000, but I could be missing context.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
```diff
@@ -166,36 +168,381 @@ impl Entity for CallMCPToolExecutor {
...
+
+/// Bounds on `patternProperties` regex compilation. The patterns come from the
+/// MCP server's advertised schema, which is untrusted input. Rust's `regex`
+/// crate already guarantees linear-time *matching* (it's a finite automaton —
+/// no catastrophic backtracking), so the only attack surface is forcing a
+/// pathological *compile*: a huge or deeply-nested pattern that blows up
+/// memory/CPU while building the program. We defend by capping the number of
+/// patterns, the length of each, and the compiled program size (#10596 review).
```
vkodithala: Ditto to my earlier comment on leaving out PR-specific context.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
```diff
@@ -166,36 +168,381 @@ impl Entity for CallMCPToolExecutor {
...
+fn resolve_refs<'a>(
+    schema: &'a serde_json::Value,
+    root: &'a serde_json::Value,
+) -> &'a serde_json::Value {
+    let mut visited = HashSet::<String>::new();
+    let mut current = schema;
+    while let Some(ref_str) = current.get("$ref").and_then(|v| v.as_str()) {
+        if !ref_str.starts_with('#') || !visited.insert(ref_str.to_string()) {
+            return current;
+        }
```
vkodithala, line 510: nit: Can you add a comment here around this early-return? A little hard to grok. Maybe reference that non-internal references != supported _and_ we don't want to incur infinite recursion? Unless I'm butchering the intent.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
```diff
@@ -165,37 +167,214 @@ impl Entity for CallMCPToolExecutor {
...
+/// integers wherever the schema declares an integer type.
+fn coerce_recursive(
+    value: &mut serde_json::Value,
+    schema: &serde_json::Value,
+    root: &serde_json::Value,
+) {
+    let schema = resolve_refs(schema, root);
+
+    // `allOf` — apply every branch; can stack with sibling keywords.
+    if let Some(branches) = schema.get("allOf").and_then(|v| v.as_array()) {
```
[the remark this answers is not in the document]
vkodithala: @david-engelmann you fixed this with a `depth` param, yeah? Feel free to resolve.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
```diff
@@ -166,36 +168,381 @@ impl Entity for CallMCPToolExecutor {
...
-            continue;
+    }
+
+    // `oneOf` / `anyOf` — apply each branch the value could actually satisfy.
+    // We skip any branch the value provably violates (`const`, `enum`, `type`,
+    // `required`, or a discriminator `const`/`enum` on a present property), so
+    // we never coerce a value according to a branch it isn't governed by — e.g.
+    // a tagged union where one arm types a field as `integer` and another as
+    // `number`. For branches with no recognized discriminator the check is a
+    // no-op, preserving the prior "try them all" behavior (#10596 review).
```
vkodithala: Ditto; let's please get rid of anywhere we reference PR or review-specific context.

vkodithala (changes requested):
Hey David, thanks for the contribution. Left a few nits, avoided going any deeper since there's some fat to be trimmed from these tests. It's very weird to me that this implementation is ~300 LoC but tests are more than 2x that. Some obvious things we can reduce:

- `any_of_branch_integer_is_coerced` and `one_of_branch_integer_is_coerced` are redundant since in your implementation, these are coerced in the sample loop?
- Similar issue with `singleton_type_array_integer_is_coerced` and `nullable_integer_is_coerced`

Overall, I'd like for you to clean up these tests and re-request a review. Will get to it ASAP.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
david-engelmann: Done — scrubbed in 607a097a.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
david-engelmann: Lowered to 2_000 in 607a097a. Stays on the same order of magnitude as observed traffic (a few hundred ops per real schema) while keeping enough headroom for unusual but legitimate shapes. Tightened the doc comment to reflect the new rationale.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
david-engelmann: Done — scrubbed in 607a097a (and got the two test-file section headers that referenced #10596 too).

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
david-engelmann: Added an inline comment in 607a097a calling out the two cases: external `$ref` (we don't fetch remote schemas) and cycle detection (so chains like `A -> B -> A` terminate).

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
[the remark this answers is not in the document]
david-engelmann: Yep, that's resolved by the `depth` parameter threaded through `coerce_recursive` plus the `budget` op counter — adversarial branching now hits one of those two caps long before any single chain can do harm. Marking this resolved.

app/src/ai/blocklist/action_model/execute/call_mcp_tool.rs
david-engelmann: Done — scrubbed in 607a097a.

david-engelmann:
@vkodithala thanks for the review. Pushed `607a097a`:

**Tests trimmed** (per your callout — the four were redundant with broader coverage):
- `one_of_branch_integer_is_coerced` and `any_of_branch_integer_is_coerced` are covered by `tagged_union_coerces_through_matching_branch` plus the `branch_excluded` test family (which exercises the same iteration loop with discriminator-aware branch selection)
- `nullable_integer_is_coerced` and `singleton_type_array_integer_is_coerced` are covered by every other integer-coercion test — those all walk through the same `type`-array sample path. `null_value_for_nullable_integer_passes_through` stays since it tests the distinct null-passthrough behavior, not coercion.

Net: -68 lines of tests, 32 still cover the same surface (all green).

**Inline nits** all addressed in the same commit — replies threaded above.

Re-requesting review.

PR #11412, opened 2026-05-20
fix(tab-configs): let a focused picker own Space (#11138)

### LINKED ISSUE #11138: pressing space on tab params switches input field
### Pre-submit Checks

### Describe the bug

when creating a custom tab layout, that takes 2 parameters, a repository one and a text one. when pressing space when focussing on the text field, switches the focus to the repository field:

### To reproduce

1. create a tab config with the following layout:
```
[params.repo]                                                                                                                                          
type = "repo"
description = "Repository to open"

[params.tab_name]                                                                                                                                        
type = "text"
description = "Tab name"
```
2. enter text in the text params
3. press space (add a space to the textfield)

### Expected behavior

I should be able to define a tab name with spaces

https://github.com/user-attachments/assets/f576e0b6-b0d0-49c7-8e40-7c63f534d3b7

### Operating system (OS)

macOS

26.4.1 (25E253)

### Shell Version

### Current Warp version

v0.2026.05.13.09.15.stable_01

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

maxmilian:
## Summary

Fixes #11138. In a tab config that mixes a `repo` param with a `text` param, pressing **Space** while the text field was focused opened the repo dropdown instead of inserting a literal space, making multi-word text params impossible to enter.

This revision is rebased on top of #11718 and is **much smaller** than the earlier picker-/`FilterableDropdown`-layer implementation. Per [@moirahuang's guidance](https://github.com/warpdotdev/warp/pull/11412#discussion_r-11718-merged), now that #11718 has landed, `FilterableDropdown` owns Space itself when a picker is focused — so the tab config modal no longer needs to special-case Space at all. The fix is now a **deletion** (+10 / −50, single file).

## Root cause

`app/src/tab_configs/params_modal.rs` registered a modal-level Space binding with the scope predicate `id!("TabConfigParamsModal") & !id!("EditorView")`, on the assumption that the `!id!("EditorView")` clause would suppress the binding while a descendant text editor was focused.

But keystroke dispatch evaluates each layer's `ContextPredicate` against **that layer's own** `keymap_context` only — it never merges ancestors with descendants. When the focused view is `EditorView`:

1. At the `EditorView` layer the context contains `"EditorView"` but **not** `"TabConfigParamsModal"`, so the predicate is false. The editor has no Space `FixedBinding`, so dispatch continues up the chain.
2. At the `TabConfigParamsModal` layer the context contains `"TabConfigParamsModal"` but **never** `"EditorView"` (the modal isn't an editor), so `!id!("EditorView")` is always true. The predicate matches, `ToggleDropdown` fires, and the editor never sees the space.

## Fix

Since #11718, `FilterableDropdown` dispatches `ToggleExpanded` on an unmodified Space when the picker is focused (`app/src/view_components/filterable_dropdown.rs`). That covers the "focused dropdown owns Space" behavior natively, so the modal-level binding is redundant — and harmful, because of the dispatch semantics above. This PR removes:

- the modal-level `space` → `ToggleDropdown` `FixedBinding`,
- the now-unused `TabConfigParamsModalAction::ToggleDropdown` variant and its handler arm,
- the dead `dropdown_count` / `toggle_single_dropdown` helpers.

Result:
- **Text field focused** → Space is no longer intercepted by the modal → the editor inserts a literal space. (Fixes #11138.)
- **Picker focused** → `FilterableDropdown`'s built-in Space handler toggles it.

## Scope note

For a config whose **only** param is a dropdown, the modal still self-focuses on open (so the Enter-to-submit `FixedBinding` fires), which means Space toggles that dropdown only once it is focused, rather than immediately on open. Auto-focusing the single dropdown on open was [called out as optional by @moirahuang](https://github.com/warpdotdev/warp/pull/11412) and is intentionally left out of this PR, since #11138 concerns the mixed `repo` + `text` case.

## Visual evidence

Both clips use the mixed `repo` + `text` config below. The user-facing behavior is identical to the earlier revisions of this PR, so the recordings still apply.

**Before — stable Warp** (`v0.2026.05.18.05.32.stable_02`): with the `tab_name` text field focused, pressing **Space** opens the `repo` dropdown instead of inserting a space — issue #11138.

https://github.com/user-attachments/assets/f17c33ea-27db-424c-a783-3a89f32ad1d7

**After** (local OSS build): the focused, collapsed `repo` picker toggles open on **Space** ("the dropdown owns Space"); the `tab_name` text field accepts literal spaces (typed `my tab   New`); the `repo` dropdown stays closed throughout.

https://github.com/user-attachments/assets/1c73b097-2b62-4c38-b719-111a1ce9d561

Tab config used for the repro:

```toml
name = "PR 11412 repro"

[[panes]]
id = "main"
type = "terminal"
directory = "{{repo}}"
commands = []

[params.repo]
type = "repo"
description = "Repository to open"

[params.tab_name]
type = "text"
description = "Tab name"
default = "my tab"
```

## Test plan

- [x] `cargo fmt -p warp -- --check` clean
- [x] `cargo clippy -p warp --lib --no-deps` clean
- [x] `cargo test -p warp --lib tab_configs::params_modal` — 3/3 pass
- [ ] Reviewer environment: open the mixed `repo` + `text` tab config above, focus the text field, press Space → a literal space is inserted; focus the repo picker, press Space → the dropdown toggles.

### DIFF
--- a/app/src/tab_configs/params_modal.rs
+++ b/app/src/tab_configs/params_modal.rs
@@ -40,18 +40,14 @@ pub fn init(app: &mut AppContext) {
             TabConfigParamsModalAction::Escape,
             id!("TabConfigParamsModal"),
         ),
-        // Enter and Space only fire when no EditorView descendant is focused.
-        // When a text field or picker filter editor has focus, the editor
-        // consumes these keys and the modal handles them via event
-        // subscriptions instead (see handle_editor_event).
+        // Enter only fires when no EditorView descendant is focused. When a
+        // text field or picker filter editor has focus, the editor consumes
+        // it and the modal handles submit via event subscriptions instead
+        // (see handle_editor_event). Space is owned by FilterableDropdown
+        // itself when a picker is focused, so the modal no longer binds it.
         FixedBinding::new(
             "enter",
             TabConfigParamsModalAction::Submit,
-            id!("TabConfigParamsModal") & !id!("EditorView"),
-        ),
-        FixedBinding::new(
-            "space",
-            TabConfigParamsModalAction::ToggleDropdown,
             id!("TabConfigParamsModal") & !id!("EditorView"),
         ),
     ]);
@@ -147,7 +143,6 @@ pub enum TabConfigParamsModalAction {
     Cancel,
     Submit,
     Escape,
-    ToggleDropdown,
 }
 
 impl TabConfigParamsModal {
@@ -322,10 +317,9 @@ impl TabConfigParamsModal {
 
         self.pending_config = Some(config);
 
-        // When the only fields are dropdowns, focus the modal itself so
-        // Enter (submit) and Space (toggle dropdown) fixed bindings fire.
-        // When there are text fields, focus the first one so the user can
-        // start typing immediately.
+        // When the only fields are dropdowns, focus the modal itself so the
+        // Enter (submit) fixed binding fires. When there are text fields,
+        // focus the first one so the user can start typing immediately.
         if self.has_text_fields() {
             self.focus_field(0, ctx);
         } else {
@@ -403,35 +397,6 @@ impl TabConfigParamsModal {
             .any(|(_, _, field)| matches!(field, ParamField::Text(_)))
     }
 
-    fn dropdown_count(&self) -> usize {
-        self.param_fields
-            .iter()
-            .filter(|(_, _, field)| !matches!(field, ParamField::Text(_)))
-            .count()
-    }
-
-    fn toggle_single_dropdown(&mut self, ctx: &mut ViewContext<Self>) {
-        let mut opened = false;
-        for (_, _, field) in &self.param_fields {
-            match field {
-                ParamField::Branch { picker, .. } => {
-                    opened = picker.update(ctx, |p, ctx| p.toggle_dropdown(ctx));
-                    break;
-                }
-                ParamField::Repo { picker, .. } => {
-                    opened = picker.update(ctx, |p, ctx| p.toggle_dropdown(ctx));
-                    break;
-                }
-                ParamField::Text(_) => {}
-            }
-        }
-        // When the dropdown just closed, reclaim focus so Enter/Space
-        // fixed bindings continue to work.
-        if !opened && !self.has_text_fields() {
-            ctx.focus_self();
-        }
-    }
-
     fn focus_field(&self, index: usize, ctx: &mut ViewContext<Self>) {
         if let Some((_, _, field)) = self.param_fields.get(index) {
             match field {
@@ -503,8 +468,8 @@ impl View for TabConfigParamsModal {
 
     fn on_focus(&mut self, focus_ctx: &FocusContext, ctx: &mut ViewContext<Self>) {
         // When focus arrives directly at this view (not at a child), keep
-        // self-focus so the Enter/Space fixed bindings fire. This happens
-        // on initial open (via focus_self in on_open) and when the Modal
+        // self-focus so the Enter fixed binding fires. This happens on
+        // initial open (via focus_self in on_open) and when the Modal
         // wrapper re-focuses the body after a child (like a dropdown)
         // releases focus.
         if focus_ctx.is_self_focused() && !self.has_text_fields() {
@@ -727,11 +692,6 @@ impl TypedActionView for TabConfigParamsModal {
                 ctx.emit(TabConfigParamsModalEvent::Close);
             }
             TabConfigParamsModalAction::Submit => self.try_submit(ctx),
-            TabConfigParamsModalAction::ToggleDropdown => {
-                if self.dropdown_count() <= 1 {
-                    self.toggle_single_dropdown(ctx);
-                }
-            }
         }
     }
 }


moirahuang (commented):
sorry for the back and forth here, i think the fix we want is at the `FilterableDropdown` level. this will allow all `FilterableDropdown`s to use space to toggle the dropdown and can reduce some of the branch picker and modal specific code in this PR. i really appreciate your work on this issue, i think a `FilterableDropdown` fix will be the root cause!

maxmilian:
@moirahuang — addressed your review by moving the Space-to-toggle binding from the picker layer down to `FilterableDropdown` (commit `77c5718a`). The keymap-context flag is now `FilterableDropdownCollapsed`, advertised by `FilterableDropdown::keymap_context` only while `!is_expanded`, and the binding is registered via a new `filterable_dropdown::register_space_toggle_binding::<A>(app)` helper. Every `FilterableDropdown<A>` inherits the flag automatically; the binding itself is per-concrete-`A` because `DropdownAction` is generic.

Focus is wired through the pickers via `View::on_focus` — when the modal focuses a picker, focus flows through to the inner dropdown (and from there to its `filter_editor` via the dropdown's existing `on_focus`), so the responder chain becomes `[filter_editor, FilterableDropdown, picker, modal, ...]` and the dropdown's keymap layer is in scope without any picker-side glue.

Code drops vs. the prior revision:
- `RepoPickerAction::ToggleDropdown` and the whole `BranchPickerAction` enum gone (reverts to `String` action alias).
- Per-picker `*_COLLAPSED` flags, `keymap_context` overrides, `build_keymap_context` helpers, `toggle_dropdown` methods all removed.
- Two picker test files consolidated into one `filterable_dropdown::tests` module (6 tests → 3).

Net **-92 lines**. `cargo check` / `cargo clippy --tests -D warnings` / `cargo fmt --check` clean; 3 new unit tests + the 65 existing `tab_configs::` tests all pass.

Manually verified the original repro (mixed `repo` + `text` config) on a local OSS build — same behaviour as the previous revision's recording, which is reused in the PR description.

PR body is rewritten to describe the FilterableDropdown-layer approach. /oz-review

maxmilian:
Addressed Oz's Enter-to-submit concern (commit `13e422f2`).

**Root cause of the regression Oz flagged:** in the previous revision `FilterableDropdown::on_focus` unconditionally forwarded self-focus to its `filter_editor`, even when the dropdown was collapsed. That pulled a focused `EditorView` into the responder chain. The editor has a `enter → EditorAction::Enter` fixed binding scoped to `id!("EditorView")`, which fires leaf-first and short-circuits any ancestor Enter binding — including `TabConfigParamsModal`'s Enter-to-submit. The subscribed `EditorEvent::Enter` handler in `FilterableDropdown` then selected the currently highlighted item instead, so Enter on a dropdown-only modal silently picked the default value instead of submitting.

**Fix:** gate the filter-editor focus forward on `is_expanded`. When collapsed, focus stays on `FilterableDropdown` itself — no `EditorView` enters the chain, the modal's `enter → Submit` binding wins, dropdown-only modals submit on Enter as before. When expanded (user has explicitly opened the dropdown to type a filter query), focus moves to `filter_editor` and the existing `EditorEvent::Enter → select item` flow is preserved.

Extracted as a pure helper `should_forward_focus_to_filter_editor(is_expanded: bool) -> bool` so the per-state decision is unit-testable without a UI harness. Added a 4th unit test `on_focus_only_forwards_to_filter_editor_when_expanded`.

**Trace for the four user-visible cases:**

| Modal state | User action | Chain (leaf → root) | Matching binding | Result |
|---|---|---|---|---|
| Repo-only, picker collapsed | **Enter** | `[FilterableDropdown, RepoPicker, Modal]` | modal `enter → Submit` | submit ✓ |
| Repo-only, picker collapsed | **Space** | `[FilterableDropdown, RepoPicker, Modal]` | FilterableDropdown `space → ToggleExpanded` (CollapsedFlag matches) | open dropdown ✓ |
| Mixed repo + text, text field focused | **Space** | `[text EditorView, Modal]` | none — falls through | literal space ✓ |
| Repo dropdown expanded, filter editor focused | **Enter** | `[filter_editor, FilterableDropdown, RepoPicker, Modal]` | filter_editor `enter → EditorAction::Enter` | select item via existing `EditorEvent::Enter` subscription ✓ |

`cargo fmt --check`, `cargo clippy -p warp --lib --tests --no-deps -- -D warnings`, and `cargo nextest run -p warp 'view_components::filterable_dropdown::' 'tab_configs::'` all green (4 + 65 tests). Single squashed commit, force-pushed.

/oz-review

maxmilian:
Addressed Oz's third concern (commit `d6ed4e1f`).

**Confirmed Oz's read of the matcher is correct.** `Keymap::bindings()` returns fixed bindings via `self.fixed_bindings.iter().rev()` (`keymap.rs:454-460`), and `push_keystroke` returns the first binding whose `context_predicate` matches (`matcher.rs:320-331`). With both pickers' `init` registering Space scoped to the same generic `FilterableDropdownCollapsed` flag, the later-registered (BranchPicker, ordered after RepoPicker in `lib.rs:1617-1618`) always wins — its `DropdownAction::<String>::ToggleExpanded` dispatches even when a `FilterableDropdown<RepoPickerAction>` is focused, then no responder in the chain handles `DropdownAction<String>`, so Space silently no-ops on RepoPicker.

**Fix:** scope each `A`'s Space binding to a **per-`A`** flag derived from `TypeId::of::<A>()`:

```rust
pub fn filterable_dropdown_collapsed_flag<A: Action + Clone>() -> &'static str {
    static MAP: OnceLock<Mutex<HashMap<TypeId, &'static str>>> = OnceLock::new();
    let mut guard = MAP.get_or_init(|| Mutex::new(HashMap::new())).lock().unwrap();
    guard.entry(TypeId::of::<A>()).or_insert_with(|| {
        Box::leak(format!("FilterableDropdownCollapsed_{}", type_name::<A>()).into_boxed_str())
    })
}
```

The flag is interned in a process-wide `OnceLock<HashMap>` (one entry per concrete `A`) so the `&'static str` is stable and `keymap::Context`'s lifetime requirement is satisfied without re-leaking on every call. `register_space_toggle_binding::<A>` scopes the binding to `id!(filterable_dropdown_collapsed_flag::<A>())`; `FilterableDropdown<A>::keymap_context` advertises the same per-`A` flag. Result: each `A` only fires its own concrete `DropdownAction::<A>::ToggleExpanded`, the focused responder chain always has a handler, and adding a third `FilterableDropdown<C>` instantiation later won't reintroduce ambiguity.

**Tests** (`view_components::filterable_dropdown::tests`, 7 total — 4 new since previous revision):
- `per_action_flags_are_distinct` — two stand-in action types resolve to different flags
- `collapsed_flag_is_stable_per_action_type` — same `A` returns the same interned `&'static str` across calls
- `space_binding_predicate_does_not_match_other_action_type` — `TestAction`'s predicate misses `OtherTestAction`'s collapsed context (the exact regression Oz flagged)
- `space_binding_predicate_matches_only_collapsed_dropdown` — predicate hits collapsed, misses expanded (existing, ported to per-`A` flag)
- `keymap_context_advertises_collapsed_flag_when_closed` / `omits_collapsed_flag_when_expanded` — keymap-context advertisement (existing, ported)
- `on_focus_only_forwards_to_filter_editor_when_expanded` — Enter-regression guard (from previous revision)

`cargo fmt --check` / `cargo clippy -p warp --lib --tests --no-deps -- -D warnings` / `cargo nextest run -p warp 'view_components::filterable_dropdown::' 'tab_configs::'` all green (7 + 65 tests). Single squashed commit `d6ed4e1f`, force-pushed.

/oz-review

app/src/view_components/filterable_dropdown.rs
```diff
@@ -1,32 +1,106 @@
...
+/// collapsed `FilterableDropdown<A>` regardless of which `A` was actually
+/// focused, and the focused responder chain — which only handles its own
+/// `DropdownAction<A>` — would silently drop the action. Tying the flag to
+/// the concrete action `TypeId` guarantees exactly one matching binding per
+/// dispatch.
+///
+/// Returned `&'static str` is interned in a process-wide `OnceLock<HashMap>`
+/// (one entry per concrete `A`), so the lifetime requirement of
+/// `keymap::Context` is satisfied without leaking strings on every call.
+pub fn filterable_dropdown_collapsed_flag<A: Action + Clone>() -> &'static str {
```
moirahuang: larger architectural comment: the complexity of this function is a code smell. i talked to another team member @vorporeal and we think the right solution here is to refactor `FilterableDropdown` and `DropdownAction` to not take a generic. i have that PR here: https://github.com/warpdotdev/warp/pull/11718. once that is merged, it'll be cleaner to accomplish what you are trying to do in this PR!

moirahuang: this PR has been merged!

sorry for all of the back and forth here. i think your PR should now be simpler where we can update the tab config modal to not specially handle the space keypress toggling the dropdown since `FilterableDropdown` itself now supports that.

one optional thing to add is to have the tab config modal immediately focus on the dropdown if it is the only param. but i think this isn't strictly necessary for the issue you're trying to fix so it's up to you!

moirahuang (approved):
thank you for your contribution!

PR #9853, opened 2026-05-01
fix(mcp): show file-based MCP servers to the agent in fresh conversations (#9111)

david-engelmann:
## Description

Fixes #9111.

When an MCP server configured globally in `~/.warp/.mcp.json` was connected — green dot in Settings, tools listed — the Oz agent in a brand-new conversation could not see it. `list_relevant_mcp_context` returned no context.

### Root cause

In [`make_request_input`](app/src/ai/agent/api.rs), the file-based-MCP-servers branch was gated behind `if let Some(cwd) = session_context.current_working_directory()`. That accessor:

```rust
pub fn current_working_directory(&self) -> &Option<String> {
    self.task_store.all_exchanges_rev().find_map(|exchange| exchange.working_directory.clone())
}
```

walks prior exchanges to find a working directory. A fresh conversation has no exchanges, so cwd is `None`, the gate skips the entire file-based branch, and `get_active_file_based_servers` is never called. Globally-configured servers — which are already home-rooted candidates inside `get_servers_for_working_directory` — never make it into the agent's context.

The Settings UI uses a different read path with no cwd gate, which is why the same server shows as connected there. Two readers, divergent filtering — that's why the symptom looked like "agent ignores a server that's clearly connected."

### Fix

Thread `Option<&Path>` through `FileBasedMCPManager::get_servers_for_working_directory` and `TemplatableMCPServerManager::get_active_file_based_servers`. When `cwd` is `None`, return home-rooted (global) servers only; repo-scoped servers are still gated since they have no meaning without a cwd. The home-rooted candidate was already in `candidate_roots` inside the function — the change just lets the call go through.

Touches 4 production files (`api.rs`, `templatable_manager.rs`, `file_based_manager.rs`, `dummy_file_based_manager.rs`) and 1 test file. No other callers, no API breakage.

### Unit regression coverage

Three new tests in `app/src/ai/mcp/file_based_manager_tests.rs`:

- **`test_get_servers_for_working_directory_returns_home_servers_when_cwd_is_none`** — direct regression test for #9111. Seeds a global server at `dirs::home_dir()`, calls `get_servers_for_working_directory(None, app)`, asserts the global server is returned.
- **`test_get_servers_for_working_directory_excludes_repo_servers_when_cwd_is_none`** — guard. Seeds a server at a repo path (not home), asserts that path-rooted servers are NOT returned when cwd is `None`.
[...]

PR #11413, opened 2026-05-20
fix(mcp): surface MCP config unmarshal errors in Settings

### LINKED ISSUE #9807: Surface better errors when MCP configs fail to unmarshal
## Instructions for Warp Agent: Perform an investigation only, unless instructed otherwise. When investigating, do not take external actions like making Github comments or opening pull requests.

aloke said:

> I had an MCP server that wasn't starting and I could not figure out why.
>
> Ended up spending a ton of time debugging and turns out we require (?) an empty args field in each MCP definition for it to be valid. I've seen at least a few different MCP configs that omit args (see second screenshot) and it seems like an unnecessary requirement.
>
> Also separately, this just silently failed for me. We need better error messaging, I was only able to debug by actually getting agent mode to add debug logs to warp.

---
Linear: [APP-3075](https://linear.app/warpdotdev/issue/APP-3075/surface-better-errors-when-mcp-configs-fail-to-unmarshal)

david-engelmann:
Fixes #9807.

## Why

Before this change, all four parse-failure sites in [parse_mcp_config_file](app/src/ai/mcp/file_mcp_watcher.rs) collapsed to `safe_warn!` + `return vec![]`. The downstream `FileBasedMCPManager` applied an empty snapshot, the Settings → MCP Servers page showed no servers for the affected `(root_path, provider)` slot, and the user had no way to learn why short of adding debug logging.

Per the issue reporter:
> Ended up spending a ton of time debugging and turns out we require (?) an empty args field in each MCP definition for it to be valid. […] Also separately, this just silently failed for me. We need better error messaging, I was only able to debug by actually getting agent mode to add debug logs to warp.

## What

Carries each parse failure forward as a structured `ConfigParseError`:

```rust
pub struct ConfigParseError {
    pub path: PathBuf,
    pub provider: MCPProvider,
    pub stage: ConfigParseStage,
    pub message: String,
}

pub enum ConfigParseStage {
    Read,            // I/O failure (not NotFound — that stays \"normal\")
    TomlNormalize,   // Codex config TOML doesn't parse
    EnvSubstitute,   // \${VAR} not found in environment
    JsonParse,       // serde_json rejected the (resolved) config
}
```

- `parse_mcp_config_file` now returns `Result<Vec<…>, ConfigParseError>`. NotFound still returns `Ok(vec![])` (an unconfigured provider isn't a user error).
- `FileMCPWatcherEvent::ConfigParsed` carries `error: Option<ConfigParseError>` alongside `servers`. The split happens in a new `split_parse_result` helper so the two existing call sites stay symmetric.
- `FileBasedMCPManager` now stores `config_parse_errors: HashMap<(PathBuf, MCPProvider), ConfigParseError>`. On `ConfigParsed { error: Some(_) }` it inserts; on `ConfigParsed { error: None }` or `ConfigRemoved` it clears. Exposed via `config_parse_errors()`.
- `MCPServersListPage` renders one inline warning banner per current error above the server cards section, using the existing `ui_error_color` + `Icon::AlertTriangle` pattern from `billing_and_usage_page.rs`. The page already subscribes to `FileMCPWatcherEvent::ConfigParsed` and refreshes on emit, so the banner updates as the user edits and saves their config.

## Banner layout

```
⚠ Claude · /Users/me/.claude.json
   Invalid JSON: expected ',' or '}' at line 5 column 17

⚠ Codex · /Users/me/.codex/config.toml
   Invalid TOML: invalid table header at line 1
```

## Scope notes

The reporter raised two concerns. **Confirmed**: silent failure on config errors → fixed here. **Unverified**: claim that `args` is required for command-based MCP servers — Oz's triage [confirmed](https://github.com/warpdotdev/warp/issues/9807#issuecomment-4360287089) the current parser already accepts omitted `args` and the issue is in `needs-info` on that point with no reporter reply. Not touched in this PR.

Per-server validation drops (entries with `templatable_mcp_server_installation: None` at [file_based_manager.rs:174](app/src/ai/mcp/file_based_manager.rs#L174)) are a separate code path and a different UX surface (per-card vs file-level). Followup work.

## Test plan

- [x] `cargo build -p warp --lib` — clean
- [x] `cargo test -p warp --lib ai::mcp::file_mcp_watcher` — 8/8 pass (2 existing + 6 new)
- [x] `cargo test -p warp --lib ai::mcp::file_based_manager` — 12/12 pass (8 existing + 4 new)
- [x] `cargo test -p warp --lib ai::mcp::` — 64/64 pass overall
- [x] `cargo clippy -p warp --lib --tests -- -D warnings` — clean
- [x] `cargo fmt -p warp -- --check` — clean

New test coverage:

Watcher (`parse_mcp_config_file`):
- `parse_missing_file_is_not_an_error` — NotFound stays `Ok(vec![])`
- `parse_valid_json_returns_servers_without_error` — happy path
- `parse_unreadable_path_emits_read_stage_error` — Read stage
- `parse_invalid_codex_toml_emits_toml_normalize_error` — TomlNormalize stage
- `parse_missing_env_var_emits_env_substitute_error` — EnvSubstitute stage
- `parse_malformed_json_emits_json_parse_error` — JsonParse stage

Manager (`FileBasedMCPManager::handle_watcher_event`):
- `config_parse_error_is_recorded_on_failed_parse`
- `successful_parse_clears_prior_config_parse_error`
- `config_removed_clears_config_parse_error`
- `config_parse_errors_are_keyed_per_root_and_provider`

### DIFF
--- a/app/src/ai/mcp/dummy_file_based_manager.rs
+++ b/app/src/ai/mcp/dummy_file_based_manager.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 use uuid::Uuid;
 use warpui::{AppContext, Entity, ModelContext, SingletonEntity};
 
-use super::MCPProvider;
+use super::{ConfigParseError, MCPProvider};
 use crate::ai::mcp::templatable_installation::TemplatableMCPServerInstallation;
 
 pub struct FileBasedMCPManager {}
@@ -19,6 +19,10 @@ impl FileBasedMCPManager {
         _app: &AppContext,
     ) -> Vec<&TemplatableMCPServerInstallation> {
         vec![]
+    }
+
+    pub fn config_parse_errors(&self) -> std::iter::Empty<&ConfigParseError> {
+        std::iter::empty()
     }
 
     pub fn file_based_servers(&self) -> Vec<&TemplatableMCPServerInstallation> {
--- a/app/src/ai/mcp/file_based_manager.rs
+++ b/app/src/ai/mcp/file_based_manager.rs
@@ -9,7 +9,7 @@ use warp_core::features::FeatureFlag;
 use warp_util::local_or_remote_path::LocalOrRemotePath;
 use warpui::{AppContext, Entity, ModelContext, SingletonEntity};
 
-use super::{FileMCPWatcher, FileMCPWatcherEvent, MCPProvider};
+use super::{ConfigParseError, FileMCPWatcher, FileMCPWatcherEvent, MCPProvider};
 use crate::ai::mcp::templatable_installation::TemplatableMCPServerInstallation;
 use crate::ai::mcp::ParsedTemplatableMCPServerResult;
 use crate::settings::ai::AISettings;
@@ -28,6 +28,11 @@ pub struct FileBasedMCPManager {
     /// They are temporarily stored here and removed to emit FileBasedMCPManagerEvent::CloudEnvMcpScanComplete
     pending_scan_auto_started_servers_by_root:
         HashMap<PathBuf, HashMap<MCPProvider, HashSet<Uuid>>>,
+    /// Last unmarshal failure observed for each `(root_path, provider)` config
+    /// slot. Surfaced by the settings UI so the user sees why a config that
+    /// exists on disk produced no servers, instead of debugging via logs (see
+    /// issue #9807).
+    config_parse_errors: HashMap<(PathBuf, MCPProvider), ConfigParseError>,
 }
 
 impl FileBasedMCPManager {
@@ -48,6 +53,7 @@ impl FileBasedMCPManager {
             file_based_servers: Default::default(),
             file_based_servers_by_root: Default::default(),
             pending_scan_auto_started_servers_by_root: Default::default(),
+            config_parse_errors: Default::default(),
         }
     }
 
@@ -58,14 +64,26 @@ impl FileBasedMCPManager {
                 root_path,
                 provider,
                 servers,
+                error,
             } => {
                 self.apply_parsed_servers(root_path.clone(), *provider, servers.clone(), ctx);
+                let key = (root_path.clone(), *provider);
+                match error {
+                    Some(err) => {
+                        self.config_parse_errors.insert(key, err.clone());
+                    }
+                    None => {
+                        self.config_parse_errors.remove(&key);
+                    }
+                }
             }
             FileMCPWatcherEvent::ConfigRemoved {
                 root_path,
                 provider,
             } => {
                 self.remove_servers_for_root_provider(root_path, *provider, ctx);
+                self.config_parse_errors
+                    .remove(&(root_path.clone(), *provider));
             }
             FileMCPWatcherEvent::CloudEnvMcpScanComplete { repo_path } => {
                 self.handle_cloud_environment_scan_complete(repo_path, ctx);
@@ -98,6 +116,14 @@ impl FileBasedMCPManager {
             }
         }
         servers
+    }
+
+    /// Returns the most recent config-parse failures across all detected
+    /// `(root_path, provider)` slots. Surfaced by Settings → MCP Servers so the
+    /// user sees a structured reason when a config that exists on disk produced
+    /// no servers.
+    pub fn config_parse_errors(&self) -> impl Iterator<Item = &ConfigParseError> {
+        self.config_parse_errors.values()
     }
 
     /// Removes all tracked servers for the given `(root_path, provider)` pair,
--- a/app/src/ai/mcp/file_based_manager_tests.rs
+++ b/app/src/ai/mcp/file_based_manager_tests.rs
@@ -11,7 +11,10 @@ use warpui::{App, Entity, ModelHandle, SingletonEntity as _};
 use watcher::HomeDirectoryWatcher;
 
 use super::{CloudEnvMcpScanServer, FileBasedMCPManager, FileBasedMCPManagerEvent, MCPProvider};
-use crate::ai::mcp::{FileMCPWatcher, ParsedTemplatableMCPServerResult};
+use crate::ai::mcp::{
+    ConfigParseError, ConfigParseStage, FileMCPWatcher, FileMCPWatcherEvent,
+    ParsedTemplatableMCPServerResult,
+};
 use crate::auth::AuthStateProvider;
 use crate::settings::{AISettings, FocusedTerminalInfo};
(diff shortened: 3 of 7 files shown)


david-engelmann:
Heads up: this PR is sitting in CHANGES_REQUESTED because oz-for-oss flagged that linked issue #9807 is labeled `needs-info` rather than `ready-to-implement`. The PR itself implements the error-surfacing behavior oz's own triage comment described as actionable; just needs the readiness label flipped on #9807 (or for a reviewer to pull it in over the bot's gate) before a human review can land. Happy to provide whatever additional context the issue needs — let me know what's missing.

acarl005:
@david-engelmann Thanks for flagging. I've added `ready-to-implement`. I'll kick off another review.

acarl005:
@david-engelmann  Please be sure to include a screenshot of the change when it's done and specific steps for the reviewer to perform in order to test the changed behavior.

david-engelmann:
Thanks @acarl005 for the label flip + @oz-for-oss for the re-review. Both concerns addressed in `1e3f74fb`:

### 1. Visual evidence of the banner

11413-banner

Captured against warp-oss built from this PR with `~/.warp-oss/.mcp.json` containing `${UNDEFINED_ENV_VAR_FOR_PARSE_ERROR_DEMO}` (intentionally unresolved). The banner renders above the controls section:

- Header: `Warp · /Users/david/.warp-oss/.mcp.json` (provider · path)
- AlertTriangle icon in `ui_error_color`
- Body: `Missing or empty environment variable: UNDEFINED_ENV_VAR_FOR_PARSE_ERROR_DEMO`

Still frame: https://github.com/david-engelmann/warp-taper/blob/main/docs/sample-tape/11413-banner.png

Recipe: [scripts/recipes/11413-mcp-parse-error-banner.json](https://github.com/david-engelmann/warp-taper/blob/main/scripts/recipes/11413-mcp-parse-error-banner.json) — same warp-taper pipeline as the #9853 and #11457 evidence.

### 2. Reactivity fix — `refresh_file_based_server_cards` now notifies

Diff:

```rust
fn refresh_file_based_server_cards(&mut self, ctx: &mut ViewContext<Self>) {
    self.create_file_based_server_cards(ctx);
+   // Notify unconditionally so the parse-error banner re-renders even on
+   // ConfigParsed / ConfigRemoved events that don't create or clear any
+   // server cards (e.g. the first parse failure on a previously-empty
+   // config). Other call sites already notify via adjacent state changes,
+   // but the watcher path has no such side-effect.
+   ctx.notify();
}
```

Centralized the notify here rather than at each subscriber call site — covers both edge cases you flagged in the inline comment:

- **First parse failure on a previously-empty config** — `config_parse_errors` gets populated but no card is created/cleared, so no other reactive path fires.
- **Recovery clearing the error map** — the error entry is removed; without an explicit notify the banner could persist visually until the next unrelated state change.

Other callers (`FileBasedMCPManagerEvent::SpawnServers/DespawnServers/PurgeCredentials`, `AISettingsChangedEvent::FileBasedMcpEnabled`) already notify via adjacent state mutations they trigger, so the extra `ctx.notify()` here is at worst redundant for those paths — no spurious renders observed in the recording.

Subscriber ordering verified: `FileBasedMCPManager` registers its `FileMCPWatcher` subscription at app init (singleton model), and `list_page` subscribes when Settings is opened (later). So on `ConfigParsed`, the manager's handler runs first → `config_parse_errors` map updated → list_page's handler fires → `refresh_file_based_server_cards` → `notify()` → re-render reads the updated map.

PR is rebased onto current master.

david-engelmann:
@acarl005 explicit reviewer repro steps you asked for. The GIF above shows the end state; this is the manual path so you can verify on your own machine:

### Reproduce the parse-error banner (positive case)

1. **Set up a broken MCP config.** Write the following to `~/.warp/.mcp.json` (or `~/.warp-oss/.mcp.json` if you're testing on warp-oss):
   ```json
   {
     "broken-server": {
       "command": "/usr/bin/missing-binary",
       "args": [],
       "env": {
         "API_KEY": "${UNDEFINED_ENV_VAR_FOR_PARSE_ERROR_DEMO}"
       },
       "start_on_launch": true,
       "working_directory": null
     }
   }
   ```
   The `${UNDEFINED_ENV_VAR_…}` reference triggers `ConfigParseStage::EnvSubstitute`. If you'd rather hit a different parse stage:
   - `Read` — `chmod 000 ~/.warp/.mcp.json` after writing
   - `JsonParse` — drop a stray `,` before the closing `}`
   - `TomlNormalize` — (only fires for Codex's config, not Warp's)

2. **Launch Warp** (or restart if already running). `FileMCPWatcher` fires `ConfigParsed { error: Some(...) }` for the broken file.

3. **Open Settings → MCP Servers** (`Cmd+,` → search "MCP" → click "MCP servers" tab).

4. **Expected:** banner renders above the controls, in `ui_error_color`, with:
   - Header: `Warp · /Users/<you>/.warp/.mcp.json` (provider · path)
   - Body: `Missing required environment variable: Missing or empty environment variable: UNDEFINED_ENV_VAR_FOR_PARSE_ERROR_DEMO`
   - `AlertTriangle` icon to the left.

### Verify the notify fix specifically (oz concern #2)

The 2026-06-10 oz review flagged that the banner *"won't reliably update when a parse failure only changes `config_parse_errors`"* — i.e. the watcher path didn't notify when no card was created or cleared. To exercise that path:

1. **Start with an empty `~/.warp/.mcp.json`** (`{}`). Open Settings → MCP Servers — no banner, no cards.
2. **With Settings still open**, edit `~/.warp/.mcp.json` to introduce the broken config from step 1 above (save the file from your editor).
3. **Expected (with the fix in `1e3f74fb`):** banner appears within ~500ms (the `FileMCPWatcher` debounce), even though no server card was ever created or cleared. Before the fix, `refresh_file_based_server_cards` rebuilt the (still-empty) card map without calling `ctx.notify()`, so the banner stayed hidden until some unrelated re-render forced it.
4. **Reverse:** delete the broken config so the file is `{}` again. Banner disappears on the next debounce.

### Cleanup

```bash
rm ~/.warp/.mcp.json   # or restore your real config
```

Happy to record an explicit before/after recording of the appear-on-edit / disappear-on-cleanup transition if useful — the GIF above is the single steady-state capture.

david-engelmann:
@acarl005 — `1e3f74fb` covered both of oz's concerns (the `refresh_file_based_server_cards` notify path + the GIF), and the 06-11 follow-up has the explicit reviewer steps you asked for. Stale-bot's now showing the final 3-day warning. Mind retriggering `/oz-review` or dropping `no-autoclose` so this doesn't time out?

PR #11418, opened 2026-05-20
Fix multiple connection across tabs causes 'unable to load content'

kevinyang372:
### Problem
Opening code review in a second tab for the same remote SSH repo causes ALL diff editors across both tabs to show "Unable to load file content".

### Architecture Context
`DiffStateModel` is shared across tabs — `WorkingDirectoriesModel` (one per window) deduplicates models by `LocalOrRemotePath`. Since multiple SSH sessions to the same host share the same `HostId`, the `RemotePath` key is identical, so all tabs share one `DiffStateModel`. `CodeReviewView` instances, however, are per-tab (keyed by `(pane_group_id, repo_path)`).

Within the **same tab**, multiple panes to the same host/repo share both the `CodeReviewView` and `DiffStateModel`, and `on_open` returns early via the `is_open` guard — so this issue does NOT occur in the same-tab case.

### Root Cause
When Tab 2 creates a new `CodeReviewView` against the shared (already-`Loaded`) `DiffStateModel`, `on_open` detects `view_needs_diffs=true` (fresh view) + `model_already_loaded=true` → `force=true` → `replay_latest_diffs()`.

`replay_latest_diffs` reconstructed `GitDiffWithBaseContent` from cached `GitDiffData` via `From<&GitDiffData>`, which hard-codes `content_at_head: None` for every file. Both tabs' views subscribe to the shared model, so both receive this content-less event and rebuild with no editor state.

### Explored Solutions

**Option A: Store full `GitDiffWithBaseContent` in `InternalRemoteDiffState`** — preserves content across replays but stores full file content in memory for every changed file. Rejected due to memory cost.

**Option B: Per-tab `DiffStateModel` for remote repos** — key `DiffStateModelMap` by `(pane_group_id, LocalOrRemotePath)` for remote repos, giving each tab its own model and session. Eliminates cross-tab interference entirely and fixes session anchoring. Rejected as over-scoped for this fix: local models should remain shared (one file watcher per repo), creating an asymmetry, and duplicate server-side subscriptions add complexity. Could be revisited later.

**Option C (chosen): Replace `replay_latest_diffs` with a real `GetDiffState` RPC** — sends an actual request to the remote server, which already handles this correctly: `RemoteDiffStateManager::subscribe` spawns a content reload (`spawn_content_reload` → `load_diffs_with_content_for_mode`) and responds **only to the requesting connection** via `resolve_pending_responses`. Content reloads are coalesced server-side (`if !already_in_flight`).

### Why replacing `replay_latest_diffs` is safe
`replay_latest_diffs` was added to fix a race where a new `CodeReviewView` against an already-`Loaded` shared `DiffStateModel` would be permanently stuck in loading. The underlying issue was that no `NewDiffsComputed` event would fire for the new view. That fix is still needed — but the **mechanism** was wrong: it replayed from stripped cached data instead of asking the server for fresh data with content. The new `fetch_fresh_snapshot` method sends a `GetDiffState` RPC without transitioning to `Loading` (avoiding a flash on Tab 1), and the server response flows through `apply_snapshot` normally.

Trade-off: Tab 1's view gets a redundant `invalidate_all` rebuild when the server response arrives (since both views subscribe to the shared model). But the data has content, so the rebuild is correct — consistent with how local `force=true` handles the same pattern.

## Linked Issue
- [x] N/A (reported via internal feedback)

## Testing
- Verified fix by opening two tabs SSH'd to the same remote host/repo and confirming both tabs' code review panels load correctly
- Ran `cargo nextest run -- code_review::diff_state::remote` — all 22 tests pass

CHANGELOG-BUG-FIX: Fixed code review panel showing "Unable to load file content" when opening the same remote SSH repo in multiple tabs.

### DIFF
--- a/app/src/code_review/code_review_view.rs
+++ b/app/src/code_review/code_review_view.rs
@@ -764,17 +764,14 @@ impl CodeReviewView {
             }
         }
 
-        // If the view is freshly created (state is None/loading) but the model
-        // already holds computed diffs (DiffState::Loaded), a non-forced load
-        // will no-op because the model thinks it's done. Force a reload so
-        // the model re-emits NewDiffsComputed and the view gets populated.
-        let view_needs_diffs = matches!(self.state(), CodeReviewViewState::None);
-        let model_already_loaded = matches!(self.diff_state(ctx), DiffState::Loaded);
-        let force = view_needs_diffs && model_already_loaded;
-
+        // Always reload diffs on open. For local, this re-reads the
+        // filesystem. For remote, this re-emits the model's current state
+        // (and will make an RPC once that path is wired). We pass
+        // should_fetch_base: false because re-opening the panel doesn't
+        // need to fetch the base branch from origin.
         self.diff_state_model.update(ctx, |model, ctx| {
             model.set_code_review_metadata_refresh_enabled(true, ctx);
-            model.load_diffs_for_current_repo(force, ctx);
+            model.load_diffs_for_current_repo(false, ctx);
         });
     }
 
--- a/app/src/code_review/diff_state/mod.rs
+++ b/app/src/code_review/diff_state/mod.rs
@@ -537,7 +537,7 @@ impl DiffStateModel {
             Self::Remote(remote) => {
                 if force {
                     remote.update(ctx, |remote, ctx| {
-                        remote.replay_latest_diffs(ctx);
+                        remote.fetch_fresh_snapshot(ctx);
                     });
                 }
             }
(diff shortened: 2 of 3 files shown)


kevinyang372:
* **#11418** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11418?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11418?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

MaggieShan (approved):

PR #11420, opened 2026-05-20
Allow AI for all plans provided they have autoreload enabled

tylerlam-warp:
## Description
I noticed a bug where we were preventing AI usage if the user had no credits remaining even if they had autoreload enabled on the client.

Left a comment in the code

## Testing
Confirmed that a Build plan non-admin can trigger autoreload if they have no credits remaining.

## Agent Mode


app/src/ai/request_usage_model.rs
```diff
@@ -375,7 +375,7 @@ impl AIRequestUsageModel {
     /// 2. user has overage enabled
     /// 3. user has bonus grants (either team grants or user grants)
     /// 4. user's team plan has pay-as-you-go enabled (enterprise only)
-    /// 5. user's team is on enterprise with bonus grants auto-reload enable (enterprise only)
+    /// 5. user's team has bonus grants auto-reload enabled
```
tylerlam-warp: I'm not exactly sure why this was enterprise only in the past but I think this was much less of a issue in the past because a admin can buy team scoped credits and the whole team can drain from that, so a non-admin trying to autoreload was less likely? But now that non-admins have user scoped credits, all non-admins will likely need to autoreload at some point

IsaiahWitzke (approved):

PR #11426, opened 2026-05-20
View attached images in agent.

szgupta:
Once you attach an image to an agent conversation, you can't view them. This PR makes that possible - clicking on a submitted image (via the attachment chip) will open a lightbox view of the submitted images. Works both pre- and post-submission.

https://www.loom.com/share/3e20c2f8f8364c5595236803fdb78e56

### DIFF
--- a/app/src/ai/blocklist/block.rs
+++ b/app/src/ai/blocklist/block.rs
@@ -6607,35 +6607,42 @@ impl TypedActionView for AIBlock {
                 });
             }
             AIBlockAction::OpenSubmittedAttachmentLightbox { image_index } => {
-                let submitted_images = self
+                let decoded_images = self
                     .model
                     .inputs_to_render(ctx)
                     .iter()
                     .filter_map(|input| input.context())
                     .flat_map(|contexts| contexts.iter())
                     .filter_map(|context| match context {
-                        AIAgentContext::Image(image) => Some(image.clone()),
+                        AIAgentContext::Image(image) => Some(image),
                         _ => None,
+                    })
+                    .enumerate()
+                    .filter_map(|(submitted_image_index, image)| {
+                        let image_bytes =
+                            match base64::engine::general_purpose::STANDARD.decode(&image.data) {
+                                Ok(image_bytes) => image_bytes,
+                                Err(error) => {
+                                    log::warn!(
+                                        "Failed to decode submitted image attachment for lightbox: {error}"
+                                    );
+                                    return None;
+                                }
+                            };
+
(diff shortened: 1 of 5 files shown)


app/src/terminal/input.rs
```diff
@@ -14538,8 +14540,21 @@ impl Input {
...
         ))
         .with_close_button(close_button)
-        .build()
-        .finish()
+        .build();
+
+        if matches!(chip.attachment_type, AttachmentType::Image) {
+            let preview_chip_index = chip.index;
+            EventHandler::new(attachment_chip.finish())
+                .on_left_mouse_down(move |ctx, _, _| {
```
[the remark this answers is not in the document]
szgupta: This works fine because the close button is deeper in the element tree so gets first dibs at handling the event.

app/src/ai/blocklist/block.rs
```diff
@@ -6595,6 +6601,66 @@ impl TypedActionView for AIBlock {
...
+                                continue;
+                            }
+                        };
+
+                    let asset_id = format!(
+                        "submitted-attachment-lightbox-{}-{submitted_image_index}",
+                        self.client_ids.client_exchange_id
+                    );
+                    AssetCache::handle(ctx).update(ctx, |asset_cache, ctx| {
+                        asset_cache.insert_raw_asset_bytes::<ImageType>(
```
alokedesai, line 6647: Does this get automatically evicted or do we need to manually do it?

app/src/ai/blocklist/block.rs
```diff
@@ -6595,6 +6601,66 @@ impl TypedActionView for AIBlock {
...
+            }
+            AIBlockAction::OpenSubmittedAttachmentLightbox { image_index } => {
+                let submitted_images = self
+                    .model
+                    .inputs_to_render(ctx)
+                    .iter()
+                    .filter_map(|input| input.context())
+                    .flat_map(|contexts| contexts.iter())
+                    .filter_map(|context| match context {
+                        AIAgentContext::Image(image) => Some(image.clone()),
```
alokedesai: Any reason not to reference count the image data? Seems kinda crazy that we clone the entire image data every time we try to view an image

alokedesai (approved):

app/src/ai/blocklist/block.rs
szgupta: There's a hard cap. Added a comment

app/src/ai/blocklist/block.rs
szgupta: Avoided the clone entirely by decoding in the iteration.

PR #11428, opened 2026-05-20
Fix deadlock in secret redaction due to mutex ordering mismatch.

vorporeal:
## Description

This fixes a potential deadlock in secret redaction caused by inconsistent mutex acquisition order between the read and write paths. Previously, the secret-redaction DFA, meta regex, and regex-level metadata were stored behind separate locks. If a reader acquired one lock at the same time that a writer acquired another, each path could then block forever waiting for the lock held by the other path.

The secret-redaction state is now stored as a single `SecretsRegex` snapshot containing the meta regex, grid-search DFAs, and level metadata. That snapshot is wrapped in an `Arc`, and the global state only stores the current `Arc`. Readers clone the current `Arc` under a very short critical section, then use that same snapshot for both DFA matching and level classification. Writers compile the replacement meta regex and DFAs before touching global state, then swap in one new `Arc` only after both structures have been built successfully.

This gives us a few stronger guarantees:

- There is only one global lock involved in publishing or reading secret-redaction state, so the original lock-order deadlock is no longer possible.
- If either the meta regex or DFA compilation fails, the current secret-redaction state is left unchanged, matching the function's documented behavior.
- Grid redaction uses one generation-consistent snapshot for both match detection and secret-level lookup, avoiding mixed old/new regex state during updates.
- Regex compilation no longer happens while holding the global lock, improving responsiveness while secret-redaction settings are refreshed.

The global lock is now a `Mutex` rather than an `RwLock`. Since readers only hold the lock long enough to clone an `Arc`, the critical section is extremely short, and the simpler/faster synchronization primitive is a better fit than reader/writer locking.

## Testing

- Ran `cargo fmt` via `cf`.

### DIFF
--- a/app/src/ai/blocklist/block/secret_redaction.rs
+++ b/app/src/ai/blocklist/block/secret_redaction.rs
@@ -6,7 +6,7 @@ use warpui::elements::{MouseStateHandle, PartialClickableElement, SecretRange};
 use warpui::platform::Cursor;
 
 use crate::ai::agent::{AIAgentOutput, AIAgentTextSection, AgentOutputText};
-use crate::terminal::model::secrets::{SecretLevel, REGEX_LEVEL_METADATA, SECRETS_REGEX};
+use crate::terminal::model::secrets::{SecretLevel, SecretsRegex, SECRETS_REGEX};
 
 use super::{AIBlockAction, TextLocation};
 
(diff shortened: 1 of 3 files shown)


vorporeal:
* **#11428** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11428?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

moirahuang (approved):

PR #11444, opened 2026-05-20
ci(specs): add validate_specs.sh script and docs references

### LINKED ISSUE #11350: Entry::build_tree walks are not cancellable — closing the file tree view leaves CPU spike running until in-flight walks finish
## Summary

`repo_metadata::entry::Entry::build_tree` recursive walks are not cancellable. When the file tree view becomes inactive and drains its `registered_lazy_loaded_paths`, the unregistration tears down the watcher but does **not** interrupt walks already in flight on background-executor threads. Those walks continue to completion (or until they hit `MAX_FILES_PER_REPO`), wasting CPU on a tree no subscriber is listening to anymore.

This is a follow-up to #11349 (file-tree-view recursive-watcher CPU issue). The fixes there (failed-walk cache + system-dir exclude list) reduce how often `Entry::build_tree` is invoked. This issue is about making invocations that are already running cancellable.

## Observed behavior

After identifying the file tree view as the watcher registrant (sibling issue), the user-side workaround is "close the file tree view sidebar". Empirically confirmed on Windows 11 against a locally-built `warp-oss.exe` (upstream `master` at `b376889`, plus one unrelated `log::warn!` line in `orchestration_event_streamer.rs` that does not touch `repo_metadata`):

1. With the file tree view rooted at `C:\Users\<user>`, CPU spikes to ~100% across `background-executor-*` threads.
2. Closing the file tree view → CPU **does not drop immediately**. It takes ~10 seconds for the spike to subside.
3. The CPU drop coincides with `ExceededMaxFileLimit` failure log messages for the offending subtree, indicating that the in-flight walks ran to completion (hit the `MAX_FILES_PER_REPO` quota) before the executor threads went idle.

This is the smoking gun for a missing cancellation path.

## Root cause

`Entry::build_tree` (`crates/repo_metadata/src/entry.rs:130-227`) takes `&mut Option<u32>` for a file budget and otherwise runs to completion. It has no `CancellationToken`, no `AtomicBool` stop-flag, and no other mechanism to bail between directory entries.

The view-side teardown at `app/src/code/file_tree/view.rs:388-398`:

```rust
ctx.unsubscribe_to_model(&self.repository_metadata_model);
self.unsubscribe_from_active_file_model(ctx);
let repository_metadata_model = self.repository_metadata_model.clone();
let paths: Vec<_> = self.registered_lazy_loaded_paths.drain().collect();
repository_metadata_model.update(ctx, move |model: &mut RepoMetadataModel, ctx| {
    for path in &paths {
        model.remove_lazy_loaded_path(path, ctx);
    }
});
```

…calls `remove_lazy_loaded_path` for each drained path. That descends (via refcount) into `remove_repository` (`crates/repo_metadata/src/local_model.rs:428`) which calls `watcher.unregister_path`. The watcher stops emitting new events, but any `Entry::build_tree` task already enqueued or in flight on a background executor keeps walking.

For a home-directory-rooted view that has already kicked off a walk into `AppData\Local\Temp` (or any large subtree), this means up to `MAX_FILES_PER_REPO = 100_000` files of `lstat` + `stat` + `dunce::canonicalize` per-entry work has to finish before the executor thread frees up.

## Proposed fix

Plumb a cancellation token through `Entry::build_tree` so it can bail between directory entries when its owning subscriber goes away.

Sketch (subject to taste — `CancellationToken` from `tokio-util`, `Arc<AtomicBool>`, or whatever pattern matches `repo_metadata`'s existing async-task abstraction):

```rust
pub fn build_tree(
    root: &Path,
    files: &mut Vec<FileMetadata>,
    gitignores: &mut Vec<...>,
    file_budget: Option<&mut u32>,
    max_depth: usize,
    current_depth: usize,
    ignored: &IgnoredPathStrategy,
    cancel: &CancellationToken,  // <-- new
) -> Result<Entry, BuildTreeError> {
    if cancel.is_cancelled() {
        return Err(BuildTreeError::Cancelled);
    }
    // ... existing per-entry loop, with cancel check at the top of each iteration
}
```

At call sites in `compute_file_tree_mutations`, hand the task a `CancellationToken` cloned from a token owned by the `LocalRepoMetadataModel` entry for that path. When `remove_repository` runs (or when a refcount-driven `remove_lazy_loaded_path` reaches zero), `cancel()` the token **before** calling `unregister_path` so the executor thread observes cancellation on its next iteration.

This bounds the wasted work after teardown to "at most one directory's worth of `read_dir` + per-entry stats", not "the whole subtree up to `MAX_FILES_PER_REPO`".

## Why this is filed separately from A + B

The two fixes in the sibling issue (failed-walk cache + system-dir exclude list) are small, localized, and defensive — they reduce *how often* `Entry::build_tree` is invoked. Cancellation is a structural change to the walker's signature and to every call site, and the right shape depends on conventions the Warp team already uses elsewhere in `repo_metadata`. Splitting it out keeps the first PR landable while this discussion happens.

## Tests to add

- An in-flight `build_tree` walk observes cancellation and returns early when its owning subscriber unregisters.
- The total CPU spent by an executor thread after `remove_lazy_loaded_path` is bounded by a single directory's stat cost, not by the full subtree size.

## Related

- #11349 — sibling A+B issue (failed-walk cache + system-dir exclude list)
- #7561 — umbrella "idle CPU on Windows" report
- #8471 — debouncing in `repo_metadata/watcher.rs` (complementary)

---

🤖 _Generated by Claude Code on behalf of @cbeaulieu-gt_

Kartalops:
## Summary

Add script/validate_specs.sh — a CI-ready validation script that checks spec directories are well-formed.

The script verifies every subdirectory with an issue-number prefix contains at least one product spec and one tech spec file (case-insensitive). Contributor-named directories are skipped.

## Why this matters

Contributors sometimes open spec PRs with incomplete directories — missing PRODUCT.md or TECH.md files. This adds a machine-checkable way to catch these before review.

## Changes

- Add script/validate_specs.sh with exit-code validation
- Reference validate_specs.sh in CONTRIBUTING.md Opening a Spec PR section
- CI runs validate_specs.sh on every PR

## Validation

bash
./script/validate_specs.sh
exits 0 if all spec dirs are complete, 1 if any are missing product/tech files

## Risk/Compatibility

No runtime behavior changes. Only adds a CI/build-time check.

---

Closes #11350 (Entry::build_tree walks are not cancellable)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,7 +94,7 @@ Issues labeled `ready-to-spec` need a spec before code can begin. A spec consist
 
 To open a spec PR:
 
-1. Add `specs/GH<issue-number>/product.md` and `specs/GH<issue-number>/tech.md`. See [`specs/GH408/`](specs/GH408/), [`specs/GH1063/`](specs/GH1063/), and [`specs/GH1066/`](specs/GH1066/) for examples of well-structured specs, and browse the rest of [`specs/`](specs/) for more. The [`/write-product-spec`](.agents/skills/write-product-spec/SKILL.md) and [`/write-tech-spec`](.agents/skills/write-tech-spec/SKILL.md) skills are available to scaffold these for you.
+1. Add `specs/GH<issue-number>/product.md` and `specs/GH<issue-number>/tech.md`. See [`specs/GH408/`](specs/GH408/), [`specs/GH1063/`](specs/GH1063/), and [`specs/GH1066/`](specs/GH1066/) for examples of well-structured specs, and browse the rest of [`specs/`](specs/) for more. The [`/write-product-spec`](.agents/skills/write-product-spec/SKILL.md) and [`/write-tech-spec`](.agents/skills/write-tech-spec/SKILL.md) skills are available to scaffold these for you. Run [`./script/validate_specs.sh`](https://github.com/warpdotdev/warp/blob/master/script/validate_specs.sh) to check your spec is well-formed before opening the PR.
 2. Use the PR as the home for product and technical discussion.
 3. Once the specs are approved, implementation generally continues on the same PR. In rarer cases — for example, if a large spec is merged on its own so the implementation can be broken up — it can move to a linked follow-up PR.
 
--- a/WARP.md
+++ b/WARP.md
@@ -36,6 +36,21 @@ Environment variables:
 - `./script/run-clang-format.py -r --extensions 'c,h,cpp,m' ./crates/warpui/src/ ./app/src/` - Format C/C++/Obj-C code
 - `find . -name "*.wgsl" -exec wgslfmt --check {} +` - Check WGSL shader formatting
 
+### Spec Validation
+
+Before opening a spec PR, validate that your spec directory is well-formed:
+
+```bash
+./script/validate_specs.sh   # macOS/Linux
+bash script/validate_specs.sh  # alternative
+```
+
+Each spec under `specs/` must contain:
+- `product.md` (or `PRODUCT.md`) — product spec
+- `tech.md` (or `TECH.md`) — tech spec
+
+The script skips contributor-named directories (e.g. `zachlloyd/`) that group related specs.
+
 ### Platform Setup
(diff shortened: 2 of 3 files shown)


cbeaulieu-gt:
@okwn I think you linked the wrong issue # for closing. The issue I reported doesn't have anything to do with spec compliance.

PR #11453, opened 2026-05-20
feat: support local→cloud handoff snapshot in remote SSH sessions

kevinyang372:
## Description

Local→cloud handoff was completely broken for remote SSH sessions. The snapshot pipeline assumed local filesystem access, causing every step (path resolution, git diff, file reads) to fail silently — the cloud agent always started with no snapshot context.

### Architecture

The remote server daemon already has `ServerApiProvider` (HTTP client + auth + warp-server URL) and can run git commands locally on the remote host. Instead of piping bytes back through SSH, the daemon gathers patches and uploads directly to GCS, returning only the `InitialSnapshotToken` string.

### Changes

**Proto** (`remote_server.proto`):
- `UploadHandoffSnapshot` / `UploadHandoffSnapshotResponse` messages (field 26)

**Daemon** (`handoff_snapshot.rs`, `server_model.rs`):
- Handler runs `derive_touched_workspace` + `upload_snapshot_for_handoff` locally on the remote host
- Uses `ServerApiProvider::get_ai_client()` and `get_http_client()` for direct GCS upload

**Client** (`workspace/view.rs`, `client/mod.rs`, `manager.rs`):
- Detects SSH sessions via `active_session_is_local()`
- Delegates to daemon via `RemoteServerClient::upload_handoff_snapshot()`
- Graceful fallback to `SkippedEmptyWorkspace` when daemon unavailable or old
- Uses `Vec<String>` instead of `Vec<PathBuf>` for cross-platform path safety
- Extracted `settle_handoff_snapshot_result` shared helper for local/remote paths
- Sets empty `TouchedWorkspace` for remote path to satisfy auto-submit gate

## Test Plan

Tested manually:
1. SSH into remote host → `cd alacritty` → `& can you explain the README.md?`
2. Daemon logs show successful snapshot upload (2/2 files uploaded)
3. Cloud agent receives the snapshot token and starts with repo context

---

[Oz conversation](https://staging.warp.dev/conversation/6752c0b1-25b0-4b68-8370-acd213c4d89d) | [Plan](https://staging.warp.dev/drive/notebook/stIJta0ZQre6swqZUjvhyH)

### DIFF
--- a/app/src/remote_server/handoff_snapshot.rs
+++ b/app/src/remote_server/handoff_snapshot.rs
@@ -1,0 +1,81 @@
+//! Daemon-side handler for the `UploadHandoffSnapshot` RPC.
+//!
+//! When the client triggers a local-to-cloud handoff from a remote SSH session,
+//! the daemon runs this module to gather git patches and orphan file contents
+//! from the remote host's filesystem and upload them to GCS via the existing
+//! [`upload_snapshot_for_handoff`] pipeline. Because the daemon IS on the remote
+//! host, all filesystem and git operations are genuinely local — no SSH
+//! tunneling overhead.
+
+use std::path::PathBuf;
+use std::sync::Arc;
+
+use anyhow::Result;
+
+use crate::ai::agent_sdk::driver::upload_snapshot_for_handoff;
+use crate::ai::blocklist::handoff::touched_repos::derive_touched_workspace;
+use crate::server::server_api::ai::{AIClient, InitialSnapshotToken};
+
+/// Gather the workspace snapshot from the given paths and upload it.
+///
+/// 1. Resolves each path to an absolute location (relative paths are resolved
+///    against `working_directory` when provided).
+/// 2. Runs [`derive_touched_workspace`] to discover git roots and orphan files.
+/// 3. Calls [`upload_snapshot_for_handoff`] to build patches, allocate a token,
+///    and upload everything to GCS.
+///
+/// Returns `Ok(Some(token))` when the upload succeeds and a token was minted,
+/// `Ok(None)` when the workspace was empty or the manifest failed, and `Err`
+/// for hard failures (auth, network).
+pub(crate) async fn gather_and_upload_handoff_snapshot(
+    paths: Vec<String>,
+    working_directory: Option<String>,
+    ai_client: Arc<dyn AIClient>,
+    http: &http_client::Client,
+) -> Result<Option<InitialSnapshotToken>> {
+    // Resolve raw path strings to absolute PathBufs.
+    let cwd = working_directory.as_deref().map(PathBuf::from);
+    let resolved_paths: Vec<PathBuf> = paths
+        .into_iter()
+        .filter_map(|raw| {
+            let raw = raw.trim().to_string();
+            if raw.is_empty() {
+                return None;
+            }
+            let candidate = PathBuf::from(&raw);
+            if candidate.is_absolute() {
+                Some(candidate)
+            } else if let Some(cwd) = &cwd {
+                Some(cwd.join(candidate))
+            } else {
+                log::warn!("Skipping relative path with no working_directory: {raw}");
+                None
+            }
+        })
+        .collect();
+
+    if resolved_paths.is_empty() {
+        log::info!("Handoff snapshot: no resolved paths; skipping upload");
+        return Ok(None);
+    }
+
+    log::info!(
+        "Handoff snapshot: deriving workspace from {} path(s)",
+        resolved_paths.len()
+    );
+
+    // Derive the touched workspace — finds git roots and orphan files.
+    // On the daemon these are all local filesystem operations.
+    let workspace = derive_touched_workspace(resolved_paths).await;
+
+    let repo_paths: Vec<PathBuf> = workspace.repos.iter().map(|r| r.git_root.clone()).collect();
+    let orphan_file_paths = workspace.orphan_files;
+
+    log::info!(
+        "Handoff snapshot: {} repo(s), {} orphan file(s)",
+        repo_paths.len(),
+        orphan_file_paths.len()
+    );
+
+    upload_snapshot_for_handoff(repo_paths, orphan_file_paths, ai_client, http).await
+}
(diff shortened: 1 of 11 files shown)


kevinyang372:
* **#11453** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11453?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11453?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/remote_server/handoff_snapshot.rs
```diff
@@ -0,0 +1,75 @@
...
+) -> Result<Option<InitialSnapshotToken>> {
+    let resolved_paths: Vec<PathBuf> = paths
+        .into_iter()
+        .filter_map(|raw| {
+            let raw = raw.trim().to_string();
+            if raw.is_empty() {
+                return None;
+            }
+            let candidate = PathBuf::from(&raw);
+            if candidate.is_absolute() {
```
alokedesai: why do we need to trim here but not in the normal local case?

app/src/remote_server/server_model.rs
```diff
@@ -2440,6 +2444,88 @@ impl ServerModel {
...
+                    }
+                    Err(e) => {
+                        log::warn!(
+                            "Handoff snapshot upload failed \
+                             (request_id={request_id_for_response}): {e:#}"
+                        );
+                        UploadHandoffSnapshotResponse {
+                            initial_snapshot_token: None,
+                            success: false,
+                            error: Some(format!("{e:#}")),
```
alokedesai: Do we need any typed errors over the wire? Is this an error we need to deserialize?

app/src/workspace/view.rs
```diff
@@ -13531,23 +13533,127 @@ impl Workspace {
...
+    ///
+    /// For remote SSH sessions, delegates the gather+upload work to the remote
+    /// server daemon via `UploadHandoffSnapshot` instead of doing it locally.
     #[cfg(all(feature = "local_fs", not(target_family = "wasm")))]
     fn spawn_handoff_snapshot_upload(
-        paths: Vec<PathBuf>,
+        paths: Vec<String>,
         pane_view: ViewHandle<TerminalView>,
         model_handle: ModelHandle<AmbientAgentViewModel>,
+        remote_session_id: Option<SessionId>,
```
alokedesai: seems like we should just require the user pass a sessionID and we internally determine if it's a local or remote session?

Generally seem problematic that we someone can just pass `None` here

app/src/workspace/view.rs
```diff
@@ -13531,23 +13533,127 @@ impl Workspace {
...
+    /// `paths` are raw path strings (not `PathBuf`) because for remote SSH
+    /// sessions they represent paths on the remote host and must not be
+    /// interpreted through the local OS's path encoding.
+    ///
+    /// For remote SSH sessions, delegates the gather+upload work to the remote
+    /// server daemon via `UploadHandoffSnapshot` instead of doing it locally.
     #[cfg(all(feature = "local_fs", not(target_family = "wasm")))]
     fn spawn_handoff_snapshot_upload(
-        paths: Vec<PathBuf>,
+        paths: Vec<String>,
```
alokedesai: can we do the path --> string conversion deeper in the stack? for the actual API, path is definitely right

app/src/workspace/view.rs
```diff
@@ -13531,23 +13533,127 @@ impl Workspace {
...
+                        // (`touched_workspace.is_some()`) is satisfied.
+                        model.set_pending_handoff_workspace(Default::default(), model_ctx);
+                        // Map the proto response into Result<Option<InitialSnapshotToken>>
+                        // so the shared settle helper can handle it.
+                        let mapped = match result {
+                            Ok(resp) if resp.success => {
+                                if let Some(token) = resp.initial_snapshot_token {
+                                    let token: InitialSnapshotToken =
+                                        serde_json::from_value(serde_json::Value::String(token))
+                                            .expect("InitialSnapshotToken is a String newtype");
```
alokedesai: why expect here

app/src/workspace/view.rs
```diff
@@ -13531,23 +13533,127 @@ impl Workspace {
...
+                                let error_msg = resp.error.unwrap_or_default();
+                                Err(anyhow::anyhow!(
+                                    "Remote handoff snapshot failed: {error_msg}"
+                                ))
+                            }
+                            Err(err) => {
+                                Err(anyhow::anyhow!(err)
+                                    .context("Remote handoff snapshot RPC failed"))
+                            }
+                        };
```
alokedesai: Not blocking but I feel like we need a cleaner boundary between the remote code server and the actual functions that wrap it

In this case, we call some async function that directly returns raw remote code types in the workspace.

In reality, I think we want a module (or set of modules) that do the translation for us so we actually operate on more abstract types instead of leaking remote code proto types all over the codebase

app/src/workspace/view.rs
```diff
@@ -14124,18 +14230,41 @@ impl Workspace {
...
+            let mut p: Vec<String> = extract_paths_from_conversation(&source_conversation)
+                .into_iter()
+                .map(|pb| pb.to_string_lossy().into_owned())
+                .collect();
             if let Some(pwd) = source_pwd {
-                p.push(pwd);
+                p.push(pwd.to_string_lossy().into_owned());
             }
-            p
+            (p, None)
```
alokedesai: feels like this could be fully consolidated by

1) adding a new active_session_path function that returns `RemoteOrLocalPath`
2) keeping the paths as `PathBuf`s (if we can)
3) Call into `spawn_handoff_snapshot_upload` with session ID and the paths

app/src/workspace/view.rs
kevinyang372: I am not sure if I fully understand the value of that thin translation layer tho. My mental model of the current contract is the proto definition is the schema for the request / response type a caller should be expect by calling the method. And they should be able to read out of it to populate their internal states

app/src/remote_server/server_model.rs
kevinyang372: I don't think we need typed errors here. There is no actionable errors from the client's side that would require a different UX depending on the error type

app/src/workspace/view.rs
kevinyang372: We cannot do path to string conversion deeper. This is also used for remote sessions which means we cannot assume local OS encoding

alokedesai (approved):

PR #11457, opened 2026-05-21
fix(privacy): refresh SECRETS_REGEX when Safe Mode is toggled at runtime

### LINKED ISSUE #11262: SECRETS_REGEX silently empty at startup despite persisted custom_secret_regex_list — local MCP-save redaction check never fires

david-engelmann:
Fixes #11262.

## Why

\`CustomSecretRegexUpdater\` only subscribed to \`PrivacySettingsChangedEvent::CustomSecretRegexList\`, which fires when the *user-secret regex list itself* mutates. Toggling Safe Mode at runtime doesn't mutate the list — it only flips the boolean that gates redaction — so the subscription never fires and the in-memory \`SECRETS_REGEX\` DFA stays in whatever state it had at app start.

For a user whose regex list was loaded from TOML and who toggled Safe Mode ON at runtime, this meant the local MCP-save redaction check from PR #10839 silently fell through:

1. \`find_secrets_in_text\` ran against an empty DFA
2. \`contains_secrets\` returned \`false\`
3. \`should_block_save_for_secrets(safe_mode=true, enterprise=false, contains=false)\` returned \`false\`
4. The save proceeded → cloud sync rejected → user saw \`Failed to create mcp server\` toast (server caught what the local check should have)

## What

Adds a second subscription in [\`CustomSecretRegexUpdater::new\`](app/src/terminal/secret_regex_updater.rs) for \`SafeModeSettingsChangedEvent::SafeModeEnabled\`. When Safe Mode flips, the updater recompiles \`SECRETS_REGEX\` from the current (persisted) regex list.

## Implementation note: what the existing emission actually does

I initially thought the bug was that \`add_all_recommended_regex\` and \`remove_user_secret_regex\` failed to emit \`CustomSecretRegexList\`. While implementing I confirmed via [\`crates/settings/src/macros.rs:345-361\`](crates/settings/src/macros.rs#L345-L361) that the macro-generated \`set_value\` already emits exactly that event for any setting mutation. So the missing piece really is the SafeMode subscription — both mutator paths fire \`CustomSecretRegexList\` correctly already.

Adds 3 regression tests that pin this contract so it doesn't silently regress:

- \`add_all_recommended_regex_emits_custom_secret_regex_list_event\`
- \`add_all_recommended_regex_when_already_populated_does_not_emit\` (early-return guard)
- \`remove_user_secret_regex_emits_custom_secret_regex_list_event\`

## Test plan

- [x] \`cargo build -p warp --lib\` — clean
- [x] \`cargo test -p warp --lib settings::privacy\` — 3/3 pass
- [x] \`cargo clippy -p warp --lib --tests -- -D warnings\` — clean
- [x] \`cargo fmt -p warp -- --check\` — clean
- [x] Recipe re-run pending: replay [scripts/recipes/11262-secrets-regex-startup-empty.json](https://github.com/david-engelmann/warp-taper/blob/main/scripts/recipes/11262-secrets-regex-startup-empty.json) against a patched build; the save should be blocked and the toast surfaced.

## Repro recording

The repro recording (against an unpatched build) lives in warp-taper:

#11262 reproduction

## Changelog

`CHANGELOG-BUG-FIX: Custom secret redaction patterns are now applied immediately when Safe Mode is toggled during a session. Previously a regex list restored from settings at startup could leave the in-memory matcher empty until the next restart, silently skipping the redaction check when saving an MCP server configuration.`


app/src/settings/privacy.rs
```diff
@@ -827,3 +827,132 @@ impl Entity for PrivacySettings {
...
 impl SingletonEntity for PrivacySettings {}
+
+#[cfg(test)]
+mod tests {
+    //! Regression coverage for #11262: every code path that mutates the
+    //! user-secret regex list must emit
+    //! [`PrivacySettingsChangedEvent::CustomSecretRegexList`] so the
+    //! subscribing `CustomSecretRegexUpdater` can recompile the in-memory
+    //! `SECRETS_REGEX` DFA. Without these events the local MCP-save
+    //! redaction check silently falls through.
```
lucieleblanc: Remove this comment.

app/src/settings/privacy.rs
```diff
@@ -827,3 +827,132 @@ impl Entity for PrivacySettings {
...
+        }
+        count
+    }
+
+    #[test]
+    fn add_all_recommended_regex_emits_custom_secret_regex_list_event() {
+        // The recommended-defaults path was a silent mutator before #11262:
+        // it pushed ~20 regexes into `user_secret_regex_list.set_value`
+        // without emitting `CustomSecretRegexList`, so the global DFA was
+        // never recompiled. Verify the event now fires.
```
lucieleblanc: No need to include the diff reasoning in the comment. The comment should explain what is being tested; users interested in the "why"/history can look at this PR via `git blame`.

lucieleblanc (commented):
Hi @david-engelmann, the code looks pretty good, but I'm not sure what the GIF screen recording in the PR description is supposed to be showing. I can't tell if the issue is the stilted agent computer use or the video quality. Could you manually record the interaction and upload a .mov or .mp4 please? I'll be happy to approve these changes once there are actual before-and-after feature interaction screen captures. Thank you!

PR #11458, opened 2026-05-21
Clear AI block find highlights when the find bar closes

### LINKED ISSUE #11212: Find highlights on AI blocks persist after closing find bar
## Summary
Closing the find bar clears highlights in terminal blocks, but highlights on AI blocks remain visible until the pane is refocused.

## Expected behavior
When the find bar is dismissed, find highlights should clear immediately for both terminal blocks and AI blocks.

## Steps to reproduce
1. Open find in a pane with both terminal and AI blocks.
2. Search for a term that matches in AI blocks.
3. Dismiss the find bar.
4. Observe that AI block highlights remain until the pane is refocused.

## Warp-internal reference
[original report (Warp internal Slack)](https://warpdev.slack.com/archives/C08KTPNQN65/p1779121694711269)

maxmilian:
## Summary

Closing the find bar left stale find highlights on AI blocks until the pane was refocused. `close_find_bar` now clears find matches so AI block child views repaint immediately. Fixes #11212.

## Root cause

`TerminalView::close_find_bar` (`app/src/terminal/view.rs`) only set `is_find_bar_open` to `false` and called `ctx.notify()` on the terminal view itself. AI block find highlights are gated on `is_find_bar_open()` at render time (`app/src/ai/blocklist/block/view_impl.rs`), but AI blocks are separate child views — nothing notified them to repaint when the find bar closed, so their last-rendered highlights persisted until another repaint (e.g. pane refocus). Terminal grid blocks were unaffected because the terminal view repaints itself on close.

## Fix

`close_find_bar` now also calls `find_model.clear_matches()`. That clears each registered rich-content child view's cached find state and notifies it to repaint — mirroring how `open_find_bar` runs find on open. Both the sync (`clear_matches`) and async (`clear_results`) find paths already iterate registered rich-content views, so the fix covers both.

## Tests

Added `close_find_bar_clears_ai_block_find_highlights` in `app/src/terminal/view_tests.rs`: builds an AI block with a searchable query, highlights a match, closes the find bar, and asserts the AI block's find match count drops to `0`. The test **fails before this change** (highlights persist, `left: 1, right: 0`) and **passes after**. `cargo fmt --check`, `cargo clippy --tests -D warnings`, and the test all pass locally.

The test drives the AI block's `run_find` directly because the blocklist find pipeline only visits rich-content views once they have been laid out, which does not happen in a headless test; this is noted in a comment in the test.

## Visual evidence

**Before** (stable v0.2026.05.20.09.21.stable_03) — highlights persist after closing find bar until pane refocus:

https://github.com/user-attachments/assets/706ff7e9-7769-430c-ae6d-c2eabdfbef3c

**After** (this PR) — highlights clear immediately on close:

https://github.com/user-attachments/assets/dff04677-fa85-4035-8cd6-4d374d74f29d

## Scope

- Only `close_find_bar` is touched; opening find and live find updates are unchanged.
- The new `#[cfg(test)]` accessors (`FindState::match_count`, `AIBlock::find_match_count`) exist solely so the regression test can observe AI block find state.

CHANGELOG-BUG-FIX: Fixed find highlights lingering on AI blocks after closing the find bar.

### DIFF
--- a/app/src/ai/blocklist/block.rs
+++ b/app/src/ai/blocklist/block.rs
@@ -4072,6 +4072,13 @@ impl AIBlock {
 
     /// Handles find match focus changes by auto-expanding collapsed reasoning blocks
     /// that contain the focused match.
+    /// The number of cached find matches for this AI block. Test-only; used to
+    /// assert that find highlights are cleared when the find bar closes.
+    #[cfg(test)]
+    pub(crate) fn find_match_count(&self) -> usize {
+        self.find_state.match_count()
+    }
+
     fn handle_find_match_focus_change(&mut self, ctx: &mut ViewContext<Self>) {
         // Get the currently focused match ID from the terminal's find model
         let focused_match_id = self
--- a/app/src/ai/blocklist/block/find.rs
+++ b/app/src/ai/blocklist/block/find.rs
@@ -37,6 +37,12 @@ impl FindState {
     pub(super) fn location_for_match(&self, id: RichContentMatchId) -> Option<&FindMatchLocation> {
         self.matches.get(&id)
     }
+
+    /// The number of find matches currently cached for this block.
+    #[cfg(test)]
+    pub(crate) fn match_count(&self) -> usize {
+        self.matches.len()
+    }
 }
 
 impl FindableRichContentView for AIBlock {
--- a/app/src/terminal/find/model.rs
+++ b/app/src/terminal/find/model.rs
@@ -523,6 +523,23 @@ impl TerminalFindModel {
         ctx.emit(FindEvent::UpdatedFocusedMatch);
     }
 
+    /// Notifies every registered rich-content child view (e.g. AI blocks) to
+    /// drop its cached find state and repaint, **without** touching the active
+    /// find run's options/config.
+    ///
+    /// Callers that just need stale highlights to disappear (e.g.
+    /// `close_find_bar`) must use this rather than [`Self::clear_matches`].
+    /// On the async path, `clear_matches` routes through
+    /// `AsyncFindController::clear_results`, which also drops
+    /// `current_find_options` — losing the query that `open_find_bar` later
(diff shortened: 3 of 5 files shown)


maxmilian:
Addressed Oz's async-path concern (commit `b32cf40f`).

The previous revision called `find_model.clear_matches()` from `close_find_bar`. On the async-find path that helper routes through `AsyncFindController::clear_results`, which also resets `current_find_options` — losing the query that `open_find_bar` reads back via `active_find_options` to restore the previous search.

**Fix:** added `TerminalFindModel::clear_rich_content_matches` — a minimal helper that only notifies registered rich-content child views to drop their cached find state and repaint, **without** touching the active find run's options/config. `close_find_bar` now calls this instead of `clear_matches`. The sync (block-list) path is unaffected because it already preserved options via `run.cleared()`.

**Regression coverage:**
- `close_find_bar_clears_ai_block_find_highlights` (existing, sync path — still green)
- `close_find_bar_preserves_options_on_async_find_path` (new) — enables `FeatureFlag::AsyncFind`, runs find with query `"needle"`, closes find bar, asserts `active_find_options()` still returns `Some("needle")`. Verified the test catches the regression by reverting the fix in place: test fails with `left: None, right: Some("needle")`.

`cargo fmt --check`, `cargo clippy -p warp --lib --tests --no-deps -- -D warnings`, and `cargo nextest run -p warp 'find::'` all green (29/29 find tests + both regression tests). Single squashed commit, force-pushed.

/oz-review

seemeroland (approved):
Nice fix, thanks! I noticed in your video a separate issue - the "list" match in the query is at the wrong position. It looks like the character offset in the query is not taking into account the `/agent` which is prefixed to the query when we render the highlight. Will create a separate issue for that

https://github.com/warpdotdev/warp/issues/11697

PR #11460, opened 2026-05-21
Support remote project rules through repo metadata standing results

moirahuang:
## Description

- Extend project-rule support to remote sessions while using the same discovery, cache, and lookup flow as local sessions.
- Read local and remote `WARP.md` / `AGENTS.md` paths from `RepoMetadataModel` standing-query results. `ProjectContextModel` listens for repository and standing-result changes, then updates its shared `path_to_rules` cache for either a local repository or a host-qualified remote repository.
- Keep file I/O at the app boundary: the app-provided `read_project_rule_contents` function reads local files with `async_fs` and remote files through `RemoteServerManager`. `ProjectContextModel` owns refresh ordering and reconciliation, including per-repository generation checks so an older asynchronous read cannot overwrite a newer refresh.
- Preserve `LocalOrRemotePath` throughout rule lookup, agent context, and the Rules UI so identical remote paths on different hosts remain distinct. Global rules and persisted rule records remain local-only.
- **Existing behavior preserved:** Manually adding any workspace explicitly requests its initial project-rule indexing before `WorkspaceAdded` is emitted. If the workspace is a standalone folder that is not already tracked, that indexing call first registers it with `RepoMetadataModel`; an already-detected repository skips registration but still receives the explicit initial refresh.
- **New behavior in this PR:** Initial hydration now discovers rule paths from `RepoMetadataModel` standing-query results and uses the same shared `ProjectContextModel` refresh path as later metadata events. Subsequent rule changes refresh from metadata events for both local and remote repositories.
- Remove the separate remote-rule reconciliation model. Also remove lookup helpers that silently treated `Path` values as local and replace the stored optional `Arc<dyn Fn>` reader with a stateless function pointer passed to refresh operations.

## Linked Issue

No issue

## Testing

- `./script/format`
- `cargo check -p ai -p warp --all-targets`
- `cargo nextest run -p ai project_context::model::tests` — 28 passed
- `cargo nextest run -p warp ai::metadata_project_rules::tests` — 3 passed
- `cargo nextest run -p warp ai::blocklist::context_model::tests` — 11 passed
- `cargo clippy -p ai -p warp --all-targets -- -D warnings`
- `git diff --check`

## Agent Mode

CHANGELOG-IMPROVEMENT: Remote sessions now discover and apply project rules such as [WARP.md](http://WARP.md) and [AGENTS.md](http://AGENTS.md).

### DIFF
--- a/app/src/ai/blocklist/context_model.rs
+++ b/app/src/ai/blocklist/context_model.rs
@@ -9,6 +9,7 @@ use std::sync::Arc;
 use ai::project_context::model::ProjectContextModel;
 use parking_lot::FairMutex;
 use warp_core::features::FeatureFlag;
+use warp_util::local_or_remote_path::LocalOrRemotePath;
 use warpui::{
     AppContext, Entity, EntityId, ModelContext, ModelHandle, SingletonEntity, WeakModelHandle,
 };
@@ -394,6 +395,25 @@ impl BlocklistAIContextModel {
     /// If `is_user_query` is true, includes blocks, selected text, and images as context.
     /// If false, excludes these user-specific contexts but includes everything else.
     pub fn pending_context(&self, app: &AppContext, is_user_query: bool) -> Vec<AIAgentContext> {
+        let current_working_directory_location = self.current_pwd().and_then(|path| {
+            PathBuf::from_str(&path)
+                .ok()
+                .and_then(|path| path.canonicalize().ok())
+                .map(LocalOrRemotePath::Local)
+        });
+        self.pending_context_for_location(
+            app,
+            is_user_query,
+            current_working_directory_location.as_ref(),
+        )
+    }
+
+    pub fn pending_context_for_location(
+        &self,
+        app: &AppContext,
+        is_user_query: bool,
+        current_working_directory_location: Option<&LocalOrRemotePath>,
+    ) -> Vec<AIAgentContext> {
         let pwd = self.current_pwd();
         let is_pwd_indexed = if cfg!(feature = "agent_mode_evals") {
             // In evals, we want to disable file outline based search. Full
@@ -406,15 +426,9 @@ impl BlocklistAIContextModel {
                 })
         };
 
-        let project_rules = if let Some(pwd) = pwd.clone().and_then(|path| {
-            PathBuf::from_str(&path)
-                .ok()
-                .and_then(|s| s.canonicalize().ok())
-        }) {
-            ProjectContextModel::as_ref(app).find_applicable_rules(&pwd)
-        } else {
-            None
-        };
+        let project_rules = current_working_directory_location.and_then(|pwd| {
+            ProjectContextModel::as_ref(app).find_applicable_rules_at_location(pwd)
+        });
 
         let mut context = Vec::new();
 
@@ -451,14 +465,14 @@ impl BlocklistAIContextModel {
         // Always include project rules if available
         if let Some(rules) = project_rules {
             context.push(AIAgentContext::ProjectRules {
-                root_path: rules.root_path.to_string_lossy().into(),
+                root_path: rules.root_path.display_path(),
                 active_rules: rules
                     .active_rules
                     .into_iter()
                     .map(|rule| {
                         let line_count = rule.content.lines().count();
                         FileContext {
-                            file_name: rule.path.to_string_lossy().into(),
+                            file_name: rule.path.display_path(),
                             content: AnyFileContent::StringContent(rule.content.clone()),
                             line_range: None,
                             last_modified: None,
--- a/app/src/ai/blocklist/context_model.rs
+++ b/app/src/ai/blocklist/context_model.rs
@@ -3,7 +3,6 @@
 
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 use std::sync::Arc;
 
 use ai::project_context::model::ProjectContextModel;
@@ -394,21 +393,7 @@ impl BlocklistAIContextModel {
     /// Returns `AIAgentContext` for the blocks to be included in the current AI query.
     /// If `is_user_query` is true, includes blocks, selected text, and images as context.
     /// If false, excludes these user-specific contexts but includes everything else.
-    pub fn pending_context(&self, app: &AppContext, is_user_query: bool) -> Vec<AIAgentContext> {
-        let current_working_directory_location = self.current_pwd().and_then(|path| {
-            PathBuf::from_str(&path)
-                .ok()
-                .and_then(|path| path.canonicalize().ok())
-                .map(LocalOrRemotePath::Local)
-        });
-        self.pending_context_for_location(
-            app,
-            is_user_query,
-            current_working_directory_location.as_ref(),
-        )
-    }
-
-    pub fn pending_context_for_location(
+    pub fn pending_context(
         &self,
         app: &AppContext,
         is_user_query: bool,
@@ -426,9 +411,8 @@ impl BlocklistAIContextModel {
                 })
         };
 
-        let project_rules = current_working_directory_location.and_then(|pwd| {
-            ProjectContextModel::as_ref(app).find_applicable_rules_at_location(pwd)
-        });
+        let project_rules = current_working_directory_location
+            .and_then(|pwd| ProjectContextModel::as_ref(app).find_applicable_rules(pwd));
 
         let mut context = Vec::new();
 
--- a/app/src/ai/blocklist/context_model.rs
+++ b/app/src/ai/blocklist/context_model.rs
@@ -399,6 +399,8 @@ impl BlocklistAIContextModel {
         is_user_query: bool,
         current_working_directory_location: Option<&LocalOrRemotePath>,
     ) -> Vec<AIAgentContext> {
+        // `pwd` is the shell-reported path used for directory context and local indexing.
+        // The location is passed separately because it preserves remote host identity for rules.
         let pwd = self.current_pwd();
         let is_pwd_indexed = if cfg!(feature = "agent_mode_evals") {
             // In evals, we want to disable file outline based search. Full
--- a/app/src/ai/metadata_project_rules.rs
+++ b/app/src/ai/metadata_project_rules.rs
@@ -1,0 +1,268 @@
+use std::collections::HashMap;
+
+use ai::project_context::model::{ProjectContextModel, ProjectRule};
+use futures::future::{BoxFuture, FutureExt as _};
+use remote_server::proto::{
+    file_context_proto, FileContextProto, ReadFileContextFile, ReadFileContextRequest,
+};
+use repo_metadata::{
+    RepoMetadataEvent, RepoMetadataModel, RepositoryIdentifier, StandingQueryContent,
+};
+use warp_util::local_or_remote_path::LocalOrRemotePath;
+use warp_util::remote_path::RemotePath;
+use warpui::{AppContext, Entity, ModelContext, SingletonEntity};
+
+use crate::remote_server::manager::RemoteServerManager;
+
+pub(crate) struct MetadataProjectRulesModel {
+    refresh_generations: HashMap<RepositoryIdentifier, u64>,
+    next_refresh_generation: u64,
+}
+
+type ProjectRuleContentsFuture =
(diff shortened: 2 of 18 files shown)


moirahuang:
* **#11921** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11921?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11460** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12047** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12047?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#12051](https://github.com/warpdotdev/warp/pull/12051) <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12051?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#12035** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12035?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/ai/src/project_context/model.rs
```diff
@@ -174,16 +177,26 @@ impl ProjectRules {
...
+    path_to_rules: HashMap<LocalOrRemotePath, ProjectRules>,
     /// Queued repository updates for paths that have an in-flight processing task.
     /// The presence of a key indicates an active async task for that path;
     /// the Vec holds updates that arrived while that task was running.
     #[cfg(feature = "local_fs")]
     pending_updates: HashMap<PathBuf, Vec<RepositoryUpdate>>,
+    /// Most recent full filesystem scan generation for each local root, so a slower older
+    /// fallback scan cannot replace newer rule state.
+    #[cfg(feature = "local_fs")]
+    local_scan_generations: HashMap<PathBuf, u64>,
```
moirahuang: agent mode pointed out that there can be  multiple asynchronous filesystem scans for the same local root so this prevents older scans from overwriting newer ones

app/src/ai/metadata_project_rules.rs
```diff
@@ -0,0 +1,110 @@
...
+use futures::future::{BoxFuture, FutureExt as _};
+use remote_server::proto::{
+    file_context_proto, FileContextProto, ReadFileContextFile, ReadFileContextRequest,
+};
+use warp_util::local_or_remote_path::LocalOrRemotePath;
+use warpui::{AppContext, SingletonEntity};
+
+use crate::remote_server::manager::RemoteServerManager;
+
+pub(crate) fn read_project_rule_contents(
```
moirahuang, line 13: aka ProjectRuleContentReader

crates/ai/src/project_context/model.rs
```diff
@@ -1,29 +1,60 @@
...
         use warp_util::standardized_path::StandardizedPath;
     }
 }
 
-#[derive(Debug, Default, Clone)]
+pub type ProjectRuleContents = Vec<(LocalOrRemotePath, String)>;
+/// App-provided transport for reading the exact rule paths discovered by repository metadata.
+///
+/// This remains injected because remote file reads are implemented in the app crate.
+pub type ProjectRuleContentReader = fn(
```
moirahuang, line 25: aka read_project_rule_contents which is used to read the actual project rule contents

crates/ai/src/project_context/model.rs
```diff
@@ -303,111 +345,130 @@ impl ProjectContextModel {
...
-            .flat_map(|results| results.project_rules())
-            .filter(|content| !content.is_directory)
-            .filter_map(|content| content.path.to_local_path())
-            .collect::<Vec<_>>();
-        let existing_rules = self
-            .path_to_rules
-            .get(&project_root)
-            .cloned()
-            .unwrap_or_default();
+        let rule_paths = standing_project_rule_paths(
```
moirahuang, line 368: converts standing query results into local or remote values

app/src/ai/blocklist/context_model.rs
```diff
@@ -406,15 +411,8 @@ impl BlocklistAIContextModel {
...
-        let project_rules = if let Some(pwd) = pwd.clone().and_then(|path| {
-            PathBuf::from_str(&path)
-                .ok()
-                .and_then(|s| s.canonicalize().ok())
-        }) {
-            ProjectContextModel::as_ref(app).find_applicable_rules(&pwd)
-        } else {
-            None
-        };
+        let project_rules = current_working_directory_location
```
moirahuang, line 416: we're already typed, that simplifies this logic here

crates/ai/src/project_context/global_rules.rs
```diff
@@ -159,7 +160,8 @@ impl GlobalRules {
                     me.global_rules.rules.insert(
                         file_path.clone(),
                         ProjectRule {
-                            path: file_path.clone(),
+                            // Global rule sources are watched under the local home directory.
+                            path: LocalOrRemotePath::Local(file_path.clone()),
```
moirahuang, line 164: purposefully keeping local

crates/ai/src/project_context/model.rs
```diff
@@ -303,111 +345,130 @@ impl ProjectContextModel {
...
+                    discovered_rules: Vec::new(),
+                    deleted_rules,
+                }));
+            }
+            ctx.emit(ProjectContextModelEvent::PathIndexed);
+        }
+    }
+
+    #[cfg(feature = "local_fs")]
+    fn apply_project_rules(
```
moirahuang, line 443: this is pulling out the logic from the original callsite

app/src/ai/blocklist/context_model.rs
```diff
@@ -393,7 +393,12 @@ impl BlocklistAIContextModel {
     /// Returns `AIAgentContext` for the blocks to be included in the current AI query.
     /// If `is_user_query` is true, includes blocks, selected text, and images as context.
     /// If false, excludes these user-specific contexts but includes everything else.
-    pub fn pending_context(&self, app: &AppContext, is_user_query: bool) -> Vec<AIAgentContext> {
+    pub fn pending_context(
+        &self,
+        app: &AppContext,
+        is_user_query: bool,
+        current_working_directory_location: Option<&LocalOrRemotePath>,
```
kevinyang372, line 400: Do we need to pass in current_working_directory_location as a separate parameter? How's that different from pwd (other than the typing?)

app/src/terminal/view/init_project/model.rs
```diff
@@ -187,7 +186,7 @@ impl InitProjectModel {
             && !*CodeSettings::as_ref(ctx).auto_indexing_enabled;
 
         let has_pending_project_scoped_rules = ProjectContextModel::as_ref(ctx)
-            .find_applicable_project_rules(path)
+            .find_applicable_project_rules(&LocalOrRemotePath::Local(path.to_path_buf()))
```
kevinyang372, line 189: Can you run /init in remote session? (eitherway don't need to address it here)

crates/ai/src/project_context/model.rs
```diff
@@ -161,13 +196,13 @@ impl ProjectRules {
...
+#[derive(Default)]
 pub struct ProjectContextModel {
     /// Mapping from directory path to list of rule files found in that directory
-    path_to_rules: HashMap<PathBuf, ProjectRules>,
-    /// Latest metadata-backed async refresh per project root.
+    path_to_rules: HashMap<LocalOrRemotePath, ProjectRules>,
+    /// Latest metadata-backed async refresh per repository.
     #[cfg(feature = "local_fs")]
-    rule_refresh_generations: HashMap<PathBuf, u64>,
+    rule_refresh_generations: HashMap<RepositoryIdentifier, u64>,
```
kevinyang372, line 208: Why RepositoryIdentifier here instead of LocalOrRemotePath?

crates/ai/src/project_context/model.rs
```diff
@@ -279,16 +311,26 @@ impl ProjectContextModel {
                     ctx.emit(ProjectContextModelEvent::PathIndexed);
                 },
             );
+
+            let remote_repo_ids = RepoMetadataModel::as_ref(ctx)
```
kevinyang372, line 320: Will we even have any remote repos on app startup (this is called on initialize_app yeah?)

crates/ai/src/project_context/model.rs
kevinyang372, line 368: Nice

crates/ai/src/project_context/model.rs
```diff
@@ -303,111 +345,130 @@ impl ProjectContextModel {
...
     ) {
-        let Some(project_root) = repo_path.to_local_path() else {
+        self.rule_refresh_generations.remove(repo_id);
+        let Some(project_root) = repo_id.to_local_or_remote_path() else {
             return;
         };
-        self.rule_refresh_generations.remove(&project_root);
         if let Some(rules) = self.path_to_rules.remove(&project_root) {
-            let deleted_rules = rules.all_rule_paths().cloned().collect();
+            if matches!(repo_id, RepositoryIdentifier::Local(_)) {
```
kevinyang372, line 431: Why we filter by local only?

kevinyang372 (approved):

app/src/ai/blocklist/context_model.rs
moirahuang: Yes, we need both today. `pwd` is the shell-reported path used for directory context and local indexing, while `current_working_directory_location` comes from the active session and preserves remote host identity. Deriving the rule lookup path from `pwd` would incorrectly treat a remote path such as `/repo` as local. I added a comment clarifying the distinction.

app/src/terminal/view/init_project/model.rs
moirahuang: `/init` is not currently supported for a remote session. The flow canonicalizes the pwd on the client filesystem and then performs local-only indexing, LSP, filesystem checks, and symlink operations, so the explicit `Local` path is intentional here. Remote `/init` support would need to be handled separately.

crates/ai/src/project_context/model.rs
moirahuang: The generation tracks an async refresh initiated by a `RepoMetadataEvent`, so it is keyed by the exact repository identity carried by metadata events and removed using that same identity on `RepositoryRemoved`. `RepositoryIdentifier` also preserves the standardized local repository identity; `LocalOrRemotePath` is the file/location representation used after conversion. I added a comment clarifying this invariant.

crates/ai/src/project_context/model.rs
moirahuang: Usually there will not be remote repos on a cold startup, but `RepoMetadataModel` is registered before `ProjectContextModel`, and a remote snapshot can arrive before `ProjectContextModel` subscribes to metadata events. Refreshing already-tracked remote IDs closes that race and hydrates any preexisting snapshots. I added a comment explaining this.

crates/ai/src/project_context/model.rs
moirahuang: This filter only gates `KnownRulesChanged`, whose delta contains local `PathBuf`s and is consumed by local persistence. Remote rules are still removed from `path_to_rules`, and `PathIndexed` is still emitted so in-memory state and UI update; remote removals should not be written into the local rules database. I added a comment clarifying that distinction.

app/src/ai/blocklist/context_model.rs
moirahuang: agent mode wrote this but agreed^ i'm going to keep `current_working_directory_location`

app/src/terminal/view/init_project/model.rs
moirahuang: ^this is true, going to keep it as `Local` here

crates/ai/src/project_context/model.rs
moirahuang: pushed this comment^

crates/ai/src/project_context/model.rs
moirahuang: `KnownRulesChanged` is consumed by `PersistedWorkspace` to update the local project-rules database

crates/ai/src/project_context/model.rs
moirahuang: ^added this comment

PR #11482, opened 2026-05-21
Handle session reconnecting for remote server executor

MaggieShan:
## Description
* Context: https://warpdev.slack.com/archives/C0AMRA82ZKL/p1779143889865059
* We're getting a lot of client error related to the run command request failing when the session is disconnected - this pr ensures that we're not sending any run command requests unless the remote server client is actually connected 
  * We now hold a `RwLock<Option<Arc<RemoteServerClient>>>` so that we can clear/set a `RemoteServerClient` during disconnection and reconnection 
  * We add a `SessionReconnecting` event to cover 
* We are getting flooded with client errors from sending run command request when the session is already disconnected. I don’t think this is an actual user-facing error but we should avoid sending the request when we know the remote server is disconnected (https://warp.metabaseapp.com/question/42606-daily-client-request-error-rate-by-operation-error-type) 

## Testing
Verified locally that on disconnection the run command request does not get sent + client error is not logged. 

## Agent Mode

### DIFF
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -413,6 +413,7 @@ repo_metadata = { workspace = true, features = ["test-util"] }
 
 [target.'cfg(not(target_family = "wasm"))'.dev-dependencies]
 tokio = { workspace = true, features = ["rt-multi-thread", "macros", "process", "time"] }
+tokio-util = { workspace = true, features = ["compat"] }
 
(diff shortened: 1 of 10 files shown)


MaggieShan:
Closing in favour or https://github.com/warpdotdev/warp/pull/11558

PR #11484, opened 2026-05-21
Update telemetry for code review over remote sessions

MaggieShan:
## Description 
* Wires is_local through existing code review view events 
  * Some(true) => local, Some(false) => remote, None => no repo
* Moves event emission to code review for access to repo path
* Tracks load diff latency 
* Emits events for decode or empty errors   

## Testing 
Verified by running locally with_sandbox_telemetry

## Agent mode
- [x] Warp Agent Mode

### DIFF
--- a/app/src/ai/agent/conversation.rs
+++ b/app/src/ai/agent/conversation.rs
@@ -2403,27 +2403,6 @@ impl AIConversation {
                         Some(api::message::Message::UpdateReviewComments(comments)) => {
                             if let Some(comments_op) = comments.operation.as_ref() {
                                 if let Some(active_code_review) = self.code_review.as_mut() {
-                                    let is_local = match comments_op {
-                                        api::message::update_review_comments::Operation::AddressReviewComments(
-                                            addressed_comments,
-                                        ) => active_code_review
-                                            .pending_comments
-                                            .iter()
-                                            .filter(|item| {
-                                                addressed_comments
-                                                    .comment_ids
-                                                    .iter()
-                                                    .any(|comment_id| {
-                                                        item.id.to_string() == *comment_id
-                                                    })
-                                            })
-                                            .find_map(|item| {
-                                                item.diff
-                                                    .file_path
-                                                    .as_ref()
-                                                    .map(|path| path.is_local())
-                                            }),
-                                    };
                                     let resolved_count = update_comment_from_comment_operation(
                                         active_code_review,
                                         comments_op.clone(),
@@ -2431,7 +2410,6 @@ impl AIConversation {
                                     if resolved_count > 0 {
                                         send_telemetry_from_ctx!(
                                             CodeReviewTelemetryEvent::CommentResolved {
-                                                is_local,
                                                 resolved_count
                                             },
                                             ctx
--- a/app/src/ai/agent/conversation.rs
+++ b/app/src/ai/agent/conversation.rs
@@ -2403,6 +2403,27 @@ impl AIConversation {
                         Some(api::message::Message::UpdateReviewComments(comments)) => {
                             if let Some(comments_op) = comments.operation.as_ref() {
                                 if let Some(active_code_review) = self.code_review.as_mut() {
+                                    let is_local = match comments_op {
+                                        api::message::update_review_comments::Operation::AddressReviewComments(
+                                            addressed_comments,
+                                        ) => active_code_review
+                                            .pending_comments
+                                            .iter()
+                                            .filter(|item| {
+                                                addressed_comments
+                                                    .comment_ids
+                                                    .iter()
+                                                    .any(|comment_id| {
+                                                        item.id.to_string() == *comment_id
+                                                    })
+                                            })
+                                            .find_map(|item| {
+                                                item.diff
+                                                    .file_path
+                                                    .as_ref()
+                                                    .map(|path| path.is_local())
+                                            }),
+                                    };
                                     let resolved_count = update_comment_from_comment_operation(
                                         active_code_review,
                                         comments_op.clone(),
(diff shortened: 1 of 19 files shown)


app/src/ai/agent/conversation.rs
```diff
@@ -2409,13 +2409,35 @@ impl AIConversation {
                         Some(api::message::Message::UpdateReviewComments(comments)) => {
                             if let Some(comments_op) = comments.operation.as_ref() {
                                 if let Some(active_code_review) = self.code_review.as_mut() {
+                                    let is_local = match comments_op {
```
kevinyang372: I don't think resolve comment needs a is_local field and this seems like a lot of logic. I will lean towards removing it

app/src/code/editor/view/actions.rs
```diff
@@ -1071,8 +1069,6 @@ impl TypedActionView for CodeEditorView {
             }
             RevertDiffHunk { line_range } => {
                 if FeatureFlag::RevertDiffHunk.is_enabled() {
-                    send_telemetry_from_ctx!(CodeReviewTelemetryEvent::RevertHunkClicked, ctx);
```
kevinyang372, line 1069: Why is this removed?

app/src/code/editor/model.rs
```diff
@@ -3831,9 +3829,6 @@ impl CoreEditorModel for CodeEditorModel {
 
 impl CodeEditorModel {
     pub fn open_comment_line(&mut self, line: &EditorLineLocation, ctx: &mut ModelContext<Self>) {
-        // Telemetry: comment editor opened for a new inline review comment.
```
kevinyang372, line 3829: Why are we refactoring the CommentEditorOpened event?

app/src/code_review/diff_state/local.rs
```diff
@@ -1412,10 +1414,12 @@ impl LocalDiffStateModel {
...
             }
-            Err(e) => {
+            Err(_) => {
                 send_telemetry_from_ctx!(
-                    CodeReviewTelemetryEvent::CalculateDiffMetadataFailed {
-                        error: e.to_string()
+                    CodeReviewTelemetryEvent::LoadMetadataFailed {
+                        is_local: Some(true),
+                        mode: self.mode.clone(),
+                        error: "Load metadata failed".to_string(),
```
kevinyang372: Also why are we no longer preserving the error here?

app/src/code_review/diff_state/local.rs
```diff
@@ -1462,10 +1466,12 @@ impl LocalDiffStateModel {
...
         ctx: &mut ModelContext<Self>,
     ) {
-        if let Err(e) = &diffs.changes {
+        if diffs.changes.is_err() {
             send_telemetry_from_ctx!(
                 CodeReviewTelemetryEvent::LoadDiffFailed {
-                    error: e.to_string(),
+                    is_local: Some(true),
+                    mode: self.mode.clone(),
+                    error: "Load diff data failed".to_string(),
```
kevinyang372: ditto

app/src/code_review/diff_state/remote.rs
```diff
@@ -215,6 +216,14 @@ impl RemoteDiffStateModel {
             Ok((metadata, state, diffs)) => self.apply_snapshot(metadata, state, diffs, ctx),
             Err(error) => {
                 log::warn!("RemoteDiffStateModel: invalid diff state snapshot: {error}");
```
kevinyang372: Actually instead of logging telemetry, would it make more sense to make this warn statement log::error so we could track these in sentry?

One thing we should do a pass on is the message here doesn't contain PII or IDs (e.g. Session ID)

app/src/code_review/code_review_view.rs
```diff
@@ -693,13 +695,21 @@ pub struct CodeReviewView {
     code_review_footer: Option<ViewHandle<CodeFooterView>>,
     /// Active git-operation dialog overlay (commit / push / publish), if open.
     git_dialog: Option<ViewHandle<GitDialog>>,
+    /// Timestamp captured when a diff load is initiated. Cleared when the
+    /// corresponding `DiffLoadCompleted` telemetry event is emitted, so the
+    /// elapsed time between request and completion can be reported.
+    diff_load_start_time: Option<Instant>,
```
kevinyang372: Instead of putting this on the overall CodeReviewView struct, should it belong to InternalRemoteDiffState / InternalDiffState in diff state model? Then we could pop and log the load time once we transition from loading -> loaded / error

kevinyang372 (approved):

app/src/code/editor/model.rs
MaggieShan: We're moving it so it's instead emitted from the code review view so we have access to the is_local field

app/src/code/editor/view/actions.rs
MaggieShan: Similarly it's been moved to https://github.com/warpdotdev/warp/pull/11484/changes/BASE..ec853652b6d299f7029226e201a8d5a108b6def4#diff-d09eaac6bf49cd8e951261cea427613054c5f168fc24ea245dfb698626dfeec8R5649-R5654

app/src/code_review/diff_state/remote.rs
MaggieShan: That's a good point - can confirm that the error message themselves don't include any PII and mainly come from the `try_from` fns that send static strings. Updated to send error logs with `safe_error!` instead of telemetry events

app/src/code_review/code_review_view.rs
MaggieShan: Good point - will update!

PR #11494, opened 2026-05-21
Fix AI blocks not appearing due to conversation existing in more than one pane

seemeroland:
## Description
Fix queries and output not showing up due to a conversation existing in more than one pane. A conversation and its blocks should only be in one pane at a time. This PR enforces this by notifying when a conversation becomes active in a new terminal view, so that old terminal views can delete the AI and command blocks for that conversation.

https://github.com/warpdotdev/warp/pull/10241/changes#diff-fe3dd831b48d66196f87f13cd7a57f0e3660444b8f4fd9f6a5718be5fa617e40 introduced `mark_active_conversation_id` to not transfer ownership. https://github.com/warpdotdev/warp/pull/10327 uses it in the restoration path. The first PR mentions it's short term, and @advait-m mentioned these original fixes are probably no longer needed, but would like review from you and @szgupta to verify. My understanding is that a conversation must never be in more than one terminal view, even with orchestration, so it should be correct to enforce this when setting active conversation id.

I'm also unsure what should be the source of truth for where the conversation lives. I'm using the ai history model for this, but it seems like the agent view controller also has this info.

## Testing

### Screenshots / Videos

https://www.loom.com/share/e0af242e4f2f42dcbe77b1753dca5177

CHANGELOG-BUG-FIX: Fixed an issue that could prevent an AI query and output from showing up


app/src/terminal/view/load_ai_conversation.rs
```diff
@@ -443,21 +443,7 @@ impl TerminalView {
...
-                // `restore_conversation_after_view_creation` decides after
-                // restoration whether the current view should become the
-                // canonical owner (regular terminal panes) or remain a
-                // non-owning mirror (conversation transcript viewers).
-                history_model.mark_active_conversation_id(
-                    active_conversation_id,
-                    self.view_id,
-                    ctx,
-                );
+                history_model.set_active_conversation_id(active_conversation_id, self.view_id, ctx);
```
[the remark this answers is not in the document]
seemeroland: It _should_ not be possible to transfer the conversation to a new pane while it's still in progress - entrypoints should be navigating to where it lives.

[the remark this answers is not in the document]
advait-m: note that it's possible to open a conversation in a "new" pane while it's in progress but we're NOT actually "transferring" the conversation, we're just changing where/how we're rendering the terminal view i.e. it's always in the same pane but it could be same pane switch, new pane split or new tab split.

Tested this manually and confirmed it still works as expected!

app/src/terminal/view/agent_view.rs
```diff
@@ -146,9 +151,31 @@ impl TerminalView {
                 );
                 self.show_error_toast(e.to_string(), ctx);
             }
+        } else if let Some(conversation) = in_memory_conversation {
+            self.restore_conversation_after_view_creation(
+                RestoredAIConversation::new(conversation),
+                false,
+                ctx,
+            );
+            if let Err(e) = self.try_enter_agent_view(
```
advait-m, line 163: nit: I think the restore call enters agent view as a side effect, so the follow-up entry sees "was already in agent view = true" when it shouldn't? For origins that auto-trigger only when already in agent view, the initial prompt could auto-submit. 

potential fix: capture "was in agent view" before the restore call, or split the restore so it doesn't implicitly enter agent view and enter once after (matching the live branch)

trace:
- `restore_conversations_from_block_params`
- `ai_context_model.set_pending_query_state_for_existing_conversation`
- `agent_view_controller.try_enter_agent_view`

advait-m (approved):
I tried this out myself manually with in-progress convos + finished convos and opening in new pane/tab + closing for orchestration. Seemed to all work as expected!

szgupta:
I'm gonna defer to @zachbai here, but we need to make sure we don't regress #10241 again. It's a really bad experience - conversation halts randomly.

seemeroland:
That issue happened when running two parallel conversations in split panes? I don't see why that should be an issue since they should be different conversation ids, unless it was a workaround around a deeper issue with the same conversation actually existing in both panes.

zachbai (approved):

app/src/terminal/view/agent_view.rs
seemeroland: This was the fix to restoring in the active pane, but agree this function should always restore and callers should do the liveness check before calling this.

Moving this fix within `restore_conversation_in_active_pane`

seemeroland: Actually that doesn't address the trace you mentioned. I'm adding an arg passed down the restoration path to not enter the agent view if the caller is going to enter the agent view afterward with a different origin. Seems like this occurs in several other places

advait-m: sounds good!

PR #11496, opened 2026-05-21
[WARP-XXXX] Enable custom endpoint usage tracking and display

dagmfactory:
## Description

Enables the CustomInferenceEndpoints feature flag in stable builds and adds proper token usage tracking for custom endpoints with alias display.

### Changes
- Enable CustomInferenceEndpoints feature flag in stable builds
- Add custom endpoint token usage aggregation in conversation footer
- Display custom endpoint aliases in usage views instead of raw config keys
- Add unit tests for custom endpoint usage metadata parsing and display
- Wire up custom endpoint usage in shared session replay

### Testing
- Added unit tests in conversation_tests.rs for custom endpoint usage metadata
- Added llms_tests.rs for custom endpoint display label resolution
- Added conversation_usage_view_tests.rs for UI display formatting

Screenshot 2026-05-20 at 8 12 35 PM

Screenshot 2026-05-21 at 3 53 18 PM

CHANGELOG-IMPROVEMENT: Custom inference endpoints now show token usage with friendly alias names in Agent Mode.

### DIFF
--- a/app/src/ai/agent/conversation.rs
+++ b/app/src/ai/agent/conversation.rs
@@ -115,7 +115,7 @@ fn footer_model_token_usage(
     for (config_key, usage) in &usage_metadata.custom_endpoint_token_usage {
         let label = custom_endpoint_display_label(config_key);
         let entry = token_usage
-            .entry(format!("custom_endpoint~{config_key}"))
+            .entry(config_key.clone())
             .or_insert_with(|| ModelTokenUsage {
                 model_id: label,
                 custom_endpoint_config_key: Some(config_key.clone()),
--- a/app/src/ai/agent/conversation.rs
+++ b/app/src/ai/agent/conversation.rs
@@ -70,7 +70,7 @@ use crate::terminal::model::block::{
     AgentInteractionMetadata, AgentViewVisibility, BlockId, SerializedAIMetadata, SerializedBlock,
 };
 use crate::ui_components::icons::Icon;
-use crate::{BlocklistAIHistoryModel, GlobalResourceHandlesProvider, LLMPreferences};
+use crate::{BlocklistAIHistoryModel, GlobalResourceHandlesProvider};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum TodoStatus {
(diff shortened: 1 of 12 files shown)


app/src/ai/blocklist/controller/shared_session.rs
```diff
@@ -500,6 +500,7 @@ impl BlocklistAIController {
                 .map(|conversation| stream_finished::ConversationUsageMetadata {
                     context_window_usage: conversation.context_window_usage(),
                     credits_spent: conversation.credits_spent(),
+                    platform_credits_spent: 0.0,
```
danielpeng2, line 503: Just checking, was `platform_credits_spent` a field someone else added in the protos?

app/src/ai/agent/conversation.rs
```diff
@@ -87,6 +87,60 @@ impl TodoStatus {
     }
 }
 
+fn footer_model_token_usage(
+    usage_metadata: &stream_finished::ConversationUsageMetadata,
+    custom_endpoint_display_label: impl Fn(&str) -> String,
```
danielpeng2: nit: do we need to pass a closure in here? Might be simpler to just pass in an AppContext and read from `crate::ai::llms::LLMPreferences::as_ref(ctx)` directly to get the label.

crates/persistence/src/model.rs
```diff
@@ -1091,15 +1091,22 @@ pub fn token_usage_category_display_name(category: &str) -> String {
 #[derive(Debug, Serialize, Deserialize, Clone, Default)]
 pub struct ModelTokenUsage {
     pub model_id: String,
+    /// Server-side custom endpoint config key used to faithfully replay usage metadata.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub custom_endpoint_config_key: Option<String>,
```
danielpeng2: Instead of storing the config key, I wonder if we can use the existing `model_id` field and store the custom model id/alias in there. Seems convenient to handle the `config_key` -> `model_id` translation once upfront and not worry about it later on.

app/src/ai/blocklist/controller/shared_session.rs
dagmfactory: yep https://github.com/warpdotdev/warp-proto-apis/blob/c67de64fc4949f693a679552dc88cebc9f7d0180/apis/multi_agent/v1/response.proto#L111

danielpeng2 (approved):

PR #11555, opened 2026-05-22
Enable async find on dogfood, add toggle for Preview/Stable

vkodithala:
## Description

Surfaces an "Asynchronous find" toggle on the Features settings page so non-dogfood users can opt into this experimental feature, while keeping it force-enabled (no toggle) on dogfood channels.

Behavior by channel:

- **Dogfood (Local/Dev)**: `FeatureFlag::AsyncFind` is in `DOGFOOD_FLAGS`, so the feature is force-enabled and no toggle is shown.
- **Preview/Stable**: the flag is off, so the toggle appears on the Features page with `default: false`. Users opt in via the toggle.

The feature flag itself is the discriminator between these two modes — there's no separate channel-aware default baked into the setting. Promoting the feature through the normal flag-promotion ladder (add to `PREVIEW_FLAGS`, then `RELEASE_FLAGS`) automatically hides the toggle and force-enables the feature; no code changes here required.

On channels where the toggle is visible, a small "BETA" chip is rendered next to the title to signal the experimental nature of the feature, styled to match the `TitleChip` pattern used elsewhere (e.g. MCP server cards).

Concrete changes:

- `crates/warp_features/src/lib.rs`: add `FeatureFlag::AsyncFind` to `DOGFOOD_FLAGS` so the feature is force-on for Local/Dev.
- `app/src/terminal/settings.rs`: add the `async_find_enabled` setting (`terminal.async_find_enabled`) with `default: false`, and a `TerminalSettings::is_async_find_enabled()` helper that returns `FeatureFlag::AsyncFind.is_enabled() || *self.async_find_enabled`.
- `app/src/terminal/find/model.rs`: route the async-find init through `TerminalSettings::is_async_find_enabled()` so callers don't need to track both the flag and the setting.
- `app/src/settings_view/features_page.rs`: add the `AsyncFindWidget` to the Features page, gated by `!FeatureFlag::AsyncFind.is_enabled()` so it only renders on channels where the user can meaningfully toggle. Adds a free `render_beta_chip` helper for the inline "BETA" pill.

## Testing

Confirmed via manual testing on Preview/Stable-like builds (flag off, toggle visible) and dogfood-like builds (flag on, toggle hidden, feature force-enabled).

- [x] I have manually tested my changes locally with `./script/run` — toggle appears with the BETA chip on the Features page when the flag is off, is hidden when the flag is on, and async find engages/disengages with the setting.

## Agent Mode

CHANGELOG-IMPROVEMENT: A new, faster implementation of find is now available as an opt-in setting under Features → Asynchronous find. This will help keep the UI responsive on large outputs.

### DIFF
--- a/app/src/settings_view/features_page.rs
+++ b/app/src/settings_view/features_page.rs
@@ -43,7 +43,8 @@ use super::settings_page::{
     HEADER_PADDING, TOGGLE_BUTTON_RIGHT_PADDING,
 };
 use super::{
-    features, flags, DisplayCount, SettingsAction, SettingsSection, ToggleSettingActionPair,
+    features, flags, render_beta_chip, DisplayCount, SettingsAction, SettingsSection,
+    ToggleSettingActionPair,
 };
 use crate::appearance::Appearance;
 use crate::default_terminal::DefaultTerminal;
@@ -7284,10 +7285,12 @@ impl SettingsWidget for AsyncFindWidget {
     type View = FeaturesPageView;
 
     fn search_terms(&self) -> &str {
-        "async asynchronous find search terminal ai background"
+        "async asynchronous fast find search"
     }
 
     fn should_render(&self, _app: &AppContext) -> bool {
+        // Here, the feature flag being enabled means the feature is force-enabled,
+        // so we don't need to render the toggle.
         !FeatureFlag::AsyncFind.is_enabled()
     }
 
@@ -7336,26 +7339,9 @@ impl SettingsWidget for AsyncFindWidget {
             switch,
             appearance,
             Some(
-                "Run find on a background thread to keep the UI responsive on large outputs."
+                "Use an improved implementation of find to keep the UI responsive while searching for matches on large outputs."
                     .into(),
             ),
         )
     }
-}
-
-/// Small inline pill rendered next to a settings label to mark a feature as beta.
-fn render_beta_chip(appearance: &Appearance) -> Box<dyn Element> {
-    let theme = appearance.theme();
-    let chip_color = theme.sub_text_color(theme.surface_3()).into_solid();
-    Container::new(
-        Text::new_inline("BETA", appearance.ui_font_family(), 10.)
-            .with_color(chip_color)
-            .finish(),
-    )
-    .with_background(theme.surface_3())
-    .with_corner_radius(CornerRadius::with_all(Radius::Pixels(3.)))
-    .with_horizontal_padding(4.)
-    .with_vertical_padding(1.)
-    .with_margin_left(8.)
-    .finish()
 }
--- a/app/src/settings_view/features_page.rs
+++ b/app/src/settings_view/features_page.rs
@@ -36,10 +36,11 @@ use super::keybindings::KeyBindingModifyingState;
 #[cfg(feature = "local_tty")]
 use super::settings_page::render_sub_sub_header;
 use super::settings_page::{
-    add_setting, build_reset_button, render_body_item, render_body_item_label,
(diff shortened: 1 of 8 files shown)


vkodithala:
* **#11555** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11555?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11555?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/terminal/find/model.rs
```diff
@@ -237,8 +237,8 @@ impl FindModel for TerminalFindModel {
 }
 
 impl TerminalFindModel {
-    pub fn new(terminal_model: Arc<FairMutex<TerminalModel>>) -> Self {
-        let async_find_controller = if FeatureFlag::AsyncFind.is_enabled() {
+    pub fn new(terminal_model: Arc<FairMutex<TerminalModel>>, ctx: &AppContext) -> Self {
+        let async_find_controller = if TerminalSettings::as_ref(ctx).is_async_find_enabled() {
```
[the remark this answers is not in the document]
vkodithala: KI, it's not actually worth setting up this infra b/c this feature will eventually be default-enabled (not a toggle).

app/src/settings_view/features_page.rs
```diff
@@ -7256,3 +7274,88 @@ impl SettingsWidget for GraphicsBackendWidget {
...
+    fn render(
+        &self,
+        view: &Self::View,
+        appearance: &Appearance,
+        app: &AppContext,
+    ) -> Box<dyn Element> {
+        let ui_builder = appearance.ui_builder();
+
+        let label = render_body_item_label::<FeaturesPageAction>(
+            "Asynchronous find".into(),
```
vorporeal, line 7306: might want to use a different user-facing name here (or description), to tie it back to "why should i enable this?"

e.g.: could make the description "Uses an improved find implementation to keep UI responsive while searching for matches"

app/src/terminal/settings.rs
```diff
@@ -133,6 +134,18 @@ define_settings_group!(TerminalSettings, settings: [
...
+    // Opt-in toggle for running terminal find on a background thread. Only consulted on
+    // channels where `FeatureFlag::AsyncFind` is off; channels with the flag on force the
+    // feature on and hide this toggle. See `is_async_find_enabled` for the composite check.
+    async_find_enabled: AsyncFindEnabled {
+        type: bool,
+        default: false,
+        supported_platforms: SupportedPlatforms::ALL,
+        sync_to_cloud: SyncToCloud::Globally(RespectUserSyncSetting::Yes),
+        private: false,
+        toml_path: "terminal.async_find_enabled",
```
vorporeal: thoughts on making this `experimental.async_find_enabled`?

given the fact that this is only a temporary toggle, i think i prefer the idea of it showing up as:

```toml
[experimental]
async_find_enabled = true
```

vorporeal (approved):

app/src/terminal/settings.rs
vkodithala: Yeah, agreed here and fixed. Not critical for this PR, but maybe in the future if we expect to have more experimental features, we should add a top-level Settings section "Experimental". I opted against this for now, and just included a "BETA" inline with the toggle, but maybe worth considering!

app/src/settings_view/features_page.rs
vkodithala: Good callout, improved the description here!

PR #11573, opened 2026-05-22
Remove prompt requirement from cloud run initiation

harryalbert:
## Description

This is the first step in a broader changeset to allow for local to cloud handoff with no prompt. This PR is completely structural, just changing it so that the prompt in `SpawnAgentRequest` is optional. This is paired with a server-side PR to make the prompt optional on the server as well.

The next step is to actually add the product surface for promptless local to cloud handoff. The overall plan is described in the broad product/tech spec in this PR, and then this changeset is described in `STAGE-1.md` (the PR to follow will have its own `STAGE-2.md`)

## Agent Mode


harryalbert:
* **#11576** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11576?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11573** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11573?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11573?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

specs/REMOTE-1499/PRODUCT.md
```diff
@@ -0,0 +1,56 @@
...
+3. Empty-prompt handoff with snapshotted workspace changes: the client substitutes `"Apply the workspace changes from my previous session."` on the wire alongside the snapshot token. The cloud agent's first user-role turn carries that intent and the snapshot rehydration runs as before. The same string is shown verbatim in the queued-prompt indicator block.
+4. The cloud pane's "setting up…" Cloud Mode Setup V2 UI transitions out properly after environment setup, even when the cloud agent does not fire a first LLM exchange. The viewer learns the setup phase has ended via a new shared-session-protocol marker rather than depending on a first `AppendedExchange` event.
+5. Ships unconditionally on Stage 2 — no new feature flag. The entry points remain gated by `OzHandoff && HandoffLocalCloud` and the source-content guardrail.
+## Non-goals
+- Cloud→cloud empty `Continue` submission (deferred follow-up; out of scope here — see "Deferred follow-ups" below).
```
captainsafia, line 18: I wonder if we actually need prompt-less support for cloud-cloud handoff. The main value prop there would be to allow you to "prewarm" a cloud environment by running all the setup steps then sending a prompt but I could make the case that we should get to a point where setup is quick enough that you don't need to queue it up separately.  
  
I can see the case for prompt-less support for local-cloud to give the feeling that you've transported to a cloud environment without triggering any work but the same doesn't seem to apply for cloud-cloud, IMO. Thoughts?

specs/REMOTE-1499/PRODUCT.md
```diff
@@ -0,0 +1,56 @@
...
+   - Type `/handoff` with no argument.
+   All three funnel through the same `start_local_to_cloud_handoff` path in `app/src/workspace/view.rs:13652-13663` and produce the same wire payload.
+2. Empty-prompt handoff against an in-progress local agent source: the client substitutes `"Continue"` on the wire (local→cloud only) so the cloud agent picks up the conversation context coherently. The same string is shown verbatim in the queued-prompt indicator block — wire and display are coupled by design.
+3. Empty-prompt handoff with snapshotted workspace changes: the client substitutes `"Apply the workspace changes from my previous session."` on the wire alongside the snapshot token. The cloud agent's first user-role turn carries that intent and the snapshot rehydration runs as before. The same string is shown verbatim in the queued-prompt indicator block.
+4. The cloud pane's "setting up…" Cloud Mode Setup V2 UI transitions out properly after environment setup, even when the cloud agent does not fire a first LLM exchange. The viewer learns the setup phase has ended via a new shared-session-protocol marker rather than depending on a first `AppendedExchange` event.
```
captainsafia, line 15: What do we think this marker will look like? I assume it's similar to the one we currently use for environment created?

specs/REMOTE-1499/TECH.md
```diff
@@ -0,0 +1,41 @@
...
+### Entry-point unification + client-side substitution
+Three entry points — the "Hand off to cloud" footer chip, `&` + Enter on an empty buffer, and `/handoff` with no argument — converge in `start_local_to_cloud_handoff` (`app/src/workspace/view.rs`) when the source-content guardrail passes. The guardrail (`source_conversation_has_content` in `app/src/ai/blocklist/handoff/mod.rs`) requires the active source conversation to exist and have at least one exchange. Per-entry-point fallbacks when the guardrail fails are specified in PRODUCT.md and detailed in STAGE-2.md.
+At submit time, `build_handoff_spawn_request` (`app/src/terminal/view/ambient_agent/model.rs`) chooses the wire prompt against the captured `source_conversation_active` bool and the post-upload `InitialSnapshotToken`:
+- empty + in-progress source + snapshot → `"Continue. Apply the workspace changes from my previous session."`
+- empty + in-progress source, no snapshot → `"Continue"`
```
captainsafia, line 19: How do these synthetic prompts appear in the agent conversation that is later triggered from this run?

captainsafia (approved):
The actual code changes here seem legit. I left some non-blocking comments on the spec we're using for this work.

specs/REMOTE-1499/PRODUCT.md
harryalbert: oh I was referring to the shared session event, not a like visual marker. The setup phase concluding is pretty clear because the "Setting up environment..." UI disappears leaving just the regular input. 

This'll be even cleaner when we have the new queued prompt UI, because as part of that change we'll allow the user to send inputs during the setup phase and just queue them up, meaning their first input will still be sent when the env setup is done

specs/REMOTE-1499/PRODUCT.md
harryalbert: yeah, I kind of agree for Oz cloud mode sessions. For non-oz sessions though, I think it is probably a cleaner experience to click continue in the tombstone and immediately kick the session back off. I'm not strongly opinionated there though

specs/REMOTE-1499/TECH.md
harryalbert: They just appear as regular user prompts, which I think makes sense at least as a v0? I don't think there's any reason to hide them and they make clear what's going on in each case (and there's no weird or bulky like system prompt language)

PR #11576, opened 2026-05-22
Update handoff surfaces to allow for empty prompts (and auto-handoff on click)

harryalbert:
## Description

Building off of the changes from the PR below this one in the stack, we now want to change all of the client handoff surfaces to allow for handing off without a prompt (and to take advantage of this instant handoff flow).

Specifically, we change it so that:
* clicking the handoff chip (when the prompt is empty) automatically kicks off handoff
   * if the conversation is in progress when handing off, a "continue" prompt is auto-inserted so the agent continues its work stream in the cloud
   * if the conversation has some snapshots to restore, a prompt is inserted to do snapshot restoration. Otherwise, we would skip snapshot restoration entirely (or do it without a user visible prompt)
   * if neither of these are true, we just handoff with no prompt
* typing /handoff with no prompt also automatically kicks off handoff (with the same prompt semantics as above)
* We listen for the new `AmbientSetupPhaseEnded` session sharing message from [this PR](https://github.com/warpdotdev/session-sharing-protocol/pull/17) to exit the environment setup phase, removing our dependency on a prompt event to do this.

## Testing

### Screenshots / Videos

https://www.loom.com/share/e748b81a3e22418f9e9836acf56c06c1

## Agent Mode


harryalbert:
* **#11576** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11576?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11576?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#11573** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

Cargo.lock
```diff
@@ -11918,7 +11918,6 @@ dependencies = [
 [[package]]
 name = "session-sharing-protocol"
 version = "0.0.0"
-source = "git+https://github.com/warpdotdev/session-sharing-protocol.git?rev=5a0ad6135809feee9da2e9efae8bd6b54b89172e#5a0ad6135809feee9da2e9efae8bd6b54b89172e"
```
harryalbert: I'll update this once the session sharing proto is merged

app/src/terminal/input.rs
```diff
@@ -2559,8 +2559,31 @@ impl Input {
...
                 }
-                AgentInputFooterEvent::OpenHandoffPane => {
-                    me.activate_cloud_handoff_compose(HandoffEntryPoint::FooterChip, ctx);
+                AgentInputFooterEvent::HandoffChipClicked => {
+                    // Auto-handoff only when the input buffer is empty and the
+                    // source conversation has content. Otherwise enter `&`
+                    // compose mode so any in-flight prompt is preserved and
+                    // the user can refine before forking.
+                    #[cfg(all(feature = "local_fs", not(target_family = "wasm")))]
+                    let auto_handoff = me.editor.as_ref(ctx).buffer_text(ctx).trim().is_empty()
```
captainsafia, line 2570: It looks like this check is duplicated across the entrypoints. Cna we add a shared guardrail here? I think the spec mentions something to that effect as well.

app/src/terminal/view/ambient_agent/model.rs
```diff
@@ -627,19 +651,48 @@ impl AmbientAgentViewModel {
         ctx.emit(AmbientAgentViewModelEvent::HandoffSnapshotUploadFailed { error_message });
     }
 
+    /// Build the `SpawnAgentRequest` for a handoff submit. Callers MUST
+    /// normalize an empty prompt to `None` before calling; the empty-prompt
+    /// arms below treat `None` as the empty-prompt signal. All substitutions
+    /// are local-to-cloud-only; the server never sees an in-progress signal
+    /// it has to interpret.
     #[cfg(all(feature = "local_fs", not(target_family = "wasm")))]
     fn build_handoff_spawn_request(
```
captainsafia, line 674: Should we be defensie about the cases where a caller passes in an empty string here?

app/src/terminal/view/ambient_agent/model.rs
```diff
@@ -627,19 +651,48 @@ impl AmbientAgentViewModel {
...
         ctx: &AppContext,
     ) -> SpawnAgentRequest {
         let config = Some(self.build_default_spawn_config(ctx));
-        let (prompt, mode) = extract_user_query_mode(prompt);
+        let has_snapshot_content = initial_snapshot_token
+            .as_ref()
+            .is_some_and(|token| !token.as_str().is_empty());
+
+        let raw_wire_prompt = match (prompt, should_inject_continue, has_snapshot_content) {
+            (Some(p), _, _) => Some(p),
```
captainsafia, line 690: Nit: these string literals feel important enough in this implementation that they might merit their own constants?

captainsafia (approved):
Looks good overall! Left some comments inline.

app/src/terminal/view/ambient_agent/model.rs
harryalbert: yeah good call

app/src/terminal/input.rs
harryalbert: can do

app/src/terminal/view/ambient_agent/model.rs
harryalbert: yeah good call, makes sense

Issue #9159, opened 2026-04-21
Mac editing shortcuts (cmd+arrows, cmd+delete, option+delete) not forwarded to TUI apps using Kitty keyboard protocol


Mamdouh66:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [x] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [x] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When a TUI app enables the Kitty keyboard protocol (CSI u) for its input, Warp does not forward standard macOS editing shortcuts to the app. The same shortcuts work correctly in Ghostty and Kitty with the same app and the same protocol enabled, so this appears to be a Warp-side encoding gap.

### To reproduce

1. Install Claude Code (npm i -g @anthropic-ai/claude-code)
2. Run with the fullscreen renderer enabled: CLAUDE_CODE_NO_FLICKER=1 claude
3. Type some text into the prompt
4. Try the standard Mac editing shortcuts:
  - cmd+← / cmd+→ — jump to start/end of line
  - cmd+delete — delete whole line
  - option+← / option+→ — jump by word
  - option+delete — delete previous word

### Expected behavior

Each shortcut performs its standard macOS line-editing action inside the Claude Code prompt (as it does in Warp's normal shell prompt, and as it does in Ghostty when running the same Claude Code binary in NO_FLICKER mode).

Actual

The shortcuts are either silently dropped or produce no effect inside the TUI's input. The app's input handler never receives a usable keypress. Falling back to raw control sequences (Ctrl+A, Ctrl+E, Ctrl+U, Ctrl+W) works, confirming the app's input layer is fine — Warp simply isn't translating the Mac shortcuts into the Kitty-encoded sequences the app is listening for

### Operating system (OS)

macOS

### Operating system and version

26.3.1 (a) (25D771280a)

### Shell Version

zsh 5.9 (arm64-apple-darwin25.0)

[...]

Issue #2686, opened 2023-02-22
The "Option" key on Mac not working properly when using with Neovim open in Warp


nguyenanhhao221:
### Describe the bug

Normally I would use the "Option" key to jump each word, or delete a word. Which it working really good in Warp without any configuration. 

But if I use warp to open an editor like Vim or Neovim. This function stop working and you cannot delete the entire word, jump in each word. Even the "Cmd" and "Backspace" don't delete the whole line

### To Reproduce

1. Open Warp
2. Run any file with vim or neovim
3. Press "i" to enter "Insert mode" for vim / neovim
4. Type words, test with Options and Cmd button

### Expected behaviour

Should behave normally for options and cmd button even when in Vim/Neovim

### Operating System

MacOS

### Does this block you from using Warp daily?

Yes, this issue prevents me from using Warp daily.

### Warp Internal (ignore): linear-label:b8107fdf-ba31-488d-b103-d271c89cac3e

None

dannyneira:
Hi @nguyenanhhao221 Thanks for letting us know, please try enabling the Option keys are Meta in the `Settings > Features > Keys`, and see if that helps with the neovim/vim issue.

![CleanShot 2023-02-22 at 15 38 17](https://user-images.githubusercontent.com/16809145/220776466-aba166ba-f1ce-40fd-b120-9606b17b8118.jpg)

fveauvy:
It worked for me.

Unfortunately once `Left Option key is Meta` was enable I encountered issue #2364.

iyogeshjoshi:
I have tried it, but it didn't work for me, and I'm using `v0.2024.02.13.08.02.stable_00` version. Though it's working fine on Mac terminal and iTerm2

igorgn:
this is reproducible for me on Neovim `0.11.6` 
`Left Option Key is Meta` _disabled_:  simple Backspace
`Left Option Key is Meta` _enabled_: changes behavior to simple ESC (back to Normal mode)

Warp `v0.2026.02.25.08.24.stable_01`

PR #11605, opened 2026-05-23
Fix line-editing shortcuts in alt-screen TUIs

### LINKED ISSUE #807: OPTION/CMD + ARROW LEFT/RIGHT or BACKSPACE not working as intended when Warp is in long-running command mode
### Describe the bug

```shell
Whenever I elevate my rights to root through SSH I cannot use either the OPTION or CMD with ARROW LEFT/RIGHT or  to either skip the whole word, skip to the end/start of the line or deleting the next/previous word or the whole line.
```

### To Reproduce

Steps to reproduce:
1. SSH to a RHEL based server(havn't tested on anything else) as normal user(needs wheel permissions)
2. Execute `sudo su -` to elevate to root
3. Write something like "asdf qwerty"
4. Try to use 'CMD+ARROW-LEFT' 
5. You can also try to 'CMD+BACKSPACE'
6. Also you can try to 'OPTION+ARROW-LEFT'
7. You could also move the cursor to the start of the line and do the commands with ARROW-RIGHT or DELETEKEY

### Expected behaviour

```shell
4. When having the cursor at the end of the line this should skip both words and get back to the start of the line
5. When having the cursor at the end of the line this should delete the whole line
6. When having the cursor at the end of the line this move the cursor to the first letter in the second word(in my example it should move the cursor in front of q)
```

### Screenshots

_No response_

### Operating System

MacOS

### OS Version

12.2.1

### Additional context

_No response_

mohamedmagdy17593:
## Description

When a readline-style TUI runs in alt-screen mode (Claude Code fullscreen, opencode, vim insert mode, etc.), `cmd+Left` / `cmd+Right` / `cmd+Backspace` / `cmd+Delete` / `alt+Backspace` no longer perform their line-editing actions on macOS — they fall through to plain bytes.

**Root cause:** the `EditableBinding`s in `view/init.rs` were gated on `LongRunningCommand`, which is mutually exclusive with `AltScreen` (`view.rs:27201-27209`). The fallback encoder wildcards the `cmd` modifier (`escape_sequences.rs:301-340`), so `cmd` was silently dropped.

**Fix:** extend the predicate on five bindings to `(LongRunningCommand | AltScreen)` — same pattern already used by `shift-tab` at `init.rs:545`.

`alt-left` / `alt-right` predicates are intentionally unchanged, so the encoder path keeps emitting `ESC[1;3D` / `ESC[1;3C` for vim, helix, and KKP-aware TUIs (no behavior change for those users).

## Linked Issue

Fixes #807 (open since 2022, labeled `ready-to-implement`)
Refs #2686, #9159

## Testing

Verified inside Claude Code `/tui fullscreen`:

| Key | Before | After |
|---|---|---|
| `cmd+Left` | `ESC[D` (1 char left) | `0x01` (jump to start) |
| `cmd+Right` | `ESC[C` (1 char right) | `0x05` (jump to end) |
| `cmd+Backspace` | `0x7f` (delete 1) | `0x15` (delete to line start) |

`alt+Left` / `alt+Right` bytes unchanged.

https://github.com/user-attachments/assets/8657e489-4126-44eb-847c-143e5f81d4a6

## Agent Mode

CHANGELOG-BUG-FIX: Fixed macOS line-editing shortcuts (cmd+arrows, cmd+backspace, cmd+delete, alt+backspace) not working inside fullscreen TUIs like Claude Code and opencode.

### DIFF
--- a/app/src/terminal/view/init.rs
+++ b/app/src/terminal/view/init.rs
@@ -500,14 +500,18 @@ pub fn init(app: &mut AppContext) {
         // We already have bindings for home/end (the keybindings for this on Linux and Mac) that
         // send the correct control sequence to the PTY.
         .with_mac_key_binding("cmd-left")
-        .with_context_predicate(id!("Terminal") & !id!("IMEOpen") & id!("LongRunningCommand")),
+        .with_context_predicate(
+            id!("Terminal") & !id!("IMEOpen") & (id!("LongRunningCommand") | id!("AltScreen")),
+        ),
         EditableBinding::new(
             "terminal:executing_command_move_cursor_end",
             "Move cursor end within an executing command",
             TerminalAction::ControlSequence(vec![escape_sequences::C0::ENQ]),
         )
         .with_mac_key_binding("cmd-right")
-        .with_context_predicate(id!("Terminal") & !id!("IMEOpen") & id!("LongRunningCommand")),
+        .with_context_predicate(
+            id!("Terminal") & !id!("IMEOpen") & (id!("LongRunningCommand") | id!("AltScreen")),
+        ),
         EditableBinding::new(
             "terminal:executing_command_delete_word_left",
             "Delete word left within an executing command",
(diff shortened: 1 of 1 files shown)


mohamedmagdy17593:
@cla-bot check

mohamedmagdy17593:
https://github.com/user-attachments/assets/8657e489-4126-44eb-847c-143e5f81d4a6

/oz-review

mohamedmagdy17593:
Valid concern — bindings fire before the KKP encoder (`warpui_core/src/core/app.rs:1985` runs the matcher first; `alt_screen_element.rs:869` only reaches `to_escape_sequence` if nothing matched).

This isn't new though — the existing `shift-tab` binding at `app/src/terminal/view/init.rs:546-557` uses the same `(LongRunningCommand | AltScreen)` predicate and has the same effect (overrides the `CSI 9;2u` that KKP would have produced per `kitty_keyboard_protocol.rs:42-44`). We're matching the established pattern, not introducing a new hazard.

Zero impact on the motivating bug (#807, anthropics/claude-code#58796): those TUIs don't enable KKP.

Happy to follow up with a separate PR that adds a `KittyKeyboardProtocol` context tag and gates all the affected bindings (including the existing `shift-tab`) with `& !id!("KittyKeyboardProtocol")` — feels like a cleaner home for that since the scope is broader than these five predicates.

PR #11621, opened 2026-05-24
Double-click the hidden-section bar to fully expand

### LINKED ISSUE #11622: Double-click the hidden-section bar to fully expand

When the Code Review panel's diff editor has a large hidden context section, the user has to click the gutter expand button multiple times — once per 25 lines — to reveal the whole thing. Add double-click as a shortcut: double-clicking the gutter expand button fully expands the hidden section in one action.

Single-click keeps the existing chunked behavior so users can still choose to reveal a little at a time.

For hunks separated by hundreds of unchanged lines (or hidden regions at the top/bottom of a file), users currently click 4–10 times to see all the context between or around hunks. There's no keyboard shortcut and no "expand all" affordance at the gutter level.

### Additional context

- macOS / Windows / Linux all conventionally use double-click for "do the maximum version of this gesture" on similar controls.
- The chunked single-click behavior should remain as the granular option.
- Adjacent gutter interactions (diff-hunk sliver toggle, add-context, revert) shouldn't change.

### Operating system (OS)

macOS, Windows, Linux

3 (Moderate)

bradleyjames:
## Description

The diff editor collapses long runs of unchanged context into **hidden sections**, each rendered as a full-width bar with one or two gutter expand buttons (`▲`/`▼`) beside it. Single-clicking a gutter button reveals ~25 lines at a time, so a large hidden region takes many clicks to fully reveal.

This PR makes **double-clicking the bar** fully expand the entire hidden section in one action. Each gesture gets its own target — the gutter arrows keep their instant chunk-at-a-time behavior, and the full-width bar handles "reveal everything." Because the gestures live on separate targets, single-click stays instant with **no debounce** (the established VS Code / Monaco `hideUnchangedRegions` pattern).

Also:
- A single click on the bar is now a no-op — in particular it no longer clamps to a text offset and starts a text selection.
- The bar is discoverable: pointer cursor, a hover highlight, and a "Double-click to expand" tooltip on hover.
- Adjacent gutter interactions (diff-hunk sliver toggle, add-as-context, revert) are unchanged.

**Implementation:** `RenderableHiddenSection` (the bar) becomes a `Hoverable` that dispatches a new `RichTextAction::hidden_section_double_clicked`, reusing the existing `HiddenSectionExpansion { ExpansionType::Both }` expansion path. `HiddenBlockConfig` now carries a `MouseStateHandle` (mirroring `BlockItem::TaskList`) so the bar's hover state persists across re-layouts. No debounce, no platform double-click interval, no `DeferredCall`.

Specs: `specs/GH11622/product.md`, `specs/GH11622/tech.md`.

## Linked Issue

Closes #11622

## Testing

Automated coverage:
- **Unit tests** for `RenderContentTreeRef::first_hidden_section_line_range` (the bar's range resolution) in `crates/editor/src/render/model/mod_tests.rs` — including the no-hidden-section case.
- **Integration test** `test_code_review_double_click_fully_expands_hidden_section` (`crates/integration/src/test/code_review.rs`): boots a real Warp instance, sets up a git repo whose diff produces hidden sections, opens the code review panel, and fully expands the first hidden section via the same `Hidden`-block → range → `Both` path the bar's double-click uses, asserting exactly one hidden section is removed (a chunked reveal would leave the count unchanged). Size-agnostic, so it also covers small sections.

Manual coverage:
- Double-click the bar → full expansion in one transition, no intermediate chunk.
- Single-click the bar → no-op, no text selection.
- Single-click a gutter arrow → instant chunked reveal (unchanged).
- Hover the bar → pointer cursor, hover highlight, and the "Double-click to expand" tooltip.
- Adjacent gutter interactions (sliver toggle, add-as-context, revert) unchanged.

Demo recording of the bar double-click + tooltip to be attached:
https://github.com/user-attachments/assets/83f84370-9fae-403f-a14e-2b5a9faaf403

Demo recording of the single click - passivity:
https://github.com/user-attachments/assets/ed629048-fb59-4267-8f1d-a69472a90859

## Agent Mode

CHANGELOG-IMPROVEMENT: Double-click a collapsed hidden section in the diff editor to fully expand it in one action.

### DIFF
--- a/crates/editor/src/render/element/hidden_section.rs
+++ b/crates/editor/src/render/element/hidden_section.rs
@@ -3,10 +3,12 @@ use std::ops::Range;
 use warp_core::ui::appearance::Appearance;
 use warp_core::ui::theme::color::internal_colors;
 use warpui_core::elements::{
-    Container, CrossAxisAlignment, Empty, Flex, Hoverable, MouseStateHandle, ParentElement,
+    ChildAnchor, Container, CrossAxisAlignment, Empty, Flex, Hoverable, MouseStateHandle,
+    OffsetPositioning, ParentAnchor, ParentElement, ParentOffsetBounds, Stack,
 };
 use warpui_core::geometry::vector::vec2f;
 use warpui_core::platform::Cursor;
+use warpui_core::ui_components::components::UiComponent;
 use warpui_core::{
     AfterLayoutContext, AppContext, Element, LayoutContext, SingletonEntity, SizeConstraint,
     WeakViewHandle,
@@ -40,19 +42,43 @@ impl RenderableHiddenSection {
         let theme = appearance.theme();
         let base_background = internal_colors::fg_overlay_1(theme);
         let hover_background = internal_colors::fg_overlay_2(theme);
+        let ui_builder = appearance.ui_builder();
 
         let element = Hoverable::new(mouse_state, move |state| {
             let row = Flex::row()
                 .with_child(Empty::new().finish())
                 .with_cross_axis_alignment(CrossAxisAlignment::Center);
-            let background = if state.is_hovered() {
+            let hovered = state.is_hovered();
+            let background = if hovered {
                 hover_background
             } else {
                 base_background
             };
-            Container::new(row.finish())
+            let bar = Container::new(row.finish())
                 .with_background(background)
-                .finish()
+                .finish();
+
+            if !hovered {
+                return bar;
+            }
+
+            // On hover, float a tooltip explaining the double-click gesture,
+            // centered just below the bar (mirrors how `Button` shows its tooltip).
+            let mut stack = Stack::new().with_child(bar);
+            let tooltip = ui_builder
+                .tool_tip("Double-click to expand".to_string())
+                .build()
+                .finish();
+            stack.add_positioned_overlay_child(
+                tooltip,
+                OffsetPositioning::offset_from_parent(
+                    vec2f(0., 8.),
+                    ParentOffsetBounds::WindowByPosition,
+                    ParentAnchor::BottomMiddle,
+                    ChildAnchor::TopMiddle,
+                ),
+            );
+            stack.finish()
         })
         // A single click on the bar does nothing, but registering the handler makes the
         // Hoverable consume the press so it does not fall through to text selection.
--- a/crates/editor/src/render/element/hidden_section.rs
+++ b/crates/editor/src/render/element/hidden_section.rs
@@ -1,4 +1,5 @@
 use std::ops::Range;
+use std::time::Duration;
 
 use warp_core::ui::appearance::Appearance;
 use warp_core::ui::theme::color::internal_colors;
@@ -48,8 +49,9 @@ impl RenderableHiddenSection {
             let row = Flex::row()
                 .with_child(Empty::new().finish())
                 .with_cross_axis_alignment(CrossAxisAlignment::Center);
-            let hovered = state.is_hovered();
-            let background = if hovered {
+            // The hover highlight tracks the mouse immediately, while the tooltip
+            // below is gated on `is_hovered()` so it respects the hover-in delay.
+            let background = if state.is_mouse_over_element() {
                 hover_background
             } else {
                 base_background
@@ -58,7 +60,7 @@ impl RenderableHiddenSection {
                 .with_background(background)
                 .finish();
 
-            if !hovered {
+            if !state.is_hovered() {
                 return bar;
             }
 
@@ -66,7 +68,7 @@ impl RenderableHiddenSection {
             // centered just below the bar (mirrors how `Button` shows its tooltip).
             let mut stack = Stack::new().with_child(bar);
             let tooltip = ui_builder
-                .tool_tip("Double-click to expand".to_string())
+                .tool_tip("Double-click to expand all lines".to_string())
                 .build()
                 .finish();
             stack.add_positioned_overlay_child(
@@ -91,6 +93,7 @@ impl RenderableHiddenSection {
                 ctx.dispatch_typed_action(action);
             }
         })
+        .with_hover_in_delay(Duration::from_millis(500))
         .with_cursor(Cursor::PointingHand)
         .finish();
 
--- a/crates/editor/src/render/element/hidden_section.rs
+++ b/crates/editor/src/render/element/hidden_section.rs
@@ -1,14 +1,17 @@
 use std::ops::Range;
+use std::sync::Arc;
 use std::time::Duration;
 
 use warp_core::ui::appearance::Appearance;
 use warp_core::ui::theme::color::internal_colors;
 use warpui_core::elements::{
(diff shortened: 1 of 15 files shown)


bradleyjames:
@oz-agent here's the issue and yes, it's not marked as ready-to-implement yet: https://github.com/warpdotdev/warp/issues/11622

bradleyjames:
@oz-agent assuming the issue will be flipped to ready-to-implement soon, can you review the PR?

crates/integration/src/test/code_review.rs
```diff
@@ -650,3 +651,61 @@ pub fn test_code_review_scroll_preserved_second_file() -> Builder {
...
+            TestStep::new("Wait for the code review panel to load file diffs")
+                .set_timeout(Duration::from_secs(20))
+                .add_assertion(assert_code_review_loaded()),
+        )
+        .with_step(
+            TestStep::new("Wait for the diff to collapse unchanged context into hidden sections")
+                .set_timeout(Duration::from_secs(20))
+                .add_assertion(assert_min_hidden_sections(TEST_FILE_NAME, 1)),
+        )
+        .with_step(expand_first_hidden_section_and_assert_full_reveal(
```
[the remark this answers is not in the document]
bradleyjames: This is a deliberate tradeoff. The integration test asserts the expand *outcome* through the same `Hidden`-block → full range → `ExpansionType::Both` resolution the bar's `on_double_click` produces (`first_hidden_section_line_range` → `expand_hidden_section`), and there's a unit test for that resolution.

Driving the literal `Hoverable::on_double_click` event would require giving the bar a registered hit-test position (a `SavePosition` wrap) plus a synthetic double-click step — infra we don't use for any other plain `Hoverable`/`Button` interaction in the integration suite. The only hover/click tooltip tests here target features that expose explicit view-state (e.g. `is_secret_tooltip_open`, `get_command_x_ray`), which this bar intentionally doesn't. The "`on_double_click` fires on `click_count == 2`" behavior itself is `Hoverable` framework logic with its own coverage, and the end-to-end gesture was verified manually (recording on the PR).

So I left the event-path test out to avoid bespoke scaffolding for framework glue. Happy to add it if you'd prefer the extra regression guard on the wiring.

kevinchevalier (approved):
Thanks for implementing this.

bradleyjames:
@kevinchevalier , had to push some fixes (e.g. formatting and master was broken so had to rebase on the fix).

crates/editor/src/render/element/hidden_section.rs
```diff
@@ -1,36 +1,98 @@
...
+            }
 
-        let element = Container::new(row.finish())
-            .with_background(internal_colors::fg_overlay_1(theme))
-            .finish();
+            // On hover, float a tooltip explaining the double-click gesture,
+            // centered just below the bar (mirrors how `Button` shows its tooltip).
+            let mut stack = Stack::new().with_child(bar);
+            let tooltip = ui_builder
+                .tool_tip("Double-click to expand".to_string())
```
bholmesdev:
```suggestion
                .tool_tip("Double-click to expand all lines".to_string())
```

This is a bit clearer about what will happen

crates/editor/src/render/element/hidden_section.rs
```diff
@@ -1,36 +1,98 @@
...
+                .build()
+                .finish();
+            stack.add_positioned_overlay_child(
+                tooltip,
```
bholmesdev: I would also prefer a delay on when the tooltip appears, rather than showing right away

crates/editor/src/render/element/hidden_section.rs
```diff
@@ -1,36 +1,98 @@
...
+        parent_view: WeakViewHandle<V>,
+        app: &AppContext,
+    ) -> Self {
         let appearance = Appearance::as_ref(app);
         let theme = appearance.theme();
+        let base_background = internal_colors::fg_overlay_1(theme);
+        let hover_background = internal_colors::fg_overlay_2(theme);
+        let ui_builder = appearance.ui_builder();
+
+        let element = Hoverable::new(mouse_state, move |state| {
```
bholmesdev, line 59: I appreciate the inclusion of a cursor pointer, but it made me realize that we do _not_ use cursor pointer for the original button that expands by increments of 25 lines. For consistency, I'd like to implement there as well

bholmesdev (changes requested):
I have a few nits above, but my primary concern is the performance degradation on-scroll when expanding a larger diff. This example expands 1000 lines, and creates noticeable frame drops while scrolling that particular diff. I think we'd need to explore renderer improvements for the code review panel before feeling comfortable merging this.

https://github.com/user-attachments/assets/b46a0726-f5ce-4282-96f8-5969d2ce77aa

crates/editor/src/render/element/hidden_section.rs
bradleyjames: done

crates/editor/src/render/element/hidden_section.rs
bradleyjames: done

crates/editor/src/render/element/hidden_section.rs
bradleyjames: done

bradleyjames:
> I have a few nits above, but my primary concern is the performance degradation on-scroll when expanding a larger diff. This example expands 1000 lines, and creates noticeable frame drops while scrolling that particular diff. I think we'd need to explore renderer improvements for the code review panel before feeling comfortable merging this.
> 
>  CleanShot.2026-06-22.at.21.18.16.mp4

Is the issue created by this feature or making it more easily to observe?

bholmesdev:
Thanks for those fixes, and sorry for the delay! Spoke to @kevinyang372 about the performance issues and agreed it shouldn't block this feature since it's a pre-existing issue. I also removed the hover bg color on the bar, since it looked a bit awkward visually (see below). The cursor change and tooltip remain. This seems good to merge.

As a follow-up, I'm making a PR to add a "show all lines" button (or equivalent) to the code review overflow menu. This should be more discoverable than the double-click, and it'll offer a way to see all full files at once.

image

bholmesdev (approved):

bradleyjames:
> As a follow-up, I'm making a PR to add a "show all lines" button (or equivalent) to the code review overflow menu. This should be more discoverable than the double-click, and it'll offer a way to see all full files at once.

Very cool and thank you.  Yeah, I normally just want everything expanded when reviewing code.

bholmesdev:
https://github.com/user-attachments/assets/4c46e60e-44b8-483f-b601-efcdbde6cb66

Further UX tweak: decided to use an explicit hyperlink callout (with a tooltip still) rather than a double-click for clarity. This is in-line with how the Codex app handles expansion.

As for the "expand all" button I noted earlier: There's more UX concerns than I initially thought. I built a prototype and realized there will be confusion with scroll position when hitting "expand all." It's also not terribly useful without navigation options to step through the changes. This would need a wider rethink of the code review panel.

Thinking this change will make the code review panel easier to use for now.

PR #11655, opened 2026-05-25
fix: Preserve custom models during LLM preference reconciliation

KelvinJRosado:
## Description
With the current logic, only the built-in models are checked when preferences are updated/validated (such as on app startup). As a result, setting a custom model as the default will only persist until this preference update runs. This PR fixes this by extending that model check to also include custom models that the user defined. 

## Linked Issue
Closes https://github.com/warpdotdev/warp/issues/8562 and https://github.com/warpdotdev/warp/issues/11564

## Testing
I tested manually by adding a custom BYOK model and verifying that I could use it for a new agent session. Then I quit the app and re-opened to verify that my default model was still set to my custom BYOK model.

- [ x] I have manually tested my changes locally with `./script/run`

### Updated Profile With Custom Default Model

image

image

### State After Quitting and Re-Opening

image

PR #11656, opened 2026-05-25
Preserve custom endpoint model selections in profiles

ildunari:
## Summary
- keep saved execution-profile model ids when custom inference is temporarily unavailable
- continue clearing only transient pane-level custom model overrides in that state
- rely on existing runtime model resolution to fall back safely until custom models are available again

## Why
Custom endpoint model selections are stored in execution profiles as local custom model ids. During workspace/team updates, custom inference can temporarily appear unavailable while entitlements or endpoint data are still loading. The previous sanitizer wrote `None` into every profile field that referenced a custom model, permanently resetting the profile back to the default agent model, usually `auto`.

Preserving the profile value avoids losing the user's selection. The existing `get_preferred_*_model` paths already gate custom model resolution on custom inference availability, so unavailable custom models still fall back safely at runtime without mutating the saved profile.

## Test plan
- `cargo fmt --check`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer cargo test -p warp ai::llms_tests`

Note: the first test attempt failed because local `xcode-select` points at missing `/Applications/Xcode-beta.app/Contents/Developer`; reran with installed `/Applications/Xcode.app/Contents/Developer`.

### DIFF
--- a/app/src/ai/llms.rs
+++ b/app/src/ai/llms.rs
@@ -859,8 +859,6 @@ impl LLMPreferences {
             .map(|info| info.id.clone())
             .collect();
         let mut updated_agent_mode = false;
-        let mut updated_coding = false;
-        let mut updated_other = false;
 
         self.base_llm_for_terminal_view.retain(|_, id| {
             let keep = !custom_ids.contains(id);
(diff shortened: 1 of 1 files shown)


gbks121:
Looks like it is related to https://github.com/warpdotdev/warp/pull/11655

PR #11671, opened 2026-05-25
Prototype of background, per-window computer use on macOS.

vorporeal:
## Description

Why: we want the agent to be able to drive a *specific* background window without raising it or stealing the cursor/focus, and to reason about which window to act on next. That requires a per-action notion of "what surface am I targeting", window-scoped screenshots the agent can map coordinates onto, and a window list it can pick from — none of which the screen-only API could express.

What this wires up (additive/optional):
- `computer_use` crate: a public `Target` (`Screen | Window { window_id, pid }`), carried per-action via `TargetedAction` and per-screenshot via `ScreenshotParams.target`, threaded through `perform_actions`. Adds `WindowInfo`/`CapturedWindow`, `enumerate_windows()`, and `background_supported()`.
- macOS: per-action routing (HID tap for screen, background per-window delivery for a window), window-scoped capture via `screencapture -l <id>` (returns pixel size + scale), and on-screen window enumeration (id, pid, app, title, bounds, layer). For a `Window` target, incoming coordinates are interpreted as window-local pixels in the captured screenshot and remapped to global physical pixels before delivery.
- `ai` + `app`: new `target` flows from the proto `ComputerUseTarget` into the action pipeline; results now carry a refreshed `windows` list plus `captured_*` metadata, and `RequestComputerUse` reports `background_supported`. Conversation restore round-trips the new fields.

Background delivery (public APIs only): events are posted straight to the target process with `CGEventPostToPid`, window-addressed, so they never enter the global event stream or move the real cursor. To make an inactive window accept them, we reproduce what a first click on an inactive window does — putting it into the AppKit-active input state — but without raising the app:
- Per-process event taps (`CGEventTapCreateForPid`) on the previously-frontmost app drop the focus-change messages macOS would use to switch the user's frontmost application.
- An `appKitDefined` activation primer plus a single center-of-window "primer" click (absorbed by the activation flow rather than firing a UI action) flip the window into the active input state.
- Activation is cached in a process-global registry keyed by `(pid, window)`, so a session activates a window once and reuses it across later turns without re-sending the center click; concurrent sessions targeting different windows coexist.

This replaces an earlier prototype that relied on private/undocumented SkyLight APIs; the delivery and activation paths above use only public AppKit/Core Graphics/Core Foundation APIs.

The approach is adapted from the one described here: https://x.com/bridge_surf/status/2057416247319618039

Notes / assumptions (prototype scope):
- Coordinate remap assumes the target window is on the main display and shares its backing scale; multi-display / mixed-scale handling is a follow-up (noted in code).
- Headless-safe: the activation, keycode-translation, and display-scale paths run inline on the calling thread and never `dispatch_sync` to the main dispatch queue. In headless CLI agent runs the main thread blocks on a channel and never services that queue, so a synchronous main-queue dispatch from a worker thread would deadlock. `main_display_scale_factor` computes the backing scale via Core Graphics, and the keycode cache serializes Carbon calls through a process-wide mutex.
- macOS only; Linux/Windows/noop actors accept the new API but ignore window targeting.

## Testing
- `cargo check -p computer_use`, `-p ai`, `-p warp` all pass.
- `cargo clippy -p computer_use -p ai` (default features → macOS path), `-p computer_use --bins`, and `-p warp` all pass with `-D warnings`.
- `cargo fmt -- --check` is clean.
- Manually validated end-to-end on macOS in a headless agent run: background mouse movement, clicks, and keyboard input drive a non-frontmost window without raising it, stealing the cursor, or deadlocking.

_Conversation: https://staging.warp.dev/conversation/a4df24cd-cdbf-4ef9-988a-332c10c8c492_

_Run: https://oz.staging.warp.dev/runs/019e6149-28e6-73ef-ba18-d5048032054c_


vorporeal:
* **#11671** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11671?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11671?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/computer_use/src/mac/mod.rs
```diff
@@ -1,28 +1,124 @@
...
+/// For a `Window` target the incoming coordinates are window-local pixels in the captured window
+/// screenshot's space; they are translated to global points using the backing scale of the
+/// display containing the window, then encoded for the existing screen-pixel mouse pipeline.
+/// `Screen` targets retain the legacy global-pixel behavior.
+fn remap_action_for_target(action: &Action, target: Target) -> Result<Action, String> {
+    let Target::Window { window_id, .. } = target else {
+        return Ok(action.clone());
+    };
+    let remap = |p: Vector2I| -> Result<Vector2I, String> {
+        let info = window::window_by_id(window_id)
```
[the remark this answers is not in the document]
vorporeal: Fixed in `pr review`. A `Target::Window` with `window_id == 0` is now rejected at the actor boundary in `perform_actions` with an explicit error, rather than failing later in `window_by_id(0)`. `0` was only ever an "unknown" sentinel (the CLI default, and unparseable wire ids); a well-behaved caller always echoes a real id from the enumerated window list, so the documented PID+coordinate contract has been removed. Also updated the `Target::Window.window_id` doc, and the `use_computer` CLI now requires `--window-id` whenever `--pid` is given.

crates/computer_use/src/mac/mod.rs
```diff
@@ -35,11 +131,30 @@ impl super::Actor for Actor {
...
+                targeted.target
+            } else {
+                Target::Screen
+            };
+
+            // Route this action to its target: the HID tap for screen actions, or directly to the
+            // owning process for a window action (without raising it or moving the cursor).
+            let post_target = post_target_for(target);
+            self.mouse.set_target(post_target);
+            self.keyboard.set_target(post_target);
```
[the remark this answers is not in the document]
vorporeal: Fixed in `pr review`. Window-targeted keyboard actions now activate the target window before posting key events. Added `activate_window_for_keyboard`, which resolves the window via `window_by_id` and calls `activation::ensure_activated` before `TypeText` and `KeyDown`, mirroring the lazy activation the mouse path performs from the event location. It is idempotent, so it does not re-send the activation primer across keystrokes or turns.

crates/computer_use/src/mac/mouse.rs
```diff
@@ -161,28 +213,204 @@ impl Mouse {
...
+    use std::ffi::c_void;
+    use std::sync::OnceLock;
+
+    type SetWindowLocationFn = unsafe extern "C" fn(*mut c_void, CGPoint);
+    // The macOS value of `RTLD_DEFAULT`, used to search all loaded images for the symbol.
+    const RTLD_DEFAULT: *mut c_void = -2isize as *mut c_void;
+
+    static RESOLVED: OnceLock<Option<SetWindowLocationFn>> = OnceLock::new();
+    let resolved = RESOLVED.get_or_init(|| unsafe {
+        let sym = libc::dlsym(RTLD_DEFAULT, c"CGEventSetWindowLocation".as_ptr());
```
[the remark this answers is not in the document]
vorporeal: Keeping the private `CGEventSetWindowLocation` — there is no public alternative for this delivery path. There is no public CGEvent field or setter for the window-local location, and on the `postToPid` path AppKit reads that field directly because the WindowServer does no hit-testing. Every independent implementation of background window-targeted posting resolves the same private symbol via `dlsym`, alongside the private fields 51/58 already used here. The reliance on these private event-routing details is documented in the `set_window_location` doc comment.

crates/computer_use/src/mac/mod.rs
danielpeng2, line 194: we don't need to activate the window for KeyUp events?

crates/computer_use/src/mac/mod.rs
```diff
@@ -58,9 +201,11 @@ impl super::Actor for Actor {
                     self.mouse.scroll(direction, distance)?;
                 }
                 Action::TypeText { text } => {
+                    activate_window_for_keyboard(target);
                     self.keyboard.type_text(text)?;
                 }
                 Action::KeyDown { key } => {
+                    activate_window_for_keyboard(target);
                     self.keyboard.key_down(key)?;
```
danielpeng2, line 191: nit: since the keyboard is target-aware, it might make more sense to have `keyboard.type_text` and `keyboard.key_down` handle this implicitly?

crates/computer_use/src/lib.rs
```diff
@@ -45,14 +45,120 @@ pub fn create_actor() -> Box<dyn Actor> {
...
+#[cfg(macos)]
+pub fn experimental_list_windows() -> Result<String, String> {
+    Ok(imp::list_windows())
+}
+
+/// Experimental: lists on-screen windows. Unsupported on this platform.
+#[cfg(not(macos))]
+pub fn experimental_list_windows() -> Result<String, String> {
+    Err("Window listing is only supported on macOS.".to_string())
+}
```
danielpeng2: what's the difference between this and `enumerate_windows`? can we just use `enumerate_windows` instead?

danielpeng2 (approved):

crates/computer_use/src/mac/mod.rs
vorporeal: No — `KeyUp` always follows a `KeyDown` that already triggered activation, so by the time a `KeyUp` arrives the window is already active. A lone `KeyUp` with no prior `KeyDown` to that window would be a no-op regardless. Added a comment to `key_up` in `keyboard.rs` documenting this.

crates/computer_use/src/mac/mod.rs
vorporeal: Good call. Moved activation into `Keyboard` to match the pattern in `Mouse` (which calls `activation::ensure_activated` from inside `post_event`). `Keyboard` now stores `window_context: Option<(u32, pid_t)>` alongside `post_target`, populated by `set_target` (which now accepts `Target` instead of `PostTarget`). `ensure_window_activated` is called at the start of `key_down` and `type_text`; `mod.rs` no longer has any activation logic for keyboard events.

crates/computer_use/src/lib.rs
vorporeal: `experimental_list_windows` returns window bounds (x, y, width, height) in addition to the fields in `WindowInfo`, which is why it uses a separate type and code path. We considered unifying by adding bounds to `WindowInfo`, but `WindowInfo` is converted to/from the `api::WindowInfo` proto where bounds are not represented, so proto-converted instances would always have zero bounds. Keeping the two separate (slim `WindowInfo` for targeting/wire, richer diagnostic type for the CLI) avoids that awkwardness. Updated the `experimental_list_windows` doc comment to call this out.

PR #11691, opened 2026-05-26
fix: suppress input for github action session shares

captainsafia:
## Description
Suppresses the universal input and cloud continuation UI for ambient session shares whose task source is `AgentSource::GitHubAction`.

Changes include:
- Tracking ambient task source on the ambient agent view model when viewing existing sessions.
- Treating GitHub Action-sourced tasks as unsupported for cloud continuation/follow-up UI.
- Hiding the terminal input box for GitHub Action ambient sessions.
- Clearing stale tombstone/follow-up state if a session resolves as GitHub Action-sourced.

## Linked Issue

Slack request: universal input should not show on GitHub Action-sourced session shares, and hand-off follow-ups should not trigger.

## Testing
- [x] `cargo +1.92.0 fmt --manifest-path /workspace/warp/Cargo.toml -p warp`
- [x] `cargo +1.92.0 test --manifest-path /workspace/warp/Cargo.toml -p warp github_action -- --nocapture`
- [x] `cargo +1.92.0 test --manifest-path /workspace/warp/Cargo.toml -p warp on_session_share_ended -- --nocapture`
- [x] `cargo +1.92.0 test --manifest-path /workspace/warp/Cargo.toml -p warp on_ambient_agent_execution_ended -- --nocapture`

### Screenshots / Videos
- Loom: https://www.loom.com/share/beeaa40e63724b4a89bdacbf57f8dd2a

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/46c8366c-c57e-4f9b-9f48-d3679ad3801d_
_Run: https://oz.staging.warp.dev/runs/019e64f7-3b06-78ce-b9b2-2dd0d8471602_

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -7431,6 +7431,9 @@ impl TerminalView {
         if self.conversation_ended_tombstone_view_id.is_some() {
             return false;
         }
+        if self.is_github_action_ambient_agent_session_from_model(model, app) {
+            return false;
+        }
         if self.has_active_cli_agent_input_session(app) {
             return true;
         }
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -7431,7 +7431,7 @@ impl TerminalView {
         if self.conversation_ended_tombstone_view_id.is_some() {
             return false;
         }
-        if self.is_github_action_ambient_agent_session_from_model(model, app) {
+        if self.blocks_cloud_followups_for_ambient_agent_session_from_model(model, app) {
             return false;
         }
         if self.has_active_cli_agent_input_session(app) {
@@ -20460,6 +20460,14 @@ impl TerminalView {
         ctx: &mut ViewContext<Self>,
     ) -> bool {
         if !FeatureFlag::HandoffCloudCloud.is_enabled() {
+            return false;
+        }
+        let blocks_cloud_followups = {
+            let model = self.model.lock();
+            self.blocks_cloud_followups_for_ambient_agent_session_from_model(&model, ctx)
+        };
+        if blocks_cloud_followups {
+            self.pending_cloud_followup_task_id = None;
             return false;
         }
         let Some(task_id) = self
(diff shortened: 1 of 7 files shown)


app/src/terminal/view.rs
```diff
@@ -7431,6 +7431,9 @@ impl TerminalView {
         if self.conversation_ended_tombstone_view_id.is_some() {
             return false;
         }
+        if self.is_github_action_ambient_agent_session_from_model(model, app) {
```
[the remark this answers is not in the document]
bnavetta: +1 to checking if the task supports followups in `try_submit_pending_cloud_followup`

app/src/terminal/view/ambient_agent/model.rs
```diff
@@ -762,6 +766,10 @@ impl AmbientAgentViewModel {
         self.task_id
     }
 
+    pub(in crate::terminal::view) fn is_github_action_source(&self) -> bool {
```
bnavetta: Thoughts on generalizing this to a `can_follow_up_in_cloud` check? I think Github Actions aren't the only task source type this applies to - if you're looking at the transcript or session share for a local interactive or CLI task, that likely also doesn't support cloud followups.

bnavetta (approved):

app/src/terminal/view/ambient_agent/model.rs
captainsafia: Ah, good call! I laser focused on the original GHA bug here. I'll confirm if we are showing the interactive input for these run types and adjust if that's the case.

captainsafia: After playing with it, I think we're covered by [the existing continuation UI check](https://github.com/warpdotdev/warp/blob/a79721a5c78159a2405d74d25ee30c87b0dfb9ae/app/src/terminal/view/shared_session/cloud_conversation_continuation.rs#L54) which verifies if the user has the correct ACLs on a run. For local runs, the user would not have edit access and would get the tombstone view in this case. 

I did end up cleaning up this method to make naming a bit less GHA specific in the event we actually wanna add more checks here.

PR #11698, opened 2026-05-26
Carry ancestor gitignore status when checking gitignores top-down

lucieleblanc:
Warp builds file trees for features like repo metadata, codebase context, and file indexing. Those traversals need to respect `.gitignore`, but were repeatedly asking the gitignore matcher to re-check a path’s parent directories. This is unnecessary in top-down traversals: the traversal order means we already know whether a parent is ignored.

This PR carries the parent's ignore state. Descendants of ignored directories now inherit that ignored state directly, avoiding unnecessary ancestor matching while preserving behavior for standalone path checks like watcher events.


crates/repo_metadata/src/entry.rs
```diff
@@ -115,12 +138,14 @@ impl Entry {
             gitignores.push(gitignore);
         }
 
-        let path_is_ignored = matches_gitignores(
-            &curr_path,
-            is_dir,
-            &*gitignores,
-            true, /* check_ancestors */
-        ) || is_git_internal_path(&curr_path);
+        let path_is_ignored = ancestor_is_ignored
```
lucieleblanc, line 141: Here's the core logic change: if the parent is ignored, we can short-circuit the other checks for the child.

crates/ai/src/index/full_source_code_embedding/codebase_index.rs
```diff
@@ -2150,13 +2150,14 @@ impl CodebaseIndex {
...
                             // Ignore paths that are excluded by .gitignore, end with .git, or are symlinks.
-                            if matches_gitignores(
-                                &child_path,
-                                is_dir,
-                                &*gitignores,
-                                false, /* check_ancestors */
-                            ) || child_path.ends_with(".git")
+                            if child_path.ends_with(".git")
                                 || child_path.is_symlink()
+                                || matches_gitignores(
```
lucieleblanc, line 2155: Reordered to short-circuit expensive `gitignore` matching when possible.

kevinchevalier: Nice!

kevinchevalier (approved):
Nice! This all makes sense to me.

PR #11701, opened 2026-05-26
Keep gitignore checking scoped during full traversals

lucieleblanc:
Warp builds local file trees and code indexes by walking a repo and applying `.gitignore` rules along the way. Previously, recursive traversal kept every gitignore matcher it discovered in one growing active list, so sibling directory rules stayed active longer than they should and later paths were checked against unnecessary matchers.

This PR adds a scoped GitignoreStack that keeps all discovered gitignore matchers for later use, but only checks the matchers that are active for the current directory path. `Entry::build_tree `and codebase snapshot diff/add traversal now push a directory’s gitignore while descending into that directory and pop it when returning to the parent. This reduces unnecessary gitignore matching work.


lucieleblanc:
This needs a more involved redesign. Closing for now.

PR #11714, opened 2026-05-26
Fix CreateProcessW failure caused by REG_MULTI_SZ environment variables

### LINKED ISSUE #8524: Why does it show an error message as soon as I open it?It's completely unusable.
### Pre-submit Checks

Failed to create shell process C:\Program Files\PowerShell\7\pwsh.exe: 参数错误。 (0x80070057): 参数错误。 (0x80070057)

_No response_

### Additional context

_No response_

### Operating system (OS)

Windows

1 (Not too important)

None

acarl005:
## Description
Skip registry environment variables that are not `REG_SZ` or `REG_EXPAND_SZ` when building the environment block for `CreateProcessW`.

`REG_MULTI_SZ` values contain embedded null bytes which corrupt the environment block — each null is treated as an entry separator, producing malformed entries without `=` that cause `CreateProcessW` to fail with error `0x80070057` ("The parameter is incorrect").

This was observed in a user's log where a third-party driver had stored a `REG_MULTI_SZ` value (`DriverYz2`) in the registry environment, which prevented Warp from spawning any shell process.

This matches how [Windows Terminal](https://github.com/microsoft/terminal/blob/1ba28b298f677d838c3a2e457a8a1f569bff6299/src/inc/til/env.h#L247-L259) handles registry environment variables — it only processes `REG_SZ` and `REG_EXPAND_SZ` types.

## Linked Issue

Closes #8524

## Testing

TODO: Send the user a build.

### DIFF
--- a/app/src/terminal/local_tty/windows/environment.rs
+++ b/app/src/terminal/local_tty/windows/environment.rs
@@ -338,7 +338,13 @@ fn reg_value_to_string(value: &RegValue, key: &str) -> anyhow::Result<OsString> 
     };
 
     // These are null-terminated, but we don't want the terminator here. We add it back later.
-    os_str.map(|v| v.to_string_lossy().trim_end_matches('\0').into())
+    os_str.map(|v| {
+        let s = v.to_string_lossy();
+        match s.find('\0') {
+            Some(pos) => s[..pos].into(),
+            None => v,
+        }
+    })
 }
 
 /// Best-effort lowercase transformation of an OsString.
--- a/app/src/terminal/local_tty/windows/environment.rs
+++ b/app/src/terminal/local_tty/windows/environment.rs
@@ -297,6 +297,15 @@ fn add_user_env(env: &mut BTreeMap<OsString, EnvEntry>) {
 }
 
 fn reg_value_to_string(value: &RegValue, key: &str) -> anyhow::Result<OsString> {
+    // Only REG_SZ and REG_EXPAND_SZ are valid for environment variables.
+    // https://github.com/microsoft/terminal/blob/1ba28b298f677d838c3a2e457a8a1f569bff6299/src/inc/til/env.h#L247-L259
+    if value.vtype != RegType::REG_SZ && value.vtype != RegType::REG_EXPAND_SZ {
+        anyhow::bail!(
+            "Unsupported registry type {:?} for env var {key:?}",
+            value.vtype
+        );
+    }
+
     let key_lower = key.to_ascii_lowercase();
     // RegType::REG_EXPAND_SZ requires expansion of nested env vars, e.g. %USERPROFILE%\AppData to
     // C:\Users\andy\AppData


acarl005:
### Where `DriverYz2` comes from

The root cause was discovered by decoding the `lpEnvironment_bytes` from a user's `warp_preview.log` on a Windows machine (user "Edsion", ThinkBook with Intel Arc graphics, Chinese locale).

The environment block contained:
```
DriverYz2=1737203622\0
18124331\0
```

`DriverYz2` is not a standard Windows environment variable. It appears to be set by a third-party driver utility or hardware driver on the user's system. The value `1737203622` resembles a Unix timestamp (≈ Jan 18, 2025), likely tracking a driver install or update date.

The critical issue is that the value was stored in the registry as `REG_MULTI_SZ` instead of `REG_SZ`. A `REG_MULTI_SZ` value is a list of null-terminated strings (`string1\0string2\0\0`), so when Windows expands it into the process environment block, it produces two entries:
1. `DriverYz2=1737203622\0` — valid
2. `18124331\0` — **malformed** (no `=` sign)

The malformed entry causes `CreateProcessW` to reject the entire environment block with error `0x80070057` ("The parameter is incorrect" / 参数错误), preventing Warp from spawning any shell process.

Both [Windows Terminal](https://github.com/microsoft/terminal/blob/1ba28b298f677d838c3a2e457a8a1f569bff6299/src/inc/til/env.h#L247-L259) and [Alacritty](https://github.com/alacritty/alacritty/blob/master/alacritty_terminal/src/tty/windows/conpty.rs#L246-L287) avoid this by only processing `REG_SZ` and `REG_EXPAND_SZ` types when reading environment variables from the registry.

acarl005:
### Why this disproportionately affects Chinese Windows users

The `DriverYz2` variable appears to come from a Chinese driver management tool (驱动 = "driver" in Chinese). China has a large ecosystem of third-party driver utilities — tools like 驱动总裁 (DrvCeo), 驱动精灵 (Driver Genius), 驱动人生 (Driver Talent), etc. — that are widely used because many Chinese OEM machines ship without proper driver packages, and Microsoft's built-in driver catalog has historically had weaker coverage for hardware sold primarily in the Chinese market.

These tools often write metadata to the registry (install timestamps, version tracking, etc.) and some of them incorrectly use `REG_MULTI_SZ` instead of `REG_SZ` for values under the `Environment` registry keys. This is likely how `DriverYz2` ended up as a `REG_MULTI_SZ` in this user's environment — a bug in whatever driver tool they used.

Users outside China rarely encounter these tools, which is why this issue has been almost exclusively reported by Chinese users.

abhishek-pandya (approved):
Awesome find! I'm realizing now that I should've just copied the Windows Terminal approach from the start. Is there anything else there that looks interesting to copy over?

acarl005:
> Is there anything else there that looks interesting to copy over?

No low-hanging fruits that I can find.

acarl005:
I'm going to merge without confirmation that this fixes the specific issue with `DriverYz2` since we can see that Windows Terminal does this.

PR #11718, opened 2026-05-26
update FilterableDropdown to use trait for SelectActionAndClose instead of generic on entire DropdownAction

moirahuang:
## Description
Out of [this PR](https://github.com/warpdotdev/warp/pull/11412), @vorporeal and I discussed that it would be better to refactor `FilterableDropdown` to not have a generic on `DropdownAction` but to instead use a Box dyn trait on `SelectActionAndClose` so that the other actions don't have to specify the generic.

## Testing
Locally ran and saw no regressions in filterable dropdown behavior + space toggles the dropdown if the dropdown is focused

### Screenshots / Videos
https://github.com/user-attachments/assets/21aefd19-8120-48f4-bbc9-63cb257caee0

## Agent Mode

### DIFF
--- a/app/src/view_components/dropdown.rs
+++ b/app/src/view_components/dropdown.rs
@@ -94,9 +94,10 @@ impl DropdownStyle {
 /// A dropdown menu view. The view renders each DropdownItem. When a menu item is clicked,
 /// on_click_action_name is dispatched, with the value of the corresponding menu item.
 ///
-/// The default item action type is `()`, so callers that only need a static dropdown shell can
-/// name `Dropdown` without specifying an action type. Callers that dispatch item selections should
-/// use a concrete action type.
+/// The item action type powers typed item helpers like `set_items` and
+/// `set_selected_by_action`. Callers that populate the menu with already-erased rich
+/// [`MenuItem<DropdownAction>`] values through `set_rich_items` do not need to name a concrete
+/// item action type.
 pub struct Dropdown<A: DropdownItemAction = ()> {
     is_expanded: bool,
     disabled: bool,
@@ -443,8 +444,9 @@ where
 
     /// Set items from rich menu items.
     ///
-    /// Rich menu items carry erased [`DropdownAction`]s, so callers are responsible for ensuring
-    /// each item dispatches an action handled by the parent view.
+    /// Rich menu items already carry erased [`DropdownAction`]s. The dropdown dispatches selected
+    /// item actions through normal action propagation, so callers should ensure each action is
+    /// handled by an appropriate view in the containing view hierarchy.
     pub fn set_rich_items(
         &mut self,
         items: impl IntoIterator<Item = MenuItem<DropdownAction>>,
--- a/app/src/view_components/dropdown.rs
+++ b/app/src/view_components/dropdown.rs
@@ -1,4 +1,5 @@
 use std::fmt::Debug;
+use std::marker::PhantomData;
 
 use pathfinder_color::ColorU;
 use warpui::elements::{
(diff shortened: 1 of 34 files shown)


app/src/view_components/filterable_dropdown.rs
```diff
@@ -216,43 +231,30 @@ where
...
             ctx.notify();
         }
     }
 
-    /// Set items from rich menu items (MenuItem). This passes the rich menu items to the
-    /// internal dropdown but also extracts searchable DropdownItem objects for filtering.
+    /// Set items from rich menu items (MenuItem). This preserves the rich menu items for
+    /// filtering and passes the filtered items to the internal dropdown.
+    ///
+    /// Rich menu items carry erased [`DropdownAction`]s, so callers are responsible for ensuring
```
moirahuang: @vorporeal would appreciate your advice on if this is ok? rich items are no longer as strictly typed but i'm not sure that the right thing to do here is to make it strictly typed and use `A`. what do you think?

app/src/view_components/filterable_dropdown.rs
```diff
@@ -26,6 +28,17 @@ use crate::menu::{Event as MenuEvent, Menu, MenuItem, MenuVariant};
 use crate::ui_components::icons;
 
 const EMPTY_DROPDOWN_HEIGHT: f32 = 50.0;
+const FILTERABLE_DROPDOWN_COLLAPSED: &str = "FilterableDropdownCollapsed";
+
+pub fn init(app: &mut AppContext) {
+    use warpui::keymap::macros::*;
+
+    app.register_fixed_bindings([FixedBinding::new(
```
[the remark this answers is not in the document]
moirahuang: added

moirahuang: @vorporeal flagging that i'm adding a `FixedBinding` for space here, i know we discussed why we shouldn't have it at the modal level but lmk if you think it's not right to have for `FilterableDropdown`. i'm not opposed to refactoring it to use an event / action flow. i don't know of a clear reason for / against it in this case, am curious for your thoughts

moirahuang: i just switched it over. i can also move it back to using a `FixedBinding` if that's better

app/src/view_components/dropdown.rs
```diff
@@ -110,10 +147,13 @@ pub struct Dropdown<A: Action + Clone> {
     /// confirmation card pickers.
     use_overlay_layer: bool,
     match_menu_width_to_top_bar: bool,
+    // The menu stores erased `DropdownAction`s internally, but the public API remains generic over
+    // the caller's concrete item action type (`set_items`, `set_selected_by_action`, etc.).
+    _item_action_type: PhantomData<A>,
```
moirahuang, line 153: to preserve the generic without storing it

app/src/view_components/filterable_dropdown.rs
```diff
@@ -378,8 +369,10 @@ where
     }
 
     fn focus(&mut self, _delta: usize, ctx: &mut ViewContext<Self>) {
-        ctx.focus(&self.filter_editor);
-        ctx.notify();
+        if self.is_expanded {
```
moirahuang, line 373: prevent stealing focus when dropdown is expanded

app/src/view_components/filterable_dropdown.rs
```diff
@@ -482,11 +479,30 @@ where
...
+        let filter_editor = self.filter_editor.clone();
+        let top_bar_element = top_bar
             .build()
             .on_click(|ctx, _, _| {
-                ctx.dispatch_typed_action(DropdownAction::<A>::ToggleExpanded);
+                ctx.dispatch_typed_action(DropdownAction::ToggleExpanded);
+            })
+            .finish();
+
+        EventHandler::new(top_bar_element)
```
moirahuang, line 494: main work to handle the space key press

app/src/view_components/dropdown.rs
```diff
@@ -60,7 +93,11 @@ impl DropdownStyle {
 
 /// A dropdown menu view. The view renders each DropdownItem. When a menu item is clicked,
 /// on_click_action_name is dispatched, with the value of the corresponding menu item.
-pub struct Dropdown<A: Action + Clone> {
+///
+/// The default item action type is `()`, so callers that only need a static dropdown shell can
+/// name `Dropdown` without specifying an action type. Callers that dispatch item selections should
+/// use a concrete action type.
```
vorporeal: ooc, where do we make use of this "static dropdown shell"?  i'd expect all uses of a dropdown to want to do _something_ when an item is selected, no?  or do they simply react to the fact that _something_ was selected, and read out the value of the current selection and branch on it?

app/src/view_components/filterable_dropdown.rs
vorporeal: i'm not sure there's much inherent value in strictly typing the actions for menu items.  action dispatch propagates up the view hierarchy; looser requirements on actions in menu items could actually be nice here (in that it allows you to have menu items dispatch actions which are handled at different levels of the view hierarchy).

with that said, the comment here isn't _quite_ accurate - "callers are responsible for ensuring each item dispatches an action handled by the parent view" is overly-prescriptive.  callers are responsible for ensuring items dispatch actions that will be handled by _some_ view in the view hierarchy that contains them, but it can be any ancestor, not just their parent.  generally speaking, the expectation is that the parent view will be the one expecting the actions here, but it's certainly not a requirement.

in any case, the previous implementation never provided any guarantees that the parent was able to handle the action - there was no direct linkage between the parent view's action type and the type here.  the typing guaranteed that all actions were of one specific type, but type mismatches were equally possible.

vorporeal (approved):

app/src/view_components/dropdown.rs
moirahuang: yeah you're right that "static dropdown shell" isn't right here since call sites expect item selection to actually do something. i've updated the comment

app/src/view_components/filterable_dropdown.rs
moirahuang: updated the comment to not be so prescriptive. makes sense that it can be the ancestor, not just the parent + also makes sense that the previous implementation didn't provide guarantees that the parent was able to handle the action so no need to try to enforce in this implementation

PR #11719, opened 2026-05-26
Size ask-question cards to their content instead of a fixed height

coolcom200:
## Description
Make the ask-user-question card size to its content instead of a fixed-height cap.

- **Single-question cards** grow to fit their content, up to an 800px cap (previously limited to 320px), so long markdown questions/options aren't squeezed into a small, heavily-scrolling panel.

- **Multi-question cards** size to their **tallest** question — up to a 520px cap (raised from 320px) — and hold that height as you navigate between questions, scrolling only when the tallest question itself exceeds the cap. This avoids the extra scrolling the old single-line height estimate caused for wrapped/markdown options, and keeps the card from resizing as you move between questions of different lengths.

Demo Loom: https://www.loom.com/share/c6ba9190113f4b2da79fdc518434a64d

### Why
The original 320px cap forced excessive scrolling for long single questions. Naive content-based sizing isn't enough on its own, because the framework can only measure the *currently visible* question — so a multi-question card would resize every time you navigated to a question of a different length.

### How
- **Single-question:** `MainAxisSize::Min` + `ConstrainedBox::with_max_height`, so the card grows to fit and scrolls past the cap.
- **Multi-question:** lay out an off-screen, non-interactive **measurement copy** of every question inside a `Stack` (which sizes to its tallest child). Only the visible question is painted; the copies just contribute to sizing, so the card collapses to the tallest question and never resizes during navigation.
- Each copy is built from that question's **saved draft** and goes through the **same scrollable wrapper** as the live body, so the reserved height matches the real answer state — including a long saved "Other…" answer — and a question that overflows the cap measures identically (no shift when you open it).
- New `MeasureOnly` element — lays out its child for sizing but never paints it or routes events to it, clamping the reported size to the incoming constraint.
- New `NumberShortcutButtonBuilder::build_measurement_element` — builds an option's visual element without click/keyboard handling, used to build the copies' options.

### Changes
- `app/src/ai/blocklist/inline_action/ask_user_question_view.rs` — sizing logic, `MeasureOnly`, measurement copies, shared `body_column`/`with_body_insets` helpers.
- `app/src/ai/blocklist/block/number_shortcut_buttons.rs` — `build_measurement_element`.

## Linked Issue

## Testing
Ran on the `warp` crate (the whole diff lives there; CI runs the full-workspace matrix):
- `./script/format`
- `cargo nextest run -p warp ask_user_question` — 29/29 passed

Existing unit tests cover the questionnaire state machine, not layout, so manual verification of the rendering is still needed:

1. Single short question → collapses, no scrollbar.
2. Single long question → grows to ~800px, then scrolls.
3. Multi-question, mixed lengths → sizes to the tallest; ←/→ never changes the card height.
4. Multi-question, tallest exceeds the 520px cap → holds at the cap and scrolls; still stable on navigation.
5. Hidden copies → no ghost text; clicks, number/arrow keys, and text selection affect only the visible question.
6. "Other…" input opens/grows/typeable; multiselect "(select all that apply)" suffix shows; a saved "Other…" answer doesn't resize the card when revisited.

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/ad7b06d9-f299-464c-987f-156a15b14cc9_

### DIFF
--- a/app/src/ai/blocklist/inline_action/ask_user_question_view.rs
+++ b/app/src/ai/blocklist/inline_action/ask_user_question_view.rs
@@ -63,6 +63,7 @@ const ASK_USER_QUESTION_ACTIVE: &str = "AskUserQuestionActive";
 
 pub(crate) const ASK_USER_QUESTION_AUTO_ADVANCE_DELAY: Duration = Duration::from_millis(300);
 pub(crate) const ASK_USER_QUESTION_MAX_CONTAINER_HEIGHT: f32 = 320.;
+pub(crate) const ASK_USER_QUESTION_SINGLE_MAX_CONTAINER_HEIGHT: f32 = 520.;
 pub(crate) const ASK_USER_QUESTION_OPTION_BUTTON_VERTICAL_SPACING: f32 = 4.;
 pub(crate) const ASK_USER_QUESTION_TEXT_TOP_PADDING: f32 = 16.;
 pub(crate) const ASK_USER_QUESTION_TEXT_BOTTOM_PADDING: f32 = 8.;
@@ -99,16 +100,22 @@ fn ask_user_question_header_height(appearance: &Appearance, app: &AppContext) ->
 fn ask_user_question_container_height(
     max_option_count: usize,
     appearance: &Appearance,
+    is_single_question: bool,
     has_nav_footer: bool,
     app: &AppContext,
 ) -> f32 {
+    let max_height = if is_single_question {
+        ASK_USER_QUESTION_SINGLE_MAX_CONTAINER_HEIGHT
+    } else {
+        ASK_USER_QUESTION_MAX_CONTAINER_HEIGHT
+    };
     let mut natural_height = ask_user_question_header_height(appearance, app)
         + ask_user_question_text_height(appearance, app)
         + estimated_min_height_for_all_options(max_option_count, appearance.monospace_font_size());
     if has_nav_footer {
         natural_height += standard_message_bar_height(app) + 1.;
     }
-    natural_height.min(ASK_USER_QUESTION_MAX_CONTAINER_HEIGHT)
+    natural_height.min(max_height)
 }
 
 fn ask_user_question_auto_advance_enabled(is_multiselect: bool, is_last_question: bool) -> bool {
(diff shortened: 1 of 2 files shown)


app/src/ai/blocklist/inline_action/ask_user_question_view.rs
```diff
@@ -1292,12 +1260,29 @@ impl AskUserQuestionView {
                 .with_corner_radius_override(CornerRadius::with_top(Radius::Pixels(8.)))
                 .render_header(app, Some(header_right.finish())),
         );
+
+        // The body sizes to its content, capped at `body_max_height` (scrolling beyond that). For
+        // multi-question cards we stack an off-screen, non-interactive measurement copy of every
+        // question behind the visible one. `Stack` sizes to its tallest child, so the card collapses
+        // to the height of the tallest question and stays fixed as the user navigates, rather than
+        // resizing per question (only the visible question can be measured directly otherwise).
```
coolcom200, line 1279: Any concerns with this approach seems a little unconventional.

app/src/ai/blocklist/inline_action/ask_user_question_view.rs
```diff
@@ -1263,20 +1228,23 @@ impl AskUserQuestionView {
     fn render_active(&self, appearance: &Appearance, app: &AppContext) -> Option<Box<dyn Element>> {
         let theme = appearance.theme();
         let current = self.session.current()?;
-        let mut question_text = current.question.question.clone();
-        if current.question.is_multiselect() {
-            question_text.push_str(" (select all that apply)");
-        }
+        let question_text = Self::question_display_text(current.question);
+        let is_single_question = !self.session.has_multiple_questions();
         let has_nav_footer = self.session.has_multiple_questions();
```
coolcom200: Probably should just reuse the above variable. Will update.

dagmfactory (approved):

szgupta:
Loom looks good to me

app/src/ai/blocklist/inline_action/ask_user_question_view.rs
```diff
@@ -1842,6 +1928,58 @@ pub(crate) fn render_text_with_markdown_support(
     }
 }
 
+/// An element that lays out its child—so the child contributes to sizing—but never paints it or
+/// routes events to it, clamping the reported size to the incoming constraint. Combined with
+/// `Stack`, this lets a multi-question card measure off-screen copies of every question and size to
+/// the tallest one, without showing or interacting with them.
+struct MeasureOnly {
+    child: Box<dyn Element>,
+}
```
szgupta, line 1937: @alokedesai would you say this is an anti-pattern?

szgupta (approved):

app/src/ai/blocklist/inline_action/ask_user_question_view.rs
coolcom200: Discussed offline and concluded that it's a bit hacky, but is scoped to just the ask question tool and we don't have a cleaner solution at the moment.

PR #11722, opened 2026-05-27
[QUALITY-768] Fix orchestration session restoration after Warp restart

cephalonaut:
## Description

Demo: https://www.loom.com/share/46634b5c2513469aa4e69cb0f09cd03a

Restarting Warp could leave orchestration sessions in four broken states: child agents rendered as "Unknown agent" in the pill bar, restored remote-child panes showed a blank "New conversation" placeholder instead of the cloud transcript, a cloud-parent orchestration restored correctly on the first restart but lost its cloud-mode shape on the second, and the on-disk prune split orchestration trees across restarts. This PR addresses all four.

**Eager orchestration-child hydration at startup.** `initialize_historical_conversations` now inserts orchestration children into `BlocklistAIHistoryModel::conversations_by_id` at load time, so the pill bar and transcript name resolver find them before any pane materializes.

**Direct `AmbientAgentTask` inspection for remote-child hydration.** Restored remote-child placeholders went through `AgentConversationsModel::resolve_open_action`, which once the local placeholder existed collapsed "navigate to local" and "hydrate cloud transcript" into the same outcome — so the transcript was never loaded. Hidden-pane hydration now inspects the `AmbientAgentTask` directly via a pure `decide_remote_child_hydration_action` and routes all four arms through a single tombstone gate keyed on whether the task is in a terminal state.

**Cloud-mode shared-session viewer snapshot/restore loop.** A cloud-parent orchestration's pane is a shared-session viewer locally. The restore path used the non-deferred `viewer::TerminalManager::new(..)` which hardcoded `is_cloud_mode = false`, so the restored `TerminalView` had no `ambient_agent_view_model`. Restart 1 still rendered correctly because the network's `JoinedSuccessfully` handler spins up an `OrchestrationViewerModel` regardless, but the snapshot path keys off `ambient_agent_view_model` to choose between `LeafContents::AmbientAgent { task_id }` and the fallback empty `LeafContents::Terminal` — so the next shutdown silently dropped the `task_id` and restart 2 re-materialized the parent as a plain local terminal with no orchestration UI. The fix threads `is_cloud_mode: bool` through `create_shared_session_viewer` and passes `true` from the two ambient-agent restoration call sites; the restored pane's shape now matches a freshly created cloud-mode pane.

**Tree-aware persisted-conversation prune.** The on-disk prune treated rows individually, so parents and children of the same orchestration session could be evicted independently and the next boot would re-encounter a split tree. The prune is now tree-aware: orchestration trees are sorted freshest-first by `max(member.last_modified_at)` and kept atomically, with the freshest tree always retained intact. The retention cap moved from 100 to 200 to fit typical orchestration session sizes.

A fifth issue originally in scope — local-no-harness Oz children persisting an optimistic stub task alongside the real upgraded root — was independently addressed by PR #11814 (QUALITY-774); this PR was rebased onto that fix and drops its earlier read-time stub filter.

The post-facto technical spec at `specs/QUALITY-768/TECH.md` documents the implementation as it ships.

## Testing

Manual validation: local-local restart shows correct agent names; local-remote restart shows the cloud transcript instead of the blank placeholder; cloud-parent restart-restart preserves the orchestration UI on both restarts; tree-aware prune keeps orchestration trees together on disk; baseline single-conversation restore unchanged.

Automated coverage added across the dispatch decision, the cloud-transcript merge seam, the cloud-mode viewer snapshot contract, and the tree-aware eviction. `cargo fmt`, `cargo clippy --all-targets --features local_fs -- -D warnings`, and `cargo nextest run` across the affected crates all pass.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed orchestration session restoration so the pill bar, child transcript name resolution, restored remote-child cloud transcripts, and cloud-parent panes all work correctly after a Warp restart.


app/src/pane_group/mod.rs
```diff
@@ -3766,6 +3815,328 @@ impl PaneGroup {
...
+        child_conversation: AIConversation,
+        parent_pane_id: PaneId,
+        task_id: AmbientAgentTaskId,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        let child_id = child_conversation.id();
+
+        // Idempotency guard — see fn doc.
+        if let Some(existing_pane_id) = self.child_agent_panes.get(&child_id).copied() {
+            if self.has_pane_id(existing_pane_id) && child_conversation.exchange_count() > 0 {
```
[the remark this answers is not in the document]
cephalonaut: Addressed in a58bf8747: the inner guard now returns whenever there is a live tracked pane, regardless of `exchange_count()`. A second mid-hydration call from `restore_missing_child_agent_panes_for_parent` will hit this guard and bail out before inserting a duplicate hidden pane.

app/src/pane_group/mod.rs
```diff
@@ -3766,6 +3815,328 @@ impl PaneGroup {
...
+            }
+            RemoteChildHydrationAction::Fallback => {
+                // No live session, no server token: attach to the
+                // (possibly empty) ambient session and insert the
+                // conversation-ended tombstone so we're never worse than
+                // the pre-fix behaviour.
+                self.apply_existing_ambient_task_to_pane(pane_id, child_id, task_id, ctx);
+                if let Some(terminal_view) = self.terminal_view_from_pane_id(pane_id, ctx) {
+                    terminal_view.update(ctx, |view, ctx| {
+                        view.insert_conversation_ended_tombstone_with_resolved_cta(ctx);
```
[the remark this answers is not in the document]
cephalonaut: Addressed in a58bf8747: `RemoteChildHydrationAction::Fallback` now carries `task_is_terminal`, and the dispatch arm routes through `attach_ambient_session_and_maybe_tombstone` instead of unconditionally tombstoning. An `ActiveUnattachable` run that lands in `Fallback` keeps `task_is_terminal: false` and skips the tombstone. Added a unit test (`decide_remote_child_hydration_active_unattachable_without_token_falls_back_non_terminal`) to lock it in.

app/src/ai/blocklist/agent_view/orchestration_pill_bar_tests.rs
```diff
@@ -4,6 +4,176 @@ use super::*;
 // `app/src/ai/blocklist/orchestration_topology_tests.rs`. These tests stay
 // focused on the pill bar's own dispatch behavior.
 
+/// Fix C: the data layer that `OrchestrationPillBar::pill_specs` reads must
```
seemeroland: super nit: assume C refers to the spec, but don't need it in the comment on the test

app/src/ai/blocklist/block/view_impl/orchestration_tests.rs
```diff
@@ -252,6 +252,156 @@ fn participant_for_agent_id_uses_pill_style_child_agent_avatar() {
     });
 }
 
+/// Fix C: a restored child run id (persisted via run_id in the SQLite
```
seemeroland: same

app/src/pane_group/mod.rs
```diff
@@ -3768,6 +3822,327 @@ impl PaneGroup {
         }
     }
 
+    /// Drains entries from `pending_remote_child_hydrations` for which task
+    /// data is now available, hydrating each hidden child pane in place.
+    fn process_pending_remote_child_hydrations(&mut self, ctx: &mut ViewContext<Self>) {
```
seemeroland, line 3827: nit: I feel like it would be nice to move the orchestration specific functions into a separate file, since this is already really long

app/src/ai/blocklist/history_model.rs
```diff
@@ -2360,6 +2361,99 @@ impl BlocklistAIHistoryModel {
...
+            .ok_or_else(|| {
+                anyhow!(
+                    "merge_cloud_tasks_into_existing_conversation: \
+                     local placeholder {local_placeholder_id} not found in conversations_by_id; \
+                     refusing to construct a detached merged conversation"
+                )
+            })?;
+
+        // Placeholder owns orchestration linkage; cloud supplies
+        // tasks/title/server metadata.
```
seemeroland: Kind of confused by this - where does the linkage live? If it's just the `local_placeholder_id` needing to exist in the history model and here we're populating the data, can we directly insert the `cloud_conversation` for that local conversation id, instead of reconstructing it again here?

Reconstructing the `AgentConversationData` to pass into `new_restored` seems circular, since we already did that when fetching the `cloud_conversation`.

app/src/ai/blocklist/history_model.rs
```diff
@@ -2360,6 +2361,99 @@ impl BlocklistAIHistoryModel {
         Ok(conversation)
     }
 
+    /// Rebuilds the local placeholder identified by `local_placeholder_id`
+    /// from the cloud `tasks` + `cloud_conversation`, keeping the
+    /// placeholder's local id and orchestration linkage (parent ids,
+    /// agent_name, run_id, is_remote_child, pinned) authoritative. Returns
+    /// `Err` when the placeholder isn't loaded so the caller can fall back
+    /// instead of silently producing a detached conversation.
+    pub fn merge_cloud_tasks_into_existing_conversation(
```
seemeroland: Would prefer a different name to make clear this is only to be used for this placeholder orchestration child case - don't want people misusing this

seemeroland (commented):
Thanks for the fix! Lgtm, only blocking question is on whether `merge_cloud_tasks_into_existing_conversation` needs to recreate the conversation

app/src/ai/blocklist/agent_view/orchestration_pill_bar_tests.rs
cephalonaut: Done in 4f84fd147 — dropped the "Fix C:" lead-in (and the matching body comments that referenced "Before Fix C" / "After Fix C" in this same file).

app/src/ai/blocklist/block/view_impl/orchestration_tests.rs
cephalonaut: Done in 4f84fd147 — dropped the "Fix C:" lead-in.

app/src/pane_group/mod.rs
cephalonaut: Strong agreement that mod.rs is too long and the orchestration cluster is a natural split point. Deferring to a Phase 1.5 follow-up PR to keep this PR scoped to the restoration-bug fix.

Proposed layout for the follow-up: expand `child_agent.rs` into a `child_agent/` directory with `mod.rs` (today's creation paths), `restoration.rs` (lazy hidden-child-pane materialization), and `hydration.rs` (remote-child cloud-transcript hydration, including `RemoteChildHydrationAction` + `decide_remote_child_hydration_action`). Plus a sibling `ambient_pane_restoration.rs` for the top-level ambient-pane restoration cluster. Shared subscription wiring stays in `mod.rs` since both pending maps drive off it.

Mechanical move, no behavioral change — I'll open it as a separate PR.

app/src/ai/blocklist/history_model.rs
cephalonaut: Good question — the linkage lives on the **local placeholder**, not the cloud transcript. The reason this isn't `conversations_by_id.insert(local_placeholder_id, cloud_conversation)` is that there's no single source of truth here — the merged record mixes both:

- **Placeholder authoritative** (linkage the cloud transcript can't reconstruct): `parent_conversation_id`, `is_remote_child`, `pinned`.
- **Placeholder-preferred, cloud fallback** (local value wins when present, cloud's value used otherwise so we don't lose data on a stale placeholder): `parent_agent_id`, `agent_name`, `orchestration_harness_type`, `run_id`.
- **Cloud authoritative** (server-side state the placeholder doesn't know): `server_conversation_token`, `conversation_usage_metadata`, `forked_from_server_conversation_token`, `artifacts_json`, `last_event_sequence`, `server_metadata`.
- **Reset on merge**: `reverted_action_ids`, `root_task_is_optimistic`, `autoexecute_override`.

Three concrete reasons we can't reuse `cloud_conversation` as-is:

1. `cloud_conversation.id()` is whatever id was reconstructed from the server token; we need the merged record under `local_placeholder_id` so all the orchestration topology indexes (`children_by_parent`, `child_agent_panes`, run-id lookups) stay valid.
2. Cloud has no notion of `parent_conversation_id` / `is_remote_child` / `pinned` / local `run_id` — using cloud's `AgentConversationData` directly would silently drop those.
3. We need `reassign_exchange_ids()` post-rebuild so the merged conversation's block ids don't collide with anything still on disk for the placeholder. `new_restored(local_id, tasks, AgentConversationData)` is the canonical way to do that — it's the same pattern as `insert_forked_conversation_from_tasks` directly above.

In 4f84fd147 I pulled the field selection out into a named helper, `merged_remote_child_placeholder_conversation_data(&placeholder, &cloud)`, with the four buckets documented on its docstring — should make the placeholder-vs-cloud intent legible without having to read the inline init block.

app/src/ai/blocklist/history_model.rs
cephalonaut: Renamed in 4f84fd147 to `hydrate_remote_child_placeholder_with_cloud_transcript`. The new name matches the verb already used at the pane-group layer (`hydrate_remote_child_transcript_in_place`, `hydrate_task_backed_hidden_child_pane`) and makes the precondition explicit. Docstring also calls out that it's narrowly scoped to the remote-child placeholder hydration path so future callers don't reach for it as a general primitive.

app/src/pane_group/mod.rs
cephalonaut: ^ is agent written, but I basically concur. Would be easier to do as a followup that just moved functions.

app/src/ai/blocklist/history_model.rs
cephalonaut: ^ not sure I actually believe all of this but the gist is correct - this merges two different data sources together which is why we can't just overwrite.

seemeroland (approved):

PR #11729, opened 2026-05-27
Add IAP (Identity-Aware Proxy) support for staging

IsaiahWitzke:
on staging builds, the new iap manager will inject Proxy-Auth headers in the http client headers.

A similar thing is done with the `AuthState` / `AuthManager`...

#### Also includes a staging-only view of iap credentials:

image

## Audit logging

we dont get to see WHO is making the requests (we need to generate an iap token via service account impersonation because of _reasons_) so it looks like all requests are coming in from the same IAP user (although... you'll still need to be logged in normally with an Auth Bearer token as yourself like normal (except for non-auth'd endpoints))

but we can see who & when this SA got impersonated with [these audit logs](https://console.cloud.google.com/logs/query;query=protoPayload.serviceName%3D%22iamcredentials.googleapis.com%22%0AprotoPayload.methodName%3D%22GenerateAccessToken%22%0AprotoPayload.request.name%3D%22projects%2F-%2FserviceAccounts%2Fiap-access@warp-server-staging.iam.gserviceaccount.com%22;cursorTimestamp=2026-04-24T18:09:41.026309487Z?project=warp-server-staging)

### DIFF
--- a/app/src/server/iap.rs
+++ b/app/src/server/iap.rs
@@ -9,8 +9,10 @@ use warp_core::channel::IapConfig;
 use warpui::r#async::Timer;
 use warpui::{Entity, ModelContext, SingletonEntity};
 
+use crate::view_components::DismissibleToast;
+use crate::workspace::{ToastStack, WorkspaceAction};
+
 const PROACTIVE_REFRESH_BUFFER: Duration = Duration::from_secs(5 * 60);
-const FALLBACK_TOKEN_VALIDITY: Duration = Duration::from_secs(55 * 60);
 const INJECTED_TOKEN_ENV_VAR: &str = "WARP_IAP_TOKEN";
 
 const BASE_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(30);
@@ -22,6 +24,18 @@ const MAX_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(5 * 60);
 const MAX_FAILURE_RETRIES: u32 = 5;
 
 #[derive(Debug, Clone)]
+pub struct CachedToken {
+    pub token: String,
+    pub expires_at: Instant,
+}
+
+impl CachedToken {
+    fn valid_token(&self) -> Option<String> {
+        (self.expires_at > Instant::now()).then(|| self.token.clone())
+    }
+}
+
+#[derive(Debug, Clone)]
 pub enum IapCredentialsState {
     Missing,
     /// A credential fetch is in progress. `previous` carries the last
@@ -29,14 +43,13 @@ pub enum IapCredentialsState {
     /// outbound requests while we're refreshing so that proactive refreshes
     /// (i.e. refresh the token 5min before exp) don't prevent active requests.
     Refreshing {
-        previous: Option<(String, Instant)>,
+        previous: Option<CachedToken>,
     },
-    Loaded {
-        token: String,
-        expires_at: Instant,
-    },
+    Loaded(CachedToken),
     Failed {
         message: String,
+        // in case the last token still works... we can try to use that for a couple more mins
+        previous: Option<CachedToken>,
     },
     /// Represents a terminal state in the iap creds state machine.
     /// The gcloud refresh loop will never run, and an IAP challenge is logged
@@ -47,6 +60,17 @@ pub enum IapCredentialsState {
     EnvInjected {
         token: String,
     },
+}
+
+impl IapCredentialsState {
+    fn previous_token(&self) -> Option<CachedToken> {
+        match self {
+            IapCredentialsState::Loaded(cached) => Some(cached.clone()),
+            IapCredentialsState::Refreshing { previous }
+            | IapCredentialsState::Failed { previous, .. } => previous.clone(),
+            IapCredentialsState::EnvInjected { .. } | IapCredentialsState::Missing => None,
+        }
+    }
 }
 
 pub struct IapState {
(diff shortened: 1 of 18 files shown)


app/src/server/iap.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    fn schedule_failure_retry(&mut self, ctx: &mut ModelContext<Self>) {
+        if self.consecutive_failures >= MAX_FAILURE_RETRIES {
+            log::warn!(
+                "IAP token fetch failed {MAX_FAILURE_RETRIES} times in a row; giving up until \
+                 manual refresh or server challenge"
+            );
+            return;
+        }
```
tylerlam-warp, line 261: Out of curiosity, was there something with the IAP infrastructure which motivated you to build retry logic for this?

app/src/server/iap.rs
```diff
@@ -0,0 +1,358 @@
...
+            let secs_remaining = exp.checked_sub(now)?;
+            Some(Instant::now() + Duration::from_secs(secs_remaining))
+        })
+        .unwrap_or_else(|| {
+            log::warn!(
+                "Could not parse `exp` claim from IAP token; falling back to {}s validity",
+                FALLBACK_TOKEN_VALIDITY.as_secs()
+            );
+            Instant::now() + FALLBACK_TOKEN_VALIDITY
+        })
```
tylerlam-warp: Wondering -- why did you elect to default to a fallback token duration of 55 minutes here instead of just directly failing

app/src/server/mod.rs
```diff
@@ -3,6 +3,10 @@ pub mod cloud_objects;
 pub mod datetime_ext;
 pub mod experiments;
 pub mod graphql;
+// IAP items are only referenced from native code paths; on wasm the
+// module compiles but every function is dead code.
+#[cfg_attr(target_family = "wasm", allow(dead_code))]
+pub(crate) mod iap;
```
tylerlam-warp, line 9: food for thought: what if we didn't compile this package on WASM? 

```
#[cfg(not(target_family = "wasm"))]
```

Would mean that we have additional lift to have that macro on every function that calls the iap module though

tylerlam-warp: Ah i guess it would get pretty annoying here: https://github.com/warpdotdev/warp/pull/11729/changes#diff-c6d8f3ed1bb80b80b97bdf4133e13dc8bb10887f26178482a2bc2f918ae4bd28R1600

app/src/server/server_api.rs
```diff
@@ -521,6 +542,59 @@ impl ServerApi {
...
+            return false;
+        }
+        if http_client::iap::is_iap_challenge(response.status(), response.headers()) {
+            log::warn!(
+                "Received IAP challenge (status {}); notifying IapManager",
+                response.status()
+            );
+            let _ = self
+                .event_sender
+                .try_send(ServerApiEvent::IapChallengeReceived);
```
tylerlam-warp: I may be missing some context here so please correct me if I am -- but as far as I understand, we don't retry the current request if the request fails and we get a IAP challenge.

Should we be retrying the current request? The technical lift is probably higher and not sure if this happens enough to make the trouble worth it, but seems like it'll make the experience less jank if I don't have my workspace metadata call fail once in a blue moon

app/src/server/iap.rs
```diff
@@ -0,0 +1,358 @@
...
+                        manager.consecutive_failures = 0;
+                        log::info!("IAP token refreshed");
+                        ctx.emit(IapManagerEvent::StateChanged);
+                        ctx.notify();
+                        manager.schedule_next_refresh(expires_at, ctx);
+                    }
+                    Err(err) => {
+                        let message = format!("{err:#}");
+                        log::warn!("IAP token fetch failed: {message}");
+                        state.set_failed(message);
```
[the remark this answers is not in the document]
tylerlam-warp: seems like a valid point?

app/src/server/server_api.rs
tylerlam-warp: Should we check the result of this `try_send`, in the event it fails?

app/src/settings_view/main_page.rs
```diff
@@ -1052,6 +1071,118 @@ impl LogoutWidget {
...
+            .with_style(UiComponentStyles {
+                font_size: Some(12.),
+                padding: Some(Coords::uniform(6.).left(16.).right(16.)),
+                ..Default::default()
+            })
+            .build()
+            .on_click(|ctx, _, _| {
+                ctx.dispatch_typed_action(MainPageAction::RefreshIapCredentials);
+            })
+            .finish();
```
tylerlam-warp, line 1169: nit: Hide this button if the state is env injected?

tylerlam-warp (approved):
looks good, left some comments! Super excited, will make life easier for all of us

app/src/server/iap.rs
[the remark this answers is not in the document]
IsaiahWitzke: fixed

app/src/server/mod.rs
IsaiahWitzke: yeah, you're right... it'd get messy

app/src/server/iap.rs
IsaiahWitzke: fixed to error out loudly. good point

app/src/server/iap.rs
IsaiahWitzke: no... i think this mightve been a oz-reviewer bot on the internal repo when i made v1 of this PR there that told me to do it...

i think its fair though, for intermittent inet in airports etc this is kinda nice

app/src/settings_view/main_page.rs
IsaiahWitzke: im going to ignroe this nit:)

if its env-injected either:
a. warp client is being run via cli, and therefore this button is effectively hidden anyway (the entire UI is gone lol)
b. it was set in my env before i started the warp gui for some reason?! in which case i almost certainly DO want to refresh the token lol

app/src/server/server_api.rs
IsaiahWitzke: ill log it, but AM is telling in both cases, there's nothing really actionable to do on the err

```
/// An error returned from [`Sender::try_send()`].
#[derive(PartialEq, Eq, Clone, Copy)]
pub enum TrySendError<T> {
    /// The channel is full but not closed.
    Full(T),

    /// The channel is closed.
    Closed(T),
}
```

image

PR #11745, opened 2026-05-27
fix: persist custom/BYOK model selection across profile reconciliation (#11564)

### LINKED ISSUE #11564: Default AI profile model does not persist across restarts
## Summary
The selected model for the default AI profile does not always persist over time. After changing the default profile's model multiple times, Warp can later revert that profile to an older model selection after restart.

## Problem
The user changes the model on their `default` AI profile frequently and expects the latest selected model to remain saved. For the past few months, the selected model has often not persisted reliably. On the latest Warp restart today, the default model reverted to `gpt-5.3-codex`, which the user believes was set several default-model changes ago.

## Expected behavior
When the user changes the model on the `default` AI profile, Warp should persist that model selection and continue using it across restarts and over time until the user changes it again.

## Actual behavior
The default AI profile can revert to an older previously selected model after some time or after restart. The report specifically observed the default model reverting to `gpt-5.3-codex` on startup, despite the user believing they had changed the default model multiple times since then.

1. Open Warp and edit the `default` AI profile.
2. Change the selected model to a new value.
3. Continue using Warp and change the default profile's model occasionally over time.
4. Restart Warp.
5. Observe whether the `default` profile still uses the most recently selected model.

Known limitation: the user has seen this often over several months, but has not found reliable reproduction steps yet.

## Artifacts
None attached.

## Warp version
v0.2026.05.19.09.24.dev_00

## Operating system
macOS 26.3.1 (build 25D771280a)

## Possible source references
- `app/src/ai/execution_profiles/mod.rs`: likely central execution profile model/persistence surface.
- `app/src/ai/execution_profiles/editor/manager.rs`: likely handles profile editing/update flows.
- `app/src/settings_view/execution_profile_view.rs`: likely user-facing settings/editor view for execution profiles.

jasonpaulso:
## Summary

Fixes #11564 — the default AI profile's selected model (especially custom/BYOK models) silently reverts to `auto`.

## Root cause

`LLMPreferences::reconcile_disabled_model_preferences()` (`app/src/ai/llms.rs`) clears any saved profile model it considers unusable. It validated the saved id **only against the server-provided model lists** (`models_by_feature.{agent_mode,coding,cli_agent}`):

```rust
self.models_by_feature.agent_mode.usable_info_for_id(effective_base_model_id, ctx)
```

Custom/BYOK models don't live there — they're built separately into `custom_llms`, keyed by a `config_key` UUID that never appears in the server `choices`. Every model **getter** resolves against both collections (e.g. `get_preferred_base_model` does `.info_for_id(&id).or_else(|| self.custom_llm_info_for_id_if_enabled(&id, app))`), but reconciliation did not. So a valid custom selection looked "unusable," reconciliation called `set_base_model(profile_id, None, ...)`, and the profile fell back to the default `auto` model.

This explains why the revert was intermittent and not restart-only: `reconcile_disabled_model_preferences` runs on **startup, server model-list refresh, auth-complete, network-online, teams-change, and BYOK-key updates** — any of which mid-session would wipe the custom selection (matching reports of reverts with no restart, on new agent sessions, and after a usage/credits refresh).

## Fix

Make reconciliation's usability check symmetric with the getters: a saved id is usable if it resolves to a usable server model **or** an enabled custom/BYOK model. Applied to the `base`, `coding`, and `cli_agent` fields. Computer-use is intentionally left unchanged — its getter has no custom-model resolution path, so a custom id can't be stored there.

## Tests

Added two GPUI regression tests in `app/src/ai/llms_tests.rs`:

- `reconcile_preserves_custom_base_model_for_default_profile` — a saved custom/BYOK base model survives reconciliation (would have failed before this fix).
- `reconcile_clears_unknown_base_model_for_default_profile` — guards against over-correcting: an id matching neither a server model nor a known custom endpoint is still cleared.

Both pass (`cargo test -p warp --lib reconcile_`).

## Manual verification

Built and ran the patched client (`./script/run`, WarpOss/`Channel::Oss`). Set a custom/BYOK base model on the default profile and exercised the reconcile triggers (restart, new agent session, network toggle); the selection now persists instead of reverting to `auto`.

### DIFF
--- a/app/src/ai/llms.rs
+++ b/app/src/ai/llms.rs
@@ -1178,6 +1178,16 @@ impl LLMPreferences {
     ///
     /// Called both when the model list is refreshed from the server and when
     /// BYOK API keys change (since `RequiresUpgrade` usability is BYOK-aware).
+    ///
+    /// A saved selection is considered usable if it resolves to a usable
+    /// server model *or* an enabled custom/BYOK model (`custom_llms`). This
+    /// must stay symmetric with the model getters (e.g. `get_preferred_base_model`),
+    /// which resolve against both collections — otherwise a custom model id
+    /// (a `config_key` UUID absent from the server `choices`) is treated as
+    /// unusable here and the profile is silently reset to the default `auto`
+    /// model on restart. Custom-model clearing when custom inference is
+    /// *disabled* is handled separately by
+    /// `sanitize_disabled_custom_model_preferences`.
     fn reconcile_disabled_model_preferences(&self, ctx: &mut ModelContext<Self>) {
         let profiles_model = AIExecutionProfilesModel::handle(ctx);
         profiles_model.update(ctx, |profiles, ctx| {
@@ -1191,7 +1201,10 @@ impl LLMPreferences {
                     let effective_base_model_usable = self
                         .models_by_feature
                         .agent_mode
-                        .usable_info_for_id(effective_base_model_id, ctx);
+                        .usable_info_for_id(effective_base_model_id, ctx)
+                        .or_else(|| {
+                            self.custom_llm_info_for_id_if_enabled(effective_base_model_id, ctx)
+                        });
                     let effective_base_model_unusable = effective_base_model_usable.is_none();
                     let effective_base_model_is_configurable = effective_base_model_usable
                         .is_some_and(|info| info.context_window.is_configurable);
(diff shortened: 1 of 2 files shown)


jasonpaulso:
recheck

bholmesdev:
Apologies for reopening above. Did some investigation, and I'm closing this issue again since the custom/BYOK reconciliation bug described here was fixed by #11655. However, we do still have reported edge cases on the parent issue #11564, and we're tracking a fix for this on the server separate from this.

PR #11749, opened 2026-05-27
Basic vertical tab grouping

johnturcoo:
## Description

Basic functionality for vertical tab grouping. Tabs in the vertical tabs panel can now be organized into collapsible groups with their own header, chevron, and action buttons. All behavior is gated behind `FeatureFlag::GroupedTabs` — with the flag off this PR is a no-op.

## Features supported
- **New "New tab group" entry in the unified new-session menu** (the `+` button dropdown in the vertical tabs panel). Selecting it creates a fresh group containing one new tab.
- **Group container rendering** in the vertical tabs panel:
  - Header with chevron + group name + `N tabs` subtitle
  - Single-click anywhere on the header (or the chevron) toggles collapse/expand
  - Kebab + close action buttons appear on header hover
  - Closing a group closes all of its member tabs
- **Member tabs render indented** inside the group container, with the group's hover/active background painting across the whole group instead of per-tab
- **Group-aware new tab placement**: when the active tab is in a group, new tabs inherit the group's `group_id` and stay clustered with it. The user's `NewTabPlacement` setting is honored within the group's bounds — `AfterCurrentTab` lands immediately after the active member, and `AfterAllTabs` lands at the end of the group's contiguous run (not at the end of the panel)

Other features such as group renaming, dragging tabs into/out of groups, dragging groups, moving tabs between groups, and other menu options will be implemented in a follow up PR.

Figma designs: https://www.figma.com/design/yTHUBNWdYjQquFugM4qjA5/Tab-grouping?node-id=6001-10704&m=dev

Note there are some minor UI changes that are not completely up to spec, still reaching alignment and will update in follow up PR.

## Linked Issue
https://linear.app/warpdotdev/issue/APP-4609/vertical-tab-grouping-basic-functionality

## Testing

- [x] Creating a new tab group from the `+` menu creates the group + one member tab
- [x] Adding new tabs from within the group keeps the group contiguous
- [x] Single-click on the header toggles collapse; the chevron icon flips
- [x] Closing the group (via the `X`) closes every member tab
- [x] Confirmed flag-off behavior is unchanged (menu item hidden, no group rendering)

### Screenshots / Videos

https://www.loom.com/share/dfb13128f5f34585bb54880274cbb84b

### DIFF
--- a/app/src/workspace/tab_group.rs
+++ b/app/src/workspace/tab_group.rs
@@ -1,0 +1,22 @@
+//! Tab group data model. Gated at runtime by `FeatureFlag::GroupedTabs`.
+
+use uuid::Uuid;
+
+/// Stable identity for a tab group.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct TabGroupId(pub Uuid);
+
+impl TabGroupId {
+    pub fn new() -> Self {
+        Self(Uuid::new_v4())
+    }
+}
+
+/// A group of tabs. Member tabs point to their
+/// group via `TabData::group_id`.
+#[derive(Clone)]
+pub struct TabGroup {
+    pub id: TabGroupId,
+    pub name: Option<String>,
+    pub collapsed: bool,
+}
--- a/app/src/workspace/tab_group.rs
+++ b/app/src/workspace/tab_group.rs
@@ -12,6 +12,12 @@ impl TabGroupId {
     }
 }
 
+impl Default for TabGroupId {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 /// A group of tabs. Member tabs point to their
 /// group via `TabData::group_id`.
 #[derive(Clone)]
--- a/app/src/workspace/tab_group.rs
+++ b/app/src/workspace/tab_group.rs
@@ -26,3 +26,20 @@ pub struct TabGroup {
     pub name: Option<String>,
     pub collapsed: bool,
 }
+
+impl TabGroup {
+    /// Creates a new, untitled, expanded tab group with a fresh id.
+    pub fn new() -> Self {
+        Self {
+            id: TabGroupId::new(),
+            name: None,
+            collapsed: false,
+        }
+    }
+}
+
+impl Default for TabGroup {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -467,6 +467,7 @@ use crate::workspace::header_toolbar_editor::{HeaderToolbarEditorEvent, HeaderTo
 use crate::workspace::header_toolbar_item::HeaderToolbarItemKind;
 use crate::workspace::one_time_modal_model::OneTimeModalModel;
 use crate::workspace::sync_inputs::SyncedInputState;
+use crate::workspace::tab_group::{TabGroup, TabGroupId};
 use crate::workspace::tab_settings::TabCloseButtonPosition;
 use crate::workspace::toast_stack::{
     ToastStack, ToastStack as WorkspaceToastStack, ToastStackEvent as WorkspaceToastStackEvent,
@@ -933,6 +934,8 @@ pub struct Workspace {
     tab_bar_hover_state: MouseStateHandle,
     tab_fixed_width: Option<f32>,
     traffic_light_mouse_states: TrafficLightMouseStates,
+    /// Tab groups in this workspace, keyed by id.
+    pub(crate) tab_groups: HashMap<TabGroupId, TabGroup>,
     tab_rename_editor: ViewHandle<EditorView>,
     pane_rename_editor: ViewHandle<EditorView>,
     vertical_tabs_search_input: ViewHandle<EditorView>,
@@ -3085,6 +3088,7 @@ impl Workspace {
             hovered_tab_index: None,
             tab_bar_hover_state: Default::default(),
             traffic_light_mouse_states: Default::default(),
+            tab_groups: HashMap::new(),
             tab_rename_editor: Self::tab_rename_editor(ctx),
             pane_rename_editor: Self::pane_rename_editor(ctx),
             vertical_tabs_search_input: Self::vertical_tabs_search_input(ctx),
@@ -6481,7 +6485,7 @@ impl Workspace {
             #[cfg(not(feature = "local_fs"))]
             NewSessionMenuItem::CreateNewTabConfig => {}
             NewSessionMenuItem::CreateNewTabGroup => {
-                // TODO(johnturcoo): implement tab group creation.
+                self.create_new_tab_group(ctx);
             }
         }
     }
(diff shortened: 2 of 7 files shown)


crates/warp_features/src/lib.rs
```diff
@@ -943,15 +943,14 @@ pub const DOGFOOD_FLAGS: &[FeatureFlag] = &[
     FeatureFlag::SshRemoteServer,
     FeatureFlag::DragTabsToWindows,
     FeatureFlag::RemoteCodebaseIndexing,
-    FeatureFlag::GroupedTabs,
```
peicodes, line 946: How do we feel about this flag staying on for dogfood? Are we confident that there's nothing breaking in this PR?

I ask because if this is purely additive and the entrypoints are specific actions the user takes (like right-click add to group), this is totally fine to just turn on

You don't have to make the announcement right away but it'll get you early feedback if there are bugs

app/src/workspace/tab_group.rs
```diff
@@ -0,0 +1,28 @@
+//! Tab group data model. Gated at runtime by `FeatureFlag::GroupedTabs`.
+
+use uuid::Uuid;
+
+/// Stable identity for a tab group.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct TabGroupId(pub Uuid);
```
peicodes, line 7: Curious if we have this pattern elsewhere or we just use Uuid directly. We don't implement any new methods on `TabGroupId` - are there plans to? I ask because that would be a compelling reason to use TabGroupId > Uuid

app/src/workspace/view.rs
```diff
@@ -6481,7 +6485,7 @@ impl Workspace {
             #[cfg(not(feature = "local_fs"))]
             NewSessionMenuItem::CreateNewTabConfig => {}
             NewSessionMenuItem::CreateNewTabGroup => {
-                // TODO(johnturcoo): implement tab group creation.
+                self.create_new_tab_group(ctx);
```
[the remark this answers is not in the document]
peicodes: nit: Not a bad suggestion but not strictly necessary.

app/src/workspace/view.rs
```diff
@@ -6635,6 +6639,98 @@ impl Workspace {
     #[cfg(not(feature = "local_fs"))]
     fn save_current_tab_as_new_config(&mut self, _tab_index: usize, _ctx: &mut ViewContext<Self>) {}
 
+    /// Creates a new tab group containing a single new tab.
+    fn create_new_tab_group(&mut self, ctx: &mut ViewContext<Self>) {
+        let group_id = TabGroupId::new();
+        self.tab_groups.insert(
+            group_id,
+            TabGroup {
```
peicodes: I would create a helper on `TabGroup` that initializes itself given params. It's better controlled than just directly using the struct, which makes it more difficult in the future to make updates across the board

app/src/workspace/view.rs
```diff
@@ -6635,6 +6639,98 @@ impl Workspace {
...
+            false,
+            ctx,
+        );
+        let new_tab_index = self.active_tab_index;
+        if let Some(tab) = self.tabs.get_mut(new_tab_index) {
+            tab.group_id = Some(group_id);
+        }
+
+        // Cluster the new group after the last existing grouped tab.
+        let target_index = self
```
peicodes: thought: this seems pretty hefty which makes me smell a data structure issue. Would be curious to know why we have to do this

app/src/workspace/view.rs
```diff
@@ -11053,6 +11149,12 @@ impl Workspace {
 
         let is_new_terminal = matches!(panes_layout, PanesLayout::SingleTerminal(_));
         let is_restoration = matches!(panes_layout, PanesLayout::Snapshot(_));
+        // Capture the active tab's group membership so the new tab can inherit it.
+        let active_tab_group_id = if FeatureFlag::GroupedTabs.is_enabled() && !is_restoration {
```
peicodes, line 11128: It's fine to be exhaustive with feature flags like this, but my read is that `tab.group_id` would always return None if the ff is off

This could be incorrect though, because of hydration and persistence. You should test what happens when you start creating tab groups and then turn off the feature flag, because rollback is a potential outcome we wanna ensure works

app/src/workspace/view.rs
```diff
@@ -11078,10 +11180,24 @@ impl Workspace {
 
         match new_tab_placement_setting {
             NewTabPlacement::AfterAllTabs => {
-                self.tabs.push(TabData::new(new_pane_group));
+                // When inheriting a group, land at the end of the group's
+                // contiguous run instead of past it so the setting is
+                // honored within the group's bounds.
+                let insert_idx = active_tab_group_id
+                    .and_then(|gid| {
```
peicodes, line 11162: This looks familiar :) maybe worth extracting or using some shared fn?

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -820,6 +838,13 @@ struct TabGroupDragState {
     insert_after_index: Option<usize>,
 }
 
+/// One visible member of a tab group, captured during the render walk.
+struct TabGroupMember<'a> {
+    tab_index: usize,
+    filtered_pane_ids: Option<&'a [PaneId]>,
+    is_last_visible: bool,
```
peicodes: why do we need this?

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -1724,27 +1749,67 @@ fn render_groups(
         groups = groups.with_spacing(TABS_MODE_ITEM_SPACING);
     }
 
-    for (visible_tab_index, (tab_index, filtered_pane_ids)) in visible_tabs.iter().enumerate() {
-        // Insert ghost slot before this tab group if the drop would land here.
-        if ghost_insertion_index == Some(*tab_index) {
+    // Consecutive tabs sharing a group_id collapse into a single group container.
+    let total_visible = visible_tabs.len();
+    let mut i = 0;
+    while i < total_visible {
```
peicodes, line 1748: While loops scare me. They can almost always be replaced with declarative loops 

We can talk through this sync and maybe come up with a way to fix this

peicodes (commented):
Left a first round of comments, let's address the clear ones first and pair program to address the ones you're unsure of.

crates/warp_features/src/lib.rs
johnturcoo: Although the tab group experience is not fully working, I agree we can keep this flag on for dogfood, everything here is additive so nothing should break

app/src/workspace/tab_group.rs
johnturcoo: There are no plans to add any new methods on `TabGroupId`, just thought it would be good for type safety and self-documenting code. This pattern exists for `pub struct AIConversationId(Uuid);` and `pub struct AmbientAgentTaskId(NonNilUuid);`, but both of these have additional methods

app/src/workspace/view.rs
johnturcoo: I am updating this, tab groups should always be added to the top of the list, so we can avoid this weird logic

app/src/workspace/view.rs
johnturcoo: Your read is correct, this ff check is quite redundant. As for the feature flag turning off, tabs preserve there order and lose their groups.

app/src/workspace/view/vertical_tabs.rs
johnturcoo: The struct existed because the per tab `insert_after_index` calculation (which used to be inline at the call site) moved across a function boundary when grouping was added. Inside `render_grouped_tab_container` we no longer have the outer loop's position context, so we threaded `is_last_visible` through each member. However, we can just pass this var through without the struct.

app/src/workspace/view/vertical_tabs.rs
johnturcoo: I think that a while loop is best here since we want to step once for ungrouped tabs and step by N for tabs in a group. 

Alternatively I can see `chunk_by` being used, but curious on your opinion here!

peicodes: Let's talk about this sync

peicodes (approved):
Stamping to unblock since this is behind a FF. 

I have three notes:
1. Let's sync on that while loop so I can better understand what's happening (can happen after merge)
2. Let's make sure we test the FF off case (but it sounds like you already did)
3. After you've already received a review, don't do any more squashes. It breaks Github's feature to review only changes since the last review

PR #11767, opened 2026-05-27
Fix WeakModelHandle::upgrade creating zombie handles after last strong handle dropped

kevinchevalier:
## Description
Fix a crash where `WeakModelHandle::upgrade` succeeds after the last strong handle has been dropped, creating a stale `ModelHandle` that panics with "circular model reference" when read.

**Root cause:** `WeakModelHandle::upgrade` only checked `app.models.contains_key()` — but during event processing, the model is still in the store even after all strong handles are dropped (it hasn't been cleaned up by `remove_dropped_items` yet). This allowed `upgrade` to create a new strong handle to a model that was already logically dead, causing a ref-count bounce (0→1) that left the entity in an inconsistent state.

**Fix:** `WeakModelHandle::upgrade` now also checks `RefCounts::is_model_dropped` — if the model's ref count has hit zero and it's pending removal, the upgrade returns `None` instead of creating a zombie handle.

**Production evidence:** [Sentry issue 7502150565](https://warpdotdev.sentry.io/issues/7502150565/) — crash in `<AppContext as ModelAsRef>::model::<GitRepoStatusModel>` during rendering, triggered after sleep/wake + remote server disconnect, where the `GitRepoStatusModel` subscription was dropped and re-acquired via `WeakModelHandle::upgrade` during event processing.

## Linked Issue
- Sentry: https://warpdotdev.sentry.io/issues/7502150565/

## Testing
Added a regression test (`test_weak_handle_fails_after_last_strong_handle_dropped_in_event_callback`) that:
1. A subscriber model holds the only strong handle to an emitter
2. The emitter emits an event
3. In the subscriber's event callback, the last strong handle is dropped
4. A `WeakModelHandle::upgrade` attempted immediately afterward returns `None`

Without the fix, the upgrade succeeds (creating a zombie handle). With the fix, it correctly returns `None`.

Full `warpui_core` test suite passes (276 tests, 0 failures).

## Agent Mode


kevinchevalier:
* **#11767** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11767?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warpui_core/src/core/ref_count_tests.rs
vorporeal, line 1: should this just be `ref_count_tests.rs`?  (and would we want to move `RefCounts` to a `ref_counts.rs` file?)

vorporeal (approved):

PR #11772, opened 2026-05-27
Warp Control CLI v2: contract and spec sync

zachlloyd:
## Description
Root branch of the Warp Control CLI v2 review stack.

This PR establishes the contract, security model, and first working slice for controlling a running Warp instance through the standalone `warpctrl` CLI. Major components include:
- A shared typed local-control protocol and exact-action allowlist used by both the app and CLI.
- The initial `warpctrl` command surface for instance discovery, app metadata, tab creation, structured output, and shell completions.
- An app-side control bridge with deterministic targeting, typed handlers, and exact-action authorization.
- Per-instance filesystem discovery, an owner-authenticated Unix credential broker, and an authenticated loopback HTTP request path.
- A protected Settings > Scripting enablement model, including isolated owner-only Linux fallback storage for the local-control setting.
- Product, technical, and security specs describing the current foundation, trust boundaries, and requirements for future catalog expansion.

The latest update simplifies credentials to authorize one exact action, clarifies the relationship between discovery records, broker sockets, and HTTP endpoints, and keeps the stronger Linux fallback behavior isolated from existing secure-storage consumers.

Demo: https://www.loom.com/share/03740b8c7c40400f9c36eec41d8befca
PR walkthrough: https://www.loom.com/share/b97c3204dcf04dac850aade7d596cc21
Walkthrough URL: https://e94574b9.warp-pr-walkthroughs.pages.dev/

## Linked Issue
- [x] Where appropriate, screenshots or a short video of the implementation are included above.

## Testing
- [x] `./script/format`
- [x] `cargo check -p warpui_extras -p settings -p warp --features warp_control_cli`
- [x] `cargo clippy --workspace --exclude warp_completer --exclude command-signatures-v2 --all-targets --tests -- -D warnings`
- [x] `cargo clippy -p warpui_extras -p settings -p warp --features warp_control_cli --all-targets -- -D warnings`
- [x] `cargo nextest run --no-fail-fast --workspace --exclude command-signatures-v2 local_control` — 45 passed
- [x] `git diff --check` and `git diff --cached --check`
- [ ] Exact primary presubmit workspace Clippy command is blocked before linting because `command-signatures-v2` requires Yarn 4.0.1 via Corepack while this environment resolves Yarn 1.22.22; the workspace run excluding only that unrelated crate passes.
- [ ] Linux-only secure-storage tests are not runnable on this macOS checkout because no Linux Rust target is installed.

### Screenshots / Videos
Demo and walkthrough links are included above.

## Agent Mode
- [x] Warp Agent Mode - This PR was created and updated via Warp's AI Agent Mode

## Warp Agent artifacts
Conversation: https://staging.warp.dev/conversation/8a7dca3a-a416-4378-bd49-f19fb78d6e2e


crates/local_control/src/protocol.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+/// Top-level request sent by a local-control client to a Warp instance.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct RequestEnvelope {
+    pub protocol_version: u32,
+    pub request_id: Uuid,
+    #[serde(default)]
+    pub target: TargetSelector,
+    pub action: Action,
```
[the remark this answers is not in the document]
zachlloyd: Addressed in `ed284890` and propagated through the `zach/warp-cli-v2/*` stack. `TECH.md` now documents the implemented `RequestEnvelope` / `Action` serde shape, with `action.kind` and `action.params` nested under `action`, and also updates the response examples to match the tagged `ResponseEnvelope` / `ControlResponse` implementation.

app/src/local_control/resolver.rs
```diff
@@ -0,0 +1,164 @@
...
+fn active_or_single_window_id(
+    ctx: &mut ModelContext<LocalControlBridge>,
+    action: ActionKind,
+) -> Result<WindowId, ControlError> {
+    if let Some(window_id) = ctx.windows().active_window() {
+        return Ok(window_id);
+    }
+    let window_ids = ctx.window_ids().collect::<Vec<_>>();
+    match window_ids.as_slice() {
+        [window_id] => Ok(*window_id),
```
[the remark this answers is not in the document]
zachlloyd: Addressed in `88b0bd79` and propagated through the `zach/warp-cli-v2/*` stack. We kept the implemented resolver behavior and updated `PRODUCT.md`, `SECURITY.md`, and `TECH.md` to document it: an omitted or active window selector first uses the active Warp window, may fall back to the sole existing window when that is the only deterministic target, returns `missing_target` when no window exists, and returns `ambiguous_target` when multiple windows exist with no active window.

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+The visible UI setting is not enough by itself. The authoritative mode must be stored in protected local storage that ordinary same-user apps cannot update by writing a plist, settings database row, registry key, JSON file, or synced cloud preference. This avoids turning outside-Warp control into a feature that any process can silently enable before invoking `warpctrl`.
+Current foundation implementation note: the mode is represented in the typed `LocalControlSettings` group as a private, local-only setting. The implemented setting must use `private: true`, `SyncToCloud::Never`, an explicit private storage key, and no `toml_path`, so it is excluded from `settings.toml`, the generated settings schema, Settings Sync, Warp Drive, and user-editable or server-backed settings surfaces. This private-settings path is an interim storage boundary, not the final protected-storage requirement; before public shipment, this authoritative mode must move to platform protected storage where available.
```
Legoben: still nervous that this could affect settings / state that _would_ persist across the network; ie a bad actor creates a malicious notebook, which opens next time the user opens the app/WASM.

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+Enablement requirements:
+- The mode is local-only and must not sync through Settings Sync, Warp Drive, or server-backed user preferences.
+- The implemented foundation setting must remain private and absent from user-visible settings files, generated schemas, local-control settings read/write commands, and any allowlisted settings mutation catalog.
+- Only the running Warp app, through the Settings > Scripting UI, should be able to change the authoritative mode.
+- `warpctrl`, shell scripts, config files, command-line flags, registry edits, defaults writes, and direct local-control protocol requests must not be able to enable or widen the mode.
```
Legoben, line 104: is there a (realistic) tamperproof way to do this? don't these just live in sqlite?

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+```
+## Discovery registry
+Each participating Warp process writes a discovery record in a secure per-user local-control directory. Discovery records are metadata, not a full control-authority model.
+A discovery record should contain:
+- opaque `instance_id`;
+- PID and process start timestamp;
+- channel and build metadata;
+- protocol version and supported capability summary;
+- loopback endpoint for the instance-local control listener;
+- credential reference or bootstrap credential metadata, not necessarily the full control credential.
```
Legoben: could these be issued JIT?

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+- Metadata reads require an explicit `read_metadata` grant.
+- Underlying data reads require an explicit `read_underlying_data` grant.
+- App-state mutations require an explicit `mutate_app_state` grant.
+- Metadata/configuration mutations require an explicit `mutate_metadata` or `mutate_configuration` grant.
+- Underlying data mutations require an explicit `mutate_underlying_data` grant and should require approval or policy for unattended automation.
+- User-authenticated data reads or mutations require an explicit `authenticated_user` grant and an allowed authenticated action family in addition to the data-category grant.
+- Integrations should receive the narrowest grant needed for the configured workflow.
+The broker must not issue broad authority merely because the request came from the signed `warpctrl` binary. It should evaluate the requested permission category, target scope, configured policy, execution context, and whether user approval is required. The CLI must not mint its own authority. It can request, load, and present credentials, but the app bridge remains the enforcement point for these safety grants.
```
Legoben: I'm not sure how reasonable this is - imo the callers shouldn't rly need to know/care about scopes. it should be the app/settings page/consent screens responsibility to manage these. if an app requests something out of scope and not explicitly denied, a new consent screen pops up.

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+- loopback endpoint for the instance-local control listener;
+- credential reference or bootstrap credential metadata, not necessarily the full control credential.
+Discovery rules:
+- Records must be readable only by the owning user.
+- POSIX records must use owner-only permissions such as `0600` for files and a non-world-readable directory.
+- Windows records must live under the current user's app data directory with ACLs limited to the current user, Administrators, and SYSTEM.
+- When outside-Warp control is disabled, records must not publish actionable control endpoints or credential references for external clients. A minimal disabled-status record is acceptable only if it contains no authority.
+- The CLI must prune or ignore stale records whose PID is gone or whose health/protocol check fails.
+- If multiple compatible instances are ambiguous, the CLI must require explicit `--instance` selection.
+- Discovery metadata must not expose terminal contents, environment variables, auth tokens for cloud services, raw local-control credentials, or mutating capability grants.
```
Legoben, line 299: I think discovery should also turn on "only when needed"? maybe thats handled in the settings page? maybe theres a non-discovery way for scripts to ask "enable scripting mode"?

rly want to prevent execution happening because either:
1) discovery exists passively in the background
2) user turned this feature on for something one-off and forgot about it

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+- validation rules;
+- a documented state/data category and permission category;
+- a documented `requires_authenticated_user` value;
+- a documented allowed execution context, including whether external clients can run it or whether it is limited to verified Warp-terminal invocations;
+- local app-side safety-grant checks;
+- deterministic target resolution;
+- a handler that reuses existing user-visible app behavior where possible;
+- typed success and error responses.
+Adding a new action should be additive and reviewable: extend the protocol enum, implement validation, map the action to a state/data category, declare whether it requires an authenticated user, declare its allowed execution contexts, add a handler, and add tests for authentication, safety-policy denial, authenticated-user denial, selector failure, and success behavior.
+## Browser and localhost protections
```
Legoben, line 427: sure, but more than browsers can send HTTP requests- shouldn't _rely_ on CORS/etc protection here

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+Loopback is not sufficient by itself because browsers can send requests to localhost.
+Required protections:
+- No permissive CORS on control endpoints.
```
Legoben, line 431: LLM: DNS-rebinding / missing Host/Origin validation against the browser vector. Browser defense currently rests on "no permissive CORS" + token secrecy. But CORS governs whether JS can read the response, not whether the request is sent and side effects happen

Add an explicit rule: reject any request carrying an Origin header, and reject any request whose Host isn't literally 127.0.0.1:<port>.

specs/warp-control-cli/SECURITY.md
```diff
@@ -0,0 +1,505 @@
...
+- POSIX records must use owner-only permissions such as `0600` for files and a non-world-readable directory.
+- Windows records must live under the current user's app data directory with ACLs limited to the current user, Administrators, and SYSTEM.
+- When outside-Warp control is disabled, records must not publish actionable control endpoints or credential references for external clients. A minimal disabled-status record is acceptable only if it contains no authority.
+- The CLI must prune or ignore stale records whose PID is gone or whose health/protocol check fails.
+- If multiple compatible instances are ambiguous, the CLI must require explicit `--instance` selection.
+- Discovery metadata must not expose terminal contents, environment variables, auth tokens for cloud services, raw local-control credentials, or mutating capability grants.
+## Credential model
+The full `warpctrl` catalog requires scoped credentials. A single shared full-power bearer token is not sufficient once automation, underlying data reads, app-state mutations, metadata/configuration mutations, and underlying data mutations are supported.
+### Credential properties
```
Legoben, line 303: LLM:

The bearer token is phishable by a local port squatter. Every layer authenticates the client to the app (token, proof, broker). Nothing authenticates the app to the client before the client hands over its credential. The loopback listener binds an ephemeral port and writes it into the discovery record. If a Warp process dies without gracefully unlinking its record (SIGKILL, crash, OOM), the record lingers with a now-free port. A same-user process that binds that port before pruning runs will receive the Authorization: Bearer <token> that warpctrl sends to the recorded endpoint. 

The health-check-based pruning doesn't save you — a squatter just answers the health check. This quietly undermines the macOS Keychain story: secure storage protects the token at rest so only Warp-signed code can unwrap it, but warpctrl will then transmit that unwrapped token to whoever owns the recorded port. 

The fix is mutual authentication — have the client verify a server-held secret (or a record-embedded server identity) before sending the credential — or move to a UDS.

Legoben (commented):
only looked thru `README.md` and `SECURITY.md` - seems overall good to me but with many footguns if something is off in implementation

specs/warp-control-cli/PRODUCT.md
```diff
@@ -0,0 +1,508 @@
...
+## Problem
+Warp already has rich interactive actions, but they are primarily reachable through UI, keybindings, menus, or deeplinks. Agents can use native tools for files, code, shell commands, MCP calls, and many context reads, but they cannot reliably operate Warp's own product surfaces: arranging the user's workspace, focusing the correct pane, opening Warp Drive objects, presenting settings, or recovering from ambiguous UI state. Developers also cannot reliably compose those same actions into shell scripts, demos, automation, or agent workflows, and there is no general local protocol for addressing a specific running Warp instance, window, pane, session, terminal block, Warp Drive object, or other uniquely named Warp entity.
+## Goals / Non-goals
+Goals:
+- Provide a first-class, scriptable standalone `warpctrl` binary for controlling running Warp app processes.
```
alokedesai: Biggest feedback from reading the product / tech spec is that I think creating a separate binary for `warpctrl` is probably not right.

1) It's going to cause distribution headaches . Users will either have to download a separate binary or we'll have to bundle the binary as an additional resource in our existing app bundle and then put it on the right path
2)  We have to make the protocol version and channel agnostic, which makes the protocol more complicated than it needs to be. 

Can we just bundle this into the existing Warp CLI binary? Similar to how we set up `oz` as a wrapper script that just calls out to the actual warp binary, we can expose a `warpctrl` script that does the same thing but with a special `--warpctrl` flag to differentiate between the two

@bnavetta for this thoughts

specs/warp-control-cli/SECURITY.md
zachlloyd: not sure i follow - mind elaborating?

specs/warp-control-cli/PRODUCT.md
zachlloyd: i didn't realize that's how oz worked.  

my only concern with that approach is around potential latency of spawning a full warp process on every invocation.  if that's not an issue in practice, i def agree with doing this the same way as oz (and that was in fact my intention, it's a pretty big miss that it wasn't done that way)

bnavetta: Yeah, I think bundling into the existing binary makes sense.

Latency shouldn't be an issue - the CLI is set up so it doesn't have to fully bootstrap the app for commands that don't need it. For example, `oz dump-debug-info` is more-or-less instantaneous, while `oz whoami` is slower because it bootstraps the app and checks auth state on the server.

zachlloyd: perfect.  super helpful!

specs/warp-control-cli/SECURITY.md
Legoben: thought that if you opened a notebook locally, that "last viewed" state would persist to the cloud, which doesn't actually seem to be the case

zachlloyd: [Warp Agent] Added a clarification that local-control actions which open/focus/view cloud-backed objects must not create unexpected synced durable side effects unless the action is intentionally classified and authorized as a synced-state mutation. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/SECURITY.md
zachlloyd: [Warp Agent] Updated the spec and implementation so the authoritative local-control mode uses Warp's secure storage provider instead of ordinary private preferences/UserDefaults/SQLite-style storage. The spec now frames this as tamper-resistant platform storage, not a perfect same-user tamperproof guarantee. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/SECURITY.md
zachlloyd: [Warp Agent] Clarified that this is the implemented flow: discovery publishes endpoint/broker metadata only, and warpctrl requests a short-lived action-scoped credential just in time from /v1/control/credentials. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/SECURITY.md
zachlloyd: [Warp Agent] Updated the spec to make clear callers request typed actions/capabilities, while the app-owned broker maps those requests to grants, policy, and consent prompts. This matches the implementation, where the CLI requests credentials for an ActionKind and the app bridge/broker enforce the metadata-derived permission category. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/SECURITY.md
zachlloyd: [Warp Agent] Clarified that discovery must not publish actionable endpoints or broker references unless the protected mode currently enables that invocation context, and noted future temporary/session-scoped enablement UX. The implementation already matches the key requirement by omitting endpoint/broker metadata unless outside-Warp control is enabled. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/SECURITY.md
zachlloyd: [Warp Agent] Expanded the browser/localhost section to state that CORS is not the primary control and that non-browser local clients must still be gated by credentials, invocation-context checks, app-side authorization, and endpoint hardening. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/SECURITY.md
zachlloyd: [Warp Agent] Added the Host/Origin requirements to the spec and implemented them on both credential and control endpoints: requests with Origin are rejected, and Host must exactly match the selected 127.0.0.1:<port>. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/SECURITY.md
zachlloyd: [Warp Agent] Clarified the stale-port/mutual-auth analysis: the current JIT broker avoids sending a stored Warp-held bearer token to a stale discovery port, but any future stored/bootstrap credential flow must verify server identity first or use UDS/peer identity. Resolved in https://github.com/warpdotdev/warp/commit/ff18b065.

specs/warp-control-cli/TECH.md
```diff
@@ -0,0 +1,603 @@
...
+  "action": {
+    "kind": "tab.create",
+    "params": {}
+  }
+}
+```
+Recommended success response shape matches `ResponseEnvelope` and the tagged `ControlResponse::Ok` variant:
+```json
+{
+  "protocol_version": 1,
```
alokedesai, line 94: we shouldn't really need versions at all anymore if the warpctrl and warp GUI are always the same version (which should be the case now)

app/src/local_control/handlers/metadata.rs
```diff
@@ -0,0 +1,80 @@
...
+use serde::Serialize;
+use warp_core::channel::ChannelState;
+#[derive(Serialize)]
+struct InstanceResponse<'a> {
+    action: &'static str,
+    instance_id: Option<&'a str>,
+    pid: u32,
+    channel: String,
+    app_id: String,
+    app_version: Option<&'static str>,
```
alokedesai: Do we need these anymore if we're using the same binary as the parent Warp GUI app. We should know at build time, no?

crates/local_control/src/catalog.rs
```diff
@@ -0,0 +1,472 @@
...
+        KeybindingGet => { name: "keybinding.get", status: Stub, authenticated_user: false, contexts: Any, state: MetadataRead, target: Keybinding, params: BindingName, result: KeybindingMetadata },
+    }
+
+    action {
+        ActionList => { name: "action.list", status: Stub, authenticated_user: false, contexts: Any, state: MetadataRead, target: Action, params: None, result: CapabilityList },
+        ActionInspect => { name: "action.inspect", status: Stub, authenticated_user: false, contexts: Any, state: MetadataRead, target: Action, params: ActionName, result: CapabilityMetadata },
+    }
+
+    surface {
+        SurfaceSettingsOpen => { name: "surface.settings.open", status: Stub, authenticated_user: false, contexts: Any, state: AppStateMutation, target: Surface, params: PageQuery, result: Acknowledgement },
```
alokedesai: what is the name used for? Is it user visible?

script/macos/bundle
```diff
@@ -558,10 +558,20 @@ EOF
...
   chmod +x "$CLI_SCRIPT_PATH"
 
+  WARPCTRL_SCRIPT_PATH="$BUNDLED_RESOURCES_DIR/bin/warpctrl"
+  echo "Creating warpctrl wrapper script..."
+  cat > "$WARPCTRL_SCRIPT_PATH" << 'EOF'
+#!/bin/bash
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+exec -a "$0" "$script_dir/../../MacOS/WARP_BIN_PLACEHOLDER" --warpctrl "$@"
+EOF
+  sed -i '' "s/WARP_BIN_PLACEHOLDER/$WARP_BIN/" "$WARPCTRL_SCRIPT_PATH"
```
alokedesai: we'll only want to do this if warpctrl is enabled, no?

alokedesai (approved):

app/src/local_control/handlers/metadata.rs
Legoben: +1 - the less info we can include in these the better generally

app/src/local_control/bridge.rs
```diff
@@ -0,0 +1,129 @@
...
+                ),
+            );
+        }
+        match request.action.kind {
+            ActionKind::InstanceList => {
+                if let Err(error) =
+                    ensure_action_allowed(grant.invocation_context, request.action.kind, ctx)
+                {
+                    return ResponseEnvelope::error(request.request_id, error);
+                }
```
Legoben: should this `ensure_action_allowed` be brought out of the match blocks? less likely to forget one in the future

app/src/local_control/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+                Self {
+                    _runtime: None,
+                    control_endpoint: None,
+                    registered_instance: None,
+                }
+            }
+        }
+    }
+
+    fn start(ctx: &mut ModelContext<Self>) -> Result<Self, ControlError> {
```
Legoben: maybe me missing it, but how to we prevent this from running multiple times?

app/src/local_control/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+                Json(ErrorResponseEnvelope::new(ControlError::with_details(
+                    ErrorCode::InvalidRequest,
+                    "failed to decode local-control credential request",
+                    err.to_string(),
+                ))),
+            )
+                .into_response();
+        }
+    };
+    if request.protocol_version != PROTOCOL_VERSION {
```
Legoben: seems like there should be a helper that this + bridge + any consumers after this PR use to check PROTOCOL, enablement status, etc. Want to avoid these getting left out in the future (or forgetting a place if a new check is added)

app/src/local_control/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+        ),
+    };
+    let status = match &response.response {
+        ControlResponse::Ok { .. } => StatusCode::OK,
+        ControlResponse::Error { .. } => StatusCode::BAD_REQUEST,
+    };
+    (status, Json(response)).into_response()
+}
+
+pub(crate) fn validate_loopback_headers(
```
Legoben, line 672: these are spoof-able- nice to have but shouldn't be relied on

app/src/local_control/mod.rs
```diff
@@ -0,0 +1,469 @@
...
+                    _runtime: None,
+                    control_endpoint: None,
+                    registered_instance: None,
+                }
+            }
+        }
+    }
+
+    fn start(ctx: &mut ModelContext<Self>) -> Result<Self, ControlError> {
+        ensure_feature_enabled()?;
```
Legoben, line 195: looks like this only checks the flag- I think we also shouldn't start this server unless the user has opted in via the settings. Wdyt?

For an app to ask the user to consent for the first time, it could call a CLI or use some other channel?

app/src/local_control/permissions.rs
```diff
@@ -0,0 +1,82 @@
...
+pub(super) fn ensure_action_allowed(
+    context: InvocationContext,
+    action: ActionKind,
+    ctx: &mut ModelContext<LocalControlBridge>,
+) -> Result<(), ControlError> {
+    let settings = LocalControlSettings::as_ref(ctx);
+    ensure_settings_allow_action(settings, context, action)
+}
+
+pub(crate) fn ensure_settings_allow_action(
```
Legoben, line 55: spec mentions different permissions levels for different actions - assume thats still to be done?

app/src/settings/local_control.rs
```diff
@@ -0,0 +1,282 @@
...
+        let validated = self.validate(new_value);
+        if self.value() != &validated || self.is_explicitly_set != explicitly_set {
+            self.inner = validated;
+            self.is_explicitly_set = explicitly_set;
+            Self::emit_changed(ctx, settings::ChangeEventReason::LocalChange);
+        }
+        Ok(())
+    }
+
+    fn set_value_from_cloud_sync(
```
Legoben, line 157: how does this get used? i don't think we want this to be cloud sync-able

crates/local_control/src/auth.rs
```diff
@@ -0,0 +1,213 @@
...
+};
+
+/// Bearer token used to authorize a single scoped local-control credential.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct AuthToken(String);
+
+impl AuthToken {
+    pub fn generate() -> Self {
+        let mut bytes = [0u8; 32];
+        rand::rngs::OsRng.fill_bytes(&mut bytes);
```
Legoben, line 24: is the the right / most secure random to use?

specs/warp-control-cli/TECH.md
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. I kept the request protocol version as a defensive schema guard for stale copied JSON, stale wrappers, and future external clients, but updated TECH.md so it no longer implies normal compatibility negotiation between separately versioned CLI and GUI binaries.

app/src/local_control/handlers/metadata.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. I removed app_version from the metadata responses while keeping channel, app_id, and protocol_version for discovery/diagnostics and defensive request validation.

crates/local_control/src/catalog.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. I documented that name is the stable public CLI/API action identifier exposed through discovery/help/wire payloads, e.g. tab.create.

script/macos/bundle
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. The macOS app-bundle wrapper is now created only when the warp_control_cli build feature is present, and the standalone warpctrl artifact path errors if that feature is missing.

app/src/local_control/bridge.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. I moved the action-allowed check out of the individual dispatch arms so implemented actions pass through the shared bridge validation path before handler execution.

app/src/local_control/mod.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. LocalControlServer now refreshes from settings, starts only when not already running, and stops/clears discovery when disabled, so the singleton model can exist without repeatedly starting listeners.

app/src/local_control/mod.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. I added a shared protocol-version validation helper and reused the common feature/protocol/settings checks in the credential and bridge paths to reduce drift as new consumers are added.

app/src/local_control/mod.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. I clarified that the Host/Origin checks are browser-origin hardening and stale-endpoint checks only; scoped bearer credentials and grant validation remain the authorization boundary.

app/src/local_control/mod.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. The listener now starts only when Settings enables outside-Warp control, and it shuts down/clears discovery when the setting is disabled. First-consent UX remains separate follow-up work.

app/src/local_control/permissions.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. The current grant now carries the action permission category from catalog metadata, and grant validation rejects permission-category mismatches. Broader user-facing permission policy/UI can still land in later slices.

app/src/settings/local_control.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. The local-control mode remains private and SyncToCloud::Never, cloud-sync writes are ignored for this setting, and tests assert both the private/never-sync invariant and that cloud sync cannot enable local control.

crates/local_control/src/auth.rs
zachlloyd: [Warp Agent] Addressed in https://github.com/warpdotdev/warp/commit/6bc74d369fc7bec28fc0754ea25f1fba332b72f4. I added a doc comment explaining that these bearer tokens use 32 bytes from the operating-system CSPRNG via OsRng because they are authentication material, not deterministic or fast userspace randomness.

zachlloyd:
[Warp Agent] Re: https://github.com/warpdotdev/warp/pull/11772#pullrequestreview-4382800980 — agreed. The implementation has a lot of security-sensitive footguns if it drifts from the specs, so we are validating that the implementation matches the spec, including using the validate-implementation-matches-spec skill from common-skills.

Legoben (approved):

PR #11779, opened 2026-05-27
Store canonicalized PWD for sessions to avoid re-canonicalizing a bunch of times.

### LINKED ISSUE #9920: Tab switching becomes sluggish for WSL tabs opened via UI
### Pre-submit Checks

### Describe the bug

I'm currently on windows 11. Whenever I open Warp it starts the terminal session using `powershell` by default.
Sometimes I need to work on WSL so I enter the environment using `wsl` commands.  Now if multiple tabs are open in this way where they are either running powershell or wsl via command. The tab switching works fine, without any visible lag.

However, if  I open new tabs via the topbar icon and selecting the `debian` session (as shown in the image), switching between those tabs feels sluggish. It even sometimes introduce a delay of 1-2 seconds between tab switching 

Image

<br><br/>

In the attached video
- the first three terminals are opened either `powershell` or opened via `wsl` command.
- the last two tabs are opened via UI selecting the `debian`  session.
## Video

https://github.com/user-attachments/assets/fbeb01be-897f-4755-9de6-57ad68475644

### To reproduce

Steps to reproduce
- Open multiple terminal tabs using the topbar icon and select any WSL distro.
- try switching between those tabs repeatedly
Notice the visible lag on switching the tabs.

### Expected behavior

Tab switching should remain smooth and shouldn't give a delay.

_No response_

### Operating system (OS)

Windows

Windows 11

### Shell Version

GNU Bash 5.2.37

### Current Warp version

v0.2026.04.27.15.32.stable_03

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

acarl005:
## Description

This PR mainly fixes an issue where switching to a WSL tab freezes the app for a second or two. The root cause is `std::fs::canonicalize` (or anything derived from that) being called repeatedly on WSL paths. Canonicalizing a path through WSL, e.g. `\\WSL$\Ubuntu\home\username\somefile` is slow because it crosses the VM boundary. It's not **so** slow that we have to avoid it entirely. However, it's slow enough that calling it 10+ times on the main thread in succession causes noticeable lag. This is actually happening. The lag grows linearly with the number of tabs in the workspace.

To make things worse, some callees are defensively re-canonicalizing even though their callers had already canonicalized a path, e.g.

- caller: https://github.com/warpdotdev/warp/blob/831de4bec9bad81bd66fe35e384e026d89d13c6f/app/src/terminal/view.rs#L13727
- callee: https://github.com/warpdotdev/warp/blob/831de4bec9bad81bd66fe35e384e026d89d13c6f/crates/repo_metadata/src/repositories.rs#L227

I propose fixing this with the following:

1. Canonicalize the PWD after precmd and storing that as a `CanonicalizedPath` on a new field `TerminalView::canonical_pwd`.
2. Propagate `TerminalView::canonical_pwd` down to all relevant callees on the code path for activating atab.
3. (When pragmatic) If any callees are redundantly canonicalizing, remove the redundant canonicalization and change the path parameter to type `CanonicalizedPath` to enforce the requirement without duplicating work.

It should be noted that not _all_ callees actually want/need canonicalized paths. We still need to expose the original path for those cases. We will keep the existing reader methods for that, but rename them in order to make it very clear that these paths are for display purposes.

- Rename `TerminalView::pwd_if_local` to `display_pwd_if_local`. Add an analogous `canonical_pwd_if_local`.
- Rename `BlockMetadata::current_working_directory` to `display_working_directory`.

Most of this diff is just updating callers for this rename.

### Alternatives Considered

Why not use `StandardizedPath` instead of `CanonicalizedPath`? I prefer `CanonicalizedPath` as it will prevent more bugs. The problem with `StandardizedPath` is that it wraps `TypedPathBuf` which itself is an enum for UNIX or Windows paths.

```
pub enum TypedPathBuf {
    Unix(UnixPathBuf),
    Windows(WindowsPathBuf),
}
```

That allows a UNIX system to hold a Windows path or vice-versa. For this case, we actually want to enforce that the path has the native host encoding. This is relevant for Windows specifically. It can have sessions that use UNIX paths, e.g. in the case of Git Bash or WSL. However, all the callees that take paths should be dealing with the Windows-native equivalents, i.e. as `\\WSL$\Ubuntu\home\username\somefile` not `/home/username/somefile`. Because `CanonicalizedPath` wraps `PathBuf`, it prevents it from holding a path with the wrong platform encoding.

## Linked Issue

Closes #9920

## Testing

The issue can be repro'd by opening a bunch of WSL tabs, it may take 10-20 tabs on a fast machine to start noticing the lag. Just activate any of the WSL tabs and it will freeze the UI for a bit.

On this branch, there is no longer perceptible lag.

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -107,6 +107,7 @@ use regex::Regex;
 #[cfg(not(target_family = "wasm"))]
 use repo_metadata::repositories::DetectedRepositories;
 use repo_metadata::repositories::RepoDetectionSource;
+use repo_metadata::CanonicalizedPath;
 use serde::Serialize;
 use serde_json::json;
 use session_sharing_protocol::common::{
@@ -2563,7 +2564,7 @@ pub struct TerminalView {
     /// Canonical (IO-validated) working directory, cached once when the shell reports a new CWD
     /// via [`ModelEvent::BlockMetadataReceived`]. `None` for remote sessions or when the directory
     /// no longer exists on disk.
-    canonical_pwd: Option<repo_metadata::CanonicalizedPath>,
+    canonical_pwd: Option<CanonicalizedPath>,
 
     block_text_selection_start_position: Option<Vector2F>,
 
@@ -11126,7 +11127,7 @@ impl TerminalView {
                                         };
 
                                         let Ok(active_directory) =
-                                            repo_metadata::CanonicalizedPath::try_from(
+                                            CanonicalizedPath::try_from(
                                                 active_directory,
                                             )
                                         else {
@@ -22428,7 +22429,7 @@ impl TerminalView {
 
     /// Returns the cached canonical CWD without any filesystem I/O.
     /// Populated once per [`ModelEvent::BlockMetadataReceived`] event for local sessions.
-    pub fn canonical_pwd_if_local(&self) -> Option<&repo_metadata::CanonicalizedPath> {
+    pub fn canonical_pwd_if_local(&self) -> Option<&CanonicalizedPath> {
         self.canonical_pwd.as_ref()
     }
 
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -13069,9 +13069,9 @@ impl TerminalView {
         self.any_session_contains_remote_blocks |= self.active_block_is_considered_remote(ctx);
         self.update_focused_terminal_info(ctx);
 
-        if let Some(working_directory) = self.display_pwd_if_local(ctx) {
+        if let Some(cwd) = self.canonical_pwd_if_local() {
+            let path_buf = cwd.as_path_buf().clone();
             CodebaseIndexManager::handle(ctx).update(ctx, |manager, _ctx| {
-                let path_buf = PathBuf::from(&working_directory);
                 manager.handle_session_bootstrapped(&path_buf);
             });
         }
@@ -13514,9 +13514,8 @@ impl TerminalView {
 
     // Show or hide codebase index speedbump depending when a settings change happens.
     fn check_codebase_index_speedbump_on_settings_changed(&mut self, ctx: &mut ViewContext<Self>) {
-        if let Some(working_directory) = self.display_pwd_if_local(ctx) {
-            let path_buf = PathBuf::from(&working_directory);
-            self.update_repo_banner_state(path_buf, ctx);
+        if let Some(cwd) = self.canonical_pwd_if_local() {
+            self.update_repo_banner_state(cwd.as_path_buf().clone(), ctx);
         }
     }
 
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -11058,6 +11058,15 @@ impl TerminalView {
                             active_directory.to_string()
                         };
 
+                        self.canonical_pwd = if is_local {
+                            CanonicalizedPath::try_from(
+                                PathBuf::from(&directory_for_detection).as_path(),
+                            )
+                            .ok()
+                        } else {
+                            None
+                        };
+
                         let fut = detect_possible_git_repo(
                             session_type,
                             &directory_for_detection,
@@ -11120,17 +11129,9 @@ impl TerminalView {
                                     #[cfg(feature = "local_fs")]
                                     {
                                         let Some(active_directory) =
-                                            me.active_session_path_if_local(ctx)
+                                            me.canonical_pwd.clone()
                                         else {
                                             me.clear_git_repo_status(ctx);
-                                            return;
-                                        };
-
-                                        let Ok(active_directory) =
-                                            CanonicalizedPath::try_from(
-                                                active_directory,
-                                            )
-                                        else {
                                             return;
                                         };
 
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -2560,6 +2560,10 @@ pub struct TerminalView {
 
     sessions: ModelHandle<Sessions>,
     active_block_metadata: Option<BlockMetadata>,
+    /// Canonical (IO-validated) working directory, cached once when the shell reports a new CWD
+    /// via [`ModelEvent::BlockMetadataReceived`]. `None` for remote sessions or when the directory
+    /// no longer exists on disk.
+    canonical_pwd: Option<repo_metadata::CanonicalizedPath>,
 
     block_text_selection_start_position: Option<Vector2F>,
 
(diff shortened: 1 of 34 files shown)


app/src/workspace/tab_settings.rs
```diff
@@ -201,17 +201,17 @@ settings::macros::implement_setting_for_enum!(
 impl DirectoryTabColors {
     /// Returns the configured tab color for a directory using longest-prefix matching.
     /// Returns `None` if no configured directory is a prefix of `dir`.
+    ///
+    /// The caller should provide an already-canonical path (e.g. from
+    /// `CanonicalizedPath`) to avoid filesystem I/O on every render frame.
     pub fn color_for_directory(&self, dir: &Path) -> Option<DirectoryTabColor> {
```
acarl005, line 205: This method is called on the render path, i.e. on every frame, so it's especially important not to canonicalize here. Ideally this also would take a `CanonicalizedPath` instead of `Path` as a param. However, not all callers are passing the `TerminalView::canonical_pwd` that we added. Some callers are passing paths from code panes, and those are not canonical paths. I think they should be, but I'm trying to limit scope here.

crates/repo_metadata/src/repositories.rs
```diff
@@ -214,16 +217,20 @@ impl DetectedRepositories {
...
+        DirectoryWatcher::as_ref(ctx).get_watched_directory_for_path(&canonical)
     }
 
     /// Given a local or remote path, return its corresponding repo root.
     /// This does not run the check against the actual file system.
     /// Instead it checks against our cached path to root mapping.
+    ///
+    /// For local paths, the caller should provide an already-canonical path
+    /// (e.g. from `CanonicalizedPath`) to avoid a filesystem round-trip.
     pub fn get_root_for_path(&self, path: &LocalOrRemotePath) -> Option<LocalOrRemotePath> {
```
acarl005, line 231: Ideally this method could also enforce that local paths are already canonicalized. However, that would require changing `LocalOrRemotePath::Local(PathBuf)` to hold a `CanonicalizedPath` instead. Again, I'd like to limit scope here. That would be a massive refactor.

crates/warp_util/src/path.rs
```diff
@@ -478,15 +483,8 @@ pub fn convert_wsl_to_windows_host_path(
...
-            windows_path
+            PathBuf::try_from(windows_path).map_err(WSLPathConversionError::CouldNotConvertToPath)
         }
-    };
-    let pathbuf =
-        PathBuf::try_from(windows_path).map_err(WSLPathConversionError::CouldNotConvertToPath)?;
-    // Many directories are symlinks into the underlying file-system location in Windows.
-    match std::fs::read_link(&pathbuf) {
-        Ok(linked_file) => Ok(linked_file),
-        Err(_) => Ok(pathbuf),
```
acarl005, line 489: Didn't mention it in the PR description but we should also avoid calling 	`std::fs::read_link` on WSL paths. That is bad for a different reason. It always fails b/c `std::fs::read_link` on a Windows system looks for Windows style symlinks and if you pass it a WSL path which is a linux style symlink it just fails. Moved this call into the specific branch so that only the non-WSL path calls it.

crates/repo_metadata/src/repositories.rs
```diff
@@ -214,16 +217,20 @@ impl DetectedRepositories {
...
         match path {
             LocalOrRemotePath::Local(local_path) => {
-                let std_path = StandardizedPath::from_local_canonicalized(local_path).ok()?;
+                let std_path = StandardizedPath::try_from_local(local_path).ok()?;
```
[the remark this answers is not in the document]
acarl005: Not doing this anymore

app/src/workspace/view.rs
```diff
@@ -5242,22 +5242,18 @@ impl Workspace {
...
-            return;
-        };
-
-        let cwd_path = Path::new(&cwd);
-        let color = TabSettings::as_ref(ctx)
-            .directory_tab_colors
-            .value()
-            .color_for_directory(cwd_path)
-            .and_then(|c| c.ansi_color());
+            .and_then(|tv| tv.as_ref(ctx).canonical_pwd_if_local().cloned())
```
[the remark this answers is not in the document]
acarl005: fixed

crates/repo_metadata/src/repositories.rs
```diff
@@ -212,14 +215,19 @@ impl DetectedRepositories {
...
-        let local_path = root.to_local_path()?;
-        DirectoryWatcher::as_ref(ctx).get_watched_directory_for_path(local_path)
+        let canonical = CanonicalizedPath::try_from(path).ok()?;
+        let root = self.get_local_root_for_canonical_path(&canonical)?;
+        let root_canonical = CanonicalizedPath::try_from(root).ok()?;
+        DirectoryWatcher::as_ref(ctx).get_watched_directory_for_path(&root_canonical)
     }
 
-    /// Given a local or remote path, return its corresponding repo root.
-    /// This does not run the check against the actual file system.
```
acarl005, line 221: > "This does not run the check against the actual file system."

Untrue. It was canonicalizing which does indeed to file I/O on Windows.

acarl005:
@oz-agent fix the test failures

app/src/terminal/view.rs
```diff
@@ -2570,6 +2561,10 @@ pub struct TerminalView {
 
     sessions: ModelHandle<Sessions>,
     active_block_metadata: Option<BlockMetadata>,
+    /// Canonical (IO-validated) working directory, cached once when the shell reports a new CWD
+    /// via [`ModelEvent::BlockMetadataReceived`]. `None` for remote sessions or when the directory
+    /// no longer exists on disk.
+    canonical_pwd: Option<CanonicalizedPath>,
```
acarl005, line 2567: This part is important

app/src/terminal/view.rs
```diff
@@ -11136,6 +11058,15 @@ impl TerminalView {
...
                         };
 
+                        self.canonical_pwd = if is_local {
+                            CanonicalizedPath::try_from(
+                                PathBuf::from(&directory_for_detection).as_path(),
+                            )
+                            .ok()
+                        } else {
+                            None
+                        };
```
acarl005, line 11068: Also very important

kevinyang372 (commented):
Leaving a review here so I can get notified on re-review request

Discussed offline but the biggest feedback is we should consider using StandardizedPath instead of CanonicalPath

acarl005:
Taken over by #12492

PR #11789, opened 2026-05-27
Fix working directories clean up

MaggieShan:
## Description
* Context: https://warpdev.slack.com/archives/C08KTPNQN65/p1779918765473719 
* We were dropping diffstatemodels on pane close despite other pane groups still relying on the model 
* The fix introduces a reverse map to keep track of all pane groups that reference a repo path - this helps ensure that on pane group close, we don't drop the diffstatemodel unless that was the only pane group referencing it 

## Testing
Repro steps: 
1. Open code review view in a repo in one tab
2. Open another code review view in the same repo in another tab
3. Close the first tab 
4. See that the code review view gets in a broken state (this repros on both local and remote sessions)
  
### Screenshots / Videos
https://www.loom.com/share/893359883e684175a71f70170c46523b

## Agent Mode

### DIFF
--- a/app/src/pane_group/working_directories.rs
+++ b/app/src/pane_group/working_directories.rs
@@ -78,6 +78,141 @@ impl DiffStateModelMap {
     }
 }
 
+/// Bidirectional map of pane groups to the repository roots they reference.
+///
+/// Maintains both a forward map (`pane_group_id -> ordered set of repo paths`)
+/// and a reverse map (`repo path -> set of pane group ids that reference it`)
+/// in lockstep, so callers can answer "is this repo still referenced by any
+/// pane group?" in O(1) without scanning every pane group's set.
+///
+/// All mutations go through methods on this wrapper to guarantee the two
+/// maps stay in sync.
+#[cfg(feature = "local_fs")]
+#[derive(Default)]
+struct PaneGroupRepositoryRoots {
+    /// Forward: per-pane-group ordered set of repository roots.
+    /// IndexSet maintains insertion order so most recently added repos appear later.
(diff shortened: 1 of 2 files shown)


app/src/pane_group/working_directories.rs
```diff
@@ -78,6 +78,141 @@ impl DiffStateModelMap {
...
+                set.remove(&pane_group_id);
+                set.is_empty()
+            });
+            if became_empty {
+                self.path_to_pane_groups.remove(&path);
+                orphans.push(path);
+            }
+        }
+        orphans
+    }
```
alokedesai: nit, not blocking: I think it would be clearer if you just nuked this function and added a specific remove function. It's awkward that we can directly call this function from anywhere in this module, it shouldn't really be part of the public API.

The function could be:

```rust
fn remove_path(&mut self, pane_group_id: EntityId, path: &LocalOrRemotePath) -> bool
```

where the return value means “removing this path from this pane group made the path globally orphaned.”

Then `set_paths` could compute the paths that left the pane group and do:

```rust
removed
    .into_iter()
    .filter(|path| self.remove_path(pane_group_id, path))
    .collect()
```

and `remove_pane_group` could use the same helper over all paths owned by the pane group.

That keeps the same bidirectional map and orphan detection behavior, but makes the abstraction a lot narrower: callers aren’t directly invoking “remove this pane group from the reverse map,” they’re just removing one tracked path from one pane group and getting back whether that path is now safe to clean up. It also makes it harder for future callers to accidentally use the reverse-map helper in a way that bypasses the forward-map.

alokedesai (approved):

PR #11791, opened 2026-05-27
Additional tab grouping actions

johnturcoo:
## Description
Adds tab grouping actions plumbing and two new tab right-click menu items behind the `GroupedTabs` feature flag
- New group with tab — creates a new `TabGroup`
- Remove from group — clears the tab's group_id and repositions it just past the previous group's last remaining member

## Changes

Action variants added to `WorkspaceAction`:

•  `NewTabGroupFromTab(usize)`
•  `MoveTabToGroup { tab_index, group_id }` (wired up but not yet exposed in UI)
•  `RemoveTabFromGroup(usize)`

## Linked Issue
https://linear.app/warpdotdev/issue/APP-4614/vertical-tab-grouping-additional-actions

## Testing

### Screenshots / Videos

[Demo video](https://www.loom.com/share/ef5118e223d64768b5f137a8d806780b)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -6708,6 +6708,136 @@ impl Workspace {
         }
     }
 
+    /// Creates a new group containing the tab and moves it to the top of
+    /// the tab list.
+    fn new_tab_group_from_tab(&mut self, tab_index: usize, ctx: &mut ViewContext<Self>) {
+        if !FeatureFlag::GroupedTabs.is_enabled() {
+            return;
+        }
+        let Some(tab) = self.tabs.get(tab_index) else {
+            return;
+        };
+        let previous_group_id = tab.group_id;
+
+        let group = TabGroup::new();
+        let group_id = group.id;
+        self.tab_groups.insert(group_id, group);
+
+        self.tabs[tab_index].group_id = Some(group_id);
+        self.move_tab_to_index(tab_index, 0, ctx);
+        self.activate_tab_internal(0, ctx);
+
+        if let Some(prev_group_id) = previous_group_id {
+            self.prune_empty_tab_group(prev_group_id, ctx);
+        }
+
+        ctx.dispatch_global_action("workspace:save_app", ());
+        ctx.notify();
+    }
+
+    /// Moves the tab into `group_id`, appending it to the end of the
+    /// group's contiguous run.
+    fn move_tab_to_group(
+        &mut self,
+        tab_index: usize,
+        group_id: TabGroupId,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        if !FeatureFlag::GroupedTabs.is_enabled() {
+            return;
+        }
+        let Some(tab) = self.tabs.get(tab_index) else {
+            return;
+        };
+        if tab.group_id == Some(group_id) {
+            return;
+        }
+        let previous_group_id = tab.group_id;
+
+        self.tabs[tab_index].group_id = Some(group_id);
+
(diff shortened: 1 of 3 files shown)


app/src/workspace/view.rs
```diff
@@ -6708,6 +6708,133 @@ impl Workspace {
...
+    /// Creates a new group containing the tab and moves it to the top of
+    /// the tab list.
+    fn new_tab_group_from_tab(&mut self, tab_index: usize, ctx: &mut ViewContext<Self>) {
+        if !FeatureFlag::GroupedTabs.is_enabled() {
+            return;
+        }
+        let Some(tab) = self.tabs.get(tab_index) else {
+            return;
+        };
+        let previous_group_id = tab.group_id;
```
vkodithala, line 6721: nit, non-blocking: Should we return early if the tab already belongs to a tab group with one member? Would be a no-op in that case yea?

app/src/workspace/view.rs
```diff
@@ -6708,6 +6708,133 @@ impl Workspace {
...
+    ) {
+        if !FeatureFlag::GroupedTabs.is_enabled() {
+            return;
+        }
+        let Some(tab) = self.tabs.get(tab_index) else {
+            return;
+        };
+        if tab.group_id == Some(group_id) {
+            return;
+        }
```
vkodithala, line 6757: These should basically never be reached, but maybe not the worst idea to add `log::debug` for the future?

app/src/workspace/view.rs
```diff
@@ -6708,6 +6708,133 @@ impl Workspace {
...
+
+        self.prune_empty_tab_group(previous_group_id, ctx);
+
+        ctx.dispatch_global_action("workspace:save_app", ());
+        ctx.notify();
+    }
+
+    /// Removes a tab group from the workspace if no tabs reference it.
+    fn prune_empty_tab_group(&mut self, group_id: TabGroupId, ctx: &mut ViewContext<Self>) {
+        let has_members = self.tabs.iter().any(|t| t.group_id == Some(group_id));
```
vkodithala: Can we not just reuse `group_member_indices` here, i.e. something like:
```
let has_members = !group_member_indices(&self.tabs, group_id).is_empty()
```

app/src/workspace/view.rs
```diff
@@ -6708,6 +6708,133 @@ impl Workspace {
...
+            ctx.notify();
+        }
+    }
+
+    /// Moves the tab at `from` to position `to` (`Vec::insert` semantics).
+    /// The active-tab tracker follows the moved tab.
+    fn move_tab_to_index(&mut self, from: usize, to: usize, ctx: &mut ViewContext<Self>) {
+        if from >= self.tabs.len() {
+            return;
+        }
```
vkodithala, line 6818: Ditto to my earlier point; probably best to add a `log::debug` or comment here to track this failure mode?

app/src/workspace/view.rs
```diff
@@ -6708,6 +6708,133 @@ impl Workspace {
...
+        let adjusted_to = if to > from { to - 1 } else { to };
+        let adjusted_to = adjusted_to.min(self.tabs.len().saturating_sub(1));
+        if from == adjusted_to {
+            return;
+        }
```
vkodithala, line 6824: nit: Wanna add a comment here describing what this does? Assuming this just no-ops when we're moving a tab to the same index, but probably best to declare (a little hard to grok)

vkodithala (approved):
This looks great, left a few comments (nothing blocking) :rocket:

PR #11799, opened 2026-05-28
Clarify create file palette item

zachlloyd:
## Description
Clarifies the command palette fallback item for creating a new file so it reads `Create a file named <file-name>…` instead of the more ambiguous `Create <file-name>…`.

This keeps the existing create-file behavior unchanged and only updates the user-facing label plus nearby code comment.

Agent conversation: https://staging.warp.dev/conversation/a701caac-682d-4561-9630-7450186a1d06

https://www.loom.com/share/ea05f1cd74fc4576a48a6b3f0d0604a2

## Linked Issue

## Testing
- [x] `cargo fmt --manifest-path /Users/zach/Projects/warp_3/Cargo.toml --all`
- [x] `cargo clippy --manifest-path /Users/zach/Projects/warp_3/Cargo.toml --workspace --all-targets --all-features --tests -- -D warnings`

### Screenshots / Videos
Not attached; this is a text-only command palette label update.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/src/search/command_palette/files/search_item.rs
+++ b/app/src/search/command_palette/files/search_item.rs
@@ -161,7 +161,7 @@ impl SearchItem for CreateFileSearchItem {
         let text_color = highlight_state.sub_text_fill(appearance).into_solid();
 
         let label = Text::new_inline(
-            format!("Create {}…", &self.file_name),
+            format!("Create a file named {}…", &self.file_name),
             appearance.ui_font_family(),
             appearance.monospace_font_size(),
         )
--- a/app/src/search/command_palette/files/data_source.rs
+++ b/app/src/search/command_palette/files/data_source.rs
@@ -286,7 +286,7 @@ impl FileDataSource {
                 .collect();
 
             // If no files matched and we have a valid query and current directory,
-            // add a "Create <filename>..." option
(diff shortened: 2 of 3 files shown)


app/src/search/command_palette/files/search_item.rs
```diff
@@ -161,7 +161,7 @@ impl SearchItem for CreateFileSearchItem {
         let text_color = highlight_state.sub_text_fill(appearance).into_solid();
 
         let label = Text::new_inline(
-            format!("Create {}…", &self.file_name),
+            format!("Create a file named {}…", &self.file_name),
```
alokedesai, line 164: Does it even make sense to show this as the fallback unless you have the `files:` filter applied?

Seems really odd that the default "no results found" UX in the command palette is to suggest creating a file.

alokedesai (approved):

app/src/search/command_palette/files/search_item.rs
zachlloyd: [Warp Agent] Updated this so the create-file fallback is only added when the active query filter is files:. Unfiltered command-palette searches will now fall back to the normal "No results found" state instead of suggesting file creation.

PR #11838, opened 2026-05-28
GPT models 1M context

dagmfactory:
## Description

- Enabling 1M Context for [supported](https://developers.openai.com/api/docs/pricing) GPT models (`gpt-5.4`, `gpt-5.5`, `grape`)
- OpenAI specific long-context pricing warning when context window `>272K` (OpenAI threshold)
- Server as source of truth for which models support configurable context windows
- change is behind `GPTConfigurableContextWindow` flag
- removing stale flag `configurable_context_window`

## Testing

#### Default 

Screenshot 2026-05-28 at 3 16 36 PM

#### Warning 

Screenshot 2026-05-28 at 3 17 50 PM

## Agent Mode


app/src/ai/agent/api/impl.rs
```diff
@@ -68,13 +68,7 @@ pub async fn generate_multi_agent_output(
                 base: params.model.into(),
                 cli_agent: params.cli_agent_model.into(),
                 computer_use_agent: params.computer_use_model.into(),
-                base_model_context_window_limit: if FeatureFlag::ConfigurableContextWindow
```
dagmfactory, line 71: @coolcom200 same question here. 
Is it okay to remove this feature flag now?

coolcom200: Yes should be safe to remove.

danielpeng2 (commented):
FYI the screenshots in the description aren't loading.

I tried using long context with your server change and I keep getting errors?
Image

Also a product nit: can we make the warning in the settings appear as you drag, not only after you release the slider?

crates/warp_features/src/lib.rs
```diff
@@ -876,6 +871,8 @@ pub enum FeatureFlag {
 
     /// Gates the v2 billing and usage page redesign.
     BillingAndUsagePageV2,
+    /// Enables configurable expanded context windows for eligible GPT models.
+    GPTConfigurableContextWindow,
```
danielpeng2, line 875: ooc do we need a client-side feature flag for this? the server already drives whether long context for GPT models is enabled -- it feels sufficient to let the server handle that decision

dagmfactory: I think the server should be the source of truth for which models are and are not configurable. We can add a flag on the client side just as a control for rollout and clean it up later though.

dagmfactory:
> can we make the warning in the settings appear as you drag, not only after you release the slider

Can you send me the conversation ID? I can take a look. It's working on my end with the server changes https://www.loom.com/share/491850b7038241519ff7a45612af5e5c

app/src/ai/agent/api.rs
```diff
@@ -303,13 +303,7 @@ impl RequestParams {
                 .active_profile(terminal_view_id, app)
                 .data()
                 .clone();
```
danielpeng2: do we still need the clone here if we're going to call `context_window_limit_for_request` to get the limit directly? feels like we could simplify this part of the code now

app/src/ai/execution_profiles/mod.rs
```diff
@@ -118,6 +143,61 @@ impl AIExecutionProfileAppExt for AIExecutionProfile {
...
+                context_window_limit
+                    .or(self.context_window_limit)
+                    .unwrap_or(llm.context_window.default_max),
+            ),
+            FeatureFlag::GPTConfigurableContextWindow.is_enabled(),
+        )
+    }
+}
+
+pub(crate) fn sanitize_context_window_limit_for_request(
```
danielpeng2: this method seems like it's only used in `context_window_limit_for_request`? I have a slight preference for inlining the method, feels like it'd be a bit easier to read.

app/src/view_components/warning_box.rs
```diff
@@ -78,8 +81,12 @@ impl WarningBoxConfig {
         self.button = Some(button);
         self
     }
-}
 
+    pub fn with_formatted_title(mut self, formatted_title: FormattedTextInline) -> Self {
+        self.formatted_title = Some(formatted_title);
+        self
+    }
```
danielpeng2: the API is a bit awkward here since to construct a `WarningBoxConfig` with a formatted title, you have to go through the `new` constructor and pass in a non-formatted title that will get overwritten.

we have a couple other UI components where the label can be either text or some custom element. a good example of this is `RadioButtonItem`. it's a bit cleaner since we can use an enum to encode the "text or custom element" constraint in code.

app/src/ai/execution_profiles/mod.rs
```diff
@@ -118,6 +143,61 @@ impl AIExecutionProfileAppExt for AIExecutionProfile {
...
+    selected_limit: Option<u32>,
+    gpt_configurable_context_window_enabled: bool,
+) -> Option<u32> {
+    if !has_effective_configurable_context_window(llm, gpt_configurable_context_window_enabled) {
+        return None;
+    }
+
+    selected_limit.map(|limit| limit.clamp(llm.context_window.min, llm.context_window.max))
+}
+pub(crate) fn has_effective_configurable_context_window(
```
danielpeng2: nit: what does it mean semantically to have an "effective" configurable context window. could we name the method `has_effective_configurable_context_window`?

app/src/ai/execution_profiles/mod.rs
```diff
@@ -15,6 +16,19 @@ use crate::cloud_object::{
 use crate::server::sync_queue::QueueItem;
 use crate::settings::AISettings;
 use crate::workspaces::user_workspaces::UserWorkspaces;
+pub const LONG_CONTEXT_WARNING_THRESHOLD: u32 = 272_000;
```
danielpeng2, line 20: we can do this in another PR, but I think we should pass the threshold value down from the server. we might need to modify the graphql/protos to pass that info alongside the rest of the context window values.

having this be server driven makes it easier to support new models with different thresholds in the future. imagine OpenAI releases a new model with a different threshold value. we want to avoid needing an out-of-band client release just to support the new model.

can we specify in a comment that this threshold is only relevant for GPT 5.4 + 5.5 for now, and wire the threshold value down from the server in a follow-up PR?

crates/warp_features/src/lib.rs
danielpeng2: we can already control the rollout via the server, and avoiding the client flag seems like it would make some of the client code simpler.

not a huge deal though, up to you!

danielpeng2:
> > can we make the warning in the settings appear as you drag, not only after you release the slider
> 
> Can you send me the conversation ID? I can take a look. It's working on my end with the server changes https://www.loom.com/share/491850b7038241519ff7a45612af5e5c

I tried again on your updated server branch, and it worked fine. Might've had something weird in my local setup, ignore that error!

app/src/ai/execution_profiles/mod.rs
dagmfactory: I'm working on a pr that updates the protos to send `long_context_used` bool from server to client on stream finish, I'll include this in that pr so that the server becomes the only  source of truth for thresholds.
For now, added a comment here that `LONG_CONTEXT_WARNING_THRESHOLD` is applicable only for GPT 5.4, 5.5 and grape

danielpeng2 (approved):

crates/warp_features/src/lib.rs
```diff
@@ -952,6 +949,8 @@ pub const DOGFOOD_FLAGS: &[FeatureFlag] = &[
     FeatureFlag::RemoteCodebaseIndexing,
     FeatureFlag::GroupedTabs,
     FeatureFlag::AsyncFind,
+    FeatureFlag::OrchestrationViewerStreamer,
```
danielpeng2, line 952: Oops, I missed this earlier, but it seems like a mistake to add this flag in this change?

PR #11840, opened 2026-05-28
feat: respect feedback skill organization setting

captainsafia:
## Summary
- Add `feedbackSkillSettings` to workspace GraphQL metadata and local workspace settings.
- Gate only Warp's bundled `feedback` skill through the workspace/org setting plus the existing user preference.
- Preserve the no-workspace fallback to the user setting and leave user/project skills named `feedback` unaffected.
- Add targeted tests for the default-disabled org behavior and bundled-skill activation/read-skill enforcement.

## Related
- Server PR: https://github.com/warpdotdev/warp-server/pull/11506

## Testing
- `cargo fmt --all --manifest-path /workspace/warp/Cargo.toml`
- `cargo fmt --all --manifest-path /workspace/warp/Cargo.toml -- --check`
- `CARGO_BUILD_JOBS=1 cargo test --manifest-path /workspace/warp/Cargo.toml -p warp test_feedback_bundled_skill -- --nocapture`
- `CARGO_BUILD_JOBS=1 cargo test --manifest-path /workspace/warp/Cargo.toml -p warp bundled_feedback_skill -- --nocapture`
- `CARGO_BUILD_JOBS=1 cargo test --manifest-path /workspace/warp/Cargo.toml -p warp disabling_feedback_bundled_skill -- --nocapture`
- `CARGO_BUILD_JOBS=1 cargo test --manifest-path /workspace/warp/Cargo.toml -p warp test_read_skill_executor -- --nocapture`

_Conversation: https://staging.warp.dev/conversation/89cb8a8e-12b8-4645-957e-b9810d70dcd0_
_Run: https://oz.staging.warp.dev/runs/019e6f23-21c3-7ba2-887c-dccf04499003_
_Plans_:
- _[Org-level built-in feedback skill disablement](https://staging.warp.dev/drive/notebook/XEC6JhQw85MdIYOywOZVWc)_

### DIFF
--- a/app/src/workspaces/user_workspaces.rs
+++ b/app/src/workspaces/user_workspaces.rs
@@ -1426,6 +1426,25 @@ impl UserWorkspaces {
             .unwrap_or_default()
     }
 
+    pub fn is_feedback_bundled_skill_enabled(&self, app: &AppContext) -> bool {
+        match self
+            .current_team()
+            .map(|team| {
+                team.organization_settings
+                    .feedback_skill_settings
+                    .setting
+                    .clone()
+            })
+            .unwrap_or(AdminEnablementSetting::RespectUserSetting)
+        {
+            AdminEnablementSetting::Disable => false,
+            AdminEnablementSetting::Enable => true,
+            AdminEnablementSetting::RespectUserSetting => {
+                *AISettings::as_ref(app).feedback_bundled_skill_enabled
+            }
+        }
+    }
+
     pub fn is_ai_allowed_in_remote_sessions(&self) -> bool {
         self.current_team()
             .map(|team| {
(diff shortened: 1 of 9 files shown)


app/src/workspaces/user_workspaces.rs
```diff
@@ -1426,6 +1426,25 @@ impl UserWorkspaces {
...
+    pub fn is_feedback_bundled_skill_enabled(&self, app: &AppContext) -> bool {
+        match self
+            .current_team()
+            .map(|team| {
+                team.organization_settings
+                    .feedback_skill_settings
+                    .setting
+                    .clone()
+            })
+            .unwrap_or(AdminEnablementSetting::RespectUserSetting)
```
peicodes, line 1438: I think this needs to be default disable to achieve our goal of preventing accidental usage

peicodes: And to confirm - this is a completely new setting correct? So the first time this runs after ship, it will use this default?

peicodes (commented):
I think we have to make the default fallback for enablement `Disabled` before we merge

app/src/workspaces/user_workspaces.rs
captainsafia: It's disabled in the default that is configured on the server side via the enterprise setting. I believe if we disabled it here, we run the risk of disabling it for people _without_ an enterprise team which would break back-compat with the existing experience for users in general.

PR #11841, opened 2026-05-28
Improve EditFiles diff match failure errors

vkodithala:
## Description

When `edit_files` fails to locate one or more search blocks, the agent previously received a generic `Could not apply all diffs to <file>.` with no indication of *which* block failed or *where* it was expected, so retries were largely guesswork. This enriches the failure into an actionable, self-correcting signal for both the search/replace and V4A (`apply_patch`) diff formats.

### How it works
- Diff matching now records a structured `DiffMatchFailure` for every unmatched block, capturing the block's search text and — when the block carried line numbers — the 1-indexed range it was expected to match. These are collected on `DiffMatchFailures` (now `Clone` instead of `Copy`) alongside the existing failure counts.
- The conversation error (`DiffApplicationError::to_conversation_message`) now enumerates each failed block: `Expected line N.` / `Expected lines N-M.` when the range is known, followed by the verbatim `Search:` content, prefixed with an instruction to make each block match the file exactly and retry. A co-occurring no-op ("changes already made") still renders alongside on its own line.
- V4A hunks don't carry line numbers, so the search excerpt is reconstructed from the hunk's combined pre-context, removed, and post-context lines (`v4a_hunk_search_text`), giving the model back the same context it originally sent.

### Bounding and sensitive-content handling
- Each excerpt is capped at `MAX_DIFF_MATCH_FAILURE_BYTES` (1 KB) via `capped_failure_text`, truncated on a UTF-8 char boundary with a `... [truncated]` marker, so a single oversized block can't crowd out other failed hunks or bloat the retry message.
- Search snippets are user content, so they are kept out of any remotely-collected surface: the detail vector is `#[serde(skip)]`ped out of telemetry payloads, and the `safe_warn!` release-log path now logs only the failure counts (unmatched / noop / missing-line-numbers). The full snippets remain only in the local-only (`full:`) log arm.

Context here: https://warpdev.slack.com/archives/C09769R5GBT/p1778528642584199?thread_ts=1778528430.538039&cid=C09769R5GBT.

## Linked Issue

N/A

## Testing

Quick walkthrough of what this looks like in practice: https://www.loom.com/share/d8cbf155bf884cbdb30019665dc40dce.

- [x] `cargo fmt`
- [x] `cargo test -p ai diff_validation`
- [x] `cargo test -p warp test_format_match_error`

## Agent Mode

Conversation: https://staging.warp.dev/conversation/93cc055d-2e87-4c89-a84c-a468c5b10c43

CHANGELOG-NONE


crates/ai/src/diff_validation/mod.rs
```diff
@@ -323,14 +324,69 @@ fn remove_extra_line_num_prefix(replace: String) -> String {
...
 pub struct DiffMatchFailures {
     /// Failures to perform a fuzzy match with content.
     pub fuzzy_match_failures: u8,
     /// The <search> and <replace> content was identical.
     pub noop_deltas: u8,
     /// Search blocks that are missing line numbers.
     pub missing_line_numbers: u8,
+    /// The search/replace payloads that failed to fuzzy match. Skipped for telemetry.
+    #[serde(skip)]
+    pub fuzzy_match_failure_details: Vec<DiffMatchFailure>,
```
vkodithala, line 342: I considered making this an independent struct to avoid coupling with telemetry (which necessitates the #[serde(skip)] here) but wanted to avoid extra plumbing. Let me know if you have any thoughts here; happy to restructure as needed.

vkodithala:
* **#11841** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11841?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/11841?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/ai/src/diff_validation/mod.rs
```diff
@@ -323,14 +324,52 @@ fn remove_extra_line_num_prefix(replace: String) -> String {
...
 pub struct DiffMatchFailures {
     /// Failures to perform a fuzzy match with content.
     pub fuzzy_match_failures: u8,
     /// The <search> and <replace> content was identical.
     pub noop_deltas: u8,
     /// Search blocks that are missing line numbers.
     pub missing_line_numbers: u8,
+    /// The search blocks that failed to fuzzy match. Skipped for telemetry.
+    #[serde(skip)]
+    pub fuzzy_match_failure_details: Vec<DiffMatchFailure>,
```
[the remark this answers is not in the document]
vkodithala: This is valid; should be resolved in latest!

szgupta (commented):
@vkodithala can we make the error here a structured enum type at the API level, so that we can control formatting of the error message (e.g. exact verbiage) as we please at the harness (server) level?

vkodithala:
@szgupta Took a look. Agree with this direction high-level, but is this in-scope for these changes? This seems like a longstanding proto-level issue, see [here](https://github.com/warpdotdev/warp-proto-apis/blob/varoon/lrc-add-preempt/apis/multi_agent/v1/task.proto#L1149) where `ApplyFileDiffsResult::Error` is defined as a plain string vs. enum. Would require changes on client, server, and protos that I'm very happy to get in, but would likely be best on an upstream PR. Wdyt?

app/src/ai/blocklist/action_model/execute/request_file_edits/diff_application.rs
```diff
@@ -648,12 +688,17 @@ async fn apply_search_replace<F, Fut>(
                 && let Some(failures) = fuzzy_match_diffs.failures.as_ref()
             {
                 safe_warn!(
-                    safe: ("Failure(s) applying diff: {failures:?}"),
+                    safe: (
+                        "Failure(s) applying diff: {} unmatched, {} noop, {} missing line numbers",
+                        failures.fuzzy_match_failures,
+                        failures.noop_deltas,
+                        failures.missing_line_numbers
+                    ),
```
vkodithala, line 696: To confirm my understanding, is the idea here that we're changing the signature from `{failures:?}` to avoid surfacing fuzzy match failure details in telemetry? Presumably because this is UGC and not something we'd want to log?

crates/ai/src/diff_validation/mod.rs
```diff
@@ -323,14 +324,52 @@ fn remove_extra_line_num_prefix(replace: String) -> String {
         .join("\n")
 }
 
-#[derive(Debug, Default, Clone, Copy, PartialEq, Serialize)]
+const MAX_DIFF_MATCH_FAILURE_BYTES: usize = 1_000;
```
vkodithala, line 330: We should add a comment explaining why limiting this to 1k bytes is a reasonable decision. IIUC we do this to avoid pushing an unnecessarily long search/replace query as an LLM response (which could corrupt/bloat the context window), so we limit each failure's surfaced identifier/search query here.

vkodithala (commented):
Few comments that I'd like some clarification on. None blocking.

crates/ai/src/diff_validation/mod.rs
```diff
@@ -323,14 +324,55 @@ fn remove_extra_line_num_prefix(replace: String) -> String {
...
     /// Search blocks that are missing line numbers.
     pub missing_line_numbers: u8,
+    /// The search blocks that failed to fuzzy match. Skipped for telemetry.
+    #[serde(skip)]
+    pub fuzzy_match_failure_details: Vec<DiffMatchFailure>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct DiffMatchFailure {
+    pub search: String,
```
szgupta, line 347: What is `search` here? Is this the error message?

szgupta: Oh I see now this is the same search block that failed. Do we need this? The model will already know what the search block was from the tool call

app/src/ai/blocklist/action_model/execute/request_file_edits/diff_application.rs
```diff
@@ -151,6 +157,40 @@ impl DiffApplicationError {
     }
 }
 
+fn append_fuzzy_match_failure_details(message: &mut String, match_failures: &DiffMatchFailures) {
+    if match_failures.fuzzy_match_failure_details.is_empty() {
+        return;
+    }
+
+    message.push_str(
+        " Update each failed search block to match the file exactly, then retry. The following search blocks could not be matched:",
```
szgupta, line 166: Hrm - this feels like a prompt. I think it should just be the specific error

szgupta:
@vkodithala can you attach a before and after screenshot of how the tool call result is formatted in the LLM request?

crates/ai/src/diff_validation/mod.rs
vkodithala: (authored by an agent after some back/forth): agreed the full echo was redundant since the model already has the original tool call. Implemented in 108f4b5b9: `DiffMatchFailure` no longer carries the search text at all, and the LLM-facing error now identifies the failed block only by its original 1-based index within that tool call (e.g. "Search block 2."), so it can be reconciled against the model's own request without repeating any content.

app/src/ai/blocklist/action_model/execute/request_file_edits/diff_application.rs
vkodithala: (authored by an agent after some back/forth): agreed. Replaced the imperative retry sentence with factual wording in a9d83ba13: the lead-in is now "The following search blocks did not match the current file contents:" followed by the specific failed block identifier(s), with no instructions on how to fix it. A capable model can infer the remediation from the specific diagnostic alone.

vkodithala:
(authored by an agent after some back/forth): @szgupta Following up on your before/after screenshot request (https://github.com/warpdotdev/warp/pull/11841#issuecomment-5133237873) — here's a Loom walking through the tool-result formatting before and after this change: https://www.loom.com/share/1bfdc52c180e44669d8f2f5af5504a35

szgupta (approved):

PR #11842, opened 2026-05-28
Support tab group renaming

johnturcoo:
## Description

Adds inline rename support for tab groups in the vertical tabs panel.

- Double-click a group header to edit its name
- Newly-created groups (`Create new group`, `New group from tab`) auto-open the rename editor so users can name them immediately)
- Empty input is treated as cancel — the existing name is preserved

## Changes
- **`workspace/action.rs`** — New `WorkspaceAction::RenameTabGroup(TabGroupId)`
- **`workspace/util.rs`** — Added `tab_group_being_renamed: Option<TabGroupId>` to `WorkspaceState` with `set/clear/is_any/is/get` accessors. 
- **`workspace/view.rs`**
  - New `tab_group_rename_editor` editor view 
  - `finish_tab_group_rename` / `cancel_tab_group_rename` commit/abort the rename, restoring focus to the active tab.
  - `rename_tab_group` opens the inline editor, seeded with the existing name or "New Group" with the text pre-selected.
  - `create_new_tab_group` and `new_tab_group_from_tab` now dispatches `RenameTabGroup` via `dispatch_typed_action_deferred` so the group's header mounts in the next render before focus is requested.
- **`workspace/view/vertical_tabs.rs`** — Group header conditionally renders the rename editor when `is_tab_group_being_renamed(group.id)` is true; double-click on the header dispatches `RenameTabGroup`.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4618/vertical-tab-group-renaming-support

## Testing

### Screenshots / Videos

[Demo video](https://www.loom.com/share/23a0c704896f4dd799e90884a31e86d6)

### DIFF
--- a/app/src/workspace/view/vertical_tabs.rs
+++ b/app/src/workspace/view/vertical_tabs.rs
@@ -2430,7 +2430,9 @@ fn render_tab_group_header_icon_button(
 }
 
 /// Renders the header row for a tab group: chevron, title + "N tabs", and (on hover)
-/// kebab + close buttons. Single-clicking outside the per-button regions toggles collapse.
+/// kebab + close buttons. Single-clicking outside the per-button regions toggles collapse;
+/// double-clicking opens the inline rename editor.
+#[allow(clippy::too_many_arguments)]
 fn render_grouped_tabs_header(
     group: &TabGroup,
     member_count: usize,
@@ -2438,6 +2440,8 @@ fn render_grouped_tabs_header(
     is_collapsed: bool,
     is_header_selected: bool,
     show_action_buttons: bool,
+    is_being_renamed: bool,
+    rename_editor: Option<&ViewHandle<EditorView>>,
     app: &AppContext,
 ) -> Box<dyn Element> {
     let appearance = Appearance::as_ref(app);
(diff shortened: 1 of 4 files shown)


app/src/workspace/view/vertical_tabs.rs
```diff
@@ -2565,11 +2574,17 @@ fn render_grouped_tabs_header(
...
-    hoverable = hoverable.on_click(move |ctx, _, _| {
-        ctx.dispatch_typed_action(WorkspaceAction::ToggleTabGroupCollapsed(group_id));
-    });
+    // Single-click toggles collapse; double-click opens the inline rename editor.
+    // No-op `on_right_click` keeps right-clicks from bubbling up to the panel's
+    // new-session menu handler.
+    if !is_being_renamed {
+        hoverable = hoverable.on_click(move |ctx, _, _| {
+            ctx.dispatch_typed_action(WorkspaceAction::ToggleTabGroupCollapsed(group_id));
+        });
```
vkodithala: Should we be gating the single-click action on whether the tab's being renamed? i.e. I'd still expect single-click to collapse the tab group even when being renamed - wydt?

vkodithala (approved):
Changes LG, left a comment. Good to :ship:

app/src/workspace/view/vertical_tabs.rs
johnturcoo: I agree with you here, good call out

PR #11846, opened 2026-05-28
Remove transfer conversation semantics from automatic streams

seemeroland:
## Description

Reintroduce the fix from https://github.com/warpdotdev/warp/pull/10241, since https://github.com/warpdotdev/warp/pull/11494 regressed it. The expectation is that the conversation should only be in one terminal view at a time, and setting active conversation id in the terminal view where it's already active should do nothing. But apparently that's not what's happening.

It's not clear to me why this path is being hit in a different terminal view, stealing the conversation from where it's in progress. But apparently it fixed the issue, so add it back.

### DIFF
--- a/app/src/ai/blocklist/history_model.rs
+++ b/app/src/ai/blocklist/history_model.rs
@@ -928,6 +928,7 @@ impl BlocklistAIHistoryModel {
             terminal_view_id,
         });
     }
+
     /// Marks a conversation as the active conversation for a terminal view
     /// without removing it from other views.
     ///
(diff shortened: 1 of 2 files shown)


app/src/ai/blocklist/history_model.rs
```diff
@@ -925,6 +928,39 @@ impl BlocklistAIHistoryModel {
             terminal_view_id,
         });
     }
+    /// Marks a conversation as the active conversation for a terminal view
```
szgupta, line 932: nit: missing newline above

szgupta (approved):

PR #11849, opened 2026-05-28
Add move to group sidecar menu

johnturcoo:
## Description
Adds a Move to group sidecar menu to the tab right-click context menu. When the user hovers Move to group, a panel opens beside the parent menu listing every other tab group (ordered by visual position in the tabs panel). Selecting a group dispatches WorkspaceAction::MoveTabToGroup, moving the tab into that group and dismissing both menus. Mirrors the existing new-session worktree sidecar pattern.

Gated by FeatureFlag::GroupedTabs — when disabled, the entire tab-group section of the right-click menu (including this item) is hidden.

## Changes

- `app/src/tab.rs`: Added `MOVE_TO_GROUP_LABEL` constant and threaded `&HashMap<TabGroupId, TabGroup>` through the tab menu builders so `tab_group_menu_items` can insert the submenu-parent item when other groups exist.
- `app/src/workspace/view.rs`: Added the sidecar menu view, and new helpers (`build_move_to_group_sidecar_items`, `update_move_to_group_sidecar`, `hide_move_to_group_sidecar`, `handle_move_to_group_sidecar_event`) that wire parent hover events into open/hide logic and render the sidecar as an overflow-aware positioned overlay.

## Linked Issue
https://linear.app/warpdotdev/issue/APP-4620/render-sidecar-menu-for-tab-grouping-options

## Testing

### Screenshots / Videos

[Demo video](https://www.loom.com/share/156a461564164f938a5e32206bc104b7)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -6974,6 +6974,7 @@ impl Workspace {
     ) {
         if self.show_tab_right_click_menu.is_some() {
             self.show_tab_right_click_menu = None;
+            self.hide_move_to_group_sidecar(ctx);
             ctx.notify();
             return;
         }
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -8994,7 +8994,9 @@ impl Workspace {
         };
         let current_group_id = tab.group_id;
 
-        let mut groups_with_first_index: Vec<(TabGroupId, usize)> = self
+        // Other groups paired with their first member's tab index, sorted so the menu
+        // matches panel order.
+        let sorted_other_groups = self
             .tab_groups
             .keys()
             .copied()
@@ -9004,11 +9006,9 @@ impl Workspace {
                     .next()
                     .map(|idx| (gid, idx))
             })
-            .collect();
-        groups_with_first_index.sort_by_key(|(_, idx)| *idx);
+            .sorted_by_key(|(_, idx)| *idx);
 
-        groups_with_first_index
-            .into_iter()
+        sorted_other_groups
             .map(|(group_id, _)| {
                 let label = self
                     .tab_groups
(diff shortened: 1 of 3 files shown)


app/src/workspace/view.rs
```diff
@@ -1924,7 +1942,26 @@ impl Workspace {
...
+        });
+        ctx.subscribe_to_view(&move_to_group_sidecar, move |me, _, event, ctx| {
+            me.handle_move_to_group_sidecar_event(event, ctx);
+        });
+
+        (
+            tab_right_click_menu,
+            new_session_menu,
+            new_session_sidecar,
+            move_to_group_sidecar,
```
peicodes, line 1963: nit and no need to address rn, but for future reference: I'm not a fan of unnamed tuples once they get to 3+ because it makes consumption difficult. Usually once they get to 3 I'll start using a lightweight struct with named fields. For this reason when I write code from scratch I avoid tuples altogether

app/src/workspace/view.rs
```diff
@@ -8980,9 +9027,145 @@ impl Workspace {
...
+    fn build_move_to_group_sidecar_items(
+        &self,
+        tab_index: usize,
+    ) -> Vec<MenuItem<WorkspaceAction>> {
+        let Some(tab) = self.tabs.get(tab_index) else {
+            return vec![];
+        };
+        let current_group_id = tab.group_id;
+
+        let mut groups_with_first_index: Vec<(TabGroupId, usize)> = self
```
peicodes: nit: This entire series of mapping and filtering seems to be just a series of declarative loops. We can definitely do it without a `mut`. And maybe break it up with intermediate variables?

app/src/workspace/view.rs
```diff
@@ -24195,6 +24378,46 @@ impl View for Workspace {
                         positioning,
                     );
                 }
+
+                // Sidecar menu for the "Move to group" submenu parent. Mirrors
+                // the new-session sidecar's overflow-aware left/right anchoring.
+                if self.show_move_to_group_sidecar {
```
peicodes, line 24313: I assume no need for ff here because `show_move_to_group_sidecar` won't be set yeah?

peicodes (approved):
General nit: I would love it if we start breaking up this file when making new changes

You can have multiple definitions of a class and Rust will basically combine them all. This allows you to define render functions and helpers in a separate file but still a part of the class

PR #11856, opened 2026-05-28
Fix symlinked gitignored paths in code review

MaggieShan:
## Description
* Context: https://warpdotdev.sentry.io/issues/7511484182/events/18c59c5b820d4bd7af362160ad91938b/?project=5658526 
* We're getting a group of sentry errors related to git commands failing due to being outside the repo (eg. `Git command failed: fatal: <path>/node_modules/<path>' is outside repository`) 
* ~~As per oz's investigation https://staging.warp.dev/conversation/e7f00920-692d-4314-86e3-adcb65c11ff9~~ 
  * ~~The `DirectoryWatcher`'s file watcher is registering gitignored directories like `node_modules` which follows its symlink entries and ends up registering paths outside the active repo => this means that we'd receive file watcher events for these file changes and fail to run commands like `git diff -- <file>`~~
* For symlinked paths like those in `node_modules` we weren't correctly attributing the file paths to their correct repositories which resulted in an inaccurate `is_ignored` value
  * This resulted in files that should've been ignored being queued for invalidation and failing with "is outside repository" 
* The fix includes updating how we derive `is_ignored` such that we're checking for the non-canonicalized paths which don't follow the symlinks 

## Linked Issue
[APP-4632](https://linear.app/warpdotdev/issue/APP-4632/git-command-failed-outside-repository)

## Testing
- Reproed error log + confirmed it no longer shows with fix

## Agent Mode


crates/repo_metadata/src/entry.rs
```diff
@@ -600,23 +600,65 @@ fn descend_allowlist_matches(suffix: &[Component<'_>]) -> bool {
...
+#[cfg(feature = "local_fs")]
+pub(crate) fn should_descend_into_directory(path: &Path, gitignores: &[Gitignore]) -> bool {
+    if !should_watch_directory_in_git_path(path) {
+        return false;
+    }
+    // Skip gitignored directories so we never register watches inside them.
+    // Directories are always passed `is_dir=true` here
+    // because the watcher only invokes this predicate for directories
+    // it is about to descend into.
+    !matches_gitignores(
```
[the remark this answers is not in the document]
MaggieShan: Fixed

crates/repo_metadata/src/entry.rs
```diff
@@ -703,23 +703,70 @@ fn descend_allowlist_matches(suffix: &[Component<'_>]) -> bool {
...
+    // The `.git/` allowlist is authoritative for git-internal paths: gitignore
+    // patterns must never prune the allowlisted `.git/` subtrees this watcher relies on,
+    // so the gitignore check below applies only to working-tree paths.
+    if is_git_internal_path(path) {
+        return should_watch_directory_in_git_path(path);
+    }
+    // Skip gitignored directories so we never register watches inside them.
+    // Directories are always passed `is_dir=true` here because the watcher only invokes this predicate for directories
+    // it is about to descend into.
+    !matches_gitignores(
```
kevinyang372: This doesn't seem fully right? I think we chatted before and an edge case here is if A/B is gitignored but not it's children A/B/C, we still need to descend into A/B (otherwise we will miss event from A/B/C)

crates/repo_metadata/src/entry.rs
```diff
@@ -703,23 +703,70 @@ fn descend_allowlist_matches(suffix: &[Component<'_>]) -> bool {
...
-pub fn repo_watch_filter() -> WatchFilter {
-    WatchFilter::with_filter(
-        Arc::new(should_watch_directory_in_git_path),
-        Arc::new(|path: &Path| !should_ignore_git_path(path)),
-    )
+pub fn repo_watch_filter(gitignores: Arc<Vec<Gitignore>>) -> WatchFilter {
+    let gitignores_for_descend = gitignores.clone();
+    let descend =
+        Arc::new(move |path: &Path| should_watch_directory(path, &gitignores_for_descend));
+    let emit = Arc::new(move |path: &Path| should_emit_event_for_path(path, &gitignores));
```
kevinyang372: I think we still need events for gitignored paths in repo metadata. This is needed for gitignored paths entries in the file tree updates right?

MaggieShan: Yes this is called out explicitly in https://github.com/warpdotdev/warp/pull/11856/changes/BASE..64fb4b7ccc8364dc7ea1b40931ed35db6107383b#diff-fda5fe6a7d99f2af27a511fc9ddf6aa71c3f4167b321ca4b7988bbe9cba0c027R435 where we pass an empty vec for the repometadatamodel whose subscribers like the file tree need events for gitignored paths

MaggieShan: Alternatively we can define a separate watch filter for repometadata vs. directorywatcher

crates/repo_metadata/src/entry.rs
MaggieShan: HMm that's a good point - in that case I'm thinking through 2 different approaches: 
1. We only update the emit filter to filter paths in gitignores 
2. Instead of updating the watch filter logic, we update specifically the filter logic in the localdiffstatemodel

crates/repo_metadata/src/entry.rs
kevinyang372: Did some exploration on RepositoryUpdate events in repo-metadata and a lot of them need to be able to consume gitignored data
https://staging.warp.dev/conversation/6691223f-6483-44aa-8ff9-ffa3f1116dd0

kevinyang372 (approved):

PR #11870, opened 2026-05-29
Fix internal-dev image by making sure we set the right rust default version to match the cargo version we install.

liliwilson:
## Description

Lately our `internal-dev` image (base for the Warp Dev Environment) has been turning up some "cargo clippy not installed" errors.

This is happening because we install the default version of `rustup`, but pin the cargo version that we install to the one defined in `warp/rust-toolchain.toml`. What this means is that whenever we try to run `cargo` from outside the `warp` repo, we don't have the right `cargo` version installed for that default version of `rustup` -- right now, the default rust version is 1.95.0, while we pin `cargo` to 1.92.0. 

This PR addresses this by setting the default rust toolchain version used in this docker image to match whatever we pin in `rust-toolchain.toml`, so it's consistent across the board.

## Testing

Built an image locally before and after this change, and confirmed that `cargo clippy --version` on the former fails, while on the latter works.

## Agent Mode

### DIFF
--- a/docker/agent-dev/Dockerfile
+++ b/docker/agent-dev/Dockerfile
@@ -22,7 +22,7 @@ RUN set -eux; \
     ca-certificates \
     xvfb
 
-# Install rustup. It will download the pinned version in our rust-toolchain.toml as needed.
+# Install rustup. The pinned Rust toolchain is installed below from rust-toolchain.toml.
 RUN set -eux; \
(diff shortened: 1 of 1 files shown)


bnavetta (approved):

captainsafia (approved):
Good find -- thank you for fixing!

PR #11871, opened 2026-05-29
Add support for a proper codex plugin.

liliwilson:
## Description

This PR adds support for a proper [Codex plugin](https://github.com/warpdotdev/codex-warp) to back notifications and third-party harness support in Warp, so we no longer need to rely on OSC 9 notifications or installing hooks via a sidecar image for cloud agents.

We gate all of this behind a feature flag, `CodexPlugin`, so we can make sure the plugin works in dogfood before enabling for everyone.

We also make sure that still fall back to OSC 9 notifications, but store a new field on the `CodexSessionHandler`, `structured_plugin_active`, so that when we receive our first OSC 777 and see we have an active plugin, we ignore all of the less-rich OSC 9 notifications that come afterwards.

This PR does not handle automatically installing the plugin for orchestration/cloud agent use cases. That will need to be done in a future PR.

## Testing

### Screenshots / Videos

Loom: https://www.loom.com/share/62f87191f6de4ce988f822a44b8fc8a6

Not shown in the video but have also confirmed that if we _don't_ accept the plugin, we still get OSC 9 notifications via the old path.

## Agent Mode


app/src/terminal/view.rs
```diff
@@ -12977,7 +12974,7 @@ impl TerminalView {
             cwd: None,
             project: None,
             payload: CLIAgentEventPayload {
-                plugin_version,
+                plugin_version: None,
```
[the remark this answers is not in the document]
liliwilson: Think this was a bad merge, restoring!

app/src/terminal/cli_agent_sessions/mod.rs
```diff
@@ -147,6 +147,24 @@ impl CLIAgentSession {
...
     }
 
+    /// Whether this session is backed by a connected structured plugin (rich
+    /// OSC 777 events) rather than command detection or Codex's OSC 9 fallback.
+    /// Codex's structured plugin always reports a version on connect, so a
+    /// present `plugin_version` distinguishes it from the versionless OSC 9
+    /// fallback. Other agents only ever connect via the structured plugin.
+    pub fn has_structured_plugin(&self) -> bool {
+        self.listener.is_some()
+            && (!matches!(self.agent, CLIAgent::Codex) || self.plugin_version.is_some())
```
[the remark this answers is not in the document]
liliwilson: I think this is okay because the Codex plugin always sends a version with `SessionStart`, but curious if you think there's a more robust soln here

app/src/terminal/cli_agent_sessions/listener/mod.rs
```diff
@@ -91,15 +78,16 @@ impl CLIAgentSessionHandler for DefaultSessionListener {
...
-/// from the raw text, all OSC 9 notifications are treated as `Stop` (success).
-/// The notification body becomes the event's `query` so it surfaces as the
-/// notification title in the UI.
-struct CodexSessionHandler;
+/// human-readable text. Since there's no way to distinguish notification types from the raw text,
+/// OSC 9 fallback notifications are treated as `Stop` (success).
+#[derive(Default)]
+struct CodexSessionHandler {
+    /// Whether we are using a plugin with OSC777 events or falling back to OSC9.
+    structured_plugin_active: bool,
```
harryalbert: super nit, but I'm not sure if `structured` is the correct like differentiating adjective (because the OSC 9 notifications are also structured, they're just not as detailed). Maybe hook-based (doesn't quite roll of the tongue lol)

app/src/terminal/cli_agent_sessions/mod.rs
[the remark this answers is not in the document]
harryalbert: Yeah, I _think_ this should be fine but it might be nice to have a clearer approach.

Does the like fallback system even send a `SessionStart` event (IIRC idt it does)? If not, couldn't we just index on receiving that to know if we're dealing with a structured plugin?

app/src/terminal/cli_agent_sessions/mod.rs
```diff
@@ -147,6 +147,24 @@ impl CLIAgentSession {
...
+    }
+
+    /// Whether the session surfaces trustworthy fine-grained status
+    /// (in-progress / blocked / success). True exactly when a structured
+    /// plugin is connected; Codex's OSC 9 fallback only emits opaque `Stop`
+    /// notifications and so does not qualify.
+    pub fn supports_rich_status(&self) -> bool {
```
harryalbert, line 160: nit, but it seems odd to have this fn as just a pass through to call `has_structured_plugin`. Could we just inline `has_structured_plugin` here? and get rid of that fn?

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
```diff
@@ -1,54 +1,231 @@
...
+const MARKETPLACE_REPO: &str = "warpdotdev/codex-warp";
+const MARKETPLACE_NAME: &str = "codex-warp";
+
+const PLATFORM_PLUGIN_KEY: &str = "orchestration@codex-warp";
+
+const CODEX_CONFIG_DIR: &str = ".codex";
+const CODEX_HOME_ENV: &str = "CODEX_HOME";
+
+// Keep in sync with the plugin version in warpdotdev/codex-warp.
+const MINIMUM_PLUGIN_VERSION: &str = "0.4.0";
```
harryalbert, line 27: OOC why are we starting w/ 0.4.0 if the plugin was just created?

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
```diff
@@ -1,54 +1,231 @@
...
+            .await?;
+        Ok(())
+    }
+
+    async fn update(&self) -> Result<(), PluginInstallError> {
+        if !FeatureFlag::CodexPlugin.is_enabled() {
+            return Ok(());
+        }
+        let mut log = String::new();
+        // Remove and re-add the marketplace to refresh the local snapshot, then
```
harryalbert: the reason we do this in the claude update flow is because the update flow is kind of buggy, but I think an update command would be preferable if codex has one (and it's not buggy lol)

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
```diff
@@ -1,54 +1,231 @@
...
+        } else {
+            &EMPTY_INSTRUCTIONS
+        }
+    }
+
+    fn supports_update(&self) -> bool {
+        FeatureFlag::CodexPlugin.is_enabled()
+    }
+
+    async fn install_platform_plugin(&self) -> Result<(), PluginInstallError> {
```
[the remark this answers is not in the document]
harryalbert: assuming this'll be fixed in a follow-up PR for the platform stuff?

harryalbert (approved):
Nice! didn't realize that the codex plugin supports hooks now.

Just had a few small nits but overall LGTM

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
liliwilson: We reported a plugin version from the old hackier sidecar plugin, this bumps from the latest version there that was `0.3.0`: https://github.com/warpdotdev/codex-warp-internal/blob/main/scripts/on-session-start.sh

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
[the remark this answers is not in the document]
liliwilson: indeed

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
liliwilson: What sorts of bugs were you seeing with claude's update flow? Codex has a `marketplace upgrade` similar to Claude `marketplace update`

liliwilson: Chatted offline!

app/src/terminal/cli_agent_sessions/mod.rs
liliwilson: Renaming to `has_rich_plugin` and killing this to fix the other comment too! Agree

app/src/terminal/cli_agent_sessions/mod.rs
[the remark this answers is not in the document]
liliwilson: I fixed this in 28542e24ba07e2600bf00b5ad6ab9116c03ea7b4

Instead, I added tracking for the event source on the `CLIAgentEvent` and check whether we have a "rich notification", which we store on the session. A bit cleaner IMO! Thanks for the push

PR #11887, opened 2026-05-29
feat: add retention increasing capabilities

dmichelin:
## Description

Adds client support for requesting extended shared-session retention when an agent environment setup fails.

This wires the new `session-sharing-protocol` retention API through the Warp client so an agent setup failure can preserve the shared session long enough for follow-up/debugging:

- Bumps `session-sharing-protocol` to the revision that includes `SessionRetentionReason` and `UpstreamMessage::ExtendSessionRetention`.
- Extends `AgentDriver` to request retention extension when setup fails with `AgentDriverError::EnvironmentSetupFailed`.
- Adds a terminal-driver path that emits `Event::ExtendSessionRetention` only when the terminal is an active shared-session sharer.
- Handles the new terminal view event in `TerminalManager` and forwards it to the shared-session network layer.
- Sends `UpstreamMessage::ExtendSessionRetention { reason }` from the sharer network.
- Fixes the strict warnings build by removing the unused `TerminalView::extend_shared_session_retention` wrapper and emitting the event directly from the driver path.

## Testing

Tests include testing that terminal extension does not happen before sharing is active, and once the terminal mode is shared, if setup fails then an extend session retention message is emitted.

## Agent Mode

CHANGELOG-NONE


app/src/ai/agent_sdk/driver/terminal.rs
```diff
@@ -707,3 +731,66 @@ impl TerminalDriver {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
seemeroland: nit: move tests to separate file

app/src/ai/agent_sdk/driver.rs
```diff
@@ -794,6 +805,16 @@ impl AgentDriver {
             // Success/blocked/cancelled are handled by LocalAgentTaskSyncModel.
             if let (Some(task_id), Err(err)) = (task_id, &result) {
                 report_driver_error(task_id, err, &server_api_for_error).await;
+                if matches!(err, AgentDriverError::EnvironmentSetupFailed(_)) {
```
seemeroland, line 808: Can we consolidate this with the if block directly below keeping session alive with `idle_on_complete`?

seemeroland (commented):
Nice just nits. Once these are in it'd be good to validate e2e that the webapp links to the session and it can be loaded (for up to 7 days) when environment setup fails

seemeroland (approved):

PR #11892, opened 2026-05-29
Auto-install Codex orchestration plugin for harness runs

kjankov:
## Description
Stacked on #11871 / `lili/add-codex-plugin`.

This adds the driver-side Codex support needed for Oz-hosted Codex runs to use the new Codex plugin path end to end:

- detects installed/update state for `orchestration@codex-warp`
- installs or updates the Codex orchestration plugin before driver-launched harness sessions
- verifies the platform plugin cache version after install/update
- launches Codex with `--dangerously-bypass-hook-trust` so Warp-installed hooks can run in unattended driver sessions
- adds targeted unit coverage for platform plugin state and Codex command construction

Agent artifacts:
- Conversation: https://app.warp.dev/conversation/b63f2b9a-7c35-4e27-931b-66223afbc690
- Plan: https://app.warp.dev/drive/notebook/FPRoEC4C3mMt4jMRGsFIyn

## Linked Issue
N/A

## Testing
- [x] `cargo fmt --all --check`
- [x] `CARGO_TARGET_DIR=/Users/katarina/Developer/warp/target cargo clippy --workspace --all-targets --all-features --tests -- -D warnings`
- [x] `CARGO_TARGET_DIR=/Users/katarina/Developer/warp/target cargo test --manifest-path app/Cargo.toml --lib codex_command`
- [x] `CARGO_TARGET_DIR=/Users/katarina/Developer/warp/target cargo test --manifest-path app/Cargo.toml --lib platform_plugin`

### Screenshots / Videos
N/A

## Agent Mode

CHANGELOG-NONE


app/src/ai/agent_sdk/driver/harness/codex.rs
```diff
@@ -43,6 +43,9 @@ pub(crate) struct CodexHarness;
 const CODEX_CLI_FORMAT: &str = "codex_cli";
 /// Slash command Codex's TUI recognises as a graceful shutdown.
 const CODEX_EXIT_COMMAND: &str = "/exit";
+/// Allow the Warp-installed Codex plugin hooks to run in vetted driver sessions
+/// without requiring an unattended `/hooks` review step.
+const CODEX_BYPASS_HOOK_TRUST_FLAG: &str = "--dangerously-bypass-hook-trust";
```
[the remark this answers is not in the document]
liliwilson: For this, did you consider pre-seeding the hooks values in the configs so we just skip verification for the Warp-specific hooks? That may be a safer approach if it's not too much of a pain.

Also curious about what happens if I run this and then run a new local codex session - will I be prompted to accept the orchestration hooks?

app/src/ai/document/ai_document_model.rs
```diff
@@ -16,6 +16,7 @@ use warp_editor::model::RichTextEditorModel;
 use warp_editor::render::model::RichTextStyles;
 use warpui::color::ColorU;
 use warpui::{AppContext, Entity, EntityId, ModelContext, ModelHandle, SingletonEntity, WindowId};
+#[rustfmt::skip]
```
liliwilson: Is this necessary?

app/src/ai/local_harness_setup.rs
```diff
@@ -28,13 +33,20 @@ impl LocalHarnessSetupState {
...
+        }
         Harness::Oz | Harness::Claude | Harness::OpenCode | Harness::Gemini | Harness::Unknown => {
             None
         }
+        Harness::Codex => None,
     }
 }
 
+fn local_codex_harness_is_enabled_for_testing() -> bool {
+    cfg!(debug_assertions) && FeatureFlag::LocalClaudeCodexChildHarnesses.is_enabled()
```
liliwilson: Why do we need the `debug_assertions` gate here as well as the feature flag?

crates/warp_features/src/lib.rs
```diff
@@ -906,7 +906,11 @@ static USER_PREFERENCE_MAP: [AtomicTriState; cardinality::<FeatureFlag>()] =
 static FEATURES_INITIALIZED: AtomicBool = AtomicBool::new(false);
 
 /// Features used in debugging.
-pub const DEBUG_FLAGS: &[FeatureFlag] = &[FeatureFlag::DebugMode, FeatureFlag::RuntimeFeatureFlags];
+pub const DEBUG_FLAGS: &[FeatureFlag] = &[
+    FeatureFlag::DebugMode,
+    FeatureFlag::RuntimeFeatureFlags,
+    FeatureFlag::LocalClaudeCodexChildHarnesses,
```
liliwilson: I don't think this needs to be a debug flag, right? If it's not ready for dogfooding, we can keep the flag out of `DOGFOOD_FLAGS`

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
```diff
@@ -258,43 +339,129 @@ fn check_installed(codex_dir: &Path) -> bool {
...
+    let contents = fs::read_to_string(config_path).ok()?;
+    let parsed = contents.parse::<toml_edit::DocumentMut>().ok()?;
+    let marketplace = parsed.get("marketplaces")?.get(MARKETPLACE_NAME)?;
+    Some(CodexWarpMarketplaceConfig {
+        source_type: marketplace
+            .get("source_type")
+            .and_then(|source_type| source_type.as_str())
+            .map(str::to_owned),
+    })
+}
```
liliwilson, line 400: Why do we need all of this? The approach for Claude seems much simpler with `claude_code_marketplace_has_local_override`, so curious

(And just to make sure I understand - this is to help with ease of testing, right?)

app/src/ai/agent_sdk/driver/harness/mod.rs
```diff
@@ -156,6 +156,12 @@ pub(crate) trait ThirdPartyHarness: Send + Sync {
         &[]
     }
 
+    /// Whether this harness requires its Oz platform plugin to be verified
+    /// before launch.
+    fn requires_verified_platform_plugin(&self) -> bool {
```
liliwilson, line 163: Why do we require this for codex and not claude? (Is it because of the dangerously bypass hooks flag?)

app/src/ai/agent_sdk/driver.rs
```diff
@@ -2127,29 +2127,65 @@ impl AgentDriver {
...
-            let plugin_result = if manager.needs_update() {
-                manager.update().await
-            } else if !manager.is_installed() {
-                manager.install().await
-            } else {
-                Ok(())
-            };
-            if let Err(e) = plugin_result {
-                log::warn!("Plugin installation/update failed (continuing): {e}");
+            if manager.can_auto_install() {
```
liliwilson: Very non-blocking nit: thoughts on pulling some of this into helpers for readability, since the if blocks are getting a bit gnarly? something like:
```
async fn `setup_harness_plugins` {
     // grab the plugin manager
     Self::setup_notification_plugin(manager, ...)
     Self::setup_platform_plugin(manager, ...)
}
```

app/src/ai/blocklist/action_model/execute/start_agent.rs
```diff
@@ -36,6 +37,53 @@ fn invalid_local_child_harness_error(harness_type: &str) -> String {
     }
 }
 
+fn parse_legacy_local_child_harness_command(command: &str) -> Option<(String, String)> {
```
liliwilson, line 42: Where do the inputs to this come from - is the risk here that an agent will use an old way of specifying a local harness?

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
```diff
@@ -227,28 +299,37 @@ static EMPTY_INSTRUCTIONS: LazyLock<PluginInstructions> = LazyLock::new(|| Plugi
...
+    check_plugin_installed(codex_dir, PLUGIN_KEY, PLUGIN_NAME)
+}
+
+fn check_platform_plugin_installed(codex_dir: &Path) -> bool {
+    check_plugin_installed(codex_dir, PLATFORM_PLUGIN_KEY, PLATFORM_PLUGIN_NAME)
+}
+
+fn check_plugin_installed(codex_dir: &Path, plugin_key: &str, plugin_name: &str) -> bool {
+    marketplace_plugin_version(codex_dir, plugin_name).is_some()
+        || check_legacy_plugin_enabled(codex_dir, plugin_key)
```
liliwilson: Why do we need to check both of these places? Is codex not consistent across the two

liliwilson (commented):
Left some comments with questions! Main thing I'm curious about whether it's possible to avoid using `--dangerously-bypass-hook-trust` for skipping the hooks dialog

app/src/ai/agent_sdk/driver/harness/codex.rs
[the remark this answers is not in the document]
kjankov: I considered pre-seeding hook trust, but avoided it for this PR because it would couple us to Codex's hook-trust config internals. The current approach instead fails harness setup unless the Codex orchestration marketplace plugin is installed and at the required version before launching with `--dangerously-bypass-hook-trust`.

That flag is only passed on the harness launch command; we do not persist hook-trust acceptance for the user. A separate local Codex session should still go through Codex's normal hook review behavior unless the user has independently accepted those hooks.

app/src/ai/document/ai_document_model.rs
kjankov: No — removed this. I rewrote the imports/re-exports so the formatter is stable without `#[rustfmt::skip]`.

app/src/ai/local_harness_setup.rs
kjankov: Agreed, the extra `debug_assertions` gate wasn't needed. This is now gated only by `LocalClaudeCodexChildHarnesses`.

crates/warp_features/src/lib.rs
kjankov: Agreed. I removed it from `DEBUG_FLAGS` and left it as an explicit runtime feature flag instead of enabling it by default for debug/dogfood builds.

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
kjankov: Simplified this. The manager now checks the current Codex marketplace manifest path directly instead of combining marketplace, legacy config, and cache lookups. The local marketplace override check remains only to avoid clobbering a developer's local marketplace while testing.

app/src/ai/agent_sdk/driver/harness/mod.rs
kjankov: Yes — Codex is different because the harness launch uses `--dangerously-bypass-hook-trust`, which bypasses hook trust globally for that Codex process. Claude does not need that bypass, so Codex is the one that fails setup if the platform plugin cannot be verified. I added this context to the trait docs.

app/src/ai/agent_sdk/driver.rs
kjankov: Done. I split the setup flow into `setup_harness_plugins`, `setup_notification_plugin`, `setup_platform_plugin`, and `verify_required_platform_plugin` helpers so the required-platform-plugin path is easier to follow.

app/src/ai/blocklist/action_model/execute/start_agent.rs
kjankov: The input comes from older StartAgent payloads where local mode did not set `harness_type` yet and encoded the intended local Codex launch as the prompt text. The normalization keeps those legacy requests routed through the Codex local harness instead of treating the literal `codex ...` command as an Oz child prompt. I added a comment explaining that.

app/src/terminal/cli_agent_sessions/plugin_manager/codex.rs
kjankov: It is not consistent across the old and current Codex plugin layouts, but we do not need to support both here. I simplified this to check only the current marketplace manifest path, so the legacy config/cache fallback is gone.

crates/warpui_core/src/lib.rs
liliwilson, line 1: Was this intentional?

app/src/ai/agent_sdk/driver/harness/codex.rs
```diff
@@ -85,6 +88,10 @@ impl ThirdPartyHarness for CodexHarness {
         ]
     }
 
+    fn requires_verified_platform_plugin(&self) -> bool {
+        true
```
liliwilson, line 92: Can we set this to be the feature flag (`CodexPlugin`) so that we don't enforce this if it's not possible to use the plugin? Otherwise I think we'll start failing driver setup in staging for cloud agents

app/src/ai/agent_sdk/driver/harness/codex.rs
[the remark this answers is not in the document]
liliwilson: I do worry about the user experience of running a local subagent + seeing the orchestration hooks popup in a later, unrelated local codex session. We do rely a lot on codex internals already, so it wouldn't be out of line with that to do something a bit more invasive here. 

That said, this is probably fine for now and we can iterate more later if it seems like this is something people are running into

liliwilson (approved):

PR #11903, opened 2026-05-29
Add tab group more options menu + bug fix

johnturcoo:
## Description

Adds a "more options" menu for tab groups in the vertical tabs panel, plus a fix for moving tabs around groups.

Tab group more-options menu — opens from the group header's more options button or by right-clicking the group container/header. Context-aware items:
•  Ungroup tabs, New tab in group
•  Move group up / down (shown only when there's a neighboring tab/group in that direction)
•  Rename
•  Close all tabs in group / Close other tabs / Close tabs above / Close tabs below (shown based on surrounding tabs)

Bug fix: moving a tab no longer lands inside a group — moving a tab up/down (per-tab menu / shortcut) now hops over an adjacent group as a single unit instead of wedging the tab between that group's members and breaking its contiguous run. Within-group reordering and moves past ungrouped tabs are unchanged. Implemented via a shared `hop_tab_to_index` reorder helper.

## Changes

-  `app/src/workspace/action.rs`: Added eight WorkspaceAction variants for the group menu (`ToggleTabGroupRightClickMenu`, `UngroupTabs`, `NewTabInGroup`, `MoveTabGroupUp/Down`, `CloseTabsOutsideGroup/AboveGroup/BelowGroup`) 
-  `app/src/workspace/view.rs:` Added `show_tab_group_right_click_menu` state plus `toggle_tab_group_right_click_menu`, which builds items via `tab_group_menu_items` and shows them on the shared `tab_right_click_menu` view. Added the action helpers it dispatches to (`ungroup_tabs`, `new_tab_in_group`, `move_tab_group/move_group_block`, `close_tabs_outside/above/below_group`) and the `group_member_indices/group_member_index_range` list utilities they share.
- `app/src/workspace/view.rs` (tab move fix): Added `hop_tab_to_index` and reworked `move_tab` to compute a target that hops over an adjacent group as a single unit, so moving a tab past a group no longer wedges it inside the group's contiguous run.
- All of the above is gated behind `FeatureFlag::GroupedTabs`.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4622/tab-group-more-options-menu-actions

## Testing

### Screenshots / Videos

[Menu demo video](https://www.loom.com/share/033aff80a8ac433aafd5630243d98454)

[Original bug demo](https://www.loom.com/share/25531aeabd5a48d297bf9e800b27ab8e)

[Bug fix demo](https://www.loom.com/share/3691eaac987d42a2844bc75fe943ebf0)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -959,6 +959,8 @@ pub struct Workspace {
     show_tab_bar_overflow_menu: bool,
     tab_right_click_menu: ViewHandle<Menu<WorkspaceAction>>,
     show_tab_right_click_menu: Option<(usize, TabContextMenuAnchor)>,
+    /// Open tab group more-options menu; reuses the `tab_right_click_menu` view.
+    show_tab_group_right_click_menu: Option<(TabGroupId, TabContextMenuAnchor)>,
     // TODO(CORE-2300): this used to be add_tab_dropdown_menu.
     // Because we are rolling out the change behind a feature flag,
     // keep this comment here until the feature flag is removed.
@@ -3213,6 +3215,7 @@ impl Workspace {
             show_tab_bar_overflow_menu: false,
             tab_right_click_menu,
             show_tab_right_click_menu: None,
+            show_tab_group_right_click_menu: None,
             new_session_dropdown_menu,
             show_new_session_dropdown_menu: None,
             changelog_model,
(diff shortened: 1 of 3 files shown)


app/src/workspace/view.rs
```diff
@@ -959,6 +959,8 @@ pub struct Workspace {
     show_tab_bar_overflow_menu: bool,
     tab_right_click_menu: ViewHandle<Menu<WorkspaceAction>>,
     show_tab_right_click_menu: Option<(usize, TabContextMenuAnchor)>,
+    /// Open tab group more-options menu; reuses the `tab_right_click_menu` view.
+    show_tab_group_right_click_menu: Option<(TabGroupId, TabContextMenuAnchor)>,
```
vkodithala, line 963: nit: Should this be named `tab_group_right_click_menu`? `show_` makes it seem like this should be a bool?

app/src/workspace/view.rs
```diff
@@ -6941,6 +6944,202 @@ impl Workspace {
...
+            NewSessionSource::Tab,
+            Some(ctx.window_id()),
+            None,
+            None,
+            false,
+            ctx,
+        );
+
+        // If the active tab is a member of the group, the new tab inherits this group on creation.
+        // Otherwise we must manually update it here, and place this new tab at the end of the group.
```
vkodithala, line 6960: Nice attention to detail 🫡

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -128,6 +128,11 @@ fn vtab_pane_row_position_id(pane_group_id: EntityId, pane_id: PaneId) -> String
     format!("vertical_tabs:pane_row:{pane_group_id:?}:{pane_id}")
 }
 
+/// Save-position id for a tab group header's kebab button; anchors the group menu.
+pub(crate) fn vtab_group_kebab_position_id(group_id: TabGroupId) -> String {
```
vkodithala: Very minor nit, do as you wish with this lol: can we rename this param to `tab_group_id` to disambiguate from pane groups? Saw that this is what `vtab_pane_row_position_id` did above so maybe we just wanna adopt the same.

app/src/workspace/view.rs
```diff
@@ -7002,11 +7247,38 @@ impl Workspace {
...
+        &mut self,
+        group_id: TabGroupId,
+        anchor: TabContextMenuAnchor,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        if self.show_tab_group_right_click_menu.is_some() {
+            self.show_tab_group_right_click_menu = None;
+            ctx.notify();
+            return;
+        }
```
vkodithala, line 7250: Very major nit, but can we use `.take()` here? Functionally same but might be a lil more ergonomic. You can use your discretion here though and feel free to push back (maybe this is overkill since we don't want to extract the owned val?).

app/src/workspace/view.rs
```diff
@@ -7002,11 +7247,38 @@ impl Workspace {
...
+        if !self.tab_groups.contains_key(&group_id) {
+            return;
+        }
+
+        let menu_items = self.tab_group_menu_items(group_id);
+        ctx.update_view(&self.tab_right_click_menu, |context_menu, view_ctx| {
+            context_menu.set_items(menu_items, view_ctx);
+        });
+        self.show_tab_right_click_menu = None;
```
vkodithala, line 7259: Another case for using `.take()` here; you could simplify to smth like:
```
if self.show_tab_group_right_click_menu.take().is_some() { ... }
// No need for the double `self.show_tab_right_click_menu = None`
```

app/src/workspace/view.rs
```diff
@@ -26030,6 +26466,17 @@ fn group_member_indices(
         .map(|(idx, _)| idx)
 }
 
+/// Returns the `(first, last)` index range for the contiguous run of tabs
+/// in `tabs` that belong to `group_id`, or `None` if the group has no members.
+/// The run is assumed to be contiguous (the workspace enforces this invariant);
+/// only the earliest and latest matching indices are returned.
+fn group_member_index_range(tabs: &[TabData], group_id: TabGroupId) -> Option<(usize, usize)> {
```
vkodithala, line 26402: nit: Is there a case for maybe moving this fn to `impl Workspace` and getting rid of the extra `tabs` param? From a quick scan it seems like tabs always == Workspace.tabs? Lmk

app/src/workspace/view.rs
```diff
@@ -12768,30 +13124,60 @@ impl Workspace {
...
+                match self.tabs[neighbor].group_id {
+                    // A different group sits to the right: target its last
+                    // member; the remove+insert leaves the tab just past it.
+                    Some(group_id) if Some(group_id) != moved_group_id => {
+                        group_member_index_range(&self.tabs, group_id)
+                            .map(|(_, last)| last)
+                            .unwrap_or(neighbor)
+                    }
+                    // Ungrouped neighbor, or a member of our own group: step one.
+                    Some(_) | None => neighbor,
```
vkodithala, line 13108: Very nice, these are great comments; made this code v easy to follow

app/src/workspace/view.rs
```diff
@@ -12768,30 +13124,60 @@ impl Workspace {
...
+        if moving_active_tab {
             send_telemetry_from_ctx!(TelemetryEvent::MoveActiveTab { direction }, ctx);
         } else {
-            // Don't want to change the active tab for the user due to an adjacent
-            // tab being moved left/right.
-            if new_index == self.active_tab_index {
-                self.set_active_tab_index(index, ctx);
-            }
             send_telemetry_from_ctx!(TelemetryEvent::MoveTab { direction }, ctx);
         }
```
vkodithala, line 13124: q, out-of-scope: Is there a reason we send _separate_ telemetry for moving the active tab and moving any tab? Is it important to capture this data? Just curious.

vkodithala (approved):
Great PR, left a bunch of minor comments (non-blocking, just curiosity/food for thought mostly). Amazing work!

app/src/workspace/view.rs
johnturcoo: Here I followed the existing convention in Workspace — `show_tab_right_click_menu`, `show_new_session_dropdown_menu`, and `show_header_toolbar_context_menu` are all sibling show_*: Option<…> fields where None means hidden and Some(payload) carries the anchor/positioning data. Essentially acting as a bool

app/src/workspace/view.rs
johnturcoo: These are actually clearing two different menus:  tab_group and tab respectively.

app/src/workspace/view/vertical_tabs.rs
johnturcoo: Good call

app/src/workspace/view.rs
johnturcoo: Very good call out, thanks! Going to move this to a follow up PR with other bug fixes because of wide surface area

app/src/workspace/view.rs
johnturcoo: I believe it is to track moving a tab with keybinding `shift-ctrl-left/right`, which can only be used for the active tab. However, I could be wrong. This telemetry was pre existing

PR #11923, opened 2026-05-30
Layout-aware keyboard shortcuts: fix Cmd/Ctrl shortcuts on non-Latin keyboard layouts

### LINKED ISSUE #7026: Support multiple hotkeys per command for different keyboard layouts
### Pre-submit Checks

Currently, Warp allows only one hotkey per command, which can be inconvenient for users who switch between keyboard layouts (e.g., English and non-English). I propose adding support for multiple hotkeys per command, so that a single action can be triggered by different key combinations depending on the active layout.

For example:

A command could be bound to both comand+I (English) and comand+Ш (Russian layout).

Warp would recognize both keybindings as valid triggers for the same action.

This would eliminate the need to switch layouts just to use hotkeys, improving workflow efficiency.

_No response_

### Additional context

_No response_

### Operating system (OS)

macOS

4

None

Desko77:
# Layout-aware keyboard shortcuts: fix Cmd/Ctrl shortcuts on non-Latin keyboard layouts

## The problem

On a Russian keyboard layout (or Greek, or any other non-Latin layout), Cmd+C / Ctrl+C
and most other letter shortcuts silently stop working. The physical C key reports
its layout character - "с" in Cyrillic - as winit's logical key, the binding stores
"c", and the two never match. You have to switch back to a Latin layout just to copy
a line. For anyone who types in two scripts this is a constant papercut.

## How it works

There are two ways to make a shortcut match on the physical key rather than the logical one.

1. Per-binding, in keybindings.yaml. A new `Code(NAME)` form, for example
   `editor:copy: cmd-Code(KeyC)`, where NAME is a W3C UIEvents code. That binding
   fires on the physical key regardless of the active layout. A plain `cmd-c` stays
   logical and behaves exactly as it does today.

2. A global "Smart layout-aware shortcuts" toggle (Settings -> Features -> Keys, or
   `keyboard.smart_layout_aware_bindings` in settings). With it on, single letter and
   digit logical bindings (`KeyA..KeyZ`, `Digit0..Digit9`) are promoted to physical
   matching. Symbol shortcuts like Cmd+/ are left alone on purpose - the slash sits on
   different physical keys across layouts, so promoting it would cause more trouble than
   it solves. IME composition is detected and exempted so it never interferes with
   normal typing.

Both default to off, so nothing changes for existing users until they opt in.

Three ways to make a shortcut layout-independent: flip the master switch (one click,
covers every alphanumeric binding), use the per-binding globe toggle in the keybinding
editor popup, or hand-write a `Code(NAME)` entry in the config. Bindings in physical form
get a globe indicator in the list, so it's obvious at a glance which ones survive a layout switch.

## Implementation

- No new fields on `Keystroke`. The physical form is encoded inside the existing key
  string (`Code(KeyC)`), so every struct-literal construction site compiles unchanged
  and YAML still round-trips cleanly. `match_mode()` and `physical_code()` are derived
  from the key string.
- `Event::KeyDown` carries a `physical_code` captured from winit's `physical_key`. The
  matcher pairs each pending event with its physical code and decides per pair which of
  the three regimes applies: explicit physical, smart promotion, or plain logical.
- Platform helpers for `physical_key` to/from string live in
  `crates/warpui_core/src/platform/keyboard.rs`.

## Testing

- keymap and matcher unit tests cover: `Code(NAME)` parse round-trip, rejection of
  unknown codes, explicit physical bindings matching across layouts, smart promotion of
  alphanumerics, symbols left untouched, and IME suppression.
- Built and run as a release build on Windows.

## Known gap: macOS

macOS does not capture `physical_key` from NSEvent yet, so the Smart toggle on macOS
only works when a physical code is already present. Explicit `Code(NAME)` bindings are
unaffected.

Closes #7026


Desko77:
@cla-bot check

PR #11924, opened 2026-05-30
feat(terminal): jump to latest agent message (#11816)

### LINKED ISSUE #11816: Auto-bookmark the last agent output so you can jump back after running commands
### Pre-submit Checks

In an agent view, when you run one or more commands, automatically bookmark the last agent output, and give a one-action jump back to it. The bookmark advances as the conversation continues, so it always points at the latest agent message.

When I'm in an agent conversation and then run terminal commands in the same thread, the agent's last message gets buried under all the command output (logs, commits, pushes). To reread what the agent said, I have to scroll up through all of it.

### Additional context

Different from what's there today:
  - Not the manual toggle-bookmark you place by hand. This is automatic and agent-aware.
  - Not #11120 (pinning external AI CLI sessions). That pins whole sessions. This bookmarks the last agent message within a thread for quick jump-back.
  
Context: David Stern raised this on the "How Warp Works" stream (2026-05-20, around 1:12). Ben offered it to me to write up, so filing it here.

### Operating system (OS)

Cross-platform

3

None

maxmilian:
## Description

Adds a **"Jump to latest agent message"** command (#11816). When you're in an agent conversation, `ESC` back to the terminal, and then run commands in the same thread, the agent's last message gets buried. This command takes you straight back to it.

**Key behavior:** agent messages only render *inside the agent view* — once you return to the terminal they collapse to a hidden, zero-height block. So there is nothing to scroll the terminal viewport *to*; the only way to actually show the latest agent message is to re-enter the agent view. This command does exactly that, re-entering the agent view for the most recent conversation (which lands on its latest message).

> Note: an earlier draft of this PR tried to scroll the terminal block list to the agent's `AIBlock`. Testing showed that can't work — in terminal mode those blocks are `should_hide` / zero-height, so the jump landed on the collapsed conversation-entry marker and the message was never visible. Re-entering the agent view is the correct mechanism.

**Implementation (minimal):**
- `TerminalView::jump_to_latest_agent_message` (`app/src/terminal/view.rs`): looks up the most recent conversation via `BlocklistAIHistoryModel::last_conversation_id` and calls the existing `enter_agent_view_for_conversation`. No-op when there is no agent conversation.
- New `AgentViewEntryOrigin::JumpToLatestAgentMessage` (+ telemetry mapping in `TelemetryAgentViewEntryOrigin`) so this entry point is attributed correctly.
- New no-arg `TerminalAction::JumpToLatestAgentMessage`, registered in the command palette (`init.rs`). **No default keybinding.**

It reuses `enter_agent_view_for_conversation`, which already handles the live / in-memory / load-from-disk conversation cases (with error toasts), so this adds no new state or traversal logic.

### Questions for reviewers
- **Keybinding:** registered with **no default keystroke** (command-palette only). Happy to bind whatever you'd prefer.

## Linked Issue

Fixes #11816

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. (#11816 is `ready-to-implement`)
- [x] Screenshots / short video (below).

## Testing

Repro: start an agent conversation → `ESC` to the terminal → run `seq 1 400` (or `git log`) to bury the conversation → command palette → **Jump to Latest Agent Message**. The agent view re-opens on the most recent conversation, showing its latest message.

### Screenshots / Videos

https://github.com/user-attachments/assets/7c0d9bbf-7e61-4aa5-902d-a0ece3a4e67f

## Agent Mode


maxmilian:
Thanks for the review, @oz-for-oss — addressed the IMPORTANT concern in 0ab8e5a0.

The command now follows actual agent activity instead of creation order: it uses the history model's `active_conversation_id(self.id())` (the "currently or most recently streamed" pointer in `active_conversation_for_terminal_view`), falling back to `last_conversation_id` only when there is no active conversation.

- This pointer is updated by `set_active_conversation_id` / `mark_active_conversation_id` on stream start and follow-ups, and is **not** cleared when the user presses ESC back to the terminal — so after burying the conversation and jumping back, it resolves to the conversation that last streamed, even if the user switched back to an older one.
- It uses the history-model version (keyed by `EntityId`), not the `AgentViewState` one (which goes `Inactive` on ESC and would return `None` for this exact flow).
- Consistent with the existing `ToggleQueueNextPrompt` / `ForkConversationFromLastKnownGoodState` handlers, which already prefer `active_conversation_id(self.id())`.

/oz-review

harryalbert (changes requested):
Thanks for the implementation here! However, I think that this implementation is missing a step.

I think that, when you run "jump to latest agent message", we should:
* enter the agent view for the most recent conversation (already implemented in this PR), but only if you're not already in that agent view
* scroll to the most recent _agent_ block in that agent view (not implemented yet from what I can t)

That second step is important because you can run terminal commands inside of an agent view, and you might want to use this functionality to scroll back to the most recent agent response before those terminal blocks.

maxmilian:
Thanks for the review @harryalbert — addressed both points:

**1. Already-in-view guard** — the command now only re-enters the agent view when we're not already in that conversation's view; when already in it, it scrolls in place (no re-enter churn).

**2. Scroll to the latest agent block** — instead of relying on entry's default scroll, it resolves the conversation's latest exchange from the model and scrolls to that block, so it lands on the most recent agent response (above any terminal blocks run after it) rather than the bottom.

While testing I also found the command wasn't reachable from the command palette *while in* the agent view: its context predicate matched only `Terminal`, but when the agent-view input has focus the active context is `Input` + the active-agent-view flag, so it was filtered out — which also made the in-view path (point 1) unreachable. The predicate now matches both.

Implementation note: each agent turn renders as a single block, so jump lands on the top of the latest agent block. From the terminal, entering the agent view kicks off its own scroll on a later tick, so I defer + briefly re-assert the scroll across a few frames so it isn't overridden — happy to switch to a cleaner approach if there's a "view entered / first render" hook I missed.

Tested from the terminal and from within the agent view (incl. with terminal blocks after the latest agent response); both consistently land on the latest agent block. Recording in the updated description.

app/src/terminal/view.rs
```diff
@@ -21838,6 +21838,64 @@ impl TerminalView {
...
+            .or_else(|| history.last_conversation_id(self.id()))
+        else {
+            return;
+        };
+        // Resolve the target exchange from the conversation model rather than from
+        // the currently-mounted blocks: when entering from the terminal the blocks
+        // mount over later frames, so the latest block may not exist yet this tick.
+        let Some(exchange_id) = BlocklistAIHistoryModel::as_ref(ctx)
+            .conversation(&conversation_id)
+            .and_then(|conversation| conversation.latest_exchange())
```
harryalbert: we should make sure that this is the latest _visible_ exchange. We can/should probably just use the `latest_visible_exchange` helper here

app/src/terminal/view.rs
```diff
@@ -21838,6 +21838,64 @@ impl TerminalView {
...
+                None,
+                AgentViewEntryOrigin::JumpToLatestAgentMessage,
+                conversation_id,
+                ctx,
+            );
+            // Entering scrolls the agent view to the bottom on the next render, so
+            // scrolling synchronously here would be overridden by that. Defer to a
+            // later tick — after entry's own scroll and after the blocks mount —
+            // then scroll to the target exchange (retrying while blocks mount).
+            ctx.spawn(futures::future::ready(()), move |me, _, ctx| {
```
harryalbert: this spawn (and `scroll_to_exchange_when_mounted`'s re-attempt loop) is a bit of a code smell IMO. It would be preferable to do this in a single shot where the timeline is easy to reason about and doesn't rely on async retries/polling.

I think one way we could do this is by:
* storing some field like `pending_scroll_target: Option<AIAgentExchangeId>`
* setting that in `jump_to_latest_agent_message` (and removing the `ctx.spawn entirely`)
* checking wither `pending_scroll_target` is `Some` in `after_terminal_view_layout` and, if so, setting it to `None` and scrolling to the target

That way the path is easy to reason about and doesn't require any retry loop or deferred spawn

app/src/terminal/view/init.rs
```diff
@@ -437,6 +437,22 @@ pub fn init(app: &mut AppContext) {
...
+        )
+        // Available from the terminal (enters the latest conversation's agent view)
+        // and from within an active agent view, where the rich input — not the
+        // terminal — holds focus, so its context lacks `Terminal` but carries
+        // `Input` plus the active-agent-view flag. Without the `Input` clause the
+        // command is unreachable from the command palette while in the agent view.
+        .with_context_predicate(
+            (id!("Terminal")
+                | (id!("Input")
+                    & (id!(flags::ACTIVE_AGENT_VIEW) | id!(flags::ACTIVE_INLINE_AGENT_VIEW))))
```
harryalbert: do we need the `ACTIVE_INLINE_AGENT_VIEW` check here?

harryalbert (changes requested):
Thanks for iterating! Had a few nits and one larger piece of architectural feedback

app/src/terminal/view.rs
maxmilian: Done — switched to `latest_visible_exchange()` so it lands on a block that actually renders (skips passive/hidden exchanges) instead of `latest_exchange()`.

app/src/terminal/view.rs
maxmilian: Done, implemented exactly as you outlined. There is now a `pending_agent_scroll_target: Option<AIAgentExchangeId>` field on `TerminalView`:

- `jump_to_latest_agent_message` sets it on the enter-agent-view branch and the `ctx.spawn` is gone; `scroll_to_exchange_when_mounted` is deleted entirely.
- `after_terminal_view_layout` does `if let Some(id) = self.pending_agent_scroll_target.take() { self.scroll_to_exchange(id, ctx); }`.

That hook runs after layout has mounted the target block and after the agent view's own entry scroll, so the single shot lands without the retry loop. (The already-in-view branch still scrolls synchronously since the block is already mounted.)

app/src/terminal/view/init.rs
maxmilian: No — removed it. The command always opens the full-screen agent view (`enter_agent_view_for_conversation` → `try_enter_agent_view` → `enter_agent_view_internal(FullScreen)`), which sets `ACTIVE_AGENT_VIEW`, never `ACTIVE_INLINE_AGENT_VIEW`. So the in-view re-scroll path only ever needs the full-screen flag. Predicate is now `(Terminal | (Input & ACTIVE_AGENT_VIEW)) & !IMEOpen`.

maxmilian:
@harryalbert addressed all three:

1. **Latest *visible* exchange** — now resolves via `latest_visible_exchange()` instead of `latest_exchange()`.
2. **No more spawn/retry loop** — replaced with a `pending_agent_scroll_target` field set on entry and consumed once in `after_terminal_view_layout` (after layout mounts the block and after the entry scroll), per your outline. `scroll_to_exchange_when_mounted` is deleted.
3. **Dropped `ACTIVE_INLINE_AGENT_VIEW`** — the command only ever opens the full-screen agent view, so the inline flag was never the matching one.

`cargo check -p warp` is green. Replied inline on each thread.

harryalbert (changes requested):
this is great, thanks for responding to my comments and iterating!

My last point of feedback is that it would probably be good to add some comprehensive tests for this new action. Once those are in (and this passes CI — I just approved a CI run so we can see if everything's ok there) this should be good to merge!

app/src/terminal/view.rs
```diff
@@ -21838,6 +21854,65 @@ impl TerminalView {
...
+            return;
+        }
+        // Follow actual agent activity: prefer the most recently streamed
+        // conversation (which tracks where the latest agent message landed even
+        // after the user switches back to an older conversation), falling back
+        // to the last-created one only when there is no active conversation.
+        let history = BlocklistAIHistoryModel::as_ref(ctx);
+        let Some(conversation_id) = history
+            .active_conversation_id(self.id())
+            .or_else(|| history.last_conversation_id(self.id()))
```
harryalbert: could we maybe use `latest_exchange_across_all_conversations` instead of `last_conversation_id`? I think it makes sense to prefer the active conversation id if there is one, but if not, I think going back the conversation that last streamed output (instead of the most recently created conversation) would make sense

maxmilian:
@harryalbert thanks — addressed both:

**1. Comprehensive tests** (`6ecb44ed`) — added five tests for the action in `view_tests.rs`:
- no-op when the `AgentView` flag is off (even with a conversation present)
- no-op when there are no conversations
- entering from the terminal enters the agent view for the latest conversation (`origin == JumpToLatestAgentMessage`) and records the `pending_agent_scroll_target` for the layout pass
- targets the latest *visible* exchange when a conversation has several
- when already in that conversation’s agent view, scrolls directly without re-entering or recording a new pending target

**2. `latest_exchange_across_all_conversations` instead of `last_conversation_id`** (`9d3d901a`) — the fallback now targets the single most-recently-streamed exchange across all conversations (and enters the conversation that owns it) rather than the most-recently-*created* one. The active conversation is still preferred when there is one.

`cargo nextest` (5/5), `cargo fmt`, and `cargo clippy --tests -D warnings` are all green.

harryalbert (approved):
LGTM! Thanks for iterating!

Issue #11932, opened 2026-05-30
Support Shift+Click to extend the current terminal text selection (iTerm2-style)


cola-runner:
## Problem

Warp's terminal can't extend an existing text selection with **Shift+Click** the way iTerm2 (and most terminals) can. Today every left-click in the block list starts a brand-new selection, so there's no way to:

- click a start point, scroll far down a long command's output, then Shift+Click an end point to select everything in between, or
- nudge the end of an existing selection outward without re-dragging the whole thing from scratch.

For long / multi-screen output this makes selecting a large range painful — you have to drag and rely on auto-scroll the entire way, and overshooting means starting over.

## Steps to reproduce

1. Run a command that produces a few screenfuls of output.
2. Click (or drag-select) somewhere near the top of the output.
3. Scroll down, hold **Shift**, and left-click near the bottom.

**Expected:** the selection extends from the original anchor to the clicked point (iTerm2 behavior).
**Actual:** the previous selection is discarded and a new (empty) selection starts at the click.

## Expected behavior

With a selection already present, **Shift + left-click** should extend the selection to the clicked point, keeping the original anchor fixed (and you can then drag to fine-tune). A plain click still starts a new selection.

## Notes

The selection model already has the needed primitive: dragging uses an "update the tail, keep the head" operation, while Shift+Click currently falls through to "start a new selection" instead of being wired to it.

I have a fix for this (gated behind a feature flag, scoped to the block list / scrollback for v1) and will open a PR linked to this issue.

cola-runner:
@cla-bot check

leesa-ui-w:
Hey @cola-runner, thanks for the write-up and PR #11933! Same as #3426, closing this one to consolidate. Your PR is still visible.

Issue #9963, opened 2026-05-03
Support Shift+Click to extend existing text selection in terminal output


khush2003:
## Summary
Feature request to support extending an existing terminal text selection with Shift+Click.

## Problem
Warp currently supports click-drag, smart selection (double-click), and rectangular selection, but does not support the common terminal workflow where Shift+Click extends an existing selection from its original anchor to a new click point.

This is especially frustrating on laptop touchpads because it's hard to scroll while actively selecting. A current workaround is to zoom out so the target text roughly fits on screen, select it, then zoom back in.

## Reproduction steps or desired workflow
1. Select a portion of terminal output.
2. Hold Shift.
3. Click at a different location in the output.
4. Desired: selection extends from the original anchor to the clicked location.
5. Current: previous selection clears/restarts instead of extending.

## Artifacts
None attached.

## Warp version
v0.2026.04.29.08.57.stable_01

## Operating system
Pop!_OS 24.04 LTS (Linux)

dannyneira:
We already have a request tracking this over at #923

If you could please give it a 👍 and subscribe to that request, it would help us gauge interest. We'll post any updates there, so stay tuned! 

We'll be closing this as a duplicate.

Issue #4715, opened 2024-04-08
shift+click to select text


tigsss:
### Describe the solution you'd like?

When text is already selected, I would like to be able to press shift+click and automatically select up until that point. This is a functionality in most terminals and is something I use a surprisingly amount of the time but is not present in Warp.

<img width="361" alt="image" src="https://github.com/warpdotdev/Warp/assets/25759582/9bc8fe9c-c23f-4780-bdb1-9bc3e3b78bd6">

I originally did shift from the final t in requirements.txt to the second to last t (highlighting txt). Then shift clicking the first  r in requirements should highlight the entire requirements.txt word.

If I were to select from the second to last to to the final t and then shift click the first r in requirements, it would highlight just the word requirements.

The algorithm is based on where you first select from, pressing shift-click will select all the text from either the cursor into the first select position or first select person to the cursor.

### Is your feature request related to a problem? Please describe.

The problem is that this is supported in other terminals (VSCode, ITerm) but not in Warp.

### How important is this feature to you?

5 (Can't work without it!)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

dannyneira:
Thanks for this feature request @tigsss!

To anyone else interested in this feature, please add a 👍 to the original post at the top to signal that you want this feature, and subscribe if you'd like to be notified.

mattwilliamson:
This is a huge one for me interrupting my flow constantly

Sclipper:
This keeps me from completely dropping vs Code and keeping Warp as my only editor

PR #11933, opened 2026-05-30
Add Shift+Click to extend terminal text selection (iTerm2-style)

### LINKED ISSUE #3426: Shift+click for text selection doesn't work

_No response_

### Describe the bug

I usually use Shift+click many times a day for selecting a block of text, it works in every program I know of except Warp. it works for the text that I can type for the last block, but it doesn't work for previous blocks. Please fix it

### To reproduce

On a block that is not the most recent one, select a character with a mouse, release the click, press and hold Shift on keyboard then move your cursor on a few characters or lines futher and click once.

### Expected behavior

It should the select the text between the original position where you first clicked or selected some text and the wherer you clicked last

### Screenshots

_No response_

### Operating system

MacOS

MacOS 12.5.1

### Shell Version

_No response_

### Current Warp version

v0.2023.07.18.08.03.stable_00

### Regression

### Recent working Warp date

not sure, I'm new

### Additional context

_No response_

### Warp Internal (ignore): linear-label:b8107fdf-ba31-488d-b103-d271c89cac3e

None

cola-runner:
## Summary

Adds iTerm2-style **Shift+Click to extend a terminal text selection**. With a selection already present, **Shift + left-click** moves the nearest selection boundary to the clicked point:

- clicking before the selection extends the start/head side
- clicking after the selection extends the end/tail side
- clicking inside the selection shrinks whichever side is closer

A plain click still starts a new selection, and Shift+Click+drag can continue fine-tuning the extended text selection. This makes selecting long / multi-screen output much easier: pick a start, scroll, then Shift+Click the other side instead of drag-and-auto-scrolling the whole way.

Closes #3426

(Supersedes the duplicate reports #11932, #9963, and #4715 -- same iTerm2-style Shift+Click extend behavior.)

## How it works

The shared selection model now exposes nearest-boundary extension logic, so block-list and alt-screen selections use the same behavior.

- **Selection model** (`selection.rs`): adds reusable nearest-boundary logic for moving start vs end when extending. The moved boundary remains the active update endpoint so Shift+Click+drag keeps fine-tuning the same side.
- **Block list** (`block_list_element.rs`, `view.rs`): Shift+Click dispatches `Extend` when a simple text selection already exists and the click is not on a link / revealed secret. Plain drag still updates the drag endpoint, while Shift+Click+drag continues dispatching text-selection updates instead of being misclassified as block selection.
- **Alt screen** (`alt_screen_element.rs`, `alt_screen.rs`): Shift+Click extension is wired for full-screen apps as well.

## Scope / follow-ups

- AI rich-content block internal selections are intentionally left out of scope for this PR because they are implemented differently.

## Testing

Verified after rebasing onto latest `upstream/master` (`c20645b9`) with this PR applied:

- `cargo fmt`
- `cargo test -p warp nearest_boundary --lib` -- 13 passed
- `cargo test -p warp test_alt_screen_extend_selection_without_existing_selection_noops --lib` -- 1 passed
- `cargo check -p warp`

The `nearest_boundary` coverage includes model tests plus view-level regressions for keeping plain drag updates on the drag endpoint and keeping Shift+Click+drag on the boundary selected by Extend.

Earlier macOS GUI evidence for the cross-screen block-list workflow is here: https://gist.github.com/cola-runner/37a80d8cf984abe661964e1b9d3fa9a9

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -2285,6 +2285,7 @@ pub struct TerminalViewRenderContext {
     pub link_tool_tip: Option<GridHighlightedLink>,
     pub is_terminal_focused: bool,
     pub is_terminal_selecting: bool,
+    pub is_extending_block_text_selection: bool,
     pub is_context_menu_open: bool,
     pub is_waterfall_gap_mode: bool,
     pub pane_state: SplitPaneState,
@@ -2479,6 +2480,8 @@ pub struct TerminalView {
 
     /// Whether there is an active text selection.
     is_selecting: bool,
+    /// Whether the active block-list text selection gesture began from Shift+Click Extend.
+    is_extending_block_text_selection: bool,
 
     context_menu: ViewHandle<Menu<TerminalAction>>,
 
@@ -4254,6 +4257,7 @@ impl TerminalView {
             alt_screen_scroll_top: Lines::zero(),
             horizontal_clipped_scroll_state: Default::default(),
             is_selecting: false,
+            is_extending_block_text_selection: false,
             context_menu_state: None,
             context_menu,
             hovered_secret: None,
@@ -6623,6 +6627,7 @@ impl TerminalView {
                         // selection visuals don't persist after switching selection focus to the
                         // CLI subagent view.
                         me.is_selecting = false;
+                        me.is_extending_block_text_selection = false;
                         me.block_text_selection_start_position = None;
                         me.clear_selected_text_except(Some(view.id()), ctx);
                         ctx.notify();
@@ -17789,6 +17794,9 @@ impl TerminalView {
             SelectAction::Update {
                 point, side, delta, ..
             } => self.update_alt_selection(*point, *side, delta, ctx),
+            SelectAction::Extend { point, side, .. } => {
+                self.extend_alt_selection(*point, *side, ctx)
+            }
             SelectAction::End => {
                 self.end_alt_selection(ctx);
             }
@@ -17813,6 +17821,7 @@ impl TerminalView {
             .alt_screen_mut()
             .start_selection(point, selection_type, side);
         self.is_selecting = true;
+        self.is_extending_block_text_selection = false;
 
         ctx.notify();
     }
@@ -17831,9 +17840,26 @@ impl TerminalView {
         ctx.notify();
     }
 
+    fn extend_alt_selection(&mut self, point: Point, side: Side, ctx: &mut ViewContext<Self>) {
+        if self.model.lock().alt_screen().selection().is_none() {
+            // Extend is only dispatched for an existing selection; direct action invocations
+            // without one are ignored.
+            return;
+        }
+
+        self.model
+            .lock()
+            .alt_screen_mut()
+            .extend_selection(point, side);
+        self.is_selecting = true;
+        self.is_extending_block_text_selection = false;
+        ctx.notify();
+    }
+
     fn end_alt_selection(&mut self, ctx: &mut ViewContext<Self>) {
         if self.is_selecting {
             self.is_selecting = false;
+            self.is_extending_block_text_selection = false;
             self.maybe_copy_selection_to_clipboard(ctx);
             ctx.notify();
         } else {
@@ -17844,6 +17870,7 @@ impl TerminalView {
     fn end_text_selection(&mut self, ctx: &mut ViewContext<Self>) {
         if self.is_selecting {
             self.is_selecting = false;
+            self.is_extending_block_text_selection = false;
             self.block_text_selection_start_position = None;
 
             let selected_text = {
@@ -18151,6 +18178,11 @@ impl TerminalView {
                 delta,
                 position,
             } => self.update_block_text_selection(*point, *side, *delta, *position, ctx),
+            BlockTextSelectAction::Extend {
+                point,
+                side,
+                position,
+            } => self.extend_block_text_selection(*point, *side, *position, ctx),
             BlockTextSelectAction::End => {
                 self.end_text_selection(ctx);
             }
@@ -18560,6 +18592,7 @@ impl TerminalView {
             .block_list_mut()
             .start_selection(point, selection_type, side);
         self.is_selecting = true;
+        self.is_extending_block_text_selection = false;
 
         if self.rich_content_views.is_empty() {
             ctx.notify();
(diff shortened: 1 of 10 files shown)


cola-runner:
@cla-bot check

cola-runner:
Updated the Testing section with macOS GUI verification for the cross-screen workflow, and linked the reviewed evidence:

https://gist.github.com/cola-runner/37a80d8cf984abe661964e1b9d3fa9a9

/oz-review

app/src/terminal/view.rs
```diff
@@ -18854,6 +18866,41 @@ impl TerminalView {
         ctx.notify();
     }
 
+    /// Extends the existing block-text selection to `point`, keeping the
+    /// original anchor (head) fixed and moving only the tail. This backs
```
seemeroland: I don't think this is the correct behavior, looking at what iTerm2 does. It may move either the head or the tail. If the shift-clicked point is before the head, it extends the head.

Also if the shift-clicked point is within the selection, it may update the head if the point is closer to the head (shrink the head vs shrink the tail, depending which side the point is closer to). Using the row is probably a fine approximation.

app/Cargo.toml
```diff
@@ -837,6 +837,7 @@ traces = []
 viewing_shared_sessions = []
 workflow_aliases = []
 rect_selection = []
+shift_click_extend_selection = []
```
seemeroland: nit: it should be okay to not add the feature flag. We're going to have to enable it in all channels anyway

app/src/terminal/view.rs
```diff
@@ -17988,6 +17988,13 @@ impl TerminalView {
             SelectAction::Update {
                 point, side, delta, ..
             } => self.update_alt_selection(*point, *side, delta, ctx),
+            SelectAction::Extend { point, side, .. } => {
+                // Shift+click extension is only wired up for the block list
+                // today; for the alt screen we treat it as a plain tail update
+                // so the variant is handled exhaustively and stays correct if
+                // alt-screen extension is enabled later.
+                self.update_alt_selection(*point, *side, &Lines::zero(), ctx);
```
seemeroland: It'd be great to add alt screen support as well (e.g. in vim). They should be able to share some of the logic. 

We'd probably want to add this feature to AIBlocks as well but that is implemented more differently so no need to worry about it in this PR.

app/src/terminal/view.rs
```diff
@@ -18854,6 +18866,41 @@ impl TerminalView {
...
+    fn extend_block_text_selection(
+        &mut self,
+        point: BlockListPoint,
+        side: Side,
+        position: Vector2F,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        let has_selection = self.model.lock().block_list().selection().is_some();
+        if !has_selection {
+            self.begin_block_text_selection(point, side, SelectionType::Simple, position, ctx);
```
seemeroland: Probably worth a comment that this is unexpected because extending only makes sense with a selection, and this is a fallback to treat it like a regular click. Alternatively I think it might be fine to do nothing in this case

seemeroland (commented):
Thanks for taking a stab at this!

cola-runner:
Updated this PR on top of latest `master` (`a30cc7a3`) and addressed the review feedback:

- changed Shift+Click extension to move the nearest selection boundary, matching iTerm2 behavior for extending before/after the selection and shrinking from inside it
- shared the nearest-boundary logic between block-list and alt-screen selections
- wired Shift+Click extension for alt screen selections
- removed the compile-time feature flag from the final diff
- changed the no-selection `Extend` fallback to no-op with a comment, since Extend should only be dispatched when a selection exists
- kept Shift+Click+drag as a text-selection update path so the extended selection can still be fine-tuned while Shift is held

Validation on latest `master` plus this PR:

- `cargo test -p warp nearest_boundary --lib` (6 passed)
- `cargo check -p warp`

cola-runner:
Added another small test-only commit for edge coverage:

- same-row shrinking uses column distance to pick the nearest boundary
- reversed block-list selections update the correct stored anchor/tail
- block-list and alt-screen `Extend` no-op when no selection exists

Validation after the test commit:

- `cargo test -p warp nearest_boundary --lib` (10 passed)
- `cargo test -p warp test_alt_screen_extend_selection_without_existing_selection_noops --lib` (1 passed)
- `cargo check -p warp`

cola-runner:
Hi @seemeroland, I addressed your review feedback in the latest updates:

- Shift+Click now moves the nearest selection boundary, including shrinking from inside the current selection.
- Block-list and alt-screen selections share the nearest-boundary logic.
- Alt-screen Shift+Click extension is wired.
- The compile-time feature flag was removed.
- `Extend` without an existing selection now no-ops.
- Shift+Click+drag can continue fine-tuning the extended text selection while Shift is held.
- Added edge tests for same-row shrink behavior, reversed block-list selections, and no-selection no-op behavior.

Validation:

- `cargo test -p warp nearest_boundary --lib` (10 passed)
- `cargo test -p warp test_alt_screen_extend_selection_without_existing_selection_noops --lib` (1 passed)
- `cargo check -p warp`
- GitHub CI is green on the current head (`352260410bda3d935c2e7c295555657e12923d20`).

Would you mind taking another look when you have a chance?

cola-runner:
Hi @seemeroland, following up here. I rebased this PR onto latest `master` (`f554cf76`) and re-ran the local validation:

- `cargo test -p warp nearest_boundary --lib` (10 passed)
- `cargo test -p warp test_alt_screen_extend_selection_without_existing_selection_noops --lib` (1 passed)
- `cargo check -p warp`

All previous review threads are resolved, and the current head is `8ccfbbecc97f3ac907b78b45de5d0853148b3644`.

cc @harryalbert @adrawbond-alto @BobbyWang0120 since you have recently touched nearby terminal view / mouse dispatch / terminal event paths. Would one of you be able to take a look or help route this to the right owner?

cola-runner:
Hi @seemeroland, following up after this PR went stale and became conflicting.

I rebased it onto latest `master` (`c20645b9`) and pushed the updated head `5b6e4f43d453eb4a3dff99f781879dfec1e00d13`. GitHub now reports the PR as mergeable again.

Local validation after the rebase:

- `cargo fmt`
- `cargo test -p warp nearest_boundary --lib` (10 passed)
- `cargo test -p warp test_alt_screen_extend_selection_without_existing_selection_noops --lib` (1 passed)
- `cargo check -p warp`

cc @harryalbert @adrawbond-alto @BobbyWang0120 in case one of you can help review or route this to the right owner.

cola-runner:
Hi @seemeroland, one more update before review: I did another self-review pass and found a boundary-wiring issue around Shift+Click+drag.

Fixed in `2eba005927ef47209e597d2abd220bc3a60e7f48`:

- plain block-list drag updates now keep using the drag endpoint instead of re-running nearest-boundary selection
- Shift+Click Extend now moves the nearest boundary and keeps that moved boundary as the active update endpoint, so Shift+Click+drag continues fine-tuning the same side
- added model/view regression coverage for the block-list and alt-screen cases

Validation after the fix:

- `cargo fmt`
- `cargo test -p warp nearest_boundary --lib` (13 passed)
- `cargo test -p warp test_alt_screen_extend_selection_without_existing_selection_noops --lib` (1 passed)
- `cargo check -p warp`

PR #11986, opened 2026-06-01
Gate per-prompt node version detection on the Node.js chip

szgupta:
## Description
The shell `warp_precmd` hook spawned `node --version` (plus `dirname` subprocesses for directory walks) on **every** prompt whenever the user was in a Node.js project inside a git repo. This added noticeable latency between commands. Worse, the work was not tied to whether the **Node.js Version** context chip was actually shown, so removing the chip from the prompt did not stop it — the shell kept computing `node_version` on every precmd.

**What/Why/How:**
- **What:** Stop running per-prompt `node --version` detection when it isn't needed, and make it cheap when it is.
- **Why:** Eliminate per-command latency caused by spawning the Node interpreter (and `dirname`) on every prompt. This directly contradicted the hook's own warning to avoid external commands in performance-sensitive code.
- **How:**
  - Gate the detection behind a new `WARP_PROMPT_NODE_VERSION_ENABLED` env var. The bootstrap treats any value other than `"0"` as enabled, so detection is only skipped when explicitly disabled (paths that don't set it, e.g. remote sessions, keep prior behavior).
  - The client sets the env var at PTY spawn based on whether `ContextChipKind::NodeVersion` is enabled in the Warp prompt, the agent footer, or the CLI agent footer (the chip value is fed by the same precmd payload wherever it's displayed).
  - Cache the resolved version in shell globals keyed on `PWD` + `PATH`, so `node --version` only re-runs when the directory or `PATH` changes (`PATH` changes on `nvm use`, so the chip still updates correctly). In steady state this is a string comparison with zero `node` spawns.
  - Replace the `dirname`-in-a-loop directory walks with fork-free path stripping (`${dir%/*}` in zsh/bash, `string replace` in fish; PowerShell already used `.Parent`).
  - Threaded a `node_version_chip_enabled` flag through `PtyOptions` and the unix/docker/Windows spawn paths, including `WSLENV` forwarding.

Applies to all four shells: zsh, bash, fish, and pwsh.

Conversation: https://staging.warp.dev/conversation/7b589f92-1c67-4956-967b-45cb9bb774db
Plan: https://staging.warp.dev/drive/notebook/HKwsjkZDAOYIWDpsADg2I2

## Linked Issue
No linked issue (performance fix surfaced during investigation).

## Testing
- `cargo check -p warp --lib` passes.
- `cargo fmt` produced no changes beyond the edited files.
- `cargo clippy -p warp --lib --tests` passes with no warnings.
- Affected unit tests pass (`terminal::local_tty::unix::tests::*history_size_sentinels`), updated for the new spawn-command parameter.
- Parse-checked the edited bootstrap scripts: `zsh -n`, `bash -n`, and `fish --no-execute` all pass (PowerShell unavailable on macOS; change mirrors the existing structure).

### Screenshots / Videos
N/A — no UI changes.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed latency between commands caused by the Node.js version prompt chip running `node --version` on every prompt; detection is now skipped when the chip is disabled and cached otherwise.


szgupta:
I manually tested by running this locally:
- if i removed the nvm chip from my prompt, it wouldn't run this nvm version check
- even with the chip in my prompt, it's much faster now that it's cached

acarl005 (approved):
The bot's comment about already-running sessions not updating is valid. This means that when you enable the node version chip, it only affects newly opened sessions. The ones you already have open still won't have the node chip.

However, fixing that is complicated. We handle it for the "honor custom PS1" setting when that changes. That involves setting custom bindings with the shell's line editor. That isn't scalable.

Testing this locally on my ASUS zenbook, it actually produces a perceptible improvement. I therefore would advocate for shipping this despite the bot's comments.

szgupta:
I should have responded to the bot's comments, but I think (1) it's not unreasonable that the setting only affects new sessions (maybe I could add something in the settings to make that clearer, but doesn't seem worth it right now), (2) even when using the node chip, it should be much faster because of the caching.

PR #12028, opened 2026-06-01
Add support for authenticating against Grok subscriptions. 

szgupta:
Demo: https://www.loom.com/share/72127676984c468bb16900500f5d56f7

## Description
Makes a connected xAI/Grok (e.g. SuperGrok) subscription a first-class, per-request provider credential on the client, mirroring the existing BYO-key handling for Anthropic/OpenAI/etc. When a user connects via `/grok`, the OAuth tokens are persisted and the access token is injected into every agent request so the server's xAI harness can authenticate with the user's subscription instead of Warp's org-wide xAI key.

**What / How**
- `crates/ai` `api_keys`: new `GrokTokens { access_token, refresh_token, expires_at }` held on `ApiKeyManager` and persisted to secure storage under a dedicated `GrokOAuthTokens` key (loaded on init; deferred write/remove mirroring the existing keychain pattern). `api_keys_for_request` now sets the new `grok_oauth_access_token` proto field, gated on the same BYO-enable path as the other keys and only when the token is present and not within a small expiry skew.
- `app` `grok_oauth`: added `refresh_access_token` (OAuth `refresh_token` grant) + a shared token-POST helper.
- `app` `grok_subscription` (new): `GrokTokenRefresher` trait on `ApiKeyManager` (mirrors `AwsCredentialRefresher`) that converts `TokenResponse` → `GrokTokens`, proactively refreshes shortly before expiry, and reschedules itself. Carries over the previous refresh token when xAI omits a new one.
- `/grok` connect flow now persists tokens via `store_grok_tokens` (previously only logged metadata); startup resumes the proactive refresh loop (`GrokOauth`-gated, non-wasm).

## Testing
- [x] `cargo check -p warp --lib`
- [x] `cargo nextest run -p warp grok` (9 passed) and `cargo nextest run -p ai grok` (9 passed)
- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy -p ai -p warp --all-targets --tests -- -D warnings` clean (scoped to the changed crates due to sandbox limits)

Unit tests cover `GrokTokens` validity/expiry/`needs_refresh`, the request injection (`api_keys_for_request`), serde round-trip, and `TokenResponse` → `GrokTokens` conversion (expiry derivation + refresh-token carry-over). End-to-end OAuth/refresh against xAI was not exercised in this sandbox.

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/1ec5c9cd-a79e-4cc7-bb3d-dbcb95e0c305_
_Run: https://oz.staging.warp.dev/runs/019e8523-3b2f-765e-900a-608134f9c7fe_


crates/ai/src/grok_subscription/mod.rs
vorporeal, line 1: could this live in the `ai` crate?  would looooooove for us to not add new code to `warp` when we don't need to.

app/src/terminal/input/slash_commands/mod.rs
```diff
@@ -356,6 +356,74 @@ impl Input {
...
+        };
+
+        // Open xAI's consent screen in the user's default browser.
+        ctx.open_url(&attempt.authorize_url());
+
+        let window_id = ctx.window_id();
+        ToastStack::handle(ctx).update(ctx, |toast_stack, ctx| {
+            toast_stack.add_ephemeral_toast(
+                DismissibleToast::default(
+                    "Opening your browser to connect your Grok subscription…".to_owned(),
```
IsaiahWitzke: thoughts on including the `attempt.authorize_url()` in the toast too?

i.e. like our login flow's "browser didn't open? <a>copy url</a>"

IsaiahWitzke: AM research:

Yes — the toast already has built-in affordance hooks, so no changes to `dismissible_toast.rs` rendering are needed. `DismissibleToast<A>` exposes three ways to attach interactivity (`app/src/view_components/dismissible_toast.rs`):

1. **`with_link(ToastLink<A>)`** (`:356`) — renders a bold link on the right. `ToastLink::with_onclick_action(A)` (`:256`) dispatches a typed action on click, and `with_keystroke` (`:261`) can even show a shortcut. ← best fit for a "Copy link" affordance.
2. **`with_action_button(ViewHandle<ActionButton>)`** (`:367`) — a full button; more plumbing.
3. **`with_on_body_click(...)`** (`:374`) — whole-body click, but it **dismisses** the toast on click (the `ClickBody` handler removes it at `:213`), so it's wrong for copy.

## Recommended approach (the link)
The Grok toast is a `DismissibleToast<WorkspaceAction>` (it goes through the workspace `ToastStack`, `app/src/workspace/toast_stack.rs:26`), so the link's action type is `WorkspaceAction`. Wiring:
- Add a `WorkspaceAction` variant, e.g. `CopyToClipboard(String)`, and handle it in the Workspace action handler with the same clipboard call the login flow uses (`app/src/auth/auth_view_body.rs:897`):
```rust
ctx.clipboard().write(ClipboardContent {
    plain_text: url.clone(),
    paths: Some(vec![url]),
    ..Default::default()
});
```
- Build the toast with the link:
```rust
DismissibleToast::default("Opening your browser to connect your Grok subscription…".to_owned())
    .with_link(
        ToastLink::new("Copy link".to_owned())
            .with_onclick_action(WorkspaceAction::CopyToClipboard(attempt.authorize_url())),
    )
```
## Two caveats worth deciding on first
1. **The toast lives only ~4 seconds.** The workspace ephemeral stack is `DismissibleToastStack::new(Duration::from_secs(4))` (`app/src/workspace/view.rs:3070`). A "Copy link" the user has 4s to click is a weak fallback. Options:
   - Use `add_persistent_toast` instead (the `ToastStack` model already exposes it, `toast_stack.rs:34`) so it stays until dismissed, **or**
   - Skip the click entirely: copy the URL to the clipboard up front and just reword the toast to *"Opening your browser… (login link copied to clipboard)"*. Zero timing pressure, no new action plumbing — arguably the better UX for a fallback.
2. **`WorkspaceAction::CopyToClipboard` is reusable**, but check there isn't already a generic copy action you can reuse before adding one.

My lean: option 2 (auto-copy + mention in the toast text) is simpler and more robust than a 4-second-window link. But if you want the explicit affordance, the link path is clean.

Want me to implement one of these? If so, which — the auto-copy-and-mention version, the "Copy link" action link, or a persistent toast with the link?

IsaiahWitzke (approved):
ooc: what does the UX look like when the refresh token is expired?

i see:

```
    // Leave the existing (soon-to-expire) token in place; the server
    // remains the authority and will reject it if it's truly invalid.
```

will there be a follow-up PR to react to invalid grok creds? Want to make sure the user sees they need to do `/grok` again if their refresh token is messed up / out of date

app/src/ai/agent/api/convert_conversation.rs
```diff
@@ -1763,6 +1765,9 @@ fn create_cancelled_result_for_tool_call(
             return None;
         }
         ToolType::Subagent(_) => return None,
+        // WaitForEvents is a server-side orchestration tool with no local
+        // result representation.
+        ToolType::WaitForEvents(_) => return None,
```
szgupta: fyi: @cephalonaut this got pulled in

szgupta:
Gonna follow up on token expiry. There is a mechanism for automatically refreshing but upon closer look, there are cases where you can get stuck in a bad state. In that case, users would at least get an error and then could disconnect+reconnect their sub.

PR #12036, opened 2026-06-01
Add proper support for host scoped requests

kevinyang372:
## Description

This PR adds first-class support for **host-scoped requests** in the remote-server stack and migrates all existing callers onto it.

### Why we need explicit host-scoped requests

A single remote host (one `remote-server` daemon) can be reached through **multiple proxy connections** at once — one per SSH session / Warp tab connected to that host. Most request/response traffic is *session-scoped*: the response must come back on the exact connection that sent it. But a class of operations are logically tied to the **host**, not the session that happens to be issuing them: writing/reading/deleting files, saving/opening buffers, discarding changes, codebase indexing, fragment-metadata lookups, handoff snapshot uploads, etc.

Previously these were modeled as ordinary session-scoped client requests. That created two problems:

1. **No failover.** If the session that issued the request disconnected while the request was in flight, the response was lost and the caller hung until it timed out, even though a sibling connection to the same host was still alive and could have carried the work/response.
2. **Lifecycle owned by the wrong layer.** `RemoteServerClient` (a single connection) owned the request lifecycle, so it had no way to retry on, or accept a response from, a *different* connection.

Host-scoped requests fix this by moving lifecycle ownership up to `RemoteServerManager` (which sees *all* connections for a host) and letting the daemon deliver a response on **any** open connection to that host.

### How it works — client and server

**Client / proxy side (`crates/remote_server`):**
- `RemoteServerManager` owns `pending_host_requests` (keyed by `RequestId`). `send_host_request` dispatches the message through *some* connected session for the host (`dispatch_host_scoped_request`) **before** registering the pending entry, so a dispatch that fails on a dead outbound channel fails the caller immediately instead of leaking a pending entry.
- Each pending request records the `dispatched_session_id` and the original message so it can be **retried on a sibling connection**. If the writer task fails to flush a host-scoped message before it reaches the daemon, it emits `ClientEvent::HostScopedWriteFailed`; the manager (`handle_host_scoped_write_failed`) re-dispatches the same `RequestId` on another connected session, or fails fast with `AllSessionsDisconnected` if none remain — no waiting for the timeout.
- `RemoteServerClient::send_host_scoped` returns `Result` (so dispatch failures are visible), and unmatched responses are forwarded on a dedicated `host_response_tx` channel that the manager drains to match against `pending_host_requests`.
- A per-request timeout is armed via an `AbortHandle` stored on the pending entry and cancelled the moment the request resolves, so in-flight (not cumulative) requests bound the number of live timer tasks.
- When all sessions for a host disconnect, `fail_pending_host_requests_for_disconnected_hosts` fails the remaining pending requests.

**Daemon side (`app/src/remote_server/server_model.rs`):**
- The daemon tracks in-flight host-scoped requests in `host_scoped_requests` (`RequestId → originating ConnectionId`).
- `send_server_message` delivers the response to the originating connection, but for host-scoped responses it **fails over to any other open connection** — both when the original connection's sender is gone *and* when `try_send` fails on it (`send_host_scoped_response_via_alternate_connection`). Non-host-scoped responses are never re-routed.

### Callsites that needed migration

**New manager API (entry points):**
- `RemoteServerManager::send_host_request` — core dispatcher (caller controls the `RequestId`).
- `RemoteServerManager::send_host_scoped_request` — convenience wrapper that mints a `RequestId`.
- `RemoteServerManager::host_request_handle` — returns a `HostRequestHandle` for pure-async callers that don't hold `&mut RemoteServerManager`.

**New `HostRequestHandle` typed methods** (so call sites never build `host_scoped_request::Message` wrappers or match on `server_message::Message`): `write_file`, `delete_file`, `save_buffer`, `open_buffer`, `read_file_context`, `get_fragment_metadata_from_hash`, `upload_handoff_snapshot`, plus the generic `send`.

**Migrated callers:**
- `app/src/ai/blocklist/action_model/execute/read_files.rs` — remote file reads now go through `HostRequestHandle::read_file_context`.
- `app/src/code/global_buffer_model.rs` — buffer save via the host request path.
- `crates/warp_files/src/lib.rs` — file save/delete via the host request path (and removed the now-dead `resolve_remote_client` helper).
- `crates/remote_server/src/manager.rs` — `discard_files` and codebase-index operations dispatched as host-scoped requests.
- Response parsing centralized in `crates/remote_server/src/host_response.rs`.

## Testing

- [x] `cargo check -p warp --lib --tests` and `cargo test -p remote_server` pass (80 tests).
- [x] Added daemon failover unit tests in `app/src/remote_server/server_model_tests.rs`: response fails over when the target send fails, when the target connection is gone, and confirms non-host-scoped responses are *not* re-routed.

## Agent Mode


kevinyang372:
* **#12036** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12036?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12036?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/remote_server/src/host_response.rs
```diff
@@ -0,0 +1,77 @@
...
+        other => Err(unexpected_variant("DeleteFile", other)),
+    }
+}
+
+/// Interprets a [`ServerMessage`] as the result of a `DiscardFiles` request.
+///
+/// Unlike the file-operation responses above, an empty `result` is treated as
+/// an error: the daemon always populates a `success`/`error` variant, so a
+/// missing one indicates a malformed response rather than a benign default.
+pub fn discard_files_result(msg: &ServerMessage) -> Result<(), String> {
```
moirahuang, line 75: why does `DiscardFiles` have different semantics than `DeleteFile` where missing = malformed response?

app/src/remote_server/server_model.rs
```diff
@@ -1208,21 +1278,44 @@ impl ServerModel {
...
+    /// request is still in flight.
     fn send_server_message(
-        &self,
+        &mut self,
         conn_id: Option<ConnectionId>,
         request_id: Option<&RequestId>,
         message: server_message::Message,
     ) {
+        // Check if this is a host-scoped request eligible for failover
+        // delivery BEFORE removing from tracking.
```
moirahuang: sorry i'm a little confused why we remove from tracking in this case?

crates/remote_server/src/host_response.rs
moirahuang, line 1: are there any guards we can add to make sure that newly added variants are properly operations? this is correctly for current variants but what if we add a new one that needs to be handled in host_response.rs here?

moirahuang (approved):
approving to unblock but i also locally asked an agent to review, here's the responses:

```
{
  "verdict": "REJECT",
  "body": "## Overview\nThis PR introduces host-scoped request envelopes and manager-owned request routing so operations can survive loss of the originating proxy connection.\n\n## Concerns\n- `GetDiffState` and `OpenBuffer` still establish daemon-side state keyed to the originating connection, so rerouting only their response cannot preserve their subscription semantics after that connection disconnects.\n- Retrying a host request after an ambiguous transport write failure can run non-idempotent operations twice because the daemon has no request-ID deduplication or replay cache.\n- Timed-out/aborted host requests are removed from the manager but remain in the daemon's `host_scoped_requests` map while another connection stays alive.\n- A malformed host-scoped response with a parseable request ID is no longer delivered as an immediate protocol error; it is left pending until the 120-second manager timeout.\n- This changes user-visible remote-host reliability and buffer/diff behavior, but the PR description and discussion contain no screenshot or recording demonstrating end-to-end failover behavior. Please include screenshots or a screen recording demonstrating it working end to end.\n\n## Verdict\nFound: 2 critical, 2 important, 1 suggestions\n\n**Request changes**",
  "comments": [
    {
      "path": "app/src/remote_server/server_model.rs",
      "line": 722,
      "side": "RIGHT",
      "body": "🚨 [CRITICAL] `GetDiffState` (and `OpenBuffer` above) cannot be made host-scoped just by rerouting its response: each handler registers `conn_id` as a subscription, and `deregister_connection` removes that pending/subscribed state when the originating session drops. The failed-over caller can therefore time out or open a buffer without receiving subsequent updates. Keep subscription-creating operations session-scoped, or explicitly transfer/reissue the subscription on the failover connection."
    },
    {
      "path": "crates/remote_server/src/manager.rs",
      "line": 1285,
      "side": "RIGHT",
      "body": "🚨 [CRITICAL] A transport write error does not prove the daemon never accepted this request, yet this retries it while the daemon has no request-ID deduplication. Retrying side-effecting operations such as `DiscardFiles`, `ResolveConflict`, or snapshot upload can execute them twice. Add daemon-side idempotency/replay handling before retrying, or do not automatically retry operations that are not safely repeatable."
    },
    {
      "path": "app/src/remote_server/server_model.rs",
      "line": 831,
      "side": "RIGHT",
      "body": "⚠️ [IMPORTANT] This tracking entry is removed only when a response is sent or all connections vanish. The new manager timeout sends `Abort` while sibling connections may stay alive, but `handle_abort` never removes `host_scoped_requests`, so every timed-out/cancelled host request leaks in the long-lived daemon. Remove the host-scoped entry when aborting the target request."
    },
    {
      "path": "crates/remote_server/src/client/mod.rs",
      "line": 846,
      "side": "RIGHT",
      "body": "⚠️ [IMPORTANT] Host responses are forwarded to the manager only after successful decoding. In the existing `Decode(_, Some(request_id))` path, a host request is absent from this client's session-scoped map and is only logged, leaving the manager request pending until its 120-second timeout. Forward a request-keyed protocol failure to the manager so host-scoped callers fail promptly."
    }
  ]
}

```

PR #11559, opened 2026-05-22
Move local and remote project skills to use RepoMetadataModel

moirahuang:
## Description

Migrates local project skills away from `ProjectSkillSubscriber` in the normal indexed-repository path and instead discovers project skill files through `RepoMetadataModel`. Gitignored skills subtrees are deliberately added to `RepoMetadataModel`

If local repository metadata indexing fails (for example, when a repository is too large to index), the existing local project-skill watcher fallback remains available.

## Testing

Added unit tests + locally tested

### Screenshots / Videos

Regular skills
https://github.com/user-attachments/assets/ce2f6abf-ecf6-4f77-8848-313b41b0a3d7

## Agent Mode

- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

Co-Authored-By: Oz [oz-agent@warp.dev](mailto:oz-agent@warp.dev)

PR #12040, opened 2026-06-02
Avoid project skill refresh scans for unrelated repo updates

moirahuang:
## Description

In the PR https://github.com/warpdotdev/warp/pull/11559/changes, we introduced triggering `refresh_project_skills_for_repo()` on every `FileTreeEntryUpdated`. This PR attempts to address that by having `FileTreeEntryUpdated` only trigger `refresh_project_skills_for_repo()` if there is a skills relevant update by passing through the exact updates.

Tested by locally creating a repository with approximately 90,000 files and on dev, unrelated file edits entered the expensive project-skill refresh path on the main thread. With this change, identical unrelated edits no longer invoke project-skill refresh traversal, while edits to `.agents/skills/repro/SKILL.md` still refresh correctly.

I'm purposefully not entirely moving the full tree traversal off the main thread. This already existed before the PR, the regression is specifically that we're triggering the scan too often on unnecessary file changes. It'd still be good to address the full tree traversal on the main thread separately, specifically for the large repo case.

## Testing
Manual regression validation:
- On dev: making an unrelated edit captured `refresh_project_skills_for_repo` at approximately 1146 ms and `find_symlinked_skill_files_in_tree` at approximately 695 ms on the main thread.
- Positive control: editing `/tmp/warp-skill-refresh-repro-7reije9h/.agents/skills/repro/SKILL.md` in the patched build still triggered `SkillWatcher::refresh_project_skills_for_repo` (approximately 1600 ms) through `find_project_skill_files_in_tree`.
- Regression check: sampling the patched build for 45 seconds during 30 edits to `/tmp/warp-skill-refresh-repro-7reije9h/packages/pkg-000/src/unrelated.txt` captured zero matches for `SkillWatcher`, `refresh_project_skills_for_repo`, `find_project_skill_files_in_tree`, `find_skill_files_in_tree`, `find_symlinked_skill_files_in_tree`, `get_repo_contents`, or `collect_contents_recursive`.

- [x] I have manually tested my changes locally with a `WarpLocal` build launched via `wrun`.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed project skill refreshes causing UI stalls when unrelated files change in large repositories.


crates/repo_metadata/src/local_model.rs
```diff
@@ -75,6 +75,9 @@ pub enum RepositoryMetadataEvent {
     /// The file tree's [`Entry`] was updated.
     FileTreeEntryUpdated {
         path: StandardizedPath,
+        /// Applied incremental metadata when this update is represented as a delta.
+        /// `None` indicates a conservative lazy-load or opaque entry update.
+        update: Option<RepoMetadataUpdate>,
```
kevinyang372: I think we should have an enum to make this extra clear. Something like:
```
enum MetadataUpdateType {
     FullReplace,
     IncrementalUpdate(RepoMetadataUpdate)
}
```

app/src/ai/skills/file_watchers/utils.rs
```diff
@@ -27,6 +29,44 @@ fn local_or_remote_path_for_repo_path(
...
+        let removed_path = local_or_remote_path_for_repo_path(repo_id, removed_path);
+        known_skill_files
+            .into_iter()
+            .flatten()
+            .any(|known_skill_file| known_skill_file.starts_with(&removed_path))
+    }) {
+        return true;
+    }
+
+    let is_local_repo = matches!(repo_id, RepositoryIdentifier::Local(_));
```
kevinyang372, line 51: Hmmm why are we filtering by local repo here?

app/src/ai/skills/file_watchers/utils.rs
```diff
@@ -27,6 +29,44 @@ fn local_or_remote_path_for_repo_path(
...
+    update.update_entries.iter().any(|entry_update| {
+        if is_local_repo
+            && is_project_provider_path(&entry_update.parent_path_to_replace.to_local_path_lossy())
+        {
+            return true;
+        }
+
```
kevinyang372, line 58: Do we have to always query the entire update subtree? Is there any pre-filtering we could apply here that allows us to early return?

kevinyang372 (approved):
Approving to unblock

app/src/ai/skills/file_watchers/utils.rs
moirahuang: we currently only support symlinking in the local path. it'll be possible after https://github.com/warpdotdev/warp/pull/12047, i can make a follow-up PR to refactor here once that PR is merged

app/src/ai/skills/file_watchers/utils.rs
moirahuang: i can't really think of a good pre-filter solution here because skills can be anywhere in the subtree?

crates/repo_metadata/src/local_model.rs
moirahuang: added

PR #12053, opened 2026-06-02
Perf/ async‑find query‑refinement fast‑path

### LINKED ISSUE #12037: perf: async‑find query‑refinement fast‑path
### Pre-submit Checks

## Summary

Optimize async terminal find when the user refines an existing query by typing additional characters.

Today, async find detects strict query refinements, but still falls back to a full rescan of terminal content. For example, when a user searches for `foo` and then continues typing `foobar`, the implementation restarts a broad scan instead of narrowing the already-known result set.

## Problem

A strict query refinement has an important property:

- Every match for `foobar` must already have matched `foo`.

Because of that, a full rescan is unnecessary work. The current behavior wastes CPU and adds latency while typing in Cmd+F, especially with large scrollback histories.

## Desired Behavior

When the new query is a strict prefix refinement of the previous query, async find should avoid rescanning the whole buffer and instead:

1. Reuse prior results as the candidate set.
2. Skip blocks that had no matches for the previous query.
3. Rescan only the relevant rows within blocks that previously matched.
4. Complete immediately when the previous query produced zero matches.

The behavior should preserve existing correctness for:

- case-sensitive vs case-insensitive search
- regex-disabled refinement handling
- focused match ordering and navigation
- AI/rich-content block integration
- dirty-range updates and scrollback truncation

## Scope

This is a targeted performance improvement in the async-find refinement path, centered around `filter_results_for_refinement` in `app/src/terminal/find/model/async_find.rs`.

The intended approach is consistent with maintainer guidance in this issue thread:

- naive improvement: enqueue only blocks that previously contained matches
- stronger improvement: enqueue only the relevant dirty rows within those blocks

## Validation

Acceptance should include tests that show:

- refinement searches return the same matches as a full rescan
- blocks with no prior matches are skipped
- zero prior matches complete immediately
- refinement is only used once the prior search is complete
- non-refinement edits still use the normal full-scan path

A benchmark demonstrating lower latency for incremental query typing on realistic large terminal histories would also be useful.

## Related Work

- Original implementation PR: #12053
- Spec PR: #12192

KRRT7:
Closes #12037

The async‑find implementation is designed to run a full‑scan of every terminal block every time the user types a new search query.

This works fine for unrelated queries, but it is highly inefficient when the new query is merely a refinement of the previous one (e.g., the user types foo and then continues typing foobar).

A refinement is a strict prefix: any line that matches foobar must have already matched foo.
Because of that property we can avoid re‑scanning the whole buffer and instead:

Reuse the matches we already have – only rows that previously contained a match need to be examined again.
Skip blocks with no prior matches – if a block never contributed a match for foo, it cannot possibly contain a match for foobar.
Finish instantly when the previous result set was empty – there is nothing to rescan


vkodithala:
Hey @KRRT7, really appreciate the contribution. You're killing it! Per our contribution guidelines [here](https://github.com/warpdotdev/warp/blob/master/CONTRIBUTING.md#prs-opened-without-a-linked-issue), make sure you link PRs like this to issues, via "Development" in the sidecar, or manually in the PR description. Go ahead and re-run "/oz-review" once you've done this! It should be tagged to me, and I'll give it a review once it passes Oz's checks.

Lemme know if our contribution guidelines can be made clearer here :)

PR #12072, opened 2026-06-02
Fix 3p agent prompt queueing bugs during Cloud Mode setup

harryalbert:
## Description

There were a few small bugs with prompt queueing for cloud mode setup that this PR fixes:

1. We don't support prompt queueing for 3p agents, so we shouldn't allow you to queue multiple prompts for 3p agents during cloud mode setup

2. Kind of unrelatedly (but I noticed this when testing this PR), during Cloud Mode setup, the cloud agent (sharer) runs environment setup commands and emits a `CommandExecutionStarted` event per command. The shared-session viewer handled every non-agent `CommandExecutionStarted` by calling `unfreeze_and_clear_agent_input`, which cleared the editor buffer — wiping any follow-up the user was composing after each setup command. The viewer now skips that clear while it is a cloud-agent viewer in the pre-first-exchange phase so that your input isn't wiped during setup

## Testing

https://www.loom.com/share/f7837d6db0a84753937b62fbf66369a8

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/452cc86b-c930-407a-8861-7952013c7b1f_
_Run: https://oz.staging.warp.dev/runs/019e88ba-800b-7d87-bf93-d1e5cefd3e97_

### DIFF
--- a/app/src/terminal/shared_session/viewer/terminal_manager.rs
+++ b/app/src/terminal/shared_session/viewer/terminal_manager.rs
@@ -1095,17 +1095,16 @@ impl TerminalManager {
                     // In cloud-mode startup (before the first exchange), shared-session input
                     // sync reflects environment setup commands. Skip applying remote edits so
                     // the visible input isn't populated with setup-command text.
-                    if FeatureFlag::CloudModeSetupV2.is_enabled()
-                        && {
-                            let model = view.model.lock();
-                            is_cloud_agent_pre_first_exchange(
-                                view.ambient_agent_view_model(),
-                                view.agent_view_controller(),
-                                &model,
-                                ctx,
-                            )
-                        }
-                    {
+                    let skip_during_setup = FeatureFlag::CloudModeSetupV2.is_enabled() && {
+                        let model = view.model.lock();
+                        is_cloud_agent_pre_first_exchange(
+                            view.ambient_agent_view_model(),
+                            view.agent_view_controller(),
+                            &model,
+                            ctx,
+                        )
+                    };
+                    if skip_during_setup {
                         return;
                     }
                     view.apply_viewer_shared_session_input_update(block_id, operations.clone(), ctx);
--- a/app/src/terminal/view/ambient_agent/mod.rs
+++ b/app/src/terminal/view/ambient_agent/mod.rs
@@ -139,15 +139,9 @@ pub fn create_cloud_mode_view(
     (terminal_view, terminal_manager)
 }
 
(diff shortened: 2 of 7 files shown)


harryalbert:
* **#12072** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12072?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12072?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/terminal/shared_session/viewer/terminal_manager.rs
```diff
@@ -1095,17 +1095,16 @@ impl TerminalManager {
...
-                            let model = view.model.lock();
-                            is_cloud_agent_pre_first_exchange(
-                                view.ambient_agent_view_model(),
-                                view.agent_view_controller(),
-                                &model,
-                                ctx,
-                            )
-                        }
-                    {
+                    let skip_during_setup = FeatureFlag::CloudModeSetupV2.is_enabled() && {
```
liliwilson, line 1098: Does this conflict with what the comment above is trying to do? Seems like this clause was added specifically for the pre-first-exchange case so just want to make sure

app/src/terminal/view/ambient_agent/mod.rs
```diff
@@ -160,7 +166,7 @@ pub fn is_cloud_agent_pre_first_exchange(
 
     let is_in_pre_first_exchange_status = matches!(
         view_model.status(),
-        Status::WaitingForSession { .. } | Status::AgentRunning
+        Status::Setup | Status::WaitingForSession { .. } | Status::AgentRunning
```
liliwilson: Why do we need to add `Setup` here? AFAICT this state is pre-agent-prompt-submission

liliwilson (approved):
Queue UI is 🔥 Silent demo is a performance art

app/src/terminal/shared_session/viewer/terminal_manager.rs
harryalbert: this is just a no-op refactor (honestly could probably revert). Was just extracting this into a variable so that it's easier to understand what we're checking/why

app/src/terminal/view/ambient_agent/mod.rs
harryalbert: hmm, I actually don't think we do, good call

PR #12074, opened 2026-06-02
Fix crash when trying to use the microphone on macOS.

vorporeal:
## Description

The `authorizationStatusForMediaType` API returns an unsigned 64-bit integer, but our code was storing the result as a `i32`.  When we moved from using the `objc` crate to the `objc2` crate, I think the new crate performs more careful checking of these types, and started to produce a runtime error as a result of the type mismatch.

The `objc2` crate family includes `objc2-av-foundation`, which includes concrete APIs for all of the code here.  I've moved over to using that, improving safety and clarity.

## Testing

- [x] Manually tested that accessing the microphone crashes on `master` but succeeds on this branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8755,7 +8755,7 @@ dependencies = [
  "libc",
  "objc2 0.5.2",
  "objc2-core-data",
- "objc2-core-image",
+ "objc2-core-image 0.2.2",
  "objc2-foundation 0.2.2",
  "objc2-quartz-core 0.2.2",
 ]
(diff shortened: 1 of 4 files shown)


vorporeal:
* **#12074** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12074?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12074?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

alokedesai (approved):

PR #12082, opened 2026-06-02
Fix generic string object update dependencies

evelyn-with-warp:
## Description
Fixes a sync queue dependency gap where several generic string object update variants did not wait for earlier create/update queue items for the same object.

When a local generic string object was created and then updated before the create response mapped its client ID to a server UID, the update could race ahead and call `UpdateGenericStringObject` with a UID the server could not find. That matches the observed GraphQL error:

`graphql response for Some("UpdateGenericStringObject") had errors ... "Not found: Object UID not found"`

This change routes all generic string object update variants through `get_update_dependencies`, including AIFact, MCPServer, AIExecutionProfile, and TemplatableMCPServer.

## Linked Issue
Slack thread: feedback-app `1780425090.994989`

## Testing
- `cargo test --manifest-path /workspace/warp/Cargo.toml -p warp test_sync_queue_generic_string_object_update_depends_on_pending_create`
- `cargo test --manifest-path /workspace/warp/Cargo.toml -p warp sync_queue_dependency`
- `git -C /workspace/warp --no-pager diff --check`
- `cargo fmt --manifest-path /workspace/warp/app/Cargo.toml -- --check`

## Agent Mode


app/src/server/sync_queue.rs
seemeroland: I'm thinking we should remove this catch all and list all variants that explicitly don't have dependencies. It seems too easy to miss when people add new variants

seemeroland (approved):
lgtm! As a nit, remove the bug fix changelog

app/src/server/sync_queue.rs
evelyn-with-warp: Addressed in the latest pushes: removed the catch-all in add_inferred_dependencies and listed the explicit no-dependency QueueItem variants. I also made update_items_with_new_revision exhaustive and extended the GSO regression test to cover chained update revision propagation.

app/src/server/sync_queue.rs
```diff
@@ -426,14 +426,18 @@ impl SyncQueue {
...
+            | QueueItem::UpdateCloudPreferences { id, .. }
+            | QueueItem::UpdateEnvVarCollection { id, .. }
+            | QueueItem::UpdateWorkflowEnum { id, .. }
+            | QueueItem::UpdateAIFact { id, .. }
+            | QueueItem::UpdateMCPServer { id, .. }
+            | QueueItem::UpdateAIExecutionProfile { id, .. }
+            | QueueItem::UpdateTemplatableMCPServer { id, .. }
+            | QueueItem::UpdateCloudEnvironment { id, .. }
+            | QueueItem::UpdateScheduledAmbientAgent { id, .. }
+            | QueueItem::UpdateCloudAgentConfig { id, .. } => self.get_update_dependencies(id),
```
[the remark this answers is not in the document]
evelyn-with-warp: Addressed as part of the follow-up: queued generic string object update variants now receive the latest revision after create/update success, and the regression test now verifies create → update → update keeps the second update on the propagated revision.

PR #12105, opened 2026-06-02
Fix queued prompt paper cuts

harryalbert:
## Description

There were a few small issues w/ queued prompts that I bundled into this PR (from the bug bash). The issues that are fixed here were:
* /queue would queue a prompt if you ran it in an empty conversation. It should just auto-send the prompt, as the conversation is not in progress
* /queue (specifically the slash command, not the queue mode) didn't work during an LRC. This was because the slash command was not allowed in LRCs, but it should be
* deleting a queued prompt would re-insert it into your input. This was intentional before the send-now button, but is not not necessary
* all of the text in the queued prompt was auto-highlighted when you started editing the prompt. Again, this was intentional (following prior art from the vertical tabs), but doesn't make as much sense for regular prompt editing (where you're less likely to be deleting the entire prompt)
* The "send now" button should make it clear that, durring an LRC, the prompt will be sent to the sub-agent. The copy now reads "Send to full terminal use agent" while an LRC is in-progress.

## Testing

### Screenshots / Videos

https://www.loom.com/share/560c3dcabda246caa5bba8312747af90

## Agent Mode

### DIFF
--- a/app/src/terminal/input/slash_commands/mod.rs
+++ b/app/src/terminal/input/slash_commands/mod.rs
@@ -1063,9 +1063,12 @@ impl Input {
                 let history = BlocklistAIHistoryModel::handle(ctx);
                 // An empty conversation defaults to `InProgress` even though nothing is
                 // running, so exclude it here to auto-send rather than queue.
-                let should_queue = history.as_ref(ctx).conversation(&conversation_id).is_some_and(
-                    |c| !c.is_empty() && (c.status().is_in_progress() || c.status().is_blocked()),
-                );
+                let should_queue = history
+                    .as_ref(ctx)
+                    .conversation(&conversation_id)
+                    .is_some_and(|c| {
+                        !c.is_empty() && (c.status().is_in_progress() || c.status().is_blocked())
+                    });
 
                 if should_queue {
                     QueuedQueryModel::handle(ctx).update(ctx, |model, ctx| {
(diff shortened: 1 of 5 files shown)


harryalbert:
* **#12105** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12105?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12105?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/terminal/input/slash_commands/mod.rs
```diff
@@ -1061,12 +1061,16 @@ impl Input {
                 };
 
                 let history = BlocklistAIHistoryModel::handle(ctx);
-                let is_in_progress = history
+                // An empty conversation defaults to `InProgress` even though nothing is
```
moirahuang, line 1064: not at all in scope of prompt queueing but do you know why an empty convo defaults to `InProgress`?

moirahuang (approved):

app/src/terminal/input/slash_commands/mod.rs
harryalbert: I don't, but I'm thinking we might want to change it because I feel like it's caused some issues before. Might ask an agent to find any possible regressions and then just change in dogfood and see what happens

moirahuang: yeah semantically an empty convo not being `InProgress` makes sense to me

PR #12107, opened 2026-06-02
Don't steal focus while navigating an AI block's code diff

advait-m:
Loom: https://www.loom.com/share/e957af04b66c40639ee7eb8498b485cc

## Description
Fixes focus being stolen to the prompt input while navigating an AI block's code diff hunks (up/down) during an agent turn.

At a "request boundary" mid-turn (one response stream finishes / a new exchange streams in), automatic focus handling re-targeted the *active* AI block; since that block has nothing actionable and the input is visible during streaming, focus bounced to the input — interrupting diff-hunk navigation on an earlier block. The same happened at end-of-turn via `redetermine_terminal_focus`.

**How:**
- Added `TerminalView::is_navigating_ai_block`.
- `focus_ai_block_if_self_focused` skips the steal when focus is inside an AI block, unless the target block is blocked on user input (so approval gates still grab focus).
- `redetermine_terminal_focus` excludes the "navigating an AI block" case, alongside the existing find-bar / block-filter exclusions.
- `try_steal_focus` is intentionally left unchanged.

This appears to be a long-standing issue (the focus code predates the public-release squash), not a regression from prompt queueing.

## Linked Issue
N/A — surfaced via internal Slack report (focus stealing while reviewing agent code diffs).

## Testing

`cargo fmt` and `cargo clippy -p warp -- -D warnings` pass.

CHANGELOG-BUG-FIX: Fixed focus jumping to the prompt input while navigating an AI block's code diff hunks with the arrow keys during an agent turn.

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -8209,15 +8209,14 @@ impl TerminalView {
             return;
         }
         let target_needs_attention = block.as_ref(ctx).is_blocked_on_user_confirmation(ctx);
-        if !target_needs_attention && self.is_navigating_ai_block(ctx) {
-            return;
+        if target_needs_attention || !self.is_any_ai_block_focused(ctx) {
+            block.update(ctx, |block, ctx| block.try_steal_focus(ctx));
         }
-        block.update(ctx, |block, ctx| block.try_steal_focus(ctx));
     }
 
     /// Returns `true` if focus is inside any AI block (e.g. the user is arrowing
     /// through a code diff's hunks).
-    fn is_navigating_ai_block(&self, ctx: &mut ViewContext<Self>) -> bool {
+    fn is_any_ai_block_focused(&self, ctx: &mut ViewContext<Self>) -> bool {
         self.rich_content_views.iter().any(|rich_content| {
             rich_content
                 .ai_block_metadata()
@@ -10796,7 +10795,7 @@ impl TerminalView {
         let reset_focus = ctx.is_self_or_child_focused()
             && !self.find_bar.is_self_or_child_focused(ctx)
             && !self.block_filter_editor.is_self_or_child_focused(ctx)
-            && !self.is_navigating_ai_block(ctx);
+            && !self.is_any_ai_block_focused(ctx);
         if reset_focus {
             self.redetermine_global_focus(ctx);
         }
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -8194,6 +8194,9 @@ impl TerminalView {
     /// don't steal focus from the user if they've focused another part of the app
     /// (e.g. another session).
     ///
+    /// Skips the steal when the user is navigating another AI block / code diff
+    /// (e.g. arrowing diff hunks), unless the target block is blocked on user input.
+    ///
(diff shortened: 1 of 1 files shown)


app/src/terminal/view.rs
```diff
@@ -8202,9 +8205,24 @@ impl TerminalView {
         block: &ViewHandle<AIBlock>,
         ctx: &mut ViewContext<Self>,
     ) {
-        if ctx.is_self_or_child_focused() {
-            block.update(ctx, |block, ctx| block.try_steal_focus(ctx));
+        if !ctx.is_self_or_child_focused() {
+            return;
+        }
+        let target_needs_attention = block.as_ref(ctx).is_blocked_on_user_confirmation(ctx);
```
vorporeal, line 8211: i think the complexity will make it out-of-scope for this, but in theory, i think a block that requires user confirmation should only get focused if it is on-screen?

could simplify to "only gets focused if the scroll position is at the top/bottom" - if the scroll position is elsewhere, it indicates that the user is engaged with other parts of the blocklist.  (and if we did that, i could imagine us adjusting focus when the user scrolls all the way to the top/bottom)

again, i think as-written is fine for now, but we could do some more tuning here in a follow-up.

app/src/terminal/view.rs
```diff
@@ -8202,9 +8205,24 @@ impl TerminalView {
...
+        if !target_needs_attention && self.is_navigating_ai_block(ctx) {
+            return;
         }
+        block.update(ctx, |block, ctx| block.try_steal_focus(ctx));
```
vorporeal: readability nit: instead of doing the early return on the negative here ("if we don't need attention and..."), i think it would be clearer to invert the condition here and apply it around the `block.update()`:
```suggestion
        if target_needs_attention || !self.is_navigating_ai_block(ctx) {
            block.update(ctx, |block, ctx| block.try_steal_focus(ctx));
}
```

app/src/terminal/view.rs
```diff
@@ -8202,9 +8205,24 @@ impl TerminalView {
...
+    }
+
+    /// Returns `true` if focus is inside any AI block (e.g. the user is arrowing
+    /// through a code diff's hunks).
+    fn is_navigating_ai_block(&self, ctx: &mut ViewContext<Self>) -> bool {
```
vorporeal: nit: `is_any_ai_block_focused()` is semantically much clearer as a name.  "navigating" isn't a well-defined verb in our application vocabulary.

vorporeal (approved):

app/src/terminal/view.rs
advait-m: agreed, done!

app/src/terminal/view.rs
advait-m: agreed; renamed!

app/src/terminal/view.rs
advait-m: gotcha - yeah being aware of on vs off screen would be good here; filed https://linear.app/warpdotdev/issue/APP-4658/only-focus-blocks-for-needs-user-attention-if-its-on-screen to track this

PR #12110, opened 2026-06-02
Support dragging actions for horizontal tab groups

johnturcoo:
## Description
Adds dragging support for horizontal tab groups, building on the basic rendering from `johnturco/app-4641-horizontal-tab-grouping-basic-rendering`. Horizontal tab bars now support:

- Dragging a whole group container left/right to reorder the group as a block.
- Dragging an individual group member out of the group, or a non-grouped tab into a group.
- Dragging the lone member of a single-tab group, which drags the entire group rather than orphaning the tab.
- Stable behavior when dragging across a collapsed group 

The implementation re-uses the existing vertical-tab group-drag plumbing (`StartGroupDrag` / `DragGroup` / `DropGroup`, `move_group_block`, `assign_tab_to_group`, `hop_tab_to_index`) and generalizes the axis-specific helpers (`neighbor_drag_rect`, `target_group_at_axis`, `on_group_drag`) so both layouts share one code path.

### `workspace/view.rs`
Wired up a group-level `Draggable` in `render_horizontal_tab_group` that dispatches the existing `StartGroupDrag` / `DragGroup` / `DropGroup` actions, with `DragAxis::HorizontalOnly` and child-mouse-down deferral so member tabs still drag individually. Generalized the axis-specific helpers used by both layouts:

- `target_group_at_axis` — replaces the prior `target_group_at_y` / `target_group_at_x` pair; one `is_vertical`-parameterized hit-test.
- `neighbor_drag_rect` — now takes `is_vertical` and falls back to the visible group container rect when the neighbor is a hidden collapsed-group member.
- `on_group_drag` — generalized over both axes; uses neighbor-midpoint thresholds for vertical and neighbor-edge thresholds for horizontal so each layout matches its per-tab drag feel.

### `tab.rs`
`TabComponent` gains `grouped_member` / `sole_grouped_member` flags, set via a new `for_grouped_member(is_sole_member)` builder. Grouped members get the regular per-tab `Draggable`; the sole-member case suppresses it so the parent group's `Draggable` picks up the drag.

### `workspace/view/vertical_tabs.rs`
Added `htab_group_position_id`, the horizontal counterpart to the existing `vtab_group_position_id`, used by the new hit-test and drag-rect helpers.

## Linked Issue
https://linear.app/warpdotdev/issue/APP-4649/horizontal-tab-group-dragging

## Testing

- [x] Tested cross window drags, currently dragging from another window into a group is not supported

### Screenshots / Videos

[Demo video](https://www.loom.com/share/e83f6e40cbf54bbeb134a7ed8f189e87)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -84,11 +84,12 @@ use warpui::clipboard::ClipboardContent;
 use warpui::elements::Percentage;
 use warpui::elements::{
     Align, Border, CacheOption, ChildAnchor, ChildView, Clipped, ConstrainedBox, Container,
-    CornerRadius, CrossAxisAlignment, Dismiss, DispatchEventResult, DraggableState, DropTarget,
-    Element, Empty, EventHandler, Expanded, Fill as ElementFill, Flex, Highlight, Hoverable,
-    Icon as WarpUiIcon, Image, MainAxisAlignment, MainAxisSize, MouseInBehavior, MouseStateHandle,
-    OffsetPositioning, ParentAnchor, ParentElement, ParentOffsetBounds, PositionedElementAnchor,
-    PositionedElementOffsetBounds, Radius, Rect, SavePosition, Shrinkable, Stack, Text,
+    CornerRadius, CrossAxisAlignment, Dismiss, DispatchEventResult, DragAxis, Draggable,
+    DraggableState, DropTarget, Element, Empty, EventHandler, Expanded, Fill as ElementFill, Flex,
+    Highlight, Hoverable, Icon as WarpUiIcon, Image, MainAxisAlignment, MainAxisSize,
+    MouseInBehavior, MouseStateHandle, OffsetPositioning, ParentAnchor, ParentElement,
+    ParentOffsetBounds, PositionedElementAnchor, PositionedElementOffsetBounds, Radius, Rect,
+    SavePosition, Shrinkable, Stack, Text,
 };
 use warpui::fonts::{Properties, Weight};
 use warpui::geometry::vector::{vec2f, Vector2F};
@@ -110,7 +111,7 @@ use warpui::{
 
 use self::vertical_tabs::telemetry::{VerticalTabsDisplayOption, VerticalTabsTelemetryEvent};
 use self::vertical_tabs::{
-    pane_summary_kind, render_detail_sidecar, render_settings_popup,
+    htab_group_position_id, pane_summary_kind, render_detail_sidecar, render_settings_popup,
     render_summary_pane_kind_icon_circle, render_summary_pane_kind_icons, vtab_group_position_id,
     SummaryPaneKind, SummaryPaneKindIcons, VerticalTabsPanelState,
     VERTICAL_TABS_SETTINGS_BUTTON_POSITION_ID,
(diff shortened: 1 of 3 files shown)


peicodes:
Visual feedback: I think the tab group looks a little small - it has a different size than standard tabs, I feel like they should match?

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -138,6 +138,13 @@ pub(crate) fn vtab_group_position_id(group_id: TabGroupId) -> String {
     format!("vertical_tabs:group:{group_id:?}")
 }
 
+/// Save-position id for a horizontal tab group's container rect, used for
+/// drop hit-testing and as the collapsed-group fallback in horizontal-axis
+/// drag math.
+pub(crate) fn htab_group_position_id(group_id: TabGroupId) -> String {
```
peicodes, line 144: does something like this exist for vtabs? Or is it taking a different approach?

app/src/workspace/view.rs
```diff
@@ -18818,23 +18823,66 @@ impl Workspace {
...
+        let container = Container::new(row.finish())
+            .with_border(
+                Border::all(1.)
+                    // Left border only on the first slot to avoid double borders.
+                    .with_sides(false, is_first_in_bar, false, true)
+                    .with_border_fill(internal_colors::fg_overlay_1(theme)),
+            )
+            .finish();
+
+        // Skip the group `Draggable` while another drag (a member tab via the
```
peicodes: How does this happen? I would expect there to be some kind of mutex on drag or that the drag is initiated by clicking and holding something (so it wouldn't be possible to initiate twice without release)

app/src/workspace/view.rs
```diff
@@ -18818,23 +18823,66 @@ impl Workspace {
...
+        // inner per-tab `Draggable`, a cross-window tab drag, or another
+        // group's drag) is already active, so the two drags don't fight for
+        // the same mouse input.
+        let is_this_group_dragging = group.draggable_state.is_dragging();
+        let skip_group_draggable =
+            tab_bar_state.is_any_tab_dragging && !is_this_group_dragging;
+        let group_id = group.id;
+        let group_draggable_state = group.draggable_state.clone();
+        let positioned_container: Box<dyn Element> = if skip_group_draggable {
```
peicodes: nit: A guard clause makes this more readable. E.g. 

```
if tab_bar_state.is_any_tab_dragging && !is_this_group_dragging {
  return container;
}
...
```

peicodes (commented):
Just questions to help my understanding - happy to stamp once I do

app/src/workspace/view/vertical_tabs.rs
johnturcoo: Yes, the equivalent for vtabs is found directly above this (`vtab_group_position_id`). The reason for its existence is to handle two edge cases when dragging tabs into/out of groups: 
1. The group is expanded, can we drag into the first or last position of the group without dragging past the current first/last tab in that group (hit testing for group bounds)
2. The group is collapsed, we must drag the tab past the entire group and not compare with a neighboring tab (which would be inside the group)

app/src/workspace/view.rs
johnturcoo: The `DraggableState` mutex is enough here. This is redundant. Removing it

peicodes (approved):

PR #12122, opened 2026-06-03
Stop watching gitignored directories in the repo file watcher

kevinyang372:
## Description

**What:** Make Warp's repository file watcher stop registering filesystem watches on **gitignored directories** (`node_modules`, build output, vendored deps, etc.), while still watching directories that consumers explicitly care about via registered "ignored-path interests".

**Why:** On large monorepos the watcher recursively registered an inotify watch for every non-`.git` directory, including gitignored ones. On a customer's remote-server daemon this ballooned to ~11 GB RSS, with ~98% of the live heap inside `notify::inotify::EventLoop::add_watch` (the per-directory watch table + cloned path buffers). The descend filter only pruned `.git/` internals, never gitignored content.

**How:** The watcher's descend predicate (`repo_watch_filter`, used by both `DirectoryWatcher` and `LocalRepoMetadataModel`) is now gitignore-aware:

- New `should_watch_repo_directory(path, gitignores, interests)`:
  1. `.git/` internals keep the existing allowlist (`should_watch_directory_in_git_path`).
  2. Any directory on the path to — or inside — a registered ignored-path **interest** is always watched (reuses `matches_ignored_path_interest`, which also matches the ancestor prefixes leading to an interest).
  3. Otherwise, gitignored directories are pruned.
- Pruning uses **ancestor-aware** matching (`matched_path_or_any_parents`), which guarantees the watcher's monotonicity invariant: a child of an ignored directory is also reported ignored, so we never accept a descendant after rejecting its parent. Directory-only re-include negations (e.g. `parentdir/*` + `!parentdir/*/`) keep working because the re-included directory matches as not-ignored on its own path (last-match-wins).
- `repo_watch_filter(gitignores, interests)` is now parameterized; both watch-registration sites pass the repo's root + global gitignores (matching the existing `is_ignored` tagging) and their interest list.
- `DirectoryWatcher` gains an `ignored_path_interests` list + `register_ignored_path_interests`; `app/src/lib.rs` registers the skill-provider paths on it too (they were already registered on `RepoMetadataModel`), so gitignored skill directories (e.g. `.agents/skills`) stay watched on both the LSP/MCP path and the file-tree/skills path.

**Consumer impact:** Skills are covered on both watcher paths via interests. Project rules already skip gitignored files on live updates; MCP config files are normally tracked. The emit predicate is unchanged (gitignored files inside a watched, non-ignored directory are still emitted and tagged `is_ignored`).

**Known follow-ups (not in this PR):**
- LSP servers that register `didChangeWatchedFiles` globs pinned to a gitignored base directory: plumb those base dirs in as interests. Default behavior otherwise matches typical editors (no events for gitignored paths). LSP already filters to server-registered globs.
- Nested per-directory `.gitignore` files are not consulted by the descend filter (same limitation as the existing `is_ignored` tagging); this can only over-watch, never miss events.
- Consolidating the two repo_metadata watchers and adding a watch-count metric/cap.

## Linked Issue

No linked GitHub issue. This is an internal reliability fix surfaced by heap profiling of a customer's remote-server daemon.

## Testing

Added unit tests in `crates/repo_metadata/src/entry_tests.rs` for `should_watch_repo_directory`:
- gitignored directory is pruned (and its descendants, ancestor-aware);
- a registered interest is descended to and into, even under a gitignored ancestor;
- nested case: `a/b` ignored but `a/b/c` is an interest → full prefix descended, ignored sibling pruned;
- directory-only re-include negation (`parentdir/*` + `!parentdir/*/`) descends subdirectories while the loose ignored file stays filtered;
- `.git` allowlist behavior preserved.

Validation run locally:
- `cargo test -p repo_metadata --features local_fs` — 72 passed (2 pre-existing flaky ignored), including the 5 new tests.
- `cargo clippy -p repo_metadata --features local_fs --all-targets` — clean.
- `./script/format` / `cargo fmt -- --check` — clean.
- `cargo check --workspace --all-targets` — compiles (incl. `lsp`, `warp` app crate).
- Targeted `cargo-clippy-diff origin/master HEAD` — no new lint violations.

Full `cargo clippy --all-targets --all-features -D warnings` and `cargo nextest run --workspace` will run in CI.

## Agent Mode

CHANGELOG-BUG-FIX: Reduced memory usage of the file watcher on large projects by no longer watching gitignored directories (e.g. node_modules, build output).

### DIFF
--- a/crates/repo_metadata/src/entry.rs
+++ b/crates/repo_metadata/src/entry.rs
@@ -73,7 +73,7 @@ pub(crate) struct BuildTreeOptions<'a> {
     pub max_depth: usize,
     pub current_depth: usize,
     pub ignored_path_strategy: &'a IgnoredPathStrategy,
-    pub ignored_path_interests: &'a [PathBuf],
+    pub force_included_paths: &'a [PathBuf],
     pub budget_exceeded_behavior: BudgetExceededBehavior,
 }
 
@@ -128,7 +128,7 @@ impl Entry {
         ignored_path_strategy: &IgnoredPathStrategy,
         budget_exceeded_behavior: BudgetExceededBehavior,
     ) -> Result<Self, BuildTreeError> {
-        Self::build_tree_with_ignored_path_interests_and_ancestor(
+        Self::build_tree_with_force_included_paths_and_ancestor(
             path,
             files,
             gitignores,
@@ -137,23 +137,24 @@ impl Entry {
                 max_depth,
                 current_depth,
                 ignored_path_strategy,
-                ignored_path_interests: &[],
+                force_included_paths: &[],
                 budget_exceeded_behavior,
             },
             false,
         )
     }
 
-    /// Builds a tree of entries from a given path, loading ignored paths that match
-    /// one of the supplied component-sequence interests instead of leaving them lazy.
-    pub(crate) fn build_tree_with_ignored_path_interests(
+    /// Builds a tree of entries from a given path, eagerly loading any path that
+    /// matches one of the supplied force-included paths instead of leaving it
+    /// lazy (see [`BuildTreeOptions::force_included_paths`]).
+    pub(crate) fn build_tree_with_force_included_paths(
         path: impl Into<PathBuf>,
         files: &mut Vec<FileMetadata>,
         gitignores: &mut Vec<Gitignore>,
         remaining_file_quota: Option<&mut usize>,
         options: BuildTreeOptions<'_>,
     ) -> Result<Self, BuildTreeError> {
-        Self::build_tree_with_ignored_path_interests_and_ancestor(
+        Self::build_tree_with_force_included_paths_and_ancestor(
             path,
             files,
             gitignores,
@@ -174,7 +175,7 @@ impl Entry {
         ignored_path_strategy: &IgnoredPathStrategy,
         ancestor_is_ignored: bool,
     ) -> Result<Self, BuildTreeError> {
-        Self::build_tree_with_ignored_path_interests_and_ancestor(
+        Self::build_tree_with_force_included_paths_and_ancestor(
             path,
             files,
             gitignores,
@@ -183,7 +184,7 @@ impl Entry {
                 max_depth,
                 current_depth,
                 ignored_path_strategy,
-                ignored_path_interests: &[],
+                force_included_paths: &[],
                 budget_exceeded_behavior: BudgetExceededBehavior::StopAndLazyLoad,
             },
             ancestor_is_ignored,
@@ -191,7 +192,7 @@ impl Entry {
     }
 
     #[allow(clippy::too_many_arguments)]
-    pub(crate) fn build_tree_with_ignored_path_interests_and_ancestor(
+    pub(crate) fn build_tree_with_force_included_paths_and_ancestor(
         path: impl Into<PathBuf>,
         files: &mut Vec<FileMetadata>,
         gitignores: &mut Vec<Gitignore>,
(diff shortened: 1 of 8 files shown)


kevinyang372:
* **#12211** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12176** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12176?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12122** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12122?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12122?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/repo_metadata/src/entry.rs
```diff
@@ -703,21 +703,79 @@ fn descend_allowlist_matches(suffix: &[Component<'_>]) -> bool {
...
+///    ([`should_watch_directory_in_git_path`]).
+/// 2. Any directory on the path to — or inside — a registered ignored-path
+///    interest is always watched, so consumers (e.g. skills) keep getting
+///    updates for ignored subtrees they explicitly care about. This mirrors
+///    the tree builder's lazy-load exception
+///    ([`matches_ignored_path_interest`]) and, because that match also returns
+///    `true` for the ancestor prefixes leading to an interest, it preserves
+///    the watcher's monotonicity invariant.
+/// 3. Otherwise, gitignored directories are pruned so we don't register
+///    inotify watches on `node_modules`, build output, vendored deps, etc.
```
alokedesai: Do exposing the internal details really matter? I feel like all we need to say is that this function determins whether a given directory should be watched or not. This current rustdoc is mad confusing

crates/repo_metadata/src/entry.rs
```diff
@@ -703,21 +703,79 @@ fn descend_allowlist_matches(suffix: &[Component<'_>]) -> bool {
...
+pub fn should_watch_repo_directory(
+    path: &Path,
+    gitignores: &[Gitignore],
+    ignored_path_interests: &[PathBuf],
+) -> bool {
+    if is_git_internal_path(path) {
+        return should_watch_directory_in_git_path(path);
+    }
+
+    if matches_ignored_path_interest(path, ignored_path_interests) {
```
alokedesai: what does `matches_ignored_path_interest` mean? That's inscrutable to me. Same with `ignored_path_interests`

crates/repo_metadata/src/watcher.rs
```diff
@@ -41,6 +42,12 @@ pub struct DirectoryWatcher {
 
     /// Handle to the internal processing queue model that orders scan & update tasks.
     processing_queue: ModelHandle<TaskQueue>,
+
+    /// Component-sequence paths that should stay watched even when gitignored
+    /// (e.g. skill provider directories). Fed into the watch descend filter so
+    /// consumers keep getting updates for ignored subtrees they care about.
+    #[cfg_attr(not(feature = "local_fs"), allow(dead_code))]
+    ignored_path_interests: Vec<PathBuf>,
```
alokedesai: this is just saying that we want to lazy load / lazy watch a directory even if we aren't watching it's parent, right?


If so, can we just make that more clear and use the existing loading mechanism?

crates/repo_metadata/src/watcher.rs
```diff
@@ -92,6 +100,27 @@ impl DirectoryWatcher {
...
+        }
+    }
+
+    /// Registers component-sequence paths that should stay watched even when
+    /// gitignored. Mirrors `LocalRepoMetadataModel::register_ignored_path_interests`
+    /// but applies to the watcher backing `Repository` subscribers (LSP, MCP).
+    /// Must be called before repositories begin watching for the interest to
+    /// take effect on already-registered watches.
+    ///
+    /// Kept as a `Vec` (deduped on insert): registration happens a handful of
```
alokedesai: epic nit: this rustdoc shouldn't leak internal implementation details

crates/repo_metadata/src/watcher.rs
```diff
@@ -92,6 +100,27 @@ impl DirectoryWatcher {
...
+    /// times at startup, while `start_watching_directory` reads it far more
+    /// often and benefits from a cheap clone.
```
alokedesai: How does the vec result in a cheap clone?

alokedesai (approved):
My main question is whether this still allows for watching a gitignored directory if it's lazy loaded? That's obviously important for us to continue to support.

crates/repo_metadata/src/watcher.rs
kevinyang372: It was kept as a HashSet and I told the agent to make it Vec to avoid converting it everytime we start watching a directory. Will remove this comment here

kevinyang372:
> My main question is whether this still allows for watching a gitignored directory if it's lazy loaded? That's obviously important for us to continue to support.

@alokedesai There shouldn't be regression in most of the cases (MacOS and Windows doesn't go through the tree descend code path to register inode watchers)

One case that won't work after this is expanding a gitignored folder in Linux file tree. This will be fixed in the follow-up to this PR (we will register non-recursive watcher as user expands the folder)

Issue #11914, opened 2026-05-29
Code Review panel briefly shows repository error when switching repos


seemeroland:
## Description

When the Code Review panel has multiple repositories available and the user switches to another repository, the panel can briefly show the error state:

> Code review panel only works in a repository

After a short delay, the selected repository eventually loads and the panel recovers.

## Reproduction context

- Code Review panel has two repositories available.
- Switching from one repository to the other triggers the transient error state.
- The report was from local development, not a devbox.

## Expected behavior

Switching between repositories in the Code Review panel should either keep showing a loading state or transition directly to the selected repository without showing the “only works in a repository” error.

## Actual behavior

The panel briefly shows the repository error state before eventually showing the selected repository.

## Public-safety notes

This issue was filed from an internal Slack report, but the public issue only includes the generic UI behavior and sanitized reproduction context. The Slack text and replies did not contain customer details, credentials, internal architecture, logs, stack traces, or other sensitive content. A ZIP attachment was present in Slack but was not inspectable by the triage fetcher, so it was omitted from this public issue.

Warp-internal Slack report: https://warpdev.slack.com/archives/C08KTPNQN65/p1780093159997789

PR #12126, opened 2026-06-03
Fix "cannot detect diffs" error

MaggieShan:
## Description
* This fixes an issue where the code review view would occasionally flash the "cannot detect diffs" error despite being in a git repo with valid diffs to render 
* The bug is that a local diff state model defaults to `DiffState::NotInRepository` on creation which the view maps to this error message
  * This used to work when the view also treated `NotInRepository` as a loading state if `repo_path.is_some()`, but with the refactors made to support a remote diff state model, the semantics for `NotInRepository` changed to be a valid error where the path was definitively not inside a git repo
* The fix introduces a new starting state where `DiffState::Detecting` explicitly means that we're actively detecting the repo before loading the diffs 

## Linked Issue
Reported in: 
* https://github.com/warpdotdev/warp/issues/11914
* https://warpdev.slack.com/archives/C08KTPNQN65/p1780093159997789
* https://warpdev.slack.com/archives/C08KTPNQN65/p1780349380531869

## Testing
https://www.loom.com/share/7bab61e0fe624f4d8a5cdfc440e28258

## Agent Mode

### DIFF
--- a/app/src/code_review/diff_state/local.rs
+++ b/app/src/code_review/diff_state/local.rs
@@ -79,6 +79,10 @@ const BIDI_CHARS: [char; 9] = [
 /// and changes against the main branch.
 #[derive(Clone, Default)]
 enum InternalDiffState {
+    /// Repo detection has been kicked off but hasn't completed yet.
+    /// We don't yet know whether the path is inside a git repository, so this is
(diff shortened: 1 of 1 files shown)


kevinchevalier (approved):
Nice! This makes the state clear!

PR #12111, opened 2026-06-02
Add client-side retry for oz starting shared sessions

seemeroland:
## Description

Motivation for this: https://linear.app/warpdotdev/issue/REMOTE-1802/session-sharing-server-times-out-creating-session

When we fail to start session-sharing-server for retryable reasons in an oz run, we make up to 3 attempts, with each attempt getting 5s (total 20s wait on the oz agent driver).

## Testing

Tested with local session-sharing-server injecting errors. The first two succeed on the third attempt and run proceeds normally. Manual user sharing also still works

Inject two internal server errors
```
2026-06-03T18:14:04Z [INFO] [warp::terminal::view] Terminal bootstrapped with pending shared session; attempting to share
2026-06-03T18:14:04Z [INFO] [warp::terminal::view::shared_session::view_impl] Emitting request to start sharing current session
2026-06-03T18:14:04Z [INFO] [warpui_core::core::app] dispatching global action for workspace:save_app
2026-06-03T18:14:04Z [INFO] [warp::terminal::local_tty::terminal_manager] Starting shared session
2026-06-03T18:14:04Z [INFO] [warp::terminal::local_tty::terminal_manager] Shared session local lifecycle: event=start_requested session_id=None source_type=ambient_agent source_task_id=Some("019e8eb1-3119-7119-9616-17506db20222") trigger=terminal_view_start_sharing
2026-06-03T18:14:04Z [INFO] [warp::terminal::shared_session::sharer::network] Shared session sharer lifecycle: event=create_attempt_started session_id=None source_type=ambient_agent source_task_id=Some("019e8eb1-3119-7119-9616-17506db20222") attempt=1 max_attempts=3
2026-06-03T18:14:04Z [INFO] [warp::terminal::model::session] Loading history from file /Users/rolandhuang/.zsh_history for shell zsh
2026-06-03T18:14:04Z [INFO] [warp::terminal::shared_session::sharer::network] Connecting to session sharing server
2026-06-03T18:14:04Z [INFO] [warp::terminal::shared_session::sharer::network] Connected to session sharing server; preparing initialization
2026-06-03T18:14:04Z [INFO] [warp::ai::agent_sdk::driver] Starting 0 existing and 0 ephemeral MCP servers
2026-06-03T18:14:04Z [INFO] [warp::terminal::shared_session::sharer::network] Sent session sharing initialization message
2026-06-03T18:14:04Z [INFO] [warp::ai::agent_sdk::driver] Selecting base agent model auto-efficient (from agent driver)
[...]

PR #12140, opened 2026-06-03
fix: recover shared-session viewer initial joins

captainsafia:
## Description
- Add the PRODUCT and TECH specifications for recoverable initial shared-session viewer joins.
- Retry transient websocket/connect failures before the viewer receives its initial join acknowledgement, with bounded exponential backoff and stale-attempt protection.
- Transition failed initial joins to an in-pane error surface instead of leaving the pane loading indefinitely, exposing same-pane Retry only for recoverable failures.
- Stop heartbeat work between attempts and keep established-session reconnect semantics separate from initial join recovery.

## Linked Issue
- Linear: [REMOTE-1661](https://linear.app/warpdotdev/issue/REMOTE-1661/errors-joining-a-cloud-mode-session-can-leave-the-pane-in-a-broken)

## Demo
- Loom: https://www.loom.com/share/547be46ef74d4ddcb696c8eebf4a0b4e

## Testing
- `./script/format --check`
- `cargo check --manifest-path Cargo.toml -p warp --tests`
- `cargo test --manifest-path Cargo.toml -p warp initial_join -- --nocapture`
- `cargo test --manifest-path Cargo.toml -p warp failed_viewer_join -- --nocapture`
- `cargo clippy --manifest-path Cargo.toml --workspace --exclude warp_completer --all-targets --tests -- -D warnings`
- `cargo clippy --manifest-path Cargo.toml -p warp_completer --all-targets --tests -- -D warnings`
- Attempted `./script/presubmit`; its full-workspace `cargo test --no-run` phase was terminated by `SIGKILL` in the sandbox after formatting and Clippy completed successfully.

## Agent Mode

CHANGELOG-NONE

_Conversation: https://staging.warp.dev/conversation/a5926e17-47ce-4726-a86b-aa2da38db2bc_
_Run: https://oz.staging.warp.dev/runs/019e8e03-950a-7cc7-9bde-6662dcf1932d_
_Plans_:
- _[Make Shared-Session Viewer Initial Joins Recoverable](https://staging.warp.dev/drive/notebook/bmcEn9Se998lvDKKc0x9pK)_


captainsafia:
@seemeroland There's a small overlap in the diff between this and what you're doing in #12111 but this addresses an issue with the viewers joining the session share as opposed to initializing the session share.

app/src/terminal/shared_session/viewer/network.rs
```diff
@@ -120,11 +123,15 @@ pub struct Network {
     initial_load_mode: SharedSessionInitialLoadMode,
 
     stage: Stage,
+    initial_join_attempt_id: u64,
```
seemeroland: I think we can move this state within `Stage::BeforeJoined` since they're strongly coupled. Would also simplify some of the checks checking against that stage

app/src/terminal/shared_session/viewer/terminal_manager.rs
```diff
@@ -1503,9 +1514,31 @@ impl TerminalManager {
...
+                    let status = model.shared_session_status();
+                    (
+                        status.failed_viewer_join_error().is_some(),
+                        status.can_retry_failed_viewer_join(),
+                    )
+                };
+                if can_retry_failed_initial_join {
+                    Self::update_current_network(&current_network, ctx, |network, ctx| {
+                        model.lock().set_write_to_pty_events_for_shared_session_tx(
+                            network.write_to_pty_events_tx(),
```
seemeroland: Do we actually need to store the `write_to_pty_events_tx` on the network to reset it here on retry? I thought this would be the same terminal manager which would have already set the same `write_to_pty_events_tx` before creating the network

seemeroland (commented):
Nice!

app/src/terminal/shared_session/viewer/network.rs
captainsafia: Good call! Fixed this.

app/src/terminal/shared_session/viewer/terminal_manager.rs
captainsafia: I was aiming for symetry with the receiver here but you're right that the network doesn't actually do anything with the sender at the moment outside of relying it to he terminal manager. I think it's probably fine to split the two up.

captainsafia:
@seemeroland would appreciate a re-review on this one too

PR #12144, opened 2026-06-03
Keep ssh-host (and other generator) completions when one Include fails (fixes #9417)

### LINKED ISSUE #9417: SSH config Include with Windows ssh.exe Git Bash
### Pre-submit Checks

### Describe the bug

In SSH config you can include mingw style
`Include /c/Users/yourname/.ssh/extra_config`
or windows style
`Include C:\Users\yourname\.ssh\extra_config`

If you are using windows style Warp is not showing any completion proposal.

This is fine as long as you use the mingw ssh, but if you chose to use Windows ssh.exe at Git Bash installation then mingw style does not work.
Adding both include style is fine for Windows ssh.exe, but Warp autocompletion broke.

There is no workaround by using mingw ssh for me because of firewall configuration I have to use Windows ssh.exe.
The only workaround is to copy the included file in the .ssh directory and use relative path.

You could do two things :
- either ignore Include which to not resolve well, but still propose autocompletion with other include content (in that case I will put both style in my config)
- and/or support Windows path style (in that case I can put only Windows include style in my config)

### To reproduce

See the bug report

### Expected behavior

_No response_

_No response_

### Operating system (OS)

Windows

Windows 11

### Shell Version

_No response_

### Current Warp version

_No response_

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

maxmilian:
## Summary

Fixes #9417. On Windows + Git Bash with Windows `ssh.exe`, an `~/.ssh/config`
containing a Windows-style `Include C:\Users\...\extra_config` line caused Warp
to show **no** ssh host completions at all. This makes the completion engine
degrade gracefully: a generator command that exits non-zero but still wrote
usable output now keeps that output instead of discarding every suggestion.

## Root cause

Warp's ssh-host completion generator `cat`s `~/.ssh/config` plus every
`Include`d file and pipes the result through `awk`. On Windows/Git Bash a
drive-letter `Include C:\Users\...` path is mis-resolved by the POSIX generator
as a relative path, so that one `cat` fails and the **whole** command exits
non-zero — even though the readable hosts were already written to stdout.

In both the legacy and v2 generator argument paths the engine mapped
`CommandExitStatus::Failure` straight to an empty result
(`crates/warp_completer/src/completer/engine/argument/{legacy,v2}.rs`), so a
single unreadable `Include` wiped out **all** ssh host completions instead of
degrading to the hosts it could read.

## Fix

Always parse whatever made it to stdout, and only bail when a failing command
produced nothing usable (empty or whitespace-only stdout). The success path is
unchanged; on failure we now surface the partial output rather than dropping it.
`output.to_string()` reads stdout only (never stderr), so no error text leaks
into completions. Applied identically to the legacy and v2 paths.

## Tests

- `test_generator_failure_with_partial_output_still_suggests` — a generator
  whose command fails-with-output no longer drops its suggestions, and a sibling
  generator that succeeded is unaffected. Fails before the fix, passes after.
- `test_generator_failure_with_blank_output_suggests_nothing` — a failure with
  only-whitespace stdout still suggests nothing (guards the `trim()` check so no
  bogus empty completion leaks).
- Extends `MockGeneratorContext` with `with_failing_command(...)` to model a
  command that exits non-zero with partial stdout.

Local: `cargo test -p warp_completer` green, `cargo fmt --check` clean,
`cargo clippy -p warp_completer --all-targets` clean.

## Scope

Intentionally **not** adding Windows drive-letter path resolution to the
generator — this PR implements option (a) from the issue ("ignore an Include
that doesn't resolve, but still propose completions from the rest"). The hosts
in the unreadable Windows-style `Include` remain uncompletable; the readable
hosts now survive instead of taking the whole list down with them. Native
Windows-path Include support can be a follow-up.

Note this graceful-degrade applies to *all* script generators, not just
ssh-host: any generator whose command exits non-zero with stdout content now
has that content parsed. In practice failing commands write diagnostics to
stderr, not stdout, so this is the desired behavior.

### DIFF
--- a/crates/warp_completer/src/completer/engine/argument/legacy.rs
+++ b/crates/warp_completer/src/completer/engine/argument/legacy.rs
@@ -718,36 +718,37 @@ async fn generate_suggestions_for_argument_type(
                 return vec![];
             };
 
-            match output.status {
-                CommandExitStatus::Success => {
-                    let Ok(output_string) = output.to_string() else {
-                        return vec![];
-                    };
+            // A generator command can exit non-zero yet still have written
+            // usable output before failing. The motivating case (#9417) is the
+            // ssh-host generator: it `cat`s `~/.ssh/config` and every `Include`d
+            // file, so one unreadable `Include` (e.g. a Windows `C:\Users\...`
+            // path the POSIX generator mis-resolves) fails the whole command --
+            // even though the readable hosts are already on stdout. Parse what
+            // we got instead of dropping every suggestion; only bail when a
+            // failing command produced nothing to parse.
+            let Ok(output_string) = output.to_string() else {
+                return vec![];
+            };
+            if output.status == CommandExitStatus::Failure && output_string.trim().is_empty() {
+                return vec![];
+            }
 
-                    let results = generator.on_complete(&output_string);
+            let results = generator.on_complete(&output_string);
 
-                    let internal_suggestions = results
-                        .suggestions
-                        .into_iter()
-                        .map(Into::into)
-                        .filter_map(|suggestion: Suggestion| {
-                            let match_type = matcher
-                                .get_match_type(parsed_token.as_str(), suggestion.display.as_str());
-                            match_type
-                                .map(|match_type| MatchedSuggestion::new(suggestion, match_type))
-                        });
+            let internal_suggestions = results.suggestions.into_iter().map(Into::into).filter_map(
+                |suggestion: Suggestion| {
+                    let match_type =
+                        matcher.get_match_type(parsed_token.as_str(), suggestion.display.as_str());
+                    match_type.map(|match_type| MatchedSuggestion::new(suggestion, match_type))
+                },
+            );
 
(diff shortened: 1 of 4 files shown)


maxmilian:
Closing this PR.

While the engine-level change is correct in isolation (a generator command that exits non-zero with partial stdout should keep that output, covered by the added unit tests), I was unable to validate that it actually changes the user-visible behavior reported in #9417.

On testing, ssh host completion draws from multiple redundant generators/sources, so a single failing `Include` does **not** wipe out the completion list — the readable hosts still surface through another source. I could not reproduce the "no completions at all" symptom outside of the original Windows-specific setup, so I can't demonstrate that this change fixes the reported behavior end to end.

Rather than land a change whose user-facing impact I can't confirm, I'm withdrawing it. If a maintainer with the Windows/Git-bash repro confirms the failing generator is the sole host source there, the engine fix here may still be useful as a starting point.

PR #12165, opened 2026-06-03
Add auto-handoff sleep discoverability notification

dplakon:
## Description
Adds a discoverability notification that surfaces on wake when the Mac slept while an in-progress **local** agent run was interrupted and **auto-handoff-on-sleep is off**. The notification (toast + inbox) explains the feature and lets the user enable it inline.

- **Enable** turns on `auto_handoff_on_sleep_enabled`. Once on, the prompt no longer fires.
- **Dismiss** closes the current notification; it is *not* permanently suppressed and resurfaces on the next sleep interruption.
- Gated behind a new `FeatureFlag::AutoHandoffSleepPrompt` (added to `DOGFOOD_FLAGS`), and only fires when cloud handoff is otherwise available.

### How it works
- `AutoCloudHandoffController` records a pending prompt at `CpuWillSleep` when an eligible in-progress local run *would* have handed off but the setting is off (reuses the existing `AutoCloudHandoffEligibility` gate), then surfaces it at `CpuWasAwakened`.
- New `NotificationAction`/`NotificationActionButtonsRow` add inline action buttons to notifications (mirrors `ArtifactButtonsRow`), rendered in both the toast and the inbox via the shared `render_notification_item_content`.
- New `NotificationOrigin::AutoHandoffSleepPrompt` keeps it a single de-duped item, independent of the interrupted conversation's own status notifications.
- Telemetry: `AmbientAgent.Handoff.SleepPrompt.{Shown,Enabled,Dismissed}`.

## Linked Issue
N/A — discoverability follow-up to sleep auto-handoff.

## Testing
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings`, and `notifications::item` unit tests pass.
- [x] Manually exercised on macOS: sleep/wake with an in-progress local agent and the setting off surfaces the prompt; Enable flips the setting, Dismiss closes it and it returns on the next sleep.

### Screenshots / Videos
Screenshot 2026-06-03 at 2 53 08 PM

## Agent Mode

[Warp conversation](https://staging.warp.dev/conversation/d7edf7c7-c4d7-4857-b93a-4080e06bb5bd) · [Plan](https://staging.warp.dev/drive/notebook/rra4szH0TH53r1T83CjYHs)

CHANGELOG-IMPROVEMENT: Prompt to enable auto-handoff-on-sleep when a local agent run is interrupted by sleep.


app/src/ai/agent_management/agent_management_model.rs
```diff
@@ -449,6 +452,80 @@ impl AgentNotificationsModel {
         self.notifications.push(item);
         ctx.emit(AgentManagementEvent::NotificationAdded { id });
     }
+
+    /// Creates the auto-handoff sleep discoverability prompt notification.
+    /// Called on wake when an in-progress local agent run was interrupted by
+    /// sleep while `auto_handoff_on_sleep_enabled` was off. The prompt carries
+    /// Enable/Dismiss action buttons handled by `handle_notification_action`.
+    pub(crate) fn add_auto_handoff_sleep_prompt(
```
harryalbert, line 460: nit, but should this fn be called `add_auto_handoff_sleep_notification` (just following the pattern from above)?

app/src/ai/agent_management/agent_management_model.rs
```diff
@@ -449,6 +452,80 @@ impl AgentNotificationsModel {
...
+        let id = item.id;
+        self.notifications.push(item);
+        send_telemetry_from_ctx!(CloudAgentTelemetryEvent::SleepPromptShown, ctx);
+        ctx.emit(AgentManagementEvent::NotificationAdded { id });
+    }
+
+    /// Handles a click on a notification action button (e.g. the sleep prompt's
+    /// Enable/Dismiss). Applies the action's side effect, then removes the
+    /// notification.
+    pub(crate) fn handle_notification_action(
```
harryalbert, line 502: hmm, it seems a little weird to me that this would live on the `agent_management_model`? We handle other notification actions in `AgentNotificationToastStack` IIRC so I'd expect this to be there as well.

it seems like the only thing we use `self` for here is to remove the notification at the end, so we could maybe just expose that remover publicly instead of having this logic all live in this model

app/src/ai/agent_management/notifications/action_buttons.rs
harryalbert, line 1: I might be missing something, but this module seems like a bit of overkill to me — we already have the action button component, and it feels like all this adds is the ability to make a row of buttons which we could maybe just do inline

app/src/workspace/auto_handoff.rs
```diff
@@ -102,6 +104,10 @@ impl AutoCloudHandoffRequest {
 }
 pub(crate) struct AutoCloudHandoffController {
     attempted_conversation_ids: HashMap<AIConversationId, AutoCloudHandoffAttemptState>,
+    /// Set at sleep time when an eligible in-progress local agent run would have
+    /// been handed off but `auto_handoff_on_sleep_enabled` is off. Consumed on
+    /// wake to surface the discoverability prompt.
+    pending_sleep_prompt: Option<EntityId>,
```
harryalbert, line 110: this variable name reads to me like it's a prompt that we would send on-wake (or on-sleep)

app/src/workspace/auto_handoff.rs
```diff
@@ -133,9 +140,141 @@ impl AutoCloudHandoffController {
...
+
+    /// At sleep time, records a pending discoverability prompt when an eligible
+    /// in-progress local agent run *would* have auto-handed-off but the sleep
+    /// setting is off. The actual notification is surfaced on wake by
+    /// [`Self::maybe_show_sleep_prompt`].
+    fn maybe_record_sleep_prompt(&mut self, ctx: &mut ModelContext<Self>) {
+        self.pending_sleep_prompt = None;
+
+        if !FeatureFlag::AutoHandoffSleepPrompt.is_enabled() {
+            log::info!("auto-handoff sleep prompt: skipping at sleep, feature flag disabled");
```
harryalbert, line 159: super nit, but do we need this log (and the logs below)?

harryalbert (commented):
Had a few small nits and architecture thoughts.

Also, at a product level, do you think it makes sense to show this every time that the user closes their computer w/ a conversation in progress, even if they've dismissed it? oo1, I'd expect us to only surface this notification once and then not surface it again once a user has dismissed it

app/src/ai/agent_management/agent_management_model.rs
dplakon: I had it set to forever dismiss at first, but I want to optimize for usage a bit. I think it's going be pretty rare that this triggers, and when it does it only shows in the notification inbox.

PR #12166, opened 2026-06-03
Allow partial build of repo metadata for repos exceeding max limits

kevinyang372:
## Description
This makes `repo_metadata` build a **partial** file tree for git repos that exceed the indexing file limit, instead of falling back to a shallow `depth=1` tree.

### Motivation
- **The `depth=1` fallback is awkward to build on.** Today, when a repo exceeds `MAX_FILES_PER_REPO`, we rebuild the entire repo as a shallow `depth=1` tree with lazily-loaded subdirectories. That shape is hard for downstream consumers to use: migrating skills discovery and project-context/rules onto repo metadata would each need to implement custom, on-demand expansion of a lazily-loaded tree just to find files below the first level. A partial-but-real tree removes that burden.
- **It unblocks non-recursive watcher registration for lazily-loaded repos.** As we move toward registering file watchers non-recursively for lazily-loaded directories, a budgeted/partial tree gives us much better watcher behavior on large repos (we watch what we actually loaded rather than the whole tree).

### Approach
- **Conservative, bounded "unbounded" indexing.** Other popular editors (VSCode, Zed) effectively index the whole non-ignored tree with no hard cap. We move in that direction but stay conservative for now with an upper bound (`MAX_FILES_PER_REPO`), since gitignored directories are already lazy and don't count toward it.
- **Load everything until the budget, then mark the rest lazy.** When the file budget is exhausted, the builder stops descending and leaves the remaining directories as unloaded placeholders (lazy-loaded on demand) instead of erroring or collapsing the whole repo to a single level. Every node scanned before the budget is kept.
- **BFS instead of DFS.** The builder now does a queue-driven, level-order breadth-first traversal rather than recursive depth-first. This spends the budget evenly and shallow-first (all of level 1, then level 2, …), which is far more useful for the file tree / `@`-context / file search than DFS's "first subtree fully, rest empty," and matches Zed's queue-based scan order.
- **Paths of interest are always included.** Registered `ignored_path_interests` (e.g. skill provider directories like `.agents/skills`) are always expanded, even past the budget, so discovery-critical files stay reachable regardless of repo size.

Consumers that must not operate on a partial tree can opt into the previous behavior: `Entry::build_tree` now takes a `BudgetExceededBehavior` parameter, and codebase embedding passes `FailFast` (the file limit there is an intentional cost cap).

## Linked Issue

## Testing
- Added `repo_metadata` unit tests covering: breadth-first coverage with the remainder left unloaded on budget exhaustion, interest paths loading past the budget, directories/gitignored files not consuming the budget, full coverage within budget, and `FailFast` erroring vs. succeeding.
- `cargo nextest run -p repo_metadata --features local_fs` passes; `cargo clippy` clean on `repo_metadata` + `ai`; `cargo check -p warp` builds.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/crates/repo_metadata/src/entry.rs
+++ b/crates/repo_metadata/src/entry.rs
@@ -1,5 +1,6 @@
 #![cfg_attr(not(feature = "local_fs"), allow(dead_code))]
 
+use std::collections::VecDeque;
 use std::io;
 use std::path::{Component, Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering};
@@ -47,6 +48,20 @@ pub enum IgnoredPathStrategy {
     Include,
 }
 
+/// What the tree builder does when the per-build file budget is exhausted.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BudgetExceededBehavior {
+    /// Stop descending and leave the remaining directories as unloaded
+    /// placeholders (lazy-loaded on demand). The build still succeeds with a
+    /// partial, breadth-first tree. This is the default for the shared file
+    /// tree, `@`-context, and skill discovery.
+    StopAndLazyLoad,
+    /// Abort the build and return [`BuildTreeError::ExceededMaxFileLimit`].
+    /// Use this for consumers that must not operate on a partial tree — e.g.
+    /// codebase embedding, where the file limit is an intentional cost cap.
+    FailFast,
+}
+
 /// Filesystem entry.
 #[derive(Debug, Clone)]
 pub enum Entry {
@@ -59,15 +74,7 @@ pub(crate) struct BuildTreeOptions<'a> {
     pub current_depth: usize,
     pub ignored_path_strategy: &'a IgnoredPathStrategy,
     pub ignored_path_interests: &'a [PathBuf],
-}
-
-impl<'a> BuildTreeOptions<'a> {
-    fn child(self) -> Self {
-        Self {
-            current_depth: self.current_depth + 1,
-            ..self
-        }
-    }
+    pub budget_exceeded_behavior: BudgetExceededBehavior,
 }
 
 #[derive(Debug, Clone, Copy, Hash, Eq, PartialEq)]
@@ -108,6 +115,9 @@ impl Entry {
     /// Builds a tree of entries from a given path, handling gitignored files and directories.
     /// After max_depth is reached, all children are lazy-loaded to prevent deeply nested trees.
     /// IgnoredPathStrategy determines what happens when ignored files are encountered.
+    /// `budget_exceeded_behavior` controls what happens once the file budget is
+    /// exhausted (see [`BudgetExceededBehavior`]).
+    #[allow(clippy::too_many_arguments)]
     pub fn build_tree(
         path: impl Into<PathBuf>,
         files: &mut Vec<FileMetadata>,
@@ -116,6 +126,7 @@ impl Entry {
         max_depth: usize,
         current_depth: usize,
         ignored_path_strategy: &IgnoredPathStrategy,
+        budget_exceeded_behavior: BudgetExceededBehavior,
     ) -> Result<Self, BuildTreeError> {
         Self::build_tree_with_ignored_path_interests_and_ancestor(
             path,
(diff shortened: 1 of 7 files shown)


kevinyang372:
* **#12166** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12166?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12166?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

kevinyang372:
cc @alokedesai

crates/repo_metadata/src/entry.rs
```diff
@@ -270,51 +312,64 @@ impl Entry {
...
+                        ) {
+                            Ok(EvaluatedEntry::File { ignored }) => {
+                                if quota == Some(0)
+                                    && options.budget_exceeded_behavior
+                                        == BudgetExceededBehavior::FailFast
+                                {
+                                    return Err(BuildTreeError::ExceededMaxFileLimit);
                                 }
-                                Err(_) => None,
+                                let metadata =
```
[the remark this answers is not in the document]
kevinyang372: This feels very edge casy and the quota is not a strict hard cap. I think it's fine to keep it as is

crates/repo_metadata/src/entry.rs
```diff
@@ -182,85 +195,114 @@ impl Entry {
...
+                    // path to a registered ignored-path interest (e.g. skill
+                    // provider directories) are always expanded so
+                    // discovery-critical files stay reachable. With `FailFast` we
+                    // keep descending and abort below as soon as a file would
+                    // exceed the budget.
+                    let should_expand = match options.budget_exceeded_behavior {
+                        BudgetExceededBehavior::FailFast => true,
+                        BudgetExceededBehavior::StopAndLazyLoad => {
+                            quota.is_none_or(|remaining| remaining > 0)
+                                || matches_ignored_path_interest(
```
moirahuang, line 269: curious why we let ignored past interests go through even if quota is 0?

crates/repo_metadata/src/entry.rs
```diff
@@ -182,85 +195,114 @@ impl Entry {
...
+                        continue;
+                    }
+
+                    let entries = match std::fs::read_dir(&job.path) {
+                        Ok(entries) => entries,
+                        Err(e) => {
+                            // Preserve existing behavior: failing to read the
+                            // root directory propagates, while an unreadable
+                            // nested directory is left as an unloaded placeholder.
+                            if job.is_root {
```
moirahuang, line 285: is this still the existing behavior we want to preserve?

crates/repo_metadata/src/local_model.rs
```diff
@@ -54,8 +56,15 @@ use crate::file_tree_update::{
...
-const MAX_FILES_PER_REPO: usize = 100_000;
+/// Maximum number of non-ignored files to index eagerly per repository.
+///
+/// This is a high safety ceiling, not the common case: gitignored directories
+/// are lazy placeholders and never consume this budget, so only repositories
+/// with an enormous number of *tracked* files reach it. When the budget is
+/// exhausted the builder stops descending breadth-first and leaves the
+/// remaining directories as unloaded placeholders (lazy-loaded on demand)
+/// rather than failing or collapsing the tree to a single level.
+const MAX_FILES_PER_REPO: usize = 200_000;
```
moirahuang, line 67: sorry i'm a bit confused by this, i thought we were going to just index the whole thing? or maybe set a memory limit, not a hardcoded max number of files?

moirahuang (approved):
approving to unblock

crates/repo_metadata/src/entry.rs
kevinyang372: This is the "special case" paths like skill directories

crates/repo_metadata/src/local_model.rs
kevinyang372: discussed offline

crates/repo_metadata/src/entry.rs
kevinyang372: Yeah I think we should preserve the existing behavior (at least for this PR)

PR #12176, opened 2026-06-03
Register non-recursive watchers for Linux

kevinyang372:
## Description
On Linux, navigating a remote (or local) session into a large non-git directory such as `/workspaces` registered a **recursive** filesystem watch over the entire subtree. Because a non-git parent has no root `.gitignore`, the gitignore-based prune (PR #12122) has nothing to prune, so every nested repo's `target/`, `node_modules/`, etc. got watched — 20,000+ inotify watches and the ~11 GB remote-server daemon memory blowup a customer reported.

This PR makes the watch for lazy-loaded (non-git) standalone roots as lazy as the directory tree itself, **on Linux only**: instead of one recursive watch on the root, we watch only the directories that are actually loaded, each `NonRecursive`, and add a watch for each subdirectory as the user expands it. Memory now scales with what the user expands rather than the whole subtree.

### Platform scoping
The per-directory inotify cost is Linux-specific. macOS (FSEvents) and Windows (ReadDirectoryChangesW) watch a tree recursively with a single OS handle, so recursive watching there is cheap and unaffected. Git repos remain recursively watched on all platforms (they rely on gitignore pruning for correctness).

### Key changes (`crates/repo_metadata/src/local_model.rs`)
- Introduced a `RepoWatchMode` enum that is the stored per-repo watch state: `RecursiveOnRoot`, or `LazyNonRecursive { watched_dirs }` which carries the set of directories currently watched under a lazy root (root + expanded subdirs).
- `add_repository_internal` takes the mode, maps it to the watcher's `RecursiveMode`, records it, and (when replacing a prior lazy registration, e.g. a lazy root upgraded to a git repo) unregisters stale per-directory watches.
- `index_lazy_loaded_path` registers the root `NonRecursive` on Linux (`Recursive` elsewhere).
- `load_directory` and the watcher-event handler register a `NonRecursive` watch on each newly loaded subdirectory and record it.
- `remove_repository` matches on the recorded mode to unregister every tracked per-directory watch (lazy) or just the root (recursive).
- Event processing is unchanged: a single `BulkFilesystemWatcher`/debouncer funnels all events, and `handle_watcher_event` routes each path to its owning root by longest-prefix match, so non-recursive watches need no new plumbing.

## Linked Issue

This is a follow-up to #12122 (gitignore-aware pruning for git-repo recursive watches); this PR bounds non-git lazy roots, which that prune cannot help.

## Testing
- Added unit tests in `crates/repo_metadata/src/local_model_tests.rs`:
  - `index_lazy_loaded_path_tracks_only_root` — lazy root records `LazyNonRecursive` with only the root on Linux, `RecursiveOnRoot` elsewhere.
  - `load_directory_tracks_expanded_subdir_for_lazy_root` — expanding a subdir adds its watch on Linux.
  - `recursive_repo_uses_recursive_watch_mode` — git repos record `RecursiveOnRoot` and aren't tracked as lazy.
  - `remove_lazy_loaded_path_clears_tracked_watches` — teardown clears all tracked watches and repo state.
- `cargo nextest run -p repo_metadata --features local_fs` (new + adjacent tests pass), `cargo check -p repo_metadata` (default, non-`local_fs`), `./script/format`, and `cargo clippy -p repo_metadata --features local_fs --all-targets -- -D warnings` all clean.
- Manual: built/deployed the remote-server daemon to a Linux devbox, `cd /workspaces`, and confirmed via `[watch_inode_debug]` logging (added in #12122) and `/proc/<pid>/fdinfo` that the watched-directory count drops from ~20,000 to a small number that grows only as folders are expanded; git repos opened at their root are unaffected.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed excessive memory usage (and inotify watch exhaustion) on Linux when a Warp session navigated into a large non-git directory; lazy directory roots are now watched incrementally as folders are expanded instead of recursively up front.

### DIFF
--- a/crates/repo_metadata/src/entry.rs
+++ b/crates/repo_metadata/src/entry.rs
@@ -1027,6 +1027,15 @@ pub fn repo_watch_filter(
 ) -> WatchFilter {
     let should_watch =
         move |path: &Path| should_watch_repo_directory(path, &gitignores, &force_included_paths);
+    // INVARIANT: gitignore lives only in the descend predicate
+    // (`should_watch_repo_directory`), never in the emit predicate. The descend
+    // predicate is consulted only by the inotify backend (Linux) at recursive
+    // registration time, so pruning gitignored dirs is purely a Linux
+    // watch-cost control. The emit predicate must NOT filter gitignored paths:
+    // on FSEvents/ReadDirectoryChangesW (macOS/Windows) the descend predicate is
+    // ignored and the recursive root watch still delivers gitignored events, and
+    // on Linux on-demand non-recursive watches likewise rely on those events
+    // flowing through. It only suppresses non-allowlisted `.git/` internals.
     WatchFilter::with_filter(
         Arc::new(should_watch),
         Arc::new(|path: &Path| !should_ignore_git_path(path)),
--- a/crates/repo_metadata/src/entry.rs
+++ b/crates/repo_metadata/src/entry.rs
@@ -1027,15 +1027,6 @@ pub fn repo_watch_filter(
 ) -> WatchFilter {
     let should_watch =
         move |path: &Path| should_watch_repo_directory(path, &gitignores, &force_included_paths);
-    // INVARIANT: gitignore lives only in the descend predicate
-    // (`should_watch_repo_directory`), never in the emit predicate. The descend
-    // predicate is consulted only by the inotify backend (Linux) at recursive
-    // registration time, so pruning gitignored dirs is purely a Linux
-    // watch-cost control. The emit predicate must NOT filter gitignored paths:
-    // on FSEvents/ReadDirectoryChangesW (macOS/Windows) the descend predicate is
-    // ignored and the recursive root watch still delivers gitignored events, and
-    // on Linux on-demand non-recursive watches likewise rely on those events
-    // flowing through. It only suppresses non-allowlisted `.git/` internals.
     WatchFilter::with_filter(
         Arc::new(should_watch),
         Arc::new(|path: &Path| !should_ignore_git_path(path)),
--- a/crates/repo_metadata/src/local_model.rs
+++ b/crates/repo_metadata/src/local_model.rs
@@ -4,9 +4,7 @@
 //! This module provides a singleton model that manages repository metadata across
 //! all repositories tracked by Warp.
 
-use std::collections::HashMap;
-#[cfg(feature = "local_fs")]
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
@@ -49,7 +47,6 @@ use crate::telemetry::RepoMetadataTelemetryEvent;
 use crate::{gitignores_for_directory, matches_gitignores, RepoMetadataError};
 cfg_if::cfg_if! {
     if #[cfg(feature = "local_fs")] {
-        use std::collections::HashSet;
         use notify_debouncer_full::notify::RecursiveMode;
         use crate::entry::repo_watch_filter;
         use crate::repositories::{DetectedRepositories, DetectedRepositoriesEvent};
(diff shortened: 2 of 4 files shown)


kevinyang372:
* **#12211** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12176** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12176?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12176?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12122** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12122?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/repo_metadata/src/local_model.rs
```diff
@@ -367,6 +395,30 @@ impl LocalRepoMetadataModel {
...
+                        // tree mutations are applied (keeps live-created dirs fresh).
+                        let new_dirs: Vec<StandardizedPath> = if matches!(
+                            model.repo_watch_modes.get(&repo_path),
+                            Some(RepoWatchMode::LazyNonRecursive { .. })
+                        ) {
+                            mutations
+                                .iter()
+                                .filter_map(|mutation| {
+                                    let path = match mutation {
+                                        FileTreeMutation::AddDirectorySubtree {
```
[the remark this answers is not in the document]
kevinyang372: This is a bug in the current lazy loading file tree behavior -- we shouldn't be eagerly expanding the subtree on file updates when the tree itself is lazy loaded

crates/repo_metadata/src/entry.rs
```diff
@@ -1075,6 +1075,15 @@ pub fn repo_watch_filter(
...
         move |path: &Path| should_watch_repo_directory(path, &gitignores, &force_included_paths);
+    // INVARIANT: gitignore lives only in the descend predicate
+    // (`should_watch_repo_directory`), never in the emit predicate. The descend
+    // predicate is consulted only by the inotify backend (Linux) at recursive
+    // registration time, so pruning gitignored dirs is purely a Linux
+    // watch-cost control. The emit predicate must NOT filter gitignored paths:
+    // on FSEvents/ReadDirectoryChangesW (macOS/Windows) the descend predicate is
+    // ignored and the recursive root watch still delivers gitignored events, and
+    // on Linux on-demand non-recursive watches likewise rely on those events
+    // flowing through. It only suppresses non-allowlisted `.git/` internals.
```
alokedesai: I recognize this comment is annoying (and it's a nit, you can ignore) but this comment doens't really make sense

crates/repo_metadata/src/local_model.rs
```diff
@@ -152,6 +152,47 @@ impl IndexedRepoState {
...
+/// extra non-recursive watches too.
+//
+// Without `local_fs` there is no watcher, so `NonRecursive` is never
+// constructed; the variant still exists to keep the watch-mode model intact.
+#[cfg_attr(not(feature = "local_fs"), allow(dead_code))]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum RootWatchMode {
+    /// A single recursive watch on the root covers the whole subtree. Used for
+    /// git repos (which rely on gitignore pruning in the watch descend filter)
+    /// and, on macOS/Windows, for lazy non-git roots where a recursive OS watch
```
alokedesai, line 182: Is there any reason we want to do this on Mac / windows for a lazy repo? It still means that we need to do a bunch of work on the cpu when a file changes, right?

crates/repo_metadata/src/local_model.rs
```diff
@@ -186,6 +227,11 @@ pub struct LocalRepoMetadataModel {
     /// alias rather than using a one-to-one bidirectional map.
     #[cfg(feature = "local_fs")]
     symlink_targets: HashMap<PathBuf, HashSet<SymlinkTarget>>,
+    /// How each tracked repository is registered with the filesystem watcher,
+    /// including any on-demand per-directory watches recorded for teardown. See
+    /// [`RepoWatch`].
+    #[cfg(feature = "local_fs")]
+    repo_watches: HashMap<StandardizedPath, RepoWatch>,
```
alokedesai, line 248: I feel like this is really just a sumtree of watched directories (that also includes the explicitly watched directories). Not blocking in any way, just the `RepoWatch` abstraction with extra_dirs is a little clunky IMO

crates/repo_metadata/src/local_model.rs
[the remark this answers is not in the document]
alokedesai: ohh so the code from the other PR removes this and fixes it at its source? if so, makes sense

alokedesai (approved):

alokedesai:
approved but did not review this very closely. Biggest thought is that the extra_dir logic is extremely complicated and makes the logic very hard to follow. I think

crates/repo_metadata/src/local_model.rs
kevinyang372: This avoids additional tree scans and keeping more states for non-recursive watchers

For Mac / Windows, there is little extra cost for registering a recursive watcher on the root node. I think it's worth keeping this behavior untouched

crates/repo_metadata/src/local_model.rs
kevinyang372: Yeah agreed

kevinyang372:
@alokedesai I tried iterating on a few different approaches to clean it up and it's not straightforward. I do think the best path forward is to follow the sumtree based approach

PR #12177, opened 2026-06-03
[QUALITY-793] Reduce agent event stream error log spam

cephalonaut:
## Description
Reduce client-side Error-level log spam from the reconnecting agent event SSE stream.

What changed:
- Gate agent event stream Error-level retry logs on error actionability.
- Emit Error only once when a failure streak crosses the configured threshold; later retries in the same streak remain Warn-level until recovery.
- Preserve typed SSE transport/status errors so non-actionable network/server conditions are classified correctly.
- Register `HttpStatusError` with `ErrorExt`, classifying 408, 429, and 5xx as non-actionable.

Why:
- Fixes QUALITY-793, where prolonged network or server-side failures could produce repeated client Error-level logs such as “Agent event stream failed 732 consecutive times”.

Plan: https://staging.warp.dev/drive/notebook/Xwz7CpHzilfHd1N5YbAllp
Conversation: https://staging.warp.dev/conversation/2bbce176-bb87-44f0-bcc2-9bcfc3fbce4e

## Linked Issue
https://linear.app/warpdotdev/issue/QUALITY-793/reduce-frequency-of-client-error-logging

## Testing
- `cargo fmt --all --check`
- `./script/format --check`
- `cargo test -p warp agent_events::driver_tests`
- `cargo check -p warp`
- `PATH=/tmp/warp-corepack:$PATH cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`
- `PATH=/tmp/warp-corepack:$PATH cargo clippy -p warp_completer --all-targets --tests -- -D warnings`
- `git --no-pager diff --check`

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/src/ai/agent_events/driver.rs
+++ b/app/src/ai/agent_events/driver.rs
@@ -6,6 +6,7 @@ use async_trait::async_trait;
 use futures::future::Either;
 use futures::StreamExt;
 use instant::Instant;
+use warp_core::errors::AnyhowErrorExt as _;
 use warpui::r#async::Timer;
 
 use crate::server::retry_strategies::is_transient_http_error;
@@ -69,7 +70,7 @@ pub(crate) struct AgentEventDriverConfig {
     /// before upstream infrastructure times it out (for example, before Cloud
     /// Run's 20-minute streaming timeout).
     pub proactive_reconnect_after: Option<Duration>,
-    /// Failure count at which reconnect logging is escalated from debug to warn.
+    /// Failure count at which actionable reconnect failures are reported at Error level.
     /// This only affects log severity; retry behavior stays the same.
     pub failures_before_error_log: usize,
 }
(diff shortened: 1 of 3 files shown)


app/src/ai/agent_events/driver.rs
```diff
@@ -214,16 +215,25 @@ impl AgentEventSource for ServerApiAgentEventSource {
...
                 Err(err) => {
-                    let anyhow_err = match &err {
-                        reqwest_eventsource::Error::InvalidStatusCode(status_code, _) => {
+                    let anyhow_err = match err {
+                        reqwest_eventsource::Error::InvalidStatusCode(status_code, response) => {
+                            let body = response
+                                .text()
+                                .await
+                                .unwrap_or_else(|err| format!("(no response body: {err:#})"));
                             let status_err = HttpStatusError {
```
vorporeal, line 224: seems a little odd that we're using `presigned_upload::HttpStatusError` here; i don't love that coupling.

app/src/server/server_api/presigned_upload.rs
```diff
@@ -26,6 +27,13 @@ pub struct HttpStatusError {
     pub body: String,
 }
 
+impl ErrorExt for HttpStatusError {
+    fn is_actionable(&self) -> bool {
+        !matches!(self.status, 408 | 429 | 500..=599)
```
vorporeal: while in some ways you could say that 5xx errors are "non-actionable", we generally treat them as actionable because they usually indicate some problem on _our_ end, that we are responsible for addressing (as opposed to a client networking issue).

given the fact that this is (apparently) widely used across modules (despite being in `presigned_upload.rs`), i don't think we should define 5xx as unconditionally non-actionable.

app/src/ai/agent_events/driver.rs
cephalonaut: I know - it appears to be widely used. I can move it somewhere convenient as a followup.

app/src/server/server_api/presigned_upload.rs
cephalonaut: Fair - I've relaxed it to exclude 5xx.

vorporeal (approved):

PR #12200, opened 2026-06-04
Support multi tab selection in vertical tabs

johnturcoo:
## Description
Adds shift-click range selection and cmd-click multi tab select to the vertical tabs sidebar. Clicking a tab while holding shift **selects** every tab between the currently active tab and the clicked tab, and renders an accent highlight on the selected range. Clicking a tab while holding command, individual **selects** this tab and renders an accent highlight. Clicking a tab without shift (or starting a drag, switching panes, etc.) clears the selection.

Further actions that allow creating groups from these selected tabs, moving/removing these tabs to/from a group will be supported in a follow up PR.

NOTE: UI is not polished here, and is going to be updated in a follow up PR.

## How it works

Added a new `ClickWithModifiersHandler` handler to differentiate regular click from shift clicks and command clicks (would love some eyes on this). In the vertical tabs pane, when an element is shift clicked, we dispatch a new `on_click_with_modifiers` action rather than setting focus to this tab/pane via a regular click. This sets a new flag on all tabs and panes within the given range (from current active tab to new element that is shift-clicked) to identify that this tab/pane is now selected as part of a range. A similar situation occurs for a command click, but marks the clicked tab as **selected** individually. This is cleared on next click, or drag action. If there is a group within the range, it is automatically expanded.

Rendering logic is updated so that tabs/panes are highlighted when they are part of a multi select (with an edge case for panes so that there is differentiation between hover and selected). Helper functions are added to `app/src/workspace/view/tab_grouping.rs` to support range selection actions, this is where all existing tab grouping logic is going to be ported to. Shift click is gated behind a feature flag so that behavior remains unchanged when it is off.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4642/shift-click-range-group-selection

## Testing

### Screenshots / Videos

[Demo video](https://www.loom.com/share/17c6cc1be36342a59555455e64d66600)

This video displays shift clicking a range of tabs, in both tab and pane view. It is cleared when another tab is selected or a drag begins. Also shows groups auto expanding when they are within a shift click range.

It also displays command clicking individual tabs to show they are also marked as selected.

### DIFF
--- a/app/src/workspace/view/tab_grouping.rs
+++ b/app/src/workspace/view/tab_grouping.rs
@@ -1,0 +1,87 @@
+use std::collections::HashSet;
+use warp_core::features::FeatureFlag;
+use warpui::ViewContext;
+
+use super::Workspace;
+use crate::workspace::tab_group::TabGroupId;
+use crate::workspace::util::PaneViewLocator;
+
+// TODO(johnturcoo) move tab grouping helpers here from workspace/view.rs.
+impl Workspace {
+    /// Clears the range selection.
+    pub(super) fn clear_tab_range_selection(&mut self, ctx: &mut ViewContext<Self>) {
+        let mut changed = false;
+        for tab in &mut self.tabs {
+            if tab.in_range_selected {
+                tab.in_range_selected = false;
+                changed = true;
+            }
+        }
+        if changed {
+            ctx.notify();
+        }
+    }
+
+    /// Selects the inclusive range between `anchor_index` and `clicked_index`,
+    /// expanding any collapsed groups the range crosses.
+    fn set_tab_range_selection(
+        &mut self,
+        anchor_index: usize,
+        clicked_index: usize,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        // Determine the bounds for our range selection.
+        let lo = anchor_index.min(clicked_index);
+        let hi = anchor_index.max(clicked_index);
+
+        // Identify groups in the selection range.
+        let crossed_group_ids: HashSet<TabGroupId> = self
+            .tabs
+            .get(lo..=hi)
+            .into_iter()
+            .flatten()
+            .filter_map(|tab| tab.group_id)
+            .collect();
+        let mut expanded_any = false;
+
+        // Expand any groups within the selected range, so user can see what they are selecting.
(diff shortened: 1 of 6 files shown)


peicodes:
From a product perspective, we we allow shift-select we probably also need to support ctrl-select

app/src/workspace/view/tab_grouping.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::workspace::tab_group::TabGroupId;
+use crate::workspace::util::PaneViewLocator;
+
+// TODO(johnturcoo) move tab grouping helpers here from workspace/view.rs.
+impl Workspace {
+    /// Clears the range selection.
+    pub(super) fn clear_tab_range_selection(&mut self, ctx: &mut ViewContext<Self>) {
+        let mut changed = false;
+        for tab in &mut self.tabs {
+            if tab.in_range_selected {
```
peicodes: we should have a concept of "selected" rather than "in range" so it's obvious that selection can happen for contiguous tabs

app/src/workspace/view/tab_grouping.rs
```diff
@@ -0,0 +1,87 @@
...
+                tab.in_range_selected = false;
+                changed = true;
+            }
+        }
+        if changed {
```
peicodes: nit: maybe we just clear all and notify for simplicity

app/src/workspace/view/tab_grouping.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Selects the inclusive range between `anchor_index` and `clicked_index`,
+    /// expanding any collapsed groups the range crosses.
+    fn set_tab_range_selection(
+        &mut self,
+        anchor_index: usize,
+        clicked_index: usize,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        // Determine the bounds for our range selection.
+        let lo = anchor_index.min(clicked_index);
```
peicodes: nit: let's be specific about what `hi` and `lo` actually are

peicodes (approved):

johnturcoo:
Decided to go with command click here rather than ctrl click, as this is what google chrome tabs and apple finder uses.

PR #12221, opened 2026-06-04
Avoid cloning whole file tree on view update & flatten entries

kevinyang372:
## Description

Switching tabs could beachball the app for users with large file trees. A `sample` of a hung client showed the main thread spending ~100% of its time under `FileTreeView::set_root_directories` → `update_directory_contents` → `rebuild_flatten_items_impl`, dominated by deep-cloning and dropping the entire file-tree map (`FileTreeEntry`, which wraps two large `HashMap`s). On a ~13 GB-footprint client these O(total-entries) clones cost seconds.

`FileTreeEntry` is cloned in two hot spots per displayed root on a tab switch:
1. The model→view sync (`root_dir.entry = state.entry.clone()`), which happens at ~8 sites in the view.
2. The flatten pass (`let entry_state = root_dir.entry.clone()`), cloned only to dodge a borrow conflict.

This PR wraps the inner store (`FileTreeMapStore`) in an `Arc` inside `FileTreeEntry` and mutates it via `Arc::make_mut` (copy-on-write). Cloning a `FileTreeEntry` is now an O(1) refcount bump; the full copy only happens when a holder actually mutates a shared store, and at most once per mutation batch. This makes both clone sites cheap and reduces tab-switch work to the O(visible-items) flatten.

### Why `Arc` instead of removing `Clone` from `FileTreeEntry`

The biggest reason is that **the file-tree view holds its own snapshot of the model's tree**: each `RootDirectory.entry` is an owned `FileTreeEntry` that the view copies from the model on every sync (`root_dir.entry = state.entry.clone()`). That snapshot is load-bearing — it lets the view read a stable tree during flatten, decide when to adopt new model state, and deliberately keep the old entry during a Pending re-index so the tree doesn't flash to a loading state.

The model itself never deep-clones the entry (it mutates in place), so `Clone` exists almost entirely to serve the view's snapshot. Removing `Clone` would force the view to stop owning a snapshot and instead rebuild the local sync around incremental updates (the remote model's `apply_repo_metadata_update` pattern) — a much larger rearchitecture: the local model would have to emit snapshots/deltas, the several `FullReplace` sync points would need converting, and the Pending/no-flicker handling preserved.

Wrapping the inner store in `Arc` keeps the existing snapshot semantics exactly (every holder still behaves as if it owns an independent copy) while making the copy O(1) until someone writes — a contained change that fixes the hot path without renegotiating the model/view contract. (It also keeps `Clone`, which `Arc::make_mut` requires, so the model can keep mutating its own entry in place via copy-on-write.)

### Why `Arc::make_mut` (copy-on-write) instead of interior mutability

The requirement here is **cheap snapshots with isolation**, not coordinated shared mutation — so interior mutability (`Arc<RwLock<…>>`, `RefCell`) would solve a problem we don't have while adding cost:

- **Snapshot semantics.** With a shared cell, the model and view would point at the same mutable store, so a model mutation would be observed by the view mid-read (out from under `flatten_entry_for_root`) and would break the intentional "keep old entry during Pending re-index" flicker handling. COW preserves value semantics: each holder behaves as if it owns an independent copy.
- **Reads are the hot path, and they're aliasing-shaped, not lock-shaped.** Flattening walks many nodes calling `get`/`child_paths` while mutating `item_states` (a different field). COW reads are plain `&` derefs with zero synchronization; a lock would add atomic traffic per node, and holding a read guard across the recursion risks re-entrancy/deadlock (`RefCell` would outright panic).
- **The mutation pattern is "owner mutates, readers snapshot,"** not concurrent shared writes — textbook COW. `make_mut` keeps the common in-place mutation O(1) when the entry is unshared and only forks when it's actually shared.
- **Thread-safety fit.** `make_mut` needs only `Clone`; `RefCell` is single-threaded and `RwLock`/`Mutex` buy coordination we don't need.

### Tradeoff

COW relocates rather than eliminates the steady-state clone: the view already re-clones its entry on every `FileTreeEntryUpdated` event, so after this change that clone is O(1) but the model's next mutation pays the copy via `make_mut`. Net per fs-update batch is unchanged for a single view and better when a repo is shown in multiple panes (1 copy instead of N), while the tab-switch path goes from two full clones to ~0. Peak memory is unchanged (a copy still transiently holds 2x the store).

## Linked Issue

## Testing

- `cargo test -p repo_metadata` — 91 passed, 0 failed (COW semantics preserved across insert/remove/rename/apply-update).
- `cargo check -p warp --features local_fs` — compiles cleanly.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed the app hanging when switching tabs in very large repositories.

### DIFF
--- a/crates/repo_metadata/src/file_tree_store.rs
+++ b/crates/repo_metadata/src/file_tree_store.rs
@@ -11,7 +11,11 @@ use crate::{BuildTreeError, Entry, FileId, FileMetadata, Repository};
 
 #[derive(Debug, Clone)]
 pub struct FileTreeEntry {
-    state_map: FileTreeMapStore,
+    // Wrapped in an `Arc` so cloning a `FileTreeEntry` is O(1) (a refcount
+    // bump) instead of deep-copying the whole tree. Mutations go through
+    // `Arc::make_mut`, which copies-on-write only when the store is actually
+    // shared with another holder (e.g. the model and a view).
+    state_map: Arc<FileTreeMapStore>,
     root_path: Arc<StandardizedPath>,
 }
 
@@ -40,7 +44,7 @@ impl FileTreeEntry {
     }
 
     pub fn rename_path(&mut self, path: &StandardizedPath, new_path: &StandardizedPath) -> bool {
-        self.state_map.rename_path(path, new_path)
+        Arc::make_mut(&mut self.state_map).rename_path(path, new_path)
     }
 
     pub fn load_at_path(
@@ -48,11 +52,11 @@ impl FileTreeEntry {
         path: &StandardizedPath,
         gitignores: &mut Vec<Gitignore>,
     ) -> Result<(), BuildTreeError> {
-        self.state_map.load_at_path(path, gitignores)
+        Arc::make_mut(&mut self.state_map).load_at_path(path, gitignores)
     }
 
     pub fn insert_entry_at_path(&mut self, path: Arc<StandardizedPath>, entry: Entry) {
-        self.state_map.insert_entry_at_path(path, entry);
+        Arc::make_mut(&mut self.state_map).insert_entry_at_path(path, entry);
     }
 
     pub fn child_paths(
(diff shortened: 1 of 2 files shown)


kevinyang372:
* **#12221** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12221?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12221?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

moirahuang (approved):

PR #12229, opened 2026-06-04
Support multi select grouping actions for vertical tabs

johnturcoo:
## Description
Adds group operations to the vertical tabs multi-selection from [the previous PR](https://github.com/warpdotdev/warp/pull/12200). When the user has multiple tabs selected (via shift-click or cmd-click), right-clicking any selected tab opens a new multi-tab menu with three actions: Create group from tabs, Move to group, and Remove from group. 

Create group from tabs turns the selection into a new tab group at the top of the tab list. Move to group folds the selection into an existing group, anchored at the group's first member so the group doesn't visually move. Remove from group is available only when every selected tab shares the same group; the removed tabs land just below the group's remaining members, or in the same position if no remaining members existing.

NOTE: UI is not polished here and is going to be updated in a follow up PR.

## How it works

Added new workspace actions for creating group, moving to group and removing from group when several tabs are selected. Helper functions containing the logic can be found in  `app/src/workspace/view/tab_grouping.rs`. Important note:
- Moving to group is possible if there is another group available for at least one of the selected tabs (ie Group 1 contains members A,B,C, tab list contains A,B,C,D. Selecting all 4 tabs allows move to group 1, since tab D can be moved to group 1). 
All other behavior for these actions should be as expected. Empty groups are deleted after any of these actions.

Added two new actions for clearing all selected tabs, and for toggling the multi-tab selection right click menu, which displays the "Move to group", "Remove from group" and "Create group from tabs" options. This menu is displayed on right click of a `selected` tab, otherwise the regular single-tab menu appears. Right click on a tab or pane branches on `is_in_multi_tab_selection` to determine which menu to display.

The "move to group" sidecar menu is now used for both the single-tab menu as well as the multi-tab selection menu, rendering is extracted into a shared function `add_move_to_group_sidecar_overlay`. 

The only other thing to note here: Since we support cmd + click to select any individual tab, the active tab is always assumed to be selected when any other member is selected, but is only marked as selected for a range selection. This is important when determining which menu to display when right clicking the active tab, without needing to do additional bookkeeping on every cmd + click that toggles selected tabs.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4661/group-creation-from-multi-select-for-vertical-tabs

## Testing

### Screenshots / Videos

[Demo video](https://www.loom.com/share/adbfa7c9fd794863995655316c4cdc4b)

This demo video shows a mix of shift + click range selection, cmd + click individual tab selection and a mix of different actions from the multi tab selection menu

### DIFF
--- a/app/src/workspace/view/tab_grouping.rs
+++ b/app/src/workspace/view/tab_grouping.rs
@@ -1,10 +1,14 @@
 use std::collections::HashSet;
 
+use itertools::{Either, Itertools};
 use warp_core::features::FeatureFlag;
-use warpui::ViewContext;
+use warpui::{EntityId, UpdateView, ViewContext};
 
-use super::Workspace;
-use crate::workspace::tab_group::TabGroupId;
+use super::{group_member_indices, Workspace};
+use crate::menu::{MenuItem, MenuItemFields};
+use crate::tab::MOVE_TO_GROUP_LABEL;
+use crate::workspace::action::{TabContextMenuAnchor, WorkspaceAction};
+use crate::workspace::tab_group::{TabGroup, TabGroupId};
 use crate::workspace::util::PaneViewLocator;
 
 // TODO(johnturcoo) move tab grouping helpers here from workspace/view.rs.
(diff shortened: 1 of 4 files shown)


app/src/workspace/view/vertical_tabs.rs
```diff
@@ -480,11 +481,21 @@ fn render_pane_row_element(
...
-            });
+            let anchor = TabContextMenuAnchor::Pointer(position);
+            if is_in_multi_tab_selection {
+                ctx.dispatch_typed_action(WorkspaceAction::ToggleTabSelectionRightClickMenu {
+                    tab_index,
+                    anchor,
+                });
+            } else {
+                // Right-clicking outside the multi-selection cancels it.
+                ctx.dispatch_typed_action(WorkspaceAction::ClearTabMultiSelection);
```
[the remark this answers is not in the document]
johnturcoo: This situation is impossible. When the multi tab selection menu is open, right clicking on another tab is ignored.

app/src/workspace/view/tab_grouping.rs
```diff
@@ -97,4 +101,379 @@ impl Workspace {
...
+    /// The active tab index is always included if any other tab is marked
+    /// as selected. This is to handle the edge case where we only mark other
+    /// tabs as selected via command click.
+    fn selected_tab_indices(&self) -> Vec<usize> {
+        let any_flagged = self.tabs.iter().any(|tab| tab.in_multi_selection);
+        // If no tab is part of the multi selection, return empty list.
+        if !any_flagged {
+            return Vec::new();
+        }
+        // Otherwise, the active tab must always be part of the multi tab selection.
```
peicodes, line 115: This might be contentious with the following user story:
- I have 5 tabs, 1 active
- I'm just organizing my tabs and moving some to its own group with cmd-click
- My active tab is not one of the ones I want to move

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -2248,11 +2271,23 @@ fn render_tab_group_internal(
         stack.finish()
     })
     .on_right_click(move |ctx, _, position| {
-        ctx.dispatch_typed_action(WorkspaceAction::ToggleVerticalTabsPaneContextMenu {
-            tab_index,
-            target: VerticalTabsPaneContextMenuTarget::ActivePane(active_pane_context_menu_target),
-            position,
-        });
+        let anchor = TabContextMenuAnchor::Pointer(position);
```
peicodes, line 2267: nit: we can maybe extract to a "on_right_click" handler if they're identical

peicodes (approved):

PR #12230, opened 2026-06-04
Add remote support for git operations

MaggieShan:
## Description
* Adds support for git operations over remote sessions
* Introduces new host-scoped proto RPCs 
  * Generate commit message, commit, push, view pr, create pr, and fetch committed files
* Moves git ops logic to a shared file for the remote daemon and local client 
* Wires `FileChangeEntry` through `DiffMetadataAgainstBase` and `Commit` to track per file diff stats that are rendered in the git dialog 

## Linked Issue
[APP-4512](https://linear.app/warpdotdev/issue/APP-4512/wire-up-git-ops-with-remote-code-review)

## Testing
https://www.loom.com/share/112c2add953147d7b6be857b5f655abf

## Agent Mode

CHANGELOG-NEW-FEATURE: Git operations are now supported on remote sessions


app/src/code_review/git_dialog/pr.rs
```diff
@@ -99,6 +72,27 @@ pub(super) fn new_state(
...
+/// backend keeps this metadata fresh via its filesystem watcher, so the box
+/// no longer needs a dedicated working-tree read. Safe to call on open and on
+/// every metadata refresh.
+pub(super) fn refresh_file_changes_from_metadata(
+    me: &mut GitDialog,
+    ctx: &mut ViewContext<GitDialog>,
+) {
+    let entries = me
+        .diff_state_model()
+        .read(ctx, |model, ctx| model.branch_file_entries(ctx).to_vec());
```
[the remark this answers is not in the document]
MaggieShan: Good call out - fixed with an explicit proto call to fetch the right data

app/src/code_review/diff_state/mod.rs
```diff
@@ -663,6 +739,145 @@ impl DiffStateModel {
...
+    // ── Remote git operations (async; results arrive via events) ─────
+    //
+    // These dispatch the op and return immediately; the outcome is not
+    // fire-and-forget but arrives later as a `DiffStateModelEvent` that the
+    // dialog/view observes (`GitOpCompleted` / `MetadataRefreshed` /
+    // `CommitMessageGenerated` / `BranchesReceived`).
+
+    /// Runs a commit chain on the remote host. The result arrives via
+    /// `DiffStateModelEvent::GitOpCompleted`. No-op for local repos.
+    pub(crate) fn git_commit_chain(
```
kevinyang372, line 743: It's a bit weird design wise that we have this operation that only works for the remote case

How's it modeled locally? I would consider branching at the local callsites instead of here or at least change the naming so it's clearer this is remote only

app/src/code_review/diff_state/mod.rs
```diff
@@ -663,6 +739,145 @@ impl DiffStateModel {
...
+                    ctx,
+                );
+            }),
+        }
+    }
+
+    /// Issues an AI commit-message generation request on the remote host. The
+    /// result arrives via `DiffStateModelEvent::CommitMessageGenerated`. No-op
+    /// for local repos, which generate the message client-side in the dialog.
+    pub(crate) fn generate_commit_message(
```
kevinyang372, line 777: ditto on here and below

crates/remote_server/proto/diff_state.proto
```diff
@@ -167,6 +181,11 @@ message DiffMetadata {
   repeated Commit unpushed_commits = 6;
   optional string upstream_ref = 7;
   optional PrInfo pr_info = 8;
+  // Field 9 (has_staged_changes) was removed once the commit dialog stopped
```
kevinyang372: I am confused by this comment -- why do we need this?

crates/remote_server/src/manager.rs
```diff
@@ -904,6 +967,234 @@ impl HostRequestHandle {
...
+    // so the raw git/gh message reaches the client's `user_facing_git_error`.
+
+    /// Runs the commit chain (commit, then optionally push, then optionally
+    /// create-PR) on the remote host in a single round trip, returning the
+    /// post-chain delta (refreshed unpushed commits + upstream) and any created
+    /// PR. The daemon sequences the underlying git / gh subprocesses host-local
+    /// (see `handle_git_commit_chain`), so the SSH link carries one request/response
+    /// instead of the 2–3 a client-side chain would send.
+    #[allow(clippy::too_many_arguments)]
+    pub async fn git_commit_chain(
```
kevinyang372, line 985: It's unclear what this name actually means in here and proto -- I would consider naming this commit_and_create_pr?

kevinyang372 (approved):
Nice! Didn't review too closely but we already aligned on the approach high-level

crates/remote_server/src/manager.rs
MaggieShan: The main reason it's named `git_commit_chain` is because it optionally chains a push and/or create pr, I think it'd actually be more confusing if we renamed it to `commit_and_create_pr` - this naming also matches the existing git helper fns elsewhere throughout the code

crates/remote_server/proto/diff_state.proto
MaggieShan: Removed!

app/src/code_review/diff_state/mod.rs
MaggieShan: Good callout - agreed that it was a bit weird.

Previously the local version called the git action helpers directly from the dialog and used the result to update the diff state model. I've since updated the logic to similarly go through the local diff state model instead, which should be much clearer.

Issue #9015, opened 2026-04-04
View Jupyter Notebooks


lisah2u:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

I use AI to generate and modify Jupyter notebooks that are shared with others via Colab and elsewhere. Typically, I use Claude in vscode — but don’t need the overhead of a vscode fork. It would be incredibly helpful if I could view them (or preview, if you like) in warp, not as JSON, but as human-readable notebooks. 

### Is your feature request related to a problem? Please describe.

Warp should be more friendly to data scientists by making it possible to view Jupyter notebooks properly. 

### Operating system (OS)

macOS

### How important is this feature to you?

4

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

Issue #5190, opened 2024-07-30
Support Jupyter (.ipynb) and Pluto (.jl) notebooks in Warp


dannyneira:
### Dupe Check

- [X] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3A%22Feature+requests%22) and there are no duplicates

### Describe the solution you'd like?

I was just curious whether I could open aJupyter (.ipynb) and Pluto (.jl) notebooks on the Notebooks panel.

### Is your feature request related to a problem? Please describe.

 I see unless it is .yaml or .md, that is not possible.

### How important is this feature to you?

5 (Can't work without it!)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

PR #12242, opened 2026-06-05
Support rendering jupyter notebooks

evelyn-with-warp:
## Description

Rendering jupyter notebook with `FileNotebookView`, with converting jupyter notebook json to markdown
related issue: 
- https://github.com/warpdotdev/warp/issues/5190
- https://github.com/warpdotdev/warp/issues/9015
## Linked Issue

## Testing

### Screenshots / Videos

https://www.loom.com/share/48cc3b6ad02e4d7686ab2c7c8e759ffc
## Agent Mode


crates/ipynb_parser/src/lib.rs
kevinyang372, line 1: I don't think we should convert from ipynb -> markdown -> formatted text

This essentially doubles the number of parsing we need to do with no clear benefits. I would follow the previous pattern of `from_markdown` implementation in Buffer and create a similar `from_ipynb` method. I would also consider creating a ipynb parser crate since it can be detangled from the core app logic

kevinyang372 (changes requested):
This looks great! I think my main feedback is we could reduce the change into more manageable components. On a high-level, I would break it down into:
1. Ipynb parser and integrate that into `Buffer::from_ipynb` to bootstrap a buffer from ipynb contents
2. Wire existing app-level logic to support opening an .ipynb file in our file notebook view

PR #12255, opened 2026-06-05
[QUALITY-733] Remove OrchestrationPillBar and OrchestrationViewerPillBar feature flags

cephalonaut:
## Description

Remove the `OrchestrationPillBar` and `OrchestrationViewerPillBar` feature flags, which were already in the `default` feature set and enabled for all builds.

## What

- Remove both variants from the `FeatureFlag` enum and all Cargo feature declarations
- Remove `is_enabled()` runtime checks throughout the codebase, keeping the code paths that were active when the flags were on
- Remove the legacy `ChildAgentStatusCard` view, which was only rendered when `OrchestrationPillBar` was OFF (now dead code)
- Remove the flag-guarded early return in `inherit_share_for_local_child` and simplify the condition in `maybe_add_parent_navigation_card`
- Update tests to remove the now-unnecessary feature flag overrides; delete the test that verified the disabled-flag behavior

Also picks up `OrchestrationV2` removal and `orchestration_viewer_streamer` addition from master.

## Why

Both flags were already in the `default` feature set, making all the `is_enabled()` checks permanently true. Removing them reduces technical debt and dead code.

## Agent Mode

CHANGELOG-NONE

---
[Oz conversation](https://staging.warp.dev/conversation/2cfacd32-329a-4c08-9a0c-780c618986e3)


advait-m (approved):
nice! agree good to clean-up now

PR #12257, opened 2026-06-05
Support multi tab selection actions for horizontal tabs

johnturcoo:
## Description
Adds shift-click range selection, cmd-click multi tab select, and the multi-tab right-click menu (create group from tabs, move to group, remove from group) to the horizontal tab bar. Clicking a tab while holding shift selects every tab between the currently active tab and the clicked tab, and renders an accent highlight on the selected range. Clicking a tab while holding command individually selects that tab and renders the same highlight. Right-clicking a tab that's part of a multi-tab selection opens the multi-tab menu. Plain clicks (or drags) clear the selection.

Follow up to [this PR](https://github.com/warpdotdev/warp/pull/12229)

This brings the horizontal tab bar to feature parity with the multi-select + group actions already available on the vertical tabs panel. Everything is gated behind the tab grouping feature flag, so behavior remains unchanged when it is off.

## How it works

This PR reuses all the existing multi tab selection logic from vertical tabs. Most of the changes are rendering and variable plumbing: threads `is_in_multi_tab_selection` through TabComponent, branches the existing right-click handler on it (chooses to display multi-tab menu vs single-tab menu), and extends the shared multi-tab selection menu render block in view.rs to fire for the horizontal tab bar too. 

One thing worth flagging: I added a new `MouseDownWithModifiersHandler` to Hoverable. Horizontal tabs originally used `on_mouse_down` rather than `on_click` (where the existing modifier handler lives) for dispatching click actions on tabs. I chose to reimplement this for mouse down rather than switching to this to an `on_click` to preserve existing behavior and UX. For example, currently a tab is set as active on mouse down, so starting a tab drag makes it active. If we changed this to `on_click` rather than using `on_mouse_down` this would change the UX which I believe is a regression. 

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4654/group-creation-from-multi-select-for-horizontal-tabs

## Testing

### Screenshots / Videos

[Demo video](https://www.loom.com/share/fa9909a4866842a6a12f767c609b27f1)

This demo video shows several different multi tab selection scenarios via shift-click and cmd-click, as well as different combinations of actions from the multi tab selection menu.

### DIFF
--- a/app/src/tab.rs
+++ b/app/src/tab.rs
@@ -772,6 +772,12 @@ pub struct TabComponent<'a> {
     grouped_member: bool,
     /// Set when this member is the sole tab in its group.
     sole_grouped_member: bool,
+    /// Locator pointing at this tab's focused pane.
+    locator: PaneViewLocator,
+    /// True when this tab is part of a multi-tab (count > 1) selection. Drives
+    /// both the in-selection highlight and the right-click menu dispatch
+    /// (multi-tab menu vs single-tab menu).
+    is_in_multi_tab_selection: bool,
 }
 
 /// Structure that holds TabComponent styles.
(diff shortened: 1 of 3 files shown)


peicodes:
I would highlight this behavioral difference for mousedown vs onclick for vtabs vs htabs, and tag ZB

peicodes (approved):
We didn't hear back from ZB, will merge first and clarify after since this does not change existing behavior

PR #12268, opened 2026-06-05
fix(voice): abort voice recording on Ctrl key chords; skip insert on empty transcription

### LINKED ISSUE #11415: Voice input active state blocks Ctrl key chords (e.g. Ctrl+V paste does nothing while voice is recording)
## Describe the bug

When the voice input toggle key is a modifier key (e.g. Ctrl), pressing a key chord that includes that modifier while voice recording is active does not work correctly. The chord fires against an editor that is still in the voice-recording (non-editable) state, so the action either has no effect or behaves unexpectedly.

A secondary but related issue: if voice input is started and then stopped without any speech being detected, the transcription returns an empty string. That empty string was being passed to `user_insert`, which can inadvertently clear any existing text selection in the editor.

## To reproduce

**Ctrl-chord conflict:**
1. Configure Ctrl (or another modifier) as the voice input toggle key.
2. Open a terminal input or AI input editor.
3. Type some text so the editor has content.
4. Press and hold Ctrl to start voice recording.
5. While still holding Ctrl, press Shift+V (i.e. issue the Ctrl+Shift+V / paste chord).

**Expected:** Voice recording is aborted and the paste action inserts clipboard content normally.
**Actual:** The editor remains in the voice-recording state; the paste is swallowed or has no effect.

---

**Empty transcription clears selection:**
1. Configure voice input with Ctrl as the toggle.
2. Select some existing text in the editor.
3. Tap Ctrl briefly (start + immediately stop voice recording) without speaking.
4. Wait for transcription to return.

**Expected:** The selection is preserved; the editor is unchanged.
**Actual:** The empty transcription result calls `user_insert("")`, which replaces the selection with nothing, effectively deleting it.

## Expected behavior

- Any key chord pressed while voice is recording should first abort the recording, then allow the chord to proceed normally.
- An empty transcription result should be treated as a no-op; the editor buffer must not be modified.

## Operating system (OS)

Linux (Ubuntu), but the voice input code path is cross-platform so this likely affects macOS and Windows as well.

No — the paste workaround is to release Ctrl first and then use Ctrl+V, but the behavior is surprising and easy to trigger accidentally.

Yes — this is specific to Warp's voice input feature.

kyledross:
Fixes #11415

## Description
This PR fixes two voice-input editor issues:
- Keybound editing actions could fire while voice recording still held the editor in a non-editable state.
- Empty or whitespace-only transcription results could replace an active selection with nothing.

## Changes
- Centralized voice-cancel behavior in `EditorView::handle_action`:
  - Added `EditorAction::aborts_voice_input()` (allowlist of editing/submit/mutation actions).
  - If voice input is active and `action.aborts_voice_input()` is true, `handle_action` now calls `stop_voice_input(true, ctx)` before dispatching the action.
- Removed narrow per-action/per-path abort handling that this central guard supersedes.
- Updated transcription insertion guard in `apply_transcribed_voice_input` to skip insertion for blank/whitespace results using `trim().is_empty()`.

## Tests
Added/updated voice tests in `app/src/editor/view/voice_tests.rs`:
- `test_aborts_voice_input_predicate_for_passive_and_editing_actions`
- `test_handle_action_aborts_active_voice_for_editing_action`
- `test_handle_action_keeps_active_voice_for_passive_action`
- `test_blank_transcription_does_not_modify_buffer` (now explicitly covers an active selection)
- `test_non_empty_transcription_inserts_text`
- `test_transcription_error_does_not_modify_buffer`

## Local Validation
- `./script/format`
- `cargo check -p warp`
- `cargo test -p warp --features voice_input editor::view::voice::tests::`
- `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings` (blocked locally in this environment by missing Node tooling for `command-signatures-v2` JS build)

## Agent Mode

### DIFF
--- a/app/src/editor/view/element.rs
+++ b/app/src/editor/view/element.rs
@@ -457,18 +457,6 @@ impl EditorElement {
                 ctx.dispatch_typed_action(EditorAction::ToggleVoiceInput(
                     voice_input::VoiceInputToggledFrom::Key { state: *state },
                 ));
-            } else if matches!(state, KeyState::Pressed)
-                && self.view_snapshot.voice_input_state.is_active()
-            {
-                // A modifier key other than the voice toggle key was pressed while voice is
-                // recording. Abort voice so that the resulting key chord can proceed normally.
-                //
-                // Example: if Ctrl is the voice toggle key and the user presses Ctrl+Shift+V
-                // to paste, voice is aborted when Shift is pressed. This ensures the editor
-                // is editable by the time V is pressed and the paste keybinding fires.
-                // (Keybindings are matched before element event handlers, so the abort must
-                // happen before the non-modifier key is pressed.)
-                ctx.dispatch_typed_action(EditorAction::AbortVoiceInput);
             }
         }
     }
--- a/app/src/editor/view/element.rs
+++ b/app/src/editor/view/element.rs
@@ -457,6 +457,18 @@ impl EditorElement {
                 ctx.dispatch_typed_action(EditorAction::ToggleVoiceInput(
                     voice_input::VoiceInputToggledFrom::Key { state: *state },
                 ));
+            } else if matches!(state, KeyState::Pressed)
+                && self.view_snapshot.voice_input_state.is_active()
+            {
+                // A modifier key other than the voice toggle key was pressed while voice is
+                // recording. Abort voice so that the resulting key chord can proceed normally.
+                //
+                // Example: if Ctrl is the voice toggle key and the user presses Ctrl+Shift+V
+                // to paste, voice is aborted when Shift is pressed. This ensures the editor
+                // is editable by the time V is pressed and the paste keybinding fires.
+                // (Keybindings are matched before element event handlers, so the abort must
+                // happen before the non-modifier key is pressed.)
+                ctx.dispatch_typed_action(EditorAction::AbortVoiceInput);
             }
         }
     }
--- a/app/src/editor/view/mod.rs
+++ b/app/src/editor/view/mod.rs
@@ -1093,6 +1093,11 @@ pub enum EditorAction {
     EmacsBinding,
     #[cfg(feature = "voice_input")]
     ToggleVoiceInput(voice_input::VoiceInputToggledFrom),
+    /// Abort any active voice recording without transcribing the captured audio.
+    /// Used to cancel voice input when a key chord is pressed while the voice toggle
+    /// key is held (e.g., Shift pressed before Ctrl+Shift+V).
+    #[cfg(feature = "voice_input")]
+    AbortVoiceInput,
     AttachFiles,
     SetAIContextMenuOpen(bool),
     ReadAndProcessImagesAsync {
@@ -1116,6 +1121,7 @@ impl EditorAction {
                 | EditorAction::TryToShowXRay(_)
                 | EditorAction::HideXRay
                 | EditorAction::Select(_)
+                | EditorAction::AbortVoiceInput
         )
     }
 
@@ -4210,6 +4216,13 @@ impl EditorView {
     }
 
     pub fn paste(&mut self, ctx: &mut ViewContext<Self>) {
+        // If voice input is recording, abort it first so the editor is editable for the paste.
+        // This handles the case where the voice toggle key (e.g. Ctrl) is held while the user
+        // presses a paste shortcut (e.g. Ctrl+V).
+        #[cfg(feature = "voice_input")]
+        if self.is_voice_input_active() {
+            self.stop_voice_input(true /* cancel_transcription */, ctx);
+        }
         // If this editor does not delegate paste handling, insert clipboard text content.
         // When paste handling is delegated, the parent view (e.g. the terminal input) is
         // responsible for processing the paste.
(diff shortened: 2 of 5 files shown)


advait-m:
ahaha, added label

/oz-review

advait-m:
> @advait-m
> 
> I ran into an unexpected error while working on this.
> 
> _Powered by [Oz](https://oz.warp.dev)_

😢 haha, I will take a look at this + review PR locally

app/src/editor/view/mod.rs
```diff
@@ -4175,6 +4179,13 @@ impl EditorView {
     }
 
     pub fn paste(&mut self, ctx: &mut ViewContext<Self>) {
+        // If voice input is recording, abort it first so the editor is editable for the paste.
```
advait-m: this works but it's whack-a-mole. cut (⌘X) and undo/redo (⌘Z), on mac, hit the exact same "non-editable while recording" problem and aren't covered, so bug 1 is only partly fixed. 

since every keybound action already funnels through `handle_action`, I'd think we could rather do one guard at the top of it that aborts voice for any non-voice action, then collapse the `paste()`/`handle_ctrl_c`/`escape` aborts into it. 

should fix the whole class in one spot. abort restores editability synchronously so the action still runs editable in the same call.

app/src/editor/view/voice.rs
```diff
@@ -493,7 +496,9 @@ impl EditorView {
         match result {
             Ok(transcribe_response) => {
                 log::debug!("Transcribed voice input: {transcribe_response:?}");
-                self.user_insert(&transcribe_response, ctx);
+                if !transcribe_response.is_empty() {
```
advait-m: nit: `.trim().is_empty()` to avoid whitespace transcript?

app/src/editor/view/element.rs
```diff
@@ -457,6 +468,18 @@ impl EditorElement {
...
+            {
+                // A modifier key other than the voice toggle key was pressed while voice is
+                // recording. Abort voice so that the resulting key chord can proceed normally.
+                //
+                // Example: if Ctrl is the voice toggle key and the user presses Ctrl+Shift+V
+                // to paste, voice is aborted when Shift is pressed. This ensures the editor
+                // is editable by the time V is pressed and the paste keybinding fires.
+                // (Keybindings are matched before element event handlers, so the abort must
+                // happen before the non-modifier key is pressed.)
+                ctx.dispatch_typed_action(EditorAction::AbortVoiceInput);
```
advait-m: I think we don't need this if we unify, smth like this

```
fn handle_action(&mut self, action: &EditorAction, ctx: &mut ViewContext<Self>) {
    #[cfg(feature = "voice_input")]
    if self.is_voice_input_active() && action.aborts_voice_input() {
        self.stop_voice_input(true /* cancel_transcription */, ctx);
    }

    match action { /* unchanged */ }
}
```

with
```
impl EditorAction {
    #[cfg(feature = "voice_input")]
    fn aborts_voice_input(&self) -> bool {
        // Any real editor command while recording means "not dictating" → abort.
        // Exclude the voice actions themselves.
        !matches!(self, EditorAction::ToggleVoiceInput(_) | EditorAction::AbortVoiceInput)
    }
}
```

app/src/editor/view/mod.rs
```diff
@@ -1098,6 +1111,7 @@ impl EditorAction {
                 | EditorAction::TryToShowXRay(_)
                 | EditorAction::HideXRay
                 | EditorAction::Select(_)
+                | EditorAction::AbortVoiceInput
```
advait-m: see comment above, I don't think we need to add this if we unify (can just use function)

advait-m (commented):
great fixes! A few (relatively small) comments - I think we can make sure we fix this for other actions besides paste too! @kyledross 

Tested paste fix locally and saw it working well

Feel free to ping me here once it's ready for another look!

app/src/editor/view/element.rs
advait-m: though I guess we need to be careful with the predicate here..  excluding only the voice actions means passive stuff aborts recording too. `TryToShowXRay`/`HideXRay` fire on mouse hover and `Scroll` on the wheel, so just moving the mouse while dictating would cancel the session. i'd also exclude `Scroll`/`TryToShowXRay`/`HideXRay`/`Focus` (≈ the passive set) but keep `UnhandledModifierKey` aborting. 

fwiw it's almost the same set as `should_report_active_cursor_position_updated` (mod.rs:1105), minus `UnhandledModifierKey` - would be worth a shared helper.

advait-m: thinking about it more, an allowlist could be safer too, to prevent any future actions causing problems:
```
fn aborts_voice_input(&self) -> bool {
    matches!(self,
        // text insertion / submit
        UserInsert(_) | VimUserInsert(_) | Tab | ShiftTab
        | Yank | Newline | CtrlEnter | ShiftEnter | AltEnter | Enter | CmdEnter
        | InsertLastWordPrevCommand | InsertNonExpandingSpace | InsertAutosuggestion
        // clipboard mutations
        | Cut | Paste | MiddleClickPaste
        // delete family
        | Delete | Backspace | VimBackspace
        | CutWordLeft | CutWordRight | DeleteWordLeft | DeleteWordRight
        | CutAllRight | CutAllLeft | DeleteAllRight | DeleteAllLeft
        | ClearLines | ClearAndCopyLines | CtrlC
        // history / structure edits
        | Undo | Redo | Fold | Unfold | FoldSelectedRanges | AddNextOccurrence
        | AddCursorAbove | AddCursorBelow
        // file attach (also needs editable target)
        | AttachFiles | DragAndDropFiles(_)
        // escape + deliberate-but-unbound chord (avoid stuck recording)
        | Escape | UnhandledModifierKey(_)
    )
}
```

kyledross:
Updated this PR with the latest fixes and test coverage improvements.

- Pushed `94dc378f` to the branch.
- Strengthened `test_blank_transcription_does_not_modify_buffer` so it now explicitly covers the active-selection case before applying blank/whitespace transcription.
- Updated the PR description so it matches the current implementation and test set.

Ready for re-review.

kyledross:
Quick update: this branch now includes the requested refactor and follow-up test hardening.

- Voice-cancel behavior is centralized in `handle_action` via `EditorAction::aborts_voice_input()` (allowlist approach).
- Empty/whitespace transcription handling now skips insert (`trim().is_empty()`).
- Added/updated voice tests in `app/src/editor/view/voice_tests.rs`, including explicit coverage for the active-selection blank-transcription case.
- Rebased onto latest `upstream/master` and pushed; rebase completed cleanly with no merge conflicts.

Validation run locally:
- `./script/format`
- `cargo check -p warp`
- `cargo test -p warp --features voice_input editor::view::voice::tests::`

It is ready for review.

kyledross:
Following-up on this PR.  It is ready for review.

PR #12272, opened 2026-06-05
feat: make repo pulling by agents done in parallel

dmichelin:
## Description

This PR makes repo pulling in agent mode done in parallel rather than done sequentially. Oz environments with many repos can take a long time currently to start up, and much of that time is due to pulling in multiple large repos one after another. 
## Linked Issue
https://linear.app/warpdotdev/issue/REMOTE-1867/parallelize-environment-repo-cloning-to-speed-up-startup

## Testing

### Screenshots / Videos

Before:
image

After:
image

## Agent Mode

### DIFF
--- a/app/src/ai/agent_sdk/driver/environment.rs
+++ b/app/src/ai/agent_sdk/driver/environment.rs
@@ -269,6 +269,11 @@ fn build_parallel_clone_command(repos: &[GithubRepo], shell_type: ShellType) -> 
         r#"set +e
 failed=0
 pids=""
+tmp_dir="$(mktemp -d "${TMPDIR:-/tmp}/warp-clone-logs.XXXXXX")"
+cleanup_clone_logs() {
+  rm -rf "$tmp_dir"
+}
+trap cleanup_clone_logs EXIT
 clone_repo() {
   repo_name="$1"
   repo_url="$2"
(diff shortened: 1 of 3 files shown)


app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -263,6 +264,97 @@ fn record_codebase_indexing(
...
+        script.push_str("pids=\"$pids $!\"\n");
+    }
+
+    script.push_str(
+        r#"for pid in $pids; do
+  if ! wait "$pid"; then
+    failed=1
+  fi
+done
+exit "$failed"
```
seemeroland, line 325: If cloning any repo fails, do we get any output? Especially for failed clones I think we need to see the output, but ideally we can show the output of every clone that happened in a way that's not interleaved. I think we might need to write output to a log and print it when they're done?

seemeroland:
Do you have a screenshot of what it looks like on success or when a repo clone fails?

seemeroland (approved):
Approving to unblock assuming it's been tested. Also just making sure this script will work on any common shell and not using any zsh or bash specific syntax?

dmichelin:
> Do you have a screenshot of what it looks like on success or when a repo clone fails?

image
This is what it looks like on success. I'm having a hard time manually testing as warp local doesn't seem to let me create an environment with a repo I can't clone

dmichelin:
> Approving to unblock assuming it's been tested. Also just making sure this script will work on any common shell and not using any zsh or bash specific syntax?

Yep, this is all POSIX

PR #12282, opened 2026-06-05
Cmd-k should cancel active conversation stream

seemeroland:
Still has issues: https://www.loom.com/share/94332418643845988a97b6344a5f521e

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -18635,12 +18635,6 @@ impl TerminalView {
             true
         } else {
             // Otherwise, there are some visible blocks and we need to clear stuff.
-            let active_block_is_long_running = self
-                .model
-                .lock()
-                .block_list()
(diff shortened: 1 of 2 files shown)


app/src/terminal/view.rs
```diff
@@ -18635,29 +18635,15 @@ impl TerminalView {
...
-                .active_block()
-                .is_active_and_long_running();
             let is_agent_monitoring = self
                 .model
                 .lock()
                 .block_list()
                 .active_block()
                 .is_agent_monitoring();
 
-            // If there isn't an active long running block, then "clear buffer" just starts a new convo.
```
seemeroland, line 18651: I think we should cancel the active block and cancel the conversation on cmd-k, same as ctrl-c during conversation with an active requested command (`stop_local_agent_conversation`). Then enter agent view for new conversation

PR #12313, opened 2026-06-07
Resolve Cargo target dir in script/macos/run instead of hardcoding ./target

### LINKED ISSUE #11957: Developing Warp consumes 100GB+ per checkout with no working shared-cache path (`script/macos/run` ignores `CARGO_TARGET_DIR`)
### Problem

A debug build of Warp produces 100GB+ of cargo artifacts per checkout. This compounds quickly with the workflows the repo encourages:

- **Git worktrees** — the natural way to keep several branches checked out at once (one for an in-flight PR, one for `master`, one for an experiment). Each worktree gets its own `./target` by default, so 3-4 active worktrees easily climbs past 400GB on disk.
- **Sibling clones** — same story, different mechanism.

There is no first-class story in the repo for keeping this bounded.

### The standard fix that doesn't work

Cargo's documented way to share build artifacts across sibling checkouts and worktrees is to point `~/.cargo/config.toml`'s `[build] target-dir` at a single shared directory:

```toml
[build]
target-dir = "/Users/me/.cargo-target"
```

This is a stable, documented cargo setting ([reference](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir)).

It doesn't work today because `script/macos/run` hardcodes `target/<profile>/bundle/osx/...` in six places and never reads `CARGO_TARGET_DIR` or the cargo config. `cargo bundle` honors the redirect and writes to `~/.cargo-target/debug/bundle/...`; the rest of the script — `install_name_tool`, `codesign`, the final `open` — keeps looking under `./target/` and aborts:

```
error: /Applications/Xcode.app/.../install_name_tool: can't open file:
target/debug/bundle/osx/WarpOss.app/Contents/MacOS/warp-oss (No such file or directory)
```

So contributors who want to share a cache hit this wall the first time they run the app.

### Steps to reproduce

1. Set a shared target dir in `~/.cargo/config.toml` (per the cargo docs above).
2. Run `./script/run` from a fresh Warp checkout.
3. `cargo bundle` writes the bundle to the shared dir; the next step (`install_name_tool`) fails because the script still expects `./target/`.

### Proposed fix

Resolve the target dir once near the top of `script/macos/run` using cargo's stable JSON API:

```bash
TARGET_DIR="$(cargo metadata --no-deps --format-version=1 | jq -r .target_directory)"
```

…and substitute `${TARGET_DIR}` for the hardcoded `target` in each `WARP_APP_PATH` assignment (lines 25, 28, 63, 65, 74, 76), dropping the `./` prefix on the final `open` / launch lines. Validated locally; happy to PR.

If Linux/Windows equivalents have the same hardcoding, the same fix applies there.

### Environment

- macOS, Apple Silicon
- Warp source `master`

bradleyjames:
## Problem

`script/macos/run` hardcoded `target/...` for the bundled `.app` path in six
places. When a developer points Cargo at a shared target directory — via
`CARGO_TARGET_DIR` or `build.target-dir` in `~/.cargo/config.toml` (the standard
way to avoid 100GB+ of build artifacts per checkout/worktree) — `cargo bundle`
writes to the configured directory while the rest of the script keeps looking
under `./target`. The build then fails at `install_name_tool`, which can't find
the bundled binary. Fixes #11957.

## Fix

Resolve the real target directory once via Cargo's stable metadata API:

```bash
TARGET_DIR="$(cargo metadata --no-deps --format-version 1 | jq -r .target_directory)"
```

`cargo metadata` already encodes the correct precedence (env var → `config.toml`
→ default `<workspace>/target`), so a single call covers every case. If metadata
is somehow unavailable, it falls back to `${REPO_ROOT}/target` so behavior is
never worse than before. The resolved path is absolute, so the now-incorrect
`./` prefixes are dropped from the `open`/exec lines (`.//Users/...` would
otherwise resolve relative to cwd).

`cargo metadata` + `jq` are already used by the `script/wasm/*` scripts, and
`jq` is an established build dependency.

## Testing

Ran `./script/macos/run` (OSS) end-to-end in both configurations and confirmed a
live app launch each time:

- **`target-dir` set in `~/.cargo/config.toml`** → builds, bundles, rpaths,
  codesigns, and launches from the shared dir (`~/.cargo-target/...`). Previously
  failed at `install_name_tool`.
- **Nothing configured** → resolves to the in-repo `<workspace>/target`,
  identical to the previous hardcoded behavior.

In both runs `codesign --verify` passed and the launched binary had the expected
`@executable_path/../Frameworks` rpath.

## Scope

Limited to the dev `script/macos/run` flow that the issue calls out. The release
`script/macos/bundle` and the Rust-side `app_target_dir()` (a compile-time
`<repo>/target` path, tracked separately under CORE-2805) are out of scope here.

### DIFF
--- a/script/macos/run
+++ b/script/macos/run
@@ -21,11 +21,19 @@ cd "${REPO_ROOT}"
 : "${WARP_CHANNEL:?WARP_CHANNEL must be set (invoke via ./script/run)}"
 : "${FEATURES:?FEATURES must be set (invoke via ./script/run)}"
 
+# Resolve Cargo's actual target directory rather than assuming ./target. This
+# honors a shared CARGO_TARGET_DIR / build.target-dir (e.g. set in
+# ~/.cargo/config.toml), which is where `cargo bundle` writes the .app bundle.
+TARGET_DIR="$(cargo metadata --no-deps --format-version 1 | jq -r .target_directory)"
+if [ -z "$TARGET_DIR" ] || [ "$TARGET_DIR" = "null" ]; then
+    TARGET_DIR="${REPO_ROOT}/target"
+fi
+
 if [ "$WARP_CHANNEL" = "local" ]; then
-    WARP_APP_PATH="target/debug/bundle/osx/WarpLocal.app"
+    WARP_APP_PATH="${TARGET_DIR}/debug/bundle/osx/WarpLocal.app"
     WARP_SCHEME_NAME="warplocal"
 else
-    WARP_APP_PATH="target/debug/bundle/osx/WarpOss.app"
+    WARP_APP_PATH="${TARGET_DIR}/debug/bundle/osx/WarpOss.app"
     WARP_SCHEME_NAME="warposs"
 fi
 DONT_OPEN=false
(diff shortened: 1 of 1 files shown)


alokedesai (approved):
Thanks!

PR #12327, opened 2026-06-07
Implement warpctrl with set of non-authenticated, safe commands

zachlloyd:
## Description
Implements Warp Control (`warpctrl`), a local CLI for operating a running Warp app through a typed, allowlisted command surface. Everything is gated behind `FeatureFlag::WarpControlCli` (cargo feature `warp_control_cli`).

**Architecture (external-only, same-user):**
- Filesystem discovery records (owner-only) advertise each running instance's loopback endpoint and credential broker.
- A per-process Unix-socket credential broker authenticates the calling OS user via kernel peer credentials and issues short-lived, instance-bound, exact-action credentials.
- A loopback-only HTTP listener (`/v1/control`) revalidates credentials and dispatches typed requests to the app over a main-thread bridge.
- No bearer tokens in discovery records, no CORS, Windows fails closed until broker transport exists there.

**Command surface:**
- Exactly **84 allowlisted actions** across instance, app, capability, window, tab, pane, session, input, theme, appearance, setting, keybinding, action, surface, and file families — uniformly default-authorized. (An earlier one-shot close-confirmation layer was removed; close actions flow through Warp's normal close behavior, so existing app warnings stay authoritative.)
- `input.insert` / `input.replace` only stage text — there is intentionally no way to submit or execute input. Block, Auth, Drive, and History families are excluded entirely.
- CLI entry is a hidden `--warpctrl` control mode on the channel binary (packaged as a wrapper script), with clap subcommands, instance/window/tab/pane/session selectors, pretty/JSON output, and shell completions.

**App integration:**
- New **Settings > Scripting** page with a secure-storage-backed enable toggle (default enabled); disabling immediately stops credential issuance and invalidates outstanding credentials.
- Bundled, feature-gated `warpctrl` agent skill teaching discovery-first usage of the CLI.
- A `warp-tour` skill was added mid-branch and then removed: it depends on `warpctrl tour` commands that live on the follow-up branch (`zach/warp-tour-runner`), which reintroduces the skill together with its CLI.

**Specs:** `specs/warp-control-cli/` (PRODUCT, TECH, SECURITY, README) rewritten to match the external-only design and the 84-action catalog.

## Testing
- Unit tests across the stack: catalog/CLI-parser invariants (every implemented action has a CLI route and handler), credential broker and auth tests, app-side handler and selector-resolution tests, settings-gate tests, and feature-gated bundled-skill tests.
- Manually tested locally with `./script/run --features warp/warp_control_cli` (discovery, credential flow, tab/pane/surface/theme/input commands, Scripting toggle).
- UI flows additionally validated end-to-end by a cloud agent driving the built app with computer use.

## Agent Mode

CHANGELOG-NONE


app/src/local_control/handlers/app_state.rs
```diff
@@ -0,0 +1,853 @@
...
+            ctx,
+        ),
+        ActionKind::SurfaceVerticalTabsToggle => workspace_action(
+            instance_id,
+            action,
+            WorkspaceAction::ToggleVerticalTabsPanel,
+            target,
+            ctx,
+        ),
+        ActionKind::FileOpen => file_open(instance_id, params, target, ctx),
```
Legoben, line 201: nervous about this one - need to make sure the file can only open in Warp and not otherwise.

app/src/local_control/handlers/metadata_config.rs
```diff
@@ -0,0 +1,794 @@
...
+    set_pane_name(&entry, None, ctx)?;
+    Ok(pane_mutation_result(
+        instance_id,
+        ActionKind::PaneResetName,
+        entry.pane_id,
+        entry.tab_id,
+    ))
+}
+
+pub(crate) fn theme_set(
```
Legoben, line 133: This one also makes me nervous - for sure something that gets persisted to server and across Warp usages / devices.

app/src/settings/local_control.rs
```diff
@@ -177,7 +167,7 @@ impl Setting for LocalControlModeSetting {
     }
 
     fn default_value() -> Self::Value {
-        LocalControlMode::Disabled
+        LocalControlMode::Enabled
```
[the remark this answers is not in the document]
Legoben: +1

app/src/local_control/handlers/metadata_config.rs
```diff
@@ -0,0 +1,794 @@
...
+                    ErrorCode::InvalidParams,
+                    "appearance.text.font_name cannot be empty",
+                ));
+            }
+            FontSettings::handle(ctx)
+                .update(ctx, |font_settings, ctx| {
+                    font_settings.monospace_font_name.set_value(font_name, ctx)
+                })
+                .map_err(|err| settings_write_error(ActionKind::SettingSet, err))
+        }
```
Legoben, line 616: more risky b/c this is something that we actually load from their computer

Legoben (commented):
mostly just skimmed

app/src/settings/local_control.rs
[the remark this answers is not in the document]
zachlloyd: [Warp Agent] Changed the default to be channel-based in 62a74f9d2: Scripting now defaults to enabled only on internal dogfood channels (Dev, Local); Stable, Preview, OSS, and Integration builds default to disabled, so public users must explicitly opt in via Settings > Scripting before the discovery record, broker, and endpoint are published. Specs updated to match.

https://github.com/warpdotdev/warp/commit/62a74f9d2

crates/warp_cli/src/local_control/commands.rs
```diff
@@ -94,35 +137,18 @@ pub(super) fn run_instance_command(
...
-                        println!(
-                            "{}\tpid={}\t{}\t{}",
-                            summary.instance_id, summary.pid, summary.channel, endpoint
-                        );
-                    }
-                    Ok(())
-                }
-            }
-        }
+        InstanceCommand::List => run_action_with_params(
```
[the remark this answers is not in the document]
zachlloyd: [Warp Agent] Fixed in 3959368f: `instance list` now enumerates local discovery records via `discovery::list_instances()` (with PID pruning and health probes) instead of sending an authenticated request to one selected instance, and exits 0 with an empty list when nothing is found.

https://github.com/warpdotdev/warp/commit/3959368f

app/src/local_control/handlers/app_state.rs
zachlloyd: [Warp Agent] `file.open` resolves its target via `resolve_file_target_to_open_in_warp`, which can only return Warp's code editor or markdown viewer — it can never route to an external editor, the system-default app, or the script-execution path used by `file://` URLs. Added a regression test pinning this in 62a74f9d2 (`test_resolve_file_target_to_open_in_warp_never_leaves_warp`).

app/src/local_control/handlers/metadata_config.rs
zachlloyd: [Warp Agent] Theme settings are already programmatically changeable by any same-user process that writes Warp's settings file with an external tool, so `theme.set` doesn't add a new capability — it routes the same change through a gated path (feature flag + Scripting setting + per-action scoped credential), and the change is immediately visible and reversible. With 62a74f9d2, public-channel users must also explicitly opt in to Scripting first.

app/src/local_control/handlers/metadata_config.rs
zachlloyd: [Warp Agent] Same rationale as theme.set: the font name is already settable by editing the settings file directly, so the font-loading path must already tolerate arbitrary names. warpctrl doesn't add a new loading capability, and the write is gated behind Scripting (now opt-in on public channels per 62a74f9d2).

app/src/local_control/handlers/app_state.rs
```diff
@@ -0,0 +1,853 @@
...
+    activate_target, active_target_pane_group, decode_params, focus_explicit_pane_target,
+    input_target_pane_id, reject_target_families, tab_index_from_target, target_pane_group,
+    target_pane_id, target_session_pane_id, target_window_id_for_target, target_workspace,
+};
+use crate::local_control::LocalControlBridge;
+use crate::palette::PaletteMode;
+use crate::pane_group::{ActivationReason, Direction, PaneGroupAction};
+use crate::server::telemetry::PaletteSource;
+use crate::settings_view::SettingsSection;
+#[cfg(feature = "local_fs")]
```
alokedesai, line 37: nit: that you can ignore. All of `local_control` should probably be gated by `local_fs` (it doesn't really make sense on the web)

app/src/local_control/handlers/metadata_config.rs
alokedesai: @Legoben it's no different than creating a script that edits the warp config file on disk, right?

app/src/local_control/handlers/settings_surfaces.rs
```diff
@@ -0,0 +1,317 @@
...
+        key: key.to_owned(),
+        value,
+        value_type: value_type.to_owned(),
+    }
+}
+
+pub(super) fn public_theme_name(theme: &ThemeKind) -> String {
+    match theme {
+        ThemeKind::Custom(custom) | ThemeKind::CustomBase16(custom) => custom.name(),
+        ThemeKind::InMemory(_) => "In-memory theme".to_owned(),
```
alokedesai: just use the actual name of the `InMemory` themeKind. "in-memory theme" is going to mean nothing to users

alokedesai (approved):

app/src/local_control/handlers/metadata_config.rs
zachlloyd: [Warp Agent] Leaving this path unchanged. This matches what a same-user process can already do by editing Warp's settings file directly, and the local-control action remains gated behind Scripting plus the scoped credential flow.

app/src/local_control/handlers/app_state.rs
zachlloyd: [Warp Agent] Fixed in https://github.com/warpdotdev/warp/commit/0718c5ba5b962ad788481ae83f50022692812064 by gating the app-side local_control module and singleton registration on the local_fs feature instead of only excluding wasm.

app/src/local_control/handlers/settings_surfaces.rs
zachlloyd: [Warp Agent] Fixed in https://github.com/warpdotdev/warp/commit/0718c5ba5b962ad788481ae83f50022692812064 by making local-control public theme names use ThemeKind's Display implementation, so InMemory themes return their actual configured name.

PR #12328, opened 2026-06-07
Cancel repo metadata tree walks on teardown

### LINKED ISSUE #11350: Entry::build_tree walks are not cancellable — closing the file tree view leaves CPU spike running until in-flight walks finish

chinmaydwivedi:
## Description

Fixes repo metadata file-tree work continuing after the owning repository/file-tree state is removed or replaced.

This PR moves repo metadata tree construction and lazy directory expansion onto async filesystem traversal, tracks spawned build tasks, and aborts tracked work when repository state is removed or superseded. It also guards task completions with a per-build generation so an old aborted build cannot apply stale tree state over a newer build for the same repo or directory.

For lazy directory expansion, duplicate requests for the same directory are coalesced: the first request performs the load, and later requests wait for that in-flight completion instead of starting duplicate work or returning a pending-index error. This keeps remote file-tree responses from failing when two clients or repeated expansions ask for the same directory concurrently.

This is intentionally scoped to repo metadata lifecycle work for #11350: no unrelated UI changes, broad scheduler rewrite, or unrelated refactors.

CHANGELOG-BUG-FIX: Cancel in-flight repo metadata tree walks when repository file-tree state is removed.

## Linked Issue

Fixes #11350

- [x] Manual local and SSH coverage is documented below; visual evidence is attached in the PR discussion.

## Testing

Automated validation at head `02b8c440948cea3e26c6dc4a423d3a07f9e399f9`:

- `./script/format --check`
- `TOOLCHAINS=Metal cargo check -p repo_metadata --features local_fs,test-util -j 4`
- `TOOLCHAINS=Metal cargo check -p repo_metadata --no-default-features -j 4`
- `TOOLCHAINS=Metal cargo clippy -p repo_metadata --features local_fs,test-util --all-targets -j 4 -- -D warnings`
- Ten focused lifecycle, cancellation, stale-completion, nested-repository, and duplicate-load coalescing tests passed.
- `TOOLCHAINS=Metal cargo nextest run -p repo_metadata --features local_fs,test-util -j 4`: 127 passed, 2 skipped.
- `TOOLCHAINS=Metal cargo check -p warp --bin warp-oss --features gui -j 4`
- Built and bundled `WarpOss.app`; `codesign --verify --deep --strict` passed.
- `git diff --check`

Manual local Project Explorer validation at head `02b8c440`:

- **Non-lazy git repository:** indexed the local git fixture and expanded `src/nested/main.rs`.
- **Standalone lazy-loaded directory:** expanded `subdir/deep` and opened `inside.txt` from `/tmp/warp-pr-02b8c440-lazy-root`.
- **Git-ignored lazy expansion:** expanded `node_modules/pkg` and opened `package.json`.

Visual evidence:

- Local non-lazy git repository: https://github.com/user-attachments/assets/e597aaf3-e24d-4915-983e-cfff37e2531a
- Standalone lazy-loaded directory: https://github.com/user-attachments/assets/a0eb6be0-a4f5-4ea4-abdc-a462cf74f575
- Git-ignored lazy expansion: https://github.com/user-attachments/assets/1f4dae32-a0b4-476c-8832-dcbed90719fe

Manual SSH validation at head `02b8c440`:

- Cross-built the public `warp-oss` worker for `x86_64-unknown-linux-musl` with `release_bundle,crash_reporting,standalone,agent_mode_debug,remote_codebase_indexing`.
- Installed the exact-head worker on an Ubuntu 24.04 x86_64 container reachable through the `warp-pr-local` SSH host.
- Verified the deployed worker SHA-256 (`b2b5e75a73822262b085dd2fe670d1872a14135a5e4e374314321b63fee02fb3`) matches the locally built artifact.
- Connected through Warp's SSH extension, loaded `/home/warp/remote-repo` in Project Explorer, expanded `src/nested`, and opened `remote-proof.txt`.

Visual evidence:

- Remote SSH Project Explorer at latest head: https://github.com/user-attachments/assets/ad31a815-d26b-43b9-a676-93fb69c94b19

A full `./script/presubmit` had previously passed format, inline-test-module checks, workspace clippy, clang-format, and wgslfmt before being stopped during full-workspace nextest compilation to avoid monopolizing the local machine. The exact-head package tests, clippy checks, app check, bundle verification, and manual local/SSH validation are listed above.


chinmaydwivedi:
@cla-bot check

crates/repo_metadata/src/entry.rs
kevinyang372: I think the right approach is to convert this into an async function -- this will allow the parent to abort the work with the abort handle we expose in ctx.spawn

crates/repo_metadata/src/entry.rs
```diff
@@ -336,6 +399,7 @@ impl Entry {
 
                     let child_depth = job.depth + 1;
                     for entry in entries {
+                        check_cancelled(cancellation)?;
```
kevinyang372: By converting this method into async, instead of calling check_cancelled everywhere, we could call async_fs::read_dir. As part of the await, the scheduler will check if the thread is aborted and avoid doing unnecessary extra work

kevinyang372 (changes requested):
Thanks for the contribution! I would recommend converting the underlying method to async, which will give us the right abort semantics for free

chinmaydwivedi:
Thanks for the review. I updated the PR to follow this direction:

- Converted the underlying tree walk to async and switched directory expansion from `std::fs::read_dir` to `async_fs::read_dir(...).await`.
- Removed the custom cancellation token/checks.
- Tracked the `ctx.spawn` handles per repo and abort them when repo/file-tree state is removed or a fresh full index starts.
- Kept completion callbacks guarded by the tracked task map so stale completions are ignored if teardown already removed the task.
- Updated async callers in AI indexing/outline code to use `Entry::build_tree_async(...).await`.

Focused validation run locally:

- `TOOLCHAINS=Metal PATH="$HOME/.cargo/bin:$PATH" cargo check -p repo_metadata --features local_fs,test-util`
- `TOOLCHAINS=Metal PATH="$HOME/.cargo/bin:$PATH" cargo nextest run -p repo_metadata -E 'test(remove_repository_aborts_and_drops_build_tasks) | test(build_tree_fail_fast_errors_when_budget_exceeded)' --features local_fs,test-util`
- `TOOLCHAINS=Metal PATH="$HOME/.cargo/bin:$PATH" cargo nextest run -p repo_metadata --features local_fs,test-util`
- `TOOLCHAINS=Metal PATH="$HOME/.cargo/bin:$PATH" cargo clippy -p repo_metadata --features local_fs,test-util --all-targets -- -D warnings`
- `TOOLCHAINS=Metal PATH="$HOME/.cargo/bin:$PATH" cargo check -p repo_metadata --no-default-features`
- `TOOLCHAINS=Metal PATH="$HOME/.cargo/bin:$PATH" cargo check -p ai --features local_fs`
- `git diff --check`

crates/repo_metadata/src/entry.rs
```diff
@@ -160,6 +194,27 @@ impl Entry {
         options: BuildTreeOptions<'_>,
         standing_results: &mut StandingQueryResults,
         definitions: &StandingQueryDefinitions,
+    ) -> Result<Self, BuildTreeError> {
+        futures::executor::block_on(Self::build_tree_with_standing_queries_async(
```
kevinyang372: We shouldn't use block_on to bridge between sync and async functions here. I think the much cleaner semantics is to enforce async on all build_tree methods and migrate any synchronous callers to build tree async instead

kevinyang372 (changes requested):

chinmaydwivedi:
Addressed Kevin's async API feedback in 1ff33ffa.

What changed:
- Removed production `futures::executor::block_on` bridges from `Entry` tree-building APIs.
- Made `Entry::build_tree`, `build_tree_with_standing_queries`, force-included/ignored-ancestor variants, and `Entry::load` async.
- Migrated production callers to await the async tree builders directly or schedule them via `ctx.spawn` so spawned work can be aborted through WarpUI's existing abort handle semantics.
- Updated lazy-load tests to explicitly await tracked build futures instead of depending on synchronous completion.

Validation run locally with capped cargo jobs:
- `./script/format`
- `cargo check -p repo_metadata --features local_fs,test-util`
- `cargo nextest run -p repo_metadata --features local_fs,test-util -j 4`
- `cargo check -p repo_metadata --no-default-features`
- `cargo check -p ai --features local_fs`
- `cargo check -p ai --features local_fs --tests`
- `cargo clippy -p repo_metadata --features local_fs,test-util --all-targets -- -D warnings`
- `git diff --check`

/oz-review

chinmaydwivedi:
Addressed Oz's pending-lazy upgrade race in 43853f5f.

Fix:
- `index_directory` now treats `Pending` + `lazy_loaded_paths.contains(path)` as an upgrade candidate instead of returning early.
- The full index path removes the lazy marker, aborts the in-flight lazy build, and schedules the full build.
- When upgrading from lazy-pending to full-pending, it preserves the existing pending condition so callers waiting on `repository_indexed` are not woken by a Pending -> Pending state replacement.
- Added `test_index_directory_path_upgrades_pending_lazy_loaded_non_git_path` to prove a pending lazy build can be superseded and the deep file is materialized by the full index.

Validation run locally with capped jobs:
- `./script/format`
- `cargo nextest run -p repo_metadata -E 'test(index_directory_path_upgrades_pending_lazy_loaded_non_git_path) | test(index_directory_path_upgrades_lazy_loaded_non_git_path) | test(remove_repository_aborts_and_drops_build_tasks)' --features local_fs,test-util -j 4`\n- `cargo nextest run -p repo_metadata --features local_fs,test-util -j 4`\n- `cargo check -p repo_metadata --features local_fs,test-util`\n- `cargo check -p repo_metadata --no-default-features`\n- `cargo check -p ai --features local_fs`\n- `cargo check -p ai --features local_fs --tests`\n- `cargo clippy -p repo_metadata --features local_fs,test-util --all-targets -- -D warnings`\n- `git diff --check`\n\n/oz-review

chinmaydwivedi:
Addressed Oz's two latest findings in aa5e82c4.

Fixes:
- `entry.rs` now routes directory reads through target-gated helpers: native uses `async_fs::read_dir`, while wasm uses `std::fs::read_dir`, so `repo_metadata` no longer references `async_fs` on wasm targets where it is not a direct dependency.
- Async `load_directory` completions now remember whether the target directory existed when the load was scheduled. If it did, the completion re-checks the current tree and skips applying the subtree if the directory was removed/renamed meanwhile.
- Added `load_directory_completion_skips_removed_tree_entry` to prove an in-flight load cannot resurrect a removed lazy child. Existing gitignored on-demand load coverage still passes.

Validation run locally with capped jobs:
- `./script/format`
- `cargo nextest run -p repo_metadata -E 'test(load_directory_completion_skips_removed_tree_entry) | test(load_directory_tracks_expanded_subdir_for_lazy_root) | test(load_directory_watches_expanded_gitignored_dir_for_git_repo)' --features local_fs,test-util -j 4`\n- `cargo nextest run -p repo_metadata --features local_fs,test-util -j 4`\n- `cargo check -p repo_metadata --features local_fs,test-util`\n- `cargo check -p repo_metadata --no-default-features`\n- `cargo check -p ai --features local_fs --tests`\n- `cargo clippy -p repo_metadata --features local_fs,test-util --all-targets -- -D warnings`\n- `git diff --check`\n\nWasm verification notes:\n- Installed `wasm32-unknown-unknown` and attempted `cargo check -p repo_metadata --target wasm32-unknown-unknown --no-default-features`; this is currently blocked before `repo_metadata` compiles by an existing transitive `getrandom` wasm-js feature error.\n- Also attempted `wasm32-wasip1`; this is blocked before `repo_metadata` compiles by an existing transitive Tokio wasm feature error.\n- `cargo tree -p repo_metadata --target wasm32-unknown-unknown --no-default-features -i async-fs` shows `async-fs` only comes transitively through `warpui_core`; the direct source reference in `entry.rs` is now explicitly `#[cfg(not(target_family = "wasm"))]`.\n\n

 /oz-review

crates/repo_metadata/src/entry.rs
```diff
@@ -521,6 +523,25 @@ impl Entry {
     }
 }
 
+#[cfg(target_family = "wasm")]
+async fn read_dir_entry_paths(path: &Path) -> io::Result<Vec<PathBuf>> {
+    Ok(std::fs::read_dir(path)?
```
kevinyang372: This can just use async_fs

crates/repo_metadata/src/local_model.rs
```diff
@@ -222,6 +225,9 @@ pub struct LocalRepoMetadataModel {
     standing_results: HashMap<StandardizedPath, StandingQueryResults>,
     /// Refcounts for lazily-loaded standalone paths tracked in the model.
     lazy_loaded_paths: HashMap<StandardizedPath, usize>,
+    /// Spawned filesystem tree build tasks owned by each tracked path.
+    build_tasks: HashMap<StandardizedPath, HashMap<u64, SpawnedFutureHandle>>,
```
kevinyang372: This doesn't really make sense to me -- why would you have multiple build tasks to the same repo path? Shouldn't the right behavior to be if there is already an in-progress build task, we avoid starting a new one?

kevinyang372 (commented):
Nice! Thanks for iterating on this!

I have a few comments on this. Also would love to see evidence of manual testing, specifically on the areas of the codebase you touched:
1. Local file tree (lazy loaded and non-lazy loaded git repos)
2. File tree over SSH (you could deploy to a remote machine with ./script/deploy_remote_server)

crates/repo_metadata/src/local_model.rs
chinmaydwivedi: Thanks, both points make sense.

On `read_dir_entry_paths`: agreed. I added the wasm fallback only because Oz flagged that `async-fs` was target-scoped in `repo_metadata`’s Cargo.toml,  but if the preferred direction is to use `async_fs` everywhere, I’ll make `async-fs` an unconditional dependency for this crate and remove the cfg-split helper. That keeps the traversal simpler and preserves the async cancellation/yield behavior uniformly.

On `build_tasks`: the reason I modeled it as `repo_path -> task_id -> handle` was to abort all outstanding work owned by a repo on teardown, including root builds and lazy subdirectory loads under the same repo. But I agree the current shape is too broad and makes it look like duplicate repo builds are expected. I’ll tighten this up so we don’t start duplicate in-progress work for the same build target. I’m thinking either:
 track a single root build per repo and separate keyed directory-load tasks by target dir, or key the task map by the actual build target path instead of an opaque id.

That should make the lifecycle clearer: duplicate work is deduped, teardown/reindex can still abort stale work, and separate directory loads are only concurrent when they are genuinely different targets.

I’ll also add manual testing evidence for:
local file tree behavior for lazy-loaded paths and normal git repos
remote file tree over SSH using `./script/deploy_remote_server`

chinmaydwivedi:
Screenshot 2026-06-12 at 1 25 19 AM
Screenshot 2026-06-12 at 1 23 42 AM
Screenshot 2026-06-12 at 1 22 24 AM

here are the screenshots of changes

kevinyang372 (changes requested):
@chinmaydwivedi As flagged by the oz review bot, these are not evidences testing the remote repo metadata case (as I mentioned before, this is a core area we should make sure we don't regress)

You should be able to setup SSH so that you can tunnel into your own machine. This will make it possible to test even if you don't have a remote setup

chinmaydwivedi:
https://github.com/user-attachments/assets/f49416e1-e639-433c-b92c-92b2f65a953e

02-remote-main-open
01-remote-repo-root

chinmaydwivedi:
Remote SSH manual testing evidence:
-Connected to an Ubuntu ARM64 host over SSH.
 Navigated to /home/warp/remote-repo.
- Expanded the remote repository and opened src/nested/main.rs.
- Expanded the large ignored cancel-tree directory, closed Project Explorer while its lazy load was in flight, reopened it, and confirmed the session remained responsive.
- The running remote daemon was built from this PR commit and its executable hash matched the local branch artifact.

/oz-review

crates/repo_metadata/src/local_model.rs
```diff
@@ -208,6 +211,32 @@ struct RepoWatch {
...
+impl BuildTaskKey {
+    fn repo_root(&self) -> &StandardizedPath {
+        match self {
+            Self::Repository(repo_root) | Self::Directory { repo_root, .. } => repo_root,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+struct BuildTaskGeneration(u64);
```
kevinyang372: I don't quite understand why we need to have versioned build task here -- to me there should always only be one build task per repo / path, if a new build task comes in when there is already an existing one -- we should add it to that task's completion listener

crates/repo_metadata/src/local_model.rs
```diff
@@ -208,6 +211,32 @@ struct RepoWatch {
     extra_dirs: HashSet<StandardizedPath>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+enum BuildTaskKey {
```
kevinyang372: Why there are repository vs directory build task keys? I thought repo metadata is just the file tree state for a given directory path

kevinyang372 (changes requested):
This is very close! Have a few more questions on how we are managing async tasks

crates/repo_metadata/src/local_model.rs
chinmaydwivedi: That makes sense. The generation was added to guard against the stale-completion race Oz flagged: if an old spawned task resolves around the same time it is aborted/replaced, its completion callback could otherwise remove/apply state for the newer task.
But I agree the current model makes the lifecycle harder to reason about. I’ll simplify this around one tracked build task per target path. Equivalent requests for the same path will subscribe to the existing task’s completion instead of starting/replacing another task.
The `Repository` vs `Directory` key split was meant to distinguish full repo root builds from lazy subdirectory expansions while still letting repo teardown abort all owned work. But both are ultimately file-tree builds for a directory path, so I’ll flatten that into a path-keyed task map and abort all tracked paths under the repo root on teardown/removal.
For the stale-completion case, I’ll avoid a separate semantic generation counter and instead make completion only apply if the tracked task for that path is still the same spawned task. If a stronger build needs to supersede a weaker one, like pending lazy root -> full repo index, that path will explicitly cancel/supersede the old task and notify/wait appropriately.

chinmaydwivedi:
I simplified the task model around that invariant.
Build tasks are now keyed directly by the directory path being built, so there is no separate Repository vs Directory key type. Duplicate requests for the same path subscribe to the existing task’s completion waiters instead of starting another build.
I also removed the separate BuildTaskGeneration counter. To keep the stale-completion guard from the earlier Oz finding, completion now only applies if the task currently tracked for that path has the same spawned FutureId. So, aborted work cannot be removed or applied over a newer task, but the model no longer has a separate semantic generation system.
Repo teardown still aborts all in-flight work owned by the repo by removing tracked build paths under the repo root.
/oz-review

crates/repo_metadata/src/local_model.rs
chinmaydwivedi: I wrote the combined reason in the next convo, thanks.

chinmaydwivedi:
Hi @kevinyang372 — following up on this PR. I addressed all async task-management review comments, and all review threads are now resolved.

The current head, 02b8c440, includes:
- repository-owner + target-path task identity
- directory-load-only request coalescing
- generation and entry-identity guards against stale completions
- tracked cancellation for index, lazy-load, and watcher-update tasks
- regression coverage for nested repositories and concurrent directory loads

The latest Oz review found 0 critical, 0 important, and 0 suggestions. Local and SSH Project Explorer testing evidence is also attached.

Could you please take another look when available? @oss-maintainers, please apply no-autoclose if needed while this awaits human re-review.

kevinyang372 (approved):
The code change looks good! The verification was from the previous commit. Do you mind updating the validation screenshots for the latest commit and I could approve the CIs?

chinmaydwivedi:
Screenshot 2026-07-16 at 2 22 06 PM



Remote SSH validation on latest commit 02b8c440: connected to an Ubuntu x86_64 host, loaded /home/warp/remote-repo in Project Explorer, expanded src/nested, and opened remote-proof.txt through the SSH extension.

chinmaydwivedi:
01-local-git-02b8c440
02-local-lazy-02b8c440
03-local-gitignored-02b8c440

1. **Local non-lazy git repository:** Project Explorer fully indexed the local git repository and expanded `src/nested` on commit `02b8c440`.

2. **Standalone lazy-loaded directory:** Project Explorer lazily expanded `subdir/deep` and opened `inside.txt` from a standalone non-git directory.

3. **Git-ignored lazy expansion:** Project Explorer lazily loaded the ignored `node_modules/pkg` subtree and opened `package.json`.

kevinyang372:
@chinmaydwivedi Can you address the CI failures?

chinmaydwivedi:
The latest commit is 119c32cb. The new workflows are currently waiting for maintainer approval. Could you please approve and run the CI workflows? All corresponding checks pass locally.

chinmaydwivedi:
Perfect, just waiting on workflow approval

kevinyang372:
This is merged! Thank you @chinmaydwivedi !

chinmaydwivedi:
Thank you @kevinyang372

Issue #11082, opened 2026-05-15
Sidebar agent session names get stuck on wants to run and manual rename does not replace the main title


jeff-schumann:
## Summary
Regression/UX issue: sidebar names for agent sessions are low-signal and can get stuck on "wants to run...", and manual rename appears as a tiny secondary label instead of replacing the main title.

## Problem
When multiple agents are running at once, sidebar session names do not provide enough context to distinguish sessions. The auto-updating title can remain stuck on "wants to run..." after the session has moved on, and manual rename does not appear to fully replace the session name.

## Expected behavior
- Manual rename should replace the main visible session title in the sidebar.
- Auto-generated session names should update to reflect current state and not remain stuck on stale text like "wants to run...".
- Session names should expose high-signal context (for example, agent identity and repository) so users can distinguish many active agents quickly.

## Actual behavior
- Auto-generated sidebar names can remain on "wants to run..." even after progress continues.
- Renaming a session results in what appears to be a small secondary label ("tiny rename") rather than a full rename of the main sidebar title.
- With several active agents (reported as about six), entries are hard to differentiate because names appear tied to transient commands rather than agent/repo context.

## Reproduction steps or desired workflow
1. Open Warp and start multiple agent sessions (around six).
2. Let agents run commands so session names auto-update.
3. Observe at least one sidebar entry remains stuck on "wants to run..." after the session moves past that step.
4. Attempt to rename that session from the sidebar.
5. Observe the rename appears as a small/secondary text treatment instead of replacing the main session title.

## Artifacts
None attached

## Warp version
v0.2026.05.06.15.42.stable_05

## Operating system
macOS 15.7.7 (build 24G720)

christophercolumbusdog:
I have also encountered this same issue, which is quite annoying. Any plans to fix it?

christophercolumbusdog:
I opened #12341 for the stuck-title half of this. On the `duplicate` label: the nearest match is #9525, but #9525 was closed by #9671, and that change only cleared the stale permission text on Stop, PermissionReplied, and PromptSubmit. The ToolComplete path was never covered. Answering Claude Code's AskUserQuestion emits only `tool_complete` (no `permission_replied`), so the title stays stuck until the next prompt or a Stop. #12341 applies the same fix from #9671 to that last transition.

[...]

PR #9671, opened 2026-04-30
fix: clear permission-scoped state when leaving the permission flow (#9525)

lonexreb:
## Description

Resolves #9525.

`PermissionRequest` populates `session_context.summary` (plus `tool_name` / `tool_input_preview`) so the tab can show e.g. *\"Wants to run bash: …\"*. Nothing was clearing those fields when the permission flow ended, so after the session reached `Success` the tab kept showing the stale permission text — the bug the reporter hit on Claude Code + [claude-code-warp](https://github.com/warpdotdev/claude-code-warp).

### Root cause

The title path under `AgentTabTextPreference::ConversationTitle` reads `session_context.title_like_text()` directly, which is `summary`. So even after `Stop` repopulates `query` and `response`, the tab keeps showing whatever the last `PermissionRequest` set in `summary`.

```rust
// app/src/workspace/view/vertical_tabs.rs:3081
agent_text.cli_agent_title = session.session_context.title_like_text();
```

### Fix

Introduce `CLIAgentSession::clear_permission_scoped_state()` and call it on the three transitions that close out the permission flow:

| Transition | Why |
|---|---|
| `Stop` | Session completed; permission state is no longer relevant — primary repro path. |
| `PermissionReplied` | User answered; status returns to InProgress. |
| `PromptSubmit` | Defensive: the user starting a new turn invalidates any leftover permission state (mirrors the existing `response = None` reset already on this arm). |

`PermissionRequest` itself still populates these fields, so the Blocked-state UI is unchanged.

## Linked Issue

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. (#9525 is `ready-to-implement`.)

Resolves #9525.

## Testing

Four new unit tests in `app/src/terminal/cli_agent_sessions/mod_tests.rs`:

| Test | What it checks |
|---|---|
| `stop_clears_permission_scoped_state` | Primary GH-9525 regression: `summary`/`tool_name`/`tool_input_preview` are cleared and `query`/`response` are repopulated when `Stop` follows `PermissionRequest`. |
| `permission_replied_clears_permission_scoped_state` | Covers the user-replied path; status returns to InProgress with permission state cleared. |
| `prompt_submit_clears_permission_scoped_state` | Covers the abandoned-permission path; clears `response` (existing behavior) and the permission fields (new behavior). |
| `permission_request_still_populates_summary_and_tool_fields` | Sanity: the population path is unaffected — `PermissionRequest` still sets all three fields and transitions to Blocked. |

[...]

Issue #9110, opened 2026-04-17
Tab names not consistent when running coding agents


bhrigu123:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When using claude code on warp and doing a `/rename`, I see the same name on the warp tab. But that name gets overwritten with other prompts as I work with claude. (Also i think the tab names don't update when renaming a session on codex).

Video: https://x.com/Bhrigu_Sr/status/2044992324145701050

### To reproduce

1. Run claude code
2. Rename claude session with `/rename`. That name gets shown on the tab too
3. Work on something on that claude session. The tab name gets overwritten to some other prompt.

### Expected behavior

I expected it to always show the name of the coding agent session. 

### Screenshots, videos, and logs

https://x.com/Bhrigu_Sr/status/2044992324145701050

### Operating system (OS)

macOS

### Operating system and version

Tahoe 26.4.1

### Shell Version

zsh 5.9 (arm64-apple-darwin25.0)

### Current Warp version

v0.2026.04.08.08.36.stable_05

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

lonexreb:
This appears to be a duplicate of #9102, with the same symptom (the user-set `/rename` value is overwritten by AI auto-naming on the next prompt).

[...]

Issue #9102, opened 2026-04-17
[Regression] /rename no longer persists tab name: AI session naming overrides it after each prompt (vertical tab layout)


batmat:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABUG) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agent-platform/agent/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-community/troubleshooting-and-support/sending-us-feedback#gathering-warp-logs)

### Describe the bug

(_UPDATE: actually, I realize after filing this bug report, that this behavior is specific to the "vertical tab layout" mode.
While this layout is extremely promising ❤️, this makes this option totally unusable for me for now. So I disabled this mode, and reverted to using the "standard/preexisting mode" with tab names shown on top and horizontally.
I'm hopeful this is seen as a bug and not as expected 😅. I love where it's going. Thanks again_)
### Description

Since the Warp update that added AI agent session naming, `/rename` no longer persists the sidebar tab label. It takes effect immediately, but Warp's AI-generated summary overwrites it on the next prompt.

This is a regression. Before that update, `/rename` reliably kept the tab name.

*(The new agent management UI is great btw — hopefully just a small rough edge to iron out.)*

### Actual Behavior

- Sidebar tab: reverts to AI-generated summary after each prompt
- macOS Window menu: still shows the `/rename` value

### Expected Behaviour

See below in the dedicated field

### Relation to [#8981](https://github.com/warpdotdev/Warp/issues/8981)

[#8981](https://github.com/warpdotdev/Warp/issues/8981) was filed when `/rename` had no effect on the tab at all. Here, `/rename` *does* update the tab initially, but AI session naming overrides it on the next prompt and `/rename` stops working entirely from that point. This is a regression, not a missing feature.

### To reproduce

1. Start a Claude Code session (or any AI agent session) in Warp
2. Run `/rename MY SESSION`
3. Observe the sidebar tab updates to "MY SESSION"
4. Send any follow-up prompt
[...]

PR #12339, opened 2026-06-08
Clear permission state when blocking tool completes

christophercolumbusdog:
## Description

When Claude Code asks the user a question via its `AskUserQuestion` tool, the plugin emits a `permission_request` event, which populates `session_context.summary` with text like:

```
Wants to run AskUserQuestion: {"questions":[{"question":"What decision should the benchmark...
```

When the user answers the question, the plugin emits only a `tool_complete` event. It does not emit `permission_replied` for this flow. The `ToolComplete` handler flips the status from `Blocked` back to `InProgress` but leaves `summary` populated, so the tab title (which falls back to `summary` via `title_like_text()` when `query` is unset) stays stuck on the stale "Wants to run AskUserQuestion: ..." text until the next prompt or a `Stop` event. If the `Stop` event is dropped, it stays stuck indefinitely.

#9671 fixed the same class of bug for the `Stop`, `PermissionReplied`, and `PromptSubmit` transitions. This PR applies the same fix to the remaining transition out of the permission flow: `ToolComplete` now also calls `clear_permission_scoped_state()`. The call sits behind the existing `Blocked` guard, so it only runs when the session is actually leaving the permission flow.

## Linked Issue

Addresses the "stuck on wants to run..." half of #11082 (the manual rename problem reported there is separate and tracked in #9102 / #9110).

## Testing

Added `tool_complete_clears_permission_scoped_state` to `mod_tests.rs`, mirroring the three regression tests from #9671. It builds a `Blocked` session with populated permission state, applies a `ToolComplete` event, and asserts the summary, tool name, and input preview are cleared and the status is `InProgress`.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed tab titles getting stuck on "Wants to run AskUserQuestion: ..." after answering a Claude Code question.

### DIFF
--- a/app/src/terminal/cli_agent_sessions/mod.rs
+++ b/app/src/terminal/cli_agent_sessions/mod.rs
@@ -162,10 +162,11 @@ impl CLIAgentSession {
     }
 
     /// Clears state populated by `PermissionRequest`. Called whenever the
-    /// session leaves the permission flow (the user replied, a new prompt
-    /// is submitted, or the session ends successfully) so the permission
-    /// summary doesn't leak into later UI surfaces — most visibly the tab
-    /// title, which can fall back to `summary` when `query` is unset.
+    /// session leaves the permission flow (the user replied, a blocking tool
+    /// completed, a new prompt is submitted, or the session ends successfully)
+    /// so the permission summary doesn't leak into later UI surfaces — most
+    /// visibly the tab title, which can fall back to `summary` when `query`
+    /// is unset.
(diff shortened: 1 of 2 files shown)


christophercolumbusdog:
Superseded by #12341. A branch rename on my fork (to follow the handle-prefix branch convention) deleted this PR's head ref, which closed it and GitHub won't let it reopen. The replacement also fixes the commit author email flagged by the CLA bot.

PR #12340, opened 2026-06-08
Fix ghost group reappearing bug

johnturcoo:
## Description

When the last tab of a group was closed, the group was not being deleted. This would cause it to reappear unexpectedly. 

The fix: when a tab is deleted and part of a group, call `prune_empty_tab_group` which will delete the group if it is empty

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4675/deleted-tab-group-reappearing

## Testing

### Screenshots / Videos

[Demo of bug](https://www.loom.com/share/6b858bd86979481d8cf24405c3837fe6)

[Demo after fix](https://www.loom.com/share/a942681a42774357be01edee2e0616d7)

The tab group no longer appears when dragging a tab from above where it used to be.

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -11368,7 +11368,7 @@ impl Workspace {
         self.tab_mru_order.retain(|id| *id != removed_pane_group_id);
 
         // If the closed tab was a group member, prune the group when it now
-        // has no remaining members. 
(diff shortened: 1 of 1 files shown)


vkodithala (approved):
LG, thanks for the fix!

PR #12348, opened 2026-06-08
Persist tab group on app close

johnturcoo:
## Description
Persists tab groups in the local SQLite session database so they survive an app restart, mirroring the existing behavior for tabs. When you close the app with groups defined — named or unnamed, expanded or collapsed, with any number of member tabs — they come back the same way on next launch, with the same tabs as members and in the same positions.

Also adds a color field to tab groups, replicating tab color styling.

## How it works

A new `tab_groups` table is added alongside the existing tabs table, with one row per group (name, collapsed state, and a reference to the parent window). A nullable `tab_group_id` column is added to tabs so each tab can point at its group, or stay NULL if it's ungrouped. The same three layers used for tabs already (workspace state → snapshot DTO → Diesel model) are extended to carry the new group data through the pipeline.

•  On save: insert the groups first, immediately query back the row ids SQLite assigned (since Diesel can't return them from the insert directly), and build a TabGroupId → row id map. Then when inserting tabs, we look up each tab's in-memory group id in that map and set tab_group_id to the matching row id.
•  On restore: load the groups, mint a fresh `TabGroupId` per row, and build the inverse row id → TabGroupId map. When building each restored tab, we look up its `persisted tab_group_id` row id in that map and populate the tab's in-memory `group_id`.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4637/persist-tab-group-on-app-close

## Testing

### Screenshots / Videos

[Demo video](https://www.loom.com/share/1c36609bafe745c0a6868ecfdaa0de10)

Demo of app close and reopen. Tab groups are persisted.

### DIFF
--- a/app/src/persistence/sqlite.rs
+++ b/app/src/persistence/sqlite.rs
@@ -942,6 +942,10 @@ fn save_app_state(conn: &mut SqliteConnection, app_state: &AppState) -> Result<(
                     .map(|group| NewTabGroup {
                         window_id,
                         name: group.name.clone(),
+                        color: match group.color {
+                            SelectedTabColor::Unset => None,
+                            _ => serde_yaml::to_string(&group.color).ok(),
+                        },
                         collapsed: group.collapsed,
                     })
                     .collect();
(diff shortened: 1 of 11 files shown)


app/src/persistence/sqlite.rs
```diff
@@ -930,6 +932,39 @@ fn save_app_state(conn: &mut SqliteConnection, app_state: &AppState) -> Result<(
...
+            let mut tab_group_row_ids: HashMap<TabGroupId, i32> = HashMap::new();
+            if !window.tab_groups.is_empty() {
+                let new_tab_groups: Vec<NewTabGroup> = window
+                    .tab_groups
+                    .iter()
+                    .map(|group| NewTabGroup {
+                        window_id,
+                        name: group.name.clone(),
+                        color: match group.color {
+                            SelectedTabColor::Unset => None,
```
vkodithala, line 946: Noticed we have a `Cleared` variant on `SelectedTabColor`; should we be mapping this to `None` too?

app/src/workspace/tab_group.rs
```diff
@@ -25,6 +27,7 @@ impl Default for TabGroupId {
 pub struct TabGroup {
     pub id: TabGroupId,
     pub name: Option<String>,
+    pub color: SelectedTabColor,
```
vkodithala, line 30: This is a nit, and you totally shouldn't address in this PR, but it's usually better to include functionally different changes without dependencies as separate/stacked PRs. Fine to do here for the sake of #warp-speed tho!

app/src/workspace/view.rs
```diff
@@ -11016,6 +11049,25 @@ impl Workspace {
...
+                    id: group.id,
+                    name: group.name.clone(),
+                    color: group.color,
+                    collapsed: group.collapsed,
+                })
+                .collect()
+        } else {
+            Vec::new()
+        };
+
```
vkodithala, line 11070: Very nice!

crates/persistence/build.rs
```diff
@@ -5,5 +5,11 @@ fn main() -> Result<(), std::env::VarError> {
         println!("cargo:rustc-cfg=feature=\"local_fs\"");
     }
 
+    // `embed_migrations!` bakes the migration SQL into the binary at compile
+    // time. Tell cargo to rebuild this crate whenever any migration file
+    // changes, so editing SQL during development doesn't leave a stale,
+    // cached binary behind.
+    println!("cargo:rerun-if-changed=migrations");
```
vkodithala: Unfamiliar with this codepath, but is this a throwaway diff? Would love to understand if/why we need this.

vkodithala (changes requested):
Mostly LG, a little concerned about adding to our persistence build file. Unfamiliar with this codepath so an explanation would be nice. Feel free to RR once addressed!

crates/persistence/build.rs
johnturcoo: OOPS was testing something, this is not meant to be in my PR

app/src/workspace/tab_group.rs
johnturcoo: I totally agree. Just wanted to add it here so I don't need to update the save/restore code again (#warp-speed as you said)

app/src/persistence/sqlite.rs
johnturcoo: I mirrored how tabs handle color. Cleared is treated meaningfully different than None (it represents that the user explicitly cleared the color selection, rather than never selecting a color). From my understanding this handles the edge case where there is a default tab color and the user clears it for a given tab, and thus falling through to the default (color is None) would be different behavior than having cleared the color (on restoration)

vkodithala (approved):
🚀 thanks for the fixes!

PR #12356, opened 2026-06-08
Update placement of group when created from tab(s)

johnturcoo:
## Description

Changes the placement behavior for "New group with tab" (single tab) and "Create group from tabs" (multi-tab selection) so that the new group is created in-place rather than jumping to the top of the tab list.

## How it works 

Before: both actions relocated the selected tab(s) to index 0, so the new group always appeared at the very top regardless of where the tabs were.

After: the new group is anchored at the position of the earliest tab involved:
•  Single tab — the tab stays where it is; no reordering occurs.
•  Multi-tab selection — the group block is spliced in at the earliest selected tab's original position, with the remaining tabs closing the gap.

Edge case (for both single and multi tab cases): if the anchor tab was already a member of an existing group, inserting at its raw slot would split that group's contiguous run. In that case the new group is placed just past the last surviving member of the old group instead, keeping it intact.

## Linked Issue

https://linear.app/warpdotdev/project/tab-grouping-8d9778792fa7/issues

## Testing

### Screenshots / Videos

[Demo Video](https://www.loom.com/share/7989048815d44141b406743a1c80d883)

Demo video shows some basic cases: 
- creating group from single and multi tab creates the group at the first selected tab
as well as some weird edge cases: 
- creating a group from single and multi tab where the first selected tab is in a group, lands the new group after the existing group

### DIFF
--- a/app/src/workspace/view/tab_grouping.rs
+++ b/app/src/workspace/view/tab_grouping.rs
@@ -160,8 +160,11 @@ impl Workspace {
     }
 
     /// "Create group from tabs" menu action. Group membership requires
-    /// tabs to be contiguous in the bar, so we relocate the selected tabs
-    /// to the top as a single block before binding them to the new group.
+    /// tabs to be contiguous in the bar, so we gather the selected tabs into
+    /// a single block anchored at the earliest selected tab's position before
+    /// binding them to the new group. When that earliest tab was itself in a
+    /// group, the block is placed just past that group's last remaining
+    /// member so the existing group stays contiguous instead of being split.
     pub(super) fn new_tab_group_from_selected_tabs(&mut self, ctx: &mut ViewContext<Self>) {
         if !FeatureFlag::GroupedTabs.is_enabled() {
             return;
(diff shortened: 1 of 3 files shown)


app/src/workspace/view/tab_grouping.rs
```diff
@@ -203,12 +212,40 @@ impl Workspace {
...
+        // Edge case — anchor tab WAS in an existing group G:
+        //   Other surviving members of G are still in `other_tabs`. Inserting at
+        //   `anchor_index` could land in the middle of G's run and split it.
+        //   Example: tabs = [A(G), B(G), C(G), D] and we select B and D.
+        //   other_tabs = [A(G), C(G), D]. anchor_index = 1, which points at C —
+        //   inserting there would produce [A(G), B(new), D(new), C(G)], breaking
+        //   G's contiguity. Instead we search other_tabs from the right for the
+        //   last surviving G member (C, at index 1) and insert after it (index 2),
+        //   giving [A(G), C(G), B(new), D(new)].
+        let insert_at = anchor_previous_group_id
```
peicodes, line 238: Code that strongly benefit from examples generally also greatly benefit from having tests

If this is already covered by the view tests then feel free to ignore

peicodes (approved):

PR #12357, opened 2026-06-08
Avoid cloning display menu items on every render

lucieleblanc:
## Description
Warp's working directory chip and git branch chip open popup menus that list files or branches. Each time one of these menus renders, the full list of (possibly-filtered) menu items was being copied so it could be handed to the list renderer, which requires owned data. For a working directory with many files, that meant a full allocation and copy of the list on every single frame.

This change wraps the items list in a reference-counted pointer (`Arc`) so the per-render handoff is a cheap O(1) pointer clone instead of copying the whole list. The places that actually change the list (when the search text changes, or when new directory contents arrive) still build a fresh list and wrap it in a new `Arc`. That keeps things correct because a previously rendered frame may still be holding the old list.

## Testing
Existing display menu tests cover filtering and selection behavior. No new tests are needed — this is a structural change to how the list is owned, not a logic change.

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/231a59c0-ef97-4c03-abb8-3a09a505f8b2_
_Run: https://oz.staging.warp.dev/runs/019ea845-f9f5-751b-81e8-1bc3d6a206d7_

### DIFF
--- a/app/src/context_chips/display_menu.rs
+++ b/app/src/context_chips/display_menu.rs
@@ -1,6 +1,7 @@
 use std::cmp;
 use std::collections::HashMap;
 use std::fmt::Debug;
+use std::rc::Rc;
 use std::sync::Arc;
 
 use fuzzy_match::{match_indices_case_insensitive, FuzzyMatchResult};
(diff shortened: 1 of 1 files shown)


app/src/context_chips/display_menu.rs
```diff
@@ -441,22 +446,26 @@ impl DisplayChipMenu {
         ctx.notify();
     }
 
+    // `FilteredMenuItem` holds an `Arc<dyn GenericMenuItem>`, which is
+    // intentionally not `Send`/`Sync`; the menu lives on a single (UI) thread.
+    #[allow(clippy::arc_with_non_send_sync)]
```
acarl005: Shouldn't this bc `Rc` instead of `Arc` then? The lint seems valid.

acarl005 (approved):

PR #12359, opened 2026-06-08
Allow orchestration agents to read plans by ID, even if the plan isn't loaded locally

harryalbert:
## Description

With orchestration, agents can make plans and then reference these plans when talking to other agents. These other agents should be able to read said plans, even if the plan isn't loaded locally.

The way to do this is to make it so that the `AIDocumentModel` can dynamically load plans from the cloud if the agent requests to read a plan that isn't loaded locally. For now I kept this new ability limited to agents involved in orchestration, but if we see this being useful for all agents we could easily extend it to be allowed for all agents.

This is also coupled w/ [a change on the server](https://github.com/warpdotdev/warp-server/pull/11733) to prompt agents to include exact plan IDs when communicating with children/parents so that other agents can read plans correctly.

## Testing

### Screenshots / Videos

https://www.loom.com/share/baf688b4e36247e595a94fd8e4606494

## Agent Mode

### DIFF
--- a/app/src/ai/blocklist/action_model/execute/read_documents.rs
+++ b/app/src/ai/blocklist/action_model/execute/read_documents.rs
@@ -3,10 +3,10 @@ use futures::FutureExt;
 use warpui::{AppContext, Entity, ModelContext, SingletonEntity};
 
 use super::{ActionExecution, AnyActionExecution, ExecuteActionInput, PreprocessActionInput};
-use crate::ai::agent::conversation::AIConversationId;
 use crate::ai::agent::{
     AIAgentAction, AIAgentActionType, DocumentContext, ReadDocumentsRequest, ReadDocumentsResult,
 };
+use crate::ai::blocklist::orchestration_topology::conversation_participates_in_orchestration;
 use crate::ai::blocklist::BlocklistAIHistoryModel;
 use crate::ai::document::ai_document_model::{AIDocumentId, AIDocumentModel};
 
@@ -44,9 +44,13 @@ impl ReadDocumentsExecutor {
         };
 
         let (mut documents, mut missing_documents) = read_documents(document_ids, ctx);
-        if !missing_documents.is_empty()
-            && conversation_participates_in_orchestration(conversation_id, ctx)
-        {
+        // Orchestrated children may reference plans owned by their parent conversation, so their
+        // local document model can legitimately be missing a requested plan.
+        let participates_in_orchestration = conversation_participates_in_orchestration(
+            BlocklistAIHistoryModel::as_ref(ctx),
+            conversation_id,
+        );
+        if !missing_documents.is_empty() && participates_in_orchestration {
             AIDocumentModel::handle(ctx).update(ctx, |model, ctx| {
                 for document_id in &missing_documents {
                     if let Err(error) = model.hydrate_saved_plan_from_warp_drive(
@@ -89,6 +93,7 @@ impl ReadDocumentsExecutor {
 impl Entity for ReadDocumentsExecutor {
     type Event = ();
 }
+
 /// Reads requested documents and returns the IDs that are not loaded locally.
 fn read_documents(
     document_ids: &[AIDocumentId],
@@ -114,21 +119,6 @@ fn read_documents(
         });
     }
     (documents, missing_documents)
-}
-
-/// Returns whether a conversation is a parent or child in orchestration.
-fn conversation_participates_in_orchestration(
-    conversation_id: AIConversationId,
-    ctx: &AppContext,
-) -> bool {
-    let history = BlocklistAIHistoryModel::as_ref(ctx);
-    let has_parent = history
-        .conversation(&conversation_id)
-        .is_some_and(|conversation| conversation.is_child_agent_conversation());
-    has_parent
-        || !history
-            .child_conversation_ids_of(&conversation_id)
-            .is_empty()
 }
 
 #[cfg(test)]
--- a/app/src/ai/blocklist/action_model/execute/read_documents.rs
+++ b/app/src/ai/blocklist/action_model/execute/read_documents.rs
@@ -1,12 +1,14 @@
 use futures::future::BoxFuture;
 use futures::FutureExt;
-use warpui::{Entity, ModelContext, SingletonEntity};
+use warpui::{AppContext, Entity, ModelContext, SingletonEntity};
 
 use super::{ActionExecution, AnyActionExecution, ExecuteActionInput, PreprocessActionInput};
+use crate::ai::agent::conversation::AIConversationId;
 use crate::ai::agent::{
     AIAgentAction, AIAgentActionType, DocumentContext, ReadDocumentsRequest, ReadDocumentsResult,
 };
-use crate::ai::document::ai_document_model::AIDocumentModel;
+use crate::ai::blocklist::BlocklistAIHistoryModel;
+use crate::ai::document::ai_document_model::{AIDocumentId, AIDocumentModel};
 
 pub struct ReadDocumentsExecutor;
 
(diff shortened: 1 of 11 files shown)


harryalbert:
* **#12359** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12359?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12359?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/blocklist/action_model/execute/read_documents.rs
```diff
@@ -38,22 +43,40 @@ impl ReadDocumentsExecutor {
...
-                Some(DocumentContext {
-                    document_id: *id,
-                    content,
-                    line_ranges: vec![],
-                    document_version: version,
-                })
-            })
-            .collect();
+        let (mut documents, mut missing_documents) = read_documents(document_ids, ctx);
+        // Orchestrated children may reference plans owned by their parent conversation, so their
```
cephalonaut: Is there a good reason to restrict this to orchestration? Users can see plan IDs and possibly copy-paste them around.

app/src/ai/blocklist/action_model/execute/read_documents.rs
```diff
@@ -38,22 +43,40 @@ impl ReadDocumentsExecutor {
...
-                let version = model.get_current_document(id)?.version;
-                Some(DocumentContext {
-                    document_id: *id,
-                    content,
-                    line_ranges: vec![],
-                    document_version: version,
-                })
-            })
-            .collect();
+        let (mut documents, mut missing_documents) = read_documents(document_ids, ctx);
```
cephalonaut: Could simplify to a single-pass approach that avoids re-reading already-found documents:

```rust
let participates_in_orchestration = /* compute once before loop */;

for id in document_ids {
    let doc = try_read_document(id, ctx); // returns Option<DocumentContext>, all owned
    // model borrow from as_ref ends here (NLL)

    let doc = if doc.is_none() && participates_in_orchestration {
        AIDocumentModel::handle(ctx).update(ctx, |model, ctx| {
            if let Err(e) = model.hydrate_saved_plan_from_warp_drive(*id, conversation_id, ctx) {
                log::warn!("...: {e}");
            }
        });
        try_read_document(id, ctx) // second read only for this id
    } else {
        doc
    };

    match doc {
        Some(d) => documents.push(d),
        None    => missing_documents.push(*id),
    }
}
```

app/src/ai/blocklist/action_model/execute/run_agents.rs
```diff
@@ -387,6 +478,56 @@ enum ChildSlot {
     Pending(async_channel::Receiver<StartAgentOutcome>),
 }
 
+struct PendingPlanPublication {
+    document_id: AIDocumentId,
+    save_wait: async_channel::Receiver<()>,
+}
+
+/// Publishes parent-owned plans and prepares waits for plans still being created.
+fn prepare_plan_publications(
```
cephalonaut: Would like to see plan publication split out into maybe `ai/document/plan_publication.rs` with a `wait_for_plan_publications` free function that returns a `BoxFuture<>`. That way `run_agents` doesn't need to own this logic and other callers could reuse it. The if/else split also disappears since the future resolves immediately when there's nothing to wait on:

```rust
ctx.spawn(
    async move {
        wait_for_plan_publications(pending).await; // no-op when empty
        request
    },
    move |me, request, ctx| {
        if !me.is_pending(&action_id_for_wait) { return; }
        me.dispatch_children_for_prepared_request(
            action_id_for_wait, request, parent_conversation_id, sender, ctx,
        );
    },
);
```

app/src/ai/blocklist/orchestration_topology.rs
```diff
@@ -233,6 +233,21 @@ pub fn orchestration_aware_conversation_status(
     }
 }
 
+/// Returns whether a conversation is a participant in an orchestration tree,
+/// either as a child agent (has a parent) or as an orchestrator (has children).
+pub(crate) fn conversation_participates_in_orchestration(
```
cephalonaut: Maybe worth breaking out named helpers. Note the asymmetry: `conversation_is_orchestrator` works from the index regardless of whether the conversation is loaded, while `conversation_is_child_agent` requires it to be loaded:

```rust
/// Works from the index; does not require the conversation to be loaded.
pub fn conversation_is_orchestrator(
    history: &BlocklistAIHistoryModel,
    conversation_id: AIConversationId,
) -> bool {
    !history.child_conversation_ids_of(&conversation_id).is_empty()
}

/// Requires the conversation to be loaded; returns false if not yet available.
pub fn conversation_is_child_agent(
    history: &BlocklistAIHistoryModel,
    conversation_id: AIConversationId,
) -> bool {
    history
        .conversation(&conversation_id)
        .is_some_and(|c| c.is_child_agent_conversation())
}

pub(crate) fn conversation_participates_in_orchestration(
    history: &BlocklistAIHistoryModel,
    conversation_id: AIConversationId,
) -> bool {
    conversation_is_child_agent(history, conversation_id)
        || conversation_is_orchestrator(history, conversation_id)
}
```

This also deduplicates the `child_conversation_ids_of(...).is_empty()` check in `orchestration_aware_conversation_status`.

app/src/ai/document/ai_document_model.rs
```diff
@@ -412,6 +559,58 @@ impl AIDocumentModel {
...
+    /// Hydrates a saved plan notebook into the target conversation.
+    pub(in crate::ai) fn hydrate_saved_plan_from_warp_drive(
+        &mut self,
+        ai_document_id: AIDocumentId,
+        conversation_id: AIConversationId,
+        ctx: &mut ModelContext<Self>,
+    ) -> Result<(), String> {
+        let notebook = CloudModel::as_ref(ctx)
+            .get_all_active_notebooks()
+            .find(|notebook| notebook.model().ai_document_id.as_ref() == Some(&ai_document_id))
```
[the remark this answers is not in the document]
cephalonaut: I sort of assume that the real constraint here is whether the user can access the plan?

app/src/ai/blocklist/action_model/execute/read_documents.rs
harryalbert: good call

app/src/ai/blocklist/action_model/execute/read_documents.rs
harryalbert: the only reason I was thinking about is, limiting this to just orchestration prevents wasted work if a regular agent tries to read a plan that doesn't exist. That being said, I don't really think that that should happen often, and agreed that users pasting plan IDs is a valid use case, so I'm happy to remove the gate

app/src/ai/blocklist/action_model/execute/run_agents.rs
harryalbert: great call

app/src/ai/blocklist/orchestration_topology.rs
harryalbert: good call, but this helper is unnecessary anyways now that we've removed the orchestration gate

app/src/ai/document/ai_document_model.rs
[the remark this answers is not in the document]
harryalbert: yeah user access to the plan is the enforced boundary (the `CloudModel` only contains notebooks the user can access), so I'm not worried about this

app/src/ai/document/ai_document_model.rs
```diff
@@ -286,6 +293,176 @@ impl AIDocumentModel {
...
+            .filter_map(|(document_id, document)| {
+                (document.conversation_id == conversation_id).then_some(*document_id)
+            })
+            .collect::<Vec<_>>();
+        let mut awaiting_server_backing = Vec::new();
+
+        for document_id in document_ids {
+            match self.get_document_save_status(&document_id) {
+                AIDocumentSaveStatus::Saved => {
+                    self.maybe_update_cloud_notebook_data(&document_id, ctx);
```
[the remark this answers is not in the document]
harryalbert: This is semi-reasonable, but I'm not really concerned about this race. We're doing this work _before_ spawning the cloud agent, so document sync would have to take longer than it takes to:
* spin up the tool call to start the agent
* initialize the agent in the cloud
* have the cloud agent start running and decide to read the planning doc

It just seems very unlikely and not worth a lot of extra complexity and state bookkeeping imo

cephalonaut (approved):
Nice work!

PR #12363, opened 2026-06-08
Add basic tracing support for cloud agents.

vorporeal:
## Description

Adds opt-in OpenTelemetry tracing for cloud-agent runs so we can inspect end-to-end execution when debugging latency and failures.

When `WARP_CLOUD_AGENT_OTLP_ENDPOINT` is configured, the native app installs an OTLP/HTTP trace exporter; otherwise tracing remains a no-op. Cloud-agent spans are explicitly marked, and filtering at the exporter boundary prevents unrelated application spans from being sent. This also adds spans and events across the cloud-agent lifecycle, server APIs, and driver setup, and propagates the current span through asynchronous tasks and streams.

The tracing provider is retained through the standard application shutdown path, with a best-effort active-span registry that ends still-reachable spans before shutting down the provider so long-lived asynchronous work is less likely to be lost.

## Linked Issue

None.

## Testing

- [x] Manually tested locally with `./script/run -- agent run --prompt "hello world"`.
- [x] Ran `./script/format`.
- [x] Ran `cargo check -p warp --features gui`.
- [x] Ran `cargo doc -p warp --features gui --no-deps`.

No automated tests were added; this is opt-in observability plumbing that was validated through the cloud-agent command path and focused compilation/documentation checks.

> [!NOTE]
> note from dstern: i ended up working with an agent to do an end-to-end test of the fully-authenticated flow with a local cloud agent running via a self-hosted worker host; everything appears to work.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/src/ai/agent_sdk/driver/terminal.rs
+++ b/app/src/ai/agent_sdk/driver/terminal.rs
@@ -8,8 +8,10 @@ use std::task::{Context, Poll};
 use std::time::Duration;
 
 use futures::channel::oneshot;
+use futures::TryFutureExt as _;
 use session_sharing_protocol::common::{Role, SessionId};
 use session_sharing_protocol::sharer::SessionRetentionReason;
+use tracing::Instrument as _;
 use warp_cli::share::{ShareAccessLevel, ShareRequest, ShareSubject};
 use warp_completer::completer::CommandOutput;
 use warp_core::command::ExitCode;
@@ -547,11 +549,17 @@ impl TerminalDriver {
             session_bootstrapped
                 .wait()
                 .with_timeout(TERMINAL_SESSION_BOOTSTRAP_TIMEOUT)
-                .await
-                .map_err(|_| {
+                .map_err(|err| {
                     log::error!("Timed out waiting for session bootstrap");
+                    tracing::error!(error = %err);
+
                     AgentDriverError::BootstrapFailed
                 })
+                .instrument(tracing::info_span!(
+                    "wait_for_session_bootstrapped",
+                    tags.cloud_agent = true
+                ))
+                .await
         }
     }
 
--- a/app/src/ai/agent_sdk/setup_observability.rs
+++ b/app/src/ai/agent_sdk/setup_observability.rs
@@ -2,6 +2,8 @@ use std::future::Future;
 use std::sync::Arc;
 
 use chrono::{DateTime, Utc};
+use futures::FutureExt as _;
+use tracing::Instrument as _;
 use warpui::r#async::executor::Background;
 
 use crate::ai::ambient_agents::AmbientAgentTaskId;
@@ -37,16 +39,26 @@ impl SetupClientEventReporter {
         }
     }
 
-    pub(crate) async fn record_result<T, E>(
+    pub(crate) async fn record_result<T, E: std::error::Error>(
         &self,
         step: SetupStep,
         future: impl Future<Output = Result<T, E>>,
     ) -> Result<T, E> {
+        let (step_name, span) = step.to_event_name_and_span();
+
         let start_timestamp = Utc::now();
-        let result = future.await;
+        let result = future
+            .map(|result| {
+                result.inspect_err(|err| {
+                    tracing::error!(error = %err);
+                })
+            })
+            .instrument(span)
+            .await;
         let finish_timestamp = Utc::now();
+
         self.post_setup_metric_event_best_effort(
-            step,
+            step_name,
             start_timestamp,
             finish_timestamp,
             result.is_err(),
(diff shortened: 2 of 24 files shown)


vorporeal:
* **#12373** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12373?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12370** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12370?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12364** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12363** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12363?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12363?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/agent_sdk/setup_observability.rs
```diff
@@ -181,33 +204,83 @@ pub(crate) enum SetupStep {
...
 
+macro_rules! span_and_name {
+    ($name:literal) => {
+        ($name, tracing::info_span!($name, tags.cloud_agent = true))
+    };
+}
+
 impl SetupStep {
-    fn as_event_name(self) -> &'static str {
+    fn to_event_name_and_span(self) -> (&'static str, tracing::Span) {
```
vorporeal, line 214: `tracing::info_span!()` and friends require that the span name (and the set of fields that will be recorded) are defined statically at compile time, and that the span name is a _literal_ (not just the result of a const expression).

in order to achieve that, and avoid duplicating the name, i replaced the simple enum->name mapping with an enum->(span,name) mapping, using a macro to generate both from the same input literal.

app/src/ai/agent_sdk/driver/terminal.rs
```diff
@@ -547,11 +549,17 @@ impl TerminalDriver {
...
                 .with_timeout(TERMINAL_SESSION_BOOTSTRAP_TIMEOUT)
-                .await
-                .map_err(|_| {
+                .map_err(|err| {
                     log::error!("Timed out waiting for session bootstrap");
+                    tracing::error!(error = %err);
+
                     AgentDriverError::BootstrapFailed
                 })
+                .instrument(tracing::info_span!(
```
seemeroland, line 558: nit: is this redundant because `record_result` already instruments it here? https://github.com/warpdotdev/warp/blob/4a06479bbfc85186d17b7715b8c4226cb3bc82dd/app/src/ai/agent_sdk/driver.rs#L1682

There's one other usage here but this is for local docker sandboxes (not cloud agent)
https://github.com/warpdotdev/warp/blob/4a06479bbfc85186d17b7715b8c4226cb3bc82dd/app/src/terminal/view/docker_sandbox/mod.rs#L231

app/src/ai/agent_sdk/setup_observability.rs
```diff
@@ -130,6 +151,8 @@ impl SetupClientEventReporter {
         event_name: &'static str,
         request: AgentRunClientEventRequest,
     ) {
+        tracing::info!(event_name, tags.cloud_agent = true);
```
seemeroland, line 154: What's this for?

crates/warp_core/src/interval_timer.rs
```diff
@@ -37,8 +37,10 @@ impl IntervalTimer {
     }
 
     pub fn mark_interval_end(&mut self, name: impl Into<String>) {
+        let name = name.into();
+        tracing::info!(name);
```
seemeroland, line 41: Do we want this?

seemeroland (approved):
Very nice, excited to have this!

app/src/ai/agent_sdk/setup_observability.rs
vorporeal: this is an info-level _event_ (as opposed to an info-level _span_) - think of it more like a log line.  it marks a point-in-time event, that we can see as a point on a span.

here, it allows us to track these point-in-time events that we think are worth reporting to the server.

crates/warp_core/src/interval_timer.rs
vorporeal: yeah, i added this to help us see different points during the application initialization phase - this is more useful for latency ("oh whoa it took us a bunch of time between event A and event B, we should look into that") than reliability.

app/src/ai/agent_sdk/driver/terminal.rs
vorporeal: maybe?  but i'd rather have too many spans and remove some later.

PR #12374, opened 2026-06-08
Filter large and/or binary files over the remote server

MaggieShan:
## Description
* Previously we were filtering binary files and diffs that were too large at render time - now that we've added support for remote diff states, this filtering needs to happen before sending any content over the wire 
  * We were seeing this appear as an error where the code review view was stuck loading and/or errored with "message too large"   
* The fix includes: filtering both binary files and large files over the wire + updating `DiffSize::Unrenderable` to include a reason such that we can distinguish between large files and large diff sizes 

## Linked Issue
[APP-4681](https://linear.app/warpdotdev/issue/APP-4681/filter-largebinary-files-before-sending-diff-loadsupdates)

## Testing

Before:
Screenshot 2026-06-08 at 4 33 37 PM

After:
Screenshot 2026-06-08 at 4 33 51 PM
Screenshot 2026-06-08 at 4 09 15 PM

## Agent Mode


app/src/code_review/diff_state/local.rs
```diff
@@ -1573,8 +1573,13 @@ impl LocalDiffStateModel {
             let is_binary = binary_files.contains(&file_path);
             let mut file_diff =
                 Self::get_file_diff(repo_path, &file_path, &status, is_binary, None).await?;
-            let content_at_head =
-                Self::get_file_content_at_head(repo_path, &file_path, &status).await;
+            // Never read or ship base content for binary files: it can't be
+            // inline-rendered and, after lossy UTF-8 decoding, can balloon ~3x.
+            let content_at_head = if is_binary {
```
kevinyang372, line 1578: Just confirming, there is no upstream dependency that relies on binary file content being populated yeah?

app/src/remote_server/diff_state_proto.rs
```diff
@@ -550,6 +560,23 @@ impl From<&DiffState> for proto::DiffState {
 /// Converts a `FileDiff` to proto with an optional `content_at_base`.
 /// Cannot be a `From` impl because of the extra parameter.
 pub fn file_diff_to_proto(f: &FileDiff, content_at_base: Option<&str>) -> proto::FileDiff {
+    // Decide what base content (if any) to send over the wire, adjusting the
+    // rendered size accordingly. This gating is remote-only: it runs when the
+    // daemon serializes a diff for a subscriber, never on the local in-memory
+    // path, so local rendering keeps full content regardless of size.
+    let (size, content_at_base) = if f.is_binary {
```
kevinyang372, line 567: I wonder if this should be a configuration on the diff state model (e.g. drop_diff_with_size_over_limit: Option\<usize\>)

A potential benefit of this is we could pass that into the methods of the diff state and potentially check the diff size before fully reading the content_at_base into memory. The main question is if it's possible to check the diff size before reading. If it's not possible, the upside is much more limited and we could probably skip

kevinyang372 (approved):

app/src/code_review/diff_state/local.rs
MaggieShan: Yup confirmed!

app/src/remote_server/diff_state_proto.rs
MaggieShan: That's a good suggestion - can follow up with those changes!

PR #12377, opened 2026-06-08
Introduce single-host BundledSkill abstraction

moirahuang:
## Description

Introduces `BundledSkill`, a single-host catalog that owns bundled-skill detection, activation-filtered descriptors, path-to-reference resolution, and definition lookup.

`SkillManager` now delegates its existing local bundled-skill behavior to this catalog. This is a no-op refactor that establishes the abstraction used by the stacked remote-host support PR.

## Linked Issue

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Testing

- `./script/format --check`
- `cargo test -p warp --lib ai::skills::skill_manager::tests`
- `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`
- `cargo clippy -p warp_completer --all-targets --tests -- -D warnings`
- [ ] I have manually tested my changes locally with `./script/run`

## Agent Mode

- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-NONE

Co-Authored-By: Oz <oz-agent@warp.dev>

PR #12378, opened 2026-06-08
Install bundled skills globally on remote hosts and push daemon-parsed catalogs

moirahuang:
## Description

Ships bundled skills to remote hosts as a single **global, version-independent resources install**, with the daemon parsing skills against its own filesystem and pushing the catalog to clients.

**Install** (`install_remote_server.sh`, `setup.rs`): the binary install stays exactly as before (flat, version-suffixed path, `--version` check). The installer additionally moves the artifact's `resources/` tree to `{install_dir}/bundled_resources` — a global location deliberately decoupled from the binary version: the last install wins, and slight skew against an older running daemon is accepted (it parsed its skills at startup). A tarball without resources is non-fatal. The binary `find` excludes the `resources/` tree so bundled-skill companion files named `oz-*` can never be mistaken for the executable.

**Daemon** (`server_model.rs`): at startup, parses and handlebars-renders the skills under `bundled_resources` against its own paths (`{{skill_dir}}`, `{{settings_schema_path}}`), then pushes the pre-parsed catalog as a `BundledSkillsSnapshot` notification — broadcast on parse completion and sent to each connection right after its `Initialize`. The handshake is never blocked, and no resources path is advertised in `InitializeResponse` anymore. `RequiresFile` activations are evaluated daemon-side; `RequiresMcp` ships as a wire hint the client evaluates. Parsing is deliberately not feature-flag gated: exposure is controlled on the client, where the connecting user's flag state actually lives.

**Client** (`manager.rs`, `remote.rs`, `skill_manager.rs`): the snapshot flows `ClientEvent` → `RemoteServerManagerEvent::BundledSkillsSnapshot { host_id, skills }`. `SkillManager` re-parses the daemon-rendered content, wraps paths as `RemotePath`s, and stores one catalog per connected host in `BundledSkills` (`HostDisconnected` tears it down; a fresh snapshot after reconnect replaces it wholesale). Skill selection is now host-aware: SSH sessions see the remote daemon's catalog — never the local client's — and local sessions see only the local one. Remote bundled skills resolve by path with activation re-checked at invocation.

Screenshot 2026-06-12 at 2 31 49 PM

## Linked Issue

## Testing

- Install-script integration tests cover the global resources install, last-install-wins replacement, the resources-less tarball path, the `oz-*` decoy exclusion, and removal leaving `bundled_resources` intact.
- Proto round-trip and client push-event tests for `BundledSkillsSnapshot`; daemon broadcast tests (no-op before parse, reaches all connections after).
- Conversion tests for daemon→proto serialization (activation handling) and proto→catalog (host-scoped paths, unknown-integration and invalid-path skipping).
- `SkillManager` tests for host-aware catalog selection (remote cwd → remote catalog only) and path-based resolution of remote bundled skills.
- `cargo nextest run -p remote_server` and targeted `-p warp` suites pass; `./script/format` and Clippy clean.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/src/ai/skills/bundled.rs
+++ b/app/src/ai/skills/bundled.rs
@@ -6,6 +6,7 @@ use futures::TryStreamExt;
 use warp_core::channel::ChannelState;
 use warp_core::ui::icons::Icon;
 use warp_core::{report_error, safe_warn};
+use warp_util::host_id::HostId;
 use warp_util::local_or_remote_path::LocalOrRemotePath;
 use warpui::{AppContext, SingletonEntity};
 
@@ -37,7 +38,54 @@ impl BundledSkillActivation {
     }
 }
 
-/// A bundled skill definition with its activation condition and icon.
+/// Catalogs of bundled skills for the local host and connected remote hosts.
+#[derive(Debug, Default)]
+pub struct BundledSkills {
+    local: BundledSkill,
+    remote_by_host: HashMap<HostId, BundledSkill>,
+}
+
+impl BundledSkills {
+    pub fn set_local(&mut self, bundled_skill: BundledSkill) {
+        self.local = bundled_skill;
+    }
+
+    pub fn local(&self) -> &BundledSkill {
+        &self.local
+    }
+
+    /// Installs the catalog for a connected remote host, replacing any
+    /// previous catalog from an earlier connection.
+    pub fn insert_remote(&mut self, host_id: HostId, bundled_skill: BundledSkill) {
+        self.remote_by_host.insert(host_id, bundled_skill);
+    }
+
+    /// Removes all catalog state for a disconnected remote host.
+    pub fn remove_remote(&mut self, host_id: &HostId) {
+        self.remote_by_host.remove(host_id);
+    }
+
+    /// Returns the catalog for a connected remote host.
+    ///
+    /// Agent Mode catalog selection is wired up in the stacked follow-up; until
+    /// then this read path is exercised by tests only.
+    #[cfg_attr(not(test), allow(dead_code))]
+    pub fn remote(&self, host_id: &HostId) -> Option<&BundledSkill> {
+        self.remote_by_host.get(host_id)
+    }
+
+    #[cfg(test)]
+    pub fn insert_local_for_testing(
+        &mut self,
+        id: impl Into<String>,
+        skill: ParsedSkill,
+        activation: BundledSkillActivation,
+    ) {
+        self.local.insert_for_testing(id, skill, activation);
+    }
+}
+
+/// One bundled skill definition with its activation condition and icon.
 #[derive(Debug, Clone)]
 struct BundledSkillDefinition {
     skill: ParsedSkill,
@@ -54,9 +102,21 @@ pub struct BundledSkill {
 impl BundledSkill {
     /// Detect all skill definitions bundled with Warp for the local host.
     pub async fn detect() -> Self {
+        let Some(resources_dir) = warp_core::paths::bundled_resources_dir() else {
+            return Self::default();
+        };
+        Self::detect_for_resources_dir(LocalOrRemotePath::Local(resources_dir)).await
+    }
+
+    /// Detect all client-shipped skill definitions, rendering their paths for
+    /// the host that owns `resources_dir`.
+    pub async fn detect_for_resources_dir(resources_dir: LocalOrRemotePath) -> Self {
+        let Some(source_resources_dir) = warp_core::paths::bundled_resources_dir() else {
+            return Self::default();
+        };
         let (mut definitions, figma_definitions) = futures::join!(
-            load_bundled_skill_definitions(),
-            load_figma_skill_definitions()
+            load_bundled_skill_definitions(&source_resources_dir, &resources_dir),
+            load_figma_skill_definitions(&source_resources_dir, &resources_dir)
         );
         definitions.extend(figma_definitions);
         Self { definitions }
@@ -115,17 +175,18 @@ impl BundledSkill {
 }
 
 /// Load skill definitions bundled with Warp.
-async fn load_bundled_skill_definitions() -> HashMap<String, BundledSkillDefinition> {
-    let Some(resources_dir) = warp_core::paths::bundled_resources_dir() else {
-        return HashMap::new();
-    };
-    let skills_dir = resources_dir.join("bundled").join("skills");
-    read_bundled_skills(&skills_dir)
+async fn load_bundled_skill_definitions(
+    source_resources_dir: &Path,
+    resources_dir: &LocalOrRemotePath,
+) -> HashMap<String, BundledSkillDefinition> {
+    let skills_dir = source_resources_dir.join("bundled").join("skills");
+    let host_skills_dir = resources_dir.join("bundled").join("skills");
+    read_bundled_skills(&skills_dir, &host_skills_dir, resources_dir)
         .await
         .into_iter()
         .map(|(id, skill)| {
             let icon = icon_for_bundled_skill(&id);
-            let activation = activation_for_bundled_skill(&id, &resources_dir);
+            let activation = activation_for_bundled_skill(&id, source_resources_dir);
             let bundled = BundledSkillDefinition {
                 skill,
                 activation,
@@ -137,15 +198,19 @@ async fn load_bundled_skill_definitions() -> HashMap<String, BundledSkillDefinit
 }
 
 /// Load Figma-specific bundled skills from the `figma/` subdirectory.
-async fn load_figma_skill_definitions() -> HashMap<String, BundledSkillDefinition> {
-    let Some(resources_dir) = warp_core::paths::bundled_resources_dir() else {
-        return HashMap::new();
-    };
-    let figma_skills_dir = resources_dir
+async fn load_figma_skill_definitions(
+    source_resources_dir: &Path,
+    resources_dir: &LocalOrRemotePath,
+) -> HashMap<String, BundledSkillDefinition> {
+    let figma_skills_dir = source_resources_dir
         .join("bundled")
         .join("mcp_skills")
         .join("figma");
-    read_bundled_skills(&figma_skills_dir)
+    let host_figma_skills_dir = resources_dir
+        .join("bundled")
+        .join("mcp_skills")
+        .join("figma");
+    read_bundled_skills(&figma_skills_dir, &host_figma_skills_dir, resources_dir)
         .await
         .into_iter()
         .map(|(id, skill)| {
@@ -160,9 +225,12 @@ async fn load_figma_skill_definitions() -> HashMap<String, BundledSkillDefinitio
 }
 
 /// Read bundled skill definitions from the specified directory.
-pub(crate) async fn read_bundled_skills(skills_dir: &Path) -> HashMap<String, ParsedSkill> {
+pub(crate) async fn read_bundled_skills(
+    skills_dir: &Path,
+    host_skills_dir: &LocalOrRemotePath,
+    resources_dir: &LocalOrRemotePath,
+) -> HashMap<String, ParsedSkill> {
     let mut skills = HashMap::new();
-    let context = build_bundled_skill_context();
 
     let Ok(mut entries) = async_fs::read_dir(skills_dir).await else {
         return skills;
@@ -194,9 +262,12 @@ pub(crate) async fn read_bundled_skills(skills_dir: &Path) -> HashMap<String, Pa
             );
             continue;
         };
+        let host_skill_dir = host_skills_dir.join(skill_id);
+        let context = build_bundled_skill_context(resources_dir, &host_skill_dir);
 
         // Apply variable substitution to the skill content.
         skill.content = handlebars::render_template(&skill.content, &context);
+        skill.path = host_skill_dir.join("SKILL.md");
         skills.insert(skill_id.to_owned(), skill);
     }
 
@@ -212,10 +283,14 @@ pub(crate) async fn read_bundled_skills(skills_dir: &Path) -> HashMap<String, Pa
 /// - `{{warp_cli_binary_name}}` - The CLI binary name (e.g., `warp` or `warp-cli`)
 /// - `{{warp_url_scheme}}` - The URL scheme (e.g., `warp`, `warpdev`, `warppreview`)
 /// - `{{settings_schema_path}}` - Path to the bundled JSON settings schema
+/// - `{{skill_dir}}` - Path to the bundled skill's directory
 /// - `{{settings_file_path}}` - Path to the user's settings TOML file
 /// - `{{keybindings_file_path}}` - Path to the user's keybindings YAML file
-pub(crate) fn build_bundled_skill_context() -> HashMap<String, String> {
-    let mut context: HashMap<String, String> = [
+pub(crate) fn build_bundled_skill_context(
+    resources_dir: &LocalOrRemotePath,
+    skill_dir: &LocalOrRemotePath,
+) -> HashMap<String, String> {
+    [
         (
             "warp_server_url".to_owned(),
             ChannelState::server_root_url().into_owned(),
@@ -236,20 +311,14 @@ pub(crate) fn build_bundled_skill_context() -> HashMap<String, String> {
             "keybindings_file_path".to_owned(),
             keybinding_file_path().display().to_string(),
         ),
+        (
+            "settings_schema_path".to_owned(),
+            resources_dir.join("settings_schema.json").display_path(),
+        ),
+        ("skill_dir".to_owned(), skill_dir.display_path()),
     ]
     .into_iter()
-    .collect();
-
-    if let Some(schema_path) =
-        warp_core::paths::bundled_resources_dir().map(|dir| dir.join("settings_schema.json"))
-    {
-        context.insert(
-            "settings_schema_path".to_owned(),
-            schema_path.display().to_string(),
-        );
-    }
-
-    context
+    .collect()
 }
 
 /// Returns the icon for a bundled skill, given its directory-based ID.
@@ -274,3 +343,7 @@ fn activation_for_bundled_skill(skill_id: &str, resources_dir: &Path) -> Bundled
         _ => BundledSkillActivation::Always,
     }
 }
+
+#[cfg(test)]
+#[path = "bundled_tests.rs"]
+mod tests;
--- a/app/src/ai/skills/bundled.rs
+++ b/app/src/ai/skills/bundled.rs
@@ -66,12 +66,37 @@ impl BundledSkills {
     }
 
     /// Returns the catalog for a connected remote host.
-    ///
-    /// Agent Mode catalog selection is wired up in the stacked follow-up; until
-    /// then this read path is exercised by tests only.
-    #[cfg_attr(not(test), allow(dead_code))]
     pub fn remote(&self, host_id: &HostId) -> Option<&BundledSkill> {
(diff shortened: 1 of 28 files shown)


moirahuang:
* **#12418** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12378** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12378?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12378?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#12377** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12377?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/remote_server/src/install_remote_server.sh
```diff
@@ -1,5 +1,23 @@
...
+#
+# Install layout:
+#
+#   {install_dir}/
+#   ├── bundles/{bundle_version}/      ← the artifact's shipped layout, unchanged:
+#   │   ├── {binary_name}                 the executable...
+#   │   └── resources/                    ...and its sibling resources
+#   └── {binary_name}{version_suffix}  ── symlink ─▶ bundles/{bundle_version}/{binary_name}
+#
+# The bundle preserves the executable-beside-resources pairing exactly as the
```
moirahuang: including an explanation for how we're shipping `resources/`

crates/remote_server/src/install_remote_server.sh
```diff
@@ -79,7 +105,49 @@ fi
...
-bin=$(find "$tmpdir" -type f -name 'oz*' ! -name '*.tar.gz' | head -n1)
+# The executable and its resources are siblings in the artifact. Exclude the
+# resources tree from the search: bundled skills may ship companion files
+# whose names also start with `oz`.
+bin=$(find "$tmpdir" -type f -name 'oz*' ! -name '*.tar.gz' ! -path '*/resources/*' | head -n1)
 if [ -z "$bin" ]; then echo "no binary found in tarball" >&2; exit 1; fi
+resources="$(dirname "$bin")/resources"
+if [ ! -d "$resources" ]; then echo "no resources directory found in tarball" >&2; exit 1; fi
+
+# Assemble the complete bundle in the temp dir first: assembly takes multiple
```
moirahuang: doing this in a temp dir because there's multiple steps so in case the connection drops midway, we don't have a half copy

crates/remote_server/src/install_remote_server.sh
```diff
@@ -79,7 +105,49 @@ fi
...
+# at this version; healthy bundles are skipped by the pre-install binary check.
+install_bundle_tmp="$bundles_dir/.{bundle_version}.install.$$"
+rm -rf "$install_bundle_tmp"
+mv "$staged_bundle" "$install_bundle_tmp"
+rm -rf "$bundle_dir"
+mv "$install_bundle_tmp" "$bundle_dir"
+
+# Point the historical launch path at the bundle. Clients launch the
+# version-suffixed path they have always used, while the shipped binary keeps
+# its unsuffixed name beside its resources — the symlink bridges the two, and
```
moirahuang: explaining the versioned symlink

crates/remote_server/src/setup.rs
```diff
@@ -495,15 +495,36 @@ pub fn remote_server_binary() -> String {
...
+        remote_server_binary(),
+    )
+}
+
+/// Returns the shell command to remove the current remote-server install.
+///
+/// Removes both the compatibility path and its version-scoped bundle. This is
+/// safe for legacy binary-only installs because `rm -f` also removes a regular
+/// file at the compatibility path and `rm -rf` tolerates a missing bundle.
+pub fn remote_server_removal_command() -> String {
```
moirahuang, line 514: operates on binary and bundle

app/src/remote_server/server_model.rs
```diff
@@ -79,6 +79,21 @@ use crate::terminal::model::session::command_executor::{
     ExecuteCommandOptions, LocalCommandExecutor,
 };
 
+/// Resolves packaged resources installed beside the standalone remote-server executable.
+///
+/// This deliberately does not use `warp_core::paths::bundled_resources_dir`,
+/// whose macOS behavior resolves resources inside an app bundle. Remote-server
+/// executables use the same sibling `resources` layout on every supported host.
+fn bundled_resources_dir_for_executable(executable: &Path) -> Option<PathBuf> {
+    let executable = std::fs::canonicalize(executable).ok()?;
```
moirahuang: canonicalize executable path, then looks for resources dir

app/src/remote_server/ssh_transport.rs
```diff
@@ -145,17 +145,18 @@ impl RemoteTransport for SshTransport {
...
+            // found, and 126 means it exists but is not executable. Any other
+            // non-zero exit (e.g. SSH exit 255 for a dead connection, or
+            // signal termination) is treated as a transport-level failure.
             let code = output.status.code();
             let stdout = String::from_utf8_lossy(&output.stdout);
             log::info!("Binary check result: exit={code:?} stdout={stdout}");
             match code {
                 Some(0) => Ok(true),
-                Some(126) | Some(127) => Ok(false),
+                Some(1) | Some(126) | Some(127) => Ok(false),
```
moirahuang: adding 1 to be not installed

app/src/remote_server/ssh_transport.rs
```diff
@@ -262,8 +263,8 @@ impl RemoteTransport for SshTransport {
     ) -> Pin<Box<dyn Future<Output = anyhow::Result<()>> + Send>> {
         let socket_path = self.socket_path.clone();
         Box::pin(async move {
-            let cmd = format!("rm -f {}", remote_server::setup::remote_server_binary());
-            log::info!("Removing stale remote server binary: {cmd}");
+            let cmd = remote_server::setup::remote_server_removal_command();
```
moirahuang, line 289: updating to also handle the bundle

script/deploy_remote_server
```diff
@@ -113,6 +113,7 @@ FEATURES="release_bundle,crash_reporting,standalone,agent_mode_debug,remote_code
 WARP_BIN="warp"
 BINARY_NAME="oz-local"
 REMOTE_DIR=".warp-local/remote-server"
+BUNDLE_VERSION="unversioned"
```
moirahuang: mirrors REMOTE_SERVER_ARTIFACT_VERSION_UNPINNED in setup.rs

script/deploy_remote_server
```diff
@@ -175,19 +188,38 @@ if ! ssh "$HOST" "command -v rsync" &>/dev/null; then
...
 
-echo "==> Uploading to $HOST:$REMOTE_ABS_DIR/$BINARY_NAME"
+echo "==> Uploading bundle to $HOST:$REMOTE_ABS_BUNDLE_DIR"
 
 # Upload via rsync with delta transfer and compression.
 # After the first deploy, only changed bytes are transferred.
-rsync -z -t --partial --progress \
-  "$BUILT_BINARY" \
-  "$HOST:$REMOTE_ABS_DIR/$BINARY_NAME"
+rsync -z -rlt --delete --partial --progress \
```
moirahuang: sync whole bundle directory, prevent stale skills from lingering

app/src/ai/skills/bundled.rs
```diff
@@ -160,9 +225,12 @@ async fn load_figma_skill_definitions() -> HashMap<String, BundledSkillDefinitio
 }
 
 /// Read bundled skill definitions from the specified directory.
-pub(crate) async fn read_bundled_skills(skills_dir: &Path) -> HashMap<String, ParsedSkill> {
+pub(crate) async fn read_bundled_skills(
```
kevinyang372, line 305: I am not sure I understand what this method is trying to do here. Why do we have both skills_dir and host_skills_dir? And what is resources_dir?

app/src/ai/skills/remote.rs
```diff
@@ -0,0 +1,100 @@
...
+    fn subscribe_to_remote_bundled_skills(&mut self, ctx: &mut ModelContext<Self>) {
+        let remote_server_manager = RemoteServerManager::handle(ctx);
+        ctx.subscribe_to_model(&remote_server_manager, |me, event, ctx| match event {
+            RemoteServerManagerEvent::HostConnected { host_id } => {
+                me.bootstrap_remote_bundled_skill(host_id.clone(), ctx);
+            }
+            RemoteServerManagerEvent::HostDisconnected { host_id } => {
+                me.remove_remote_bundled_skill(host_id);
+            }
+            RemoteServerManagerEvent::SessionConnecting { .. }
```
kevinyang372, line 40: Hmmm do we actually need to list out every single variation here?

crates/remote_server/src/manager.rs
```diff
@@ -946,6 +947,8 @@ pub struct RemoteServerManager {
     sessions: HashMap<SessionId, RemoteSessionState>,
     /// Reverse index: host → sessions for O(1) lookup by `HostId`.
     host_to_sessions: HashMap<HostId, HashSet<SessionId>>,
+    /// Packaged resources directory advertised by each connected remote host.
+    bundled_resources_dirs: HashMap<HostId, String>,
```
kevinyang372: I don't quite understand why we need the workflow of

Remote server advertises what bundled resources are available -> client
Client then fetches the actual content in a later request

Why don't the remote server just pushes a notification to the client on what bundled skills there are and their contents?

crates/remote_server/src/install_remote_server.sh
```diff
@@ -1,5 +1,23 @@
...
-# Installs the Warp remote server binary on a remote host.
+# Installs the Warp remote server on a remote host.
+#
+# Install layout:
+#
+#   {install_dir}/
+#   ├── bundles/{bundle_version}/      ← the artifact's shipped layout, unchanged:
+#   │   ├── {binary_name}                 the executable...
+#   │   └── resources/                    ...and its sibling resources
+#   └── {binary_name}{version_suffix}  ── symlink ─▶ bundles/{bundle_version}/{binary_name}
```
kevinyang372: Why do we need the symlink instead of just having the initialize command start `bundles/<bundle_version>/<binary_name>` directly?

resources/bundled/skills/create-skill/SKILL.md
```diff
@@ -227,7 +227,7 @@ Once all runs are done:
 
 2. **Aggregate into benchmark** — run the aggregation script from the skill-creator directory:
    ```bash
-   python -m scripts.aggregate_benchmark <workspace>/iteration-N --skill-name <name>
+   python {{skill_dir}}/scripts/aggregate_benchmark.py <workspace>/iteration-N --skill-name <name>
```
kevinyang372, line 230: Why these changes?

crates/remote_server/src/setup.rs
kevinyang372: Shouldn't we just remove the bundle here?

crates/remote_server/src/setup.rs
```diff
@@ -495,15 +495,36 @@ pub fn remote_server_binary() -> String {
...
-/// own arguments. A missing binary produces exit 127 (command not
-/// found) or 126 (not executable), and a corrupted binary will fail
-/// with a non-zero exit of its own.
+/// Exits 0 when the compatibility path is a symlink, the bundle executable is
+/// present and executable, the bundle resources directory exists, and the
+/// executable can parse its own arguments.
 pub fn binary_check_command() -> String {
-    format!("{} --version", remote_server_binary())
+    format!(
+        "test -L {} && test \"$(readlink {})\" = \"{}\" && test -x {} && test -d {} && {} --version",
```
kevinyang372: Why do we need to check the bundle as part of the binary check command? It's reasonable to have a working binary without a bundled directory no?

app/src/ai/skills/remote.rs
moirahuang: i intentionally am so we're explicit about the expected behavior, i think it's better than a wildcard?

app/src/ai/skills/bundled.rs
moirahuang: we're reading the skill content from local using `skills_dir` but we're using the host's paths for the skill's files. and `resources_dir` is the host's resources root which is used for `settings_schema_path`

crates/remote_server/src/manager.rs
moirahuang: we're not fetching the actual content in a later request, we're using what is shipped with the client

we shouldn't need to push the content right since the remote resources already has its own skill contents

crates/remote_server/src/install_remote_server.sh
moirahuang: the symlink allows us to 1) swap out to always the same path 2) we have existing places in the code that expect `remote_server_binary()`. i guess we could update that?

crates/remote_server/src/setup.rs
moirahuang: we wouldn't have the skills then, we should check for that right?

crates/remote_server/src/setup.rs
moirahuang: we should also remove the path if we go forward w the symlink case right?

resources/bundled/skills/create-skill/SKILL.md
moirahuang: now that we have `{[skill_dir}}` i think it's more robust than -m. specifically, -m resolves against sys.path which is the cwd

crates/remote_server/src/install_remote_server.sh
```diff
@@ -79,7 +90,23 @@ fi
...
 
-bin=$(find "$tmpdir" -type f -name 'oz*' ! -name '*.tar.gz' | head -n1)
+# The executable and its resources are siblings in the artifact. Exclude the
+# resources tree from the search: bundled skills may ship companion files
+# whose names also start with `oz`.
+bin=$(find "$tmpdir" -type f -name 'oz*' ! -name '*.tar.gz' ! -path '*/resources/*' | head -n1)
 if [ -z "$bin" ]; then echo "no binary found in tarball" >&2; exit 1; fi
 chmod +x "$bin"
+
+# Install the resources tree at the global, version-independent location
```
moirahuang, line 100: a simpler install to just get the global bundle installed

crates/remote_server/src/setup.rs
```diff
@@ -506,6 +506,15 @@ pub fn binary_check_command() -> String {
     format!("{} --version", remote_server_binary())
 }
 
+/// Returns the shell command to remove the current remote-server binary.
+///
+/// The global bundled resources directory is deliberately left in place:
+/// the next install overwrites it, and an older daemon that is still
+/// running parsed its skills at startup.
+pub fn remote_server_removal_command() -> String {
+    format!("rm -f {}", remote_server_binary())
```
moirahuang, line 515: now only removes the binary, leading global resources for the next install to overwrite

crates/remote_server/proto/remote_server.proto
```diff
@@ -239,6 +240,33 @@ message ErrorResponse {
...
+  // When set, the skill is only active while the named MCP integration
+  // is running on the client (e.g. "figma"). Evaluated client-side.
+  optional string requires_mcp = 6;
+}
+
+// Server → client push: the daemon's bundled skill catalog. Sent to a
+// connection after its Initialize completes (when parsing has already
+// finished) and broadcast to all connections when parsing completes.
+// Never blocks the initialize handshake.
+message BundledSkillsSnapshot {
```
moirahuang, line 266: send all skills

crates/remote_server/proto/remote_server.proto
```diff
@@ -239,6 +240,33 @@ message ErrorResponse {
   string message = 2;
 }
 
+// ── Bundled skills ────────────────────────────────────────────────
+
+// One bundled skill, pre-parsed and handlebars-rendered by the daemon
+// against its own filesystem. Named with the `Proto` suffix (like
+// `FileContextProto`) to avoid colliding with the app's `BundledSkill`.
+message BundledSkillProto {
```
moirahuang, line 248: single skill info

app/src/ai/skills/bundled.rs
```diff
@@ -37,7 +38,79 @@ impl BundledSkillActivation {
     }
 }
 
-/// A bundled skill definition with its activation condition and icon.
+/// Catalogs of bundled skills for the local host and connected remote hosts.
+#[derive(Debug, Default)]
+pub struct BundledSkills {
```
moirahuang, line 43: where we actually manage local vs. remote

app/src/ai/skills/bundled.rs
```diff
@@ -159,10 +299,14 @@ async fn load_figma_skill_definitions() -> HashMap<String, BundledSkillDefinitio
         .collect()
 }
 
-/// Read bundled skill definitions from the specified directory.
-pub(crate) async fn read_bundled_skills(skills_dir: &Path) -> HashMap<String, ParsedSkill> {
+/// Read bundled skill definitions from the specified directory, rendering
+/// handlebars variables against this host's filesystem (`resources_dir` is
+/// the resources root the skills belong to).
+pub(crate) async fn read_bundled_skills(
+    skills_dir: &Path,
```
moirahuang, line 306: allows remote to use

app/src/ai/skills/remote.rs
```diff
@@ -0,0 +1,175 @@
...
+fn mcp_integration_from_wire_id(wire_id: &str) -> Option<McpIntegration> {
+    match wire_id {
+        "figma" => Some(McpIntegration::Figma),
+        _ => None,
+    }
+}
+
+/// Converts a daemon-pushed snapshot into a catalog whose skill paths are
+/// remote paths on `host_id`.
+fn bundled_skill_from_protos(host_id: &HostId, skills: &[BundledSkillProto]) -> BundledSkill {
```
moirahuang, line 91: rparses daemon-rendered content so name/description/line-range are correct for remote

app/src/ai/skills/bundled.rs
```diff
@@ -37,7 +38,79 @@ impl BundledSkillActivation {
...
+        self.remote_by_host
+            .get(&remote_path.host_id)?
+            .skill_by_path(path)
+    }
+
+    /// Like [`Self::remote_skill_by_path`], but only returns the skill when
+    /// its activation condition is met.
+    pub fn remote_active_skill_by_path(
+        &self,
+        path: &LocalOrRemotePath,
```
kevinyang372, line 91: Can this just take in a RemotePath?

app/src/ai/skills/bundled.rs
kevinyang372: What does allows remote to use mean? Is this for the remote daemon to use?

kevinyang372: If this is for the client connecting to a remote daemon, we cannot use Path here as it assumes local os encoding

app/src/remote_server/server_model.rs
```diff
@@ -1461,6 +1523,17 @@ impl ServerModel {
             }
         }
 
+        // Push the bundled skill catalog to this connection if it is already
+        // parsed. Enqueued on the same channel as the response below, so the
+        // client buffers it as a push event during the handshake.
+        if let Some(skills) = self.bundled_skills.clone() {
```
kevinyang372, line 1529: Is there a race here where we might flush the bundled skill twice on startup?

kevinyang372 (approved):

PR #12379, opened 2026-06-08
Fix file search filtering

kevinyang372:
## Description
Fixes a regression where cmd-O (command palette file search) and `@` context file search stopped finding files that exist in the repo.

### Source issue
File search fuzzy-matches over a list of repo contents returned by `RepoMetadataModel::get_repo_contents`. That traversal caps results at `MAX_REPO_CONTENTS_RESULTS = 100` (`crates/repo_metadata/src/local_model.rs`) and was being called with no filter, so it returned the **first 100 files in traversal order**. The fuzzy matcher then only ever saw those 100 files. In any repo with more than 100 indexed files, matches outside that prefix were silently dropped — so searching for a real file often returned nothing.

In other words, the search was reversed: we truncated first and matched second, instead of matching first and truncating second.

### Fix
Push the query down into the traversal as a filter so the 100-result cap applies to **matching** files rather than the first files encountered:
- `FileSearchModel::get_repo_contents` / `get_contents_from_repo` now take the query and, for non-empty queries, build `GetContentsArgs::with_filter(...)` whose predicate strips the repo-root prefix and keeps entries where `fuzzy_match_path(relative_path, query).is_some()`. Reusing `fuzzy_match_path` keeps the filter consistent with downstream scoring and handles multi-term/wildcard queries.
- The filter is evaluated *before* an entry counts toward the cap (`collect_contents_recursive`), so a match that sorts past the first 100 entries is still returned.
- Empty (zero-state) queries keep the existing cached, unfiltered behavior. Query-specific results bypass the per-repo cache since they aren't reusable.
- cmd-O and the `@` context menu pass their finalized query into `get_repo_contents`.

### Why we don't fully remove the post-search match
A natural follow-up is "we filter in `get_repo_contents` now, so drop the per-item match in the data sources." We intentionally keep it, because it isn't a pure filter:
- **Scoring + highlighting.** The per-item `fuzzy_match_path` produces the `FuzzyMatchResult` stored on `FileSearchItem.match_result`, which drives both ranking (`score()` → `k_largest_relaxed_by_key`, plus opened-file/recency/directory boosts) and the matched-character highlighting in `render_file_search_row`. The traversal filter only returns a `bool` and discards the score and matched indices, so removing the downstream match would lose relevance ranking and highlighting.
- **Non-filtered folder sources.** Both surfaces also feed cached, *unfiltered* listings through the same code path — cmd-O's `CurrentFolder` mode and the `@` menu's `file_data_source_for_pwd`. For those, the per-item match is the only place filtering happens, so removing it would show the entire folder listing regardless of the query.

So the two passes aren't redundant: the upstream filter bounds the candidate set (the actual fix), and the downstream match scores/highlights and still filters the folder modes.

That being said, this can definitely be optimized. But I am punting on it to get the urgent fix out first. Eitherway, this will be a significant performance improvement since we are not building up a huge in-memory list of every file in the repo metadata as we search

## Linked Issue

## Testing

Automated tests (in `crates/repo_metadata/src/local_model_tests.rs`):
- `test_get_repo_contents_filter_applies_before_cap` — a matching file placed well past the 100-entry cap in traversal order is still returned, and `truncated` stays `false` when only one entry matches. This is the core guarantee of the fix.
- Existing `get_repo_contents` traversal/truncation/gitignore tests continue to pass.

Manual: cmd-O and `@` search in a large repo (this `warp` checkout) now surface deeply-nested files that previously didn't appear.

## Agent Mode

### DIFF
--- a/app/src/search/ai_context_menu/files/data_source.rs
+++ b/app/src/search/ai_context_menu/files/data_source.rs
@@ -62,7 +62,7 @@ pub fn file_data_source_for_current_repo(
                     last_opened,
                 }
             } else {
-                let contents = file_search_model.get_repo_contents(app);
+                let contents = file_search_model.get_repo_contents(&query.text, app);
                 FileSnapshot {
                     contents,
                     git_changed_files: HashSet::new(),
--- a/app/src/search/command_palette/files/data_source.rs
+++ b/app/src/search/command_palette/files/data_source.rs
@@ -109,11 +109,11 @@ impl FileDataSource {
         }
     }
 
-    fn contents(&self, app: &AppContext) -> Arc<Vec<FileSearchResult>> {
+    fn contents(&self, query: &str, app: &AppContext) -> Arc<Vec<FileSearchResult>> {
         match &self.mode {
             FileDataSourceMode::Repo => {
                 let file_search_model = FileSearchModel::as_ref(app);
-                file_search_model.get_repo_contents(app)
+                file_search_model.get_repo_contents(query, app)
             }
             FileDataSourceMode::CurrentFolder { cached_contents } => {
                 Arc::new(cached_contents.clone())
@@ -195,8 +195,6 @@ impl FileDataSource {
     > {
         let file_search_model = FileSearchModel::as_ref(app);
 
-        let contents = self.contents(app);
-
         // Strip any trailing : in case user is in the middle of typing a line / column arg.
         let query_text = query_text.strip_suffix(':').unwrap_or(query_text);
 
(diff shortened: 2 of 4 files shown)


kevinyang372:
* **#12379** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12379?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12379?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

moirahuang (approved):

PR #12383, opened 2026-06-08
Resolve tab group bugs: new tab group expansion, pane name duplication, and settings tab splitting group

johnturcoo:
## Description

This PR resolves several tab grouping bugs, including the following:
- Tab groups not expanding when tabs were added/moved to them. 
- Settings could split a group with "After current tab" placement.
- All panes in a multi-pane tab showed the same (tab) name when moved inside a tab group

The new behavior is as follows:
- Moving tab(s) to a group, or adding tab(s) to a group now expand the target group
- Opening settings via keyboard shortcuts or menu opens the settings tab after the group of the current active tab (if a group exists)
- Panes in a multi-pane tab keep their original name when moved inside a tab group, rather than inheriting the tabs name (only relevant for pane view)

## Linked Issues

[Tab group expansion](https://linear.app/warpdotdev/issue/APP-4678/adding-a-new-tab-expands-group)
[Tab naming bug](https://linear.app/warpdotdev/issue/APP-4677/named-tabs-cause-pane-renaming-when-added-to-a-group)
[Opening settings cause tab group split](https://linear.app/warpdotdev/issue/APP-4690/opening-settings-splits-a-tab-group)

Videos can be found on issues for more context.

## Testing

### Screenshots / Videos

[Demo video for adding/moving tab(s) to group](https://www.loom.com/share/0f7cb551b7bd479f9ead480b05c1e850)

[Demo video for opening settings](https://www.loom.com/share/baea696ecb264d139184ca99dff0facf)

[Demo video for tab naming fix](https://www.loom.com/share/24a31435965e4429a14c4816a3a7c19d)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -6879,6 +6879,14 @@ impl Workspace {
         }
     }
 
+    /// Ensures the group is expanded (not collapsed). No-op if the group does
+    /// not exist or is already expanded.
+    fn expand_tab_group(&mut self, group_id: TabGroupId) {
+        if let Some(group) = self.tab_groups.get_mut(&group_id) {
+            group.collapsed = false;
+        }
+    }
+
     /// Opens the inline rename editor over the given group's header.
     pub fn rename_tab_group(&mut self, group_id: TabGroupId, ctx: &mut ViewContext<Self>) {
         let Some(group) = self.tab_groups.get(&group_id) else {
@@ -6981,6 +6989,7 @@ impl Workspace {
             .map(|i| i + 1)
             .unwrap_or(self.tabs.len());
         self.tabs[tab_index].group_id = Some(group_id);
+        self.expand_tab_group(group_id);
         self.move_tab_to_index(tab_index, target_index, ctx);
 
         if let Some(prev) = previous_group_id {
@@ -7067,6 +7076,7 @@ impl Workspace {
             }
             self.move_tab_to_index(new_idx, target_index, ctx);
         }
+        self.expand_tab_group(group_id);
         ctx.notify();
     }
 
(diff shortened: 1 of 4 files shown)


app/src/workspace/view.rs
```diff
@@ -6879,6 +6879,14 @@ impl Workspace {
         }
     }
 
+    /// Ensures the group is expanded (not collapsed). No-op if the group does
+    /// not exist or is already expanded.
+    fn expand_tab_group(&mut self, group_id: TabGroupId) {
+        if let Some(group) = self.tab_groups.get_mut(&group_id) {
+            group.collapsed = false;
+        }
+    }
```
vkodithala: Do we need a `ctx.notify()` here? Unimportant since I'm assuming you dogfooded this (and demo LG), just noting since we do emit a notify in similar fn `toggle_tab_group_collapsed`. Pretty sure you call `ctx.notify()` in all uses of this fn so shouldn't matter, defer to you.

app/src/workspace/view.rs
```diff
@@ -12102,7 +12112,22 @@ impl Workspace {
                         .push(self.tabs.last().unwrap().pane_group.id());
                     self.activate_tab_internal(self.tab_count() - 1, ctx);
                 } else {
-                    let insert_idx = self.active_tab_index + 1;
+                    // When the new tab won't inherit the active tab's group
+                    // (e.g. settings, notebooks, or other restoration-based tabs)
+                    // but the active tab is inside a group, land after the group's
+                    // last member so the group isn't split in two.
+                    let insert_idx = if active_tab_group_id.is_none() {
```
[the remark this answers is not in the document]
vkodithala: Eh, don't really care if you include this. Demo LG.

app/src/workspace/view.rs
```diff
@@ -12102,7 +12112,22 @@ impl Workspace {
...
+                            .and_then(|t| t.group_id)
+                            .and_then(|gid| {
+                                group_member_indices(&self.tabs, gid)
+                                    .last()
+                                    .map(|last| last + 1)
+                            })
+                            .unwrap_or(self.active_tab_index + 1)
+                    } else {
+                        self.active_tab_index + 1
+                    };
```
vkodithala: This seems functionally correct, but the code was tricky for me to follow.

The case that we want to land the new tab after the active tab group's last member is `active_tab_group_id.is_none()` and `active_tab.is_some()`, which from my understanding is just `is_restoration`? This is a nit but might be clearer to model this conditional around `is_restoration` for clarity.

app/src/workspace/view.rs
```diff
@@ -12111,12 +12136,13 @@ impl Workspace {
             }
         }
 
-        // Inherit the active tab's group membership. D
+        // Inherit the active tab's group membership.
```
vkodithala, line 12139: Nice haha

app/src/workspace/view_tests.rs
```diff
@@ -3690,3 +3690,186 @@ fn test_new_tab_with_after_current_tab_setting_lands_after_active_tab_in_group()
         });
     });
 }
+
+// --- expand-on-add tests ---
```
vkodithala: nit: Let's remove this? Probably not helpful.

vkodithala (approved):
Few nits, non-blocking. LGTM! :rocket:

app/src/workspace/view.rs
johnturcoo: We don't need it because at all call sites we call `ctx.notify()` but then again it could be better that callers don't need to remember to call `ctx.notify()`, and its cheap.

app/src/workspace/view.rs
johnturcoo: Good call, this is definitely clearer!

PR #12420, opened 2026-06-09
Fix tab restoration bug: caused group split

johnturcoo:
## Description

Fix a bug where restoring a closed tab that previously belonged to a tab group would cause the group to split if the group had moved since the tab was closed.

Root cause: `restore_closed_tab` blindly inserted the TabData at the `tab_index` with its original `group_id` still set. If the group's contiguous block had shifted in the meantime, the restored tab would land at the wrong position — violating the group contiguity invariant and visually splitting the group.

Fix: On restore, branch on the group's current state:
•  Group still exists (`tab_groups.contains_key`): append the restored tab after the group's current last member and expand the group so it's immediately visible. This is consistent with how `move_tab_to_group` places tabs.
•  Group was pruned (all other members were also closed): clear `group_id` and fall back to the original `tab_index`. The group metadata is gone at this point (pruned when the last member closed), so there's nothing to rejoin.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4699/tab-restoration-leaves-grouped-tabs-in-broken-state

## Testing

### Screenshots / Videos

[Demo of issue](https://www.loom.com/share/112bd75f169744a688d62ba0ca4f30f9)

[Demo of fix](https://www.loom.com/share/25a139170bdd4a4d99fc501f3340ba58)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -11762,7 +11762,7 @@ impl Workspace {
     pub fn restore_closed_tab(
         &mut self,
         tab_index: usize,
-        tab_data: TabData,
+        mut tab_data: TabData,
         ctx: &mut ViewContext<Self>,
     ) {
         // When restoring a closed tab, we have to reattach its panes so that they know they're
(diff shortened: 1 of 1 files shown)


app/src/workspace/view.rs
```diff
@@ -11771,10 +11771,36 @@ impl Workspace {
...
+        // the group's current last member. If the group no longer exists (it was
+        // pruned when the tab was closed), drop the membership and fall back to
+        // the original index instead.
+        let insert_index = if let Some(group_id) = tab_data.group_id {
+            if self.tab_groups.contains_key(&group_id) {
+                // Group still exists — append after its current last member.
+                group_member_indices(&self.tabs, group_id)
+                    .last()
+                    .map(|i| i + 1)
+                    .unwrap_or(self.tabs.len())
```
vkodithala, line 11784: nit: I think this is fine for now, but an ideal solution here might be to try to preserve the closed tab's position and group in earnest. i.e., keeping some pointer like `group_index` that encodes the tab's relative index in the group for restore. 

But I'm cool w your solution for now and we can maybe rethink if people find this unexpected; I don't think it's gonna be a major drawback since tab restoration isn't rly a hot path.

vkodithala (approved):
LG! Left a very minor comment but this implementation is great. 🚀

app/src/workspace/view.rs
johnturcoo: Yes I thought about doing this as well, but I think its overkill right now.

PR #12429, opened 2026-06-09
Fix shared hover/click state across "Open skill" buttons

vkodithala:
## Description
Fixes the "Open skill" button bug reported in warp#12427: when the agent reads two or more skills in a single block (the repro here is when an agent makes parallel tool calls), hovering one "Open skill" button highlights both, and clicking is flaky/slow.

Root cause: it is **not** that we mint a new `MouseStateHandle` per button — it's the opposite. All skill banners shared a single per-block handle. `MouseStateHandle = Arc<Mutex<MouseState>>`, so the two skill banners (`ReadSkill` and `ReadFiles`) each cloned the same `open_skill_button_handle` / `read_from_skill_button_handle` stored on `AIBlockStateHandles`. Sharing the underlying `MouseState` means hover/click state is shared across every skill button in the block.

Fix: replace the two single handles with a per-action `HashMap<AIAgentActionId, MouseStateHandle>` (`skill_button_handles`), registered in `handle_updated_output` alongside the existing per-action handles (`orchestration_navigation_card_handles`, etc.). Each banner now looks up its own handle by action id, so each button has independent hover/click state. The `OpenCodeInWarp` reset now iterates the per-action handles (same "reset before navigation" intent as before).

## Linked Issue
warp#12427

## Testing
Demo [here](https://www.loom.com/share/e7dee62b8c9d46e7ad6abfabc68a8055).
- `cargo check -p warp --lib --features local_fs` passes
- `cargo clippy -p warp --lib --features local_fs` passes

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/3c4fc723-000e-4978-ae5f-5991647d4467_
_Run: https://oz.staging.warp.dev/runs/019eae3a-7601-7299-909c-f719e8af7fca_

### DIFF
--- a/app/src/ai/blocklist/block.rs
+++ b/app/src/ai/blocklist/block.rs
@@ -477,13 +477,10 @@ pub(super) struct AIBlockStateHandles {
     /// Mouse state handle for AI document created block
     ai_document_handle: MouseStateHandle,
 
-    /// Mouse state handle for 'open skill' button
-    /// from an OpenSkill action banner
-    open_skill_button_handle: MouseStateHandle,
-
-    /// Mouse state handle for 'open skill' button
-    /// from a ReadFiles action banner
-    read_from_skill_button_handle: MouseStateHandle,
+    /// Per-action mouse state handles for the 'open skill' button shown on
+    /// ReadSkill and ReadFiles action banners. Keyed by action id so that
+    /// multiple skill banners in the same block don't share hover/click state.
+    skill_button_handles: HashMap<AIAgentActionId, MouseStateHandle>,
 }
 
 #[derive(Default, Clone, Debug)]
@@ -2007,6 +2004,16 @@ impl AIBlock {
             if matches!(&action.action, AIAgentActionType::StartAgent { .. }) {
                 self.state_handles
                     .orchestration_navigation_card_handles
+                    .entry(action.id.clone())
+                    .or_default();
+            }
+
+            if matches!(
+                &action.action,
+                AIAgentActionType::ReadSkill(_) | AIAgentActionType::ReadFiles(_)
+            ) {
+                self.state_handles
+                    .skill_button_handles
                     .entry(action.id.clone())
                     .or_default();
             }
@@ -6591,10 +6598,7 @@ impl TypedActionView for AIBlock {
             } => {
                 // Resets the interaction states of ReadSkill and ReadFiles tool call banners before opening a new code pane
                 // Avoids an immediate re-hover (and stuck tooltip) while the new code pane is being created
-                for handle in [
-                    &self.state_handles.open_skill_button_handle,
-                    &self.state_handles.read_from_skill_button_handle,
-                ] {
+                for handle in self.state_handles.skill_button_handles.values() {
                     if let Ok(mut state) = handle.lock() {
                         state.reset_interaction_state();
                     }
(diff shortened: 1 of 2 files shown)


app/src/ai/blocklist/block.rs
```diff
@@ -2011,6 +2008,16 @@ impl AIBlock {
...
 
+            if matches!(
+                &action.action,
+                AIAgentActionType::ReadSkill(_) | AIAgentActionType::ReadFiles(_)
+            ) {
+                self.state_handles
+                    .skill_button_handles
+                    .entry(action.id.clone())
+                    .or_default();
+            }
```
vkodithala, line 2019: This is the main point of contention in this PR, but it's a minor one. We're essentially minting a new `MouseStateHandle` for every `ReadFiles` action (cheap), vs. computing if a read file falls within a skill directory at output time (which may not even be accurate, since skill directories change p frequently at runtime). 

I'm opting for the extra mouse handles and don't think it's too malicious, but lmk if you have strong feelings here.

harryalbert: Yeah I don't _love_ this. I've seen agents spam a bunch of file reads and it seems pretty wasteful to reserve a mouse handle for each read if we can avoid it.

How difficult would it be to tighten up the mouse handle provisioning a little more?

harryalbert (commented):
had one thought about the mouse handle provisioning that you called out. Not super blocking but I think worth thinking through a bit more

app/src/ai/blocklist/block.rs
vkodithala: Yeah, I did a pretty shitty job explaining this, mb. Basically, we mint new "Open skill" buttons at render time, [here](https://github.com/warpdotdev/warp/blob/oz-agent/skill-button-shared-hover-state/app/src/ai/blocklist/block/view_impl/output.rs#L1868-L1875), but we allocate mouse states upon receiving new output from the server, [here](https://github.com/warpdotdev/warp/blob/oz-agent/skill-button-shared-hover-state/app/src/ai/blocklist/block.rs#L2011-L2019).

The edge case is that we receive a `ReadFiles` call from the server without a corresponding skill directory at the time we receive it, and so don't allocate a state handle. Then, assume we add a skill directory that includes that file (i.e. moving a `SKILL.md` into a skill directory). At render time, we don't have a `MouseStateHandle` to allocate to the `ReadFiles` call, so we just don't render the "Open skill" button. Does that make sense?  


I will say this is a major edge case, but my oo1 is that the cost of reserving extra mouse state handles for the sake of correctness is probably OK, unless these are expensive (could be missing smth). Lmk.

vkodithala:
* **#12429** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12429?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12429?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/blocklist/block.rs
harryalbert: ahh I see what you're saying — yeah I'm not worried about that edge case, just misunderstood the scope/severity of your original comment

harryalbert (approved):

Issue #11754, opened 2026-05-27
Feature request: Persist and auto-reopen named AI chats on startup


ocnova-lab:
## Feature Request

**Summary:** Allow users to pin/bookmark named AI chats so they automatically reopen in the sidebar panel on Warp startup.

**Use case:**

I maintain several ongoing AI chat sessions in Warp's sidebar, each dedicated to a specific workflow context (e.g. "Educational processes", "Marketing", "Methodologists", "Reports", "Legal/Tax topics"). Every time I restart Warp, I have to manually find and reopen these chats from the list.

**Desired behavior:**

- Users should be able to mark specific chats as "pinned" or "startup chats"
- On Warp startup, these pinned chats appear open in the sidebar automatically (even if empty / no new messages)
- Ideally, this could be configurable via `~/.warp/settings.toml` or a dedicated config file, consistent with how tab configs work in `~/.warp/tab_configs/`

**Current behavior:**

Chats are saved (cloud storage works), but none reopen automatically — the user must manually open them each session.

**Why this matters:**

For users who use Warp AI as a structured work tool with named, context-specific chats, having to re-navigate to the same chats every session adds friction. Pinned/persistent chats would make Warp feel more like a workspace that remembers your setup — similar to how IDEs restore open tabs/panels on restart.

**Possible implementation hint:**

Could be as simple as a `pinned_chats` field in `settings.toml`, or an extension to the existing tab config format to support a `type = "ai_chat"` pane.

PR #12431, opened 2026-06-09
Resume/retry cloud runs on transport failures instead of failing them

advait-m:
# Resume cloud agent runs on transport failures (REMOTE-1894)

Specs: [`specs/REMOTE-1894/PRODUCT.md`](https://github.com/warpdotdev/warp/blob/advait/remote-1894-resume-cloud-runs-on-transport-failures/specs/REMOTE-1894/PRODUCT.md) (behavior invariants) · [`specs/REMOTE-1894/TECH.md`](https://github.com/warpdotdev/warp/blob/advait/remote-1894-resume-cloud-runs-on-transport-failures/specs/REMOTE-1894/TECH.md) (implementation + validation mapping)

## Problem

Cloud agent runs (Oz) died with terminal errors when the MAA response stream hit transient network failures (ConnectionReset, TLS close_notify, silent EOF). The client had resume machinery, but it was gated incorrectly, raced with terminal ERROR reporting, and was dropped entirely on some code paths. See REMOTE-1894 for the catalogued failure modes from the BDR pipeline runs.

### `ConversationStatus::TransientError` (client-only state)

A failure with automatic recovery pending is now a first-class, non-terminal conversation status instead of a terminal `Error` with a `will_attempt_resume` rendering flag smuggled alongside:

- Renders as "Reconnecting" with an in-progress treatment.
- `local_agent_task_sync_model` maps it to `IN_PROGRESS` ("Connection lost while receiving the agent response; attempting to resume.") so the server/UI never see a spurious terminal ERROR.
- The agent driver holds the run open for `AUTO_RESUME_TIMEOUT` (120s) while a conversation is in `TransientError`; if recovery doesn't land in time, the run ends with the last structured error.
- `RenderableAIError::will_attempt_resume` remains as a rendering-only hint; nothing makes control-flow decisions off it anymore.
- User cancellation while a recovery is parked aborts the pending resume and surfaces `Cancelled` (the conversation doesn't linger in "Reconnecting"), and terminal transitions out of `TransientError` run the same finished-conversation handling (e.g. queued prompts) as transitions out of `InProgress`.

### Strict retry/resume split

`recovery_action()` (pure function, unit-tested matrix) decides recovery for every failed attempt:

- **No client actions received yet** → re-send the identical request (in-request retry, budget of 3). Safe because nothing has executed.
- **Client actions received** → one-shot `ResumeConversation` request that reconciles from server state. Re-sending verbatim would be unsafe; resumes run with `can_attempt_resume_on_error=false` so recovery is bounded.
- **Pre-action retry budget exhausted while online** → terminal failure (same split as master; earlier revisions of this branch allowed a resume fallback here).
- Application-level failures (quota, overload) stay terminal for resume purposes: a resume would fail identically or add load the server shed.

### Hold recovery until connectivity returns

Instead of bailing when offline, pre-action retries park on `NetworkStatus::wait_until_online()` (`RetryWhenOnline`). While parked, the conversation shows `TransientError` via `WaitingForNetwork` events and flips back to `InProgress` when the retry fires. Cancellation invalidates parked retries through `current_request_id`.

### `AIApiError::StreamTruncated` synthesized in `ResponseStream`

The server always sends a `StreamFinished` event before ending the response, but the transport can truncate between chunks, surfacing as a clean EOF. `ResponseStream` now tracks `stream_finished_received` / `error_event_emitted` and synthesizes `StreamTruncated` (retryable + transient) when a stream completes without either, running the same recovery decision as any other failure. This replaces the controller-level EOF fallback (now a defensive warn). HTTP send failures don't take this path (they arrive as in-stream error events); client-side request-conversion failures — where no stream was ever created — are surfaced terminally with the original error instead of being synthesized as `StreamTruncated`.

**This case is empirically reproducible, not theoretical** (it was questioned in review whether the server contract makes it impossible). In the e2e runs below, the server's mid-stream abort surfaced client-side as a **clean EOF with no transport error event** — even over localhost with zero intermediaries:

- Every failure surfaced through the synthesized-EOF path: the warn `generate_multi_agent_output stream ended without emitting StreamFinished event.` preceded each recovery, and the only error reports carried `StreamTruncated`'s message ("Response stream ended unexpectedly before completion.").
- Zero `Transport(...)` errors anywhere in the client log — the stream's `Err`-event arm never fired (no `MultiAgent request failed, retrying…` lines).
- Server log confirmed the response ended without `StreamFinished` (`simulating transport reset (dev-only); aborting response stream`).

So all recovery in the e2e runs was driven entirely by the synthesized `StreamTruncated` path. The contract holds at the application layer — the server always *sends* the event — but a mid-stream termination (connection cut between chunks, LB/proxy idle timeout, server crash) can still surface to the client as a clean end-of-stream rather than an error; in production (Cloud Run + load balancers) there are more layers that can swallow an abrupt close the same way. The REMOTE-1894 prod runs themselves surfaced as explicit transport errors through the error-event path — the EOF variant is the same failure class surfacing differently, and master already carried a hardcoded EOF-fallback branch, evidence that clean EOFs predate this PR. Both shapes now feed the same `recovery_action` decision.

## Behavior changes (relative to master)

- Failures with automatic recovery pending no longer surface as terminal: the conversation reports `TransientError`, task sync reports `IN_PROGRESS`, and the driver holds the run open for up to 120s while recovery lands. On master, the sync model reported terminal ERROR immediately — killing the run — even when the client was about to resume (the core REMOTE-1894 bug).
- Streams that end without a `StreamFinished` event (silent EOF) now recover like any transport failure (retry pre-actions, resume post-actions). On master this path was hardcoded as a non-resumable terminal error. Request-send failures now retry too.
- Recovery waits for connectivity: pre-action retries park on `NetworkStatus::wait_until_online()` (master required `is_online` to retry and otherwise failed immediately), and the auto-resume spawn likewise waits for connectivity before sending.
- Auto-resume eligibility narrowed from `is_retryable` to `is_transient_failure`: application-level failures (quota, overload) no longer trigger an auto-resume. On master these set the resume flag.
- Unchanged from master, called out for reviewers: resume still requires received client actions, and pre-action retry budget exhaustion while online remains terminal. Earlier revisions of this branch had relaxed both; the final version restores master's split.

## Validation

- `cargo nextest` on the affected suites (response_stream recovery matrix, server_api error classification, history model status plumbing, task sync mapping): 85 passed.
- `cargo fmt`, `cargo clippy --lib --tests -- -D warnings`, `cargo check --features crash_reporting` all clean.
- E2e on oz-local with the `TransportReset` LLM mock (warp-server `simulate_maa_transport_reset`):
  - **Recovery**: single mid-stream reset → surfaced as clean EOF → synthesized `StreamTruncated` (`stream ended without emitting StreamFinished` in the client log) → conversation `TransientError` → driver "automatic recovery pending — waiting up to 120s" → resume fires → run completes with the mocked `finish_task`.
  - **Bounded failure**: cycling resets → exactly one resume attempt, then terminal run ERROR with "Warp lost connection while receiving the agent response. This is usually temporary." — no retry storm (2 LLM calls total).
  - Both scenarios exercised the synthesized silent-EOF path exclusively (see the `StreamTruncated` section above) — the stream never yielded a transport `Err` event.
  - Note: on cloud runs the user-message-append ClientActions precede the LLM call, so mid-stream failures recover via resume; the pre-action retry paths are covered by the unit-test matrix.
  - Caveat: the mock abort simulates the class of failure, not byte-level equivalence with prod ConnectionReset/close_notify. The design doesn't depend on which way it surfaces — `Err` event and silent EOF feed the same `recovery_action` decision.

Example with mock error locally:
image

## Out of scope

- bb8 session-share connection timeouts (REMOTE-1878).
- Server-side changes (in-turn LLM retries and turn-level retry landed separately in warp-server #11754/#11755/#11770).

---

CHANGELOG-BUG-FIX: Cloud agent runs and local agent conversations now automatically recover from transient network failures mid-turn (showing a "Reconnecting" status) instead of failing the run.


app/src/ai/blocklist/controller.rs
```diff
@@ -2832,9 +2837,23 @@ impl BlocklistAIController {
                         "generate_multi_agent_output stream ended without emitting StreamFinished event."
                     );
 
+                    // The stream died mid-turn without surfacing a transport error (e.g. an
+                    // abrupt EOF). Treat it like a retryable mid-stream failure and attempt an
+                    // automatic resume when this request is allowed to.
```
advait-m: from call: shouldn't be possible with server contract?

app/src/ai/blocklist/controller.rs
```diff
@@ -2832,9 +2837,23 @@ impl BlocklistAIController {
...
+                    let will_attempt_resume =
```
advait-m: from call: switch to TRANSIENT_ERROR state (introduce this)

app/src/ai/blocklist/controller/response_stream.rs
```diff
@@ -293,12 +298,18 @@ impl ResponseStream {
                     return;
                 }
 
-                // If we can't retry (because client actions were received) but the error is
-                // retryable and we're allowed to attempt a resume, signal that the controller
-                // should resume the conversation after the stream completes.
-                let should_attempt_resume = self.has_received_client_actions
```
advait-m, line 299: from call: retry is correct for no client actions (not resume) - we should wait till online if offline (we have some tooling for future)

maybe remove is_online check and delay sending till we're online

advait-m: for both retries + resumes, we should hold request till we're online

app/src/ai/blocklist/controller.rs
advait-m: ^ confirmed that this can happen with a mock transport/from example traces; copying over slack msg:

Following up on the "clean EOF without StreamFinished shouldn't be possible" piece from our call - was able to repro it with the TransportReset LLM mock on oz-local (warp-server aborts the MAA response mid-stream, dev-only flag)

Interestingly, the abort reached the client as a clean end-of-stream, not a transport error - even on localhost with zero proxies in between. In both e2e runs, the only failure-path line in the client log was the synthesized-EOF warn (stream ended without emitting StreamFinished event); the stream's Err arm never fired at all. So all recovery went through the synthesized StreamTruncated path.

When it happens: the contract holds at the app layer (as you said, we wouldn't expect this to break) - the server always sends StreamFinished - but if the connection dies between chunks (reset mid-body, server crash, etc), the body just ends and reqwest surfaces EOF rather than an error. Prod has Cloud Run + LBs that can swallow an abrupt close the same way, and it matches the REMOTE-1894 runs that died through master's old silent-EOF fallback branch.

I think we'll want to cover this case - error events and silent EOF can feed the same recovery_action decision.

app/src/ai/blocklist/controller.rs
advait-m: done

app/src/ai/blocklist/controller/response_stream.rs
advait-m: done (separate retries and resumes - refactored into a func to be very clear) and done (hold till online)

app/src/ai/blocklist/controller/response_stream.rs
```diff
@@ -263,85 +328,175 @@ impl ResponseStream {
...
+        }
+
+        // The server always sends a StreamFinished event before ending the response,
+        // but a transport cut between chunks surfaces as a clean EOF. Synthesize the
+        // failure and recover like any transient error.
+        if !self.stream_finished_received && !self.error_event_emitted {
+            log::warn!(
+                "generate_multi_agent_output stream ended without emitting StreamFinished event."
+            );
+            let truncated = Arc::new(AIApiError::StreamTruncated);
```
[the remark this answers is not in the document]
advait-m: Fixed - that path only carries deterministic request-conversion failures (send failures arrive as in-stream error events), so it shouldn't have been retried or labeled `StreamTruncated`. The original error is now surfaced terminally.

app/src/ai/agent/conversation.rs
```diff
@@ -871,6 +871,12 @@ impl AIConversation {
     pub fn status(&self) -> &ConversationStatus {
         &self.status
     }
+
+    /// Test-only setter for driving status-dependent logic directly.
+    #[cfg(test)]
+    pub(crate) fn set_status_for_test(&mut self, status: ConversationStatus) {
+        self.status = status;
+    }
```
vorporeal, line 879: i would love to know why agents fail to add blank lines before/after code blocks 😅

app/src/ai/agent/mod.rs
```diff
@@ -715,6 +715,7 @@ impl From<&AIApiError> for RenderableAIError {
             {
                 Self::transient_network_error(false, false)
             }
+            AIApiError::StreamTruncated => Self::transient_network_error(false, false),
             _ => Self::Other {
```
vorporeal: we should not use a catch-all here.  seems ripe for "oops we ended up in the wrong state" problems.

app/src/ai/agent_sdk/driver.rs
```diff
@@ -2957,6 +2957,34 @@ impl AgentDriver {
                         return;
                     }
 
+                    if conversation.status().is_transient_error() {
+                        // An automatic recovery is in flight. Don't terminate yet, but bound
+                        // the wait so the CLI doesn't hang if it never completes; a successful
+                        // recovery returns to InProgress, which cancels this deadline.
+                        log::info!("Transient error; automatic recovery pending — waiting up to {AUTO_RESUME_TIMEOUT:?}");
+                        log::info!(
```
vorporeal: these two seem redundant?  i'd remove the first one.

app/src/ai/ambient_agents/mod.rs
vorporeal: should this be a `match`?  (ideally an exhaustive one)

app/src/ai/blocklist/controller.rs
```diff
@@ -2588,6 +2591,26 @@ impl BlocklistAIController {
             .in_flight_response_streams
             .try_cancel_streams_for_conversation(conversation_id, reason, ctx)
         {
+            // A parked auto-resume was aborted above; nothing else will move the
+            // conversation out of TransientError, so surface the cancellation directly.
```
vorporeal, line 2635: can you help me understand this?  i would (naively) expect that cancelling a conversation unconditionally puts it into a cancelled state.

app/src/server/server_api_tests.rs
vorporeal, line 1: do you think we need these tests?  they're basically directly asserting some pretty straightforward match logic.

app/src/server/server_api.rs
```diff
@@ -326,6 +332,35 @@ impl AIApiError {
...
         }
     }
+
+    /// Returns whether this is a transient network/server failure where a fresh
+    /// request is likely to succeed once connectivity or the service recovers.
+    ///
+    /// Gates automatic conversation resumes; intentionally narrower than
+    /// [`Self::is_retryable`]: auto-resuming on application-level failures (quota,
+    /// overload) would fail identically or add load the server just shed.
+    pub fn is_transient_failure(&self) -> bool {
```
vorporeal: i'm not sure i understand the distinction between `is_retryable()` and `is_transient_failure()`.

for one, some of these non-transient-failures seem miscategorized.  e.g.: `ServerOverloaded` is definitely transient (unless we're _very_ overloaded, we'd want to retry and hope that things succeed on the next request).

under what circumstances would we be willing to retry an error, but not _treat_ it as a transient error?

app/src/server/server_api.rs
```diff
@@ -185,6 +185,12 @@ pub enum AIApiError {
         #[source]
         source: anyhow::Error,
     },
+
+    /// Synthesized client-side when a response stream ends without a stream-finished
+    /// event: the server always sends one, but the transport can truncate the response
+    /// between chunks, surfacing as a clean EOF.
+    #[error("Response stream ended unexpectedly before completion.")]
+    StreamTruncated,
```
vorporeal: thoughts on naming this `UnexpectedEof` or `PrematureEof`?  slightly better matches the semantics, i think.  could even go with `EofBeforeStreamFinishedEvent`, if we want to be super explicit (and not require someone to read the doc comment to understand what it means.

app/src/ai/blocklist/local_agent_task_sync_model.rs
```diff
@@ -320,6 +320,14 @@ fn map_conversation_status(
         // matches the local view.
         ConversationStatus::WaitingForEvents => (AgentTaskState::InProgress, None),
         ConversationStatus::Success => (AgentTaskState::Succeeded, None),
+        // An automatic recovery is pending: keep the run IN_PROGRESS so the cloud
+        // execution isn't torn down out from under it.
+        ConversationStatus::TransientError => (
+            AgentTaskState::InProgress,
+            Some(TaskStatusUpdate::message(
+                "Connection lost while receiving the agent response; attempting to resume.",
+            )),
```
vorporeal: where does this surface to the user in the UI?  do we actually want to show this?  or do we just "swallow" transient errors, given we hope that they'll resolve themselves?

app/src/ai/blocklist/local_agent_task_sync_model.rs
```diff
@@ -357,6 +359,21 @@ fn map_conversation_status(
...
+    error: Option<&RenderableAIError>,
+) -> (AgentTaskState, Option<TaskStatusUpdate>) {
+    match error {
+        Some(error) => classify_renderable_error(error),
+        None => (
+            AgentTaskState::Error,
+            Some(TaskStatusUpdate::message("Agent encountered an error")),
+        ),
+    }
+}
```
vorporeal, line 370: can we just have the tests call `map_conversation_status()` instead of breaking out this helper?  doing this creates an opportunity for the real production codepath (through `map_conversation_status()`) to return a different value than what is being checked in tests.

app/src/ai/blocklist/history_model.rs
```diff
@@ -1895,9 +1895,12 @@ impl BlocklistAIHistoryModel {
         });
     }
 
+    /// Marks the stream's exchanges as finished with `error`. `recovery_pending` moves
+    /// the conversation to the non-terminal `TransientError` status instead of `Error`.
```
vorporeal: second half of this comment isn't relevant to a reader, i don't think.  (common case of an agent adding unnecessary comments explaining code changes)

app/src/ai/agent/conversation.rs
```diff
@@ -2155,10 +2164,16 @@ impl AIConversation {
...
+    /// Marks the in-flight request's exchanges as finished with `error`.
+    ///
+    /// `recovery_pending` moves the conversation to the non-terminal `TransientError`
+    /// status instead of `Error`, so consumers don't treat it as dead while an
+    /// automatic recovery is in flight.
     pub fn mark_request_completed_with_error(
         &mut self,
         stream_id: &ResponseStreamId,
         error: RenderableAIError,
+        recovery_pending: bool,
```
vorporeal, line 2185: is it possible for us to define `recovery_pending` as a function of `RenderableAIError`, or is it dependent on some state that is not captured within `RenderableAIError`?

app/src/ai/agent/conversation.rs
vorporeal, line 4371: not something we need to address here; more of a general note:

having these helpers encourages us to do things like `if is_transient_error() || is_blocked()`, which is less resilient to bugs when adding new states than using exhaustive match expressions.

vorporeal (commented):
generally looks good but want to align on some data model things before approving

app/src/ai/agent/conversation.rs
advait-m: ah shoot, I thought I fixed all of these 😅, thanks, fixed; and agreed haha

app/src/ai/agent/mod.rs
advait-m: agreed; fixed (debated a small variable for generic error obj, but not sure that's worth the readability hit)

app/src/ai/agent_sdk/driver.rs
advait-m: ah agree the `AUTO_RESUME_TIMEOUT` part is redundant, removing first line

app/src/ai/ambient_agents/mod.rs
advait-m: hmm, yeah exhaustive match would be nice here - updated (pulled some common code for some match branches into a helper)!

app/src/server/server_api.rs
advait-m: yeah that's a good point, I like the rename - thinking `UnexpectedEof` is best out of these, matching Rust stdio error as well (and the semantics match that error). `EofBeforeStreamFinishedEvent` could be nice, but thinking could be problematic if `StreamFinished` naming ever drifts

app/src/ai/blocklist/history_model.rs
advait-m: agreed; cut comment

app/src/ai/agent/conversation.rs
advait-m: noted - thanks for calling out; will keep in mind & avoid these helpers/prefer exhaustive matches generally

app/src/ai/blocklist/local_agent_task_sync_model.rs
advait-m: so, on the client-side, we generally convert this to `IN_PROGRESS` (to not show errored statuses in places like pill bar, etc)

but on stuff like the Oz runs list or public API for GET `/agent/runs/{run_id}`, we would return this message (which has some marginal value perhaps)

though one catch: found a bug when digging into this - `update_agent_task` treats `status_message=None` as no-change, and recovery sends `None` on InProgress/Success, so this line lingers as the task's final `status_message` even after a successful resume (the happy path). 

clearing it properly needs a server-side fix (no clear sentinel in the input), and the value here is marginal - so i'm aligned, updating this to just going to swallow it client-side. can add it back later with proper clearing if we decide the transparency's worth it.

app/src/server/server_api.rs
advait-m: this is more of a `is_resume_eligible`

yeah, this was something I was trying to figure out the right buckets for (was originally going narrower than retries, given resuming is more expensive) - for `ServerOverloaded`, thinking was that we may not want to send the server traffic if it's already overloaded, but you're right that we probably just want to retry/resume and expect it to be back to normal hopefully.

The other one that didn't make sense to me to resume on was `QuotaLimit` - I don't think this one makes sense to retry on either? Is there some context on why we retry for that one? (not gonna change that logic) @vorporeal 

but yeah, let me just combine these two concepts, since I think I agree that the other classes of errors are fine to try resuming on (was generally thinking they might be more deterministic, but really depends on the error)

I'm renaming the original function to `is_recoverable` to cover both retry/resume

app/src/ai/blocklist/controller.rs
advait-m: it does end up `Cancelled` - but via the stream-cancel path (`try_cancel_streams_for_conversation` → `AfterStreamFinished` → `mark_response_stream_cancelled`), not a direct write. this branch only fires when there's no stream.

the exception is a conversation parked in `TransientError`: we abort its auto-resume above, and with no stream to carry the cancel, nothing else gets it off "Reconnecting" - so we set `Cancelled` directly here (we should only need this for the new state/logic). the guards just scope it to that stuck state.

adding a comment here to explain in-code too
```
// No active stream whose cancellation would mark the conversation `Cancelled`.
```

app/src/ai/agent/conversation.rs
advait-m: `will_attempt_resume`/`waiting_for_network` are still on the error but render-only now (they drive the "attempting to resume…" text). `ConversationStatus::TransientError` is the source of truth, and the `LocalAgentTaskSyncModel` sync model explicitly ignores the flag and keys off status.

I think `recovery_pending` isn't really derivable from the error: it's the output of `recovery_action`, which keys off runtime state the error doesn't carry (whether client actions already streamed, resume budget, `can_attempt_resume_on_error`, online). And post-`is_recoverable`-unification, quota/overload are resume-eligible but convert to `QuotaLimit`/`ServerOverloaded` (no `will_attempt_resume` field), so we can't read it back off the flag. 

related: `RenderableAIError` is serialized with the exchange, but `TransientError` isn't - on restart we recompute a conversation's status from its exchanges, and a finished-with-error exchange comes back as terminal `Error`. which we want, since after a restart there's no live stream or resume in flight. if we instead derived status from the error, a persisted `will_attempt_resume=true` would reload the convo as `TransientError` ("reconnecting") with nothing left to ever clear it. 

so keeping `recovery_pending` explicit and temporary likely makes sense? But open to ideas on structuring this

app/src/ai/blocklist/local_agent_task_sync_model.rs
advait-m: good point - requires a bit more plumbing, but updated tests to call `map_conversation_status`

app/src/server/server_api_tests.rs
advait-m: yeah that's true; removed

app/src/ai/agent/conversation.rs
vorporeal: yeah that makes sense.  no need to change things, was mostly curious if it was redundant state that we could avoid.

app/src/ai/blocklist/controller/response_stream_tests.rs
vorporeal, line 1: not sure we need all of these tests to evaluate some pretty simple logic?  also, the fact that we're just passing a bunch of bools to the function under test makes it hard to easily know if these are even correct.

app/src/ai/blocklist/controller.rs
vorporeal: oof, the fact that some of the logic here that ensures we're in the correct state depends on an event that's conditionally fired (depending on whether or not there's an active network request) is a bit concerning from a correctness perspective.  no concrete suggestion on what to do about it, but 😬

app/src/ai/blocklist/controller.rs
```diff
@@ -2786,6 +2815,29 @@ impl BlocklistAIController {
...
+                    log::warn!("Could not find conversation for response stream: {stream_id:?}");
+                    return;
+                };
+                // Mirror the parked-retry state on the conversation: TransientError while
+                // waiting for connectivity, back to InProgress when the retry fires.
+                let status = if *waiting {
+                    ConversationStatus::TransientError
+                } else {
+                    ConversationStatus::InProgress
+                };
```
vorporeal, line 2876: this also doesn't entirely make sense to me?  nothing else in this branch implies anything about error states - do we only ever get a `WaitingForNetwork` event in the case of an error/retry?  seems like something that theoretically could happen if there's some slow tool call (e.g.: a grep), and by the time it finishes and we're ready to make the next request, we're offline.

vorporeal (approved):

app/src/ai/blocklist/controller/response_stream_tests.rs
advait-m: yeah that's true - removing

app/src/ai/blocklist/controller.rs
advait-m: yeah, hmm, it's not ideal - root cause is the parked auto-resume lives outside `in_flight_response_streams`, so the usual stream-teardown cancel path can't fire for it, hence the direct-write special case

the clean fix is maybe making the parked resume a first-class cancelable so cancellation always flows through one completion path (expose `cancel()` interface and emit completion signal) - but that's a bigger refactor 👀 

adding in a TODO/linear for this

app/src/ai/blocklist/controller.rs
advait-m: hmm yeah, traced it - `WaitingForNetwork` is emitted from exactly one place, `defer_retry_until_online`, which is only reached via `RecoveryAction::RetryWhenOnline`. That arm only fires when a request attempt failed with a recoverable error while offline (`!has_received_client_actions && is_recoverable && has_retry_budget && !is_online`). So it's always preceded by a failed attempt.

the grep scenario does the same thing: the next request still gets attempted and fails (offline transport error) first, and only then parks the retry + emits `WaitingForNetwork`. so `TransientError` ("reconnecting") is accurate there too, I think?

you're right that the coupling was implicit though - the handler assumed `WaitingForNetwork == transient` but that's only enforced by the emitter in another file. Added comments at both the event def and the handler clarifying the invariant for if someone adds a non-error emission later.

app/src/ai/blocklist/controller.rs
seemeroland: I'm a bit confused on this logic. It looks like we added a path to mark the conversation as canceled, if we call this while there's no stream and the conversation is in transient error status? Why do we need this?

Trying to track down sources of conversation cancellation that are not initiated by user

advait-m: not a non-user cancel - this only fires from `cancel_conversation_progress` (user stop). 

a transport failure parks an auto-resume in `TransientError` with no live stream, so the usual cancel → `AfterStreamFinished` → `Cancelled` path can't fire and it'd stay stuck on "Reconnecting" - hence the direct write, scoped by `is_transient_error()`.

PR #12432, opened 2026-06-09
Fix horizontal tab dragging bugs for tab groups

johnturcoo:
## Description

Fixes two bugs in horizontal tab-bar dragging around tab groups:

**1. Dragging a tab slowly into a group from the right could split the group**

Cause: dragged tab was trying to swap positions with the last tab before registering as inside the group. 

Fix: add 8px of trailing padding inside the expanded group container, giving the right edge the same buffer the header gives the left. The entrance zone now extends past the last member's edge, so entrance always registers first. Collapsed groups are unchanged, and the inward `EDGE_MARGIN` still leaves a dead strip between adjacent groups so a tab can be dropped between them without joining either.

**2. Dragging across a collapsed group could eject one of its members**

Repro: `[collapsed][expanded (tab1)(tab2)] (tab3)` — dragging tab3 leftward across both groups would pop a member out of the collapsed group (it landed ungrouped between the two groups).

Root cause: `DragTab { tab_index }` carries an index captured at render time by the tab's `Draggable` closure. Mid-drag mutations (swaps, the hop over a collapsed block) reorder `self.tabs`, and mouse events arrive faster than frames repaint — so in-flight events can carry a stale index that now points at a different tab. After the hop shifts the collapsed members by one slot, a stale event runs the group-membership logic on a collapsed member and rips it out of its group.

Fix: `on_tab_drag` now resolves the dragged tab by identity — the tab whose `DraggableState` reports an active drag — and only trusts the event index when it matches. A bounds guard also drops late events whose index no longer points at a real tab (previously a latent panic via direct indexing in the groups block). This mirrors how group drags were already immune: `DragGroup` dispatches a stable `TabGroupId` instead of an index.

No behavior change in the steady state: when the event index is accurate the identity check is a no-op, for both horizontal and vertical layouts.

**NOTE**  The second fix is a bit of a bandaid, and I can follow up to clean up this pattern. Logically it is correct and I have tested that dragging was not regressed.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4679/dragging-a-tab-in-horizontal-tabs-causes-group-to-split

## Testing

### Screenshots / Videos

[Demo of bug 1](https://www.loom.com/share/2741c5cd82234a45a1bb43c88810b9ba)

[Demo of bug 2](https://www.loom.com/share/d588be6822e44f67b2e6fe40e569eeab)

[Demo of fix](https://www.loom.com/share/6f6d6223882c4ed2b0ef957714f96aea)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -26777,6 +26777,8 @@ impl Workspace {
         // it out of its group. The dragged tab is the one whose
         // `DraggableState` reports an active drag, so trust that identity
         // over the captured index.
+        //
+        // TODO(johnturcoo): determine the right shape for a long-term solution.
         let current_index = if self
             .tabs
             .get(current_index)
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -19027,14 +19027,19 @@ impl Workspace {
             }
         }
 
-        let container = Container::new(row.finish())
-            .with_border(
-                Border::all(1.)
-                    // Left border only on the first slot to avoid double borders.
-                    .with_sides(false, is_first_in_bar, false, true)
-                    .with_border_fill(internal_colors::fg_overlay_1(theme)),
-            )
-            .finish();
+        // Additional padding on expanded groups,
+        // allowing tabs to be dropped into the last position of the group.
+        const EXPANDED_GROUP_TRAILING_PADDING: f32 = 8.;
+        let mut container = Container::new(row.finish()).with_border(
+            Border::all(1.)
+                // Left border only on the first slot to avoid double borders.
+                .with_sides(false, is_first_in_bar, false, true)
+                .with_border_fill(internal_colors::fg_overlay_1(theme)),
+        );
+        if !is_collapsed {
+            container = container.with_padding_right(EXPANDED_GROUP_TRAILING_PADDING);
+        }
+        let container = container.finish();
 
         let group_id = group.id;
         let group_draggable_state = group.draggable_state.clone();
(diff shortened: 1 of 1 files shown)


app/src/workspace/view.rs
```diff
@@ -19027,14 +19027,19 @@ impl Workspace {
...
+        let mut container = Container::new(row.finish()).with_border(
+            Border::all(1.)
+                // Left border only on the first slot to avoid double borders.
+                .with_sides(false, is_first_in_bar, false, true)
+                .with_border_fill(internal_colors::fg_overlay_1(theme)),
+        );
+        if !is_collapsed {
+            container = container.with_padding_right(EXPANDED_GROUP_TRAILING_PADDING);
+        }
+        let container = container.finish();
```
vkodithala, line 19042: Makes sense! Per your PR description we give the same buffer to the left edge; I haven't verified, but maybe it makes sense to just make this unconditional vs. adding this in the collapsed state only? 

This is good for a hotfix but I'd make a small thread with @peterrajani describing your changes to see if he has any input.

app/src/workspace/view.rs
```diff
@@ -26762,6 +26767,31 @@ impl Workspace {
...
+            current_index
+        } else {
+            self.tabs
+                .iter()
+                .position(|tab| tab.draggable_state.is_dragging())
+                .unwrap_or(current_index)
+        };
+        if current_index >= self.tabs.len() {
+            return;
+        }
```
vkodithala, line 26796: You noted this in your PR description, but yeah, it seems very weird to me that `current_index` here deviates from the true index of the tab being dragged. Fine for now and your demo LG, but I'd add a TODO here to investigate the right shape of a solution in the future.

vkodithala (approved):
You might wanna get a postmortem review from @peicodes since he's a bit more familiar with the tab drag path from his work with cross-window tab drag. This LGTM, left a few comments to think on but non-blocking!

app/src/workspace/view.rs
johnturcoo: Will do, thanks!

PR #12456, opened 2026-06-10
Add GitHubRepoModel

MaggieShan:
## Description 
* Addresses comment in https://github.com/warpdotdev/warp/pull/10930#pullrequestreview-4349008394
* This PR refactors GitHub-specific repository information into a dedicated `GitHubRepoModel`. Previously, this logic was coupled with the general `GitRepoStatusModel`, and this change separates those concerns to improve maintainability and clarify the data flow for PR information
  * This removes the brittleness that came with explicitly registering pr info subscribers
  * Since then, `repository_info` was also added to `GitRepoStatusModel` which had the same issues and similarly executes a `gh` command - by moving both fields to their own model, we can more cleanly support callers like `BlocklistAIContextModel` which only needed the two github info fields

## Linked issue 
This is a prereq for adding support for git chips over remote environments 
[APP-4560](https://linear.app/warpdotdev/issue/APP-4560/code-feature-support-pr-chip)

## Testing 
Verified all tests pass, ran locally and verified pr and repo info is properly fetched/updated, this should be a no-op

## Agent Mode


app/src/code_review/git_status_update.rs
```diff
@@ -372,215 +284,23 @@ impl GitRepoStatusModel {
...
-                }
-            },
-        ));
-    }
-
-    /// Fetch PR info. Called by the periodic timer and after `gh`/`gt`
-    /// commands. Missing/auth setup failures suppress the default chip, but
-    /// polling continues so the chip can recover after the user's `gh` setup
-    /// changes.
-    pub(crate) fn refresh_pr_info(&mut self, ctx: &mut ModelContext<Self>) {
```
MaggieShan, line 426: note for reviewer: most of these deleted structs/fns have been copy-pasted/moved almost 1:1 to `github_repo_model.rs`

app/src/code_review/git_status_update.rs
```diff
@@ -49,14 +36,15 @@ pub struct GitStatusMetadata {
...
-/// [`GitRepoStatusModel`] instances.
+/// [`GitRepoStatusModel`] and [`GitHubRepoModel`] instances.
 ///
 /// Multiple terminals in the same repo share a single sub-model.  When the last
-/// strong handle to a sub-model is dropped, the watcher is torn down
-/// automatically.
+/// strong handle to a sub-model is dropped, the models are torn down automatically.
 pub struct GitStatusUpdateModel {
     #[cfg(feature = "local_fs")]
     repos: HashMap<PathBuf, WeakModelHandle<GitRepoStatusModel>>,
```
kevinchevalier: Mild preference to rename this field.

app/src/code_review/git_status_update.rs
```diff
@@ -120,6 +106,41 @@ impl GitStatusUpdateModel {
...
+    /// When all handles are dropped, the model and its in-flight fetches are torn down.
+    pub fn subscribe_github_repo(
+        &mut self,
+        repo_path: &Path,
+        ctx: &mut ModelContext<Self>,
+    ) -> anyhow::Result<ModelHandle<GitHubRepoModel>> {
+        let repo_path_buf = repo_path.to_path_buf();
+
+        // Check the cache for an existing live model.
+        if let Some(weak) = self.github_repo_models.get(&repo_path_buf) {
```
kevinchevalier, line 130: Oh, this is kind of unfortunate. The git stats pull is fairly expensive, so many enterprise users cannot use it. This dependency on it means they can also not use the PR chip.

That's probably okay, but if there is a simple way around it, it might be nice.

app/src/code_review/github_repo_model.rs
```diff
@@ -0,0 +1,445 @@
+#[cfg(feature = "local_fs")]
```
kevinchevalier: This entire file seems to be "local_fs", so it might be worth either conditionally importing this file, or have this file and a wasm only stub file if needed.

kevinchevalier (approved):
Generally looks good. I had a couple of thoughts, but no blockers.

app/src/code_review/git_status_update.rs
MaggieShan: Renamed to `git_repo_status_models`!

app/src/code_review/git_status_update.rs
MaggieShan: Oh that's a really good point - if we were to remove the dependency I believe there would be two main options here:
1. We rely on periodic timer as well as the `BlockCompleted` event to trigger refreshes for any `git`/`gh`/`gt` command 
2. The `GitHubRepoModel` has it's own file watcher to detect git branch changes 

I would lean towards option 1 but it won't be able to cover as many cases as a file watcher option.

MaggieShan: Spoke offline - we're ok with keeping it as is for now

PR #11571, opened 2026-05-22
Re-enable local Claude Code orchestration

kjankov:
## Description
Re-enables the local Claude Code harness for orchestration while keeping local Codex disabled.

This updates the local harness selector so Claude Code remains visible but disabled when the `claude` CLI is missing, with install guidance in the disabled row tooltip. When Claude Code is installed, the selector allows local Claude Code and silently reconciles missing/outdated Warp Claude plugins during launch.

Follow-up review updates:
- Renamed the local setup helpers away from `local_child_harnesses` since they are not child-specific.
- Added docs to the local harness setup state/helpers.
- Moved setup-state tests into a separate test file.
- Removed the separate selectable `Learn more` dropdown row.

## Linked Issue
N/A

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Testing
- [x] `cargo fmt`
- [x] `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings`
- [x] `cargo test -p warp --lib local_harness_setup`
- [x] `cargo test -p warp --lib orchestration_controls::tests`
- [x] `cargo test -p warp --lib run_agents_card_view::tests`
- [x] `cargo test -p warp --lib local_harness_launch::tests`
- [x] `cargo test -p warp --lib start_agent::tests`
- [x] Earlier targeted tests for `plugin_manager::claude::tests` and `block::tests`

- [ ] I have manually tested my changes locally with `./script/run`

### Screenshots / Videos
N/A

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

## Agent Artifacts
- Conversation: https://app.warp.dev/conversation/d85346e0-18bc-46bc-b743-78414f789f81
- Plan: https://app.warp.dev/drive/notebook/k8cNEcyP1BrVD1mbpgqlGc

CHANGELOG-OZ: Re-enabled local Claude Code child agents for orchestration when Claude Code is installed.

Co-Authored-By: Oz <oz-agent@warp.dev>

PR #12458, opened 2026-06-10
Fix issue when installing oz-harness-support on the update path

abhishek-pandya:
## Description

Fixes REMOTE-1906. The update path got regressed in https://github.com/warpdotdev/warp/pull/11571 and we would no longer clone the internal version of the repo that contains the `oz-harness-support` plugin.

## Testing

Tested locally with the self-hosted direct backend path (i.e. `oz-local`). It did not work before but now does.

## Agent Mode
- [X] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

Helped with logs!

### DIFF
--- a/app/src/terminal/cli_agent_sessions/plugin_manager/claude.rs
+++ b/app/src/terminal/cli_agent_sessions/plugin_manager/claude.rs
@@ -52,26 +52,6 @@ impl ClaudeCodePluginManager {
             .map(|path| HashMap::from([("PATH".to_owned(), path.to_owned())]));
         run_cli_command_logged("claude", args, &self.executor, env_vars, log).await
     }
-
-    async fn attempt_to_install_oz_harness_support_plugin(
-        &self,
-        log: &mut String,
-    ) -> Result<(), PluginInstallError> {
-        if self
-            .run_logged(&["plugin", "install", PLATFORM_PLUGIN_KEY], log)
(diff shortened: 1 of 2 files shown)


app/src/terminal/cli_agent_sessions/plugin_manager/claude.rs
```diff
@@ -52,6 +52,26 @@ impl ClaudeCodePluginManager {
...
+    ) -> Result<(), PluginInstallError> {
+        if self
+            .run_logged(&["plugin", "install", PLATFORM_PLUGIN_KEY], log)
+            .await
+            .is_err()
+        {
+            self.run_logged(
+                &["plugin", "marketplace", "add", PLATFORM_MARKETPLACE_REPO],
+                log,
+            )
```
liliwilson: Should we just unconditionally run `plugin marketplace add PLATFORM_MARKETPLACE_REPO` before trying to install the plugin? Feels clearer and matches precedent in `install`  for the normal plugin

liliwilson (approved):

app/src/terminal/cli_agent_sessions/plugin_manager/claude.rs
abhishek-pandya: chatted offline and we made some tweaks

PR #12462, opened 2026-06-10
Add ipynb formatting buffer

evelyn-with-warp:
## Description

Breakdown https://github.com/warpdotdev/warp/pull/12242 into
1. Ipynb parser and integrate that into Buffer::from_ipynb to bootstrap a buffer from ipynb contents
2. Wire existing app-level logic to support opening an .ipynb file in our file notebook view
This PR is part 1

### Comparison with #12242
This PR (#12462) is a strict subset of #12242. Breakdown of which changes each PR covers:

| Change (high-level) | #12242 | #12462 |
|---|:---:|:---:|
| New `ipynb_parser` crate: convert nbformat v4 notebooks → `FormattedText` (markdown/code/raw cells, saved outputs, images, ANSI stripping, size/length truncation) | ✅ | ✅ |
| Register `ipynb_parser` as workspace member + `warp_editor` dependency (`Cargo.toml`, `Cargo.lock`, `crates/editor/Cargo.toml`) | ✅ | ✅ |
| Editor buffer integration: `ContentFormat::Ipynb`, `InitialBufferState::ipynb`, `Buffer::from_ipynb` (raw-JSON fallback on parse failure), `replace()` match arm | ✅ | ✅ |
| Editor model API: `RichTextEditorModel::reset_with_ipynb` to bootstrap a buffer from notebook contents | ✅ | ✅ |
| `.ipynb` file detection helper `is_jupyter_notebook_file` + tests (`crates/warp_util/src/file_type.rs`) | ✅ | ✅ |
| New `JupyterNotebookRendering` feature flag + dogfood enablement (`crates/warp_features`, `app/src/features.rs`, `app/Cargo.toml`) | ✅ | ❌ |
| Treat `.ipynb` as an openable/notebook file type + tests (`app/src/util/openable_file_type.rs`) | ✅ | ❌ |
| File-tree / code-view wiring to open `.ipynb` in the notebook view (`app/src/code/file_tree/view.rs`, `app/src/code/view.rs`) | ✅ | ❌ |
| Notebook file model/view support for `.ipynb` (`app/src/notebooks/file/mod.rs` + tests, `app/src/notebooks/editor/model.rs`, `editor/view.rs`) | ✅ | ❌ |
| "Open in Warp" + URI routing for `.ipynb` (`app/src/terminal/view.rs`, `terminal/view/open_in_warp.rs`, `app/src/uri/mod.rs` + tests) | ✅ | ❌ |
| Tooltip + workspace view updates for opening notebooks (`app/src/util/tooltips.rs`, `app/src/workspace/view.rs`) | ✅ | ❌ |

## Linked Issue

## Testing

No user perceived changes in this PR
### Screenshots / Videos

https://www.loom.com/share/f6be00f6ab824bb094d74685050675e4
since it changed markdown image rendering by limiting image size and handling base64 string, we tested loading md file in warpdev and warplocal: markdown with images (smaller than 16MB) could be successfully opened/rendered in both (before & after this change)
## Agent Mode


crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+        push_text_output(lines, "[output image omitted: exceeds size limit]");
+        return;
+    }
+    if BASE64_STANDARD.decode(base64.as_bytes()).is_err() {
+        push_text_output(lines, "[output image omitted: invalid base64 data]");
+        return;
+    }
+    lines.push(FormattedTextLine::Image(FormattedImage {
+        alt_text: "output".to_string(),
+        source: format!("data:{mime};base64,{base64}"),
```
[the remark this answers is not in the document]
evelyn-with-warp: Addressed in `add5a9fce3`: added `asset_cache::data_uri_source`, which decodes base64 `data:` URIs into an `AssetSource::Async` (mirroring `url_source` / the mermaid asset pattern), and wired it into `resolve_asset_source_relative_to_directory` (both non-WASM and WASM). Notebook `data:` image outputs now load instead of being treated as a local file path.

crates/asset_cache/src/lib.rs
```diff
@@ -38,6 +46,48 @@ pub fn url_source(url: impl Into<String>) -> AssetSource {
     }
 }
 
+/// Creates an [`AssetSource::Async`] that decodes an inline base64 `data:` URI
+/// (e.g. `data:image/png;base64,<payload>`) into its raw bytes.
+pub fn data_uri_source(source: &str) -> Option<AssetSource> {
```
kevinyang372, line 67: Why are we making changes in our asset cache code? I don't quite understand the motivation here

crates/editor/src/content/edit.rs
```diff
@@ -76,7 +77,9 @@ pub fn resolve_asset_source_relative_to_directory(
     source: &str,
     base_directory: Option<&Path>,
 ) -> AssetSource {
-    if source.starts_with("http://") || source.starts_with("https://") {
+    if let Some(data_uri_source) = asset_cache::data_uri_source(source) {
```
kevinyang372, line 80: This change on supporting inline data URL in editor seems orthogonal to the overall PR on ipynb format parsing?

crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+    if notebook.nbformat != Some(SUPPORTED_NBFORMAT) {
+        return Err(IpynbError::UnsupportedFormat {
+            nbformat: notebook.nbformat,
+        });
+    }
+
+    let language = notebook.language();
+    let mut lines: Vec<FormattedTextLine> = Vec::new();
+
+    for cell in &notebook.cells {
```
kevinyang372: Iterating over notebook cells, converting them into our format text, and then pushing it into a Vec seems not optimal for performance due to the incremental allocation we need to do as we iterate

Could we use std::iter::Map here to do the conversion without incrementally allocating space?

crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    Ok(FormattedText::new_trimmed(lines))
+}
+
+/// Build a verbatim [`FormattedText`] fallback for content that is not a
+/// parseable notebook (malformed JSON, unsupported nbformat version, etc.). The
+/// raw content is placed in a single code block so any Markdown/HTML inside it
+/// is shown verbatim, never re-interpreted.
+pub fn raw_fallback_formatted_text(content: &str) -> FormattedText {
```
kevinyang372, line 83: Shouldn't we return an error and render that in a raw code editor (potentially with a user visible UX)?

I don't think silently falling back to a giant block of JSON is the right approach

crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+    if base64.len() > MAX_IMAGE_DATA_CHARS {
+        push_text_output(lines, "[output image omitted: exceeds size limit]");
+        return;
+    }
+    if BASE64_STANDARD.decode(base64.as_bytes()).is_err() {
+        push_text_output(lines, "[output image omitted: invalid base64 data]");
+        return;
+    }
+    lines.push(FormattedTextLine::Image(FormattedImage {
+        alt_text: "output".to_string(),
```
kevinyang372: Why are we hard-coding alt text here?

crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+fn push_image(lines: &mut Vec<FormattedTextLine>, mime: &str, value: &serde_json::Value) {
+    let base64: String = value_to_text(value)
+        .chars()
+        .filter(|c| !c.is_whitespace())
+        .collect();
+    if base64.is_empty() {
+        return;
+    }
+    if base64.len() > MAX_IMAGE_DATA_CHARS {
+        push_text_output(lines, "[output image omitted: exceeds size limit]");
```
kevinyang372: Is this the same default behavior we do for markdown images?

crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+
+/// Append a text output as a plain (unhighlighted) code block, truncating
+/// oversized output. Empty output is skipped.
+fn push_text_output(lines: &mut Vec<FormattedTextLine>, text: &str) {
+    let stripped = strip_ansi(text);
+    let text = stripped.trim_end_matches('\n');
+    if text.is_empty() {
+        return;
+    }
+    if text.chars().count() > MAX_TEXT_OUTPUT_CHARS {
```
kevinyang372: Why do we have some arbitrary text length truncation here?

crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+fn truncate_chars(s: &str, max_chars: usize) -> &str {
+    match s.char_indices().nth(max_chars) {
+        Some((idx, _)) => &s[..idx],
+        None => s,
+    }
+}
+
+/// Strip ANSI escape sequences (CSI/SGR colors, OSC, and simple escapes) so
+/// tracebacks render as readable plain text.
+fn strip_ansi(input: &str) -> String {
```
kevinyang372, line 228: Why do we need to strip ansi?

crates/ipynb_parser/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+        serde_json::Value::Array(items) => items.iter().filter_map(|v| v.as_str()).collect(),
+        _ => String::new(),
+    }
+}
+
+/// Sanitize a notebook-declared language into a safe code-block language tag.
+fn sanitize_language(raw: &str) -> String {
+    let trimmed = raw.trim();
+    let is_safe = !trimmed.is_empty()
+        && trimmed.chars().count() <= MAX_LANGUAGE_TAG_CHARS
```
kevinyang372, line 288: Don't quite understand why we need this sanitization process as well. Also seems like the fallback is we just leave an empty string?

kevinyang372 (changes requested):

crates/asset_cache/src/lib.rs
evelyn-with-warp: The motivation comes from [Oz comment](https://github.com/warpdotdev/warp/pull/12462#pullrequestreview-4478170831), as jupyter notebook supports embedding images as base64 inline, which is different from markdown image rendering as an external reference

crates/editor/src/content/edit.rs
evelyn-with-warp: it is ipynb parser [push_image](https://github.com/warpdotdev/warp/blob/2b542e46d0591d1e6b240f306787dbab8d2a62ac/crates/ipynb_parser/src/lib.rs#L204-L205) that would produce inline data, and therefore we need to support this. 
Please let me know if i misunderstood

crates/ipynb_parser/src/lib.rs
evelyn-with-warp: this is a place holder: if a cell generates image (calling plot libraries like seaborn, matplotlib..etc), it would carry no alt-text metadata and display the image in cell output, like
```
   "outputs": [
    {
     "data": {
      "image/png": some_base64
      "text/plain": [
       "<Figure size 1427.25x450 with 2 Axes>"
      ]
     },
     "metadata": {},
     "output_type": "display_data"
    }
   ```

Hence just picked the placeholder name semantically, as FormattedImage.alt_text is non_optional. 
Hard to decide if put an empty string here is better.

File: crates/ipynb_parser/src/lib.rs (hunk 8 above)
Comment (evelyn-with-warp) [reply crates/ipynb_parser/src/lib.rs]: this truncation only applies for cell output, not cell (code or markdown), to prevent cell output bloating the buffer; 
For vscode,  we will see something like this
image
Added a TODO here, if we also want to support a text truncation UI like vscode (for now i feel it's a little overhead on the scope of rendering a notebook)

File: crates/ipynb_parser/src/lib.rs (hunk 7 above)
Comment (evelyn-with-warp) [reply crates/ipynb_parser/src/lib.rs]: Great call. This is only for jupyter notebook UI optimization (more strict than markdown, but maybe not necessary). i dropped it to be consistent with markdown too. https://github.com/warpdotdev/warp/pull/12462/changes/24a89077cfb07ee25633c11873e46483875485f4
We already have [a size guard in asset cache](https://github.com/warpdotdev/warp/blob/2b542e46d0591d1e6b240f306787dbab8d2a62ac/crates/asset_cache/src/lib.rs#L22-L23) that applied to both jupyter notebook and markdown.

File: crates/ipynb_parser/src/lib.rs (hunk 9 above)
Comment (evelyn-with-warp) [reply crates/ipynb_parser/src/lib.rs:228]: called out by this [Oz-comment](https://github.com/warpdotdev/warp/pull/12462#pullrequestreview-4470471349)
The push_text_output will be rendered as plan, unhighlighted code block instead of a terminal/ansi-aware surface. `strip_ansi` in all input paths (stream, text/plain, and error) of push_text_output in order to avoid garbage rendering

File: crates/ipynb_parser/src/lib.rs (hunk 10 above)
Comment (evelyn-with-warp) [reply crates/ipynb_parser/src/lib.rs:288]: The language refers to the the notebook metadata `metadata.language_info.name`, which is attacker-controllable.
it's also a security guard called out by [Oz](https://github.com/warpdotdev/warp/pull/12242#discussion_r3360482120)

Leave an empty string means the code block will render as a plain, unhighlighted code block. Therefore, when detects the language is unknown/not safe
1. [push_code_block](https://github.com/warpdotdev/warp/blob/24a89077cfb07ee25633c11873e46483875485f4/crates/ipynb_parser/src/lib.rs#L76-L77) takes an empty string as param
2. Flows to [CodeBlockType](https://github.com/warpdotdev/warp/blob/24a89077cfb07ee25633c11873e46483875485f4/crates/editor/src/content/text.rs#L714), where lang is used to pick a grammar highlight

File: crates/ipynb_parser/src/lib.rs (hunk 7 above)
Comment (evelyn-with-warp) [reply crates/ipynb_parser/src/lib.rs]: As a follow up: silent drop maybe not desirable; therefore attempted to display a hint when images are too big to display https://github.com/warpdotdev/warp/pull/12462/changes/a2c7defbedbbf4865c0da69291a2504573857233

File: crates/ipynb_parser/src/lib.rs (hunk 4 above)
Comment (evelyn-with-warp) [reply crates/ipynb_parser/src/lib.rs]: https://github.com/warpdotdev/warp/pull/12462/changes/a8b342481e9647accee9aae04a4a7b7eae0cb87f
is this something in your mind? The only way I found with agent to avoid incremental allocation was applying counting first for a size hint, but i don't have strong feeling on which is more efficient on the bottom

File: crates/ipynb_parser/src/lib.rs (hunk 5 above)
Comment (evelyn-with-warp) [reply crates/ipynb_parser/src/lib.rs:83]: i agree on this. VScode will display a ui of invalid notebook and routed to a text editor instead. 
I attempted this change https://github.com/warpdotdev/warp/pull/12462/changes/f9e49b659d108ef2d516da77b8d41b192eeabeaa, but later reverted it as i could not test due to missing the wiring logic. Let's put the change in the following PR of wiring up to app logic and test e2e

File: crates/asset_cache/src/lib.rs (hunk 2 above)
Comment (kevinyang372) [reply crates/asset_cache/src/lib.rs:67]: I see

File: crates/ipynb_parser/src/lib.rs (hunk 5 above)
Comment (kevinyang372) [reply crates/ipynb_parser/src/lib.rs:83]: Can we change the `from_ipynb` API so it returns Result<Buffer> instead of Buffer? I would rather get the APIs in a good spot in this PR

File: crates/ipynb_parser/src/lib.rs (hunk 8 above)
Comment (kevinyang372) [reply crates/ipynb_parser/src/lib.rs]: I think we need to think through the truncation experience more. Pushing hard coded text here means if user selects all on the notebook and hit copy, we are going to record `[output truncated]` rather than the canonical text

I would rather us punting on these arbitrary truncations and think about the right modeling here

File: crates/ipynb_parser/src/lib.rs [hunk 11]
```diff
@@ -0,0 +1,411 @@
...
+        }
+        // Unknown output types are skipped.
+        _ => {}
+    }
+}
+
+/// Append a text output as a plain (unhighlighted) code block, truncating
+/// oversized output. Empty output is skipped.
+fn push_text_output(lines: &mut Vec<FormattedTextLine>, text: &str) {
+    let stripped = strip_ansi(text);
```
kevinyang372, line 198: Instead of stripping ansi -- should we have a color attribute on FormattedTextStyles?

crates/ipynb_parser/src/lib.rs
kevinyang372: ditto -- see my comment above

crates/ipynb_parser/src/lib.rs
evelyn-with-warp: Did https://github.com/warpdotdev/warp/pull/12462/changes/213f2151ada693b947c8ffa0221d7c8d00e08600 address your thought?

crates/ipynb_parser/src/lib.rs
evelyn-with-warp: which comment? was it about injecting placeholder text that violate copy canonical content? it's already been removed here https://github.com/warpdotdev/warp/pull/12462/changes/24a89077cfb07ee25633c11873e46483875485f4
now the size control shares the same with markdown in `crates/asset_cache/src/lib.rs` which will return None

crates/ipynb_parser/src/lib.rs
evelyn-with-warp: removed truncation in https://github.com/warpdotdev/warp/pull/12462/changes/213f2151ada693b947c8ffa0221d7c8d00e08600 to unblock, and revisit when truncation UI is supported.
On rendered notebook, i'd expect it's rare to have use case of selecting all the content or all the coding blocks (aka cells); This concern is mostly for selecting all the output of a cell. For this, IMHO the user story will likely to be searching for other sources to copy from, when displayed truncation hint. Also, since warp will not intend to support jupyter notebook kernal (execution), the purpose of this is rendering the content in a neat & clean way for a quick proof reading. Therefore , i'd trade the copy/paste convenience for a nicer display (just my personal preference)

crates/ipynb_parser/src/lib.rs
evelyn-with-warp: That would be desirable, and the color attribute will be applied cross-crates, not just jupyter notebook, right?Therefore shall we put it in another PR (scope); for now let remove ansi to protect notebook rendering. Added a TODO item here.

crates/editor/src/content/buffer.rs
```diff
@@ -912,6 +947,28 @@ impl Buffer {
                 selection_model.clone(),
                 ctx,
             ),
+            ContentFormat::Ipynb => match Buffer::from_ipynb(
```
kevinyang372, line 950: Instead of doing the fallback formatted text, we can just ignore updating *self in this case right? (it's initialized to be empty anyway)

This allows us to get rid of `raw_fallback_formatted_text` entirely

kevinyang372 (approved):

PR #12465, opened 2026-06-10
Add setting to enable reusing user's existing control master

kevinyang372:
### Problem
Some users manage their own SSH `ControlMaster` for a host: their ssh config sets `ControlMaster auto` + `ControlPath`, and that master owns shared state — in the reported case, a reverse-forwarded (`RemoteForward`) socket used by all of their SSH clients. Warp's SSH wrapper always forces `-o ControlMaster=yes -o ControlPath=~/.ssh/<session_id>` on warpified sessions, so connecting from Warp creates a **second, competing master** for the same host. Only one connection can bind the reverse forward, so the two masters race over it and the user's tooling becomes unreliable (`Warning: remote port forwarding failed ...`).

### Fix
Behind a new opt-in setting (`warpify.ssh.reuse_existing_control_master`, default **off**; toggleable from the Warpify settings page and the Command Palette), the SSH wrapper now looks for a live master the user already runs for the destination before creating its own:

1. `ssh -G "$@"` resolves the user's effective `ControlPath`, with tokens (`%h`/`%p`/`%r`/`%C`) expanded into a literal path.
2. `ssh -O check` verifies a master is actually alive on that socket. Both probes are local-only commands.
3. If alive, the foreground ssh runs with `ControlMaster=no` against the user's socket — Warp becomes just another client of the user's master, so there is no second master and no forward conflict. All multiplexed traffic (remote-server proxy, scp, generators) flows through the same socket unchanged.
4. On any miss (no `ControlPath` configured, dead master, path characters we can't safely embed, older ssh), the wrapper falls back to today's behavior exactly.

Master ownership is threaded end-to-end so teardown is safe: the SSH DCS hook reports a new `external_master` field, plumbed through `IsLegacySSHSession` → `SshTransport` → a new `ControlPath` enum (`WarpManaged` / `UserOwned` / `None`) on the remote-server session state. Session teardown only runs `ssh -O exit` against `WarpManaged` masters — Warp never tears down a master the user owns.

### Alternatives considered
- **Stop binding forwards from Warp's master** (e.g. force `ClearAllForwardings=yes` in the wrapper): prevents this specific conflict but still leaves two masters per host, doesn't address non-forward conflicts, and silently drops forwards users *do* want in Warp sessions.
- **Manual "attach to this socket" setting** (user supplies a ControlPath per host): avoids discovery entirely, but requires per-host configuration and can't express tokenized ControlPaths (`%C`, `%r@%h:%p`), which are the recommended and most common setup.
- **Destination-aware transport refactor**: thread the user's real ssh destination through the client so secondary commands name the host instead of `placeholder@placeholder`. Most general, but a much larger change — and unnecessary, since `ssh -G` already yields a literal expanded socket path that the existing socket-only transport consumes as-is.
- **No product change** (user-side workaround): users can move `RemoteForward` to a dedicated master started outside Warp. Works, but pushes config surgery onto every affected user and leaves the surprising two-master behavior in place.

### Rollout
Default **off** for safety: reusing a master couples Warp's session lifetime to the user's master and shares its `MaxSessions` channel budget, which existing multiplexing users may not expect. Once validated more broadly we can flip the default and keep the setting as an escape hatch.

## Linked Issue
No linked issue — this addresses a customer support request about the SSH wrapper's forced ControlMaster conflicting with user-managed masters.

## Testing

**Automated**
- Unit tests for SSH DCS hook parsing of the new `external_master` field, including the default (`false`) for hooks emitted by older bootstrap scripts.
- Updated transport/session tests for the new `SshTransport` ownership parameter; `cargo nextest` passes for the affected crates; `./script/format` + clippy clean.
- Syntax-checked all three modified bootstrap scripts (`bash -n`, `zsh -n`, `fish --no-execute`).

**Manual E2E** (`./script/run` against a remote devbox configured like the customer: `ControlMaster auto`, tokenized `ControlPath`, `RemoteForward`):
- With a user-started master running: Warp attached to it — no new Warp socket in `~/.ssh`, no port-forward failure, session fully warpified, remote-server features work over the reused socket.
- After exiting the Warp session: `ssh -O check` confirms the user's master is **still running** (Warp did not tear it down) and the reverse forward stays bound.
- Fallback: with no live master, the wrapper creates a Warp-owned master exactly as before and cleans it up on exit.
- Regression: a host with no `ControlPath` configured behaves identically to today.

## Agent Mode

[Conversation](https://staging.warp.dev/conversation/665a4e24-372b-4e95-8c19-601228ec77eb) | [Options plan](https://staging.warp.dev/drive/notebook/CCVs76IlBdsQNB5PBsaePx)

CHANGELOG-IMPROVEMENT: Added an opt-in setting (`warpify.ssh.reuse_existing_control_master`) that makes Warp's SSH wrapper attach to an existing SSH ControlMaster for the destination host instead of always creating its own, preserving user-managed masters and their port forwards.

### DIFF
--- a/app/assets/bundled/bootstrap/bash_body.sh
+++ b/app/assets/bundled/bootstrap/bash_body.sh
@@ -1008,6 +1008,40 @@ if [ -z "$WARP_BOOTSTRAPPED" ]; then
             # and use printf to convert back to plaintext
             local zsh_env_script=$(printf '%s' 'unsetopt ZLE; unset RCS; unset GLOBAL_RCS; WARP_SESSION_ID='$remote_session_id'; WARP_USING_WINDOWS_CON_PTY=@@USING_CON_PTY_BOOLEAN@@; WARP_HONOR_PS1='$WARP_HONOR_PS1'; _hostname=$(command -pv hostname >/dev/null 2>&1 && command -p hostname 2>/dev/null || command -p uname -n); _user=$(command -pv whoami >/dev/null 2>&1 && command -p whoami 2>/dev/null || echo $USER); _msg=$(printf "{\"hook\": \"InitShell\", \"value\": {\"session_id\": $WARP_SESSION_ID, \"shell\": \"zsh\", \"user\": \"%s\", \"hostname\": \"%s\"}}" "$_user" "$_hostname" | command -p od -An -v -tx1 | command -p tr -d '"'"' \n'"'"'); printf '"'"'\e]9278;d;%s\x07'"'"' $_msg; unset _hostname _user _msg' | command -p od -An -v -tx1 | command -p tr -d ' \n')
 
+            # Optionally attach to an existing ControlMaster the user already
+            # runs for this destination instead of creating our own. Resolve
+            # the user's configured ControlPath with `ssh -G` (which expands
+            # tokens like %h/%p/%r/%C into a literal path), then verify the
+            # master is alive with `ssh -O check`. Both probes are local-only
+            # commands. On any failure we fall back to creating a Warp-owned
+            # master, preserving the existing behavior.
+            local control_path="$SSH_SOCKET_DIR/$WARP_SESSION_ID"
+            local control_master_mode="yes"
+            local external_master="false"
+            if [[ "$WARP_SSH_REUSE_CONTROL_MASTER" == "1" ]]; then
+                local user_control_path=$(command ssh -G "${@:1}" 2>/dev/null | command -p awk '$1 == "controlpath" { sub(/^controlpath[ \t]+/, ""); print; exit }')
+                case "$user_control_path" in
+                    "" | none)
+                        # No ControlPath configured for this destination.
+                        ;;
+                    *[![:alnum:]._/~@:+,-]*)
+                        # The resolved path contains characters we cannot safely
+                        # embed in the SSH hook JSON below (e.g. an unexpanded %
+                        # token, quotes, or whitespace); fall back to a
+                        # Warp-owned master.
+                        ;;
+                    *)
+                        if command ssh -O check -o ControlPath="$user_control_path" "${@:1}" >/dev/null 2>&1; then
+                            # A live master exists: multiplex through it and let
+                            # the client know Warp does not own it.
+                            control_path="$user_control_path"
+                            control_master_mode="no"
+                            external_master="true"
+                        fi
+                        ;;
+                esac
+            fi
+
             # Keep remote commands up-to-date with shell.rs & bash.sh.
             # Note that in this command, we're passing a string to the remote shell. Any variable expansions need to be
             # escaped with "''" to avoid the local shell from expanding them before they're passed to the remote shell.
@@ -1015,7 +1049,7 @@ if [ -z "$WARP_BOOTSTRAPPED" ]; then
             # determine what shell is the login shell on the remote machine.  We perform a preliminary check to see if
             # the remote shell is the Bourne shell to avoid asking it to parse later lines that use syntax it doesn't
             # support.
-            command ssh -o ControlMaster=yes -o ControlPath=$SSH_SOCKET_DIR/$WARP_SESSION_ID \
+            command ssh -o ControlMaster=$control_master_mode -o ControlPath="$control_path" \
             -t "${@:1}" \
 "
 export TERM_PROGRAM='WarpTerminal'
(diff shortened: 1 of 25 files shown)


kevinyang372:
* **#12465** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12465?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12465?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/remote_server/ssh_transport.rs
```diff
@@ -251,7 +267,15 @@ impl RemoteTransport for SshTransport {
                 failure_rx,
                 host_response_rx,
                 child,
-                control_path: Some(socket_path),
+                // Tag the socket with master ownership. Teardown only runs
+                // `ssh -O exit` against Warp-managed masters; a user-owned
+                // (external) master must be left running when the Warp
+                // session exits.
+                control_path: if owns_master {
```
alokedesai: would name this variable `warp_managed_control_socket` or something a little bit more explicit

app/src/settings_view/warpify_page.rs
```diff
@@ -743,6 +767,64 @@ impl SettingsWidget for SSHWidget {
             );
         }
 
+        // Reuse-ControlMaster only applies to the legacy (ControlMaster)
+        // SSH wrapper, which is bypassed while the tmux wrapper is in use.
+        if !should_prompt_ssh_tmux_wrapper {
```
alokedesai, line 772: Is this setting meant to be permanent or just a temporary state before we enable it for everyone? If the latter, i'd advocate for a simpler approach where we only make the setting available in the file and don't expose anything in the UI

app/src/terminal/model/session.rs
```diff
@@ -532,7 +532,14 @@ impl From<&SessionType> for command_corrections::SessionType {
...
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum IsLegacySSHSession {
-    Yes { socket_path: PathBuf },
+    Yes {
+        socket_path: PathBuf,
+        /// `true` when `socket_path` points at a ControlMaster the user
+        /// already had running (the SSH wrapper attached to it instead of
+        /// creating a Warp-owned one). Warp must not tear down such a
+        /// master on session exit.
+        external_master: bool,
```
alokedesai: simlar to the above comment. This should be something like `external_control_master` or something like that

crates/remote_server/src/ssh.rs
alokedesai: thoughts on making this part of the `ControlPath` enum (where it early returns if it's a user owned socket)

app/assets/bundled/bootstrap/bash_body.sh
```diff
@@ -1008,14 +1008,48 @@ if [ -z "$WARP_BOOTSTRAPPED" ]; then
...
+            # the user's configured ControlPath with `ssh -G` (which expands
+            # tokens like %h/%p/%r/%C into a literal path), then verify the
+            # master is alive with `ssh -O check`. Both probes are local-only
+            # commands. On any failure we fall back to creating a Warp-owned
+            # master, preserving the existing behavior.
+            local control_path="$SSH_SOCKET_DIR/$WARP_SESSION_ID"
+            local control_master_mode="yes"
+            local external_master="false"
+            if [[ "$WARP_SSH_REUSE_CONTROL_MASTER" == "1" ]]; then
+                local user_control_path=$(command ssh -G "${@:1}" 2>/dev/null | command -p awk '$1 == "controlpath" { sub(/^controlpath[ \t]+/, ""); print; exit }')
```
alokedesai: Confirming you tested this across all shells?

alokedesai: consider just using `sed` here instead of `awk` to limit the number of external commands we depend on in the bootstrap script. Would also make the code slightly easier to read.

`local user_control_path=$(command ssh -G "${@:1}" 2>/dev/null | command -p sed -n 's/^controlpath //p')`

alokedesai (approved):

app/src/settings_view/warpify_page.rs
kevinyang372: I think we should have a permanent setting in case user actually want Warp to always create a master connection. There are actual use cases for this imo (e.g. user has limited number of control paths per master connection)

PR #12466, opened 2026-06-10
Add instrumentation to 3p harness plugin installation and updates

seemeroland:
## Description

We had one on `ThirdPartyHarnessPreparation`, but not for the specific things we might do as part of that. It'd be good to know which plugin and on install or update is failing.


abhishek-pandya (approved):
Nice! This will be very helpful to have and can hopefully reduce our dependency on CLI logs when something goes wrong

PR #12470, opened 2026-06-10
Make MCP server startup non-fatal for Oz agent runs

vkodithala:
## Description

Makes MCP server startup non-fatal by default for Oz agent runs.

Previously, cloud agent runs failed outright with "One or more MCP servers failed to start..." when any requested MCP server failed to reach `Running` within the 20s startup timeout. Several Oz users hit issues where some subset of their servers wouldn't spawn within the (non-configurable) timeout and thus the entire run was borked.

The shape of this solution:

- The MCP startup waiters in `AgentDriver` now wait for _every_ requested server to reach a terminal state (`Running` or `FailedToStart`) instead of failing fast on the first failure, and track per-server names so degradation is reportable (e.g. `failed to start: server_x; still starting after 20s: server_y`).
- `AgentDriverError::MCPStartupFailed` carries those details, which surface in both the strict-mode failure message and the non-strict warning.
- In `run_internal`, both the spec'd (`--mcp`) and profile MCP startup results are handled outside `record_result` (preserving `setup_mcp_server_startup` telemetry): non-strict runs log a WARN naming the unavailable servers and post a best-effort message-only `update_agent_task` status update, then continue. `MCPServerNotFound` (bad UUID config) stays fatal.
- Two new CLI flags on `oz agent run`: `--strict-mcp-startup` (restores the old fail-fast behavior) and `--mcp-startup-timeout <DURATION>` (default 20s), plumbed through `AgentDriverOptions`.

Servers that connect after the timeout still become usable mid-run: in-flight spawns are never aborted, and the per-request `mcp_context` is rebuilt from active servers each turn.

## Follow-ups

This PR sets up client-side scaffolding that lets us configure how strict we are about spawning MCP servers prior to cloud agent runs, and the duration we wait for those servers to spawn before propagating the first user query. The intended follow-up here is to surface these as configurable options via the Oz web app, which should be a fairly trivial change.

## Testing

Manually tested via local Oz CLI; here's an example of a publicly reachable MCP server (`deepwiki`) and an _unreachable_ server configured via a URL that explicitly accepts no connections. Notice that the conversation proceeds despite the failure.

`cargo nextest run -p warp -E 'test(error_classification) or test(driver_tests)'` — 32/32 pass, including a new test asserting `MCPStartupFailed` details surface in the classified message.

- `cargo nextest run -p warp_cli` — 175/175 pass (clap arg definitions).
- `./script/format` and presubmit's `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings` pass.

## Agent Mode

CHANGELOG-OZ: By default, cloud agent runs now continue without configured MCP servers that fail to start within the 20s default startup window, instead of failing the run.

Plan: https://staging.warp.dev/drive/notebook/ASfpNkdsDIpw6hkTsmooCo  
Conversation: https://staging.warp.dev/conversation/44d42331-9340-4dbe-9e87-276117c147cb

### DIFF
--- a/app/src/ai/agent_sdk/driver/error_classification.rs
+++ b/app/src/ai/agent_sdk/driver/error_classification.rs
@@ -100,15 +100,22 @@ pub fn classify_driver_error(error: &AgentDriverError) -> (AgentTaskState, TaskS
                 PlatformErrorCode::EnvironmentSetupFailed,
             ),
         ),
-        AgentDriverError::MCPStartupFailed { details } => (
-            AgentTaskState::Failed,
-            TaskStatusUpdate::with_error_code(
-                format!(
-                    "One or more MCP servers failed to start ({details}). Check that your MCP server configuration is valid and the server process is runnable."
+        AgentDriverError::MCPStartupFailed { details } => {
+            let server_lines = details
+                .iter()
+                .map(|detail| format!("- {detail}"))
+                .collect::<Vec<_>>()
+                .join("\n");
+            (
+                AgentTaskState::Failed,
+                TaskStatusUpdate::with_error_code(
+                    format!(
+                        "One or more MCP servers failed to start:\n\n{server_lines}\n\nCheck that each server's configuration is valid and that it is reachable from the agent's environment."
+                    ),
+                    PlatformErrorCode::EnvironmentSetupFailed,
                 ),
-                PlatformErrorCode::EnvironmentSetupFailed,
-            ),
-        ),
+            )
+        }
         AgentDriverError::MCPJsonParseError(msg) => (
             AgentTaskState::Failed,
             TaskStatusUpdate::with_error_code(
--- a/app/src/ai/agent_sdk/driver/error_classification.rs
+++ b/app/src/ai/agent_sdk/driver/error_classification.rs
@@ -100,10 +100,12 @@ pub fn classify_driver_error(error: &AgentDriverError) -> (AgentTaskState, TaskS
                 PlatformErrorCode::EnvironmentSetupFailed,
             ),
         ),
-        AgentDriverError::MCPStartupFailed => (
+        AgentDriverError::MCPStartupFailed { details } => (
             AgentTaskState::Failed,
             TaskStatusUpdate::with_error_code(
-                "One or more MCP servers failed to start. Check that your MCP server configuration is valid and the server process is runnable.",
+                format!(
+                    "One or more MCP servers failed to start ({details}). Check that your MCP server configuration is valid and the server process is runnable."
+                ),
                 PlatformErrorCode::EnvironmentSetupFailed,
             ),
         ),
--- a/crates/warp_cli/src/agent.rs
+++ b/crates/warp_cli/src/agent.rs
@@ -312,6 +312,15 @@ pub struct RunAgentArgs {
     /// LEGACY: MCP servers to start before executing the agent, identified by UUID.
     #[arg(long = "mcp-server", value_name = "UUID", hide = true)]
     pub mcp_servers: Vec<uuid::Uuid>,
+    /// Fail the run when any requested MCP server fails to start.
+    ///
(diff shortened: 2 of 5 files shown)


app/src/ai/agent_sdk/driver/error_classification.rs
```diff
@@ -100,10 +100,12 @@ pub fn classify_driver_error(error: &AgentDriverError) -> (AgentTaskState, TaskS
...
             ),
         ),
-        AgentDriverError::MCPStartupFailed => (
+        AgentDriverError::MCPStartupFailed { details } => (
             AgentTaskState::Failed,
             TaskStatusUpdate::with_error_code(
-                "One or more MCP servers failed to start. Check that your MCP server configuration is valid and the server process is runnable.",
+                format!(
+                    "One or more MCP servers failed to start ({details}). Check that your MCP server configuration is valid and the server process is runnable."
+                ),
```
vkodithala: context: The idea here is that we log richer errors on failed server startup in the `--strict-mcp-startup` case so users won't have to pore through logs to determine which MCPs failed.

captainsafia: Super nit but feels like we should have the details displayed after the error message or shorten the error message that we include. Specifically because I think the details have more value than the bit about checking your config.

captainsafia (approved):
Looks good to me as far as the structural changes:  
  
\- Failures will warn and log instead of erroring by default  
\- Erros are collected on a per-MCP server basis instead of all-up and logged individually

peicodes (approved):
This will require the customer to provide the new option to see a difference, correct?

vkodithala:
@peicodes Nah, `--strict-mcp-startup` is a boolean arg, so it's only set to true when provided (default False). Similar case for `--mcp-startup-timeout`; in the CLI driver we set a default value of 20s: https://github.com/warpdotdev/warp/blob/varoon/oz-mcp-failures-non-fatal/app/src/ai/agent_sdk/driver.rs#L694-L695, this is what it was prior to this change. Meaning the default = we don't block agent startup on MCP spawn and wait for 20s for servers to report back their status.

peicodes:
Got it, I had it backwards. Ty for the fix!

crates/warp_cli/src/agent.rs
```diff
@@ -312,6 +312,15 @@ pub struct RunAgentArgs {
...
     pub mcp_servers: Vec<uuid::Uuid>,
+    /// Fail the run when any requested MCP server fails to start.
+    ///
+    /// By default, MCP servers that don't start within the startup timeout are
+    /// skipped and the agent runs without their tools.
+    #[arg(long = "strict-mcp-startup")]
+    pub strict_mcp_startup: bool,
+    /// Maximum time to wait for requested MCP servers to start (e.g. `30s`, `1m`).
+    #[arg(long = "mcp-startup-timeout", value_name = "DURATION")]
+    pub mcp_startup_timeout: Option<humantime::Duration>,
```
vkodithala, line 323: context: `--strict-mcp-startup` is a boolean arg, so defaults to false when unset. Driver log in `driver::driver.rs` makes it so we set startup timeout to 20s when unset, too. 

The implication is that when both flags are unset, we revert to a 20s timeout with the only behavioral change from the pre-PR state being that we log warnings for servers that fail to start up in that window.

app/src/ai/agent_sdk/driver/error_classification.rs
vkodithala: Good feedback, restructured to make this clearer.

vkodithala:
* **#12470** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12470?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12470?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #12471, opened 2026-06-10
Fail fast when platform plugin is not correctly installed.

abhishek-pandya:
## Description

Follow up from https://warpdev.slack.com/archives/C093TUFQQSJ/p1781022276678869.

## Testing

I tested locally that using a bogus repository name for the marketplace causes a cloud mode run to fail fast.

Screenshot 2026-06-10 at 2 53 19 PM


liliwilson (approved):
Thanks Abhi!

PR #12484, opened 2026-06-11
Enhance writing logic for ai_queries sqlite db table 

evelyn-with-warp:
## Description

We don't have write cap for ai_queries (which could be dangerous), but we have [read cap for ai_queries](https://github.com/warpdotdev/warp/blob/917041b78aeed1c3e62e4be463f9c8694ee2c695/app/src/persistence/block_list.rs#L96), and those empty input lines [got filtered out](https://github.com/warpdotdev/warp/blob/917041b78aeed1c3e62e4be463f9c8694ee2c695/app/src/ai/blocklist/history_model.rs#L2015-L2023) before feeding downstream. Therefore, we get much less number of rows than expected. (85% of them are filtered out from local check with below sqlite queries)

This PR aims to fix above experiences: 
- added write cap for ai_queries based on FIFO
- skipped writing to table if input was empty

The goal is without touching user local storage, the ai_queries will be filled by non-empty records by FIFO eviction policy eventually.

## Context
when running 
```
DB="$HOME/Library/Group Containers/2BBY89MBSN.dev.warp/Library/Application Support/dev.warp.Warp-Dev/warp.sqlite"

# Total row count
sqlite3 "$DB" "SELECT COUNT(*) FROM ai_queries;"

# Rows with empty input (treats whitespace-only as empty)
sqlite3 "$DB" "SELECT COUNT(*) FROM ai_queries WHERE TRIM(input) = '[]';"
```
i observed 16745 (out of 18144 total) are empty input rows. Those rows are written in ai_queries table, but they got filtered out upon [reading ](https://github.com/warpdotdev/warp/blob/b34f4ecbdffba385d53d71161514b91aa4c8dad7/app/src/ai/blocklist/history_model.rs#L2015-L2023)(so we never used those rows and this is a waste of resource)
## Linked Issue

## Testing

## Agent Mode


app/src/pane_group/pane/terminal_pane.rs
```diff
@@ -2180,13 +2180,21 @@ fn handle_ai_history_event(
                 return;
             }
 
+            // Only query-bearing inputs (e.g. user queries) are persisted for
+            // up-arrow history. Early return and skip writing when xchanges that carry
```
seemeroland: nit: fix grammar

app/src/persistence/block_list.rs
```diff
@@ -106,15 +106,53 @@ pub(super) fn read_ai_queries(
...
+    limit: i64,
 ) -> anyhow::Result<()> {
     use schema::ai_queries::dsl::*;
 
     let new_ai_query = NewAIQuery::try_from(query.as_ref())?;
 
     Ok(conn.transaction::<_, Error, _>(|conn| {
+        // Only a genuinely new exchange grows the table.
+        let is_new_exchange = ai_queries
+            .filter(exchange_id.eq(&new_ai_query.exchange_id))
```
seemeroland, line 133: Is this indexed? This could be slow if not. Wondering if this is necessary to check

seemeroland (commented):
lgtm, just question about performance of checking if this is a new exchange

app/src/persistence/block_list.rs
evelyn-with-warp: it should be indexed here [crates/persistence/migrations/2024-08-22-182702_ai_exchange_id_unique_index/up.sql](https://github.com/warpdotdev/warp/blob/917041b78aeed1c3e62e4be463f9c8694ee2c695/crates/persistence/migrations/2024-08-22-182702_ai_exchange_id_unique_index/up.sql)

The upsert fires on both AppendedExchange and UpdatedStreamingExchange, so the same exchange_id gets upserted repeatedly while output streams. For UpdatedStreamingExchange, we don't need to run the count check and just update the table

evelyn-with-warp: if the check is cheap, then shall we keep it for performance (avoid unnecessary count & delete); 
Another option is dropping the check while maintain the accuracy : upsert first and then delete if over capacity, but then we will run the count() on every streaming update

seemeroland: I don't think we expect to ever hit a case where we're inserting a duplicate exchange id. I think it'd be fine to assume it's inserting a new row. But since it's cheap I suppose it's okay to keep

seemeroland (approved):

PR #12492, opened 2026-06-11
remove redundant path canonicalizations causing UI to freeze with WSL tabs

### LINKED ISSUE #9920: Tab switching becomes sluggish for WSL tabs opened via UI

acarl005:
## Description

This PR mainly fixes an issue where switching to a WSL tab freezes the app for a second or two. The root cause is `std::fs::canonicalize` (or anything derived from that) being called repeatedly on WSL paths. Canonicalizing a path through WSL, e.g. `\\WSL$\Ubuntu\home\username\somefile` is slow because it crosses the VM boundary. It's not **so** slow that we have to avoid it entirely. However, it's slow enough that calling it 10+ times on the main thread in succession causes noticeable lag. This is actually happening. The lag grows linearly with the number of tabs in the workspace.

To make things worse, some callees are defensively re-canonicalizing even though their callers had already canonical a path, for example:

1. When [this value](https://github.com/warpdotdev/warp/blob/3d8235443c6eadab65159e2cd5e0f27f2204be99/app/src/workspace/view.rs#L16711) is produced it [is already canonical](https://github.com/warpdotdev/warp/blob/ba25a8474cdca53414cbe52eada1cbcdd5fde8a7/app/src/terminal/view.rs#L23201-L23201).
2. That gets [saved on the `ActiveSession` singleton](https://github.com/warpdotdev/warp/blob/3d8235443c6eadab65159e2cd5e0f27f2204be99/app/src/workspace/view.rs#L16729-L16729).
3. That gets [retrieved here](https://github.com/warpdotdev/warp/blob/3d8235443c6eadab65159e2cd5e0f27f2204be99/app/src/search/ai_context_menu/view.rs#L389-L392) and then passed to `DetectedRepositories::get_root_for_path` which in turn [canonicalized again](https://github.com/warpdotdev/warp/blob/3d8235443c6eadab65159e2cd5e0f27f2204be99/crates/repo_metadata/src/repositories.rs#L226).

I propose fixing this with the following:

1. Cache a canonical version of the PWD on the `TerminalView`. Basically I added a analogous method `canonical_session_pwd_if_local`. It wraps the existing `active_session_path_if_local` method but canonicalizes it afterwards and then memoizes that result. I have consumers call this new method instead. The main culprit for this is [`DirectoryTabColors::color_for_directory`](https://github.com/warpdotdev/warp/blob/3d8235443c6eadab65159e2cd5e0f27f2204be99/app/src/workspace/tab_settings.rs#L204-L204)
2. Add a non-canonicalizing version of [`DetectedRepositories::get_root_for_path`](https://github.com/warpdotdev/warp/blob/1e43a0a0bc610b63edce624a01d1cbae8835f374/crates/repo_metadata/src/repositories.rs#L228-L228) which I call [`get_root_for_canonical_path`](https://github.com/warpdotdev/warp/blob/bb432670251e4a0ab00e8cbd8190299a239217fd/crates/repo_metadata/src/repositories.rs#L246-L246). Callers which have already canonicalized the path can call this instead to bypass the canonicalization performance penalty. I only needed to have AIContextMenu call this to fix the issue.

## Alternatives Considered

Having both [`get_root_for_path`](https://github.com/warpdotdev/warp/blob/1e43a0a0bc610b63edce624a01d1cbae8835f374/crates/repo_metadata/src/repositories.rs#L228-L228) and [`get_root_for_canonical_path`](https://github.com/warpdotdev/warp/blob/bb432670251e4a0ab00e8cbd8190299a239217fd/crates/repo_metadata/src/repositories.rs#L246-L246) is definitely not ideal. It's too easy to use the wrong method. Ideally the type system would help enforce correctness. We already have `CanonicalizedPath` but that isn't used by this `LocalOrRemotePath` wrapper. Maybe we can make it generic `LocalOrRemotePath<LocalPathT>` so that it can represent different types of local paths either `PathBuf` or `CanonicalizedPath`. I'm assuming there are _also_ situations where `StandardizedPath` would be preferred.

## Linked Issue

Closes #9920

## Testing

The issue can be repro'd by opening a bunch of WSL tabs, it may take 10-20 tabs on a fast machine to start noticing the lag. Just activate any of the WSL tabs and it will freeze the UI for a bit.

On this branch, there is no longer perceptible lag. It feels instant!

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -2453,6 +2453,13 @@ impl DropTargetData for TerminalDropTargetData {
     }
 }
 
+/// Cached result of [`TerminalView::canonical_session_pwd_if_local`].
+struct CanonicalSessionPwdCache {
+    /// Non-canonical path
+    path: PathBuf,
+    canonical: PathBuf,
+}
+
 pub struct TerminalView {
     pub model: Arc<FairMutex<TerminalModel>>,
     view_handle: WeakViewHandle<Self>,
@@ -2601,6 +2608,8 @@ pub struct TerminalView {
 
     sessions: ModelHandle<Sessions>,
     active_block_metadata: Option<BlockMetadata>,
+    /// Memoized result of `canonical_session_pwd_if_local`.
+    canonical_session_pwd_cache: RefCell<Option<CanonicalSessionPwdCache>>,
 
     block_text_selection_start_position: Option<Vector2F>,
 
@@ -4290,6 +4299,7 @@ impl TerminalView {
             sessions,
             remote_server_shimmer_handle: ShimmeringTextStateHandle::new(),
             active_block_metadata: None,
+            canonical_session_pwd_cache: RefCell::new(None),
             block_text_selection_start_position: None,
             background_executor: ctx.background_executor().clone(),
             inline_banners_state: Default::default(),
@@ -7643,6 +7653,33 @@ impl TerminalView {
         } else {
             None
         }
+    }
+
+    /// Returns the active local session's CWD, canonicalized via `dunce::canonicalize` to resolve
+    /// symlinks and normalize the path.
+    ///
+    /// The canonicalization is memoized in `canonical_session_pwd_cache`, keyed on the
+    /// non-canonical path, and only recomputed when that path changes.
+    pub fn canonical_session_pwd_if_local<C: ModelAsRef>(&self, ctx: &C) -> Option<PathBuf> {
+        let path = self.active_session_path_if_local(ctx)?;
+
+        // Return the cached value when the non-canonical path has not changed.
+        if let Some(cached) = self.canonical_session_pwd_cache.borrow().as_ref() {
+            if cached.path == path {
+                return Some(cached.canonical.clone());
+            }
+        }
+
+        let canonical = dunce::canonicalize(&path).ok()?;
+
+        if let Ok(mut cache) = self.canonical_session_pwd_cache.try_borrow_mut() {
+            *cache = Some(CanonicalSessionPwdCache {
+                path,
+                canonical: canonical.clone(),
+            });
+        }
+
+        Some(canonical)
     }
 
     pub fn input(&self) -> &ViewHandle<Input> {
@@ -11730,6 +11767,9 @@ impl TerminalView {
 
         self.active_block_metadata = Some(block_metadata.clone());
 
+        // Not needed for correctness, but this method memoizes its result. This "warms" the cache.
+        let _ = self.canonical_session_pwd_if_local(ctx);
+
         if let Some(session) = block_metadata
             .session_id()
             .and_then(|id| self.sessions.as_ref(ctx).get(id))
@@ -13642,10 +13682,9 @@ impl TerminalView {
         self.any_session_contains_remote_blocks |= self.active_block_is_considered_remote(ctx);
         self.update_focused_terminal_info(ctx);
 
-        if let Some(working_directory) = self.pwd_if_local(ctx) {
+        if let Some(working_directory) = self.active_session_path_if_local(ctx) {
             CodebaseIndexManager::handle(ctx).update(ctx, |manager, _ctx| {
-                let path_buf = PathBuf::from(&working_directory);
-                manager.handle_session_bootstrapped(&path_buf);
+                manager.handle_session_bootstrapped(&working_directory);
             });
         }
 
@@ -14086,11 +14125,10 @@ impl TerminalView {
         self.init_project(true, ctx);
     }
 
-    // Show or hide codebase index speedbump depending when a settings change happens.
+    /// Show or hide codebase index speedbump depending when a settings change happens.
     fn check_codebase_index_speedbump_on_settings_changed(&mut self, ctx: &mut ViewContext<Self>) {
-        if let Some(working_directory) = self.pwd_if_local(ctx) {
-            let path_buf = PathBuf::from(&working_directory);
-            self.update_repo_banner_state(path_buf, ctx);
+        if let Some(working_directory) = self.active_session_path_if_local(ctx) {
+            self.update_repo_banner_state(working_directory, ctx);
         }
     }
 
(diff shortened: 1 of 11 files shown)


app/src/terminal/view.rs
```diff
@@ -13682,10 +13682,9 @@ impl TerminalView {
         self.any_session_contains_remote_blocks |= self.active_block_is_considered_remote(ctx);
         self.update_focused_terminal_info(ctx);
 
-        if let Some(working_directory) = self.pwd_if_local(ctx) {
+        if let Some(working_directory) = self.active_session_path_if_local(ctx) {
             CodebaseIndexManager::handle(ctx).update(ctx, |manager, _ctx| {
-                let path_buf = PathBuf::from(&working_directory);
-                manager.handle_session_bootstrapped(&path_buf);
+                manager.handle_session_bootstrapped(&working_directory);
```
acarl005, line 13690: This isn't actually necessary for this fix. I just noticed that `pwd_if_local` is just a wrapper around `active_session_path_if_local` which returns a string instead of a path. We're converting back to a path in an unnecessary roundabout. Might as well just call `active_session_path_if_local`. There are several other places where I've made this change.

crates/warp_util/src/path.rs
```diff
@@ -478,15 +483,8 @@ pub fn convert_wsl_to_windows_host_path(
...
-            windows_path
+            PathBuf::try_from(windows_path).map_err(WSLPathConversionError::CouldNotConvertToPath)
         }
-    };
-    let pathbuf =
-        PathBuf::try_from(windows_path).map_err(WSLPathConversionError::CouldNotConvertToPath)?;
-    // Many directories are symlinks into the underlying file-system location in Windows.
-    match std::fs::read_link(&pathbuf) {
-        Ok(linked_file) => Ok(linked_file),
-        Err(_) => Ok(pathbuf),
```
acarl005, line 489: This is related but I didn't mention it in the PR description. `std::fs::read_link` is also very slow on WSL paths. It also always fails b/c the linux symlink format is different than Windows and `std::fs::read_link` only understands the Windows format when running on a Windows machine. So, I moved that inside the non-WSL branch specifically.

app/src/terminal/view.rs
```diff
@@ -2454,6 +2453,13 @@ impl DropTargetData for TerminalDropTargetData {
     }
 }
 
+/// Cached result of [`TerminalView::canonical_session_pwd_if_local`].
+struct CanonicalSessionPwdCache {
+    /// Non-canonical path
+    path: PathBuf,
```
kevinyang372, line 2462: Are these guaranteed to be local sessions? If so we need to be very explicit about it in struct naming (e.g. LocalSessionPwdCanonicalCache)

app/src/terminal/view.rs
```diff
@@ -2454,6 +2453,13 @@ impl DropTargetData for TerminalDropTargetData {
...
+    canonical: PathBuf,
```
kevinyang372: I still feel like this should be StandardizedPath (I know we debated about this but PathBuf type does not encode anything about the path being canonicalized)

app/src/terminal/view.rs
```diff
@@ -7685,6 +7655,33 @@ impl TerminalView {
         }
     }
 
+    /// Returns the active local session's CWD, canonicalized via `dunce::canonicalize` to resolve
+    /// symlinks and normalize the path.
+    ///
+    /// The canonicalization is memoized in `canonical_session_pwd_cache`, keyed on the
+    /// non-canonical path, and only recomputed when that path changes.
+    pub fn canonical_session_pwd_if_local<C: ModelAsRef>(&self, ctx: &C) -> Option<PathBuf> {
```
kevinyang372: Also I think the return type should be a StandardizedPath to emphasize this is a canonicalized path

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -5208,9 +5210,11 @@ fn compute_tab_group_color_mode(
                     .as_ref(app)
                     .local_path(app)
                     .as_deref()
+                    // TODO(andy): avoid canonicalizing on a render code path
+                    .and_then(|file_path| dunce::canonicalize(file_path).ok())
```
kevinyang372, line 5214: We should definitely avoid canonicalizing on a render path -- why do we need canonicalizing for tab colors?

acarl005: It's already happening. I'm removing 2 callers that are canonicalizing on the render path. I'm not adding any. I'm just making it explicit that we need to remove it eventually.

Canonicalization was happening internally to `fn color_for_directory` before. I removed it from the body of `fn color_for_directory` and made it the responsibility of the caller.

kevinyang372 (approved):

PR #12496, opened 2026-06-11
Pin Metal AIR bytecode target to MACOSX_DEPLOYMENT_TARGET

### LINKED ISSUE #11700: Crash on launch in Renderer::new (Metal) — regression between 0.2026.05.06 and 0.2026.05.20 on Intel Mac w/ Bronze-tier GPU
### Pre-submit Checks

### Describe the bug

Warp crashes immediately on launch with a Rust panic inside the Metal renderer initialization path. The crash is a hard `abort()` from an `unwrap()` on a `Result` returned by Metal device setup — there is no fallback or user-visible error, the app just dies before any window is drawn.

This is a **regression**: version `0.2026.05.06.15.42.stable_05` launches and runs normally on the same machine. Version `0.2026.05.20.09.21.03` crashes 100% of the time on launch. No config, prefs, or sqlite changes between the two installs.

### To reproduce

1. Install Warp `0.2026.05.20.09.21.03` on the affected hardware (see below).
2. Launch Warp.
3. App crashes before any window is drawn. `EXC_CRASH (SIGABRT)`, `abort() called`.

Reinstalling the same version reproduces 100%. Reinstalling `0.2026.05.06.15.42.stable_05` works 100%.

### Expected behavior

Either the renderer initializes successfully, or — if the GPU genuinely can't support whatever Metal feature was added between these builds — Warp falls back to a compatible code path or surfaces a clear error to the user instead of aborting.

[mac_os_crash.txt](https://github.com/user-attachments/files/28279376/mac_os_crash.txt)

### Operating system (OS)

macOS

15.7.5 (24G624)

### Shell Version

_No response_

### Current Warp version

_No response_

### Regression

### Recent working Warp date

20 May 2026

### Additional context

This machine runs macOS 15 via OpenCore Legacy Patcher, so the Bronze-tier Metal drivers are in active use rather than being a deprecated code path the OS would normally avoid. That makes it an unusually good canary for Metal feature-set regressions: anything in the renderer that newly requires Metal 3, argument buffers tier 2, or similar will surface here as a hard failure rather than degraded rendering.

I'm not asking for OCLP to be officially supported — but the same `unwrap()` would also crash any real older Intel Mac that hits the same Metal limitation, and the crash is unconditional `abort()` with no diagnostic.

System Info

- **Model:** MacBookPro11,5 (mid-2015 15" MacBook Pro)
- **CPU:** Quad-Core Intel Core i7 @ 2.8 GHz (x86_64)
- **GPUs:** Intel Iris Pro (integrated) + AMD Radeon R9 M370X (discrete) — both Bronze-tier Metal
- **macOS:** 15.7.5 (24G624)
- **Loaded GPU driver at crash:** `AMDMTLBronzeDriver` 4.8.101, `AppleIntelHD5000GraphicsMTLDriver` 18.8.4
- **Warp version (broken):** `0.2026.05.20.09.21.03`
- **Warp version (working):** `0.2026.05.06.15.42.stable_05`

None

maxmilian:
## Description

`xcrun metal` on recent Xcode toolchains emits AIR targeted at the current SDK (e.g. `air64_v27-apple-macosx15.0.0`) regardless of the `MACOSX_DEPLOYMENT_TARGET` env var. On older Metal driver stacks (legacy Bronze drivers on older Intel Macs, OCLP machines running macOS 15 with injected legacy drivers) the resulting `.metallib` is rejected during pipeline state creation, which `Renderer::new` unwraps and aborts the app on launch.

The crash is shader-cache-mediated: a fresh launch can JIT-compile the shaders, the legacy driver then writes a corrupted cache entry to `/var/folders/.../com.apple.metal/`, and every subsequent launch aborts when the driver tries to deserialize that cache. The reporter in #11700 narrowed this down to a deployment-target mismatch (`air64_v27-apple-macosx15.0.0` vs. the driver's `air64-apple-macosx14.2.0` ceiling) and provided the exact error string.

Pass `-mmacosx-version-min=$MACOSX_DEPLOYMENT_TARGET` to `xcrun metal -c` so the AIR bytecode matches the rest of the build's minimum deployment target (`10.14`, set in `.cargo/config.toml`). This mirrors the same pattern already in use for the dock tile plugin build in `app/build.rs`.

## Linked Issue

Fixes #11700

## Testing

- [x] `./script/format --check` (Rust 1.92.0 per `rust-toolchain.toml`)

I do not have a full Xcode install on this machine, only Command Line Tools — `xcrun metal` is unavailable, so `cargo check -p warpui` cannot complete the build script and `./script/run` cannot be run. The change is a one-line addition to an existing argument vector, mirroring the `MACOSX_DEPLOYMENT_TARGET` plumbing in `app/build.rs` for the dock tile plugin build.

Requesting:
- CI to confirm `xcrun metal` accepts `-mmacosx-version-min=$MACOSX_DEPLOYMENT_TARGET` and the metallib still builds.
- @trfeng91 (issue reporter — has a MacBookPro11,5 / OCLP setup with the affected Bronze driver stack and explicitly offered to test pre-release builds in the issue) to verify that launch and subsequent relaunches no longer crash.

### Screenshots / Videos

N/A — build-system change with no UI surface.

## Agent Mode

### DIFF
--- a/crates/warpui/build.rs
+++ b/crates/warpui/build.rs
@@ -81,8 +81,27 @@ fn compile_metal_shaders() {
 
     println!("cargo:rerun-if-changed={header_path}");
     println!("cargo:rerun-if-changed={metal_path}");
+    println!("cargo:rerun-if-env-changed=MACOSX_DEPLOYMENT_TARGET");
 
-    let mut compile_args = vec!["-sdk", "macosx", "metal", "-c", metal_path, "-o", air_path];
+    // Pin the AIR bytecode target to `MACOSX_DEPLOYMENT_TARGET`. Without an
+    // explicit `-mmacosx-version-min`, `xcrun metal` on recent Xcode toolchains
+    // emits AIR for the current SDK (e.g. `air64_v27-apple-macosx15.0.0`)
+    // regardless of the env var, and older Metal driver stacks reject the
+    // resulting `.metallib` during pipeline state creation. See #11700.
+    let min_macos_version = env::var("MACOSX_DEPLOYMENT_TARGET")
+        .expect("MACOSX_DEPLOYMENT_TARGET must be set for macOS builds");
+    let min_version_arg = format!("-mmacosx-version-min={min_macos_version}");
+
+    let mut compile_args = vec![
+        "-sdk",
+        "macosx",
+        "metal",
+        "-c",
+        metal_path,
+        "-o",
+        air_path,
+        &min_version_arg,
+    ];
     if cfg!(feature = "enable-metal-frame-capture") {
         compile_args.push("-frecord-sources");
         compile_args.push("-gline-tables-only");


maxmilian:
Quick follow-up after checking the shipped binaries directly, since this design was based on the reporter's experimental `-std=metal3.2` repro producing `air64_v27-apple-macosx15.0.0`.

`strings` on the embedded metallib bytes in both:
- `v0.2026.05.20.09.21.stable_03` (the build pinpointed as the regression in #11700): binary SHA-256 `30b4f1eba1c4d4c752593f6ca8911e8e09625efa020aad75ba67c880e1e9875b`, 797,137,344 bytes
- `v0.2026.06.03.09.49.stable_03` (current shipped stable): binary SHA-256 `c601cbeda784f00c92d0e2ec0338ce12d535b81b42b1ea834fa29761fef6f31b`, 801,999,088 bytes

Both binaries print `air64-apple-macosx10.14.0` for every shader (rect/glyph/image vertex+fragment). The binaries themselves are different builds — different SHA-256, ~5 MB size delta — so it's not a fingerprint collision, but the AIR deployment target is already pinned at `10.14` in shipped builds.

This is what you'd expect: `xcrun metal` honors `MACOSX_DEPLOYMENT_TARGET`, which is set to `10.14` in `.cargo/config.toml`'s `[env]` block, so it gets plumbed through to the compiler without needing the explicit flag.

What this means for this PR: forcing `-mmacosx-version-min=$MACOSX_DEPLOYMENT_TARGET` explicitly should produce a byte-equivalent metallib to today's, since the env var already does this. That makes the change more of a defense-in-depth measure (explicit > implicit, guards against `MACOSX_DEPLOYMENT_TARGET` not being set in some future build env) than a fix for the `Renderer::new` crash in #11700.

The actual stable_02 → stable_03 regression vector is elsewhere — the reporter measured a metallib SHA change (`1f8702c6…` → `f522af78…`) at the same AIR target, which points at `-std=` MSL language standard variation between Xcode toolchain versions on the build runners, or optimizer changes inside `xcrun metal`. The Bronze cache deserializer choking on those differences is the actual bug surface, not the deployment target.

Happy to either:
- Close this PR and reopen with `-std=macos-metal2.x` pinning if that's the direction you'd want; or
- Keep this open as a hardening change without the `Fixes #11700` linkage; or
- Hold off entirely while the actual regression vector gets investigated.

Whichever's most useful — let me know.

maxmilian:
## Correction: this PR's premise is wrong — retracting "Fixes #11700"

After opening this PR I did some binary archaeology on the shipped stable builds to double-check the deployment-target hypothesis. The data contradicts it, so I want to correct the record before anyone reviews this on a false premise.

**Method:** downloaded the shipped dmgs from `releases.warp.dev`, located the embedded `shaders.metallib` in each fat-binary slice by its `MTLB` magic (the metallib is embedded per-slice via `include_bytes!` at `crates/warpui/src/platform/mac/rendering/metal/renderer.rs:37`), carved it out using the container header's size field, and compared SHA-256 / AIR target triples / compiler ident strings. The two SHAs below match the ones @trfeng91 reported in #11700, so we're measuring the same artifacts.

| Build | x86_64 slice metallib | arm64 slice metallib | @trfeng91's Intel/Bronze result |
|---|---|---|---|
| `v0.2026.05.20.09.21.stable_00` | `1f8702c6…` | `1f8702c6…` | launches |
| `v0.2026.05.20.09.21.stable_01` | `1f8702c6…` | `1f8702c6…` | launches |
| `v0.2026.05.20.09.21.stable_02` | `1f8702c6…` | `f522af78…` | launches, relaunches fine |
| `v0.2026.05.20.09.21.stable_03` | `f522af78…` | `f522af78…` | crashes |
| `v0.2026.06.03.09.49.stable_03` (current stable) | `f522af78…` | `f522af78…` | — |

**1. Deployment target was never the problem.** Every one of these metallibs — including the crashing ones — already targets `air64-apple-macosx10.14.0`, with the identical compiler ident `Apple metal version 32023.883 (metalfe-32023.883)`. So `MACOSX_DEPLOYMENT_TARGET=10.14` from `.cargo/config.toml` *is* reaching the Metal compiler in the release pipeline — answering @vorporeal's "(if it's not already set by compiler infra using that envvar)": it is. The `air64_v27-apple-macosx15.0.0` error quoted in the issue came from trfeng91's deliberately modified local build (`-std=metal3.2`), not from any shipped binary. This PR's `-mmacosx-version-min` flag therefore wouldn't change a single shipped byte — defense-in-depth at best, not a fix.

**2. There are two metallib *variants*, and they don't map to "old vs. new toolchain" or to CPU architecture.** `1f8702c6` and `f522af78` are the same size (31,233 bytes), declare the same compiler version and the same AIR target, but differ in 165 bytes — including all five per-function `HASH` entries, i.e. genuinely different bitcode. Each variant is byte-identical wherever it appears (across slices and across builds weeks apart), so this is systematic, not build nondeterminism. The rollout pattern in the table — arm64 slice flips at stable_02, x86_64 slice flips at stable_03, then it's uniform through today's stable — looks like a build-runner fleet being upgraded gradually (same `metalfe` version, but e.g. a different macOS SDK / `metal_stdlib` revision on the new runners would produce exactly this kind of small codegen delta).

One correction to the thread's data: stable_00 ships `1f8702c6` in **both** slices — not `f522af78` as measured in https://github.com/warpdotdev/warp/issues/11700#issuecomment-4549494678. That also dissolves the "stable_00 wrote the corrupted cache entry" part of the narrative; the bisect results are all consistent with the simpler story below.

**3. The crash boundary is exactly the x86_64 slice switching variants.** trfeng91's Intel/Bronze stack: every build whose x86_64 slice carries `1f8702c6` launches and relaunches fine; the first build whose x86_64 slice carries `f522af78` (stable_03) is the first one that crashes. Combined with his cache experiments (fresh cache → launch succeeds; relaunch → `"Compiler encountered an internal error"` from the driver's shader-cache deserializer; CLI probe without a bundle ID → no persistent cache → never crashes), the root-cause picture is:

> **The legacy Bronze driver's shader-cache serializer handles the `1f8702c6` variant fine but writes a corrupt cache entry for the `f522af78` variant and crashes deserializing it on the next launch.** The regression was the release pipeline's compile environment changing between stable_02 and stable_03 (for the x86_64 slice), not any change in this repo — shader source, compiler flags, and deployment target are identical on both sides of the boundary.

Current stable (June 3) still ships `f522af78` in the x86_64 slice, so the crash should still reproduce today.

**Where this leaves this PR — maintainers' call:**
1. Close it (my default if I don't hear otherwise).
2. Keep it as explicit hardening — pinning what is currently an implicit env-var dependency — with the "Fixes #11700" claim dropped. (For the same reason, `-std=` pinning would also not have prevented this: both variants already come from identical compiler versions and flags.)
3. The repo-side change that would actually help users on legacy stacks is replacing the `unwrap()`s in `Renderer::new` with recoverable error handling (clear error message, possibly a one-shot retry after clearing/bypassing the pipeline cache), as trfeng91 suggested — that turns this class of driver bug from "abort before first frame" into something diagnosable. Happy to do that as a separate PR.

The fix for #11700 itself looks like it lives in Warp's internal build/release infra (what changed on the runners that produce the shipped metallib between stable_02 and stable_03), which isn't visible from the OSS repo.

Sorry for the noise — I should have done this verification before opening the PR, not after.

trfeng91:
@maxmilian You're right, nice catch. I re-checked everything from my end and your
  numbers hold up, including the stable_00 correction. And I figured out how I botched
  that measurement in the first place.

  I re-carved the metallibs from the dmgs I still had lying around from May, plus a
  fresh stable_00 download (same method as yours - `MTLB` magic, size field, SHA-256
  per slice):

  | Build | x86_64 | arm64 |
  |---|---|---|
  | May 6 `stable_05` | `1f8702c6…` | `1f8702c6…` |
  | `stable_00` (fresh download) | `1f8702c6…` | `1f8702c6…` |
  | `stable_02` | `1f8702c6…` | `f522af78…` |
  | `stable_03` | `f522af78…` | `f522af78…` |
  | `dev_00` | `f522af78…` | `f522af78…` |

  Matches your table exactly, mixed-slice stable_02 included, and everything reports
  `air64-apple-macosx10.14.0` / `metalfe-32023.883`. So yeah, deployment target theory
  is dead (mine included), and my "stable_00 wrote the corrupt cache" story goes with it.

  **How my stable_00 measurement went wrong:** the auto-updater. It replaces the .app
  in place after launch, I literally have a bundle saved as `Warp-stable_02.app` whose
  Info.plist now says `…09.21.03`, while my log from that session shows it booting as
  stable_02. So back in May I carved "stable_00" *after* running it and got stable_03's
  bytes. PSA for anyone poking at shipped Warp builds: carve from the dmg, never from a
  bundle you've launched.

  One fun extra: of the 165 differing bytes, 160 are just the five `HASH` entries. The
  real bitcode delta is **5 bytes, one per shader function, each a single extra flag
  bit** (`0x46→0x56`, `0x8d→0xad`, `0x23→0x2b`). That's all it takes to make the Bronze
  cache serializer fall over, which fits your same-compiler/different-SDK read.

  Agreed on closing this one, and +1 on the `Renderer::new` error-handling PR, that's
  the change that would've made this debuggable from day one. Happy to test it on the
  affected hardware when it's up; current stable still ships `f522af78` in the x86_64
  slice, so my repro is alive and well.

vorporeal (approved):
thanks for the fix!

EDIT: thanks for the change that does some hardening and will avoid someone in the future thinking this might not be correct :)

vorporeal:
i can try to see what changed between those two patch releases, but i'm suspicious/worried that it might be something that changed out from under us in GitHub's runner images.

vorporeal:
and yeah, replacing low-info panics with more descriptive error handling (even if it eventually panics/terminates the app) would be a very welcomed PR. :)

PR #12497, opened 2026-06-11
Surface raw error debug details in transient network error messages

advait-m:
Screenshot:
image

## Description
Follow-up to #12382 (REMOTE-971), addressing review feedback that the friendly transient-network message dropped all of the detailed debugging info: it's valuable to show the exact error details so user reports can disambiguate different causes.

**Stacked on #12431** (`advait/remote-1894-resume-cloud-runs-on-transport-failures`) — it builds on the `AIApiError::StreamTruncated` error type and the `recovery_pending` plumbing introduced there. Only the last commit is new.

Changes:
- Add a structured `RenderableAIError::TransientNetworkError { debug_details, ... }` variant, replacing the string-only construction. The raw underlying error (e.g. `Transport(reqwest::Error {...})` or `StreamTruncated`) is preserved at conversion time and rendered as a `Debug info:` line under the friendly message:

  > Warp lost connection while receiving the agent response. This is usually temporary.
  >
  > Debug info: Transport(reqwest::Error { kind: Decode, source: hyper::Error(Body, ...) })

- Drop the generic `I'm sorry, I couldn't complete that request.` apology prefix for these errors only — they carry their own complete user-facing copy, so the prefix added nothing (also per review feedback). All other error copy is unchanged.
- The debug details flow through `Display`, so telemetry (`AgentModeError`), task status sync, and desktop notification descriptions regain pre-#12382 parity with the raw error included.
- Task-sync classification for the new variant intentionally matches the previous `Other` behavior (`ERROR` + `INTERNAL_ERROR`) — no behavior change there.

## Linked Issue
Linear: REMOTE-971

## Testing
- `cargo nextest run -p warp` (targeted: `transient_network_error*`, `classify*`, history model stream-error tests) — all pass
- `./script/format`
- `cargo clippy --workspace --all-targets --all-features --tests` — clean


advait-m:
> * ⚠️ [IMPORTANT] `app/src/ai/agent/conversation.rs` still computes `will_attempt_to_resume` for `AgentModeError` telemetry with a direct `RenderableAIError::Other { will_attempt_resume: true, .. }` match. After this PR, resumable transient network errors use `RenderableAIError::TransientNetworkError`, so that telemetry field will be reported as false unless the match is updated or the existing `will_attempt_resume()` helper is used.


2nd concern isn't relevant against downstacked branch (I think bot is comparing against master maybe?)

app/src/ai/agent/mod.rs
```diff
@@ -666,9 +676,16 @@ pub enum RenderableAIError {
...
+    /// Creates a transient network error. `debug_details` is the raw underlying failure (e.g.
+    /// the transport error's debug representation), preserved so user reports can disambiguate
+    /// the different causes behind the shared user-facing copy.
+    pub fn transient_network_error(
+        will_attempt_resume: bool,
+        waiting_for_network: bool,
+        debug_details: impl Into<String>,
+    ) -> Self {
+        Self::TransientNetworkError {
+            debug_details: debug_details.into(),
```
vorporeal: can we stick the actual error in here?  is there a reason that we need to collapse this (structured data) down to a string representation, or can we keep the raw structured data, and stringify (in the desired string representation at point-of-use) later?

i think the place we're doing the conversion has an `Arc<AIApiError>`, and we're turning that into a `&AIApiError` and then converting here.  maybe we can instead change the `From` to be `From<&Arc<AIApiError>>`, and clone the `Arc<AIApiError>` for the transient case, where we want to hold onto the error itself?

advait-m: agree raw would be nice - updated to go with raw error (wasn't possible earlier due to the derives). turns out the serde/Eq/Hash derives on `RenderableAIError` were vestigial (nothing actually serialized or hashed it), so stripping them let me hold the raw error directly.

`TransientNetworkError` now carries a `kind: TransientNetworkErrorKind` instead of a flattened string. the lost-connection/truncated-stream arms hold `Api(Arc<AIApiError>)` via `From<&Arc<AIApiError>>` (just clones the arc like suggested), and the two synthetic sites get their own variants (unfinished or missing exchange). Only stringifies at `Display` time - user reports keep the full structured error now

app/src/ai/agent/mod.rs
```diff
@@ -695,14 +711,46 @@ impl RenderableAIError {
...
 }
 
-impl From<&AIApiError> for RenderableAIError {
-    fn from(value: &AIApiError) -> Self {
-        match value {
+/// The cause behind a [`RenderableAIError::TransientNetworkError`]. Kept structured (rather than
+/// collapsed to a free-form string) so user reports preserve the raw error; rendered to text only
+/// at display time.
+#[derive(Clone, Debug)]
+pub enum TransientNetworkErrorKind {
```
vorporeal, line 726: should this implement Error?  (i.e.: do a `derive(Error)` for it)

i think that would give us the `Display` impl, and feels more correct than a non-error "kind" enum

vorporeal (approved):

app/src/ai/agent/mod.rs
advait-m: ah, yep that would be nice - done (updated to `#[derive(Clone, Debug, thiserror::Error)]` and added in display strings in-line)

PR #12513, opened 2026-06-11
Add telemetry for SuperGrok subscription connect (success/failure)

szgupta:
WISOTT


app/src/settings_view/ai_page.rs
```diff
@@ -2138,6 +2148,8 @@ impl AISettingsPageView {
             }
         };
 
+        send_telemetry_from_ctx!(TelemetryEvent::SuperGrokSubscriptionConnectInitiated, ctx);
```
danielpeng2: should we move this to above L2128? Seems weird that if we hit the Err branch on L2130 that we'd have a `SuperGrokSubscriptionConnectFinished` without a `SuperGrokSubscriptionConnectInitiated`.

danielpeng2 (approved):

PR #12514, opened 2026-06-11
Add auto-handoff sleep modal and success toast

dplakon:
## Description

This PR replaces the auto-handoff sleep discoverability notification with a Figma-matched one-time modal and adds a success toast when automatic local-to-cloud handoff completes.

What changed:
- Detects eligible in-progress local agent runs when the computer is about to sleep and records that the user should see discoverability on wake if auto-handoff-on-sleep is disabled. The sleep-time evaluation shares a single eligibility pipeline (`evaluate_handoff_candidate`) with the automatic handoff trigger.
- Shows a once-per-user modal on wake explaining auto-handoff and offering to enable the setting.
- Adds a green success toast, "Handed session off to the cloud", when an automatic handoff succeeds; if success happens while the machine is sleeping, the toast is deferred until wake.
- Pauses auto-resume-after-error while the sleep modal is open so an interrupted run doesn't resume locally while the user is deciding: the resume composes `NetworkStatus::wait_until_online()` with a Condition-based `OneTimeModalModel::wait_until_auto_handoff_sleep_modal_closed()`, and is cancelled by a manual request or cancel via `pending_auto_resume_handles`.
- Removes the earlier notification/action-button implementation for this UX.
- Adds telemetry for sleep prompt shown/enabled/dismissed events.
- Adds debug-only palette actions to open/reset the modal and trigger the auto-handoff flow for testing.
- Hardens automatic handoff lookup against stale focused terminal IDs, and fixes the workspace registry at the source so restored windows re-register in `Workspace::handle_reopen`.
- Ships unconditionally to all users: there is no feature flag. The modal is gated only by the once-per-user `did_show_auto_handoff_sleep_modal` setting and the `auto_handoff_on_sleep_enabled` preference.

## Linked Issue

No linked GitHub issue.

## Testing

Manual testing:
- Verified the real sleep/wake flow surfaces the modal on wake using a dogfood build (`cargo run --bin warp`).
- Verified `[Debug] Open Auto-Handoff Sleep Modal` shows the modal.
- Verified `[Debug] Reset Auto-Handoff Sleep Modal State` resets the once-ever state for retesting.
- Verified `[Debug] Trigger Auto-Handoff to Cloud` triggers automatic handoff for an in-progress local agent run.
- Verified the green success toast appears after successful automatic handoff.

Automated validation:
- Added unit tests for the modal close-condition future (`one_time_modal_model_tests.rs`: resolves while closed, pends while open including for futures created before the modal opened, resolves on dismissal) and for cancel aborting a scheduled auto-resume (`controller_tests.rs`); `cargo nextest run -p warp` passes.
- `./script/format` passed.
- `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings` passed.
- `cargo check -p warp --lib` and `cargo check -p warp_features` passed.
- `./script/presubmit` passed formatting/clippy/static checks, then failed during SSH integration tests because the local environment was not authorized for the IAP tunnel / backend endpoints. Representative failure: `ConnectionCreationError: Error while connecting [4033: 'not authorized']` in `integration::shell_integration_tests::test_legacy_ssh_into_bash`. No feature-specific presubmit failure was observed.

No new integration test was added; the flow depends on system sleep/wake and cloud handoff state, and we agreed to continue without adding integration coverage for this PR.

### Screenshots / Videos

Screenshot 2026-06-11 at 12 20 02 PM

Screenshot 2026-06-10 at 11 19 07 AM

## Agent Mode

Agent context:
- Conversation: https://staging.warp.dev/conversation/d7edf7c7-c4d7-4857-b93a-4080e06bb5bd
- Plan artifact: https://staging.warp.dev/drive/notebook/rra4szH0TH53r1T83CjYHs

CHANGELOG-IMPROVEMENT: Warp now offers a one-time prompt to enable automatic local-to-cloud handoff when your computer sleeps during a local agent run, and shows a confirmation toast when an automatic handoff completes.


app/src/ai/blocklist/controller.rs
```diff
@@ -2001,6 +2015,59 @@ impl BlocklistAIController {
...
+    /// open, in which case the resume is parked until the modal closes so it doesn't race the
+    /// user's enable/dismiss decision.
+    fn auto_resume_after_error_or_pause_for_modal(
+        &mut self,
+        conversation_id: AIConversationId,
+        ctx: &mut ModelContext<Self>,
+    ) {
+        if OneTimeModalModel::as_ref(ctx).is_auto_handoff_sleep_modal_open() {
+            self.auto_resumes_paused_for_auto_handoff_modal
+                .insert(conversation_id);
```
harryalbert: do we resume the conversation after the user chooses one of the two options in the modal? I think we probably should

app/src/workspace/view/auto_handoff_sleep_modal/view.rs
```diff
@@ -0,0 +1,302 @@
...
+use crate::ui_components::icons::Icon;
+use crate::view_components::action_button::{
+    ActionButton, ActionButtonTheme, ButtonSize, PrimaryTheme, SecondaryTheme,
+};
+
+const MODAL_WIDTH: f32 = 420.;
+const HERO_HEIGHT: f32 = 92.;
+const HERO_IMAGE_PATH: &str = "async/png/onboarding/auto_handoff_sleep_banner.png";
+
+fn modal_background(appearance: &Appearance) -> Fill {
```
harryalbert: nit: I don't think this needs to be a separate fn. we can just inline. Same with all of the fns below that just pass through an appearance value. IMO the extra fns just add an extra layer of complexity where inlining would probably be clearer

app/src/workspace/auto_handoff.rs
```diff
@@ -1,18 +1,29 @@
...
+use crate::ai::ambient_agents::telemetry::CloudAgentTelemetryEvent;
 use crate::settings::AISettings;
 use crate::system::{SystemStats, SystemStatsEvent};
 use crate::terminal::view::TerminalView;
+use crate::view_components::DismissibleToast;
 use crate::BlocklistAIHistoryModel;
 
+/// Body text of the toast shown after an automatic handoff completes
+/// successfully.
+const AUTO_HANDOFF_SUCCESS_TOAST_TEXT: &str = "Handed session off to the cloud";
```
harryalbert: nit: I think we can just inline this

app/src/workspace/auto_handoff.rs
```diff
@@ -132,9 +164,168 @@ impl AutoCloudHandoffController {
...
+                ctx,
+            );
+        });
+    }
+
+    /// At sleep time, records a pending discoverability prompt when an eligible
+    /// in-progress local agent run *would* have auto-handed-off but the sleep
+    /// setting is off. The actual modal is surfaced on wake by
+    /// [`Self::maybe_show_sleep_prompt`].
+    fn maybe_record_sleep_prompt(&mut self, ctx: &mut ModelContext<Self>) {
```
harryalbert: A lot of this pipeline/logic is pretty much the same as `prepare_handoff_request`, right?

 I think we can get rid of a lot of the duplication by extracting the shared core into something like `evaluate_handoff_candidate(...) -> Result<Candidate, AutoCloudHandoffSkipReason>` (extending the skip-reason enum with the precondition failures like NoFocusedConversation / TerminalNotFound). Then both this and `prepare_handoff_request` become thin wrappers, and the sleep path reads as:
```
// In handle_system_stats_event, CpuWillSleep:
match Self::evaluate_handoff_candidate(&self.attempted_conversation_ids, ctx) {
    Err(reason) => log::info!("auto handoff: skipping at sleep: {reason:?}"),
    Ok(candidate) if setting_on => self.dispatch_handoff(candidate, trigger, ctx),
    Ok(_) => self.pending_sleep_prompt = true, // discoverability prompt on wake
}
```

app/src/workspace/one_time_modal_model.rs
```diff
@@ -136,6 +139,75 @@ impl OneTimeModalModel {
...
+        should_show
+    }
+
+    #[cfg(debug_assertions)]
+    pub fn force_open_auto_handoff_sleep_modal(&mut self, ctx: &mut ModelContext<Self>) {
+        self.set_auto_handoff_sleep_modal_open(true, ctx);
+    }
+
+    #[cfg(test)]
+    pub fn set_auto_handoff_sleep_modal_open_for_test(
```
harryalbert: what's the point of this for_test fn? It seems like it just passes through to the `set_auto_handoff_sleep_modal_open` fn

app/src/workspace/one_time_modal_model.rs
```diff
@@ -136,6 +139,75 @@ impl OneTimeModalModel {
...
+            }
+        });
+
+        let should_show = !matches!(ChannelState::channel(), Channel::Integration);
+        self.set_auto_handoff_sleep_modal_open(should_show, ctx);
+        should_show
+    }
+
+    #[cfg(debug_assertions)]
+    pub fn force_open_auto_handoff_sleep_modal(&mut self, ctx: &mut ModelContext<Self>) {
```
harryalbert: what's the point of this fn? Couldn't we just call `set_auto_handoff_sleep_modal_open(true, ctx)` directly?

app/src/ai/blocklist/controller.rs
harryalbert, line 1: I don't love keeping all of modal related state on the blocklist controller. Especially for a one time modal, adding this extra state and fn on top of the controller feels like bleeding of concerns to me.

The auto-resume already composes on `wait_until_online`(controller.rs:2025). I think we could give `OneTimeModalModel` a `wait_until_no_modal_open()` fn using the same Condition pattern `NetworkStatus` uses (app/src/network.rs (24-88)). Then the resume becomes:
```
let wait = async move { wait_online.await; wait_modal_closed.await };
let handle = ctx.spawn(wait, move |me, _, ctx| { ... resume ... });
```
and we can keep `pending_auto_resume_handles` as the single source of truth. That would let us get rid of `auto_resumes_paused_for_auto_handoff_modal`, the subscription, the drain method, etc.

app/src/workspace/auto_handoff.rs
```diff
@@ -224,13 +450,35 @@ impl AutoCloudHandoffController {
...
             .find_map(|(window_id, workspace)| {
                 let terminal_view = workspace.as_ref(ctx).terminal_view(terminal_view_id, ctx)?;
                 Some((window_id, workspace, terminal_view))
-            })
+            });
+        if from_registry.is_some() {
+            return from_registry;
+        }
+
+        // The registry can be empty or stale: `on_window_closed` unregisters
```
harryalbert: I don't love papering over this registry issue instead of fixing at the source. If it's not too big of a refactor, I think ideally we'd just fix the registry in this PR so that it re-registers windows on restore. 

If that is too big of a refactor, I think we should put this fallback in `WorkspaceRegistry` directly (so that it can be used by other callers) with a TODO comment to fix at the source later

harryalbert (commented):
Cool change! Just had some nits and a few larger thoughts about some possible refactors

app/src/ai/blocklist/controller.rs
dplakon: yes, it resumes

app/Cargo.toml
```diff
@@ -669,6 +669,7 @@ default = [
     "git_credential_refresh",
     "oz_handoff",
     "handoff_local_cloud",
+    "auto_handoff_sleep_prompt",
```
harryalbert: any reason we're feature flagging this if we're putting it straight into prod? Feels like we could probably just get rid of the flag IMO

app/src/workspace/auto_handoff.rs
```diff
@@ -251,8 +421,16 @@ pub(crate) fn trigger_auto_handoff_to_cloud(
...
-    AutoCloudHandoffController::handle(ctx).update(ctx, |controller, ctx| {
-        controller.trigger(trigger, ctx);
+    AutoCloudHandoffController::handle(ctx).update(ctx, |_, ctx| {
+        // Defer the trigger to the next executor turn so it runs outside any
+        // in-progress view update. `update_view` temporarily removes a view
+        // from its window while updating it, so when this is dispatched from
+        // a workspace action (e.g. the debug palette entry), a synchronous
+        // workspace lookup would not find the dispatching workspace — its
+        // registry weak-handle fails to upgrade and `views_of_type` misses it.
+        ctx.spawn(async {}, move |controller, _, ctx| {
```
harryalbert: this `ctx.spawn` deferral is a bit of a code smell. Is this only needed for the debug palette entry? If so, I'd rather either:
* scope the ctx spawn to the debug palette case specifically, or
* get rid of the debug palette entry entirely
as opposed to using `ctx.spawn` for the normal case where it's not needed

harryalbert (commented):
Just had one more thought and one question. After those two are resolved, should be good to go!

harryalbert (approved):
nice

PR #12516, opened 2026-06-11
Refresh expired Grok OAuth tokens at request time instead of silently dropping them

szgupta:
## Description

**What:** Connected SuperGrok subscriptions could silently stop authenticating. Once the stored OAuth access token came within 60s of expiry, the client dropped it from `api_keys` on MAA requests — no logs, no UI signal. The proactive refresh loop could also fail to ever arm, because `grok_refresh_allowed` is computed once at `ApiKeyManager` init from the BYO policy (often before auth/team state has loaded) and was only re-synced on `TeamsChanged`. Diagnosed from a real session: zero refresh activity for 10.5 hours after connecting, the 6-hour token expired, and every subsequent request silently omitted it.

**Changes:**
- Always send the stored Grok access token, even past expiry — the server is the authority on validity (`GrokTokens::access_token_for_request` replaces the expiry-gated `valid_access_token`).
- Request-time background refresh safety net: when a request is built and the stored token is nearing (or already past) expiry, `ApiKeyManager::refresh_grok_tokens_if_needed` kicks off a background refresh. It also re-syncs the BYO policy mirror from the freshly evaluated policy, so the proactive refresh loop re-arms after a stale-policy startup — fixing the root cause above. The triggering request is never delayed; it carries the currently stored token.
- Refresh attempts are guarded by an in-flight flag so the proactive timer and the request-time safety net can't issue overlapping refreshes; a failed attempt no longer kills the refresh loop permanently (the next request re-triggers it).

**Follow-up (separate PR):** a blocking, one-shot refresh at request time for tokens already past hard expiry, so the first request after a long idle period goes out with a fresh token instead of relying on the stale one.

## Linked Issue

N/A — diagnosed from local logs and keychain state ([conversation](https://staging.warp.dev/conversation/b2d9cfb0-4ca0-4da5-8815-4d297b1f0034)).

## Testing

- Unit tests in `crates/ai/src/api_keys_tests.rs`: expired tokens are still returned and included in request keys; `needs_refresh` covers near-expiry, expired, and unknown-expiry tokens.
- `cargo nextest run -p ai` (260 tests) passes; presubmit `cargo clippy` and `./script/format` pass. `app/src` is untouched except a 15-line hook in `BlocklistAIController::send_request_input`.
- Not manually tested end-to-end against xAI (requires holding a token at its expiry boundary); the failure mode was validated from logs and stored token state. The new info/error log lines make the token lifecycle observable.
- No integration test: the flow requires a real xAI OAuth token exchange, which can't be exercised hermetically.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed connected SuperGrok subscriptions silently failing to authenticate once the OAuth access token expired; tokens are now refreshed automatically in the background as expiry nears.

### DIFF
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -2372,9 +2372,8 @@ impl BlocklistAIController {
         // Safety net: if the connected Grok subscription's OAuth token is
         // nearing expiry, kick off a background refresh so upcoming requests
         // can authenticate even when the proactive refresh loop isn't
-        // running. (A token already past its expiry is instead refreshed by a
-        // blocking, one-shot attempt in `ResponseStream` before the request
-        // is sent.)
+        // running. (A token already past its expiry is instead refreshed by
+        // the blocking, one-shot attempt started below.)
         #[cfg(not(target_family = "wasm"))]
         {
             use ::ai::api_keys::ApiKeyManager;
@@ -2396,6 +2395,25 @@ impl BlocklistAIController {
         request_params.parent_agent_id = parent_agent_id;
         request_params.agent_name = agent_name;
 
+        // If the Grok OAuth token this request carries is already past its
+        // hard expiry, start a single refresh attempt for the request to
+        // block on before it is sent (awaited in the request task, off the
+        // main thread). If the attempt fails, the request proceeds with the
+        // stale token — the server is the authority on its validity.
+        #[cfg(not(target_family = "wasm"))]
+        let pending_grok_token = request_params
+            .api_keys
+            .as_ref()
+            .is_some_and(|keys| !keys.grok_oauth_access_token.is_empty())
+            .then(|| {
+                ::ai::api_keys::ApiKeyManager::handle(ctx).update(ctx, |manager, ctx| {
+                    manager.blocking_grok_refresh_for_request(ctx)
+                })
+            })
+            .flatten();
+        #[cfg(target_family = "wasm")]
+        let pending_grok_token = None;
+
         let server_conversation_token_for_identifiers =
             conversation_data.server_conversation_token.clone();
 
@@ -2412,6 +2430,7 @@ impl BlocklistAIController {
                 request_params.clone(),
                 ai_identifiers,
                 can_attempt_resume_on_error,
+                pending_grok_token,
                 ctx,
             )
         });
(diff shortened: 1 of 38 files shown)


app/src/ai/blocklist/controller.rs
```diff
@@ -2369,6 +2369,21 @@ impl BlocklistAIController {
...
+        // nearing or past expiry, kick off a background refresh so upcoming
+        // requests can authenticate even when the proactive refresh loop
+        // isn't running. This request still carries the currently stored
+        // token; the server is the authority on its validity.
+        #[cfg(not(target_family = "wasm"))]
+        {
+            use ::ai::api_keys::ApiKeyManager;
+
+            let byo_allowed = UserWorkspaces::as_ref(ctx).is_byo_api_key_enabled(ctx);
+            ApiKeyManager::handle(ctx).update(ctx, |manager, ctx| {
```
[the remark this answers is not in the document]
szgupta: Unnecessary - passive suggestions doesn't need this.

app/src/ai/blocklist/controller.rs
```diff
@@ -2369,6 +2369,21 @@ impl BlocklistAIController {
...
+                manager.refresh_grok_tokens_if_needed(byo_allowed, ctx);
```
IsaiahWitzke, line 2383: sorry, i should prob know this alr from the previous review...

do we have a way to unlink grok tokens if you stop paying your grok sub or whatever? we prob dont want to hard block you in a state where we're trying to refresh on every maa request? Also... in that case, could you even use grok models at all? like if your refresh token is stuck in secure storage, are we always going to try to send your grok tokens to the grok backend, and error out if they're invalid?

IsaiahWitzke (approved):

app/src/ai/blocklist/controller.rs
szgupta: No we don't. I think that'd be hard to build? It's like if you BYOK and then you revoke that API key with your AI provider and your requests start failing in Warp. I think that's expected. The error should be clear enough that you can then just go disconnect your grok sub and be on your merry way.

IsaiahWitzke: > sorry, i should prob know this alr from the previous review...

aha yeah, i forgot about this... this is what i was lookgin for . ag

https://github.com/warpdotdev/warp/pull/12028/changes#diff-b11a740af7e4dff6639ae52cd33a6602df9e3fdaede9aaaa94ddc31e63459516R7509-R7515

PR #12517, opened 2026-06-11
aggregate child artifacts in parent info panel

harryalbert:
## Description

For orchestrated conversations, the parent agent's artifacts section in the info panel should hold all artifacts for every child agent (local and cloud), as these are all artifacts of the same overarching conversation.

## Testing

### Screenshots / Videos

https://www.loom.com/share/2a0187e632c34b4d9df86cd4e8be3cae

## Agent Mode

### DIFF
--- a/app/src/ai/agent_conversations_model.rs
+++ b/app/src/ai/agent_conversations_model.rs
@@ -42,6 +42,7 @@ use crate::ai::artifacts::Artifact;
 use crate::ai::blocklist::orchestration_topology::orchestration_aware_conversation_status;
 use crate::ai::blocklist::{
     BlocklistAIHistoryEvent, BlocklistAIHistoryModel, ConversationStatusUpdate,
+    orchestration_topology,
 };
 use crate::ai::cloud_environments::CloudAmbientAgentEnvironment;
 use crate::ai::conversation_navigation::ConversationNavigationData;
@@ -644,6 +645,11 @@ pub struct AgentConversationsModel {
     /// Earliest RTC timestamp received while no list surface was open.
     /// On next `register_view_open`, triggers a single `fetch_tasks_updated_after`.
     dirty_since: Option<DateTime<Utc>>,
+    /// Parent run IDs for which a child-run backfill has already been attempted
+    /// this session. Ensures `ensure_child_tasks_loaded` issues at most one
+    /// `ancestor_run_id` list request per parent, even across repeated panel
+    /// refreshes or after a failed attempt.
+    child_backfill_attempted: HashSet<AmbientAgentTaskId>,
 }
 
 pub enum AgentConversationsModelEvent {
@@ -694,6 +700,7 @@ impl AgentConversationsModel {
                 task_fetch_state: HashMap::new(),
                 rtc_task_refresh_throttle_state: RtcTaskRefreshThrottleState::default(),
                 dirty_since: None,
+                child_backfill_attempted: HashSet::new(),
             };
         }
 
@@ -733,6 +740,7 @@ impl AgentConversationsModel {
             task_fetch_state: HashMap::new(),
             rtc_task_refresh_throttle_state: RtcTaskRefreshThrottleState::default(),
             dirty_since: None,
+            child_backfill_attempted: HashSet::new(),
         };
 
         // Only sync local conversations if we're not in CLI mode. Server-side data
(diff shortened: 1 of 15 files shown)


harryalbert:
* **#12517** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

harryalbert:
@liliwilson tagging you because (IIRC) you originally worked on artifacts and the info panel. Tagging @cephalonaut for vis around orchestration changes

app/src/ai/agent_conversations_model.rs
```diff
@@ -1703,6 +1692,132 @@ impl AgentConversationsModel {
...
+            return;
+        }
+
+        let ai_client = ServerApiProvider::as_ref(ctx).get_ai_client();
+        let ancestor_run_id = parent_task_id.to_string();
+        ctx.spawn(
+            async move {
+                ai_client
+                    .list_ambient_agent_tasks(
+                        INITIAL_TASK_AMOUNT,
```
cephalonaut: I think this limits us to 100 child agents which I know has been exceeded by some users. How hard would it be to paginate? Also, on failure we've already added parent_task_id to requested_child_runs_for which maybe means that we'll never try again?

app/src/ai/blocklist/orchestration_topology.rs
```diff
@@ -6,7 +6,13 @@
 //! the agent-mode usage footer's credit rollup) can walk and order the same
 //! tree without duplicating the logic.
 
+use std::collections::{HashMap, HashSet};
+
+use crate::ai::active_agent_views_model::ConversationOrTaskId;
```
cephalonaut: Maybe this is a backwards dep? Comes from a view model, while this is a lower level utility package?

app/src/ai/blocklist/orchestration_topology.rs
cephalonaut: Don't think this function is really needed anymore and can be inlined.

app/src/ai/blocklist/orchestration_topology.rs
```diff
@@ -246,6 +236,166 @@ pub fn aggregated_orchestrator_status(
...
+    }
+}
+
+/// Returns the local conversation ID represented by the given task, if this task and a
+/// conversation entry both point at the same underlying local run.
+///
+/// We first match using the orchestration agent ID (task ID / run ID under v2), and fall back
+/// to the server conversation token for cases where the task only carries conversation identity
+/// through `conversation_id`.
+pub(crate) fn conversation_id_shadowed_by_task(
```
cephalonaut: The name of this function is a bit perplexing. Is it possible there is something clearer?

cephalonaut (commented):
Thinking about this, I wonder what happens if someone generates 100 PRs in the sidebar. Does it break down?

app/src/ai/agent_conversations_model.rs
```diff
@@ -562,6 +563,10 @@ pub struct AgentConversationsModel {
     /// Earliest RTC timestamp received while no list surface was open.
     /// On next `register_view_open`, triggers a single `fetch_tasks_updated_after`.
     dirty_since: Option<DateTime<Utc>>,
+    /// Parent run IDs whose child runs we have already requested from the
+    /// server this session. Like `task_fetch_state`, this exists purely to
+    /// dedupe requests and ensure we don't try to fetch the same child run repeatedly.
+    requested_child_runs_for: HashSet<AmbientAgentTaskId>,
```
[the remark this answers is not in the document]
harryalbert: good call security bot

app/src/ai/agent_conversations_model/entry.rs
```diff
@@ -277,27 +282,6 @@ impl AgentConversationEntry {
     }
 }
 
-/// Returns the local conversation ID represented by the given task, if this task and a
-/// conversation entry both point at the same underlying local run.
-///
-/// We first match using the orchestration agent ID (task ID / run ID under v2), and fall back
-/// to the server conversation token for cases where the task only carries conversation identity
-/// through `conversation_id`.
-pub(super) fn conversation_id_shadowed_by_task(
```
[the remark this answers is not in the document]
harryalbert: good point

app/src/ai/agent_conversations_model.rs
harryalbert: pagination shouldn't be hard I think.

and good callout for the failure case, I think we can handle failures in a more nuanced way so that we retry potentially transient failures

harryalbert:
> Thinking about this, I wonder what happens if someone generates 100 PRs in the sidebar. Does it break down?

​Good question. I think it's pretty unlikely that someone actually _wants_ to see > 100 artifacts in the sidebar (or would get any value from it). My strawman would be that we should cap at something like 25 artifacts and then show a "+n more..."

app/src/ai/blocklist/orchestration_topology.rs
harryalbert: yeah fair point

app/src/ai/blocklist/orchestration_topology.rs
harryalbert: yeah you're right, not super clear. I think `local_conversation_id_for_task` is much clearer

harryalbert:
here's what that new "+n..." UI looks like:

PR #12532, opened 2026-06-11
Send explicit ConversationHandoff marker on user inputs after cloud-local handoff

vkodithala:
## Description

We were seeing a bunch of issues (example [here](https://warpdev.slack.com/archives/C09769R5GBT/p1780514354474139)) where cloud-local handoff produced errors related to subagent tool call results with unmatched/mismatched tool calls. 

I shipped a [hotfix](https://github.com/warpdotdev/warp-server/pull/11577) to gracefully handle mismatched tool call results, but synced with @vorporeal and noticed that there was a larger issue here, which was that for cloud-local handoff in the middle of a running LRC, we actually continued from _within_ the CLI subagent, which is the wrong behavior since the command it was monitoring was cancelled on handoff.

The right behavior here would've been to, on the server, recognize that the user's query originated from a handoff event, and amend task context to exit out of that subagent before processing the user's query. This necessitates three changes:
1. A new `ConversationHandoff` input variant sent alongside the first user input after handoff, implemented in this [protos PR](https://github.com/warpdotdev/warp-proto-apis/pull/321).
2. Client changes to propagate that input to the server upon a handoff event, implemented here. Tl;dr on my changes here: disambiguate handoff events from regular forked conversations and pass a `ConversationHandoff` input via `RequestParams` alongside the first user input after a handoff event.
3. Server changes to handle `ConversationHandoff` markers in our existing input interceptors. These changes basically check if we're within a CLI subagent, cancel that subagent's tool calls, and then cancel its parent tool call and watched shell command.

This is designed to be extensible; we parse `CloudtoLocal` and `LocaltoCloud` events, pass the former as a marker to the server, and correct task state in the CLI subagent case. In the future, it might be helpful to encode some sort of prompt context around handoff events on the server, but deemed those changes OOS for this fix.

## Testing

Manually tested; here's a [demo](https://www.loom.com/share/df662434600643df9e234e08fc05591a) with a little tl;dr on how this all works.

## Agent Mode
- [X] Warp Agent Mode - This PR was created via Warp's AI Agent Mode


vkodithala:
* **#12532** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12532?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12532?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/agent/api/impl.rs
```diff
@@ -53,6 +53,10 @@ pub async fn generate_multi_agent_output(
         redaction::redact_inputs(&mut params.input);
     }
 
+    if params.pending_conversation_handoff.is_some() {
+        params.input.insert(0, AIAgentInput::ConversationHandoff);
+    }
```
vkodithala, line 59: nit: Probably best to add a comment here explaining why we do this. Maybe something like "if an incoming request indicates conversation handoff, we add a `ConversationHandoff` marker before all other inputs so that taskCtx is corrected properly server-side".

app/src/ai/agent/mod.rs
```diff
@@ -2629,6 +2629,8 @@ pub enum AIAgentInput {
         config: OrchestrationConfig,
         status: OrchestrationConfigStatus,
     },
+
+    ConversationHandoff,
```
vkodithala, line 2720: nit: Also best to add a comment. This seems to be conventional; see other `AIAgentInput` variants. Maybe something like "Sent on the next outbound request after conversation handoff", or something.

app/src/ai/agent/mod.rs
```diff
@@ -2841,6 +2845,21 @@ impl AIAgentInput {
...
+                static_query_type: None,
+                ..
+            } | Self::ActionResult { .. }
+                | Self::MessagesReceivedFromAgents { .. }
+                | Self::EventsFromAgents { .. }
+                | Self::PassiveSuggestionResult { .. }
+                | Self::OrchestrationConfigUpdate { .. }
+                | Self::ConversationHandoff
+        )
+    }
```
vkodithala: question: I worry about directly enumerating which `AIAgentInput` variants can be converted into `UserInput`s, since this may go out of sync with our conversion helpers in `convert_to.rs`. Is there a way we can colocate/better handle to avoid these going out-of-sync?

app/src/ai/blocklist/controller.rs
```diff
@@ -207,6 +207,18 @@ pub struct RequestInput {
...
+        let mut inputs = self.all_inputs();
+        let Some(first_input) = inputs.next() else {
+            return false;
+        };
+
+        if inputs.next().is_none() {
+            first_input.converts_to_user_input()
+        } else {
+            self.all_inputs().any(AIAgentInput::converts_to_user_input)
+        }
```
vkodithala: nit: I don't think we actually need a conditional here.

app/src/ai/blocklist/controller.rs
```diff
@@ -2700,6 +2721,10 @@ impl BlocklistAIController {
                                         history_model.conversation_mut(&conversation_id)
                                     {
                                         conversation.clear_forked_from_server_conversation_token();
+                                        if should_clear_pending_handoff {
+                                            conversation.clear_pending_conversation_handoff();
+                                            conversation.write_updated_conversation_state(ctx);
+                                        }
```
vkodithala, line 2815: nit: We should definitely add a comment here explaining why we do this. IIUC the idea's that when we send a `ConversationHandoff` marker that's processed by the server, we consume it so it isn't also sent in follow-ups.

app/src/workspace/view.rs
```diff
@@ -15138,6 +15145,7 @@ impl Workspace {
                 FORK_PREFIX,
                 true,
                 title_override.as_deref(),
+                None,
```
vkodithala, line 15539: We probably want to enumerate a new variant on PendingConversationHandoff that keeps the local-cloud state, right? Even though this is OOS and isn't explicitly handled in this PR?

app/src/ai/agent/api/impl.rs
vkodithala: Done — added a short comment explaining that the marker is prepended so the server can correct task context before processing the rest of the user input batch.

app/src/ai/agent/mod.rs
vkodithala: Done — added a brief variant comment for ConversationHandoff.

app/src/ai/blocklist/controller.rs
vkodithala: Done — simplified this to use any(...) directly.

app/src/ai/blocklist/controller.rs
vkodithala: Done — added a short comment explaining that the pending marker is consumed once it reaches the server so follow-ups do not resend it.

app/src/workspace/view.rs
vkodithala: Done — added a distinct PendingConversationHandoff::LocalToCloud variant and wired the local-to-cloud fork path to set it.

app/src/ai/agent/mod.rs
vkodithala: Done — agreed this should not duplicate conversion knowledge on AIAgentInput. I moved the eligibility helper into convert_to.rs and made it share the actual convert_input_to_user_input logic on a cloned input, then RequestInput calls that through the api module.

crates/persistence/src/model.rs
```diff
@@ -1031,6 +1031,11 @@ impl<'de> Deserialize<'de> for PersistedAutoexecuteMode {
         })
     }
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Copy, PartialEq, Eq)]
+pub enum PendingConversationHandoff {
+    CloudToLocal,
```
vkodithala, line 1022: This is currently unused (the scope of this PR is to handle the cloud-local handoff case), but thought it might be nice to enumerate in case we wanted to propagate/handle local-cloud handoff in the future. Added a comment explaining this.

app/src/terminal/input/slash_commands/mod.rs
```diff
@@ -1141,21 +1141,9 @@ impl Input {
...
-                // Move any pending attachments out of the source input so they travel with the
-                // initial prompt into the continued local pane and no longer linger on the
-                // original input. Only drain them when a non-empty prompt will actually be sent;
-                // the fork drops attachments when there is no initial prompt, which would
-                // silently discard them.
-                let initial_attachments =
-                    self.maybe_take_attachments_for_initial_prompt(argument, ctx);
-
-                ctx.dispatch_typed_action(&WorkspaceAction::ForkAIConversation {
+                ctx.dispatch_typed_action(&WorkspaceAction::ContinueConversationLocally {
```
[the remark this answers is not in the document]
vkodithala: Yeah, this is valid. Added new `initial_attachments` param to the handoff path in latest, should be fixed.

app/src/workspace/view.rs
```diff
@@ -15138,6 +15145,7 @@ impl Workspace {
                 FORK_PREFIX,
                 true,
                 title_override.as_deref(),
+                Some(PendingConversationHandoff::LocalToCloud),
```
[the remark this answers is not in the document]
vkodithala: Also valid and fixed in latest; set to `None` in the local-to-cloud handoff case (unhandled for now).

app/src/ai/agent/api/convert_to_tests.rs
```diff
@@ -106,6 +106,23 @@ fn git_context_deserializes_legacy_string_pull_request_number() {
...
+    };
+
+    assert!(matches!(
+        user_inputs
+            .inputs
+            .first()
+            .and_then(|input| input.input.as_ref()),
+        Some(api::request::input::user_inputs::user_input::Input::ConversationHandoff(_))
+    ));
+}
```
vkodithala: nit: Do we actually need this? Other tests in this file, i.e.`transfer_control...` below (which tests that a TransferShellCommandControl snapshot converts to a tcr) are non-trivial, whereas our type conversion seems pretty trivial.

app/src/ai/agent/api/impl.rs
```diff
@@ -52,6 +52,11 @@ pub async fn generate_multi_agent_output(
     if params.should_redact_secrets {
         redaction::redact_inputs(&mut params.input);
     }
+    // Place the handoff marker first so the server can correct task context
+    // before processing the rest of the user input batch.
```
vkodithala: We should add a note here around why this is important. Something like, "task state should be corrected (i.e. if we're in a CLI subagent, we should pop back to the primary agent) before processing the user's next input after a handoff event".

app/src/ai/agent/api.rs
```diff
@@ -95,6 +96,7 @@ pub struct RequestParams {
     pub input: Vec<AIAgentInput>,
     pub conversation_token: Option<ServerConversationToken>,
     pub forked_from_conversation_token: Option<ServerConversationToken>,
+    pub pending_conversation_handoff: Option<PendingConversationHandoff>,
```
vkodithala, line 100: We should add a comment explaining what this is, and why we set it on `RequestParams`. From my understanding, upon a handoff event, we should send a 'conversation handoff' marker with the next user-initiated interaction so that the agent's aware of any changes in their machine,status of running shell commands, etc.

app/src/ai/agent/api.rs
```diff
@@ -149,6 +151,7 @@ pub struct ConversationData {
     pub tasks: Vec<warp_multi_agent_api::Task>,
     pub server_conversation_token: Option<ServerConversationToken>,
     pub forked_from_conversation_token: Option<ServerConversationToken>,
+    pub pending_conversation_handoff: Option<PendingConversationHandoff>,
```
vkodithala, line 160: q: Why is it necessary that we initialize a pending handoff state on both `ConversationData` and `RequestParams`?

app/src/ai/blocklist/controller.rs
```diff
@@ -208,6 +208,17 @@ pub struct RequestInput {
...
+        self.all_inputs().any(|input| {
+            matches!(
+                input,
+                AIAgentInput::UserQuery {
+                    static_query_type: None,
+                    ..
+                }
+            )
+        })
+    }
```
vkodithala, line 214: I believe we have a new helper fn `display_query` on `AIAgentInput` that spits back user input if it exists (see AIBlock, which calls it in `has_user_input`). I think there are two things we should consider doing here:
1. In a previous change, I renamed this function from `user_query` to `display_query`. `display_query` feels too similar to the fn `display_user_query` below it - which actually calls this function. We should take a look at all uses to determine its intent and modify its name here.
2. We may want to consider using `display_query` vs. manual iteration over all inputs here?

app/src/ai/blocklist/controller_tests.rs
```diff
@@ -1,22 +1,121 @@
...
+        static_query_type: None,
+        referenced_attachments: HashMap::new(),
+        user_query_mode: UserQueryMode::Normal,
+        running_command: None,
+        intended_agent: None,
+    }
+}
+
+#[test]
+fn request_params_only_snapshots_pending_handoff_for_user_queries() {
```
vkodithala: If we change the semantics on when we send `ConversationHandoff` markers to user _inputs_ and not just user queries, we should probably update this test.

app/src/terminal/input/slash_commands/mod.rs
```diff
@@ -1043,20 +1043,11 @@ impl Input {
                     AgentManagementTelemetryEvent::SlashCommandContinueLocally,
                     ctx
                 );
-
-                // Move any pending attachments out of the source input so they travel with the
-                // initial prompt into the continued local pane and no longer linger on the
-                // original input. Only drain them when a non-empty prompt will actually be sent;
-                // the fork drops attachments when there is no initial prompt, which would
-                // silently discard them.
```
vkodithala, line 1051: Is there a reason we got rid of this comment? Looks like we're still moving pending attachments out of the source & into the pane after fork?

app/src/workspace/view.rs
```diff
@@ -24593,19 +24601,26 @@ impl TypedActionView for Workspace {
...
                     None,
                     false,
                     None,
-                    None,
-                    vec![],
-                    ForkedConversationDestination::SplitPane,
+                    initial_prompt.clone(),
+                    initial_attachments.clone(),
+                    *destination,
+                    Some(PendingConversationHandoff::CloudToLocal),
```
vkodithala, line 24916: Is there a reason we weren't threading initial prompt/attachment context through `ContinueConversationLocally` prior to this change? Feels like that was the wrong behavior, but this seems like a fairly large change so I'm curious.

app/src/ai/agent/mod.rs
```diff
@@ -2815,7 +2823,7 @@ impl AIAgentInput {
     /// (typed queries, slash commands, skill invocations, etc.). Unlike
     /// [`Self::is_user_query`], which strictly matches the `UserQuery` variant,
     /// this returns `Some` for several input variants.
-    pub fn display_query(&self) -> Option<String> {
+    pub fn user_input_text(&self) -> Option<String> {
```
vorporeal: i'm not sure this renaming is entirely semantically correct?  like, the `CloneRepository` and `CreateNewProject` arms don't really feel like "user input text", but they do return `Some(...)`.

crates/persistence/src/model.rs
```diff
@@ -1031,6 +1038,8 @@ pub struct AgentConversationData {
     /// The server conversation ID of the source conversation if this conversation was forked.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub forked_from_server_conversation_token: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub pending_conversation_handoff: Option<PendingConversationHandoff>,
```
vorporeal, line 1042: why is this something we're persisting?  i would expect `pending_conversation_handoff` to be transient state that exists basically just as part of a request?

if you open up a pane to do a cloud-to-local handoff, but don't actually send the initial request, we'll persist that? (and therefore need to restore the fact that we're in a pre-handoff state later?)

skills-lock.json
vorporeal: can we revert the changes to this file?  i have a PR in-flight to automate the creation of PRs to update the lock files, and would like to validate it works properly by having it generate the PR to do this update

app/src/ai/blocklist/controller_tests.rs
```diff
@@ -1,22 +1,135 @@
...
+        static_query_type: None,
+        referenced_attachments: HashMap::new(),
+        user_query_mode: UserQueryMode::Normal,
+        running_command: None,
+        intended_agent: None,
+    }
+}
+
+#[test]
+fn request_params_only_snapshots_pending_handoff_for_user_inputs() {
```
vorporeal, line 71: super nit: this test got inserted in the middle of helper functions.  not a hard blocker, but boy would i love it if agents were better about this.

vorporeal (approved):

skills-lock.json
vkodithala: yeah, totally. i did see your PR, should be super helpful once it goes in. removed in latest

app/src/ai/agent/mod.rs
vkodithala: agree with you here, reverted back to `display_query`. the original idea was that the fn `display_user_query` used `display_query`, which felt a little confusing, but i think 'display' more closely maps to how the string returned from this fn is used

app/src/ai/blocklist/controller_tests.rs
vkodithala: ugh yeah good callout. fixed in latest

crates/persistence/src/model.rs
vkodithala: yeah, this basically just solves for a pretty narrow edge case where the user:

- hands off conv
- restarts app before sending their first query
- we still want to unravel the CLI subagent stack on their next query after quit/restart

hence the persistence, which i thought was pretty cheap since we already send a `conversation_data` JSON blob to SQLite and this would just be an additional field. would love your take on whether this is worth including

vorporeal: goal here is correctness, right?  if this is required for correctness in that user flow, we should include it.

(my question about whether or not to include it wasn't about cost, but cognitive complexity, and wanting to make sure i understood the problem being solved by its inclusion)

vkodithala: yup, goal is correctness, and this makes it so our solution is robust across shutdown/restart events. your understanding is correct :smiley:

PR #12540, opened 2026-06-11
Fix cmd-enter starting new agent conversations

MaggieShan:
## Description
* Context: https://warpdev.slack.com/archives/C08KTPNQN65/p1781038190411049 
  * Noticed a regression where selecting a block then hitting `cmd-enter` did not properly pull the selected blocks into a new agent conversation 
* The bug:
  * Git bisected the bug and found that it stems from https://github.com/warpdotdev/warp/pull/10095 which changed the focus from the input editor to the terminal on block selection if the setting `Preserve input focus on block selection` was disabled 
  * This meant that `cmd-enter` was getting triggered from the terminal context instead of the input context which triggered `StartNewAgentConversation` that didn't go through the logic to include pending context, it called `enter_agent_view` directly 
* The fix:
  * Routes `StartNewAgentConversation` to `TerminalView::enter_agent_view_for_new_conversation` which properly attaches pending context 
  * Added `StartNewAgentConversation` to the cmd-enter keybinding from the terminal context, so we can differentiate between different entry origins and handle accordingly 
    * Specifically `AgentViewEntryOrigin::Keybinding` also needs to check for a double confirmation before starting a new conversation 

## Testing
Before:
https://www.loom.com/share/32ee798cdcb846b28accef42e80869c1 

After:
https://www.loom.com/share/cb50d392472c4ab9870249bab88e0e38

## Agent mode 

### DIFF
--- a/app/src/ai/blocklist/agent_view/agent_view_block.rs
+++ b/app/src/ai/blocklist/agent_view/agent_view_block.rs
@@ -259,7 +259,7 @@ impl View for AgentViewEntryBlock {
             // If the agent_view_block's conversation no longer exists,
             // we assume that it has been deleted.
             return render_deleted_state(
-                self.origin,
+                self.origin.clone(),
                 self.cached_title.clone(),
                 appearance,
                 are_block_dividers_enabled,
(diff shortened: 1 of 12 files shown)


zachbai (approved):
Nice fix!

PR #12541, opened 2026-06-11
Reorganize git repo models

MaggieShan:
## Description
* WISOTT 
* Builds on https://github.com/warpdotdev/warp/pull/12456 + split of https://github.com/warpdotdev/warp/pull/12498 
* Updates the existing models `GitHubRepoModel` and `GitRepoStatusModel` such that each model is moved to its own folder with a wrapper and local variant (remote variant to be added) 
* Renames `GitStatusUpdateModel` to `GitRepoModels` and moves the model to its own file 

## Linked issue 
* [APP-4590: remote git chips](https://linear.app/warpdotdev/issue/APP-4590/improve-remote-git-related-chip-experience)
* [APP-4560: remote pr chip](https://linear.app/warpdotdev/issue/APP-4560/code-feature-support-pr-chip)

## Testing
Should be a no-op
- [x] I have manually tested my changes locally with `./script/run`

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

PR #12542, opened 2026-06-11
Add remote support for git chips

MaggieShan:
## Description
* Adds support for various git/github-backed features like prompt chips and git operations dialog over a remote environment 
* Builds on https://github.com/warpdotdev/warp/pull/12541 + split of https://github.com/warpdotdev/warp/pull/12498
* Adds remote variants of each model and updates wrappers to support a repo backed by `LocalOrRemotePath`
*  Remote repositories now also source PR information from the `GitHubRepoModel`, removing the need for `DiffStateModel` to manage PR-specific state - deprecated the field and associated methods 
* Introduces new proto calls to subscribe to and fetch github info, as well as subscribe to and receive git status updates 

## Linked issue 
* [APP-4590: remote git chips](https://linear.app/warpdotdev/issue/APP-4590/improve-remote-git-related-chip-experience)
* [APP-4560: remote pr chip](https://linear.app/warpdotdev/issue/APP-4560/code-feature-support-pr-chip)

## Testing 
https://www.loom.com/share/2a3a8ddbc83b400e84c0c2d60586edf6

## Agent Mode

CHANGELOG-NEW-FEATURE: Adds support for git chips over remote environments

### DIFF
--- a/app/src/code_review/git_repo_model/remote.rs
+++ b/app/src/code_review/git_repo_model/remote.rs
@@ -1,0 +1,88 @@
+use remote_server::manager::{RemoteServerManager, RemoteServerManagerEvent};
+use warp_util::remote_path::RemotePath;
+use warpui::{Entity, ModelContext, SingletonEntity};
+
+use super::{GitRepoStatusEvent, GitStatusMetadata};
+use crate::remote_server::proto;
+
+/// Client-side per-repo git status for a repository on an SSH host.
+///
+/// Presents the same read surface as [`super::LocalGitRepoStatusModel`] and emits the
+/// same [`GitRepoStatusEvent`]s so the unified [`super::GitRepoStatusModel`] can
+/// substitute it transparently (mirrors `RemoteDiffStateModel`).
+///
+/// Holds the latest status for its `(host_id, repo_path)`. On construction
+/// (and again on reconnect) it sends an `UpdateGitStatus` notification
+/// requesting the daemon to push the current snapshot; live watcher updates
+/// then arrive as `GitStatusPush` messages filtered by `(host_id, repo_path)`.
+/// `HostDisconnected` preserves stale data.
+pub struct RemoteGitRepoStatusModel {
+    remote_path: RemotePath,
+    metadata: Option<GitStatusMetadata>,
+}
+
+impl Entity for RemoteGitRepoStatusModel {
+    type Event = GitRepoStatusEvent;
+}
+
+impl RemoteGitRepoStatusModel {
+    pub fn new(remote_path: RemotePath, ctx: &mut ModelContext<Self>) -> Self {
+        let mgr = RemoteServerManager::handle(ctx);
+        ctx.subscribe_to_model(&mgr, Self::handle_manager_event);
+        let model = Self {
+            remote_path,
+            metadata: None,
+        };
+        model.request_snapshot(ctx);
+        model
+    }
+
+    fn handle_manager_event(
+        &mut self,
+        event: &RemoteServerManagerEvent,
+        ctx: &mut ModelContext<Self>,
+    ) {
+        match event {
+            RemoteServerManagerEvent::GitStatusPushReceived {
+                host_id,
+                repo_path,
+                metadata,
+            } if host_id == &self.remote_path.host_id && repo_path == &self.remote_path.path => {
+                self.apply_push(metadata, ctx);
+            }
+            RemoteServerManagerEvent::HostConnected { host_id }
+                if host_id == &self.remote_path.host_id =>
+            {
+                self.request_snapshot(ctx);
+            }
+            _ => {}
+        }
+    }
+
+    pub(super) fn request_snapshot(&self, ctx: &mut ModelContext<Self>) {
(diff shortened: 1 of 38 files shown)


app/src/code_review/git_repo_model/remote.rs
```diff
@@ -0,0 +1,88 @@
...
+use warpui::{Entity, ModelContext, SingletonEntity};
+
+use super::{GitRepoStatusEvent, GitStatusMetadata};
+use crate::remote_server::proto;
+
+/// Client-side per-repo git status for a repository on an SSH host.
+///
+/// Presents the same read surface as [`super::LocalGitRepoStatusModel`] and emits the
+/// same [`GitRepoStatusEvent`]s so the unified [`super::GitRepoStatusModel`] can
+/// substitute it transparently (mirrors `RemoteDiffStateModel`).
```
moirahuang: can we pull logic out to be shared with `RemoteGitRepoStatusModel`, `RemoteDiffStateModel`, and `RemoteGitHubRepoModel`?

specs/remote-git-repo-status/TECH.md
```diff
@@ -0,0 +1,65 @@
...
+Git status is shared per-repo model state and is small enough to remain a notification-triggered push stream:
+- `UpdateGitStatus { repo_path }` is a `Notification`. It asks the daemon to subscribe/create the per-repo git-status model and push the current `GitStatusPush` snapshot when metadata is available. It has no request id, no response, and is best-effort.
+- `RemoteGitRepoStatusModel` sends that notification on construction and again on `HostConnected`; live watcher changes arrive later as `GitStatusPush` messages filtered by `(host_id, repo_path)`.
+GitHub info is also shared cached state, but explicit refreshes use host-scoped request/response so the client can track request completion and surface PR-refresh loading state:
+- `GetGitHubPrInfoRequest { repo_path }` is a `HostScopedRequest`. The daemon validates the path, subscribes/creates the per-repo `GitHubRepoModel` (whose own lifecycle drives `GitHubPrInfoPush` broadcasts), runs `git::get_pr_for_branch` for the requesting client, and returns `GetGitHubPrInfoResponse`. It does not separately trigger `refresh_pr_info`, which would run a redundant `gh pr view`.
```
moirahuang: i wonder if there's something we can consolidate here so we have one source of truth for `gh pr view`? specifically, it feels like we should be able to fetch once and have that inject the info all at once instead of having to fetch multiple times from different places?

crates/remote_server/src/manager.rs
```diff
@@ -1195,6 +1200,71 @@ impl HostRequestHandle {
...
     }
 
+    /// Fetches PR info for the current branch on the remote host. `Ok(None)`
+    /// means the lookup completed and the branch has no PR.
+    pub async fn get_github_pr_info(
+        &self,
+        repo_path: &StandardizedPath,
+    ) -> Result<Option<crate::proto::PrInfo>, HostRequestError> {
+        let msg = self
+            .send(crate::proto::host_scoped_request::Message::GetGithubPrInfo(
```
moirahuang: nit: do we maybe want to import these at the top instead of inlining?

crates/remote_server/src/manager.rs
```diff
@@ -3256,29 +3327,68 @@ impl RemoteServerManager {
...
     }
 
-    /// Fetches the committed branch files (`merge_base(HEAD, main)..HEAD`) for
-    /// the current branch on the remote host and emits
-    /// `GetCommittedBranchFilesResponse` with the result.
-    pub fn git_get_committed_branch_files(
+    /// Sends an `UpdateGitStatus` notification (fire-and-forget) to the
+    /// remote server for the given host.
+    pub fn update_git_status(&self, host_id: HostId, repo_path: &StandardizedPath) {
+        if let Some(client) = self.client_for_host(&host_id) {
```
moirahuang, line 3263: is it possible to have 2 terminal panes in 2 different repos in a client? do we need to be specific about which exactly pane we connect to instead of just generally returning to the client?

moirahuang (changes requested):

moirahuang (approved):

crates/remote_server/src/manager.rs
MaggieShan: Yes it's possible but we don't need to be specific about which pane we connect to since the info comes back as a push that is matched against repo path and host id

specs/remote-git-repo-status/TECH.md
MaggieShan: Good point - the server model should also be calling the `LocalGithubRepoModel` for its pr/repo info, I've updated it such that the request triggers a pr info refresh from the model + relies solely on the push events so that we have one source of truth.

app/src/code_review/git_repo_model/remote.rs
MaggieShan: Can you expand on this? The "shared" logic between the models is mainly wrt subscribing to and handling the manager events - but the actual handling logic and which events each model cares about is still different. 

I pulled the shared `host_id` and `path` checks into `RemotePath` and updated the comment to be more precise but let me know if there was something else you were thinking about.

PR #12555, opened 2026-06-12
cancel in-progress stuff when cmd-k a conversation

### LINKED ISSUE #12483: cmd-k to clear doesn't clear previous conversation context
### Pre-submit Checks

### Describe the bug

https://warpdev.slack.com/archives/C08KTPNQN65/p1780640536267589

### To reproduce

Press cmd-k while in an agent conversation. the new conversation seems to continue the old conversation

### Expected behavior

You get a new conversation with fresh context

https://www.loom.com/share/ffc2b36f2e5447beb021a68cc1e6c065

[log_excerpt.txt](https://github.com/user-attachments/files/28817585/log_excerpt.txt)

### Operating system (OS)

macOS

?

### Shell Version

_No response_

### Current Warp version

_No response_

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

acarl005:
## Description

See here: https://warpdev.slack.com/archives/C08KTPNQN65/p1780700091713279?thread_ts=1780640536.267589&cid=C08KTPNQN65

The bug happens with the following scenario:

1. Ask the agent to run a command, e.g. `sleep 10`. Let's call this conversation "A".
2. As the agent is responding but *before* `sleep 10` actually runs, press cmd-k to clear. This starts a new conversation, conversation "B".
3. The `sleep 10` runs in "B".
4. After `sleep 10` completes, the block goes back to "A".

The `sleep 10` command should never appear in "B" since it was requested from "A".

This PR has a 2-layer approach to the fix. Either is sufficient for this bug, but implementing both is prudent for robustness.

1. Cancel in-progress conversation when cmd-k is pressed. It seems that we assumed once a conversation is no longer active in the view, it shouldn't be doing things like executing tool calls anymore. Therefore, we should cancel it if we are hiding the conversation in order to start a new one.
2. Encode that invariant for executed command by checking, before we execute them, if the active conversation is _not_ the same conversation that it belongs to. If not, cancel the command.

## Linked Issue

(partly) fixes #12483

## Testing

On this branch, the same repro steps will cancel conversation "A" as soon as cmd-k is pressed.

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -18644,6 +18644,12 @@ impl TerminalView {
                 .block_list()
                 .active_block()
                 .is_agent_monitoring();
+            let is_agent_driving_command = self
+                .model
+                .lock()
+                .block_list()
+                .active_block()
+                .is_agent_driving_command();
 
             // If there isn't an active long running block, then "clear buffer" just starts a new convo.
             if !active_block_is_long_running {
@@ -18673,7 +18679,7 @@ impl TerminalView {
                     ctx,
                 );
                 true
-            } else if is_agent_monitoring {
+            } else if is_agent_monitoring || is_agent_driving_command {
                 // Otherwise, if the agent is monitoring this long-running block,
                 // then clear just that block and leave the rest of the blocklist in tact.
                 self.model.lock().clear_screen(ClearMode::ActiveBlock);
@@ -18710,8 +18716,14 @@ impl TerminalView {
             .block_list()
             .active_block()
             .is_agent_monitoring();
+        let is_agent_driving_command = self
+            .model
+            .lock()
+            .block_list()
+            .active_block()
+            .is_agent_driving_command();
 
-        if is_agent_monitoring {
+        if is_agent_monitoring || is_agent_driving_command {
             return;
         }
 
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -18632,27 +18632,18 @@ impl TerminalView {
             true
         } else {
             // Otherwise, there are some visible blocks and we need to clear stuff.
-            let active_block_is_long_running = self
-                .model
-                .lock()
-                .block_list()
-                .active_block()
-                .is_active_and_long_running();
-            let is_agent_monitoring = self
-                .model
-                .lock()
-                .block_list()
-                .active_block()
-                .is_agent_monitoring();
-            let is_agent_driving_command = self
-                .model
-                .lock()
-                .block_list()
-                .active_block()
-                .is_agent_driving_command();
+            let (is_long_running, is_agent_monitoring, is_agent_driving_command) = {
+                let model = self.model.lock();
+                let active_block = model.block_list().active_block();
+                (
+                    active_block.is_active_and_long_running(),
+                    active_block.is_agent_monitoring(),
+                    active_block.is_agent_driving_command(),
+                )
+            };
 
             // If there isn't an active long running block, then "clear buffer" just starts a new convo.
-            if !active_block_is_long_running {
+            if !is_long_running {
                 // Cancel any in-progress work in the conversation being left.
                 if let Some(conversation_id) = self
                     .agent_view_controller
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -18647,6 +18647,26 @@ impl TerminalView {
 
             // If there isn't an active long running block, then "clear buffer" just starts a new convo.
             if !active_block_is_long_running {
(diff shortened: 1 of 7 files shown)


MaggieShan (commented):
Had a few questions: 
* Was testing this and can confirm the case for triggering `cmd-k` while the agent is still generating the command is fixed, but I can still get into a weird state if I `cmd-k` after the command is running https://www.loom.com/share/0377da7646b046f09530a6b91591108a 
* Is there a way we can add tests to prevent this from regressing? 
* Product-wise: I wonder if it'd be a clearer to users if we require a double confirmation before starting a new conversation with `cmd-k` (similarly to what's done for `cmd-enter` when already in a conversation) - for users used to `cmd-k` simply clearing the buffer, it may not be as clear that this would also start a new conversation  
  Screenshot 2026-06-16 at 2 16 52 PM

app/src/terminal/view.rs
```diff
@@ -18644,19 +18644,69 @@ impl TerminalView {
...
             // If there isn't an active long running block, then "clear buffer" just starts a new convo.
             if !active_block_is_long_running {
+                // Cancel any in-progress work in the conversation being left.
+                if let Some(conversation_id) = self
+                    .agent_view_controller
+                    .as_ref(ctx)
+                    .agent_view_state()
+                    .active_conversation_id()
+                {
+                    let is_in_progress = BlocklistAIHistoryModel::as_ref(ctx)
```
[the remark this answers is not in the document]
acarl005: Actually that case behaves just fine. "blocked" is a fine state. Users can go back to that conversation and approve if they want.

acarl005:
@MaggieShan 

> but I can still get into a weird state if I cmd-k after the command is running

Fixed! Needed to check for an additional state. I made sure this case matches the behavior when the command is running and the agent is monitoring it, i.e. it clears the running command's grid.

> Is there a way we can add tests to prevent this from regressing?

Added some unit tests, needed to add some new mocks for this.

> Product-wise: I wonder if it'd be a clearer to users if we require a double confirmation before starting a new conversation with `cmd-k`

Ah that's a good Q. My intuition would be not to require double confirmation. We do require it for cmd-enter and ctrl-c which I think makes sense. ctrl-c does a lot of other stuff, e.g. exiting a long-running command, exiting insert mode in Vim keybindings, clearing the input editor, etc. If we didn't require a double-press, it would be super easy to accidentally exit your conversation with ctrl-c. cmd-k is different b/c it isn't as overloaded with other functions. Also, cmd-k is already a pretty destructive action in a normal terminal + shell setup, e.g. in iTerm it deletes the whole screen+scrollback. If users are pressing cmd-k I think we have more confidence in their intent, and the way we handle it in the agent view, starting a new convo (or clearing the grid for the active running command), maps pretty cleanly to the idea of clearing in a normal terminal.

That said, I don't think it's critical to make that change in this PR. I think this change is strictly better than the existing functionality. I'm very open to adding that if people find this too destructive. What do you think?

MaggieShan:
> Ah that's a good Q. My intuition would be not to require double confirmation. We do require it for cmd-enter and ctrl-c which I think makes sense. ctrl-c does a lot of other stuff, e.g. exiting a long-running command, exiting insert mode in Vim keybindings, clearing the input editor, etc. If we didn't require a double-press, it would be super easy to accidentally exit your conversation with ctrl-c. cmd-k is different b/c it isn't as overloaded with other functions. Also, cmd-k is already a pretty destructive action in a normal terminal + shell setup, e.g. in iTerm it deletes the whole screen+scrollback. If users are pressing cmd-k I think we have more confidence in their intent, and the way we handle it in the agent view, starting a new convo (or clearing the grid for the active running command), maps pretty cleanly to the idea of clearing in a normal terminal.
> 
> That said, I don't think it's critical to make that change in this PR. I think this change is strictly better than the existing functionality. I'm very open to adding that if people find this too destructive. What do you think?

That's a good point on user intent with `cmd-k` compared to `ctrl-c` - agreed in that case for keeping the current product behaviour! Also agreed any changes related to this point would be out of scope for this pr 👍

app/src/terminal/view.rs
```diff
@@ -18644,19 +18644,69 @@ impl TerminalView {
                 .block_list()
                 .active_block()
                 .is_agent_monitoring();
+            let is_agent_driving_command = self
```
MaggieShan: nit: can we combine the access to the model and active lock so that we're not acquiring it multiple times?

MaggieShan (approved):
Thanks for iterating - approving to unblock!

app/src/terminal/view.rs
acarl005: borrow checker made it a tad awk but done!

PR #12573, opened 2026-06-12
Drain queued prompts per conversation when a sibling block masks turn end

harryalbert:
## Description

@vkodithala ran into an issue where queued prompts sometimes weren't being sent in sub-agent conversations. This issue is semi-flaky (no consistent repro), but looking at Varoon's logs with the agent I think I pinpointed at least one issue.

The problem was that, when a conversation's turn ends, `handle_ai_controller_event` receives `BlocklistAIControllerEvent::FinishedReceivingOutput` and decides whether to fire conversation-finished side effects. This decision relied on global pane helpers, specifically `active_ai_block` and `last_ai_block`. However, an orchestration pane hosts the lead and local child conversations in a single TerminalView, so when one conversation's turn ends while a sibling is mid-turn, those helpers observe the sibling's still-streaming block instead. In this edge-case, the finish reason comes back None, so we don't drain the finished event, meaning we don't send the queued prompt.

The fix is, when the pane-global path yields no finish reason (and no subagent is active), fall back to a new conversation-scoped lookup (finish_reason_for_conversation) that finds the finished conversation's own most recent block, and drain its queued prompts from there.

## Testing
- Added `finish_reason_is_scoped_to_the_finished_conversation`, which builds a pane with a finished block and a later streaming sibling block and asserts the scoped lookup reports `Complete` for the finished conversation, `None` for the streaming one, and `None` for an unknown conversation.

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/15494e48-16a7-4830-9e12-09f3e5cc1152_
_Run: https://oz.staging.warp.dev/runs/019ebc23-56cd-7e96-8928-b6d51eeb3017_

### DIFF
--- a/app/src/ai/blocklist/block/model.rs
+++ b/app/src/ai/blocklist/block/model.rs
@@ -233,6 +233,7 @@ pub mod testing {
         input: Vec<AIAgentInput>,
         output: Shared<AIAgentOutput>,
         model_id: LLMId,
+        is_streaming: bool,
     }
 
     impl FakeAIBlockModel {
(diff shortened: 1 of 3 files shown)


harryalbert:
* **#12573** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12573?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12573?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

vkodithala (approved):
this makes sense to me and seems like the right behavior in any case. thanks for the fix and great explanation/comments!

Issue #12329, opened 2026-06-08
Support reliable CLI agent notifications over SSH and tmux


eclipse1228:
# Support reliable CLI agent notifications over SSH + tmux

## Summary

Warp's third-party CLI agent notification UI works for local Codex and Claude Code sessions, but the same agent lifecycle events are unreliable or missing when the agent runs inside an SSH session and tmux. This makes it hard to identify which tab needs attention when several remote agent sessions are running.

This affects the user-visible notification pipeline:

- top tab status and unread badge
- vertical tab status and unread badge
- Warp notification center / native notification
- running vs completed/waiting state transitions

## User Problem

I usually run agent CLIs inside remote SSH sessions and tmux. When multiple Warp tabs are open, I need Warp to show which tab is still running and which tab has completed or needs input.

Without this, I have to manually inspect every remote tmux pane to find the agent that finished.

## Environment

Local:

- OS: macOS 15.6.1 (24G90), arm64
- Warp repo base commit tested locally: `1c2d4ccb`
- tmux: 3.6a
- Codex CLI: 0.137.0
- Claude Code: 2.1.156

Remote:

- SSH session launched from Warp into a remote shell
- shell: zsh in the observed test session
- tmux: control-mode / warpified tmux session
- SSH mode: Warp SSH / warpified remote session

## Reproduction

1. Build and run Warp OSS locally.
2. Open a new Warp tab.
3. SSH into a remote host.
4. Start a tmux session.
5. Run Claude Code or Codex inside the tmux session.
6. Submit a prompt and wait for the agent to complete.
7. Observe top tabs, vertical tabs, and notification center.

Expected:

- Prompt submit marks the corresponding tab as running.
- Agent completion marks that same tab as unread/complete.
- Warp shows a notification.
- Opening/focusing the completed tab clears the unread completion badge.

Actual:

- In remote SSH + tmux sessions, completion notifications and tab status can be missing.
- Depending on the agent and transport path, running state can be visible while completion is not, or neither running nor completion state appears.

## What I Verified Locally

Manual validation after local patches:

- Local Codex: running and completion tab state works.
- Local Claude Code: completion notification works.
- SSH + tmux Codex: running/completion notification works with `warp-agent-notify` helper env.
- SSH + tmux Claude Code: completion notification works when Claude Code emits raw OSC 777 via `terminalSequence`; plugin-side tmux DCS wrapping must only be used for direct `/dev/tty` fallback.
[...]

Issue #11925, opened 2026-05-30
SSH session disconnects when agent resumes command execution after takeover in SSH terminal


mario-meng:
## Summary
SSH sessions established via manual `ssh` command in Warp terminal consistently disconnect when re-entering agent mode after a takeover. The disconnection occurs at the moment the agent begins executing commands. This is a regression that appeared in recent versions.

## Problem
When using Warp's agent mode over an active SSH session, taking over from the agent and then re-entering agent mode causes the SSH connection to drop. The agent reports "The SSH session has exited" and can no longer run commands. This blocks any continued agent-assisted work over SSH without re-establishing the connection, and the issue recurs on every subsequent attempt.

## Expected behavior
The SSH session should remain connected when switching between agent mode and manual terminal control. Re-entering agent mode after a takeover should allow the agent to continue executing commands in the existing SSH session.

## Actual behavior
The SSH session exits immediately when the agent begins executing commands after a takeover. The agent reports it can no longer run commands because the SSH session has exited. Reconnecting and re-entering agent mode produces the same result every time.

## Reproduction steps
1. Open Warp terminal.
2. Run `ssh user@host` to connect to a remote server.
3. Enter agent mode and let the agent execute commands over SSH.
4. Take over from the agent (exit agent mode).
5. Re-enter agent mode.
6. Observe: SSH session disconnects when the agent attempts to execute a command.

## Artifacts
None attached

## Warp version
v0.2026.05.20.09.21.stable_03

## Operating system
macOS 26.4.1

mario-meng:
https://github.com/user-attachments/assets/539a3a90-80b0-4a1b-aa94-2894c9ec2758  
At the end, the entire screen froze — pressing any key had no response, and there was no output or prompt at all. The only option was to close it.

david-engelmann:
The takeover-handoff failure mode is more specific than a generic SSH drop, but it's adjacent to session-resilience work I've been doing on a fork: a `ManagedSession` state machine that decouples "session is alive" from "warpify state". If the disconnect is happening because the agent-mode re-entry tears down a connection that the manual session expected to outlive the mode switch, that's exactly what the state machine is designed to address.

Not trying to preempt @oz-for-oss' env questions to @mario-meng (those are still load-bearing — root cause could be remote shell config, not the handoff path). But if/when the env details narrow this down to a Warp-side handoff bug, happy to share the `ManagedSession` shape and discuss whether it's the right fit.

PR #12574, opened 2026-06-12
feat(session-journal): new crate for resequencable event ring + JSONL mirror (R3.1 foundation)

david-engelmann:
## Description

Foundation for SSH session resilience — a per-session ring buffer of monotonically-sequenced events with an optional on-disk JSONL mirror. When a connection drops mid-task, a reconnecting client asks `replay_since(Some(N))` to catch up to the live tail instead of starting mid-stream. When the requested cursor is older than the oldest surviving entry, the snapshot reports `replay_gap` so the caller can fall back to a full reload.

Refs [#11925](https://github.com/warpdotdev/warp/issues/11925) (SSH session disconnects when agent resumes), [#12329](https://github.com/warpdotdev/warp/issues/12329) (CLI agent notifications over SSH+tmux). This PR ships the **journal only** — the `attach(since_seq)` RPC integration with `RemoteServerManager` is a follow-up phase. Surfacing now so the data structure can be reviewed independently of the SSH-transport wiring.

The design is a Warp port of the production journal shipping in [helmor](https://github.com/david-engelmann/helmor) (~5 months of soak time across daemon restarts + reattach storms). I prepared a roadmap describing the full chain at [warp-roadmap-r3](https://github.com/david-engelmann/warp/blob/david/r3.1-session-journal/CHANGELOG.md) — happy to break out a separate design discussion if helpful.

## Public surface

```rust
let mut journal: EventJournal<MyEvent> = EventJournal::default();
let seq: u64 = journal.append(event);

// On reattach:
let snapshot = journal.replay_since(client_cursor); // Option<u64>
// snapshot.entries: Vec<JournalEntry<MyEvent>>
// snapshot.replay_gap: Option<u64>  ← Some(earliest_available) when the cursor is too old
// snapshot.head_seq: u64
```

Generic over `P` so consumers can plumb whatever event type (terminal output chunks, agent deltas, structured notifications) without the crate baking in domain knowledge.

Optional disk mirror via `JournalDiskWriter<P>` (JSONL, flushed per-line) lets the journal survive process restarts. Read back with `read_journal_entries(path)`.

## Design choices

- Per-journal seq starts at 1, monotonic, never resets. `head_seq() == 0` before the first append.
- Capacity-bounded eviction (default 1024 entries). Oldest dropped when ring is full. **Eviction does NOT decrement the seq counter** — that would hand out duplicate seqs on the next append.
- Disk write failures detach the writer (logged once) but don't abort the in-memory append. Durability for one event is less important than dropping a live event the consumer is waiting on.
- Malformed JSONL lines on replay are logged + skipped, not fatal.
- Zero capacity is clamped to 1 so the contract stays simple.

## Testing

- 18 unit tests covering:
  - **Sequence behaviour:** monotonic from 1; `head_seq=0` before first append; head advances; empty-state predicates.
  - **Ring/eviction:** zero-capacity clamps to 1; eviction drops oldest; eviction does NOT reset the seq counter (regression-protector for the "after eviction, next append gets seq 1 again" footgun).
  - **`replay_since` permutations:** `None`→full ring; in-range cursor→skips already-seen; cursor at head→empty no-gap; cursor below evicted floor→reports `replay_gap`; cursor at/above floor→no gap; empty ring→no entries, no gap.
  - **Disk:** persists as JSONL; replay roundtrip recovers entries; malformed lines skipped; empty lines skipped; `append_to` preserves existing file.

```
$ cargo test -p warp_session_journal
running 18 tests
... (all 18)
test result: ok. 18 passed; 0 failed
```

- `cargo fmt -p warp_session_journal --check` clean
- `cargo clippy -p warp_session_journal --tests --no-deps -- -D warnings` clean

## What this does NOT do (deferred)

- No consumer yet — this PR is the journal crate in isolation, ready to be wired into `RemoteServerManager` in a follow-up.
- No `attach(since_seq)` RPC — that's the next phase. Adds an `Attach` variant to the remote protocol, calls `replay_since` on the server, ships entries back over the existing transport.
- No auto-reconnect / liveness loop — separate phase building on the journal + attach surface.

## Server API dependencies

N/A — pure client-side crate, no protocol or server changes.

## Agent Mode

(none — internal plumbing crate with no user-facing change in this PR)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15280,6 +15280,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "warp_session_journal"
+version = "0.1.0"
+dependencies = [
+ "log",
+ "serde",
+ "serde_json",
+ "tempfile",
+]
+
+[[package]]
 name = "warp_terminal"
 version = "0.0.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ warp_ripgrep = { path = "crates/warp_ripgrep" }
 warp_search_core = { path = "crates/warp_search_core" }
 warp_server_auth = { path = "crates/warp_server_auth" }
 warp_server_client = { path = "crates/warp_server_client" }
+warp_session_journal = { path = "crates/warp_session_journal" }
 warp_terminal = { path = "crates/warp_terminal" }
 warp_util = { path = "crates/warp_util" }
 warp_web_event_bus = { path = "crates/warp_web_event_bus" }
--- a/crates/warp_session_journal/Cargo.toml
+++ b/crates/warp_session_journal/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "warp_session_journal"
+version = "0.1.0"
+edition = "2021"
+authors = ["Warp Team <dev@warp.dev>"]
+publish.workspace = true
+license.workspace = true
+
+[dependencies]
+log.workspace = true
+serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
+
+[dev-dependencies]
+tempfile.workspace = true
--- a/crates/warp_session_journal/src/lib.rs
+++ b/crates/warp_session_journal/src/lib.rs
@@ -1,0 +1,344 @@
+//! Per-session ring buffer of resequencable events with an optional
+//! on-disk JSONL mirror.
+//!
+//! Foundation for SSH session resilience: when a connection drops
+//! between a long-running task and the desktop client, the client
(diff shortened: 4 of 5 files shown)


david-engelmann:
Closing this one — filed prematurely. The anchor issues ([#11925](https://github.com/warpdotdev/warp/issues/11925) and [#12329](https://github.com/warpdotdev/warp/issues/12329)) don't carry \`ready-to-implement\` yet, and I shouldn't be putting speculative foundation crates into the review queue before they're scoped.

Code stays on \`david/r3.1-session-journal\` so it's preserved if/when the journal+\`attach(since_seq)\` design gets a maintainer green-light. Apologies for the noise.

PR #12581, opened 2026-06-12
[REV-1599] Add the GEAP credential engine: WIF mint, refresh lifecycle, and request attachment

JaidenRatti:
## Description

PR 2 of the **REV-1599 — GEAP (Gemini Enterprise Agent Platform) BYOLLM** client work, implementing the credential engine from the client tech spec (`specs/REV-1599/TECH.md`). Builds on #12537. 

In this PR, the client mints a short-lived Google Cloud access token through Workload Identity Federation, rooted in the user's existing Warp session (meaning the client doesn't have to do anything special to enable GEAP), keeps it fresh, and attaches it to eligible agent requests.

### What's here

**`crates/ai/src/geap_credentials.rs` (new)**
- `GeapCredentials` (private fields; the only token egress is the `From` conversion into the wire type), `needs_refresh()` with a 5-minute lead window, and `access_token_for_request()` that never consults expiry (always-send: Google stays the authority on validity and tells us if the token is valid or not). 
- `GeapMintBinding` / `GeapRequestGate` — a token is only attachable for the exact `(user, audience, SA)` it was minted for; sign-out/account switches and admin config changes invalidate it at attach time.
- `GeapCredentialsState`: `Missing | Disabled | Refreshing { previous } | Loaded | Failed` — `Refreshing` carries the previous token so requests keep authenticating during the ~1-3s re-mint (tokens stay until replaced).

**`crates/ai/src/api_keys.rs`** — `geap_credentials_state` on `ApiKeyManager` with a deduped setter, and the attach matrix in `api_keys_for_request` (now takes the expected `GeapRequestGate`, keeping the function a pure `&self` read).

**`app/src/ai/geap_credentials.rs` (new)** 
- The 3-leg mint: fresh Warp OIDC JWT per mint (`IssueTaskIdentityToken`) → STS token exchange (RFC 8693) → optional SA impersonation (`generateAccessToken`); skipped entirely when no `gcpSaEmail` is configured (the federated token is used directly).
- The refresh guard all triggers funnel through: gate-off → `Disabled` (tokens never retained while disabled), unconfigured → `Missing`, one mint at a time, skip-if-valid under the mint binding.
-One-shot timer at `expires_at − 5min` (clamped to a 60s floor so there's no hot mint loops), plus a request-time safety net that re-arms a parked chain — retries are demand-driven, never an unbounded background loop.
- Failure policy: a failed background re-mint keeps the previous token and parks; `Failed { per-leg message }` is reserved for first/forced mints. Per-leg errors (`MintIdentityToken` / `ExchangeToken` / `ImpersonateServiceAccount`) carry actionable copy, capped detail, and never token material.

All cfg'd out of wasm like `aws_credentials` (web clients simply never carry a token; the server's no-token policy applies.

### Follow-up PRs
- Settings > Warp Agent toggle row (with "Enabled by your workspace" under `ENFORCE`). A manual **Refresh credentials** button (`force_refresh_geap_credentials` is already in place)
- The inline credential error view with auto-remint + Retry consuming the per-leg `Failed` state and the server's `InvalidGeminiEnterpriseCredentialsError`.

## Linked Issue

Linear: [REV-1599](https://linear.app/warpdotdev/issue/REV-1599)

- [x] Where appropriate, screenshots or a short video of the implementation are included below (not applicable — no UI in this PR).

## Testing

- `cargo nextest run -p ai -p warp -E 'test(geap) or test(api_keys_for_request) or test(workspaces::)'` — 67 passed post-rebase (39 GEAP-specific)
- `cargo clippy -p ai -p warp --all-targets --tests -- -D warnings` — clean
- `./script/format` — clean

Unit coverage: the full attach matrix (binding mismatch per field, expired-token-still-attached, previous-served-while-refreshing, omission for `Missing`/`Disabled`/`Failed`/first-mint); pure helpers (gate trimming/normalization, STS `expires_in` → JWT-expiry fallback, RFC 3339 impersonation expiry incl. invalid-timestamp rejection, timer-delay floor, error-detail truncation, per-leg error copy); and `App::test` harness tests for the refresh guard and safety net (gate-off drops tokens, skip-if-valid, in-flight no-op even when forced, expired → `Refreshing { previous: Some }`, binding-mismatch re-mint, parked-chain re-arm, pure no-op when gate off).

Manual E2E against a real GCP project (`warp-geap-test-2026`) + local warp-server. When the admin configures the correct information (simulated via graphiql for now), GEAP supported models correctly route to GEAP. When the admin configures an incorrect project id, location, the error message propagates to the client. 

## Agent Mode


app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,643 @@
...
+    }
+    // The previous token (if any) keeps serving requests while the re-mint is
+    // in flight — tokens stay until replaced, so a request landing in the
+    // ~1-3s mint window still authenticates.
+    let previous = match manager.geap_credentials_state() {
+        GeapCredentialsState::Loaded {
+            credentials,
+            minted_for,
+            ..
+        } => Some((credentials.clone(), minted_for.clone())),
```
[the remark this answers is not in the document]
JaidenRatti: Fixed, `previous` now only captured when its binding matches the binding of the mint that's being started. Test updated as well

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,643 @@
...
+fn apply_geap_mint_result(
+    manager: &mut ApiKeyManager,
+    result: Result<GeapCredentials, LoadGeapCredentialsError>,
+    minted_for: GeapMintBinding,
+    force: bool,
+    ctx: &mut ModelContext<ApiKeyManager>,
+) {
+    // Gate flipped off mid-mint: discard the result; no token is retained
+    // while disabled.
+    if current_geap_request_gate(ctx).is_none() {
```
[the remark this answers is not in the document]
JaidenRatti: Fixed, we check the current binding again rather then just checking if the gate is on or off. We discard and remind unless it still matches `minted_for`.

crates/ai/src/geap_credentials.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Refresh the access token this long before its hard expiry so a request
+/// never races the expiration. Possibly-expired tokens are still sent (Google
+/// is the authority on validity), so this lead time is purely about keeping
+/// the token *fresh*, never about when it stops being *sent*.
+pub const GEAP_REFRESH_LEAD_TIME: Duration = Duration::from_secs(5 * 60);
+
+/// A short-lived Google Cloud access token minted through WIF. Lives only in
+/// memory and is never persisted; the fields are private so the only egress of
+/// the raw token is the conversion into the request wire type.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
JaidenRatti: @IsaiahWitzke is this a real concern? I see we have the same thing for [aws here](https://github.com/warpdotdev/warp/blob/3f83932cd127b5d17bb456a6d725a86d5d2651f3/crates/ai/src/aws_credentials.rs#L9-L10). Is this also in need of a specific redacting `Debug`?

app/src/ai/agent/api.rs
```diff
@@ -237,9 +237,17 @@ impl RequestParams {
         let user_workspaces = UserWorkspaces::as_ref(app);
         let api_key_manager = ApiKeyManager::as_ref(app);
         let is_byo_enabled = user_workspaces.is_byo_api_key_enabled(app);
+        // The expected Gemini Enterprise (GEAP) mint binding when the gate is
+        // on; `None` skips GEAP attachment entirely. Computing it here keeps
+        // `api_keys_for_request` a pure `&self` read without `AppContext`.
```
IsaiahWitzke: nit: i think these comments from agent mode are redundant... like the comment on the `current_geap_request_gate` defn is basically saying the same thing.

[this happens to me on the regular](https://github.com/warpdotdev/warp-server/pull/11831#discussion_r3400169131)

crates/ai/src/geap_credentials.rs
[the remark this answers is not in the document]
IsaiahWitzke: yeah, i guess we should be conservative here, yeah... 

do u you mind implementing a custom `Debug` for AWS for me too?

crates/ai/src/geap_credentials.rs
```diff
@@ -0,0 +1,149 @@
...
+/// build site while the GEAP enablement gate is on. Passing it into
+/// [`ApiKeyManager::api_keys_for_request`](crate::api_keys::ApiKeyManager::api_keys_for_request)
+/// keeps that function a pure `&self` read: the caller owns the policy
+/// evaluation (auth, admin availability, enablement mode, member toggle), and
+/// `None` means the gate is off and GEAP credentials are skipped entirely.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct GeapRequestGate {
+    pub user_uid: String,
+    pub audience: String,
+    pub sa_email: Option<String>,
```
IsaiahWitzke: so it was kinda surprising to me that a null sa_email here is a valid state... was going to comment "you should make this not optional" before discovering this [totally will work](https://docs.cloud.google.com/iam/docs/workload-download-cred-and-grant-access#by-group). nice find!

however, we could do a little better job of explaining in code what's going on here / what a Null sa_email actually does... wdyt of something like this?

```
enum GeapPolicy {
    Disabled,                  // gate off
    Unconfigured,              // enabled, empty audience  -> Missing
    Mintable { user_uid: String, audience: String, federation: GeapFederation },
}

enum GeapFederation {
    DirectWif,                         // 2-legged: STS token used directly
    ServiceAccount { email: String },  // 3-legged: + generateAccessToken
}
```

also i think theres comments all around that mention "3 legged flow" without a mention that "we swap to 2 legged if `sa_email.is_none()`, so might need to do a little comment audit.

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,683 @@
...
+            sts_response.access_token,
+            Some(federated_expires_at),
+        ));
+    };
+
+    // Leg 3: SA impersonation. IAM authorizes this only if the pool identity
+    // holds `roles/iam.workloadIdentityUser` on the SA — the customer's
+    // control point for who may become the SA.
+    let url = format!(
+        "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{sa_email}:generateAccessToken"
```
IsaiahWitzke: this should prob be a const up at the top along with our other constants

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,683 @@
...
+        .post(STS_TOKEN_URL)
+        .form(&StsTokenExchangeRequest {
+            grant_type: TOKEN_EXCHANGE_GRANT_TYPE,
+            audience: &config.audience,
+            scope: CLOUD_PLATFORM_SCOPE,
+            requested_token_type: ACCESS_TOKEN_TYPE,
+            subject_token: &identity_token.token,
+            subject_token_type: ID_TOKEN_TYPE,
+        })
+        .send()
```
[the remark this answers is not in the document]
IsaiahWitzke: not a bad point

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,683 @@
...
+/// Legs 2 and 3 of the mint: exchanges the Warp OIDC JWT at Google STS for a
+/// federated token, then (when configured) impersonates the workspace's
+/// service account for the final ~1h access token. Runs entirely off the
+/// request path via `http_client` (the Compat-wrapped reqwest required to run
+/// off-Tokio on the warpui executor).
+async fn exchange_identity_token_for_geap_credentials(
+    identity_token: TaskIdentityToken,
+    config: &GeapWifConfig,
+) -> Result<GeapCredentials, LoadGeapCredentialsError> {
+    // Leg 2: STS token exchange (RFC 8693). Google validates the issuer
```
IsaiahWitzke: once again... AM's repeated comments are a bit much

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,683 @@
...
+    access_token: String,
+    /// RFC 3339 timestamp; always present in IAM responses.
+    expire_time: String,
+}
+
+/// Legs 2 and 3 of the mint: exchanges the Warp OIDC JWT at Google STS for a
+/// federated token, then (when configured) impersonates the workspace's
+/// service account for the final ~1h access token. Runs entirely off the
+/// request path via `http_client` (the Compat-wrapped reqwest required to run
+/// off-Tokio on the warpui executor).
```
IsaiahWitzke: nit: could we clean up the func doc comment here...? i think this is talking about the http_client internals which doesn't really make sense to do at this level? i think mostly everywhere where we make http requests from the UI thread, we're using "Compat-wrapped reqwest" so i dont know if its worth pointing out here?

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,683 @@
...
+/// Cap on provider error detail captured into states/logs. Bodies are
+/// sanitized to this length and never contain token material.
+const ERROR_DETAIL_MAX_CHARS: usize = 512;
+
+const STS_TOKEN_URL: &str = "https://sts.googleapis.com/v1/token";
+const CLOUD_PLATFORM_SCOPE: &str = "https://www.googleapis.com/auth/cloud-platform";
+const TOKEN_EXCHANGE_GRANT_TYPE: &str = "urn:ietf:params:oauth:grant-type:token-exchange";
+const ID_TOKEN_TYPE: &str = "urn:ietf:params:oauth:token-type:id_token";
+const ACCESS_TOKEN_TYPE: &str = "urn:ietf:params:oauth:token-type:access_token";
+/// Requested lifetime for the impersonated SA access token (leg 3).
```
IsaiahWitzke: nit: delete the comment

lol i think the name of the var does a good job explaining what it does by itself

IsaiahWitzke: and the same goes for the comments on the `LoadGeapCredentialsError` enum members below

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,683 @@
...
+const GEAP_IDENTITY_TOKEN_DURATION: Duration = Duration::from_secs(60 * 60);
+
+/// Floor on the proactive refresh timer delay so a near-expired store (e.g. a
+/// badly skewed local clock) cannot spin mint -> store -> re-mint as a hot
+/// loop; the floor rate-limits timer-driven re-mints to once per minute.
+const GEAP_MIN_TIMER_DELAY: Duration = Duration::from_secs(60);
+
+/// Cap on provider error detail captured into states/logs. Bodies are
+/// sanitized to this length and never contain token material.
+const ERROR_DETAIL_MAX_CHARS: usize = 512;
```
IsaiahWitzke: was there a real reason you have this? i.e. in testing were error messages huge?

also, this feels like more of a thing that the UI code should be concerned about?

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// `roles/iam.workloadIdentityUser` binding or disabled API).
+    ImpersonateServiceAccount(String),
+}
+
+impl std::fmt::Display for LoadGeapCredentialsError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::MintIdentityToken(detail) => write!(
+                f,
+                "Failed to authenticate with Warp while minting Gemini Enterprise credentials \
```
IsaiahWitzke: this all feels like it should live at the UI level if its going to be user facing...

you could probably do:

```
enum LoadGeapCredentialsError {
    MintIdentityToken { detail: String }, // <- or maybe just include the entire Error in the detail?
    ExchangeToken { status: Option<StatusCode>, detail: String },
    ImpersonateServiceAccount { status: Option<StatusCode>, detail: String },
}

...

let identity_token = token_future.await.map_err(|err| {
    LoadGeapCredentialsError::MintIdentityToken { detail: format!("{err:#}") }
})?;
...
// request/parse failure → no status, full chain
.map_err(|err| LoadGeapCredentialsError::ExchangeToken {
    status: None,
    detail: format!("request failed: {err:#}"),
})?;

// non-success response → carry the real status + raw body (untruncated)
if !status.is_success() {
    let body = response.text().await.unwrap_or_default();
    return Err(LoadGeapCredentialsError::ExchangeToken {
        status: Some(status),
        detail: body,
    });
}
...
let expires_at = parse_generate_access_token_expiry(&impersonation.expire_time)
    .map_err(|detail| LoadGeapCredentialsError::ImpersonateServiceAccount {
        status: None,
        detail,
    })?;
...


// and elsewhere, we change GeapCredentialsState to transit thru the unformatted error:

// GeapCredentialsState
enum GeapCredentialsState {
  Failed { error: LoadGeapCredentialsError },   // was: Failed { message: String }
}
```

then the logic to display the correct user-facing message lives at the UI layer, and you don't have to worry about maxing error messages out to 512 nor nice human facing messages at this point...

and i also dont think it matters to log::error the full untruncated details either

app/src/lib.rs
```diff
@@ -1350,6 +1352,10 @@ pub(crate) fn initialize_app(
         let mut manager = ::ai::api_keys::ApiKeyManager::new(ctx);
         #[cfg(not(target_family = "wasm"))]
         manager.subscribe_to_settings_changes(ctx);
+        // Gemini Enterprise (GEAP) credential refresh triggers: workspace
+        // settings saves / team changes and the member's enablement toggle.
+        #[cfg(not(target_family = "wasm"))]
+        manager.subscribe_to_geap_settings_changes(ctx);
```
IsaiahWitzke: i think we should be putting this behind a FF (see below example with Suraj's Grok code)

this way we can dogfood internally without launching to everyone

crates/ai/src/geap_credentials.rs
IsaiahWitzke: then u can :axe: `GeapWifConfig` too. `Mintable { audience, federation }` becomes the config

IsaiahWitzke: and then u can also probably consider killing `GeapMintBinding` too?

crates/ai/src/geap_credentials.rs
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    pub fn expires_at(&self) -> Option<SystemTime> {
+        self.expires_at
+    }
+
+    /// The access token whenever it is non-empty — regardless of expiry.
+    /// Possibly-expired tokens are still sent so Google stays the final
+    /// authority on validity.
```
IsaiahWitzke: i think this comment is duplicated like 3 times in this pr... : /

IsaiahWitzke: also the entire comment for the needs_refresh below can prob be axe'd too... its like 2x more tokens than the code its describing lol

IsaiahWitzke (approved):
looking good!

nothing blocking... 2 big points tho:
a. the code could be made a lot simpler with a cleaner `GeapPolicy` enum
b. 75% of the claude-genned comments can be 🪓ed lol

crates/ai/src/geap_credentials.rs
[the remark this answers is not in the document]
JaidenRatti: for sure 👍

app/src/ai/agent/api.rs
JaidenRatti: will clean up all of the comments here. 

I guess this is what happens when models are rlhf’d on readability #goodharts..

app/src/ai/geap_credentials.rs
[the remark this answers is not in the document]
JaidenRatti: agreed, adding a timeout

app/src/ai/geap_credentials.rs
JaidenRatti: yes, this comment got oddly specific..

app/src/ai/geap_credentials.rs
JaidenRatti: 🫡

crates/ai/src/geap_credentials.rs
JaidenRatti: "Google stays the final authority" 😵‍💫. lol yea going to purge a lot of these comments.

app/src/ai/agent/api.rs
IsaiahWitzke: > goodharts

ty i will be using this vocab going fwd!

crates/ai/src/geap_credentials.rs
JaidenRatti: I agree that there is a bit of redundancy right now with the identical structs + some confusion with the state being implied by sa_email being null. 

I like the `GeapPolicy` and `GeapFederation` you proposed - it makes the directwif vs service account options more explicit. I agree with then definitely axing `GeapWifConfig`.

I see a world where `GeapMintBinding` could stay. I think it does help keep the staleness check a simple `==`, one named type reused across Mintable/Loaded/Refreshing instead of the same three fields inlined (and compared) in each.

Option A: Keep `GeapMintBinding`: 
```
struct GeapCredentials { 
    access_token, 
    expires_at 
} 

enum GeapFederation { 
    DirectWif, 
    ServiceAccount { email: String } 
}

struct GeapMintBinding {
    user_uid: String,
    audience: String,
    federation: GeapFederation,
}

enum GeapPolicy {                                                 
    Disabled,
    Unconfigured,
    Mintable(GeapMintBinding) // reuses binding                               
}

enum GeapCredentialsState {
    Missing,
    Disabled,
    Refreshing { previous: Option<(GeapCredentials, GeapMintBinding)> }, 
    Loaded { credentials: GeapCredentials, loaded_at: SystemTime, minted_for: GeapMintBinding },
    Failed { message: String },
}
```

Option B: `GeapPolicy + GeapFederation` + 🪓 `GeapWifConfig` and `GeapMintBinding`

```
struct GeapCredentials { access_token, expires_at } 

enum GeapFederation { 
    DirectWif, 
    ServiceAccount { email: String } 
}

enum GeapPolicy {                                         
    Disabled,
    Unconfigured,
    Mintable { user_uid: String, audience: String, federation: GeapFederation },
}

// GeapCredentialsState would then look like
enum GeapCredentialsState {
    Missing,
    Disabled,
    Refreshing { previous: Option<(GeapCredentials, String, String, GeapFederation)> },   // inlined tuple
    Loaded {
        credentials: GeapCredentials,
        loaded_at: SystemTime,
        minted_user_uid: String,        
        minted_audience: String,
        minted_federation: GeapFederation,
    }, // gets a little messy with all the fields
    Failed { message: String },
}
```

But not married to this.

IsaiahWitzke: ah yeah, option A was what i was _trying_ to suggest -- nice

app/src/ai/geap_credentials.rs
JaidenRatti: this is a good call, I like this design more.

app/src/ai/geap_credentials.rs
JaidenRatti: no real reason for this. and yea - agreed on this being more of a ui thing.

app/src/ai/geap_credentials.rs
```diff
@@ -0,0 +1,542 @@
...
+        .timeout(GEAP_MINT_REQUEST_TIMEOUT)
+        .send()
+        .await
+        .map_err(|err| LoadGeapCredentialsError::ExchangeToken {
+            status: None,
+            detail: format!("request failed: {err:#}"),
+        })?;
+    let status = response.status();
+    if !status.is_success() {
+        let body = response.text().await.unwrap_or_default();
```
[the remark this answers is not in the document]
JaidenRatti: Intentional. We dropped the 512-char ERROR_DETAIL_MAX_CHARS cap and now carry the untruncated provider body in the structured `LoadGeapCredentialsError`, with user-facing copy + truncation moved to the UI layer. These are Google STS/IAM error responses, so detail never contains the access token or the Warp OIDC JWT.

PR #11725, opened 2026-05-27
Fix conversation creator profile display

seemeroland:
## Summary
- preserve the AI conversation creator public profile in full and metadata-only GraphQL conversions
- prefer the server-provided creator profile (including photo URL) when rendering conversation details and entries
- add a regression test that prevents falling back to an unrelated UID initial when a creator profile is available

## Validation

Tested with https://github.com/warpdotdev/warp-server/pull/11468
Log in to one account, view conversation from different account
Screenshot 2026-06-02 at 11 37 48 AM

- `cargo check --manifest-path /workspace/warp/Cargo.toml`
- `cargo test --manifest-path /workspace/warp/Cargo.toml -p warp test_from_conversation_prefers_server_creator_profile`
- `git -C /workspace/warp --no-pager diff --check`

Co-Authored-By: Oz <oz-agent@warp.dev>

_Conversation: https://staging.warp.dev/conversation/1d22fd10-69a0-488e-a343-0c530639f4c1_
_Run: https://oz.staging.warp.dev/runs/019e66f6-e731-7ab9-9d11-bde60d8f7e5c_
_This PR was generated with [Oz](https://warp.dev/oz)._

PR #12585, opened 2026-06-12
[REMOTE-1782] Show departed participants' avatars on their agent queries in shared sessions

ianhodge:
## Description

When viewing a live shared session (most commonly a running cloud agent session), the avatar next to a user prompt showed the **viewing user's own** name/photo instead of the person who actually sent the prompt, whenever that person was no longer connected to the session.

**Why:** The AI block resolves the exchange's `response_initiator` (a session-sharing `ParticipantId`) through `PresenceManager::get_participant()`, which only returns *present* participants. Once the initiator disconnected, the lookup failed and the view fell back to the current user's auth state — so every later viewer saw themselves as the prompt author. (PR #11725 fixed the analogous bug for conversation *transcripts*, but live sessions resolve avatars through presence, not server conversation metadata.)

**How:** The presence manager already tracks departed participants in `absent_viewers` with their last-known profile data — it just wasn't consulted. This PR adds `PresenceManager::get_participant_attribution()`, which resolves present participants and the sharer as before, and falls back to absent viewers' last-known display name/photo (without a live presence color). The AI block query avatar now uses this attribution lookup. All live-presence call sites (cursors, role checks, block selections) intentionally keep using `get_participant()` and continue to exclude absent participants.

Note: this does **not** address follow-ups injected via warp-server's identity-less `/sessions/inject` path (those are attributed to the sharer by design today); that's a larger cross-service change tracked separately.

## Linked Issue

Linear: [REMOTE-1782](https://linear.app/warpdotdev/issue/REMOTE-1782/show-original-user-avatar-when-viewing-a-conversation-transcript-or) — the live shared-session half of the report (the transcript half shipped in #11725).

## Testing

Automated coverage added in `presence_manager_tests.rs`:
- `attribution_resolves_absent_viewer_profile_without_presence_color` — regression test: an absent viewer is not resolvable via `get_participant` but is via the new attribution lookup
- `attribution_still_resolves_viewer_after_they_leave` — present → absent transition keeps attribution working and drops the presence color
- `attribution_resolves_sharer_with_presence_color`
- `attribution_returns_none_for_unknown_participant`

Validation run locally:
- `cargo nextest run -E 'test(presence_manager)'` — 9/9 pass
- `./script/format` — clean

Manual two-client repro (sharer runs an agent prompt; viewer B joins with executor role, sends a follow-up, disconnects; viewer C joins) is the end-to-end check — before this change C sees their own avatar on B's prompt, after it C sees B's name/photo.

## Agent Mode

[Conversation](https://staging.warp.dev/conversation/0f4a1187-15d5-4337-8c18-33bd9b192652)

CHANGELOG-BUG-FIX: Fixed shared session viewers seeing their own avatar on agent prompts sent by participants who have since left the session.

### DIFF
--- a/app/src/ai/blocklist/block/view_impl.rs
+++ b/app/src/ai/blocklist/block/view_impl.rs
@@ -959,7 +959,9 @@ impl View for AIBlock {
             }
             // Derive the display info for the participant who initiated this exchange.
             // For non-shared sessions, this is just the current user.
-            // For shared sessions, this is the user who initiated the request.
+            // For shared sessions, this is the user who initiated the request, even if
+            // they have since left the session (in which case we use their last-known
+            // profile data without a live presence color).
             let (avatar_display_name, profile_image_path, avatar_color) = self
                 .model
                 .response_initiator(app)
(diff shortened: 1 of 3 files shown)


ianhodge:
Working on an e2e manual test right now too

seemeroland (commented):
Would be good to test this. I have a feeling this might fix if we were both viewers in the shared session and then you left. But if you ran some queries, you left, and I later joined, I don't think `absent_viewers` will be populated with viewers who left before I joined.

PR #12586, opened 2026-06-12
Apply NLD history match to agent prompt history backed by ai_queries

evelyn-with-warp:
## Description

Fuzzy-match the input buffer against both shell command history and agent prompt history, locking NLD to whichever mode has the later matched timestamp. 

- loaded most recent 2000 rows from ai_queries
- chained with new conversation since app started
- run fuzzy match with command and prompt history (both sorted by start_ts desc), with early return upon match found. 
## Linked Issue

## Testing

```
RUST_LOG=debug ./script/run --features nld_heuristic_v2,nld_classifier_v3,nld_prompt_history_match,warpui/log_named_telemetry_events
```

### Screenshots / Videos

https://www.loom.com/share/06c6ef1c669d49818d39f5c4c04a9c30
## Agent Mode


app/src/persistence/block_list.rs
```diff
@@ -106,6 +106,53 @@ pub(super) fn read_ai_queries(
...
+            schema::ai_queries::columns::input,
+            schema::ai_queries::columns::start_ts,
+        ))
+        .filter(schema::ai_queries::columns::input.ne("[]"))
+        .order_by(schema::ai_queries::columns::id.desc())
+        .limit(limit)
+        .load::<(String, NaiveDateTime)>(conn)?
+        .into_iter()
+        .filter_map(|(input, start_ts)| {
+            let inputs: Vec<PersistedAIInputType> = serde_json::from_str(&input).ok()?;
```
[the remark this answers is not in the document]
evelyn-with-warp: this is one-time read, not triggered at every keystroke per user type. so it's fine to not narrow read trading for reuse existing struct

app/src/persistence/block_list.rs
```diff
@@ -106,6 +106,53 @@ pub(super) fn read_ai_queries(
...
+    read_nld_prompts_with_limit(conn, MAX_AI_QUERIES_TO_READ_FOR_NLD)
+}
+
+/// Split out from [`read_nld_prompts`] so tests can exercise the read cap with a small
+/// limit instead of inserting thousands of rows.
+fn read_nld_prompts_with_limit(
+    conn: &mut SqliteConnection,
+    limit: i64,
+) -> Result<Vec<(String, DateTime<Local>)>, diesel::result::Error> {
+    Ok(schema::ai_queries::table
```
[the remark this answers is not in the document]
evelyn-with-warp: it's fine to bring back the filtered out prompt for this use case, as long as it is a prompt which aims to lock NLD to ai mode

script/linux/bundle
```diff
@@ -215,6 +215,11 @@ if [[ "$ARTIFACT" == "cli" || "$ARTIFACT" == "warpctrl" ]]; then
 elif [[ "$ARTIFACT" == "app" ]]; then
   # All channels ship the v3 classifier and v2 heuristic.
   FEATURES="$FEATURES,gui,nld_classifier_v3,nld_heuristic_v2"
+  # The prompt-history match feature is still in development, so only enable
+  # it for local/dev builds.
+  if [[ "$RELEASE_CHANNEL" == "local" || "$RELEASE_CHANNEL" == "dev" ]]; then
+    FEATURES="$FEATURES,nld_prompt_history_match"
+  fi
```
szgupta: Let's use normal feature flags for this because it doesn't affect bundle size

app/src/ai/blocklist/history_model.rs
```diff
@@ -249,6 +250,13 @@ pub struct BlocklistAIHistoryModel {
     /// history.
     persisted_queries: Vec<PersistedAIInput>,
 
+    /// Prompt-history candidates for NLD input classification, read once from `ai_queries` at
+    /// startup.
+    nld_persisted_prompts: Vec<(Arc<str>, DateTime<Local>)>,
+
+    /// User-query prompts submitted during the current session,
+    nld_session_prompts: Vec<(Arc<str>, DateTime<Local>)>,
```
szgupta: Do these need to be separate? I was thinking this would just be `prompt_history` and it would be initialized using sqlite and updated as new prompts are submitted.

Also, it's worth calling out how / why these are different from `persisted_queries` right above.

app/src/ai/blocklist/input_model.rs
```diff
@@ -810,16 +817,41 @@ impl BlocklistAIInputModel {
...
+                                .rev()
+                                .map(|(command, start_ts)| (command.as_str(), *start_ts)),
                             HISTORY_ENTRY_MATCH_CUTOFF,
                         )
-                        .await
-                        {
-                            return (InputType::Shell, InputTypeAutoDetectionSource::HistoryMatch);
+                        .await;
+
+                        futures_lite::future::yield_now().await;
```
szgupta: Why do we have a manual yield here?

app/src/persistence/block_list.rs
```diff
@@ -106,6 +106,53 @@ pub(super) fn read_ai_queries(
...
+
+/// Reads recent AI query prompts (text and submission time) for NLD prompt-history
+/// matching, newest-first.
+///
+/// Rows with an empty serialized input (`[]`) are excluded in SQL before the limit is
+/// applied: rows written before empty inputs were skipped at write time remain in the
+/// table until FIFO eviction removes them. The result is intentionally not deduped;
+/// deduplication happens after these rows are combined with in-memory conversation
+/// queries in [`crate::ai::blocklist::BlocklistAIHistoryModel::nld_prompt_history`].
+pub(super) fn read_nld_prompts(
```
szgupta: nit: here and everywhere else, we shouldn't name this "nld". The prompt history is used by NLD but doesn't need to be coupled to it so explicitly. We might use prompt history for other stuff (e.g. ctrl-r menu)

szgupta (commented):
Did an initial pass!

script/linux/bundle
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/12586/changes/6435a32e6f52e4d51b3d4d85ef36e1547cad27e6

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: performance wise to support early abortion of fuzzy match , if next keystroke is triggered;

app/src/ai/blocklist/history_model.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/12586/changes/183d26866ee0b5b90977d95a96d8b6501cd4e8ef now we have prompt history to cover both, but still it only serves for nld history match for now; 
https://github.com/warpdotdev/warp/pull/12586/changes/e2d8ec20690e343dd716833c26cdcffe695b1b33 Also added todo in comment for future refactor reference: consolidate `prompt_history` and `persisted_query` when uparrow prompt history( ctrl + r) supports pagination

app/src/ai/blocklist/history_model.rs
```diff
@@ -244,11 +245,17 @@ pub struct BlocklistAIHistoryModel {
...
+    // TODO: When up-arrow prompt history supports pagination, consolidate
+    // `persisted_queries` and `prompt_history` (both seeded from the same `ai_queries`
+    // read) and share the in-memory conversation appending done by `all_ai_queries`.
     /// AI queries that were read from the SQLite DB. These exchanges do not contain as much
     /// information as the other exchanges we store because they are only used for display in
     /// history.
     persisted_queries: Vec<PersistedAIInput>,
 
+    /// Prompt-history candidates seeded once from `ai_queries` at startup via
+    prompt_history: Vec<(Arc<str>, DateTime<Local>)>,
```
szgupta: This should be `String`, not `str`. Also do we need an `Arc` here (or any reference counting)?

app/src/ai/blocklist/history_model.rs
```diff
@@ -244,11 +245,17 @@ pub struct BlocklistAIHistoryModel {
     /// conversations out of local history and navigation.
     conversation_transcript_viewer_terminal_view_ids: HashSet<EntityId>,
 
+    // TODO: When up-arrow prompt history supports pagination, consolidate
+    // `persisted_queries` and `prompt_history` (both seeded from the same `ai_queries`
+    // read) and share the in-memory conversation appending done by `all_ai_queries`.
```
szgupta: nit: this comment reads better on `prompt_history`

app/src/ai/blocklist/history_model.rs
szgupta: Also, can we make a helper type instead of using a tuple like this? Will make it easier to build on top of in the future

app/src/ai/blocklist/history_model.rs
```diff
@@ -980,18 +995,41 @@ impl BlocklistAIHistoryModel {
...
-            .ok_or(UpdateHistoryError::ConversationNotFound(
-                request_input.conversation_id,
-            ))?;
+            .get_mut(&conversation_id)
+            .ok_or(UpdateHistoryError::ConversationNotFound(conversation_id))?;
+
+        // That first exchange is the synthetic orchestrator prompt (not user input)
+        // which the NLD prompt history needs to exclude.
+        let is_synthetic_orchestrator_prompt = conversation.is_child_agent_conversation()
+            && conversation.root_task_exchanges().next().is_none();
```
szgupta, line 1028: nit: is there a better way to tell if the input was a user input vs "system"?

app/src/persistence/block_list.rs
```diff
@@ -87,25 +87,67 @@ impl TryFrom<&PersistedAIInput> for NewAIQuery {
...
-    // Only load at most 100 AI queries; there's a very low chance that the user
-    // will ever try rerunning AI queries older than this duration and loading
-    // all AI queries in perpetuity has performance implications on app startup.
-    // TOOD(alokedesai): Consider loading all AI queries by paginating the SQL query.
-    const MAX_AI_QUERIES_TO_READ: i64 = 100;
     Ok(schema::ai_queries::table
         .select(AIQuery::as_select())
         .order_by(schema::ai_queries::columns::start_ts.desc())
-        .limit(MAX_AI_QUERIES_TO_READ)
+        .limit(MAX_AI_QUERIES_READ_LIMIT)
```
szgupta, line 108: This will be a material change, right? We're going from reading 100 rows to 2k. Just making sure this will be performant enough?

app/src/ai/blocklist/input_model.rs
szgupta: The `await` before this will give the executor a change to yield to a different task. Don't think we need to manually do this.

app/src/ai/blocklist/input_model.rs
```diff
@@ -928,40 +966,85 @@ impl Entity for BlocklistAIInputModel {
...
+/// Returns `None` when neither source matched
+/// - command-only match -> Shell
+/// - prompt-only match -> AI
+/// - both matched: the entry with the later timestamp wins. When the command
+///   has no timestamp (e.g. a history-file entry) or the timestamps are equal,
+///   we cannot prove the prompt is more recent, so we preserve the existing
+///   Shell short-circuit.
+fn resolve_history_match(
+    command_match: Option<Option<DateTime<Local>>>,
+    prompt_match: Option<Option<DateTime<Local>>>,
```
szgupta: Why is this a nested `Option`? Can we flatten it? If not, what do the different states (e.g. `Some(None)`) mean?

app/src/ai/blocklist/input_model.rs
```diff
@@ -810,16 +818,46 @@ impl BlocklistAIInputModel {
...
-                        if has_any_close_matches(
+                        // Iterate commands newest-first so the first match is the
+                        // most-recent matching command.
+                        let command_match = most_recent_close_match(
                             &buffer_cloned,
-                            history_entries.iter().map(AsRef::as_ref),
+                            history_entries
+                                .iter()
+                                .rev()
+                                .map(|(command, start_ts)| (command.as_str(), *start_ts)),
```
szgupta, line 716: Noticed we weren't reversing this before. is this intentional? what's the functional difference?

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: removed here https://github.com/warpdotdev/warp/pull/12586/changes/90daf3cf4f1475477a281a4621ff01233c84cc31

app/src/ai/agent/mod.rs
```diff
@@ -2893,6 +2893,14 @@ impl AIAgentInput {
         Some(query)
     }
 
+    /// Returns the raw user query text for the [`Self::UserQuery`] variant.
+    pub fn user_query(&self) -> Option<String> {
+        match self {
+            AIAgentInput::UserQuery { query, .. } => Some(query.clone()),
+            _ => None,
+        }
+    }
```
evelyn-with-warp, line 2967: different with `display_user_query` this will not include /command; NLD only judge on pure user input

app/src/ai/blocklist/history_model.rs
evelyn-with-warp: [Warp Agent] We use `Arc<str>` instead of `Arc<String>`/`String` because we never grow the prompt text in place — we only append new entries to the list — so the immutable, single-allocation `Arc<str>` is the right fit. We need the `Arc` because `prompt_history_candidates()` hands an owned clone of the whole list to a detached async autodetection task on every keystroke, and reference counting keeps that clone to cheap pointer bumps instead of deep-copying up to 2000 prompts.

Resolved in https://github.com/warpdotdev/warp/commit/ac26edf9dc9700c54638aecd19ebae5710508af1

app/src/persistence/block_list.rs
evelyn-with-warp: stress test (from cloud build) is like ~ 10ms - 30 ms, will confirm in dogfood, if that will cause user perceived performance regression

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: we didn't need time stamp before; this rev secures the first match returned will be the most recent record , as the sqlitedb read returned in asc manner

app/src/ai/blocklist/history_model.rs
evelyn-with-warp: [Warp Agent] There isn't a cleaner runtime signal today — `is_child_agent_conversation()` only knows about parent IDs, and the synthetic orchestrator prompt is itself a real `UserQuery`, so `has_user_query()` can't tell it apart. The proper fix is an explicit `is_agent_initiated` flag on the MAA `UserQuery` proto so the distinction survives server restoration; that's tracked as QUALITY-636.

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: Not recommended to flatten — the nested `Option` represents three distinct states:

- `None` → no match in this source
- `Some(Some(ts))` → matched at time `ts`
- `Some(None)` → matched, but with no timestamp (e.g. the only ts missing source is a command from the shell history file such as `.zsh_history`) 

So we have the flexibility on how to handle match when timestamp is None; atm if both command and prompt have text match, but command does not have time stamp, we will lock to AI)

app/src/ai/blocklist/history_model.rs
evelyn-with-warp: https://github.com/warpdotdev/warp/pull/12586/changes/aa3e219c5d9366231966397a5a8aefb9afc78c01

app/src/ai/blocklist/input_model.rs
szgupta: Can you create a private enum helper type to enumerate these distinct states? nested option is hard to understand

app/src/ai/blocklist/input_model.rs
```diff
@@ -810,16 +818,41 @@ impl BlocklistAIInputModel {
...
-                    // close matches to short-circuit as shell input.
+                    // close matches against command history and agent prompt history.
                     if let Some(history_entries) = history_entries {
-                        if has_any_close_matches(
+                        // Iterate commands newest-first so the first match is the
+                        // most-recent matching command.
+                        let command_match = most_recent_close_match(
                             &buffer_cloned,
-                            history_entries.iter().map(AsRef::as_ref),
+                            history_entries
```
szgupta, line 713: Can you make sure it's documented that history_entries are sorted in ascending order?

szgupta (approved):

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: Done — replaced the nested `Option<Option<DateTime<Local>>>` with a private `HistoryMatch` enum (`NoMatch` / `MatchedAt(ts)` / `MatchedWithoutTimestamp`), which `most_recent_close_match` now returns and `resolve_history_match` matches on.

app/src/ai/blocklist/input_model.rs
evelyn-with-warp: Done — added a comment at the `history_entries` gathering site noting that `History::commands` returns entries in ascending (oldest-first) order, and that downstream matching reverses them to iterate newest-first. The prompt-history gathering and the reversal sites are documented similarly.

PR #12587, opened 2026-06-12
Add setting to auto-queue prompts until the end of an LRC (and default it to enabled)

harryalbert:
## Description

Before this change, prompts submitted while an agent was driving a long-running command were sent to the agent immediately, steering it mid-command. This was called out by an enterprise customer (and by ZL) as a bit annoying, with the expected behavior being that their prompt is sent to the main agent instead of the LRC agent. The LRC agent does have the ability to pass prompts back up to the main agent, but the prompt still isn't in prompt history or clearly sent to the main agent in the conversation.

To fix this, this PR adds (and auto-enables) LRC prompt queueing while an agent is in control of a long-running command. Prompts submitted during the command get queued (with a "queued until the command finishes" note in the queue panel) and are sent to the agent in order once the command finishes. The queue chip and ghost text reflect this state, and toggling the chip off mid-command is scoped to just that command — your conversation's queue mode reverts to whatever it was before once the command ends.

This is controlled by a new "Default long-running command submission mode" setting (Settings → AI, right below the default prompt submission mode dropdown) with two options: "Queue until command finishes" (the default) and "Send immediately" (the old behavior). It only applies — and is only shown — when the default prompt submission mode is Interrupt, since in Queue mode prompts already queue until the end of the full response.

I think these invariants are a little confusing when written out, but in the actual product I think the behavior is pretty clear (see demo for product walkthrough). ZL also gave his stamp on this solution/product behavior.

## Testing

### Screenshots / Videos

https://www.loom.com/share/f74acb178a9b40819b100f43d4a1466a

## Agent Mode


harryalbert:
* **#12587** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12587?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12587?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

vkodithala:
nit: We should prob add an entry to our changelog with this PR, and also update our [docs](https://docs.warp.dev/agent-platform/capabilities/full-terminal-use/#how-full-terminal-use-works) on FTU to reference how this all works with prompt queueing!

app/src/ai/blocklist/queued_query.rs
```diff
@@ -161,6 +167,9 @@ struct ConversationQueueState {
     /// dispatched and cleared when it finishes; keeps the queue accepting new rows while the
     /// agent is idle and gates the next drain until the command completes.
     command_in_flight: bool,
+    /// Manual queue toggle made while an agent controls a long-running command. Cleared when
+    /// the command ends; never touches `queue_next_prompt_override`.
+    queue_next_lrc_prompt_override: Option<bool>,
```
vkodithala: I like that this is restricted to the LRC and resets upon exit, definitely feels like the right product behavior!

app/src/settings/ai.rs
```diff
@@ -532,6 +532,71 @@ impl PromptSubmissionMode {
...
+    description = "What happens when a prompt is submitted while an agent controls a long-running command.",
+    rename_all = "snake_case"
+)]
+pub enum LongRunningCommandSubmissionMode {
+    /// Send the prompt to the agent immediately, steering it mid-command.
+    SendImmediately,
+    /// Queue the prompt and send it to the agent when the command finishes
+    /// (default).
+    #[default]
+    QueueUntilCommandCompletes,
```
vkodithala, line 563: I worry about making this the default. Chiefly I'm concerned about the case where users manually tag in a CLI subagent (i.e. via running a command manually and hitting Cmd+Enter). I feel like the intention there would be that their query goes to the subagent (but do think you're right that for agent-initiated commands, sent user queries are probably intended for the primary).

I don't think it's a huge deal and am fine with shipping as-is and seeing how folks feel, but wanted to call out in case there was a middle ground we could strike here between agent-invoked and user-invoked commands.

app/src/terminal/view.rs
```diff
@@ -5543,6 +5544,31 @@ impl TerminalView {
...
+    /// queued rows, which wait for the end of the full response.
+    pub(crate) fn send_lrc_queued_prompts(
+        &mut self,
+        conversation_id: AIConversationId,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        let rows: Vec<(QueuedQueryId, String)> = QueuedQueryModel::as_ref(ctx)
+            .queue(conversation_id)
+            .iter()
+            .filter(|row| row.origin() == QueuedQueryOrigin::LrcAutoQueue)
```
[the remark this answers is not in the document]
vkodithala: Is this valid?

harryalbert:
> nit: We should prob add an entry to our changelog with this PR, and also update our [docs](https://docs.warp.dev/agent-platform/capabilities/full-terminal-use/#how-full-terminal-use-works) on FTU to reference how this all works with prompt queueing!

​Changelog entry is in place (it's just hidden in the PR description, but you can see it if you try to edit the description), and docs change is in progress ✅ . I'll submit it once this PR is merged

vkodithala (approved):
Few comments, nothing blocking. Excited for this to go out + apologies for the delay here!

app/src/settings/ai.rs
harryalbert: hmm, that's a good point. Yeah I think only using this behavior for commands _initiated by the agent_ would be reasonable. I'll make that change

app/src/terminal/view.rs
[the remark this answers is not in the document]
harryalbert: eh, kind of. It's an edge case that I'm not super worried about though, and I don't think there's a good solution short of cancelling the entire conversation if the user is editing a prompt, an then auto-resuming when the user is done, which feels just as obstructive.

I think for now the current approach makes sense, and then if people have issues we can circle back

PR #12595, opened 2026-06-12
Modify existing tab and group functionality to enforce pinning invariant

johnturcoo:
## Description

Updates existing tab and tab-group operations so they respect the new invariant for pinned items: effectively pinned tabs (individually pinned, or members of a pinned group) always form a contiguous block at the front of the tab list. Without these changes, several existing flows could land tabs/groups inside the pinned region without ever being pinned. 

**NOTE**: It is important to understand that pinning a tab and group are independent events and that a tab within a pinned group is not equivalent to a pinned tab. Tabs can be in the pinned region as part of a pinned group.

TLDR all pinned items must appear before all unpinned items. 

This PR **also** fixes a bug with tab groups. If the first/last tab in a group was repositioned with the "move tab up/down" action in the tab menu, it could be removed from the group and cause it to split. This is now not possible. Tabs can only be repositioned within their group unless explicitly removed from group or dragged out. 

## How it works 

All of the changes are to the existing tab grouping functions. Rather than explaining any logic, I will explain the product behavior below. This is quite dense, happy to go over sync:

Updated several tab/group operations so that new tabs/groups land after the pinned region, when applicable:

- New tab placement: creating a new tab ensures the tab lands after the pinned region, unless it is part of a group 
- Creating new tab groups: new tab groups are now inserted after the pinned region, instead of the beginning of the tab list
- Move to group: moving one or multiple tabs to a group remove the individual pin state on all of those tabs
- Remove from group: removing one or multiple tabs from a group cause them to land after the pinned region, if that group was pinned (rather than just after the group)
- Group from selection: creating a group from one or more tabs unpins any selected tab(s) and positions the group where the first selected tab originally was (if this was within the pinned region, it is moved to after the pinned region)
- Ungrouping: when a group is pinned and ungrouped, all the members are moved to after the pinned region
- Reorder gating: created new helper function's `can_move_tab()` and `can_move_tab_group()` to gate the move tab up/down options in the tab menus (same for groups). This ensures that tabs/groups can not be moved into the pinned region, and can not be moved out of the tab lists bounds. Additionally, it prevents tabs from being moved out of a group, which fixes the bug described above.

Added a few new helpers that are used throughout this PR, probably helpful to scan:
- `pinned_boundary_index()`: used to find the boundary between pinned and unpinned items, for deciding where certain items should land
- `clamp_to_unpinned_region(index)`: pushes an index to the beginning of the unpinned region/after the pinned region, if the index param is within the pinned region. Used whenever we want something to land at 'index' but it can not land within the pinned region
- `index_after_group(group_id)`: determines the index directly after the last tab in a group. Used for appending to groups and placing items directly after a group. 

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4721/modify-existing-tab-and-group-functionality-to-respect-pinning

and 

https://linear.app/warpdotdev/issue/APP-4729/fix-bug-with-move-tabs-option

## Testing

### Screenshots / Videos

[Demo of bug](https://www.loom.com/share/3042f8f022464a888f59ca45df601a47)

[Demo of fix](https://www.loom.com/share/e205d80c65ae414bab713d4017b6951f)

- the demo is just showing that the 'move up' option is not available for the first tab, and the 'move down' option is not available for the last tab

[Demo of grouping behavior with pinning](https://www.loom.com/share/1f13a351b8744a7d8b738487f4c632bd)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -6824,16 +6824,16 @@ impl Workspace {
             ctx,
         );
         let new_tab_index = self.active_tab_index;
+
+        // Ensure that new tab groups always land below pinned items, but above
+        // any other items in the tab list.
+        let target = self.pinned_boundary_index();
+
         if let Some(tab) = self.tabs.get_mut(new_tab_index) {
             tab.group_id = Some(group_id);
         }
 
-        // New tab groups always land at the top of the tab list.
-        if new_tab_index != 0 {
-            let tab = self.tabs.remove(new_tab_index);
-            self.tabs.insert(0, tab);
-            self.active_tab_index = 0;
-        }
+        self.move_tab_to_index(new_tab_index, target, ctx);
 
         ctx.dispatch_global_action("workspace:save_app", ());
         ctx.notify();
@@ -6919,10 +6919,15 @@ impl Workspace {
             });
     }
 
-    /// Creates a new group containing the tab, anchored at the tab's original
-    /// position. If the tab was pulled out of an existing group, leaving it in
-    /// place would split that group's contiguous run, so the new group lands
-    /// just past the old group's last remaining member instead.
+    /// Creates a new group containing the tab. The new group is unpinned
+    /// regardless of whether the source tab was pinned — tab pinning and
+    /// group pinning are independent concepts, and the user can pin the new
+    /// group explicitly after creation. The block lands at the source tab's
+    /// original slot, except:
+    /// * If the tab was in an existing group, anchor past that group's last
+    ///   remaining member so the old group stays contiguous.
+    /// * If the tab was effectively pinned, clamp past the pinned region so
+    ///   the new (unpinned) group doesn't land inside the pinned area.
     fn new_tab_group_from_tab(&mut self, tab_index: usize, ctx: &mut ViewContext<Self>) {
         if !FeatureFlag::GroupedTabs.is_enabled() {
             return;
@@ -6937,15 +6942,21 @@ impl Workspace {
         let group_id = group.id;
         self.tab_groups.insert(group_id, group);
 
-        self.tabs[tab_index].group_id = Some(group_id);
+        // If the tab we are making a group from, is already in a group
+        // our target index is directly after the group's last member.
+        // Otherwise it is the tab's current index.
+        let natural_target = previous_group_id
+            .and_then(|gid| self.index_after_group(gid))
+            .unwrap_or(tab_index);
 
-        // When the tab leaves an existing group, reposition it just past that
-        // group's last remaining member so the old group stays contiguous.
-        if let Some(prev_group_id) = previous_group_id {
-            if let Some(last) = group_member_indices(&self.tabs, prev_group_id).last() {
-                self.move_tab_to_index(tab_index, last + 1, ctx);
-            }
-        }
+        // Our target index should always be after all existing pinned items.
+        let target = self.clamp_to_unpinned_region(&self.tabs, natural_target);
+
+        self.tabs[tab_index].group_id = Some(group_id);
+        // Pin state is cleared when a new group with this tab is created.
+        self.tabs[tab_index].pinned = false;
+
+        self.move_tab_to_index(tab_index, target, ctx);
 
         // The move above may have shifted the tab; the new group has exactly
         // one member, so its position is the new active index.
@@ -6985,11 +6996,13 @@ impl Workspace {
         }
         let previous_group_id = tab.group_id;
 
-        let target_index = group_member_indices(&self.tabs, group_id)
-            .last()
-            .map(|i| i + 1)
+        let target_index = self
+            .index_after_group(group_id)
             .unwrap_or(self.tabs.len());
         self.tabs[tab_index].group_id = Some(group_id);
+
+        // Moving a tab to a group, clear its pinned state.
+        self.tabs[tab_index].pinned = false;
         self.expand_tab_group(group_id, ctx);
         self.move_tab_to_index(tab_index, target_index, ctx);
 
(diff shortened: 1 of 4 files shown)


vkodithala:
this is a drive-by comment, but it seems like we're doing too many things with this PR. we're (1) enforcing a pinned tab invariant in tab grouping and (2) fixes a bug with regular tab repositioning. these should almost always be two separate PRs; it helps me review easier/quicker and it's easier to roll back if for ex. (2) was the wrong change. 

totally fine for now but just noting for posterity/future engineering work since i think this is something most ppl appreciate! p.s. agents are really good at separating grouped changes into different PRs, i've been just getting agent mode to do this for me a bunch

app/src/workspace/view/tab_grouping.rs
```diff
@@ -146,7 +146,7 @@ impl Workspace {
     /// Re-seats `active_tab_index` so the previously-active pane group stays
     /// visually active across a tab reorder. Pass the pane group id captured
     /// before the reorder; no-op if it can't be found.
-    fn restore_active_tab_index(&mut self, pane_group_id: Option<EntityId>) {
+    pub(super) fn restore_active_tab_index(&mut self, pane_group_id: Option<EntityId>) {
```
vkodithala, line 149: i like that you used `super` here to control exposure, this is a good pattern 💯

app/src/workspace/view/tab_grouping.rs
```diff
@@ -478,13 +490,32 @@ impl Workspace {
...
             .count()
     }
 
+    /// Pushes `idx` past the leading effectively-pinned tabs in `tabs` if it
+    /// falls inside that prefix.
+    pub(super) fn clamp_to_unpinned_region(&self, tabs: &[TabData], idx: usize) -> usize {
+        let boundary = tabs
+            .iter()
+            .take_while(|tab| self.is_tab_effectively_pinned(tab))
+            .count();
```
vkodithala: nit: why not just use `pinned_boundary_index` instead of recomputing inline? helps avoid some code duplication

app/src/workspace/view.rs
```diff
@@ -7032,12 +7051,39 @@ impl Workspace {
...
+                let drained: Vec<TabData> = self.tabs.drain(first..=last).collect();
+                // Recompute boundary now that the (formerly group-pinned)
+                // block has been removed from the list.
+                let target = self.pinned_boundary_index();
+                // Place the ungrouped tabs at their new target index.
+                self.tabs.splice(target..target, drained);
+                self.restore_active_tab_index(active_pane_group_id);
+            }
+        }
+
```
vkodithala, line 7086: nice job, and well explained. this was very easy for me to understand. leaving the codebase better than you found it!

app/src/workspace/view.rs
```diff
@@ -7098,14 +7173,14 @@ impl Workspace {
         if !FeatureFlag::GroupedTabs.is_enabled() {
             return;
         }
+        if !self.can_move_tab_group(group_id, direction) {
+            return;
+        }
```
vkodithala, line 7178: drive-by and out-of-scope: i wonder if it's weird that we no-op here to users. i'd like to give this a try in dogfood first, but maybe we consider adding a toast when they drag a pinned tab/group to a region where it can't live explaining why we no-op. worth thinking about

app/src/workspace/view.rs
```diff
@@ -7254,6 +7328,11 @@ impl Workspace {
 
         let previous_group_id = self.tabs[tab_index].group_id;
         self.tabs[tab_index].group_id = group_id;
+        // Entering a group: drop any individual pinned flag. The group's own
+        // `pinned` flag now governs whether this tab is in the pinned region.
+        if group_id.is_some() {
+            self.tabs[tab_index].pinned = false;
+        }
```
vkodithala, line 7335: i'm not sure and not tied to this, but i wonder if we should persist the pinned-ness of the tab even on group moves? take the following repro, i:
- pin tab
- move tab to pinned group
- move tab out of pinned group

would i expect that tab to still be pinned, since i set that state earlier? wdyt?

app/src/workspace/view.rs
```diff
@@ -13366,13 +13463,54 @@ impl Workspace {
...
     /// Moves the tab at `index` one slot left/right, where a "slot" is either a
     /// single tab or an entire adjacent group. If the neighbor in the move
     /// direction belongs to a *different* group, the tab hops over that whole
     /// group so it never lands inside the group's contiguous run. Moving past an
     /// ungrouped tab, or reordering within the tab's own group, is an ordinary
     /// one-slot move.
     fn move_tab(&mut self, index: usize, direction: TabMovement, ctx: &mut ViewContext<Self>) {
+        if !self.can_move_tab(index, direction) {
+            return;
+        }
```
vkodithala, line 13513: ditto to my earlier comment on considering a toast, out of scope tho

app/src/workspace/view.rs
vkodithala: basically i wonder if the invariant should be:
- show pinned icon on tab if tab.pinned and gid != pinned
- if gid == pinned, and tab moved out of group, tab moves to last position in pinned section if its internal state is pinned

if that makes sense

vkodithala (approved):
exceptional work, i've enjoyed your attention to detail and focus on fixing all of these bugs. this is looking awesome - keep sending these my way!

few comments, none blocking

app/src/workspace/view.rs
johnturcoo: This is honestly just a safety net, since the move tab group option will not be available in the menu for this case. My next PR handles dragging, and it should be more clear to the user that unpinned items can not be moved into the pinned region.

app/src/workspace/view.rs
johnturcoo: I thought about this for a while as well, and don't think there is a clear right answer. In google chrome tabs, moving a pinned item to a group is treated as a destructive action that unpins the tab. This is a bit of a unique scenario since we also support pinning tab groups (which most other platforms do not). I think we will see what users think from dogfood and preview, and can update accordingly. 

IMO, I think tab pinning and group pinning should be independent of each other and there should be no "hidden states" (a tab is in a group and there is no indication it is pinned until it is dragged out, would be weird, even if the tab was pinned originally)

PR #12599, opened 2026-06-12
Add paste-the-code fallback for SuperGrok (xAI) OAuth login

bholmesdev:
## Description
Fixes the SuperGrok (xAI) subscription connect flow for users who can't complete the browser redirect. xAI's consent screen sometimes **displays an authorization code** for the user to paste back into Warp ("Copy the code below into Grok Build to finish signing in") instead of redirecting to the loopback callback — for example when the browser/network blocks the cross-origin Private Network Access fetch to `127.0.0.1`. Previously the connect flow relied solely on that loopback redirect, so these users hit:

> Couldn't connect SuperGrok: failed to read the Grok OAuth callback request

…and had no way to finish signing in. This reflects the Slack report that the SuperGrok auth loop could not be completed on the latest stable build.

This PR adds a **paste-the-code fallback** alongside the existing loopback flow:

- **`crates/ai/src/grok_subscription/oauth.rs`** — `OauthAttempt::manual_code_exchange()` returns an opaque `ManualCodeExchange` that captures the attempt's PKCE verifier (cloned, so the loopback callback server keeps running in parallel) and exchanges a user-pasted code for tokens, reusing the same loopback redirect URI from the original authorize request. The CSRF `state` check is intentionally skipped here because the code is copied out of band; PKCE (the verifier bound to the original challenge) protects the exchange. Blank codes are rejected before any network call.
- **`app/src/settings_view/ai_page.rs`** — while a connect attempt is in progress, the main row button changes to a disabled **"Connecting"** state and a **"Paste sign-in code"** input appears beneath the row. Pasting into that input auto-submits the code (matching Warp's existing auth-token paste behavior); manually typed codes can still be submitted with Enter. The code exchange persists tokens via `ApiKeyManager`. Whichever path completes first connects the subscription; stale completions are ignored. If the loopback path times out/fails before manual entry completes, Warp exits the disabled connecting state and shows **"Retry"** while keeping the manual code fallback available.

## Linked Issue
N/A — reported via `#feedback-app` Slack thread (SuperGrok login failure on stable).

## Testing
- Added unit tests in `crates/ai/src/grok_subscription/oauth_tests.rs`:
  - `manual_code_exchange_captures_attempt_verifier` — the fallback reuses the originating attempt's PKCE verifier.
  - `manual_code_exchange_rejects_blank_code` — whitespace-only codes error out before any network request.
- `cargo nextest run -p ai grok_subscription` — all 4 tests pass.
- `cargo check -p ai` and `cargo check -p warp` — compile clean.
- Re-ran `./script/format` and `cargo check -p warp` after the UI refinements.
- Re-ran `./script/format` and `cargo check -p warp` after the review fixes.
- Removed the extra reveal step so the sign-in code input appears as soon as the connection begins. Re-ran `./script/format` and `cargo check -p warp`.
- `cargo clippy -p ai -p warp --all-features --tests -- -D warnings` — no warnings.
- `./script/format` — no changes.

UI behavior (disabled connecting state, reveal action, paste-to-submit input, and toasts) was implemented following existing API-key-editor and auth-token paste patterns; end-to-end manual verification with a live xAI account was not possible in this environment.

### Screenshots / Videos
N/A in this environment — the new row reuses existing settings input/button styling. The fallback is gated to the existing `SuperGrok` feature flag and only appears while a connect attempt is in progress.

## Agent Mode


bholmesdev:
https://github.com/user-attachments/assets/5d4b61ad-7387-486a-9d1f-12b4a76dd283

szgupta:
@bholmesdev @eluzzi5 wdyt about a simple 'paste code from clipboard' button rather than an input?

bholmesdev:
I think allowing the affordance for manual entry is good, but don't feel too strongly

eluzzi5:
@szgupta @bholmesdev I agree with Ben, I think the input field is clearer, but I would remove the step of having a button that you have to click to get there – let's just have the input field show up after the connection begins

app/src/settings_view/ai_page.rs
```diff
@@ -2193,17 +2298,15 @@ impl AISettingsPageView {
...
-                    send_telemetry_from_ctx!(
-                        TelemetryEvent::SuperGrokSubscriptionConnectFinished {
-                            error: Some("oauth_failed".to_string()),
-                        },
-                        ctx
+                        safe: ("Grok OAuth loopback callback failed; waiting for manual code entry"),
+                        full: ("Grok OAuth loopback callback failed; waiting for manual code entry: {err:#}")
                     );
-                    DismissibleToast::error(format!("Couldn't connect SuperGrok: {err}"))
+                    ctx.notify();
```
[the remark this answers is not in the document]
bholmesdev: [Warp Agent] Fixed in the latest push. I added a separate `grok_loopback_in_progress` state so loopback timeout/failure exits the disabled `Connecting` state and shows an enabled `Retry` button while keeping the manual sign-in code fallback available. The timeout path still avoids showing an error toast since the user may be expected to paste the xAI code.

app/src/settings_view/ai_page.rs
```diff
@@ -2217,6 +2320,80 @@ impl AISettingsPageView {
...
+        let code = self.grok_code_editor.as_ref(ctx).buffer_text(ctx);
+        if code.trim().is_empty() {
+            return;
+        }
+
+        ctx.spawn(
+            async move { exchange.exchange(&code).await },
+            |me, result, ctx| {
+                let window_id = ctx.window_id();
+                let toast = match result {
```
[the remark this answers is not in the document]
bholmesdev: [Warp Agent] Fixed in the latest push. The manual code completion handler now returns early when `grok_manual_code_exchange` is already cleared, so stale/double-submit completions cannot overwrite a successful loopback/manual connection or show a late failure toast.

bholmesdev:
[Warp Agent] Addressed in `b7ac6d258`: removed the extra reveal button/step, so the `Paste sign-in code` input now appears immediately once the SuperGrok connection attempt starts. Paste-to-submit and Enter-to-submit behavior are unchanged. Also updated the PR description and re-ran `./script/format` + `cargo check -p warp`.

app/src/settings_view/ai_page.rs
```diff
@@ -688,6 +688,19 @@ pub struct AISettingsPageView {
...
     custom_endpoint_edit_buttons: Vec<ViewHandle<ActionButton>>,
+
+    // SuperGrok (xAI) manual code-entry fallback. When the browser can't reach
+    // the loopback callback, xAI shows the authorization code for the user to
+    // paste back into Warp. `grok_manual_code_exchange` is `Some` only while a
+    // connect attempt is in progress, which is when the paste-code input row is
+    // shown; it captures the attempt's PKCE verifier so the pasted code can be
+    // exchanged for tokens.
+    #[cfg(not(target_family = "wasm"))]
+    grok_manual_code_exchange: Option<::ai::grok_subscription::oauth::ManualCodeExchange>,
```
szgupta: hrm, why is this in the view?

bholmesdev: [Warp Agent] Agreed — this was auth-flow state rather than pure view state. I moved the in-flight SuperGrok OAuth session into `ApiKeyManager` in `38b2b7d6b`: the manager now owns the `ManualCodeExchange`/PKCE verifier and loopback progress, and exposes a small `GrokOauthSessionState` for the settings UI. The view now only owns the paste-code editor and renders/submits based on manager state. Re-ran `./script/format`, `cargo check -p warp`, and WASM clippy.

bholmesdev:
Added the "X Premium" callout
CleanShot 2026-06-15 at 19 57 44@2x

app/src/settings_view/ai_page.rs
```diff
@@ -1748,6 +1754,46 @@ impl AISettingsPageView {
             dropdown
         });
 
+        // SuperGrok (xAI) manual code-entry fallback: a single-line input for
```
szgupta: nit: can you move this closer to the other supergrok stuff?

app/src/settings_view/ai_page.rs
```diff
@@ -688,6 +690,12 @@ pub struct AISettingsPageView {
     pending_remove_custom_endpoint_index: Option<usize>,
     custom_inference_add_button: ViewHandle<ActionButton>,
     custom_endpoint_edit_buttons: Vec<ViewHandle<ActionButton>>,
+
+    // In-flight fallback exchange for a pasted SuperGrok authorization code.
+    #[cfg(not(target_family = "wasm"))]
+    grok_oauth_attempt: Option<ManualCodeExchange>,
```
szgupta, line 698: why do we need this manual code exchange thingy? cant we just use the oauth attempt type?

app/src/settings_view/ai_page.rs
```diff
@@ -2121,13 +2176,11 @@ impl AISettingsPageView {
         /// Object id shared by the connect-flow toasts so the completion toast
         /// (success or error) automatically replaces the in-progress one.
         const CONNECT_TOAST_OBJECT_ID: &str = "grok_oauth_connect_toast";
-
```
szgupta, line 2124: nit: revert these removed lines?

app/src/settings_view/ai_page.rs
```diff
@@ -2154,7 +2207,15 @@ impl AISettingsPageView {
             }
         };
 
+        // Capture the PKCE verifier so the fallback is ready if xAI shows a
+        // code instead of redirecting.
+        self.grok_oauth_attempt = Some(attempt.manual_code_exchange());
+        self.grok_code_editor.update(ctx, |editor, ctx| {
+            editor.set_buffer_text("", ctx);
```
szgupta: nit: is there a clear buffer text api?

app/src/settings_view/ai_page.rs
```diff
@@ -2217,6 +2290,74 @@ impl AISettingsPageView {
...
+
+        use crate::view_components::DismissibleToast;
+        use crate::ToastStack;
+
+        // Shared with the browser connect-flow toasts.
+        const CONNECT_TOAST_OBJECT_ID: &str = "grok_oauth_connect_toast";
+        let Some(exchange) = self.grok_oauth_attempt.clone() else {
+            return;
+        };
+        let code = self.grok_code_editor.as_ref(ctx).buffer_text(ctx);
```
szgupta: nit: i think makes sense to materialize the buffer and pass it in as an arg to this fn

szgupta (approved):

app/src/settings_view/ai_page.rs
bholmesdev: Just looked into that. The loopback path moves the whole `OauthAttempt` into a spawned future and `finish()` consumes it, so the view can't hold onto that same value for the manual path. `OauthAttempt` also isn't `Clone` (it owns the loopback listener), and the pasted-code path only needs the PKCE verifier anyway. So `manual_code_exchange()` pulls out just that verifier as a small clonable handle. Seems like the lightweight option given the trade-offs

PR #12600, opened 2026-06-13
late async search results added to the mixer last

### LINKED ISSUE #11787: Ctrl-R command search does not prioritize command matches
### Describe the bug

An internal report says Ctrl-R command search is not prioritizing command matches as expected, making search results feel incorrect. A second internal reporter confirmed they have been seeing the same behavior for a while.

### Expected behavior

Ctrl-R command search should prioritize relevant command matches so recently or strongly matching commands are easy to find.

### Additional context

Internal Slack thread: https://warpdev.slack.com/archives/C08KTPNQN65/p1779920089771239

A linked Loom video was present in the original report, but it was not included here because it was not needed to describe the issue and was not inspected for public-safety.

acarl005:
## Description

This PR fixes an issue where command search (ctrl-r in the terminal input) usually doesn't show *any* command history items. Here is the scenario:

https://www.loom.com/share/e40d832305d542d8b93eba152072237c

To summarize the video, searching a command the appears a lot in my history, e.g. `oz-dev`, should turn up a lot of history items in the results. Instead, I only see notebooks.

Screenshot 2026-06-12 at 1 24 49 PM

I've traced this back to the behavior added in #warpdotdev/warp-internal#22987 which AFAICT is not working as intended. The intention is that "late" async results are appended to the *end* in order to avoid a re-shuffle of the results.

I have confirmed that the notebooks are returning "late" (after the timeout) in the video. However, they are actually added to the beginning, not the end as intended. You can see it in this trimmed recording. The correct results flicker very quickly and then get replaced by the notebooks

https://github.com/user-attachments/assets/3f744e63-f251-47f6-b42d-7438d44b8f1c

The problem is that the mixer results is stored in *ascending* order (highest score last). This is true whether the results are top-down or bottom-up (a separate field controls that). Underlying ordering is ascending. This means that appending to the end is actually showing the late items *first*, i.e. as the highest scoring items. This results in a total replacement of the results.

I have fixed this by prepending (inserting at the lowest-scoring edge) instead of appending late results.

## Linked Issue

Fixes #11787

### After

Screenshot 2026-06-12 at 5 12 09 PM

### DIFF
--- a/crates/warp_search_core/src/mixer.rs
+++ b/crates/warp_search_core/src/mixer.rs
@@ -205,7 +205,7 @@ impl<T: Action + Clone> SearchMixer<T> {
     ///
     /// Old results remain visible while new results are buffered. The visible result set is
     /// replaced atomically once all sources finish, or after [`INITIAL_RESULTS_TIMEOUT`] elapses.
-    /// Late-arriving async results are appended without reordering existing results.
+    /// Late-arriving async results are placed at the low-priority edge without reordering existing results.
     pub fn run_query(&mut self, query: Query, ctx: &mut ModelContext<Self>) {
         self.pending_results = Some(Vec::new());
         self.finished_sources.clear();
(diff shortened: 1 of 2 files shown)


moirahuang (approved):
really appreciate you making this fix!

PR #12605, opened 2026-06-13
Reuse code review editors on reload to fix diff flashing

### LINKED ISSUE #12606: Code Review diff flashes back to the loading skeleton on reload
## Summary
Opening the Code Review / diff panel flashes the diff content back to the gray loading skeleton and then back to content (**"content → skeleton → content"**), most noticeable right after opening the panel.

## Steps to reproduce
1. Make some changes in a git repo.
2. Open the Code Review panel.
3. Observe: the diff renders, then briefly blinks to the skeleton placeholder before reappearing.

## Root cause
`render` gates the whole view on `all_editors_loaded()`. `invalidate_all` rebuilds every editor on each `DiffStateModelEvent::NewDiffsComputed`, resetting each `is_loaded` to `false`. On open, the metadata refresh (`should_reload_diffs = true`) emits a second, identical `NewDiffsComputed`, so the already-rendered diff blanks to the skeleton until the rebuilt editors reload. (Single-file updates already reuse editors in place, so only the full-reload path flashed.)

## Proposed fix
Reconcile and reuse existing per-file editors on full reload instead of tearing the whole view down, preserving their loaded state. Implemented in #12605.

## Environment
- Platform: macOS
- Area: code review / diff view

kunish:
Closes #12606

## Problem
Opening the Code Review / diff panel flashed **"content → loading skeleton → content"**: the diff appeared, briefly reverted to the gray skeleton placeholder, then came back. The flicker was most noticeable right after opening the panel.

## Root cause
`render` gates the entire view on `all_editors_loaded()` — if any editor's `is_loaded` is `false` it renders the skeleton (`render_loading_state`) for the *whole* view. `invalidate_all` (run on every `DiffStateModelEvent::NewDiffsComputed`) tore the view down and **rebuilt every editor**, resetting each to the not-yet-loaded state.

On open this produces a redundant second pass:
1. `on_open` → initial `load_diffs_for_current_repo` → `NewDiffsComputed` → editors load → content renders.
2. The metadata refresh that fires right after open (`handle_updated_metadata_for_repo` with `should_reload_diffs = true`) emits a **second, identical** `NewDiffsComputed`.
3. `invalidate_all` rebuilds all editors → every `is_loaded` flips back to `false` → `all_editors_loaded()` returns `false` → the already-rendered diff blanks to the skeleton until the rebuilt editors reload.

Single-file updates (`SingleFileUpdated` → `update_from_single_file_diff_result`) already reuse editors in place, which is why only the full-reload path flashed.

## Fix
`invalidate_all` now **reconciles** the incoming diff against the existing file states instead of tearing the view down. New helper `reuse_file_state_if_compatible`:
- Reuses an existing editor when the file is still present, not binary, has content, and its deleted-state didn't toggle — **preserving its loaded state**.
- If the diff is byte-identical, leaves the editor untouched (covers the redundant reload on open — zero work, zero flicker).
- If the diff changed and there are no unsaved edits, refreshes it in place via the existing `apply_diff_to_code_editor(is_initial_setup = false)` path, so the base/HEAD is reset correctly on mode/branch switches. `file_diff` is only advanced once the diff has actually been applied.
- Builds a fresh (unloaded) editor only for genuinely new files; drops removed files' editors **before** `remove_deallocated_buffers` so reused global buffers aren't freed.

Because unchanged reloads keep their editors loaded, `all_editors_loaded()` stays `true` and the view never falls back to the skeleton.

## Behavior change
- **Before:** every `NewDiffsComputed` (open, refresh, mode/branch switch, redundant reload) blanked the whole diff to the skeleton, then re-rendered.
- **After:** unchanged reloads keep their content on screen; changed files refresh in place; only genuinely new files momentarily show the skeleton.

## Demonstration
The flash is a sub-frame "content → skeleton → content" flicker on the redundant post-open reload. Captured from a real-display integration run that opens the Code Review panel and then forces the same redundant full reload — the frame captured *during* the reload tells the story:

- **Before (master):** the reload rebuilds every editor, so the panel flashes back to the **"Loading open changes…"** skeleton.
- **After (this PR):** the editors are reused, so the diff content stays on screen — no flash.

https://github.com/user-attachments/assets/286b6769-077c-4d1c-bd92-40e377b7c35c

https://github.com/user-attachments/assets/98f81880-85a5-4fd4-b1c2-922cad7f67b6

## Scope / non-goals
- The intentional "skeleton until loaded" behavior on **first** load and for **genuinely new** files is preserved — this does **not** reintroduce the empty-editor flash that the skeleton was originally added to fix.
- Non-global-buffer mode (editors created via `new_loaded`, already `is_loaded = true`) never flashed and is unaffected.
- Unsaved in-editor edits are never clobbered (the reused editor and its `file_diff` are left untouched), mirroring the single-file path.

## Testing
- `cargo check -p warp --tests` — passes.
- `cargo test -p warp --lib code_review` — 119 passed, 0 failed (no regression in the incremental/single-file path).
- `cargo test -p warp --lib reuse_file_state` — 4 unit tests pass:
  - identical diff → editor reused and loaded state preserved,
  - file became binary → rebuild (returns `None`),
  - deleted-state toggled → rebuild (returns `None`),
  - unsaved changes + changed diff → editor reused, diff not re-applied, stale `file_diff` retained.

Beyond the reuse-decision unit tests, the fix is demonstrated end-to-end by the before/after capture in **Demonstration** above: a real-display integration run opens the Code Review panel and forces the redundant reload, confirming the diff no longer blinks to the gray placeholder. Manual check: open the Code Review panel on a repo with changes and confirm the diff stays put during the post-open metadata refresh.


kunish:
@cla-bot check

kunish:
Thanks for the review! Addressed all three points in `297ac55c`:

1. **`file_diff` could desync when `apply_diff_to_code_editor` is skipped** — `file_diff` is now only advanced *after* the diff has actually been applied to the editor, so it can no longer get ahead of the editor's contents (including the `repo_path() == None` path).
2. **Split list-item accounting** — `build_view_state_for_file_diffs` is now side-effect-free; all viewported-list accounting lives in the callers. `invalidate_all` adds one item per entry in `file_states_vec` in a single place, and both `update_from_single_file_diff_result` arms add explicitly. Item count/order are unchanged.
3. **Missing `has_unsaved_changes` coverage** — added `test_reuse_file_state_preserves_unsaved_editor_without_clobbering`: it marks the editor dirty, feeds a changed `FileDiffAndContent`, and asserts the editor is reused *without* re-applying the diff and that the stale `file_diff` is retained.

Verification: `cargo test -p warp --lib code_review` → 119 passed, 0 failed (no regression in the incremental path); `cargo test -p warp --lib reuse_file_state` → 4/4 pass.

kunish:
Good catch on the stale per-file **Discard** button — confirmed and fixed in 90733282.

**The bug:** the reuse path kept the whole `FileState`, including its `discard_button`, which bakes in `git_operation_blocked` at build time (disabled + no click handler while blocked; a live `ShowDiscardConfirmDialog` dispatch otherwise). A full reload that lands after a merge/rebase starts reuses unchanged files and keeps the stale button, so it can dispatch a discard while git operations should be blocked. (It's also asymmetric: a button built *while* blocked stores no click handler, so a later unblock would leave it permanently inert — `set_disabled` alone can't fix both directions.)

**The fix:** extracted `build_discard_button` and rebuild it against the current `git_operation_blocked` on every reuse, re-deriving the disabled flag and the click handler together. The button carries no loaded state, so rebuilding it doesn't touch `all_editors_loaded` and doesn't reintroduce the skeleton flash this PR fixes (that came from rebuilding *editors*). I audited the other reused `FileState` fields — `chevron_button`, `open_in_tab_button`, `add_context_button`, `copy_path_button` capture only the file's own path, and `is_expanded` is carried over intentionally — so `discard_button` was the only field with this ambient-state hazard.

Added `test_reuse_file_state_rebuilds_discard_button` (the reused file state must get a fresh Discard button). `cargo test -p warp --lib code_review` → 120 passed, 0 failed.

kunish:
Right again — fixed in 77b9e729. Same class of stale-capture as the Discard button.

The per-file **Open file** button captures the diff's first changed line (`file_line_for_open`) at build time. When a reload changes the diff for an unchanged-path file, `reuse_file_state_if_compatible` re-applies the new diff to the editor but kept the old button, so it would navigate to the previous diff's location.

**Fix:** extracted `build_open_in_tab_button` and rebuild it from the current diff on every reuse, alongside the Discard button. Like the Discard button it has no loaded state, so rebuilding it doesn't touch `all_editors_loaded` or reintroduce the skeleton flash.

I audited the remaining reused `FileState` buttons to be sure this closes the class: **chevron**, **add-context**, and **copy-path** capture only the file's own path — and the path is the reuse key, so it's invariant across a reuse — while `is_expanded` is intentionally carried over. So Discard (ambient git state) and Open file (diff-derived line) were the only two buttons capturing mutable state; both are now rebuilt.

Added `test_reuse_file_state_rebuilds_open_in_tab_button`. `cargo test -p warp --lib code_review` → 121 passed, 0 failed.

PR #12127, opened 2026-06-03
Add Chinese and English localization

WolffyCode:
## Linked Issue

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Testing

- [ ] I have manually tested my changes locally with `./script/run`

## Agent Mode
- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

Issue #1194, opened 2022-04-26
Application language switch


Quinlivanner:
### Discord username (optional)

Quickk#0001

### Describe the solution you'd like?

```text
Provide other language options for application and optimize for language localization.
(It is the interface language, not the language content support), allowing users to choose their own applicable language in the Setting, such as Chinese, so that users can better experience the product and find problems.
```

### Is your feature request related to a problem? Please describe.

```text
No, I'm not currently having problems
```

77321660:
+1

mingzaily:
+1

MacAndTea:
+1

fby0394:
Will there be a Chinese version in the future?

Quinlivanner:
> Will there be a Chinese version in the future?

Tbh i think yes if have enough check

ujjboy:
+1

dannyneira:
Two more users have expressed interest in Chinese UI support.

wenhongquan:
+1

tree-white:
+1 ！！！

suhengli:
+1

nu0l:
+1

Fanlu91:
+1

Chuhan1112:
+1

BusyHenryMa:
+1

PPPPigggg:
+1

jsonpip:
+1

nu0l:
+1

sancppp:
+1

shdjllajksd:
+1

zhangyongjie1997:
+1

zy-kkk:
+1

Gary1567:
+1

zhengjianyang:
+1

StringKe:
+1

303snowing:
+1

LucaHhx:
+1

onburn:
+1

qiaohui:
+1

zhengjianyang:
+1

qiaohui:
+1

acanyo:
> 还有两名用户表示对中文 UI 支持感兴趣。



> Two more users have expressed interest in Chinese UI support.

May I ask approximately how long it will take to implement? Many people in China are fond of this software!

Liu-Da:
+1

DevKiDMediaCreation:
[...]

PR #12607, opened 2026-06-13
Add Chinese and English localization

CyJaySong:
for https://github.com/warpdotdev/warp/issues/1194
fork form https://github.com/warpdotdev/warp/pull/12127
Because the original author not yet agreed CLA， I forked its branch and initiated a new PR


CyJaySong:
So verbose

Issue #11282, opened 2026-05-19
Phantom characters rendered at start of command when running agent code snippet


andrewjschmitt:
## Summary
When running a command from an agent prompt code snippet, phantom characters (e.g. "PK") appear visually at the very beginning of the rendered command output. These characters are not part of the actual command or its output—they are purely a rendering artifact—but they are visible in the terminal display and can confuse the agent when it reads the output.

## Problem
After the agent suggests a multi-line shell command in a code snippet and the user chooses to run it, the terminal renders extraneous characters (observed as "PK" with a distinct highlight/background color) at the very start of the command text. The characters do not exist in the command string itself—the command executes correctly—but the visual artifact is confusing and, more importantly, can cause the agent to misinterpret subsequent output when it reads back what was executed.

## Expected behavior
The rendered command output should exactly match the command that was executed, with no phantom or extra characters prepended.

## Actual behavior
Two extra characters ("PK" in the observed case) appear highlighted at the beginning of the command display. They are not part of the command and do not affect execution, but they are visible in the terminal and can mislead the agent.

## Reproduction steps
1. Start a conversation with the Warp agent (Oz).
2. Have the agent suggest a multi-line shell command in a code snippet (e.g. a `zip` + `aws lambda` deploy sequence).
3. Choose to run the command from the agent prompt code snippet.
4. Observe the rendered output: phantom characters appear at the very beginning of the command text, highlighted differently from the rest of the command.

## Artifacts
**Screenshot 1:** Terminal output after running an agent-suggested multi-line shell command, showing "PK" highlighted at the start of the command text that was not part of the original command.

<img width="184" height="51" alt="Image" src="https://github.com/user-attachments/assets/f0bccb04-6669-4272-9609-04ecfc25781a" />

## Warp version
v22.19.0 (visible in terminal prompt)

## Operating system
macOS

andrewjschmitt:
##Warp's using the omz shell
```version master (a071263)```

##Here's the snipped we clicked on the run from the code snippet
```
PKG="$(mktemp -d)/redacted.zip" && \
cd /Users/andrewschmitt/Documents/GitHub/redacted/backend && \
zip -qr "${PKG}" dist/ node_modules/ && \
aws lambda update-function-configuration --no-cli-pager --profile redacted --region us-east-1 \
  --function-name redacted --runtime nodejs22.x \
  --query 'Runtime' --output text && \
[...]

Issue #4069, opened 2024-01-14
Warp INSERTS BOGUS CHARACTERS to the displayed command and its output in some cases


doga:
### Describe the bug

Warp runs the shell commands correctly, but adds bogus characters to the command and its output in some well-defined cases:

1. When displaying a multi-line command, _Warp_ inserts a bogus character at the beginning.
2. When displaying a command output that does not end with an empty line,  _Warp_ inserts a bogus character at the end.

### To reproduce

Run this command in Warp:

```shell
curl \
 'https://qworum.net/languages.json'
```

See the bogus characters that Warp displays in the command (`c` at the beginning) and its output (`%` at the end):

```text
ccurl \
 'https://qworum.net/languages.json'
["en"]%
```

### Expected behavior

Warp should not display bogus characters that were absent in the command and its output:

```text
curl \
 'https://qworum.net/languages.json'
["en"]
```

### Screenshots

<img width="352" alt="Screenshot 2024-01-14 at 12 29 02" src="https://github.com/warpdotdev/Warp/assets/212690/ad292c35-4910-482b-8700-095cf5f7120a">

### Operating system

MacOS

### Operating system and version

12.7.2

### Shell Version

zsh 5.8.1 (x86_64-apple-darwin21.0)

### Current Warp version

v0.2024.01.09.08.02.stable_02

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this a Warp specific issue? (i.e. does it happen in Terminal, iTerm, Kitty, etc.)

- _Only in Warp_: The bogus char at the beginning of the command.
-  _Probably in all terminals (verified in macOS Terminal + zsh)_:  The bogus char at the end of the command output.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

dannyneira:
hey @doga thanks for submitting this. I've seen this kind of issue before, and it's typically caused by an unsupported plugin in your `~/.zshrc` file. The most common culprit is the iterm2_shell_integration, but we have a longer list of them here: https://docs.warp.dev/help/known-issues#list-of-incompatible-tools

Once you find it, you can disable parts of your dotfiles just for Warp by using this conditional statement:
```
# Bash and Zsh
if [[ $TERM_PROGRAM != "WarpTerminal" ]]; then
# > What you want to disable here (e.g. iterm shell integration or others) <
fi
```

[...]

PR #12619, opened 2026-06-14
Fix duplicated prefix in rendered multiline command blocks

### LINKED ISSUE #5024: Duplicate characters added in block's command render
### Dupe Check

- [X] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates

### Describe the bug

When running multi-line commands in my shell (zsh), I consistently notice rendering issues with how the command is shown in the command/output block once it's run

### To reproduce

1. Run a multi-line command
2. Observe how the command is rendered when the command & output are put in a block in the shell's history

### Expected behavior

I would expect the command portion of the previous blocks to only render the command I entered, without extra characters

### Screenshots

Entering a command

First two characters duplicated in "command" section of a block in the shell's history

### Operating system

MacOS

14.5.0

### Shell Version

zsh 5.9 (x86_64-apple-darwin23.0)

### Current Warp version

v0.2024.06.11.08.02.stable_01

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

MX48Live:
## Description

Fixes a terminal model rendering bug where completed multiline command blocks can show duplicated leading characters, such as `eecho ...`, `ccurl ...`, `mmkdir ...`, or `pprintf ...`, even though the shell receives and executes the correct command.

## Problem

When a user runs a multiline command, Warp sometimes shows extra characters at the start of the command in the completed block. The command itself still runs correctly, but the saved block can look wrong. For example, a command that should display as `echo ...` can appear as `eecho ...`, or `curl ...` can appear as `ccurl ...`.

That is confusing because the visual command history no longer matches what the user actually ran. It is especially risky for agent workflows, where the agent may read the completed block and treat the extra characters as if they were part of the command.

The duplicate characters are not coming from the bytes Warp sends to the shell. They come from the shell's line editor/redraw output while Warp is building the command grid for the completed block.

## Solution

Use the command reported by `preexec` as the source of truth for what the shell is about to execute, but only after the command grid has been finalized and only when the grid clearly contains a redraw artifact.

In practice, this means Warp now checks for one narrow shape: the finished command grid must look like a short duplicated prefix followed by the exact `preexec` command. If that shape is present, Warp rebuilds the prompt and command grid using the existing prompt boundary and the canonical `preexec` command.

The fix does not implement a cosmetic "drop first characters" rule. Commands that legitimately start with repeated characters, such as `asdf\nasdf`, are preserved.

## Linked Issue

Closes #5024.

Related evidence:

- #11282 reports phantom characters at the start of agent-run multiline commands.
- #4069 reports bogus leading characters for displayed multiline commands.

## Implementation Notes

- Finish the command grid at `preexec` before inspecting it, so multiline command text is fully available.
- Detect only the narrow redraw-artifact shape: finished command-grid text must be longer than the `preexec` command, must end with the exact `preexec` command, and the extra leading text must be a prefix of that same command.
- Rebuild the prompt and command grid from the existing prompt demarcation plus the canonical `preexec` command when that narrow condition holds.
- Preserve prompt/command boundaries and avoid changing shell input bytes.

## Testing

Automated:

- [x] `cargo test --manifest-path Cargo.toml -p warp multiline_preexec -- --nocapture`
- [x] `cargo nextest run --manifest-path Cargo.toml --no-fail-fast -p warp -E 'test(multiline)'`
- [x] `./script/format`
- [x] `./script/check_no_inline_test_modules`
- [x] `cargo clippy --manifest-path Cargo.toml -p warp --tests -- -D warnings`

New regression coverage:

- `test_multiline_preexec_reconciles_command_grid_redraw_prefix`
- `test_multiline_preexec_reconciliation_preserves_prompt_demarcation`
- `test_multiline_preexec_preserves_legitimate_repeated_command_prefix`

Manual repro cases:

1. Basic multiline echo:

```sh
echo "line one" && \
echo "line two"
```

Expected: the completed block renders `echo "line one" ...`, not `eecho ...` or `ececho ...`.

BEFORE 

AFTER

2. Valid repeated-prefix guard:

```sh
asdf
asdf
```

Expected: the command may fail with `command not found`, which is fine. The completed block should still render the command exactly as entered and should not remove the leading `a`.

BEFORE

AFTER

3. Curl-style multiline command from related reports:

```sh
curl \
'https://example.com'
```

Expected: the completed block renders `curl ...`, not `ccurl ...`.

BEFORE

AFTER

4. Zip/deploy-style multiline command:

```sh
mkdir -p /tmp/warp-multiline-repro && \
printf "hello from warp\n" > /tmp/warp-multiline-repro/index.txt && \
cd /tmp/warp-multiline-repro && \
zip -r repro.zip index.txt && \
ls -lh repro.zip
```

Expected: the completed block starts with `mkdir`, with no duplicated or phantom prefix.

BEFORE

AFTER

5. Longer pipeline:

```sh
printf "alpha\nbeta\ngamma\n" | \
grep "beta" | \
awk '{ print "matched:", $0 }'
```

Expected: the completed block starts with `printf`, with no extra leading characters.

BEFORE

AFTER 

### Screenshots / Videos

Full `./script/presubmit` was attempted locally. It failed only on SSH/GCloud integration tests (`test_ssh_into_ash`, `test_ssh_into_sh`, `test_ssh_wrapper_into_bash`, `test_ssh_wrapper_into_zsh`, `test_ssh_with_shell_override`) because this machine does not have an active GCloud account/access to Warp's `warp-ssh-integration-testing` project.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed duplicated leading characters in rendered multi-line command blocks.

### DIFF
--- a/app/src/terminal/model/block_tests.rs
+++ b/app/src/terminal/model/block_tests.rs
@@ -1296,6 +1296,109 @@ fn test_clone_command_from_blockgrid_to_empty() {
     assert_eq!(new_content, "clone1\nclone2\nclone3");
 }
 
+#[test]
+fn test_multiline_preexec_reconciles_command_grid_redraw_prefix() {
+    let cases = [
+        (
+            "ececho \"line one\" && \\\r\necho \"line two\"",
+            "echo \"line one\" && \\\necho \"line two\"",
+        ),
+        ("aasdf\r\nasdf", "asdf\nasdf"),
+    ];
+
+    for (command_grid_text, reported_command) in cases {
+        let prompt_and_command_grid = mock_blockgrid(command_grid_text);
+        let mut rprompt_grid = mock_blockgrid("");
+        rprompt_grid.finish();
+        let mut output_grid = mock_blockgrid("");
+        output_grid.finish();
+
+        let mut block = create_test_block_with_grids(
+            BlockIndex::zero(),
+            prompt_and_command_grid,
+            rprompt_grid,
+            output_grid,
+            false, /* honor_ps1 */
+        );
+        block.set_honor_ps1(false);
+
+        block.preexec(PreexecValue {
+            command: reported_command.to_owned(),
+            session_id: None,
+        });
+
+        assert_eq!(block.command_to_string(), reported_command);
+        assert_eq!(
+            block.prompt_and_command_with_secrets_unobfuscated(false),
+            reported_command
+        );
+    }
+}
+
+#[test]
+fn test_multiline_preexec_reconciliation_preserves_prompt_demarcation() {
+    let prompt = "warp> ";
+    let reported_command = "echo \"line one\" && \\\necho \"line two\"";
+    let mut block = TestBlockBuilder::new().with_honor_ps1(true).build();
+    block.prompt_marker(ansi::PromptMarker::StartPrompt {
+        kind: ansi::PromptKind::Initial,
+    });
+    for c in prompt.chars() {
+        block.input(c);
+    }
+    block.prompt_marker(ansi::PromptMarker::EndPrompt);
+    block.start();
+    for c in "ececho \"line one\" && \\".chars() {
+        block.input(c);
+    }
+    block.carriage_return();
+    block.linefeed();
+    for c in "echo \"line two\"".chars() {
+        block.input(c);
+    }
+
+    block.preexec(PreexecValue {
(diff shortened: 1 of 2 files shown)


MX48Live:
@cla-bot check

zachbai (approved):

PR #12623, opened 2026-06-14
Cancel LRC conversations after Stop takeover

vkodithala:
## Description

This fixes a bug originally reported by ZL [here](https://warpdev.slack.com/archives/C09769R5GBT/p1781279811994189) with a pretty trivial repro: tell an agent to run a long-running command, `ctrl-c` to exit the subagent, then `ctrl-c` again to cancel the command. Notice that a "Warping..." is still visible in both agent and terminal views, and doesn't disappear until the next user query.

The bug here is that when a user takes control of an agent-monitored LRC, the conversation is intentionally marked `InProgress` with the assumption that they'll hand back control to the CLI subagent. When the user cancels the command itself, we don't cancel the conversation's progress, which we should do since user no longer has the ability to resume the conversation in place; they must send a follow-up query.

This PR:

- Treats Ctrl-C in a Stop/user-control LRC state as conversation cancellation. As documented in test comments, we intentionally avoid applying this behavior to takeovers initiated by `TransferShellCommandToUser` since cancellation events here emit a server tool call result that by default resumes the conversation, hence the "in-progress" state is correct even after command cancellation.
- Still writes Ctrl-C to the PTY so the underlying command is interrupted.
- Preserves transfer/manual user-control behavior by sending Ctrl-C to the PTY without cancelling the conversation.
- Adds focused regression coverage for Stop takeover vs. transfer takeover.

This is the first PR in a two-PR stack. The follow-up PR handles the agent-requested command focus / early Ctrl-C race.

## Testing

Demo of how this works end-to-end, along with a short explainer: https://www.loom.com/share/4654d825c946467ebaf92d800b7e8926.

- [x] `./script/format`
- [x] `git --no-pager diff --check`
- [x] `cargo nextest run -E 'test(ctrl_c_after_stop_takeover_cancels_conversation) | test(ctrl_c_after_transfer_takeover_does_not_cancel_conversation)'`
- [ ] `cargo clippy --workspace --all-targets --all-features --tests -- -D warnings` — not rerun after splitting; earlier attempt in this checkout was blocked in `app/build.rs` while downloading Sentry.xcframework because `gh` returned `Bad credentials (HTTP 401)`.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed an issue where pressing Ctrl-C to stop a command monitored by Warp Agent could leave the conversation stuck showing `Warping…`.

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -2376,6 +2376,12 @@ pub struct TerminalViewStateChange {
     pub state: TerminalViewState,
     pub timestamp: Instant,
 }
+#[derive(Clone, Copy)]
+struct CtrlCActiveBlockState {
+    is_long_running: bool,
+    is_agent_in_control_of_command: bool,
+    conversation_id_to_stop: Option<AIConversationId>,
+}
 
 impl Default for TerminalViewStateChange {
     fn default() -> TerminalViewStateChange {
@@ -8496,23 +8502,31 @@ impl TerminalView {
     /// Windows users expect ctrl-c to copy if there is selected text. Otherwise,
     /// we perform the normal ctrl-c action.
     fn ctrl_c(&mut self, ctx: &mut ViewContext<Self>) {
-        let (
-            has_block_list_selection,
-            has_alt_screen_selection,
-            is_long_running,
-            is_agent_in_control_of_command,
-        ) = {
+        let (has_block_list_selection, has_alt_screen_selection, active_block_state) = {
             let model = self.model.lock();
             let has_alt_screen_selection = model.alt_screen().selection().is_some();
             let has_block_list_selection = model.block_list().selection().is_some();
             let active_block = model.block_list().active_block();
             let is_long_running = active_block.is_active_and_long_running();
             let is_agent_in_control_of_command = active_block.is_agent_in_control();
+            let conversation_id_to_stop = active_block
+                .long_running_control_state()
+                .and_then(|state| {
+                    state
+                        .user_take_over_reason()
+                        .is_some_and(UserTakeOverReason::is_stop)
+                        .then(|| active_block.ai_conversation_id())
+                })
+                .flatten();
+            let active_block_state = CtrlCActiveBlockState {
+                is_long_running,
+                is_agent_in_control_of_command,
+                conversation_id_to_stop,
+            };
             (
                 has_block_list_selection,
                 has_alt_screen_selection,
-                is_long_running,
-                is_agent_in_control_of_command,
+                active_block_state,
             )
         };
         // We don't want to copy blocks in AI input mode because those are
@@ -8524,8 +8538,7 @@ impl TerminalView {
             has_copiable_block_selection,
             has_block_list_selection,
             has_alt_screen_selection,
-            is_long_running,
-            is_agent_in_control_of_command,
+            active_block_state,
             ctx,
         );
 
@@ -8542,8 +8555,7 @@ impl TerminalView {
         has_copiable_block_selection: bool,
         has_block_list_selection: bool,
         has_alt_screen_selection: bool,
-        is_long_running: bool,
-        is_agent_in_control_of_command: bool,
+        active_block_state: CtrlCActiveBlockState,
         ctx: &mut ViewContext<Self>,
     ) {
         if has_block_list_selection {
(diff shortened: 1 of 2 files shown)


vkodithala:
* **#12667** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12667?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#12617](https://github.com/warpdotdev/warp/pull/12617) <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12617?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#12701](https://github.com/warpdotdev/warp/pull/12701) <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12701?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#12623** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12623?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12623?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/terminal/view.rs
```diff
@@ -8563,7 +8563,25 @@ impl TerminalView {
...
+                            .then(|| active_block.ai_conversation_id())
+                    })
+                    .flatten()
+            };
+
+            if let Some(conversation_id) = conversation_id_to_stop {
+                self.stop_local_agent_conversation(conversation_id, ctx);
+            } else {
+                self.user_write_ctrl_c_to_pty(ctx);
+            }
```
vkodithala: two nits:

we should add better comments around why we've implemented this. the concrete case here is: a user cancels an agent-initiated LRC that they've taken control of, then the conversation progress should be moved to cancelled (since it is, and the user has no way of routing back to it since follow-up queries 'resume' that conversation) AND the command should be cancelled, both of which stop_local_agent_conversation does.
there's some weirdness around cloud conversations that i'd like you to investigate, where if i cancel an LRC in a cloud conversation it auto-sends a "resume" query as the next message. is there something funky going on here? since "resume" is a follow-up i'm assuming our logic here still works fine (conv progress will be re-set on the resume anyways), but maybe worth a brief smoke test?

vkodithala: update - resolved comment nit on latest, and then my concern around auto-resume for cloud convs is not malicious since the cancelled conversation would be resumed via the default path anyways

app/src/terminal/view_tests.rs
```diff
@@ -5211,6 +5211,121 @@ fn inline_agent_view_exits_when_tagged_in_long_running_command_is_tagged_out() {
...
+            let conversation = BlocklistAIHistoryModel::as_ref(ctx)
+                .conversation(&conversation_id)
+                .expect("conversation should exist");
+            assert_eq!(conversation.status(), &ConversationStatus::Cancelled);
+        });
+    })
+}
+
+#[test]
+fn ctrl_c_after_transfer_takeover_does_not_cancel_conversation() {
```
vkodithala, line 5270: nit: can you document _why_ we don't cancel the conversation on transfer takeover? i think it's because we have a waiting channel/sender for TransferShellCommand that, if we cancel a terminal command, sends a result back to the server to allow it to continue/retry. but, verify with code.

app/src/terminal/view.rs
```diff
@@ -8628,7 +8628,28 @@ impl TerminalView {
                 controller.switch_control_to_user(UserTakeOverReason::Stop, ctx);
             });
         } else if is_long_running {
-            self.user_write_ctrl_c_to_pty(ctx);
+            // After cancelling an agent-initiated long-running command, the conversation
+            // itself is stopped and cannot be re-entered until the next user query, so
+            // Ctrl+C should interrupt the command and cancel conversation progress.
+            let conversation_id_to_stop = {
```
MaggieShan: nit: I wonder if it'd be cleaner to pass this in from `ctrl_c` -> `ctrl_c_internal` -> `ctrl_c_to_active_block` since `ctrl_c` already accesses the active block

MaggieShan (approved):
Tested and this LGTM!

app/src/terminal/view.rs
vkodithala: fixed in latest - good callout!

PR #12663, opened 2026-06-15
Fail fast on PTY spawn errors with specific error messages

seemeroland:
## Description

Resolves REMOTE-1935 https://linear.app/warpdotdev/issue/REMOTE-1935/investigate-terminal-bootstrapping-failures
- Note it was one user who tried multiple times

When the terminal server fails to spawn a shell (e.g. due to E2BIG / "Argument list too long"), Warp previously fell back to a direct spawn and then waited the full 60 s bootstrap timeout before surfacing a generic "Terminal session failed to start. Please try running your task again." error. This makes cloud agent failures from oversized environments take over a minute and give the user no actionable information.

This error can happen when env vars or secrets are too long. The syscall `execve` fails when spawning the initial shell with `/bin/bash -c "exec -a bash '/bin/bash' --rcfile <(echo '<bootstrap_script>')"`.

If the shell fails to spawn, we immediately return the error instead of waiting the full timeout, and give a more actionable error message. We also log some diagnostic information to see what is so large.

## Testing
 
Reproduced the original error by starting an oz run with a really long secret.

After the fix, see error message and logs
Screenshot 2026-06-15 at 6 37 42 PM

```
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner] Shell spawn env var diagnostics (names and sizes only, no values):
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]   PtyOptions env_vars (9 entries):
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "roland-long-fake-secret" — 1500025 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "OZ_CLI" — 89 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WARP_FOCUS_URL" — 68 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WARP_TERMINAL_SESSION_UUID" — 60 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WARP_SESSION_SHARING_SERVER_URL" — 52 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WARP_WS_SERVER_URL" — 50 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "OZ_RUN_ID" — 47 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WARP_SERVER_ROOT_URL" — 43 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "OZ_HARNESS" — 14 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]   Inherited process env (15 vars, ~1236 bytes total):
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "PATH" — 630 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WARP_WORKLOAD_TOKEN" — 85 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WARP_API_KEY" — 83 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "GITHUB_ACCESS_TOKEN" — 61 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "TMPDIR" — 57 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "GH_TOKEN" — 50 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "OZ_RUN_ID" — 47 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "TASK_ID" — 45 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "__CF_USER_TEXT_ENCODING" — 38 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WITH_LOCAL_SESSION_SHARING_SERVER" — 36 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "HOME" — 24 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "GIT_TERMINAL_PROMPT" — 22 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "GH_PROMPT_DISABLED" — 21 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "WITH_LOCAL_SERVER" — 20 bytes
2026-06-16T01:36:46Z [ERROR] [warp::terminal::local_tty::spawner]     "LANG" — 17 bytes
2026-06-16T01:36:46Z [ERROR] [errors::report_error] Failed to spawn pty via terminal server; falling back to spawning locally...: Received error reading message back from terminal server
2026-06-16T01:36:46Z [INFO] [warp::terminal::local_tty::unix] Starting shell /bin/zsh
2026-06-16T01:36:46Z [INFO] [warp::terminal::local_tty::unix] Successfully spawned child zsh process with pid 57986
2026-06-16T01:36:46Z [INFO] [warp::terminal::model::terminal_model] Tried to exit the alternate screen, but it was already inactive
2026-06-16T01:36:46Z [INFO] [warp::terminal::model::block] Block finished with new state DoneWithNoExecution
2026-06-16T01:36:46Z [INFO] [warpui_core::core::app] dispatching typed action: warp::pane_group::PaneGroupAction::HandleFocusChange
2026-06-16T01:36:46Z [WARN] [warpui_core::core::app] Action HandleFocusChange was dispatched, but no view handled it
2026-06-16T01:36:46Z [ERROR] [warp::ai::agent_sdk::driver::terminal] Terminal bootstrap failed: Shell spawn failed. This can happen when env vars or secrets are too long — check your image for excessively long environment variables or Oz for excessively long secrets.
```

### DIFF
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -436,7 +436,7 @@ pub enum AgentDriverError {
     NotLoggedIn,
     #[error("Saved prompt not found for id {0}")]
     AIWorkflowNotFound(String),
-    #[error("Terminal bootstrap failed: {error}")]
+    #[error("Terminal bootstrap failed")]
     BootstrapFailed {
         #[source]
         error: terminal::BootstrapError,
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -436,8 +436,11 @@ pub enum AgentDriverError {
     NotLoggedIn,
     #[error("Saved prompt not found for id {0}")]
     AIWorkflowNotFound(String),
-    #[error("Terminal bootstrap failed: {reason}")]
-    BootstrapFailed { reason: String },
+    #[error("Terminal bootstrap failed: {error}")]
+    BootstrapFailed {
+        #[source]
+        error: terminal::BootstrapError,
+    },
     #[error("Unable to share agent session")]
     ShareSessionFailed {
         #[source]
@@ -1835,6 +1838,7 @@ impl AgentDriver {
                     })
                     .await?
                     .await
+                    .map_err(|error| AgentDriverError::BootstrapFailed { error })
             })
             .await?;
 
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -436,8 +436,8 @@ pub enum AgentDriverError {
     NotLoggedIn,
     #[error("Saved prompt not found for id {0}")]
     AIWorkflowNotFound(String),
-    #[error("Terminal bootstrap failed")]
-    BootstrapFailed,
+    #[error("Terminal bootstrap failed: {reason}")]
+    BootstrapFailed { reason: String },
     #[error("Unable to share agent session")]
     ShareSessionFailed {
         #[source]
@@ -1831,8 +1831,7 @@ impl AgentDriver {
                 foreground
                     .spawn(|me, ctx| {
                         me.terminal_driver
-                            .as_ref(ctx)
-                            .wait_for_session_bootstrapped()
+                            .update(ctx, |driver, _| driver.wait_for_session_bootstrapped())
                     })
                     .await?
                     .await
--- a/app/src/ai/agent_sdk/driver/terminal.rs
+++ b/app/src/ai/agent_sdk/driver/terminal.rs
@@ -578,7 +578,7 @@ impl TerminalDriver {
                     // PTY spawn failed — use the specific reason from the view.
                     futures::future::Either::Right((Ok(reason), _)) => reason,
                     // Sender dropped without sending (shouldn't happen in practice).
-                    futures::future::Either::Right((Err(_), _)) => return Ok(()),
+                    futures::future::Either::Right((Err(_), _)) => "Internal error".to_string(),
                 }
             } else {
                 // No failure channel — wait for bootstrap with the standard timeout.
@@ -731,10 +731,7 @@ impl TerminalDriver {
                 if !self.session_bootstrapped.is_set() {
                     if let Some(tx) = self.spawn_failure_tx.take() {
                         let _ = tx.send(
-                            "Shell spawn failed. This can happen when env vars or secrets \
-                             are too long — check your image for excessively long \
-                             environment variables or Oz for excessively long secrets."
-                                .to_string(),
+                            "Shell spawn failed. Check the Warp logs for details.".to_string(),
                         );
                     }
                 }
--- a/app/src/ai/agent_sdk/driver/terminal.rs
+++ b/app/src/ai/agent_sdk/driver/terminal.rs
@@ -731,10 +731,10 @@ impl TerminalDriver {
                 if !self.session_bootstrapped.is_set() {
                     if let Some(tx) = self.spawn_failure_tx.take() {
                         let _ = tx.send(
-                            "The shell process exited before the Warp bootstrap script \
-                             completed. Check any environment variables or secrets \
-                             configured for this run."
-                            .to_string(),
+                            "Shell spawn failed. This can happen when env vars or secrets \
+                             are too long — check your image for excessively long \
+                             environment variables or Oz for excessively long secrets."
+                                .to_string(),
                         );
                     }
                 }
--- a/app/src/ai/agent_sdk/driver/terminal.rs
+++ b/app/src/ai/agent_sdk/driver/terminal.rs
@@ -724,6 +724,21 @@ impl TerminalDriver {
                     let _ = tx.send(reason.clone());
                 }
             }
+            crate::terminal::view::Event::Exited => {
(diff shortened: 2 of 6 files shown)


app/src/ai/agent_sdk/driver/terminal.rs
```diff
@@ -539,27 +548,54 @@ impl TerminalDriver {
...
-                    AgentDriverError::BootstrapFailed
-                })
-                .instrument(tracing::info_span!(
-                    "wait_for_session_bootstrapped",
-                    tags.cloud_agent = true
-                ))
-                .await
+            let reason = if let Some(failure_rx) = spawn_failure_rx {
+                // Race: bootstrap success vs. early PTY-spawn failure.
+                let success_future = session_bootstrapped
```
vorporeal: i feel like the modeling here isn't quite right.  can we merge the failure notification path into the `session_bootstrapped()` future?  like, i think it should probably return something like `Result<Result<(), PtySpawnError>, TimeoutError>`.

we'd probably need to turn `self.session_bootstrapped` into a channel receiver instead of a condition variable, but this feels more correct.

app/src/ai/agent_sdk/driver.rs
```diff
@@ -436,8 +436,8 @@ pub enum AgentDriverError {
     NotLoggedIn,
     #[error("Saved prompt not found for id {0}")]
     AIWorkflowNotFound(String),
-    #[error("Terminal bootstrap failed")]
-    BootstrapFailed,
+    #[error("Terminal bootstrap failed: {reason}")]
+    BootstrapFailed { reason: String },
```
vorporeal: can we store a structured error here instead of flattening to a `String` before this point?  if not, that's fine, but i think we are too much in the habit of "let's turn this nice structured error into a string" in a way that removes the ability for a consumer to decide on error formatting.

vorporeal (commented):
couple questions:
* should we set some limit on secret size?  if we know that a secret >XXkb will break things, that should probably be a validation error at secret creation time, yeah?
* should we use some other method of passing secrets, to avoid the limitations?

---

approving because this is fine as-is, but i think unifying `session_bootstrapped` and the new channel would make things clearer.

seemeroland:
> should we set some limit on secret size? if we know that a secret >XXkb will break things, that should probably be a validation error at secret creation time, yeah?

Yea, https://linear.app/warpdotdev/issue/REMOTE-1958/add-validation-to-prevent-secrets-128kb

> should we use some other method of passing secrets, to avoid the limitations?

I doubt there are real cases we need an env var > 128kb

app/src/ai/agent_sdk/driver/terminal.rs
```diff
@@ -39,6 +36,48 @@ use crate::terminal::view::{ConversationRestorationInNewPaneType, Event};
...
+                 Check the Warp logs for details."
+            ),
+            BootstrapError::InternalError => {
+                write!(f, "An unexpected internal error occurred during bootstrap.")
+            }
+        }
+    }
+}
+
+impl std::error::Error for BootstrapError {}
```
vorporeal, line 79: can we derive `thiserror::Error`?  pretty sure that covers both of these impls for us (and is more idiomatic)

app/src/ai/agent_sdk/driver.rs
```diff
@@ -436,8 +436,11 @@ pub enum AgentDriverError {
     NotLoggedIn,
     #[error("Saved prompt not found for id {0}")]
     AIWorkflowNotFound(String),
-    #[error("Terminal bootstrap failed")]
-    BootstrapFailed,
+    #[error("Terminal bootstrap failed: {error}")]
```
vorporeal: we shouldn't need to add `: {error}` in here - if we format with `{error:#}`, we'll get the full error chain, using `Display` formatting.

When we interpolate the error like this _and_ add `#[source]`, we end up duplicating the error in the formatted output anywhere we format the outer error using `{error:#}`.

vorporeal (approved):

app/src/ai/agent_sdk/driver/terminal.rs
seemeroland: I implemented display manually so we can branch based on `reason` being `None` or not, which I don't believe deriving supports more cleanly than this impl, but correct me if I'm wrong

PR #12671, opened 2026-06-15
Horizontal tabs pinning UI

johnturcoo:
## Description

Adds pinned state rendering to horizontal tabs. This renders a pin icon to the right of the tab name, or to the right of all tabs in a tab group.

This also adds a safety net feature flag check to the `is_effectively_pinned()` function so that behavior remains unchanged when the feature flag is off (and a tab is pinned somehow). I ack that this should be in a separate PR but have a few changes I need to stack on top of this and want to prioritize velocity. This PR is also quite simple.

## How it works

For tabs: modifies `render_close_tab_button()` to render the pin icon in place of the close button when the tab is pinned and not hovered.

For groups: adds a helper `render_horizontal_group_pin_indicator()` for rendering the pin icon. This is placed after the group title when the group is collapsed, otherwise after the last tab displayed in the group.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4698/horizontal-tabs-pinning-ui

## Testing

### Screenshots / Videos

[Demo video ](https://www.loom.com/share/8c86700ad55d46b795ce2f126910f844)

### DIFF
--- a/app/src/tab.rs
+++ b/app/src/tab.rs
@@ -72,6 +72,7 @@ const WARP_2_HOVERED_TAB_COLOR_OPACITY: Opacity = 50;
 const TAB_CLOSE_BUTTON_OPACITY: Opacity = 60;
 const TAB_CLOSE_BUTTON_WIDTH: f32 = 20.0;
 const MAX_TOOLTIP_LENGTH: usize = 80;
+pub(crate) const TAB_PIN_INDICATOR_ICON_SIZE: f32 = 16.0;
 
 const TAB_INDICATOR_SYNCED_COLOR: u32 = 0x4A93FFFF;
 
@@ -1301,6 +1302,26 @@ impl<'a> TabComponent<'a> {
                     ctx.dispatch_typed_action(WorkspaceAction::CloseTab(tab_index))
                 })
                 .finish()
+        } else if self.show_pin_indicator() {
+            // Pinned: render the pin in the exact slot the close button uses so
+            // hovering swaps icons in place without changing the layout.
+            let theme = self.appearance.theme();
+            ConstrainedBox::new(
+                Align::new(
+                    ConstrainedBox::new(
+                        Icon::PinFilledDiagonal
+                            .to_warpui_icon(theme.main_text_color(theme.background()))
+                            .finish(),
+                    )
+                    .with_width(TAB_PIN_INDICATOR_ICON_SIZE)
+                    .with_height(TAB_PIN_INDICATOR_ICON_SIZE)
+                    .finish(),
+                )
+                .finish(),
+            )
+            .with_width(ICON_DIMENSIONS)
+            .with_height(ICON_DIMENSIONS)
+            .finish()
         } else {
             ConstrainedBox::new(Empty::new().finish())
                 .with_width(ICON_DIMENSIONS)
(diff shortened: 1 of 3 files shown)


app/src/workspace/view.rs
```diff
@@ -27016,19 +27187,60 @@ impl Workspace {
...
+                // The bounds of the group that a tab could be dragged into.
+                let target_group_range =
+                    expanded_target.and_then(|gid| group_member_index_range(&self.tabs, gid));
+
+                // Assign the tab to the group we're dragging over (or clear its
+                // group if none). This intentionally leaves the pin untouched:
+                // a pinned tab dragged into a pinned group keeps its pin for the
+                // duration of the drag and only loses it on drop (see the
+                // `DropTab` handler).
                 self.assign_tab_to_group(current_index, expanded_target, ctx);
```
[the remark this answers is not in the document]
johnturcoo: outdated comment based on bad merge

app/src/workspace/view/tab_grouping.rs
```diff
@@ -482,6 +482,11 @@ impl Workspace {
     /// either because its own `pinned` flag is set (ungrouped pinned tab) or
     /// because it belongs to a pinned group.
     pub(super) fn is_tab_effectively_pinned(&self, tab: &TabData) -> bool {
+        // Safety net, ensures no behavioral changes if feature flag
+        // is off and some tabs have a pinned state saved.
+        if !FeatureFlag::PinnedTabs.is_enabled() {
+            return false;
+        }
```
vkodithala, line 489: nice!

app/src/workspace/view.rs
```diff
@@ -19192,6 +19192,12 @@ impl Workspace {
             }
         }
 
+        // Pinned + expanded: trailing pin indicator after the last member.
```
vkodithala, line 19195: added a note to your Loom, but i feel like this could get confusing. when you expanded the pinned tab group in your demo, i basically thought the last _tab_ was pinned, not the group. non-blocking and probs OOS if this aligns with mocks

app/src/tab.rs
```diff
@@ -1301,6 +1302,26 @@ impl<'a> TabComponent<'a> {
                     ctx.dispatch_typed_action(WorkspaceAction::CloseTab(tab_index))
                 })
                 .finish()
+        } else if self.show_pin_indicator() {
+            // Pinned: render the pin in the exact slot the close button uses so
```
vkodithala, line 1309: nit: this function reads a little weird, `render_close_tab_button` is probably somewhat confusing. i'd either add a comment to the top of this fn describing that we conditionally render a pinned tab indicator, or better, just rename it into something like `render_tab_right_button` (that's crude but you get the idea)

app/src/tab.rs
```diff
@@ -1629,6 +1657,8 @@ impl<'a> TabComponent<'a> {
                 )
             };
 
+        // `render_close_tab_button` renders the close button or a pin icon
```
vkodithala: if we rename the fn to something that more accurately communicates its intent we probably don't need this as a comment

vkodithala (approved):
LG, small question around UI and a few nits on fn naming. nothing too important, 🚢

app/src/workspace/view.rs
johnturcoo: Ya this is to-spec. I agree its a bit confusing though

PR #12614, opened 2026-06-13
remove dead code for WelcomePalette

acarl005:
## Description

This is dead code now. Just deleting it. Replaced by the new agent view. This was the palette that we tried out as the new "zero state" when a new tab was opened. It showed indexed codebases and recent conversations.

## Linked Issue

None, this is dead code removal.

PR #12672, opened 2026-06-15
add back pane_leaves to sqlite save_app_state which was mistakenly deleted

acarl005:
## Description

I tried to delete some dead code in #12614 but accidentally deleted something we need. Adding it back.


johnturcoo (approved):
Thanks!

PR #12702, opened 2026-06-16
Rely on server sending protocol-level ping frames, ensure one connection

seemeroland:
## Description

Was debugging https://linear.app/warpdotdev/issue/REMOTE-1883/investigate-enterprise-report-of-cant-send-follow-up-to-oz

The server was not receiving pings from the client (supposed to be sent every 5s), and the server terminates the connection after not receiving a ping from the client for 30s. The reconnects succeed immediately after the server terminates the connection.

This is because the browser throttles background timers to be at most every 60s. The wasm client was both sending pings every ~60s and checking idle timeout every ~60s. This would cause the server to terminate the connection every 60-90s which is what I saw. 

The wasm client would also end up in a state where its own idle timeout fired, showing "Offline, trying to reconnect..." but it actually was permanently disconnected:
```
INFO app/src/terminal/shared_session/viewer/network.rs:295 Viewer reconnecting: heartbeat idle timeout
warp.js:1111 INFO app/src/terminal/shared_session/viewer/network.rs:469 Attempting to reconnect to session sharing server as viewer
warp.js:1111 INFO app/src/terminal/shared_session/viewer/network.rs:487 Successfully reconnected to server as viewer
warp.js:1111 INFO app/src/terminal/shared_session/viewer/network.rs:385 Closing websocket to session sharing server as viewer
warp.js:1111 INFO app/src/terminal/shared_session/viewer/network.rs:358 Websocket to session sharing server ended
warp.js:1111 INFO app/src/terminal/shared_session/viewer/network.rs:385 Closing websocket to session sharing server as viewer
warp.js:1111 INFO app/src/terminal/shared_session/viewer/network.rs:358 Websocket to session sharing server ended
warp.js:1111 INFO app/src/terminal/shared_session/viewer/network.rs:295 Viewer reconnecting: heartbeat idle timeout
```

This was due to a race where the idle timeout could cause the client to reconnect before the original stream closed. Once the original stream closed, we called `network.close()` which closed the reconnected websocket. We prevent this and ensure old websockets are closed before reconnecting.

Remove client sending application-level pings entirely, and replace with protocol-level ping frames sent by the server every 15s, with 45s timeout. Tungstenite (library we use) automatically responds with protocol-level pong. The browser also autoresponds and this doesn't use a timer that gets throttled. The server will be responsible for terminating idle connections, and the client will not do anything but respond with pongs.

## Testing

Before this change, leaving a wasm viewer for > 1 min would result in both server and client terminating connection, and only a refresh would fix.

Tested locally, killed session sharing server and restarted, sharer and viewer still reconnecting fine. Both wasm and desktop viewer can be left indefinitely and remain connected and able to send follow ups

### DIFF
--- a/app/src/terminal/shared_session/network/heartbeat.rs
+++ b/app/src/terminal/shared_session/network/heartbeat.rs
@@ -46,8 +46,13 @@ impl Heartbeat {
     /// Starts the periodic ping and the idle timeout tracker.
     pub fn start(&mut self, ctx: &mut ModelContext<Self>) {
         self.reset_idle_timeout(ctx);
+        // Abort any existing ping timer so we don't accumulate chains across reconnects.
+        if let Some(handle) = self.periodic_ping_abort_handle.take() {
+            handle.abort();
+        }
         self.periodic_ping(ctx);
     }
+
     pub fn stop(&mut self) {
         if let Some(handle) = self.idle_timeout_abort_handle.take() {
             handle.abort();
--- a/app/src/terminal/shared_session/network/heartbeat.rs
+++ b/app/src/terminal/shared_session/network/heartbeat.rs
@@ -1,3 +1,8 @@
+// Session sharing now relies on the server sending protocol-level ping frames that the client responds to,
+// and the server initiates disconnect if the client is unresponsive.
+// This module is no longer used, but we keep the code around for now.
+#![allow(dead_code)]
+
 use std::time::Duration;
 
 use futures::stream::AbortHandle;
--- a/app/src/terminal/shared_session/sharer/network.rs
+++ b/app/src/terminal/shared_session/sharer/network.rs
@@ -1,7 +1,10 @@
 //! The "sender" of a shared session represents the sharer's end.
 //!
 //! Currently there is no way to share a session from wasm.
-#![cfg_attr(any(test, feature = "integration_tests", target_family = "wasm"), allow(dead_code))]
+#![cfg_attr(
+    any(test, feature = "integration_tests", target_family = "wasm"),
+    allow(dead_code)
+)]
 
 use std::collections::HashMap;
 use std::pin::pin;
--- a/app/src/terminal/shared_session/sharer/network.rs
+++ b/app/src/terminal/shared_session/sharer/network.rs
@@ -37,7 +37,7 @@ use session_sharing_protocol::sharer::{
 };
 use warp_core::features::FeatureFlag;
 use warpui::r#async::Timer;
-use warpui::{Entity, ModelContext, ModelHandle, RequestState, RetryOption, SingletonEntity};
+use warpui::{Entity, ModelContext, RequestState, RetryOption, SingletonEntity};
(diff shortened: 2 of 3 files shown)


app/src/terminal/shared_session/viewer/network.rs
```diff
@@ -356,13 +323,13 @@ impl Network {
...
             |network, ctx| {
                 log::info!("Websocket to session sharing server ended");
-                // Close our current websocket proxy, because we may try to reconnect and that will create a new websocket proxy.
-                // This must be done before trying to reconnect.
-                network.close();
                 if matches!(network.stage, Stage::JoinedSuccessfully) {
                     // The connection may have timed out or the server restarted.
                     log::info!("Viewer reconnecting: websocket closed by server");
                     network.reconnect_websocket(ctx);
+                } else if !matches!(network.stage, Stage::Reconnecting { .. }) {
```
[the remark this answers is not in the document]
seemeroland: I don't think this is a real issue. We only reach this function `on_websocket_connected` if the reconnection succeeded, and that puts us in `JoinedSuccessfully`. We'd only be in `Reconnecting` if the websocket hasn't connected yet, which means this function hasn't been called.

This is a guard for an old stream to prevent closing an active reconnect that's going on (should be unexpected now that `reconnect_websocket` closes old stream before reconnecting, but for safety

app/src/terminal/shared_session/network/heartbeat.rs
```diff
@@ -1,3 +1,8 @@
+// Session sharing now relies on the server sending protocol-level ping frames that the client responds to,
+// and the server initiates disconnect if the client is unresponsive.
+// This module is no longer used, but we keep the code around for now.
```
vorporeal, line 3: por que

why not just delete this?

app/src/terminal/shared_session/sharer/network.rs
```diff
@@ -1,7 +1,10 @@
 //! The "sender" of a shared session represents the sharer's end.
 //!
 //! Currently there is no way to share a session from wasm.
-#![cfg_attr(target_family = "wasm", allow(dead_code))]
+#![cfg_attr(
+    any(test, feature = "integration_tests", target_family = "wasm"),
+    allow(dead_code)
```
vorporeal, line 6: should this be `expect(dead_code)` (if we don't anticipate any of this being used)?  or just conditionally compile the whole module?

or are only parts of this dead code?  if so, can we conditionally compile them instead?

vorporeal (approved):

app/src/terminal/shared_session/network/heartbeat.rs
seemeroland: Prefer to keep it around for now until we're confident we won't want to add it back. This PR also fixes some things about it I'd like to preserve in case we want it

app/src/terminal/shared_session/sharer/network.rs
seemeroland: Only parts of it are dead code in tests, but adding it on everything is pretty annoying/tedious. It's the same pattern for wasm that already existed

PR #12703, opened 2026-06-16
Fix headless integration-test render loop

cephalonaut:
## Description
Fixes a headless integration-test render loop where `integration_tests` builds eagerly called `build_scene` from the window invalidation callback.

Integration tests run with a real winit/X11 event loop, so `request_redraw()` is sufficient. Calling `build_scene()` eagerly under Xvfb can create a loop of invalidation → scene build → synthetic mouse move / view notifications → invalidation, keeping the main thread busy and preventing async `RunTask` events from being processed. In agent-mode evals this starves completed tool-action callbacks and leaves conversations stuck `InProgress`.

The change keeps eager `build_scene()` only for unit tests, where there is no real event loop.

## Linked Issue

## Testing
- `cargo fmt -p warpui_core`
- `cargo check -p warpui_core --features integration_tests`

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/crates/warpui_core/Cargo.toml
+++ b/crates/warpui_core/Cargo.toml
@@ -10,6 +10,10 @@ license = "MIT"
 traces = []
 # Feature enabled only when app is compiled for integration tests.
 integration_tests = ["dep:minimp4", "dep:openh264"]
+# Feature enabled only when the app is compiled for agent-mode evals (which also
+# enables integration_tests). Used to opt agent-mode evals out of eager scene
+# building in the window-invalidation callback; see core/app.rs.
+agent_mode_evals = []
 # If enabled, writes named telemetry events to the info log.  Should never be
 # enabled in release builds without a PII audit of telemetry data.
 log_named_telemetry_events = []
--- a/crates/warpui_core/Cargo.toml
+++ b/crates/warpui_core/Cargo.toml
@@ -10,10 +10,9 @@ license = "MIT"
 traces = []
 # Feature enabled only when app is compiled for integration tests.
 integration_tests = ["dep:minimp4", "dep:openh264"]
-# Feature enabled only when the app is compiled for agent-mode evals (which also
-# enables integration_tests). Used to opt agent-mode evals out of eager scene
-# building in the window-invalidation callback; see core/app.rs.
-agent_mode_evals = []
+# When enabled, skip eager scene building from the window-invalidation callback
+# and rely on the normal platform redraw path instead; see core/app.rs.
+defer_scene_build = []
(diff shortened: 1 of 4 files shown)


cephalonaut:
This seems to fix the headless eval hang I was telling you about. Agent mode diagnosed as a tight CPU loop caused by this code and empirically it seems to be true.

crates/warpui_core/src/core/app.rs
```diff
@@ -2502,9 +2502,21 @@ impl AppContext {
                         // of the actual work later.
                         window.request_redraw();
 
-                        // In tests, however, there's no real event loop, so we need to do the work now.
-                        // While this _shouldn't_ be necessary in integration tests, it currently is.
```
danielpeng2, line 2506: do you happen to know why this was required before but isn't now?

danielpeng2 (commented):
do non-headless integration test runs still work after this?

crates/warpui_core/src/core/app.rs
cephalonaut: Basically, no, some of the integration tests did fail with this change. Some of the hover/focus related integration tests were depending on this code to generate a synthetic mouse move to update UI state. It was added in #17255 (viewported List) and deliberately kept in #19270, which added request_redraw but left the eager build, noting it "shouldn't be necessary in integration tests, but currently is."

So I've change the fix to only exclude evals which really do run in a real UI environment and shouldn't need this fix.

crates/warpui_core/Cargo.toml
```diff
@@ -10,6 +10,10 @@ license = "MIT"
 traces = []
 # Feature enabled only when app is compiled for integration tests.
 integration_tests = ["dep:minimp4", "dep:openh264"]
+# Feature enabled only when the app is compiled for agent-mode evals (which also
+# enables integration_tests). Used to opt agent-mode evals out of eager scene
+# building in the window-invalidation callback; see core/app.rs.
+agent_mode_evals = []
```
danielpeng2: Hmmm this is leaking knowledge of our evals framework to the UI framework, which is supposed to be app-agnostic. I'm not sure I see a way around this though. Maybe we could rename the feature to be something more generic?

cephalonaut: Oh, good point, lemme think

cephalonaut: Ok, I renamed it to defer_scene_build and tried to make the comment reasonably generic.

danielpeng2 (approved):

PR #12704, opened 2026-06-16
Fix fork button tooltip, command insertion, /continue-locally visibility

harryalbert:
## Description

When in a cloud agent conversation, the fork button in the ai block footer inserts the `/fork` command into the input even though `/fork` isn't available in cloud agent conversations. This PR changes it so that, in cloud conversations, this button now inserts `/continue-locally` instead (with a tooltip for continuing locally instead of forking).

Relatedly, while working on this PR I noticed that we don't provide the `/fork` option for cloud conversations, but we _do_ provide the `/continue-locally` button in all contexts, including local conversations. I changed it so that we get rid of the `/continue-locally` command in local conversations, as this option doesn't really make sense in that context.

Finally, I changed it so that we stop hiding the `/fork` option in cloud mode conversations. I think this command is completely valid in cloud mode conversations, even if it's duplicative of /continue-locally (which we just added for extra visibility's sake). We still filter the `/fork` option out when we're in the cloud mode v2 composer, but once we're in a regular cloud conversation I think it makes sense for this slash command to be available.

## Testing

## Demo

https://www.loom.com/share/5ab05b909ca14257b73aead5d5167b4f

## Agent Mode

### DIFF
--- a/app/src/search/slash_command_menu/static_commands/mod.rs
+++ b/app/src/search/slash_command_menu/static_commands/mod.rs
@@ -34,12 +34,15 @@ bitflags! {
         const CODEBASE_CONTEXT = 1 << 6;
         /// Requires AI to be globally enabled.
         const AI_ENABLED = 1 << 7;
+        /// Requires a non-cloud-agent context.
         const NOT_CLOUD_AGENT = 1 << 8;
+        /// Requires a cloud-agent context.
+        const CLOUD_AGENT = 1 << 9;
         /// Set on the session context iff the slash command data source was constructed via
         /// `SlashCommandDataSource::for_cloud_mode_v2` *and* `FeatureFlag::CloudModeInputV2`
         /// is enabled. Commands that require this bit are hidden everywhere except the V2
         /// cloud-mode composing input.
-        const CLOUD_AGENT_V2 = 1 << 9;
+        const CLOUD_MODE_V2_COMPOSER = 1 << 10;
     }
 }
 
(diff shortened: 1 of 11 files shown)


harryalbert:
* **#12704** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12704?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12704?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/search/slash_command_menu/static_commands/mod.rs
```diff
@@ -34,12 +34,15 @@ bitflags! {
         const CODEBASE_CONTEXT = 1 << 6;
         /// Requires AI to be globally enabled.
         const AI_ENABLED = 1 << 7;
+        /// Requires a non-cloud-agent context.
         const NOT_CLOUD_AGENT = 1 << 8;
+        /// Requires a cloud-agent context.
+        const CLOUD_AGENT = 1 << 9;
```
abhishek-pandya, line 40: Any chance we could use just one of `CLOUD_AGENT` and `NOT_CLOUD_AGENT`? It seems the truth value of one implies that the other is the opposite and I'm worried these could get out of sync.

abhishek-pandya (approved):
Nice quick win for our slash command and forking UX!

app/src/search/slash_command_menu/static_commands/mod.rs
harryalbert: unfortunately given the bit mapping method here, it's not so simple to check like !CLOUD_AGENT. I think it'd be worth a follow up to generalize and make this better, but right now it's not super easy

PR #12706, opened 2026-06-16
Fix tab group bugs for "special tabs"

johnturcoo:
## Description

When the active tab was inside a tab group, opening a "special" tab — Settings, a code file, a notebook, orchestrated agent new tab — dropped the new tab after the group instead of joining it or split the group. (settings/notebooks opened after the group by design, code files and orchestrated agent tabs were a bug that split the group) 

Demo of bugs below for more context.

Code files and orchestrated agent tabs  are built from a pane snapshot or an already-constructed pane, so unlike a regular new tab they never inherited the active tab's group, and inserting them inline would split it. 

This PR  makes them inherit the active group and land inside it, matching normal new-tab behavior — while preserving group contiguity and pinned-region ordering. Gated behind the existing grouped-tabs feature.

## How it works 

- Centralized new-tab placement and group inheritance in a single helper used by both tab-creation paths (`new_tab_index_and_group`).
    - This handles the tab placement setting (after current tab or after all tabs), placement within a group and clamps to after the pinned region if needed
- modified all call sites of `add_tab_from_existing_pane` to use this new helper to determine group ownership and insertion index
- modified `add_tab_from_existing_pane` to now assign the desired group to the newly created tabs
- simplified `add_tab_with_pane_layout` significantly by using the new helper function and removing the special case that used to push settings and notebooks past the end of a groups run

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4747/opening-a-restored-tab-joins-the-current-group

## Testing

### Screenshots / Videos

[Demo of bugs](https://www.loom.com/share/196d0645ea0e4cefa80698603950b57d)

[Demo of fix](https://www.loom.com/share/196d0645ea0e4cefa80698603950b57d)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -4110,7 +4110,7 @@ impl Workspace {
         } else {
             let home_pane = super::home::create_home_pane(ctx);
             placeholder_pane = Some(home_pane.as_pane().id());
-            self.add_tab_from_existing_pane(home_pane, 0, ctx);
+            self.add_tab_from_existing_pane(home_pane, 0, None, ctx);
 
             // If we can't start a terminal session to run the onboarding flow, show the Warp Home
             // placeholder along with Warp Drive.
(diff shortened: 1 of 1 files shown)


app/src/workspace/view.rs
```diff
@@ -16023,7 +16003,13 @@ impl Workspace {
                             };
 
                             if let Some(pane) = pane {
-                                self.add_tab_from_existing_pane(pane, workspace_tab_index, ctx);
+                                // TODO(johnturcoo) inherit the tab group on pane drop.
```
peicodes, line 16006: Is this still a TODO?

peicodes (approved):

app/src/workspace/view.rs
johnturcoo: Yes

PR #12712, opened 2026-06-16
4xx MAA errors should be FAILED (user error) status in oz

seemeroland:
## Description

4xx MAA errors should be FAILED and marked `is_user_error=true`. The nonretryable errors are already all 4xx errors except `REQUEST_TIMEOUT` and `TOO_MANY_REQUESTS`, and it's correct for these to be `is_user_error=false`


app/src/ai/blocklist/local_agent_task_sync_model_tests.rs
```diff
@@ -122,13 +122,29 @@ fn other_error_is_error_with_internal() {
...
+            is_user_error: false,
         }),
         AgentTaskState::Error,
         Some(PlatformErrorCode::InternalError),
         Some("something broke"),
     );
 }
 
+#[test]
+fn other_user_error_is_failed_with_feature_not_available() {
```
advait-m: nit: should this be named `feature_not_available` or `invalid_request`? since we aren't testing `PlatformErrorCode::FeatureNotAvailable` here

advait-m (approved):
:+1:

app/src/ai/blocklist/local_agent_task_sync_model_tests.rs
seemeroland: Good call forgot to update the name

PR #12735, opened 2026-06-17
fix: sync DockTilePlugin icon to NSApp for Stage Manager

### LINKED ISSUE #12730: Stage Manager shows the default Warp icon instead of the selected custom app icon
## Pre-submit Checks

- [x] I have searched Warp bugs and did not find a duplicate for Stage Manager showing the wrong Warp icon.
- [x] I have searched Warp known issues and did not find this listed.

## Describe the bug

When using Stage Manager on macOS, Warp's icon in Stage Manager can show the default Warp icon instead of the custom app icon selected in Warp settings. The Dock shows the selected custom icon correctly, but Stage Manager does not match it.

From a Warp team discussion, this may be because Stage Manager does not use the dock tile extension path that Warp currently relies on to override the icon.

## To reproduce

1. On macOS, open Warp.
2. Select a non-default Warp app icon in settings.
3. Enable Stage Manager.
4. Compare the Warp icon shown in the Dock with the Warp icon shown in Stage Manager.

## Expected behavior

Stage Manager should show the same selected custom Warp icon as the Dock.

## Screenshots, videos, and logs

Screenshots were observed in the original report, but I do not have files to attach here. The visible symptom is that Stage Manager shows the default Warp icon while the Dock shows the custom selected icon.

## Operating system (OS)

macOS

## Operating system and version

macOS with Stage Manager enabled. Exact version not provided.

## Shell Version

Not applicable.

## Current Warp version

Not provided.

## Regression

Unknown. This may have existed throughout custom app icon support with Stage Manager.

## Additional context

Maintainer context from Slack: Stage Manager appears not to work with the dock tile extension Warp uses to override the app icon. This issue is being opened so contributors can investigate whether Warp should also update the app-level icon path used by Stage Manager.

No.

Likely yes. The mismatch is specific to Warp's custom app icon override behavior.

rubendepaz1:
## Description

Stage Manager reads the app icon from `NSApplication.applicationIconImage`, not from `NSDockTile`. The docktile plugin only updated the dock tile view via `setContentView:`, so Stage Manager always showed the bundled default icon regardless of the user's custom selection.

This fix adds `setApplicationIconImage:` calls alongside the existing `setContentView:` calls so both the Dock and Stage Manager display the correct icon.

## Linked Issue

Closes #12730

## Testing

### Screenshots / Videos

**Before:** Dock shows the custom icon, Stage Manager shows the default icon.
**After:** Both Dock and Stage Manager show the custom icon.

CHANGELOG-BUG-FIX: Stage Manager now shows the correct custom app icon instead of always showing the default icon.

### DIFF
--- a/app/DockTilePlugin/WarpDockTilePlugin.m
+++ b/app/DockTilePlugin/WarpDockTilePlugin.m
@@ -68,6 +68,7 @@
             [self logMessage:@"User has default icon, resetting dock tile to system default"];
             [tile setContentView:nil];
             [tile display];
+            [NSApp setApplicationIconImage:nil];
             return;
         }
 
@@ -84,7 +85,10 @@
         [imageView setImageScaling:NSImageScaleProportionallyUpOrDown];
         [tile setContentView:imageView];
         [tile display];
-        
+        if (currentImage) {
+            [NSApp setApplicationIconImage:currentImage];
+        }
+
         [self logMessage:[NSString stringWithFormat:@"Dock tile updated with icon: %@", iconFileName]];
     } @catch (NSException *exception) {
         [self logMessage:[NSString stringWithFormat:@"Exception updating dock tile icon: %@\nStack trace: %@\nTile: %@", 


rubendepaz1:
@cla-bot check

rubendepaz1:
I investigated this further and the current approach does not fix Stage Manager.

Findings:
- Warp’s custom icon is applied correctly to the Dock.
- macOS stores that override on the app bundle via `Icon\r` / resource fork, which Finder/Dock respect.
- Stage Manager appears to ignore that override and continues using the bundled icon from `Contents/Resources/AppIcon.icns`.
- I tried `NSApplication.applicationIconImage`, `NSWorkspace.setIcon(_:forFile:)`, the dock tile plugin path, and `NSWindow.miniwindowImage`; none made Stage Manager use the selected custom icon in the real Stage Manager shelf.
- Replacing the bundled `.icns` is the only path I found that would affect the source Stage Manager seems to read, but it mutates the signed app bundle and breaks code signing, so it does not look viable as a normal runtime fix.

Given that, this PR should not be treated as a working fix yet. This may be a macOS limitation unless there is another Apple-supported API for updating the icon source Stage Manager uses.

PR #12738, opened 2026-06-17
Fix LRC auto-resume after Ctrl-C takeover

vkodithala:
## Description

This fixes a long-running-command Agent Mode regression where a conversation could remain stuck showing `Warping...` after a Ctrl-C takeover.

Repro flow:
1. Ask the agent to start a long-running command, e.g. `sleep 5`.
2. Press Ctrl-C while the CLI subagent is monitoring the command. This switches control back to the user.
3. Let the command complete normally.

Before this PR, the first Ctrl-C takeover used `UserTakeOverReason::Stop`, and `switch_control_to_user` intentionally cancelled only the CLI subagent monitoring stream with `CancellationReason::CLISubagentUserTakeover`. That cancellation preserves the primary conversation's `InProgress` state because the primary agent is expected to resume once the command completes.

The bug was that `on_user_block_completed` decided whether to resume by checking that the takeover reason was **not** `Stop`. That meant the Ctrl-C takeover path left the primary conversation `InProgress`, but skipped the completion-time resume, resulting in an infinite `Warping...` state. The `cmd+i` manual takeover path did not hit the bug because it uses `UserTakeOverReason::Manual`, which passed the old `!reason.is_stop()` gate.

This PR decouples auto-resume suppression from the takeover reason:

- Adds an explicit `suppress_auto_resume` bit to `AgentInteractionMetadata`, including serialized block metadata.
- Sets that bit only in the genuine teardown paths (`stop_local_agent_conversation` and `rewind_ai_conversation`) through `set_user_control_and_suppress_auto_resume`.
- Updates `on_user_block_completed` to resume user-controlled requested-command LRCs unless `suppress_auto_resume` is set.

### Takeover reason semantics preserved

This intentionally keeps `UserTakeOverReason::Stop` rather than collapsing it into `Manual`:

- `Manual` still represents an explicit manual/user-control takeover (e.g. `cmd+i`). It does not suppress auto-resume, so the primary conversation resumes when the command completes.
- `Stop` still represents Ctrl-C takeover of an agent-controlled long-running command. It remains the sentinel used by the #12623 flow: after the first Ctrl-C takes control, a second Ctrl-C cancels the whole conversation and interrupts the command.
- `Stop` no longer inherently means "do not resume when the command completes." A plain Ctrl-C takeover now has `Stop` with `suppress_auto_resume = false`, so it resumes on natural command completion.
- Teardown flows still suppress resume explicitly. When the user stops or rewinds the conversation, the block is marked with `Stop` and `suppress_auto_resume = true`, so a later command completion cannot resurrect a cancelled or rewound conversation. @MaggieShan lemme know if this is confusing, happy to chat sync. The issue here was that previously, `UserTakeOverReason::Stop` conflated two intents: suppressing auto-resume after command completion (which should only apply in the case that a conversation is rewinded/explicitly killed to avoid a stray "Resume conversation"), and cancelling the active subagent conversation (the normal Ctrl-C case).
- `TransferFromAgent` behavior is unchanged: if the agent explicitly transfers control to the user, Ctrl-C interrupts the command without cancelling the conversation.

## Testing

Tested manually; demo/explainer [here](https://www.loom.com/share/085080576cd4457996058674788f6e64).

- [x] `cargo nextest run -p warp -E 'test(completed_user_controlled_lrc) | test(ctrl_c_after_stop_takeover_cancels_conversation) | test(ctrl_c_after_transfer_takeover_does_not_cancel_conversation)'`
  - `completed_user_controlled_lrc_resumes_when_not_suppressed`
  - `completed_user_controlled_lrc_skips_resume_when_suppressed`
  - `ctrl_c_after_stop_takeover_cancels_conversation`
  - `ctrl_c_after_transfer_takeover_does_not_cancel_conversation`
- [x] `./script/format`

## Agent Mode

## Agent Context

- Conversation: https://staging.warp.dev/conversation/1a1573ce-4822-4c1b-8992-c99840689000
- Plan: https://staging.warp.dev/drive/notebook/hzKaQ4fSha6yU1LXyDrQFN

CHANGELOG-BUG-FIX: Fixed a case where taking over an agent-monitored long-running command with Ctrl-C could leave the conversation stuck showing `Warping...` after the command completed.

### DIFF
--- a/app/src/terminal/model/block/interaction_mode.rs
+++ b/app/src/terminal/model/block/interaction_mode.rs
@@ -156,15 +156,16 @@ impl Block {
         }
     }
 
-    /// Hands control to the user with the `Stop` reason and suppresses auto-resume of the
-    /// conversation once the command completes. Used by teardown paths (rewind, stop) where
-    /// the conversation has been cancelled and should not resume on its own.
-    pub fn set_user_control_and_suppress_auto_resume(&mut self) {
+    /// Hands control to the user with a non-resuming `Stop`. Used by teardown paths (rewind,
+    /// stop) where the conversation has been cancelled and must not resume when the command
+    /// completes.
+    pub fn set_user_control_for_teardown(&mut self) {
         if let InteractionMode::Agent(metadata) = &mut self.interaction_mode {
-            metadata.suppress_auto_resume = true;
             if let Some(state) = &mut metadata.long_running_control_state {
                 *state = LongRunningCommandControlState::User {
-                    reason: UserTakeOverReason::Stop,
+                    reason: UserTakeOverReason::Stop {
+                        should_auto_resume: false,
+                    },
                 };
             }
         }
@@ -231,7 +232,6 @@ impl Block {
             long_running_control_state: None,
             has_agent_written_to_block: false,
             should_hide_block: true,
-            suppress_auto_resume: false,
         })
     }
 
@@ -363,7 +363,6 @@ impl InteractionMode {
             }),
             has_agent_written_to_block: false,
             should_hide_block: false,
-            suppress_auto_resume: false,
         }))
     }
 
@@ -488,10 +487,6 @@ pub struct AgentInteractionMetadata {
     /// `true` if this block should be hidden from the user (as is the case with AI-requested
     /// commands, for example).
     should_hide_block: bool,
-
-    /// `true` if the conversation must not auto-resume when this command completes, even though
-    /// the user is in control. Set by teardown paths (rewind, stop) that cancel the conversation.
-    suppress_auto_resume: bool,
 }
 
 impl AgentInteractionMetadata {
@@ -511,7 +506,6 @@ impl AgentInteractionMetadata {
             long_running_control_state,
             has_agent_written_to_block,
             should_hide_block,
-            suppress_auto_resume: false,
         }
     }
 
@@ -558,14 +552,6 @@ impl AgentInteractionMetadata {
 
     pub fn should_hide_block(&self) -> bool {
         self.should_hide_block
-    }
-
-    pub fn should_suppress_auto_resume(&self) -> bool {
-        self.suppress_auto_resume
-    }
-
-    pub fn set_suppress_auto_resume(&mut self) {
-        self.suppress_auto_resume = true;
     }
 }
 
--- a/app/src/terminal/model/block/interaction_mode.rs
+++ b/app/src/terminal/model/block/interaction_mode.rs
@@ -156,16 +156,17 @@ impl Block {
         }
     }
 
-    /// Sets control to user with Stop reason if a long-running control state exists.
-    pub fn set_user_control_with_stop_reason(&mut self) {
-        if let InteractionMode::Agent(AgentInteractionMetadata {
-            long_running_control_state: Some(ref mut state),
-            ..
-        }) = self.interaction_mode
-        {
-            *state = LongRunningCommandControlState::User {
-                reason: UserTakeOverReason::Stop,
-            };
+    /// Hands control to the user with the `Stop` reason and suppresses auto-resume of the
+    /// conversation once the command completes. Used by teardown paths (rewind, stop) where
+    /// the conversation has been cancelled and should not resume on its own.
+    pub fn set_user_control_and_suppress_auto_resume(&mut self) {
+        if let InteractionMode::Agent(metadata) = &mut self.interaction_mode {
+            metadata.suppress_auto_resume = true;
+            if let Some(state) = &mut metadata.long_running_control_state {
+                *state = LongRunningCommandControlState::User {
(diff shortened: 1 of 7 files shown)


vkodithala:
* **#12740** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12740?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#12738** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/terminal/model/block/serialized_block.rs
```diff
@@ -124,20 +127,25 @@ impl From<AgentInteractionMetadata> for SerializedAIMetadata {
...
+        let mut metadata = AgentInteractionMetadata::new(
             value.requested_command_action_id,
             value.conversation_id,
             value.subagent_task_id,
             value.long_running_control_state,
             value.has_agent_written_to_block,
             value.should_hide_block,
-        )
+        );
+        if value.suppress_auto_resume {
```
MaggieShan: can we pass the value as it instead of doing an extra check here?

app/src/terminal/view.rs
```diff
@@ -10868,13 +10868,10 @@ impl TerminalView {
...
                     if ai_metadata.requested_command_action_id().is_some()
+                        && !ai_metadata.should_suppress_auto_resume()
                         && ai_metadata
                             .long_running_control_state()
-                            .is_some_and(|state| {
-                                state
-                                    .user_take_over_reason()
-                                    .is_some_and(|reason| !reason.is_stop())
-                            }) =>
+                            .is_some_and(|state| state.user_take_over_reason().is_some()) =>
```
MaggieShan: We should be able to use `is_user_in_control` instead

app/src/terminal/model/block/interaction_mode.rs
```diff
@@ -156,16 +156,17 @@ impl Block {
...
-            ..
-        }) = self.interaction_mode
-        {
-            *state = LongRunningCommandControlState::User {
-                reason: UserTakeOverReason::Stop,
-            };
+    /// Hands control to the user with the `Stop` reason and suppresses auto-resume of the
+    /// conversation once the command completes. Used by teardown paths (rewind, stop) where
+    /// the conversation has been cancelled and should not resume on its own.
+    pub fn set_user_control_and_suppress_auto_resume(&mut self) {
```
MaggieShan: I wonder if there is a better name for this function - it's not entirely clear why callers of this fn should set the `suppress_auto_resume` flag to true, while other callers that change LRC state don't 

IIUC we only want to suppress auto resume on "teardown" flows like stopping a conversation or rewinding it - how can we make that clearer?

app/src/terminal/model/block/interaction_mode.rs
```diff
@@ -485,6 +488,10 @@ pub struct AgentInteractionMetadata {
     /// `true` if this block should be hidden from the user (as is the case with AI-requested
     /// commands, for example).
     should_hide_block: bool,
+
+    /// `true` if the conversation must not auto-resume when this command completes, even though
+    /// the user is in control. Set by teardown paths (rewind, stop) that cancel the conversation.
+    suppress_auto_resume: bool,
```
MaggieShan: Curious if you've considered either:
* moving this flag to `LongRunningCommandControlState` if it only matters for an LRC, or
* introducing a new variant to `UserTakeOverReason` instead

imo this could help make intent clearer and harder to desync later on

MaggieShan (approved):
Approving to unblock but had a few questions!

app/src/terminal/model/block/interaction_mode.rs
vkodithala: Yeah, I agree - fixed in latest, renamed to `set_user_control_for_teardown` to more directly tie this to teardown (conversation stoppage, rewind) events.

app/src/terminal/model/block/interaction_mode.rs
vkodithala: Definitely agree re: making it harder to desync. Keeping this field on interaction mode was def wrong, especially since an active interaction mode doesn't always imply LRC. Made some changes so the `Stop` takeover reason takes a `should_auto_resume` param that's set on teardown events.

app/src/terminal/model/block/serialized_block.rs
vkodithala: Good catch, fixed!

app/src/terminal/view.rs
vkodithala: Agreed and good catch. Fixed in latest.

PR #12747, opened 2026-06-17
Fix crash when rejecting then confirming a requested command

vkodithala:
### The issue
When a requested-command approval card (the card Agent Mode shows when it asks for permission before running a shell command) had keyboard focus, the reject + confirm sequence could crash the app in debug builds.

Repro:
1. In Agent Mode, ask the agent to run a command and have it request approval (e.g. `sleep 5`).
2. With the approval card focused, press `Ctrl-C` to reject it.
3. Immediately press `Enter`.

Two problems combined to cause this:
- After the card was rejected (or accepted), keyboard focus stayed on the now-resolved card instead of returning to the terminal.
- The card's key bindings (`enter` → Accept, `ctrl-c` → Reject, etc.) were always active for a focused `RequestedCommandView`, regardless of whether the underlying action was still pending. So the stray `Enter` re-fired `Accept` against an action that was no longer pending, which tripped `debug_assert!(false, "Expected action to be requested command.")` in `BlocklistAIActionModel` and panicked debug builds.

### The fix
- **Scope the card's key bindings to when it is actually awaiting confirmation.** Added a new `RequestedActionBlocked` keymap context that is only set while the action's status `is_blocked()` (waiting for the user). Accept / reject / edit-mode bindings now require this context, and `handle_action` short-circuits those actions when the card is no longer blocked, so a late keypress is ignored.
- **Yield focus back to the terminal after a card resolves.** Added `yield_requested_action_focus_if_focused`, which emits `AIBlockEvent::FocusTerminal` when the resolved card (or a child) currently holds focus, called after both Accept and Reject. This prevents a stray `Enter` from landing on a stale card.
- **Make the action model defensive.** Downgraded the acceptance `debug_assert!(false, ...)` to a `log::warn!` so that a late acceptance for a non-pending command is logged and ignored rather than crashing.

## Linked Issue

## Testing

Manually tested and things work as expected (no app crash). Happy to share a demo as-needed, but think the code here is pretty self-explanatory.

Manual repro before the fix: reject a focused approval card with `Ctrl-C`, then press `Enter` → debug build panics on the `debug_assert!`. After the fix, focus returns to the terminal on reject, the stale `Enter` is ignored, and no crash occurs.

## Agent Mode

### DIFF
--- a/app/src/ai/blocklist/action_model.rs
+++ b/app/src/ai/blocklist/action_model.rs
@@ -1218,7 +1218,7 @@ impl BlocklistAIActionModel {
         }
 
         let Some(conversation_id) = found_conversation_id else {
-            debug_assert!(false, "Expected action to be requested command.");
+            log::warn!("Ignoring acceptance for non-pending requested command: {action_id:?}");
             return;
         };
 
--- a/app/src/ai/blocklist/block.rs
+++ b/app/src/ai/blocklist/block.rs
@@ -3438,6 +3438,7 @@ impl AIBlock {
                         ctx,
                     );
                 });
+                self.yield_requested_action_focus_if_focused(&view, ctx);
                 ctx.notify();
             }
             RequestedCommandViewEvent::EnableAutoexecuteMode => {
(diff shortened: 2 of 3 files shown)


MaggieShan (approved):
Nice!

PR #12748, opened 2026-06-17
Disable handoff while child agents are running

harryalbert:
## Summary
local to cloud handoff is not amazing for orchestrated conversations, as we don't bring the child or parent agents into the cloud conversation. We make the agent aware of this in the cloud handoff prompt, but it's still a bit borked in some cases.

The case where it's the most borked is when the child agent is still in progress, but the parent agent is moved to the cloud while the child agent is still trying to interact w/ it. The fix here is to disable handoff while child agents are in progress.

The other change bundled in here is to make sure we snapshot changes from all participants in the given orchestrated set of conversations (instead of just the parent) when handing off, so that any child agent work is preserved.

## Demo

https://www.loom.com/share/aa66467f088a4540b20b82f359731e9d

### DIFF
--- a/app/src/ai/blocklist/agent_view/agent_input_footer/mod.rs
+++ b/app/src/ai/blocklist/agent_view/agent_input_footer/mod.rs
@@ -50,9 +50,7 @@ pub(crate) use self::environment_selector::{
 };
 use crate::ai::blocklist::agent_view::is_in_cloud_context;
 use crate::ai::blocklist::history_model::{BlocklistAIHistoryEvent, BlocklistAIHistoryModel};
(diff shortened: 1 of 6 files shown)


harryalbert:
* **#12748** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12748?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12748?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #12753, opened 2026-06-17
Fix cross window dragging splitting groups

johnturcoo:
## Description

Fixes two main bugs with interactions between cross window tab dragging and tab groups. 
1. When dropping a cross window tab in the middle of a tab group, it could cause the tab group to split. Now, we only allow dropping before or after a tab group, and trying to drop in the middle anchors it past the tab group. 
2. When a tab group was collapsed, the tab list was incorrectly rendering the insertion position for the cross window tab landing position. Now the index is calculated correctly and renders the landing position accordingly. 

I think I will do a follow up PR that actually allows dropping a cross window tab into an existing tab group, however IMO this is non blocking and not high priority given the additional complexity. **Please let me know what you think.**  (This is complicated because we need to do some more calculations for group bounds, and can't just infer the group based on insertion index (inserting into the index before and after a group could mean first/last position in the group or just outside the group, also need to modify group rendering).

## How it works 

- Created a helper function `tab_bar_slots()` that converts the tab list into a list of Tab bar slots which is essentially an enum representing a group or an individual tab.  This was previously used for rendering the tab/groups in the tab list but is now also used for computing the insertion index for the cross window tab
- Created a helper function `clamp_past_group()` that clamps the `ghost tab` and the insertion index past a group, so that tabs can not land inside of groups and split them.
- Updated insertion calculations to use the group rects when applicable rather than solely the tab rects 

## Linked Issue

## Testing

### Screenshots / Videos

[Demo of bug ](https://www.loom.com/share/53318faf76e849308c91b69b25cfb59e)

[Demo of correct functionality vtabs](https://www.loom.com/share/a6e3845e5adb41e7ae12239f4209c3cd)

[Demo of correct functionality htabs ](https://www.loom.com/share/1deecc5e85b145ea96832975e29fda86)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -26915,7 +26915,6 @@ impl Workspace {
                     ..
                 } => {
                     if let Some(container_rect) = group_container_rect(window_id, group_id, ctx) {
-        
                         if container_rect.width() > MIN_VISIBLE_TAB_WIDTH
                             && rect_is_within_tab_bar(container_rect, &tab_bar_rects)
                         {
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -20149,40 +20149,10 @@ impl Workspace {
             // Ghost state for cross-window drag hovering over this tab bar.
             let ghost = drag_model.ghost_state_for_window(self.window_id);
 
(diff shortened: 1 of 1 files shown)


peicodes:
Definitely let's OOS cross-window tab dragging into groups

peicodes (approved):

PR #12754, opened 2026-06-17
[MCP][8/9] Resolve managed MCPs in agent run CLI

Legoben:
## Description

Adds runtime resolution of managed MCP servers for CLI agent runs. Previously, a `warp_id` UUID passed via `--mcp` would fail with `MCPServerNotFound` unless a local templatable installation with that exact UUID existed. This change introduces a local-first fallback: if the UUID is not found locally, the client calls the new `createManagedMcpClientConfig` GraphQL mutation to retrieve server-rendered MCP config and materializes it as an ephemeral installation for the duration of the run.

Key changes:

- Adds the `createManagedMcpClientConfig` GraphQL mutation to the schema and a corresponding `ManagedMcpClient` trait backed by `ServerApi`, exposed via `ServerApiProvider::get_managed_mcp_client()`.
- Introduces `ResolvedMcpSpecs` to carry both local UUIDs and ephemeral installations through the resolution pipeline.
- `resolve_mcp_specs` is now async and checks local installations first; only non-local UUIDs trigger a GraphQL call.
- Adds `installations_from_managed_client_config_json`, a managed-specific parsing helper that synthesizes `{{key}}` placeholder variable values for template variables not captured from `env`/`headers`, enabling command args like `["--token={{API_TOKEN}}"]` to resolve correctly via the existing `apply_secrets` path.
- Adds `AgentDriverError::ManagedMcpResolutionFailed` with UID and message, classified as `EnvironmentSetupFailed`.
- Resolved managed config is kept in memory only and is never written back to config files, snapshots, or logs.
- `ParsedTemplatableMCPServerResult` now exposes `variable_values` so the managed parsing helper can access captured values without re-parsing.

## Testing

New unit tests cover:

- Local UUID resolves without calling `ManagedMcpClient`.
- Non-local UUID calls `create_managed_mcp_client_config` and produces an ephemeral installation.
- Command config with an env placeholder resolves the secret correctly.
- Command config with an arg placeholder resolves the secret correctly.
- Env and arg placeholders coexist with literal env values.
- URL config preserves proxy URL and `Authorization` header unchanged.
- Missing secret leaves the placeholder in rendered output, matching existing `apply_secrets` behavior.
- GraphQL failure maps to `ManagedMcpResolutionFailed` with the correct UID and message.
- `ManagedMcpResolutionFailed` is classified as `AgentTaskState::Failed` with `PlatformErrorCode::EnvironmentSetupFailed`.

##

### DIFF
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -68,6 +68,7 @@ use crate::ai::mcp::parsing::{normalize_mcp_json, resolve_json, ParsedTemplatabl
 use crate::ai::mcp::templatable_manager::TemplatableMCPServerManagerEvent;
 use crate::ai::mcp::{
     JSONMCPServer, MCPServerState, TemplatableMCPServerInstallation, TemplatableMCPServerManager,
+    VariableType, VariableValue,
 };
 use crate::ai::skills::{
     filter_skills_by_spec, read_skills_from_directories, resolve_skill_repos, SkillManager,
@@ -81,6 +82,7 @@ use crate::server::server_api::ai::{AIClient, TaskStatusUpdate};
 use crate::server::server_api::harness_support::{
     HarnessSupportClient, ResolvePromptAttachedSkill, ResolvePromptRequest,
 };
+use crate::server::server_api::managed_mcp::ManagedMcpClient;
 use crate::server::server_api::ServerApiProvider;
 use crate::terminal::cli_agent_sessions::plugin_manager::{
     plugin_manager_for, CliAgentPluginManager,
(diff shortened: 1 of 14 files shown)


Legoben:
* **#12754** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12754?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12754?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

peicodes (approved):

PR #12757, opened 2026-06-17
Suppress transient "lost connection" banner during auto-recovery

advait-m:
## Description
While an automatic retry/resume is in flight after a transient transport failure, `render_failed_output` no longer surfaces the alarming **"Warp lost connection"** banner (plus raw debug info). Instead it renders nothing until recovery has actually failed, at which point the full error banner is shown as before.

Context: #12431 started synthesizing/surfacing transient transport errors (e.g. `UnexpectedEof` on a stream cut) and rendered them as "Warp lost connection… Attempting to resume" even when the auto-resume subsequently succeeded. For runs that recover on their own (the common case), this banner can be noisy and misleading.

Changes:
- `app/src/ai/blocklist/block/view_impl/common.rs`: early-return `Empty` from `render_failed_output` for `TransientNetworkError`/`Other` errors with `will_attempt_resume: true`. Simplified the now terminal-only match arms and removed the dead `with_resume_status` helper.
- `app/src/ai/blocklist/controller/response_stream.rs`: added a `log::warn!` to both `RecoveryAction::Resume` paths so auto-recovery isn't completely silent in the logs, matching the existing `RetryNow`/`RetryWhenOnline` logging. This is a one-time event (not the per-frame render path).

The full error banner (including debug info) is still shown once `will_attempt_resume` is `false` (recovery exhausted/failed).

### DIFF
--- a/app/src/ai/agent/mod.rs
+++ b/app/src/ai/agent/mod.rs
@@ -34,6 +34,7 @@ use session_sharing_protocol::common::ParticipantId;
 use task::TaskId;
 pub use telemetry::AIIdentifiers;
 use uuid::Uuid;
+use warp_core::channel::ChannelState;
 use warp_core::features::FeatureFlag;
 use warp_editor::render::model::LineCount;
 use warp_multi_agent_api::{diff_hunk as diff_hunk_api, AgentEvent, AgentType};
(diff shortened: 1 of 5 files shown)


seemeroland (commented):
lgtm, but do you have a way of testing this? Maybe we should add something on the server to kill the connection the first 2 times? I did something similar for testing retries for shared sessions

advait-m:
> lgtm, but do you have a way of testing this? Maybe we should add something on the server to kill the connection the first 2 times? I did something similar for testing retries for shared sessions

Yep - done. Mocked reset error on server:
- Confirmed I see the error + recovery if I do 1 error without this branch
- Confirmed I don't see the UI for error/recovery if I do 1 error with this branch
- Confirmed I see error if I do 2 errors with this branch

image

image

image

advait-m:
Oh yeah let me make this aggressive on Dev-only too

advait-m:
> Oh yeah let me make this aggressive on Dev-only too

done

/oz-review

seemeroland (approved):
Do we only retry one time and that's why we show it after 2 failures? Feel like we should retry 2 times

advait-m:
> Do we only retry one time and that's why we show it after 2 failures? Feel like we should retry 2 times

Right now yep, we only retry once (for this class of errors)

Yeah I agree we can retry more times - I'll split that into a separate PR

PR #12760, opened 2026-06-17
BYOK/BYOE for Enterprise Product Spec

JaidenRatti:
Product spec for BYOK/BYOE support for Enterprises


specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- A "user-managed" provider is a key or endpoint a member added in their own Warp settings - stored only on that member's device (OS secure storage). The client attaches the secret to each interactive request, where the server uses it transiently and never persists it (current BYOK/BYOE for self-serve behaviour). 
+
+### Admin Panel — Models page (Enterprise Team Admin)
+- The Models page gains a "Bring Your Own Keys & Endpoints" section alongside the existing Direct API and AWS Bedrock sections. The section has a master enable toggle. When off, no team-managed providers are active for the team and the section's configuration controls are hidden/disabled. When on, the admin can configure team keys and endpoints and they become available to members. When the admin enables team-managed BYO (adds keys/endpoints) while Direct API is still enabled, the section surfaces a prominent prompt explaining that members can still select Warp-managed models, with a one-click affordance to turn off Direct API.
```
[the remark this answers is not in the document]
JaidenRatti: As discussed in our meeting, a very clear prompt that the enterprise admin should disable direct api was determined to be sufficient.

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- Team API keys: the admin can paste a key for each first-party provider Warp supports for BYOK (e.g. OpenAI, Anthropic, Google). Each provider row shows whether a key is currently set without revealing the stored value. Saving a key persists it for the team; clearing a row removes that team key.
+- Team custom endpoints: the admin can add one or more OpenAI-compatible Chat Completions endpoints. Each endpoint has a name, URL, API key, and one or more models, where each model has a model name (sent to the endpoint) and an optional alias (shown to members). The admin can add, edit, and remove endpoints.
+  - Endpoint validation: an endpoint cannot be saved without a name, a valid URL, an API key, and at least one model with a non-empty model name. Invalid fields are indicated inline and block saving (same behaviour as current client).
```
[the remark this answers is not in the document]
JaidenRatti: We are using the pre-existing machinery of processing requests to custom endpoints.

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- The "Provided by your team" group lists the enabled team-managed providers and team-managed endpoints (name + model chips). Providers or models the admin has disabled do not appear here, mirroring the picker (disabling a model in the Admin Panel removes it from the member's picker automatically). It is read-only: members cannot edit, add, or remove these entries.
+- The team group includes a short explanation that these were configured by the team admin, are shared with everyone, and also power cloud agents.
+- The "User added keys" group is the existing self-serve BYOK/BYOE experience: paste provider keys, and add/edit/remove custom endpoints (name, URL, API key, model name + alias). It carries a one-line description clarifying that these stay on the member's device and are not available to cloud agents.
```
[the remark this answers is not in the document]
JaidenRatti: Good find.

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- User-managed models also appear in the picker when "Allow users to bring their own models" is on.
+- Currently, custom-endpoint models appear in the picker as `<model alias> (Custom • <endpoint name>)`. Since an enterprise admin may set endpoint name `a`, and the user may also set an endpoint with name `a`, we always disambiguiate by showing Team-provided endpoints as `<model alias> (Team • endpoint name)`, and user-provided endpoints as `<model alias> (Custom • endpoint name)`. i.e. we are doing Team provided custom endpoints `UNION` User provided custom endpoints. 
+- First-party provider keys (OpenAI/Anthropic/Google) never create duplicate picker entries: a member's own key for a provider takes precedence over the team key for that provider, so the standard model appears once — routing through the member's key when set, otherwise the team key. This precedence is automatic; there is no member-facing toggle in MVP.
+
+### Cloud agents (Oz)
+- Team-managed keys and endpoints are usable by Oz cloud agents for that team: a cloud agent run can perform inference through the team's configured providers without any per-member device state.
```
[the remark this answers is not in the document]
JaidenRatti: I need to look into this more.

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- Turning "Allow users to bring their own models" off does not delete a member's locally stored keys/endpoints; it disables the UI and prevents their selection/use until re-enabled.
+
+### How team-managed providers are stored, synced, and injected
+This is meant to just be a high-level overview for alignment, more details will be in the tech spec. 
+
+Team-managed configuration is split into public metadata and secrets.
+- **Public metadata** syncs from the server down to every member's client through the existing workspace-settings channel and never includes secret values. What syncs differs by provider type:
+  - **BYOK (OpenAI/Anthropic/Google):** just a per-provider boolean — whether the team has a key configured. The provider is an already-known, built-in identity, so no value or id is synced. That boolean drives the member's "configured" checkmark and tells the picker which standard models can route through a team key.
+  - **Custom endpoints:** Endpoint name + URL, and each model's name, alias, and reference id (`config_key`) — since the client knows nothing about an admin-defined endpoint otherwise. Only the endpoint's API key is withheld.
```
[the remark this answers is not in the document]
JaidenRatti: This was addressed in the next commit.

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
+# Enterprise BYOK / BYOE
+
+## Summary
+Let enterprise teams use their own LLM providers in Warp. A team admin configures shared API keys and OpenAI-compatible custom endpoints (e.g. OpenRouter, LiteLLM, Samba) once in the Warp Admin Panel, and those providers appear automatically in every member's model picker. Admins can also choose whether members may add their own local keys and endpoints. Team-managed provider keys are stored by Warp, so they work for both interactive agent requests and Oz cloud agents; member-managed providers are stored locally in secure storage (the current art), and work only for interactive requests
```
jefflloyd, line 4: We might want user-level keys to be manageable as well. Like, I wonder if a better framing of this product is "add Warp-managed BYOK/BYOE so that 1) keys can be centrally managed by a team, rather than done on a per-user basis, and 2) BYOK/BYOE work for cloud agents"

jefflloyd: I see you call this out as a non-goal below - I think in the interest of sprinting to July 2nd we can leave this as a fast follow, but when building, let's just make sure we don't paint ourselves into a corner by making it hard to add it later

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Design
+A [Loom walkthrough](https://www.loom.com/share/6eb6e0f8f0764d2a9e8ec4f886957e63) with basic flows. Treat the Loom as the visual reference for the admin section and the member settings layout; this Behavior section is the source of truth for behavior.
+
+## Behavior
+
+### Roles and terms
+- "Team admin" is a member with admin/owner permissions; only admins see and edit the Admin Panel. 
+- "Team member" is any member; members see the in-app settings surface. 
+- A "team-managed" provider is a key or endpoint an admin configured in the Admin Panel. Its secret is stored server-side by Warp and never synced to member devices. The member's client references the provider by ID; the server injects the stored secret at request time.
```
jefflloyd: More in the technical weeds, but what do you mean by ID? Are you saying that if I pick "claude opus 4.8 (max)", then the server decides if it should go to a public endpoint or a BYO endpoint? Do I as the client know what models are BYO vs public? If so, how?

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- Team custom endpoints: the admin can add one or more OpenAI-compatible Chat Completions endpoints. Each endpoint has a name, URL, API key, and one or more models, where each model has a model name (sent to the endpoint) and an optional alias (shown to members). The admin can add, edit, and remove endpoints.
+  - Endpoint validation: an endpoint cannot be saved without a name, a valid URL, an API key, and at least one model with a non-empty model name. Invalid fields are indicated inline and block saving (same behaviour as current client). 
+- "Allow users to bring their own models" toggle: when on, members may add their own local keys and custom endpoints in their Warp settings; when off, the member-facing self-serve BYO UI is disabled.
+- Cloud-agent note: the section states that team-managed keys and endpoints are stored by Warp and are therefore available to Oz cloud agents, unlike member-managed providers which never leave the member's device.
+- Saving any team-managed configuration propagates to members: the next time a member's client loads team settings, the team-managed providers (and the "allow users" permission) reflect the admin's latest saved state.
```
jefflloyd, line 38: I think it's less the team settings and more the available models, right?

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- Validation errors from saving (e.g. malformed endpoint) are shown to the admin and the prior saved state is preserved until a valid save succeeds.
+
+### Team member — Warp settings surface
+- The member's AI/Custom Inference settings present two clearly distinct groups: "Provided by your team" (team-managed, read-only) shown first, and "User added keys" (the member's own, editable) shown below. Both groups use the same visual layout (provider key rows and endpoint cards). _The API key section of the "Provided by your team" will not show the redacted API key, instead just whether the API key is configured (checkmark) or not, and whether it's active or overridden by the "user added keys"_.
```
jefflloyd, line 42: Interested to see how this ends up looking in-app!

jefflloyd: Maybe goes without saying but we should keep the page simple when possible - i.e., if you don't have any team-provided keys (either because you're not on enterprise or your enterprise hasn't supplied any) we can keep it simple. Also, bear in mind most enterprises don't have BYOK/BYOE enabled.

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+- When "Allow users to bring their own models" is on, the "User added keys" group is fully interactive as it is for self-serve users today (excluding the SuperGrok toggle). 
+
+### Model picker and routing behaviour/logic
+- Team-managed models (each team endpoint's models, shown by alias when set, and the providers the team supplies) appear in the member's model picker.
+- User-managed models also appear in the picker when "Allow users to bring their own models" is on.
+- Currently, custom-endpoint models appear in the picker as `<model alias> (Custom • <endpoint name>)`. Since an enterprise admin may set endpoint name `a`, and the user may also set an endpoint with name `a`, we always disambiguiate by showing Team-provided endpoints as `<model alias> (Team • endpoint name)`, and user-provided endpoints as `<model alias> (Custom • endpoint name)`. i.e. we are doing Team provided custom endpoints `UNION` User provided custom endpoints.
```
jefflloyd: This is a good edge case, but we might want to keep it simple if there are no conflicts (or, e.g., there are no user-supplied keys)

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Model picker and routing behaviour/logic
+- Team-managed models (each team endpoint's models, shown by alias when set, and the providers the team supplies) appear in the member's model picker.
+- User-managed models also appear in the picker when "Allow users to bring their own models" is on.
+- Currently, custom-endpoint models appear in the picker as `<model alias> (Custom • <endpoint name>)`. Since an enterprise admin may set endpoint name `a`, and the user may also set an endpoint with name `a`, we always disambiguiate by showing Team-provided endpoints as `<model alias> (Team • endpoint name)`, and user-provided endpoints as `<model alias> (Custom • endpoint name)`. i.e. we are doing Team provided custom endpoints `UNION` User provided custom endpoints. 
+- First-party provider keys (OpenAI/Anthropic/Google) never create duplicate picker entries: a member's own key for a provider takes precedence over the team key for that provider, so the standard model appears once — routing through the member's key when set, otherwise the team key. This precedence is automatic.
```
jefflloyd: Why are they different?

specs/byok-byoe-enterprise/PRODUCT.md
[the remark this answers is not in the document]
jefflloyd: I wouldn't worry too much about this - we're telling customers that their endpoints must be reachable on the open internet.

specs/byok-byoe-enterprise/PRODUCT.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## Open Questions
+
+- Should admins be able to prevent team members from adding their own API key and Custom Endpoints separately? (i.e. it's okay for users to add their own API key, but it's not okay for them to add their own Custom Endpoint). Do we want separate toggles, or should we just treat these the same? **My vote**: Unless enterprise demand for this is strong, I think this is fine to be under the same toggle.
+- Should we default to "Allow users to bring their own models" for a newly enabled team? **My vote**: No, we should not.
```
jefflloyd, line 82: Agree... I think

jefflloyd (approved):
This looks great - a lot of my comments are on the fringes of importance, you've got a great core of the spec.

Not sure if it goes here or in the tech spec, but I would encourage you to think about how you'll surface this on the policy level - like, we normally disable BYOK/BYOE for enterprises, but in this new world, will we enable them? Or is there a new kind of policy "ENTERPRISE_BYOK_BYOE" that will be enabled? I haven't figured out exactly how that should work myself yet

specs/byok-byoe-enterprise/PRODUCT.md
JaidenRatti: Agreed that is a good framing of the product (since your second point then applies for both enterprises **&** self-serve users). Will be sure to keep that capability in mind when building.

specs/byok-byoe-enterprise/PRODUCT.md
JaidenRatti: Ack. I need to update this product spec on how the model picker will work, I changed how some of this is going to work while prototyping.

specs/byok-byoe-enterprise/PRODUCT.md
JaidenRatti: Agreed on it keeping things simple where possible. Planning to reuse the existing rust client work where I can for this.

specs/byok-byoe-enterprise/PRODUCT.md
JaidenRatti: My thought for this was it's safer to err on the side of caution. It's _probably_ better to have the worst-case of an enterprise being confused that they have to manually enable user provided BYOK/BYOE as opposed to finding out their users have been routing requests through their keys.

specs/byok-byoe-enterprise/PRODUCT.md
JaidenRatti: I think it would be both. The loom in this [thread](https://warpdev.slack.com/archives/C0BA8H7S46B/p1781644299864679?thread_ts=1781634670.557639&cid=C0BA8H7S46B) (from the very first meeting) might help with visual context (it will be much more simplified than this). But what I meant here is if the team admin says "users cant bring their own models", then we need to update the clients settings somehow to prevent them from adding models. Same goes for if they want to let users bring their own models, we have to update the users settings page to let them add models. This is what I meant by "team settings".

specs/byok-byoe-enterprise/PRODUCT.md
JaidenRatti: Not 100% sure on what you are referring to.

Line 52: How we would treat BYOE

Line 53: How we would treat BYOK (of which we support users bringing keys for OAI, Anthropic, and Google). 

For BYOK, new keys don't introduce new models to the model picker. It just changes which key is used for the same model. So if the team admin provides an OAI key, the user would see GPT models as normal in their picker (with a special key icon or something to indicate that their team admin has provided a key for it). And if the user adds a key themselves, we use their key. But if I haven't answered your question pls lmk!

specs/byok-byoe-enterprise/PRODUCT.md
JaidenRatti: I'm planning to go into more depth in the tech doc on this, but will update the product doc as well. 

ID here is for custom endpoints. 

For team managed custom endpoints, the client doesn't receive an endpoint URL or API key, it receives the endpoint name, slug, model display name, and a `config_key`. If the user selects a custom model, it sends the config key as the model id, and the server resolves it to the right team endpoint at request time. 

For first party API keys, there aren't separate provider IDs per team key. If the user selects Claude Opus 4.8 (max), the client sends the normal Warp model ID. The server infers it's an Anthropic model and then applies the precedence:
1. User-provided Anthropic key, if present and allowed;
2. Team-managed Anthropic key, if configured;
3. Warp-managed inference, if allowed. 

It gets a little hairy when the user has a custom model called "Claude Opus 4.8 (max)". 

Regardless, the server is going to have to pass some information to the client (like "team provided an anthropic key"), so we can show some sort of indication (via icons in the model picker / text in the "Inference" model picker section) to the user if their request is going to hit team provided, user provided, or warp-provided.

PR #12763, opened 2026-06-17
use the "emitter_handle" parameter instead of cloning strong handle i…

acarl005:
…n EditorView subscription

## Description

This PR removes a strong handle cycle _a. la._ https://github.com/warpdotdev/warp/pull/12355

## Linked Issue

None. This is a misc cleanup.


lucieleblanc (approved):
Nice!

PR #12767, opened 2026-06-18
add emitter_handle parameter to model subscriptions

acarl005:
## Description

This PR brings `ModelContext::subscribe_to_view` and `ModelContext::subscribe_to_model` to parity with their `ViewContext` counterparts. We add this parameter so that subscribers can access the thing they subscribed to inside the closure. Without that, closures need to capture handles and it turns out that 5 callers are capturing strong handles which undermines the design of the UI framework to not hold strong handles in subscribers. This is a high risk of forming cycles and causing memory leaks.

See context: https://warpdev.slack.com/archives/C0ANKD0Q5JL/p1781737089932179

The diff is super bloated but I've left inline comments on actual salient parts.

### DIFF
--- a/app/src/settings/ai_tests.rs
+++ b/app/src/settings/ai_tests.rs
@@ -46,12 +46,11 @@ fn test_update_both_values_changed() {
 
         // Setup event tracking
         let (sender, receiver) = async_channel::unbounded();
-        let model_handle_clone = model_handle.clone();
-        model_handle.update(&mut app, move |_, ctx| {
+        app.update(|ctx| {
             let sender = sender.clone();
             ctx.subscribe_to_model(
-                &model_handle_clone,
-                move |_, _, event: &FocusedTerminalInfoEvent, _| match event {
+                &model_handle,
+                move |_, event: &FocusedTerminalInfoEvent, _| match event {
                     FocusedTerminalInfoEvent::TerminalInfoUpdated => {
                         let _ = sender.try_send(());
                     }
@@ -87,12 +86,11 @@ fn test_update_additional_value_changed() {
 
         // Setup event tracking
         let (sender, receiver) = async_channel::unbounded();
-        let model_handle_clone = model_handle.clone();
-        model_handle.update(&mut app, move |_, ctx| {
+        app.update(|ctx| {
             let sender = sender.clone();
             ctx.subscribe_to_model(
-                &model_handle_clone,
-                move |_, _, event: &FocusedTerminalInfoEvent, _| match event {
+                &model_handle,
+                move |_, event: &FocusedTerminalInfoEvent, _| match event {
                     FocusedTerminalInfoEvent::TerminalInfoUpdated => {
                         let _ = sender.try_send(());
                     }
@@ -136,12 +134,11 @@ fn test_update_no_change() {
 
         // Setup event tracking
         let (sender, receiver) = async_channel::unbounded();
-        let model_handle_clone = model_handle.clone();
-        model_handle.update(&mut app, move |_, ctx| {
+        app.update(|ctx| {
             let sender = sender.clone();
             ctx.subscribe_to_model(
-                &model_handle_clone,
-                move |_, _, event: &FocusedTerminalInfoEvent, _| match event {
+                &model_handle,
+                move |_, event: &FocusedTerminalInfoEvent, _| match event {
                     FocusedTerminalInfoEvent::TerminalInfoUpdated => {
                         let _ = sender.try_send(());
                     }
@@ -185,12 +182,11 @@ fn test_update_only_remote_toggles() {
 
         // Setup event tracking
         let (sender, receiver) = async_channel::unbounded();
-        let model_handle_clone = model_handle.clone();
-        model_handle.update(&mut app, move |_, ctx| {
+        app.update(|ctx| {
             let sender = sender.clone();
             ctx.subscribe_to_model(
-                &model_handle_clone,
-                move |_, _, event: &FocusedTerminalInfoEvent, _| match event {
+                &model_handle,
+                move |_, event: &FocusedTerminalInfoEvent, _| match event {
                     FocusedTerminalInfoEvent::TerminalInfoUpdated => {
                         let _ = sender.try_send(());
                     }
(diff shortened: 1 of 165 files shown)


crates/warpui_core/src/core/model/context.rs
```diff
@@ -67,20 +67,23 @@ impl<'a, T: Entity> ModelContext<'a, T> {
...
             .or_default()
             .push(Subscription::FromModel {
                 model_id: self.model_id,
                 callback: Box::new(move |model, payload, app, model_id| {
-                    let model = model.downcast_mut().expect("downcast is type safe");
-                    let payload: &<S as Entity>::Event =
-                        payload.downcast_ref().expect("downcast is type safe");
-                    let mut ctx = ModelContext::new(app, model_id);
-                    callback(model, payload, &mut ctx);
+                    if let Some(emitter_handle) = emitter_handle.upgrade(app) {
```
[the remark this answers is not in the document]
acarl005: > when a model subscribes to its own events, this `upgrade` returns `None`

A model should never be subscribing to it's own events. I used [dylint](https://github.com/trailofbits/dylint) and vibe-coded a custom lint rule to catch that scenario. I even purposely added a self-subscription to make sure the custom lint worked. I can guarantee that nowhere in the app do models subscribe to themselves.

crates/warpui_core/src/core/ref_count_tests.rs
```diff
@@ -8,48 +8,55 @@ use super::*;
 /// `WeakModelHandle::upgrade` attempted afterward must return `None`.
 #[test]
 fn test_weak_handle_fails_after_last_strong_handle_dropped_in_event_callback() {
```
acarl005, line 10: This test actually started failing. It's asserting that a model gets dropped in the middle of a subscription closure. Now that the closure takes a handle to the subscribee, it will prevent the subscribee from being dropped until the end of the closure.

I changed the test so that the model getting dropped isn't the same one getting subscribed to. 

This is important to flag b/c it highlights that we've changed the semantics a little bit with the drop order.

crates/warpui_core/src/core/model/context.rs
```diff
@@ -67,20 +67,27 @@ impl<'a, T: Entity> ModelContext<'a, T> {
...
-                    let model = model.downcast_mut().expect("downcast is type safe");
-                    let payload: &<S as Entity>::Event =
-                        payload.downcast_ref().expect("downcast is type safe");
-                    let mut ctx = ModelContext::new(app, model_id);
-                    callback(model, payload, &mut ctx);
+                    // Use `upgrade_unchecked` so that self-subscriptions work:
+                    // `emit_event` temporarily removes the subscriber model from
+                    // `app.models` before invoking callbacks, so the standard
+                    // `upgrade` would return `None` and silently drop the callback.
+                    if let Some(emitter_handle) = emitter_handle.upgrade_unchecked(app) {
```
acarl005: This part is actually non-trivial. See the comment here.

crates/warpui_core/src/core/model/handle.rs
```diff
@@ -226,6 +226,15 @@ impl<T: Entity> WeakModelHandle<T> {
...
     }
+
+    /// Like [`Self::upgrade`] but does not require the model to be present in `app.models`.
+    pub(in crate::core) fn upgrade_unchecked(&self, app: &AppContext) -> Option<ModelHandle<T>> {
+        if !app.ref_counts.lock().is_model_dropped(self.model_id) {
+            Some(ModelHandle::new(self.model_id, &app.ref_counts))
+        } else {
+            None
+        }
+    }
```
acarl005: New method to enable models to subscribe to themselves.

app/src/terminal/warpify/settings.rs
```diff
@@ -277,13 +277,13 @@ impl WarpifySettings {
...
-    /// model subscribe to itself after it's registered.
+    /// This is different from the typical register method, as it also ensures
+    /// that our parsed regexes stay in sync with the underlying data by
+    /// subscribing to the model's change events at the app level.
     pub fn register(ctx: &mut AppContext) {
         let handle = ctx.add_singleton_model(Self::new_from_storage);
-        handle.clone().update(ctx, |_, ctx| {
-            ctx.subscribe_to_model(&handle, |me, event, _| match event {
+        ctx.subscribe_to_model(&handle, |settings, event, ctx| {
+            settings.update(ctx, |me, _| match event {
```
acarl005, line 286: Remove the one self-subscription we have in production code.

app/src/terminal/warpify/settings_tests.rs
```diff
@@ -1,5 +1,43 @@
-#[cfg(windows)]
+use settings::Setting;
+use warpui::{App, SingletonEntity};
+
 use super::WarpifySettings;
+use crate::test_util::settings::initialize_settings_for_tests;
+
+#[test]
+fn test_parsed_subshell_commands_updated_via_self_subscription() {
```
acarl005, line 8: Added a test to make sure we didn't break the functionality with the self-subscription.

crates/warpui_core/src/core/model/context.rs
```diff
@@ -67,20 +67,31 @@ impl<'a, T: Entity> ModelContext<'a, T> {
...
+        F: 'static + FnMut(&mut T, ModelHandle<S>, &S::Event, &mut ModelContext<T>),
     {
+        // Self-subscriptions are disallowed: `emit_event` temporarily removes the subscriber model
+        // from `app.models` before invoking callbacks, so `emitter_handle.upgrade` would return
+        // `None` and silently drop the callback.
+        debug_assert_ne!(
+            handle.id(),
+            self.model_id,
+            "a model must not subscribe to its own events"
+        );
```
acarl005, line 79: Make sure we disallow self-subscriptions

app/src/settings/ai_tests.rs
```diff
@@ -46,11 +46,10 @@ fn test_update_both_values_changed() {
 
         // Setup event tracking
         let (sender, receiver) = async_channel::unbounded();
-        let model_handle_clone = model_handle.clone();
-        model_handle.update(&mut app, move |_, ctx| {
+        app.update(|ctx| {
```
acarl005, line 49: We also had some self-subscriptions happening in tests.

crates/warpui_core/src/core/model/handle.rs
acarl005: deleted. we are dis-allowing self subscriptions

vorporeal (approved):

PR #12776, opened 2026-06-18
Fix Ctrl+Tab switcher: carry tab color to icon and lead with title

### LINKED ISSUE #12692: Ctrl+Tab switcher previews ignore tab color-coding and bury the title behind the index
### Pre-submit checks

- I searched the open issues and found no duplicate (closest is #6899, which is only about the subtitle path, not color/title).

### Describe the bug

Tabs can be **renamed** and **color-coded** in the tab bar/sidebar, and those customizations show up there. But the **Ctrl+Tab switcher** (the MRU palette shown when "Cycle most recent tabs/sessions" is bound to Ctrl+Tab) ignores the color entirely and de-emphasizes the title, so every entry looks nearly identical and is hard to scan quickly.

Two concrete problems with each switcher row:

1. **No color.** A tab's color-coding is never carried into the switcher, so colored tabs are visually indistinguishable from one another. The whole point of color-coding tabs is fast visual identification, which is exactly what a switcher should support.
2. **Title is buried.** Each row is rendered as `[Tab 3] my-title`. The leading `[Tab N]` index is read first; the (often custom) title — the thing you actually recognize — comes second.

### To reproduce

1. Settings → Keyboard → set Ctrl+Tab to "Cycle through most recently used tabs" (or sessions).
2. Open several tabs, rename a few, and assign each a different color in the tab bar.
3. Press and hold Ctrl, tap Tab to open the switcher.
4. Observe: rows are all the same color, and each row leads with `[Tab N]` rather than the title.

### Expected behavior

The switcher should reflect the same customizations as the tab bar:

- Each row shows a **color indicator** matching the tab's color-coding.
- The **title leads**, with the numeric index kept as a secondary disambiguation hint.

### Root cause (from reading the source)

- `TabNavigationData` (`app/src/session_management.rs`) — the per-row model for the switcher — has no `color` field, so color never reaches the renderer. (The title *is* already passed via `display_title()`, so custom names are present — just not shown first.)
- `Workspace::tab_navigation_data()` (`app/src/workspace/view.rs`) builds that struct but never reads `tab.color()`, even though `get_tab_color()` right below it shows the color is readily available.
- The row renderer `SearchItem::render_item` / `render_icon` (`app/src/search/command_palette/tabs/search_item.rs`) renders an untinted `Icon::Navigation` and formats the label as `format!("[Tab {}] {}", tab_index, title)`.

### Proposed fix

Small, self-contained change:

1. Add `color: Option<AnsiColorIdentifier>` to `TabNavigationData`.
2. Populate it from `tab.color()` in `tab_navigation_data()`.
3. In `render_icon`, tint the leading icon with the tab color (converted via the existing `AnsiColorIdentifier::to_ansi_color` + `ColorU` path, with contrast handled by the existing `render_search_item_icon`); fall back to the default fill when a tab has no color.
4. Reorder the label to lead with the title: `format!("{}  ·  Tab {}", title, tab_index)`.

I have a working patch for this (~19 lines across the 3 files) and am happy to open a PR once this is labeled `ready-to-implement`.

### Operating system

macOS (Darwin 25.5.0)

### Shell version

zsh 5.9

GeekLuffy:
## Description
This PR improves the usability and visual hierarchy of the `Ctrl+Tab` MRU switcher. 

Specifically:
1. **Ignores Tab Colors:** The MRU switcher previously ignored custom tab color-coding. This change ensures that if a tab has a custom color (either manual override or directory default), the switcher icon is tinted with that color.
2. **Text Formatting:** Changed the layout format of switcher rows from `[Tab N] Title` to `Title · Tab N` so that the actual tab name is prioritized and easier to scan.

## Changes
- Added a `color` field (`Option<AnsiColorIdentifier>`) to `TabNavigationData` and populated it in `Workspace::tab_navigation_data()`.
- Updated `SearchItem::render_icon` to use the tab color if available, falling back to the standard highlight fill.
- Updated `SearchItem::render_item` to display labels in the `Title · Tab N` format.

## Linked Issue
Fixes #12692

## Visual Proof
ctrl_tab_switcher_demo_1781770421433

## Testing
- [x] Verified switcher renders tab icons with correct user-defined or default directory colors.
- [x] Verified tab titles are correctly displayed first, followed by the index as a trailing detail.

### DIFF
--- a/app/src/search/command_palette/tabs/search_item.rs
+++ b/app/src/search/command_palette/tabs/search_item.rs
@@ -36,7 +36,13 @@ impl SearchItemTrait for SearchItem {
         highlight_state: ItemHighlightState,
         appearance: &Appearance,
     ) -> Box<dyn Element> {
-        let color = highlight_state.icon_fill(appearance).into_solid();
+        let color = if let Some(tab_color) = self.tab.color {
+            tab_color
+                .to_ansi_color(&appearance.theme().terminal_colors().normal)
+                .into()
+        } else {
+            highlight_state.icon_fill(appearance).into_solid()
+        };
         render_search_item_icon(appearance, Icon::Navigation, color, highlight_state)
     }
 
@@ -54,7 +60,7 @@ impl SearchItemTrait for SearchItem {
         let appearance = Appearance::as_ref(app);
 
         let title_text = Text::new_inline(
-            format!("[Tab {}] {}", self.tab.tab_index, self.tab.title),
+            format!("{} · Tab {}", self.tab.title, self.tab.tab_index),
             appearance.ui_font_family(),
             appearance.monospace_font_size(),
         )
--- a/app/src/session_management.rs
+++ b/app/src/session_management.rs
@@ -7,6 +7,7 @@ use crate::context_chips::prompt_snapshot::PromptSnapshot;
 use crate::pane_group::{PaneGroup, PaneId};
 use crate::terminal::model::blockgrid::BlockGrid;
 use crate::terminal::shared_session::SharedSessionStatus;
+use crate::themes::theme::AnsiColorIdentifier;
 use crate::workspace::{PaneViewLocator, Workspace};
 
 /// Contains session metadata, including a prompt and running command (if there is one).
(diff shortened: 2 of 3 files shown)


GeekLuffy:
@cla-bot check

acarl005 (approved):

acarl005:
@GeekLuffy Thank you for the fix! This is a much better experience. Also that's dope username heh

GeekLuffy:
@acarl005 Thanks for the review and the kind words! 😄 Happy to contribute and looking forward to more!

PR #12800, opened 2026-06-18
Include `launch_mode` field in `run_internal` spans.

vorporeal:
## Description
Adds a `launch_mode` field to the root `run_internal` span so startup traces can be filtered by how Warp was started, including the specific Oz CLI command.

Tracing labels are defined exhaustively beside each CLI command enum and delegated through `CliCommand`, keeping `LaunchMode` decoupled from nested command details.

## Testing
- `cargo fmt -p warp -p warp_cli`
- `cargo check -p warp`
- Not manually tested; this only changes tracing metadata.

CHANGELOG-NONE

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -172,18 +172,6 @@ use terminal::local_shell::LocalShellState;
 pub use util::bindings::cmd_or_ctrl_shift;
 use voice::transcriber::VoiceTranscriber;
 use warp_cli::agent::AgentCommand;
-use warp_cli::api_key::ApiKeyCommand;
-use warp_cli::artifact::ArtifactCommand;
-use warp_cli::environment::EnvironmentCommand;
-use warp_cli::federate::FederateCommand;
-use warp_cli::harness_support::HarnessSupportCommand;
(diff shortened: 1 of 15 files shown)


vorporeal:
* **#12800** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

captainsafia (approved):
Looks good! It would be nice if we captured the argument keys that were passed (not values) specifically for the run command so we can discern what permutation of args a command was called with but not strictly necessary.

vorporeal:
> Looks good! It would be nice if we captured the argument keys that were passed (not values) specifically for the run command so we can discern what permutation of args a command was called with but not strictly necessary.

yeah, for now, there's a specific odd issue i'm trying to debug with this; i think it's probably best if we add additional information to trace spans as-needed.

PR #12806, opened 2026-06-18
prevent trying to use 0-byte corrupted installer file in auto-update

acarl005:
## Description

Windows auto-update was failing on my machine. When I click "update and restart" button, Warp would exit and then... no installer would take over! I found this line in the Warp logs:

```
2026-06-18T22:30:59Z [ERROR] Error relaunching app after autoupdate: %1 is not a valid Win32 application. (os error 193)
```

I inspected the installer inside %TEMP% and the installer was 0 bytes long. Unfortunately, the file was corrupted (probably a network error) and Warp doesn't try re-downloading if the file exists.

This PR adds a simple validation to make sure the file isn't empty and try re-downloading if it is. There are definitely better validations to add here but this is a good stop-gap for now.


abhishek-pandya (approved):
Nice find!

PR #12826, opened 2026-06-19
Add keyboard shortcut to toggle Markdown rendered/raw view

### LINKED ISSUE #11786: Add keyboard shortcut to toggle Markdown raw/rendered view
## Summary
A user requested a keyboard shortcut for toggling the Markdown viewer between rendered and raw/source mode. They noted that VS Code uses Cmd+Shift+V and Obsidian uses Cmd+E, so they instinctively reach for a shortcut when switching modes in Warp.

## Requested behavior
- Provide a configurable or default keybinding to toggle between rendered Markdown and raw/source Markdown in the Markdown viewer.
- Consider familiar defaults from other editors, such as Cmd+Shift+V or Cmd+E, if they do not conflict with existing Warp shortcuts.

## Context
- Internal Slack report: https://warpdev.slack.com/archives/C08KTPNQN65/p1779914198503059
- No attachments were included.

## Public-safety review
This issue was filed from a text-only feature request. It contains no customer identifiers, internal implementation details, logs, secrets, screenshots, or attachments.

maxmilian:
## Description

Add a keyboard shortcut to toggle a Markdown file pane between its rendered preview and its raw/source view, as requested in #11786.

Warp already exposes this toggle through the pane-header segmented control (rendered side) and the **View Markdown preview** context-menu entry (raw side), but there was no way to trigger it from the keyboard. This adds an editable keybinding that does so.

The shortcut defaults to **Cmd/Ctrl+E** — the Obsidian default called out in the issue. (VS Code's Cmd+Shift+V is already bound to paste in Warp, so it is intentionally avoided.) Because it is a normal editable binding, users can rebind it from the keybindings settings, which also satisfies the issue's "configurable" option.

Rendered Markdown lives in `FileNotebookView` and the raw/source view lives in `CodeView`, so the toggle is implemented as two single-direction bindings on the same key — one per view — that together form the round trip:

- `FileNotebookView` (rendered) → `ToggleMarkdownDisplayMode(Raw)` → swaps in the code pane
- `CodeView` (raw markdown) → `RenderMarkdown` → swaps back to the rendered file pane

Both bindings are registered view-wide, so each action handler is guarded to no-op on non-markdown panes (the existing segmented control and context-menu entries are already markdown-gated, so this keeps behavior consistent).

## Linked Issue

Closes #11786

## Testing
- [x] `cargo check -p warp --features local_fs`
- [x] `cargo clippy -p warp --features local_fs --all-targets -- -D warnings`
- [x] `cargo nextest run -p warp --features local_fs -E 'test(notebooks::file)'`

Added two regression tests in `app/src/notebooks/file/mod_tests.rs`:

- `test_markdown_toggle_is_noop_for_non_markdown_pane` — a pane with no markdown file ignores the toggle action.
- `test_markdown_toggle_switches_markdown_pane_to_raw` — a markdown pane switches to raw mode when the action fires.

### Screenshots / Videos

https://github.com/user-attachments/assets/06aa230e-8d71-4005-8671-e36a5e91df2d

The recording shows pressing **Cmd+E** in a Markdown file pane to switch from the rendered preview to the raw/source view, and pressing it again to switch back.

There is no separate "before" clip on purpose: this change only adds a keyboard path to the *existing* rendered/raw toggle (already reachable via the pane-header segmented control and the "View Markdown preview" context-menu entry). The panes themselves are unchanged — the new behavior is the Cmd+E shortcut itself, which the recording demonstrates.

## Agent Mode

### DIFF
--- a/app/src/code/view.rs
+++ b/app/src/code/view.rs
@@ -123,6 +123,17 @@ pub fn init(app: &mut AppContext) {
         .with_context_predicate(id!("CodeEditorView"))
         .with_key_binding("cmdorctrl-r u"),
     ]);
+
+    // Toggle a markdown code pane back to its rendered preview. Mirrors the
+    // `FileNotebookView` binding so the same shortcut toggles both directions.
+    #[cfg(feature = "local_fs")]
+    app.register_editable_bindings([EditableBinding::new(
+        "code_view:toggle_markdown_display_mode",
+        "Toggle Markdown rendered/raw view",
+        CodeViewAction::RenderMarkdown,
+    )
+    .with_context_predicate(text_entry)
+    .with_key_binding("cmdorctrl-e")]);
 }
 
 const PADDING: f32 = 4.;
@@ -2337,9 +2348,13 @@ impl TypedActionView for CodeView {
             }
             #[cfg(feature = "local_fs")]
             CodeViewAction::RenderMarkdown => {
+                // The keybinding is registered for all code panes, but only markdown
+                // files have a rendered preview to switch to. Drop the path otherwise so
+                // the action is a no-op (the context-menu entry is already markdown-gated).
                 let lor_path = self
                     .tab_at(self.active_tab_index)
-                    .and_then(|t| t.location.clone());
+                    .and_then(|t| t.location.clone())
+                    .filter(|p| is_markdown_file(std::path::Path::new(&p.display_path())));
 
                 if let Some(lor_path) = lor_path {
                     let source = self.source.clone();
--- a/app/src/code/view.rs
+++ b/app/src/code/view.rs
@@ -2351,10 +2351,17 @@ impl TypedActionView for CodeView {
                 // The keybinding is registered for all code panes, but only markdown
                 // files have a rendered preview to switch to. Drop the path otherwise so
                 // the action is a no-op (the context-menu entry is already markdown-gated).
+                //
+                // Detect markdown via the standardized path component (always
+                // `/`-separated) instead of display_path() + Path::new, which assumes
+                // local-OS encoding and would misparse a remote file's path on a
+                // cross-OS session.
                 let lor_path = self
                     .tab_at(self.active_tab_index)
                     .and_then(|t| t.location.clone())
-                    .filter(|p| is_markdown_file(std::path::Path::new(&p.display_path())));
+                    .filter(|p| {
+                        is_markdown_file(std::path::Path::new(p.path_component().as_str()))
+                    });
 
                 if let Some(lor_path) = lor_path {
                     let source = self.source.clone();
--- a/app/src/code/view.rs
+++ b/app/src/code/view.rs
@@ -2352,15 +2352,18 @@ impl TypedActionView for CodeView {
                 // files have a rendered preview to switch to. Drop the path otherwise so
                 // the action is a no-op (the context-menu entry is already markdown-gated).
                 //
-                // Detect markdown via the standardized path component (always
-                // `/`-separated) instead of display_path() + Path::new, which assumes
-                // local-OS encoding and would misparse a remote file's path on a
-                // cross-OS session.
+                // Detect markdown from the standardized path component (always
+                // `/`-separated) rather than display_path(), which is local-OS
+                // oriented and would misparse a remote file's path on a cross-OS
+                // session. is_markdown_file takes impl AsRef<Path> and only inspects
+                // the extension/file name, so feed it the standardized &str directly.
                 let lor_path = self
                     .tab_at(self.active_tab_index)
                     .and_then(|t| t.location.clone())
                     .filter(|p| {
-                        is_markdown_file(std::path::Path::new(p.path_component().as_str()))
+                        crate::util::openable_file_type::is_markdown_file(
+                            p.path_component().as_str(),
+                        )
                     });
 
                 if let Some(lor_path) = lor_path {
--- a/app/src/code/view.rs
+++ b/app/src/code/view.rs
@@ -2348,20 +2348,23 @@ impl TypedActionView for CodeView {
             }
             #[cfg(feature = "local_fs")]
             CodeViewAction::RenderMarkdown => {
-                // The keybinding is registered for all code panes, but only markdown
-                // files have a rendered preview to switch to. Drop the path otherwise so
-                // the action is a no-op (the context-menu entry is already markdown-gated).
+                // The keybinding is registered for all code panes, but only files that
+                // render in the notebook viewer have a rendered preview to switch to.
+                // Drop the path otherwise so the action is a no-op. This is the same
+                // predicate the header segmented control and the "View Markdown preview"
+                // context-menu entry use, so those existing entry points (including
+                // Jupyter notebooks) keep working unchanged.
                 //
-                // Detect markdown from the standardized path component (always
+                // Detect the file type from the standardized path component (always
                 // `/`-separated) rather than display_path(), which is local-OS
                 // oriented and would misparse a remote file's path on a cross-OS
-                // session. is_markdown_file takes impl AsRef<Path> and only inspects
-                // the extension/file name, so feed it the standardized &str directly.
+                // session. The predicate takes impl AsRef<Path> and only inspects the
+                // extension/file name, so feed it the standardized &str directly.
                 let lor_path = self
                     .tab_at(self.active_tab_index)
                     .and_then(|t| t.location.clone())
                     .filter(|p| {
-                        crate::util::openable_file_type::is_markdown_file(
+                        crate::util::openable_file_type::renders_in_warp_notebook_viewer(
                             p.path_component().as_str(),
                         )
                     });
--- a/app/src/notebooks/file/mod.rs
+++ b/app/src/notebooks/file/mod.rs
@@ -237,6 +237,13 @@ pub fn init(app: &mut AppContext) {
             FileNotebookAction::ReloadFile,
         )
         .with_context_predicate(id!("FileNotebookView")),
+        EditableBinding::new(
+            "notebookview:toggle_markdown_display_mode",
+            "Toggle Markdown rendered/raw view",
+            FileNotebookAction::ToggleMarkdownDisplayMode(MarkdownDisplayMode::Raw),
+        )
+        .with_context_predicate(id!("FileNotebookView"))
+        .with_key_binding("cmdorctrl-e"),
     ])
 }
 
@@ -1084,6 +1091,12 @@ impl TypedActionView for FileNotebookView {
                 self.context_menu.handle_action(action, ctx);
             }
             FileNotebookAction::ToggleMarkdownDisplayMode(mode) => {
+                // The rendered/raw toggle only applies to markdown files (its segmented
+                // control is only shown for them). The keybinding is registered view-wide,
+                // so ignore it on non-markdown panes.
+                if !self.is_markdown_file() {
+                    return;
+                }
                 self.markdown_display_mode = *mode;
                 self.display_mode_segmented_control
                     .update(ctx, |control, ctx| {
--- a/app/src/notebooks/file/mod.rs
+++ b/app/src/notebooks/file/mod.rs
@@ -1091,10 +1091,12 @@ impl TypedActionView for FileNotebookView {
                 self.context_menu.handle_action(action, ctx);
             }
             FileNotebookAction::ToggleMarkdownDisplayMode(mode) => {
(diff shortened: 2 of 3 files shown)


app/src/code/view.rs
```diff
@@ -2242,9 +2253,13 @@ impl TypedActionView for CodeView {
...
             #[cfg(feature = "local_fs")]
             CodeViewAction::RenderMarkdown => {
+                // The keybinding is registered for all code panes, but only markdown
+                // files have a rendered preview to switch to. Drop the path otherwise so
+                // the action is a no-op (the context-menu entry is already markdown-gated).
                 let lor_path = self
                     .tab_at(self.active_tab_index)
-                    .and_then(|t| t.location.clone());
+                    .and_then(|t| t.location.clone())
+                    .filter(|p| is_markdown_file(std::path::Path::new(&p.display_path())));
```
kevinyang372: We can not use Path::new here (this assumes local OS encoding but user could be opening a markdown file in a remote session)

We need to make sure the handler logic is session-agnostic here

kevinyang372 (changes requested):
Thanks for contributing! Have one question

maxmilian:
Good catch — fixed. The toggle now detects markdown via `location.path_component()` (the standardized, always-`/`-separated path) instead of `display_path()` + `Path::new`, so it no longer assumes local-OS path encoding and behaves correctly in a remote session. This mirrors how `left_panel.rs` and `file_tree/view.rs` already detect markdown for remote files — both feed `StandardizedPath::as_str()` to `is_markdown_file`.

For what it's worth, the same `display_path()` + `Path::new` pattern predates this PR in a few other spots (the `view.rs` segmented-control / context-menu fallback, and `FileNotebookView::is_markdown_file`). I've kept this PR focused on the toggle and left those alone, but I'm happy to make them session-agnostic in a follow-up if you'd like.

Verified: `cargo check -p warp --features local_fs`, `cargo clippy -p warp --features local_fs --tests -- -D warnings`, `cargo fmt -- --check`, and the markdown-detection tests all green.

app/src/code/view.rs
```diff
@@ -2242,9 +2253,20 @@ impl TypedActionView for CodeView {
...
+                // Detect markdown via the standardized path component (always
+                // `/`-separated) instead of display_path() + Path::new, which assumes
+                // local-OS encoding and would misparse a remote file's path on a
+                // cross-OS session.
                 let lor_path = self
                     .tab_at(self.active_tab_index)
-                    .and_then(|t| t.location.clone());
+                    .and_then(|t| t.location.clone())
+                    .filter(|p| {
+                        is_markdown_file(std::path::Path::new(p.path_component().as_str()))
```
kevinyang372: Is this fixed? We're still using Path::new here

kevinyang372 (changes requested):

maxmilian:
Good catch — you were right. In the earlier round I'd only swapped `display_path()` → `path_component()` but left the `Path::new` wrapper in, so the literal was still there. Pushed `881a202` removing it: `is_markdown_file` takes `impl AsRef<Path>` and only inspects the extension/file name, so the handler now passes the standardized `path_component()` `&str` directly — no `Path::new`. The toggle's markdown detection is session-agnostic end to end now.

maxmilian:
@kevinyang372 this is addressed now in `881a202` — the `Path::new` wrapper is gone; markdown is detected from `location.path_component()` (the standardized, always-`/`-separated path) fed straight into `is_markdown_file(..)`, so the handler no longer assumes local-OS path encoding and stays session-agnostic for remote files. Ready for another look whenever you have a moment 🙏

maxmilian:
Hi @kevinyang372 — gentle bump on this one 🙏 Your last review point (the `Path::new` wrapper assuming local-OS path encoding) was addressed in `881a202`: markdown is now detected straight from `location.path_component()` (the standardized, always-`/`-separated path) fed into `is_markdown_file(..)`, so the toggle stays session-agnostic for remote files. Happy to make any further tweaks — let me know whenever you have a moment to take another look.

maxmilian:
Hi @kevinyang372 — following up so this doesn't get auto-closed. Your `Path::new` concern (local-OS encoding leaking into remote sessions) is fully addressed as of `881a202`:

- The `Path::new(...)` wrapper is gone. Markdown detection now reads `location.path_component()` — the standardized, always-`/`-separated path component — and feeds it straight into `is_markdown_file(..)` (which takes `impl AsRef<Path>` and only inspects the extension). No local-OS path parsing, so the handler stays session-agnostic for remote files.

Happy to adjust further if there's anything still blocking — otherwise this should be ready for another look. Thanks! 🙏

maxmilian:
Hi @kevinyang372 — rebased onto latest `master` (was ~385 commits behind) to keep this from going stale. The only adaptation needed was qualifying the `is_markdown_file` import after the recent notebook-viewer refactor (`renders_in_warp_notebook_viewer`) — behavior is unchanged. Your earlier `Path::new` / local-OS encoding concern remains addressed via `path_component()` (previously `881a202`, now `33379ba` after the rebase) as described in my previous comment. `cargo check` is clean and the module tests pass. Ready for another look when you have a chance. 🙏

maxmilian:
Rebased onto latest `master` (`abea51c`) to clear the merge conflict.

The conflict came entirely from #13990 (*Move workspace crates over to the Rust 2024 edition*), which reordered the import block in `mod_tests.rs` to the 2024 rustfmt style. My side had added `FileNotebookAction` to the `super::{..}` import plus an `AuthManager` import that upstream now already carries a few lines below — so the resolution was to keep the `FileNotebookAction` addition and drop the duplicate. No logic changed.

Verified the rebase is behavior-preserving: the net diff against the new base is identical to the previous one apart from import ordering and hunk offsets. `rustfmt --edition 2024 --check` is clean on all three touched files.

@kevinyang372 — your earlier `Path::new` / local-OS encoding concern remains addressed via `path_component()` (now `a014aa1` after this rebase), as described above. Ready for another look whenever you have a chance. 🙏

maxmilian:
@kevinyang372 — sorry for the slow loop-back on your `Path::new` note; addressed now (commits `4683162e` + `a014aa1c`).

The toggle's markdown detection no longer goes through `display_path()` + `Path::new`, which was the local-OS-encoding assumption you flagged for remote sessions. It now inspects `path_component().as_str()` (the session-native, `/`-separated component) via `is_markdown_file(impl AsRef<Path>)`, so the check is session-agnostic for cross-OS remote files.

For completeness: there are two other `renders_in_warp_notebook_viewer(Path::new(&loc.display_path()))` call sites in `update_markdown_mode_segmented_control` / the notebook-viewer check, but those are pre-existing (from `13e8b61`, unrelated to this PR), so I left them untouched to keep this change surgical. Happy to file a follow-up if you'd like that same hardening applied there.

Could you take another look when you have a moment?

maxmilian:
Replying to keep this out of the auto-close path — but the review state here is stale rather than unaddressed, so I want to lay out the timeline plainly.

**@kevinyang372's two notes were about `Path::new`, and `Path::new` is gone.** Both requests landed 06-22/06-23; the fix went in on 07-20 (`4683162e` + `a014aa1c`, current head `a014aa1c`) and I summarized it here on 07-21.

Concretely, the markdown check no longer goes through `display_path()` + `Path::new` — that was the exact hazard you raised (local-OS path encoding, which misparses a remote file's path on a cross-OS session). It now feeds `is_markdown_file` the **standardized path component**, which is always `/`-separated:

```rust
.and_then(|t| t.location.clone())
.filter(|p| {
    crate::util::openable_file_type::is_markdown_file(
        p.path_component().as_str(),
    )
})
```

and `FileNotebookView` gates on `is_markdown_file()` rather than re-deriving anything from a path.

Both of your inline threads now show as **outdated** (the lines they point at no longer exist) but GitHub leaves them `unresolved`, and the `CHANGES_REQUESTED` review predates the head commit by ~4 weeks — which is why the stale bot reads this as "no activity from the author" even though the requested change shipped.

So the ask is just a state correction: could a reviewer dismiss the stale review (or re-review the current head)? Happy to rebase onto master first if that's easier for whoever picks it up. @kevinyang372 — no rush from my side, I only pinged because the bot is counting down to auto-close.

maxmilian:
Rebased onto latest `master` (`2b4a66f`) — the branch had fallen 309 commits behind. Clean rebase, diff unchanged (3 files, +110/−3), `rustfmt --edition 2024 --check` clean, and every upstream symbol the patch consumes still resolves on the new base.

I am not adding anything to the substance here — I have already laid out twice (07-21 and 07-30) that @kevinyang372's two `Path::new` requests were addressed on 07-20 in `4683162e` + `a014aa1c`. This is just a current head to review against, rather than one from June.

For what it is worth on the process side: I cannot re-request review from a fork, so the `CHANGES_REQUESTED` state stays visible even though the requests it refers to predate the current head by six weeks. If that state is what is keeping this in the stale-bot path, a re-review (or dismissing the stale one) would clear it — whichever is less work on your side.

acarl005:
@maxmilian FYI the intended flow is to re-request review from Oz after you've addressed the changes requested in the human review. I'll kick that off for you.

maxmilian:
@acarl005 thanks for kicking off the re-review — and Oz is right on both counts. Those were real regressions I introduced, not false positives, so I've fixed them rather than argued them. Pushed as `09156c7e6`.

**What was wrong.** I gated both new handlers on markdown-ness, but the actions they guard are shared with UI that is *not* markdown-only:

| handler | what actually gates the UI entry point | what I had gated on |
| --- | --- | --- |
| `FileNotebookAction::ToggleMarkdownDisplayMode` (`notebooks/file/mod.rs`) | `shows_markdown_toggle()` — markdown **or** `.ipynb` when `JupyterNotebookRendering` is on (`mod.rs:732`); it is what decides whether the header segmented control renders at all (`mod.rs:1200`) | `is_markdown_file()` |
| `CodeViewAction::RenderMarkdown` (`code/view.rs`) | `renders_in_warp_notebook_viewer()` — same markdown-or-`.ipynb` set; used by both the raw-side segmented control (`view.rs:288`) and the "View Markdown preview" context-menu entry (`view.rs:2154`) | `is_markdown_file()` |

So with the notebook feature flag on, an `.ipynb` pane would still *show* the Rendered/Raw control, but clicking it would have done nothing in either direction. That contradicts an invariant your own suite already pins — `test_load_jupyter_notebook_renders_cells` asserts `shows_markdown_toggle()` for a rendered notebook and cites "PRODUCT invariant 14".

**The fix** is to guard on the same predicate the UI uses, in both places, so the shared actions behave identically no matter which entry point dispatches them.

This does *not* walk back @kevinyang372's earlier request: the `CodeView` side still feeds the standardized `p.path_component().as_str()` rather than going through `display_path()` + `Path::new`, so it stays session-agnostic for remote files on a cross-OS session. Only the predicate changed, not the input to it.

**New regression test** — `test_markdown_toggle_switches_jupyter_notebook_pane_to_raw` opens a real `.ipynb`, asserts it is *not* a markdown file but *does* expose the toggle, then fires the shared action and requires it to reach raw mode.

Verified fail-before / pass-after: with the previous `is_markdown_file()` guard it fails (`assertion left == right failed ... left: Rendered, right: Raw`), and passes with the fix.

```
cargo check -p warp --features local_fs          # clean
rustfmt --edition 2024 --check                   # clean (3 files)
cargo test -p warp --features local_fs notebooks::file
                                                 # 11 passed
```

(`cargo-nextest` isn't installed on this machine at the moment, so I ran the same selection through `cargo test` rather than the `nextest` invocation listed in the PR description.)

Re-triggering Oz below so it reviews the fixed head.

maxmilian:
@kevinyang372 — short status note, not a request for a full re-read.

Your change request is recorded against `95aa5a08`. The `Path::new` concern it raised was fixed two days later in `881a202`: the check now takes `location.path_component().as_str()` — the standardized `/`-separated path — so it no longer assumes local-OS path semantics in a remote session. That fix survived the later rebases; the current head is `09156c7e` and its diff adds no `Path::new` call.

Oz re-reviewed `09156c7e` on Aug 4 with 0 critical / 0 important / 0 suggestions and requested human review.

The PR still shows CHANGES_REQUESTED only because the June review has not been dismissed. Dismissing it, or re-reviewing whenever convenient, would clear that. Happy to rebase if it has drifted since.

kevinyang372 (changes requested):
Rather than post-filtering on action trigger, this is better implemented by setting a specific keymap context so the keybinding is only triggerable when the markdown viewer is active.

kevinyang372:
Closing this PR as there have been too many failed attempts to get it over finish line

PR #12840, opened 2026-06-19
Invalidate cached local-file images in AI blocks when the file changes

zachlloyd:
## Description
Fixes a caching bug in inline image display in AI blocks (the PR #9993 feature): if an agent displays a local image, the file is then changed on disk, and the agent is asked to display the same path again, the second display showed the **stale, first-loaded image** instead of the updated file. This made it impossible to iterate on an image with the agent.

**Root cause:** `AssetSource::LocalFile` was keyed by path only. `AssetCache::load_asset()` reads a path once and never re-reads it, and the rendered-size `ImageCache` is keyed on the same source. There was no mtime/content check and no invalidation path for `LocalFile` (eviction only covered `AssetSource::Raw`), so a changed file at the same path kept hitting the cached decode.

**Fix:** include an optional content version in the local-file cache key.
- New `LocalFileContentVersion` (a `NonZeroU64` hash of last-modified time + file size) and a `content_version: Option<LocalFileContentVersion>` field on `AssetSource::LocalFile` (`crates/warpui_core/src/assets/asset_cache.rs`). Because it is part of the source, it flows into both the `AssetCache` (decoded bytes) and `ImageCache` (rendered) keys, so an edited file misses the cache and is re-read.
- `AssetSource::with_local_file_content_version()` reads the metadata; the AI block calls it inside the existing background link-detection task (`app/src/ai/blocklist/block.rs`), which populates the per-block `resolved_blocklist_image_sources` map.
- Stored as `NonZeroU64` so `Option<LocalFileContentVersion>` stays pointer-width (niche) and `AssetSource` does not grow (a `{u128,u64}` version tripped `clippy::large_enum_variant` on an unrelated enum via `WarpTheme`).

**Metadata is read only when the block resolves its image sources** (block creation / output completion / restore / shell-change) — never on the per-streaming-token path and never on render. Renders reuse the cached resolved source, and the render-time fallback stays stat-free (`content_version: None`). The stat also runs in the existing background `spawn_blocking` task, off the UI thread.

This is a general `AssetCache`/`ImageCache` improvement; non-blocklist callers (themes, notebooks) keep path-only caching (`content_version: None`), so their behavior is unchanged.

https://www.loom.com/share/c0eab28c73f146bb9ba531d3ba5b8081

## Linked Issue
warpdotdev/warp#12802

## Testing
- Added 3 focused unit tests in `crates/warpui_core/src/assets/asset_cache_tests.rs`:
  - a changed file (different size) produces a different cache key, while an unchanged file produces the same key
  - `for_path` returns `None` for a missing file
  - `with_local_file_content_version()` leaves non-local sources unchanged
- `cargo fmt -- --check`, `cargo check`, and `cargo clippy --tests -- -D warnings` are clean for the changed crates and the app (`warp` with `local_fs`); the new unit tests and the existing tests I touched pass.
- Could not run in this cloud sandbox: the full app test binary OOMs while **linking** (memory limit — it compiles fine), and the full wasm bundle build needs `clang` plus app-level feature wiring unavailable here. The wasm changes are cfg-gated stubs (`for_path` returns `None` on wasm). CI will run the full presubmit.

## Agent Mode

CHANGELOG-BUG-FIX: Re-displaying a local image in an AI response now reflects edits made to the file on disk instead of showing a stale cached copy.

---
_Conversation: https://staging.warp.dev/conversation/970500d6-1824-42c3-85a4-620b31258e19_
_Run: https://oz.staging.warp.dev/runs/019ed6bb-b701-78a8-8fa2-844979f25593_


crates/warpui_core/src/assets/asset_cache.rs
```diff
@@ -64,6 +69,51 @@ impl std::fmt::Debug for AsyncAssetId {
...
+impl LocalFileContentVersion {
+    /// Builds a content version by reading filesystem metadata for `path`.
+    ///
+    /// Performs blocking filesystem I/O, so this must only be called off the
+    /// render hot path (for example, once when a view resolves its image
+    /// sources), never on every frame. Returns `None` when metadata cannot be
+    /// read.
+    #[cfg(not(target_arch = "wasm32"))]
+    pub fn for_path(path: impl AsRef<std::path::Path>) -> Option<Self> {
+        let metadata = std::fs::metadata(path).ok()?;
```
kevinyang372, line 88: Some OS might not have modified time available. I think git uses a combination of modified time + file size as the fingerprint

crates/warpui_core/src/assets/asset_cache.rs
```diff
@@ -64,6 +69,51 @@ impl std::fmt::Debug for AsyncAssetId {
...
+///
+/// Used as part of an [`AssetSource::LocalFile`] cache key so that a file whose
+/// contents change on disk is treated as a distinct asset and re-read, rather
+/// than served from a now-stale cache entry. The fingerprint is a hash of the
+/// file's last-modified time and size, which are cheap to read and change
+/// whenever the file is rewritten. It is stored as a `NonZeroU64` so that
+/// `Option<LocalFileContentVersion>` stays pointer-width and does not grow the
+/// [`AssetSource`] enum.
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+pub struct LocalFileContentVersion(NonZeroU64);
```
kevinyang372: Why do we need to convert these into an int? I would keep the content version a plain structure with modified time + file size so it's explicit what it is

crates/warpui_core/src/assets/asset_cache.rs
```diff
@@ -85,11 +135,39 @@ pub enum AssetSource {
...
     },
     /// Accessible in the user's local filesystem at the provided path.
-    LocalFile { path: String },
+    LocalFile {
+        path: String,
+        /// Optional content fingerprint. When present, it makes the cache key
+        /// sensitive to on-disk changes so an edited file is re-read instead of
+        /// served stale. `None` preserves path-only caching for callers that do
+        /// not need invalidation.
+        content_version: Option<LocalFileContentVersion>,
```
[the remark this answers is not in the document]
kevinyang372: I don't think this is relevant. This is the purpose of adding content version is so that old image files don't get evicted

kevinyang372 (approved):

crates/warpui_core/src/assets/asset_cache.rs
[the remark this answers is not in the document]
zachlloyd: [Warp Agent] Added a 320 MB oldest-first cap for loaded versioned local-file assets. This preserves prior versions under normal use while preventing unbounded growth; when an old version is evicted, its corresponding rendered-image cache entries are also evicted. Fixed in https://github.com/warpdotdev/warp/commit/aa5c8407fe9e93509cd1e69699ed1430178dfc69

crates/warpui_core/src/assets/asset_cache.rs
zachlloyd: [Warp Agent] The content version continues to combine modified time with file size, while treating modified time as optional when the filesystem does not provide it. I also made those fields explicit in the content-version structure. Fixed in https://github.com/warpdotdev/warp/commit/aa5c8407fe9e93509cd1e69699ed1430178dfc69

crates/warpui_core/src/assets/asset_cache.rs
zachlloyd: [Warp Agent] Replaced the opaque hashed integer with an explicit content-version structure containing optional modified time and file size. Fixed in https://github.com/warpdotdev/warp/commit/aa5c8407fe9e93509cd1e69699ed1430178dfc69

PR #12869, opened 2026-06-20
Parallelize work done by the shell for the precmd hook.

vorporeal:
## Description

`precmd` previously collected and JSON-escaped each prompt-context field serially, adding the latency of every independent operation before the next prompt could render.

Bash and Zsh now start the independent virtualenv, Conda, Node version, Git, and Zsh `KUBECONFIG` work concurrently through process substitutions, then drain their outputs into caller-local variables with a shared `warp_read_parallel_outputs` helper. This avoids temporary files and remains compatible with Bash 3.2.

Fish cannot run shell functions concurrently, so it uses a Fish-specific approach: an uncached Node version pipeline starts before the grouped Git lookup, then independent JSON-escaping pipelines run concurrently and write into a private per-shell directory that is removed on exit. Fish falls back to serial collection if that directory cannot be created.

Across all three shells, Git head and branch stay grouped to avoid duplicate Git commands, while the Node eligibility directory walk and cache updates remain in the parent shell so cache state persists across prompts. Benchmarking the directory walk inside the Node worker showed no measurable benefit, so it remains serial. The MSYS2-specific Bash path is unchanged.

### Performance

Measured against the original serial prompt-context collection with all fields non-empty:

| Shell | Node cache | Serial | Parallel | Speedup | CPU overhead |
|---|---:|---:|---:|---:|---:|
| Bash 3.2 | Cold | 62.4 ms | 34.9 ms | 1.79x (44% lower latency) | +7% |
| Bash 3.2 | Warm | 32.1 ms | 23.0 ms | 1.39x (28% lower latency) | +13% |
| Zsh 5.9 | Cold | 55.7 ms | 29.7 ms | 1.88x (47% lower latency) | +12% |
| Zsh 5.9 | Warm | 29.1 ms | 17.3 ms | 1.68x (40% lower latency) | +14% |
| Fish 4.2 | Cold | 34.9 ms | 23.3 ms | 1.50x (33% lower latency) | +5% |
| Fish 4.2 | Warm | 11.2 ms | 7.5 ms | 1.48x (33% lower latency) | +1% |

A separate end-to-end Fish benchmark of the full `precmd` hook measured cold latency improving from 58.5 ms to 47.1 ms (1.24x / 19.5% lower, +4.6% CPU) and warm latency improving from 35.3 ms to 31.7 ms (1.11x / 10.2% lower, +1.0% CPU).

The collection benchmark ran from a package-containing checkout with a real `.git` directory, using Node v22.13.1 and Hyperfine. Each result is the per-hook mean from batches of 20 collections, with 5 warmups and 25–30 measured runs. The Fish full-hook benchmark used batches of 30 collections, with 3 warmups and 20 measured runs. CPU overhead is combined user and system time relative to the serial baseline.

## Linked Issue

N/A

## Testing

- `/bin/bash -n app/assets/bundled/bootstrap/bash_body.sh`
- `/bin/zsh -n app/assets/bundled/bootstrap/zsh_body.sh`
- `fish -n app/assets/bundled/bootstrap/fish.sh`
- Parsed generated/minified Bash, Zsh, and Fish bootstrap forms.
- Verified caller-local output assignment on Bash 3.2 and Zsh 5.9, including Git-value splitting with Zsh `KSH_ARRAYS` enabled.
- Verified byte-identical serial/parallel Fish `precmd` payloads for named and detached Git HEADs with cold and warm Node caches.
- Verified Fish's serial fallback and shared JSON escaping behavior.
- Ran the Hyperfine benchmarks documented above.

No new automated tests were added because this is performance-sensitive shell bootstrap behavior validated directly in all three target shells and through comparative benchmarks.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/assets/bundled/bootstrap/bash_body.sh
+++ b/app/assets/bundled/bootstrap/bash_body.sh
@@ -426,6 +426,19 @@ if [ -z "$WARP_BOOTSTRAPPED" ]; then
     warp_git () {
       GIT_OPTIONAL_LOCKS=0 command git "$@"
     }
+    # Reads process substitution outputs into caller-local variables. Passing all
+    # process substitutions in one function call starts their commands in parallel
+    # before this function begins reading.
+    warp_read_parallel_outputs () {
+      local output_var
+      local output_path
+      while (( $# > 0 )); do
+        output_var=$1
+        output_path=$2
+        IFS= read -r -d '' "$output_var" < "$output_path" || true
+        shift 2
+      done
+    }
 
     # Note that this is very performance sensitive code, so try not to
     # invoke any external commands in here.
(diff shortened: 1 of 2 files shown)


vorporeal:
* **#12869** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12869?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/12869?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #12872, opened 2026-06-21
fix(osc): guard params[1] access in OSC 1337 arm against missing second param

### LINKED ISSUE #12817: OSC 1337 without a second parameter panics (index out of bounds), crashing terminal output processing
### Describe the bug

Receiving an OSC `1337` (iTerm inline image protocol) escape **without a second parameter** triggers an `index out of bounds` panic in the OSC dispatcher. The triggering bytes are ordinary, untrusted PTY output, so any program — or `cat`-ing a crafted file — can crash the terminal's output-processing path. It is also reachable from a shared-session remote peer's output.

### To reproduce

```
printf '\e]1337\a'
```

(byte sequence `ESC ] 1337 BEL` = `\x1b]1337\x07`)

Unit-level repro against the ANSI parser:

```rust
let _ = parse_bytes(b"\x1b]1337\x07"); // panics during OSC dispatch
```

### Expected behavior

A malformed / parameterless OSC 1337 should be ignored (treated as unhandled), exactly like the neighboring OSC handlers do when a required parameter is missing.

### Actual behavior

Panic:

```
index out of bounds: the len is 1 but the index is 1
at app/src/terminal/model/ansi/mod.rs:1029
```

`app/src/terminal/model/ansi/mod.rs`, `osc_dispatch`, the `b"1337"` arm. The only length guard before the `match params[0]` (`params.is_empty() || params[0].is_empty()`) only guarantees `params[0]` exists, but the `1337` arm then unconditionally indexes `params[1]`. When the sequence has no second parameter, `params == [b"1337"]` (len 1), so `params[1]` is out of bounds. Every neighboring OSC arm (e.g. `133`, `777`, `52`) checks `params.len()` first; `1337` is the only one that does not. Adding a `params.len() < 2` guard (falling through to `unhandled`) fixes it.

### Environment

- Platform-independent (terminal OSC/VTE dispatch). Reproduced on macOS, warp `master`.
- Severity: a `panic` on untrusted terminal input (DoS). Reachable from local program output and from a shared-session remote peer.
- Regression: present as long as the unguarded `params[1]` access has existed in the `1337` arm.
- Only affects Warp.

I've traced the root cause and have a one-line guard plus a regression test ready locally — happy to open a PR if this is accepted.

adhavan18:
## Description

adds a `params.len() < 2` guard at the top of the `b"1337"` arm in `osc_dispatch` so a parameterless OSC 1337 sequence is treated as unhandled instead of panicking.

## Linked Issue

fixes #12817

## Root cause

`app/src/terminal/model/ansi/mod.rs`, `osc_dispatch`, the `b"1337"` arm directly accesses `params[1]` without a length guard. every neighboring OSC arm (`133`, `777`, `52`, `9277`, `9278`) checks `params.len()` before accessing any param beyond index 0 — the 1337 arm was the only exception.

when the sequence has no second parameter (`params == [b"1337"]`), `params[1]` is out of bounds → panic.

## Fix

three lines added at the top of the `b"1337"` arm:

```rust
if params.len() < 2 {
    return unhandled(params);
}
```

this is identical to the guard pattern already used by `b"133"`, `b"777"`, `b"52"`, and others.

## Testing

added a regression unit test in `mod_tests.rs`:

```rust
#[test]
fn parse_osc1337_without_second_param_does_not_panic() {
    let bytes: &[u8] = b"\x1b]1337\x07";
    let _ = parse_bytes(bytes); // must not panic
}
```

manual test: `printf '\e]1337\a'` in a running warp session no longer crashes.

## Agent Mode

CHANGELOG-BUG-FIX: fix panic when receiving a bare OSC 1337 sequence without a second parameter

### DIFF
--- a/app/src/terminal/model/ansi/mod.rs
+++ b/app/src/terminal/model/ansi/mod.rs
@@ -1026,6 +1026,9 @@ where
 
             // iTerm inline image protocol.
             b"1337" => {
+                if params.len() < 2 {
+                    return unhandled(params);
+                }
                 if params[1].starts_with(b"File=") {
                     let metadata = parse_iterm_image_metadata(params);
 
--- a/app/src/terminal/model/ansi/mod_tests.rs
+++ b/app/src/terminal/model/ansi/mod_tests.rs
@@ -1131,3 +1131,13 @@ fn parse_osc7_empty_payload_ignored() {
 
     assert!(handler.cwd_updates.is_empty());
 }
+
+#[test]
+fn parse_osc1337_without_second_param_does_not_panic() {
+    // Regression test for GH #12817: a bare `ESC ] 1337 BEL` (no second parameter)
+    // previously triggered an index-out-of-bounds panic because the 1337 arm
+    // accessed params[1] without a length guard. It must be silently ignored.
+    let bytes: &[u8] = b"\x1b]1337\x07";
+    // Must not panic.
+    let _ = parse_bytes(bytes);
+}


zachbai:
@adhavan18 please sign the CLA and re-run the CLA check with @cla-bot check when you have

zachbai:
Closing in favor of #12889

PR #12914, opened 2026-06-22
remove strong model handle captures in subscriptions

acarl005:
## Description

This PR is part of an effort to prevent future memory leaks. It is a follow-up to #12767 where we added a new param to `ModelContext::subscribe_to_model`. This PR actually utilizes that new param. This is to prevent scenarios _a. la._ #12355.

## Linked Issue

None. This is structural improvements to memory hygiene.


kevinchevalier (approved):
Great improvement, changing the signature of the callback!

PR #12926, opened 2026-06-22
Fix bug with cross-window drag + pinned tabs

peicodes:
https://www.loom.com/share/25b54d6318bb44529c72e25f7a415062

Fixes issue where dragging a tab out and back in with pinned tabs allows the drag to go above the pinned tabs.

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -27465,6 +27465,14 @@ impl Workspace {
                         self.calculate_updated_tab_index(current_index, position, ctx)
                     };
                     if new_index != current_index {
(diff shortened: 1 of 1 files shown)


johnturcoo (approved):

johnturcoo:
I think we also need to handle the case where a pinned tab is dragged out and dragged back in. It currently is showing the ghost tab in the pinned region, but then lands after the pinned items. This is because pinned items lose their pin state after they are cross window dragged.

I think we need to clamp past the pinned region in `ReorderInSource` so any tab that is dragged out and back in can not land in the pinned region, regardless of its original pin state.

Also, I assume the fix for the tab group bug is in a follow-up?

PR #12955, opened 2026-06-23
fix(terminal): DCH (CSI Pn P) must not erase cells before the cursor …

### LINKED ISSUE #12820: DCH (CSI Pn P) erases characters before the cursor when count exceeds cells to end-of-line
### Describe the bug

The DCH (Delete Character, `CSI Pn P`) handler erases characters **before** the cursor when `Pn` is larger than the number of cells remaining from the cursor to the end of the line. DCH must only delete the character under the cursor and those to its right (shifting the remainder left); cells before the cursor must never be touched.

### To reproduce

Deterministic repro (pure grid logic, 5-column row, ASCII only):

1. Write `abcde` (fills columns 0–4).
2. Move the cursor to column 3 (`CSI 4 G`).
3. Emit DCH with a count larger than the cells remaining: `ESC [ 10 P` (10 > the 2 cells from the cursor to end-of-line).

From a shell, output like `printf 'abcde\r\033[3C\033[10P'` exercises the same path. Any TUI / line editor that emits DCH with a large count near end-of-line can hit it.

Unit-level repro against the grid handler:

```rust
let mut grid = GridHandler::new_for_test(1, 5);
for c in "abcde".chars() { grid.input(c); }
grid.goto(VisibleRow(0), 3);
grid.delete_chars(10); // count (10) >> remaining (cols 5 - col 3 = 2)
assert_eq!(grid.grid_storage()[VisibleRow(0)][0].c, 'a'); // FAILS: got '\0'
```

### Expected behavior

Columns 0–2 keep `abc`; columns 3–4 become blank.

### Actual behavior

The **entire row** is cleared — `abc` (before the cursor) is destroyed:

```
  left: '\0'
 right: 'a'
```

`app/src/terminal/model/grid/ansi_handler.rs`, `fn delete_chars`:

- `let count = min(count, cols);` clamps to the full row width instead of the cells remaining from the cursor (`cols - cursor_col`).
- The trailing clear `let end = cols - count;` then falls **before** `start` when `count > cols - start`, so `row[end..]` wipes cells left of the cursor.

The sibling `insert_blank` already clamps correctly with `min(count, self.columns() - cursor.point.col)`. The fix is the same shape: `min(count, cols - start)`.

### Environment

- Platform-independent (pure terminal grid logic). Reproduced on macOS, warp `master`.
- Severity: screen content corruption / data loss (not a panic — the swap-loop indices stay in bounds).
- Regression: present as long as the current `delete_chars` clamp has been in place.
- Only affects Warp.

I've traced the root cause and have a minimal fix (`min(count, cols - start)`) plus a regression test ready locally — happy to open a PR if this is accepted.

maxmilian:
Closes #12820

GridHandler::delete_chars clamped the delete count to the full row width (min(count, cols)) instead of the cells remaining from the cursor. When the count exceeds those cells, the trailing clear (end = cols - count) falls before the cursor and wipes characters to its left -- full-line content loss. Clamp to min(count, cols - start), mirroring insert_blank's existing clamp.

Adds grid::tests regression coverage: large-count, in-bounds left-shift, and exact-remaining-count cases.

## Linked Issue

## Testing

### Screenshots / Videos

Recorded in a local OSS build (`./script/run`). The terminal prints 8 chars `PRESERVE`, leaving the cursor just past them, then emits `DCH` (`CSI 200 P`) with a count far larger than the cells remaining to end-of-line:

- **Before** (master): the trailing clear `end = cols - count` lands *left* of the cursor, so `PRESERVE` (text before the cursor) is wrongly erased and the line goes blank.
- **After** (this PR): clamping to `cols - start` keeps the clear at/after the cursor, so `PRESERVE` survives.

https://github.com/user-attachments/assets/4e0b4b65-916e-45e1-b9f4-685136c1349f

## Agent Mode

### DIFF
--- a/app/src/terminal/model/grid/ansi_handler.rs
+++ b/app/src/terminal/model/grid/ansi_handler.rs
@@ -679,10 +679,14 @@ impl ansi::Handler for GridHandler {
         let cursor = &self.grid.cursor();
         let bg = cursor.template.bg;
 
-        // Ensure deleting within terminal bounds.
-        let count = min(count, cols);
-
         let start = cursor.point.col;
+        // Clamp to the cells from the cursor to end-of-line, not the full row
+        // width. Otherwise `end = cols - count` (below) can fall *before*
+        // `start` when `count` exceeds the cells remaining, and the trailing
+        // clear then wipes cells before the cursor (#12820). Mirrors the clamp
+        // `insert_blank` already uses: `min(count, columns - cursor col)`.
+        let count = min(count, cols - start);
+
         let end = min(start + count, cols - 1);
         let num_cells = cols - end;
 
--- a/app/src/terminal/model/grid/tests.rs
+++ b/app/src/terminal/model/grid/tests.rs
@@ -1299,3 +1299,53 @@ fn wrap_cell(c: char) -> Cell {
     cell.flags.insert(Flags::WRAPLINE);
     cell
 }
+
+// DCH (CSI Pn P) with a count larger than the cells remaining from the cursor
+// to end-of-line must only blank the tail; it must never erase characters
+// before the cursor (#12820).
+#[test]
+fn delete_chars_large_count_keeps_text_before_cursor() {
+    let mut grid = GridHandler::new_for_test_with_scroll_limit(1, 5, 2);
+    grid.input_at_cursor("abcde");
+    grid.goto(VisibleRow(0), 3);
+    grid.delete_chars(10); // 10 >> the 2 cells from col 3 to end-of-line
+
+    let row = grid.row(0).unwrap();
+    assert_eq!(row[0], cell('a'));
+    assert_eq!(row[1], cell('b'));
+    assert_eq!(row[2], cell('c'));
+    assert_eq!(row[3], Cell::default());
+    assert_eq!(row[4], Cell::default());
+}
+
+// DCH within bounds shifts the trailing cells left and blanks the freed tail.
+#[test]
+fn delete_chars_shifts_trailing_cells_left() {
+    let mut grid = GridHandler::new_for_test_with_scroll_limit(1, 5, 2);
+    grid.input_at_cursor("abcde");
+    grid.goto(VisibleRow(0), 1);
+    grid.delete_chars(2); // delete 'b','c'; 'd','e' shift left
+
+    let row = grid.row(0).unwrap();
+    assert_eq!(row[0], cell('a'));
+    assert_eq!(row[1], cell('d'));
+    assert_eq!(row[2], cell('e'));
+    assert_eq!(row[3], Cell::default());
+    assert_eq!(row[4], Cell::default());
+}
+
+// DCH count exactly equal to the cells remaining clears them with no shift.
+#[test]
+fn delete_chars_count_equals_remaining() {
+    let mut grid = GridHandler::new_for_test_with_scroll_limit(1, 5, 2);
+    grid.input_at_cursor("abcde");
+    grid.goto(VisibleRow(0), 3);
+    grid.delete_chars(2); // delete 'd','e' (exactly the cells to end-of-line)
+
+    let row = grid.row(0).unwrap();
+    assert_eq!(row[0], cell('a'));
+    assert_eq!(row[1], cell('b'));
+    assert_eq!(row[2], cell('c'));
+    assert_eq!(row[3], Cell::default());
+    assert_eq!(row[4], Cell::default());
+}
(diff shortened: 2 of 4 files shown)


app/src/terminal/model/grid/ansi_handler.rs
```diff
@@ -670,10 +670,14 @@ impl ansi::Handler for GridHandler {
...
         let start = cursor.point.col;
+        // Clamp to the cells from the cursor to end-of-line, not the full row
+        // width. Otherwise `end = cols - count` (below) can fall *before*
+        // `start` when `count` exceeds the cells remaining, and the trailing
+        // clear then wipes cells before the cursor (#12820). Mirrors the clamp
+        // `insert_blank` already uses: `min(count, columns - cursor col)`.
+        let count = min(count, cols - start);
+
         let end = min(start + count, cols - 1);
         let num_cells = cols - end;
```
vorporeal, line 691: this also feels wrong to me - shouldn't `num_cells` be the difference between `start` and `end`?

maxmilian: Good question — I dug into this and I don't think `num_cells` is the count of deleted cells; it's the length of the **surviving tail that gets shifted left** into the gap, which is `cols - end`, not `end - start`.

The op runs in two phases:

1. **Shift left**: `for offset in 0..num_cells { row.swap(start + offset, end + offset) }` moves the cells in `[end, cols)` (everything after the deleted run) down to `[start, ..)`. The number of cells to move is exactly `cols - end` = `num_cells`.
2. **Blank the freed tail**: the later `let end = cols - count; row[end..]` clears the `count` cells at the end of the row.

So `end - start` (≈ `count`) is the number of *deleted* cells, handled in phase 2; `cols - end` is the number of *surviving* cells that slide left, which is what `num_cells` drives in phase 1. They're different quantities.

Quick trace, `cols=5`, row `abcde`, cursor at col 1, `delete_chars(2)`:
- `start=1`, `count=2`, `end=min(3,4)=3`, `num_cells=5-3=2`
- shift: `swap(1,3)`, `swap(2,4)` → `a d e _ _`
- blank tail: `end=cols-count=3`, clear `[3,5)` → `a d e _ _` ✓

That's the `delete_chars_shifts_trailing_cells_left` test. If `num_cells` were `end - start` the shift loop would move the wrong number of cells and corrupt the row.

Worth noting these two lines (`end`/`num_cells`) are pre-existing — this PR only tightens the `count` clamp from `min(count, cols)` to `min(count, cols - start)` so the trailing clear can't reach before the cursor (#12820). Happy to add a comment documenting the two-phase invariant if that'd help readability.

maxmilian:
Nudging this one — it's been quiet since 06-29 and I think it's waiting on a single yes/no from a reviewer.

**Recapping where the one open thread stands.** @vorporeal asked whether `num_cells` should be `end - start`. Two things worth separating:

1. **That line isn't part of this PR.** `let num_cells = cols - end;` is pre-existing — it's on `origin/master` today at `app/src/terminal/model/grid/ansi_handler.rs:687`. My diff to that file is 4 lines: replacing `min(count, cols)` with `min(count, cols - start)` plus the comment explaining why.

2. **And `cols - end` is correct for what it does.** It's the length of the *surviving tail that gets shifted left*, not the count of deleted cells:
   - phase 1, `for offset in 0..num_cells { row.swap(start + offset, end + offset) }` moves `[end, cols)` down to `[start, ..)` — that's `cols - end` cells
   - phase 2 blanks the freed tail via `row[cols - count..]` — that's where the `count` deleted cells are handled

   So `end - start` (≈ `count`) is the deleted count and it *is* used, just in phase 2. Naming it `num_cells` is admittedly opaque; happy to rename it to `tail_len` (or add a one-line comment) in a follow-up if you'd like, but I'd rather not touch unrelated pre-existing lines inside a regression fix.

**The actual change is the clamp.** Without it, `end = cols - count` can land *before* the cursor when `count` exceeds the cells remaining, and the trailing clear then wipes text to the left of the cursor — #12820. The clamp mirrors what `insert_blank` already does (`min(count, columns - cursor col)`), so DCH and ICH now agree. `tests.rs` covers the over-large count case red-then-green.

Nothing blocking on my side — if the `num_cells` naming is the only reservation, I'm happy to take it either way. Would just like to avoid it aging out.

maxmilian:
Rebased onto latest `master` (`2b4a66f`) — this branch had drifted furthest of my open PRs, 731 commits behind. Clean rebase, diff unchanged (2 files, +57/−3), `rustfmt --edition 2024 --check` clean, and all upstream symbols the patch consumes still resolve on the new base.

No change in substance; the DCH clamp and its regression test are exactly as reviewed. Flagging only so the head is current whenever this reaches the review queue.

PR #12960, opened 2026-06-23
feat: remove idle on failure to ensure setup consistency

dmichelin:
## Description
Now that we persist data in Redis we can reconstruct conversational data after environment failure. The client itself doesn't need to stay alive for this after sending a termination message, and it appears to be causing some strange behavior.

https://warpdev.slack.com/archives/C0B3LJBBT96/p1782233614823809

This removes the idle period after setup failure.

## Linked Issue
https://linear.app/warpdotdev/issue/REMOTE-2001/agent-start-flow-breaks-after-setup-command-failures-in-continued

## Testing

Tested locally by running a self hosted oz query with failed setup commands.

## Agent Mode

CHANGELOG-BUG-FIX: {{Fixed warp client idling after setup failure}}

### DIFF
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -928,10 +928,6 @@ impl AgentDriver {
                             );
                         })
                         .await;
-                    // Exit immediately after reporting the setup failure. Session data is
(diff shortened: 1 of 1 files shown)


ianhodge (approved):
nice, good find

PR #12966, opened 2026-06-23
fix: use status_error_message fallback to avoid misclassifying quota/credit errors as warp faults

dmichelin:
Loom
https://www.loom.com/share/7dcd2e3490634283a22d15ddfc296a69
## Description

When a cloud agent task fails due to quota or credit exhaustion, the client was sending two status updates in quick succession:

**Signal 1 — task polling**
The client polls the server every 3 seconds checking the task state. When the quota error kills the task, polling sees `AmbientAgentTaskState::Failed` → calls `handle_spawn_error` → sets `ConversationStatus::Error` on the local conversation. At this point the conversation's last exchange is still `Streaming` (the response from the LLM hasn't finished being processed client-side yet). `LocalAgentTaskSyncModel` handles the status change, finds no finished exchange error, and sends the generic "Agent encountered an error" to the server.

**Signal 2 — conversation stream** 
The actual quota error arrives through the streaming response channel. `mark_request_completed_with_error` is called with the structured `RenderableAIError::QuotaLimit`, which sets the exchange to `FinishedAIAgentOutput::Error`, then fires `UpdatedConversationStatus` again. This time `LocalAgentTaskSyncModel` finds the real error and sends the correct update.

So polling wins the race and sends the wrong update first, then the stream catches up and sends the right one. The fix just says: if the exchange is still streaming when the `Error` status fires, skip the update — the stream will send the only update once it's done.

**Fix:** In `on_conversation_status_updated`, when the conversation is in `Error` status but the last exchange is still `Streaming`, skip the update entirely. The stream will fire `UpdatedConversationStatus` again once `mark_request_completed_with_error` completes — at that point the exchange has the real structured error and the correct update is sent.

For **pre-spawn failures** (quota rejected before any stream starts), no exchange exists so the streaming check doesn't trigger. The `status_error_message` fallback surfaces the real error message instead of "Agent encountered an error".

## Linked Issue

Fixes REMOTE-1988

## Testing

Unit tests in `local_agent_task_sync_model_tests.rs` cover the error classification paths. The skip logic applies end-to-end when a cloud agent task fails while an active response stream is still in flight.

## Agent Mode

`CHANGELOG-BUG-FIX: Fixed quota and credit errors being misclassified as "Agent encountered an error" (warp fault) in cloud agent runs.`

### DIFF
--- a/app/src/ai/agent/conversation.rs
+++ b/app/src/ai/agent/conversation.rs
@@ -885,6 +885,13 @@ impl AIConversation {
         self.status = status;
     }
 
+    /// Test-only setter for the status error message, used to exercise
+    /// the `status_error_message` fallback in `map_conversation_status`.
+    #[cfg(test)]
+    pub(crate) fn set_status_error_message_for_test(&mut self, msg: Option<String>) {
+        self.status_error_message = msg;
+    }
+
     /// Test-only helper: appends an exchange to the root task so status-derivation
     /// logic (e.g. `map_conversation_status`) can be exercised end-to-end.
     #[cfg(test)]
--- a/app/src/ai/agent/conversation.rs
+++ b/app/src/ai/agent/conversation.rs
@@ -214,11 +214,6 @@ pub struct AIConversation {
     status: ConversationStatus,
     /// Optional detail for the current error status.
     status_error_message: Option<String>,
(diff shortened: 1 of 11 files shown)


seemeroland (approved):
Nice fix!

PR #12979, opened 2026-06-24
Make New File keybinding editable

### LINKED ISSUE #12681: `New File` action has no keyboard shortcut and isn't user-bindable
# `New File` action has no keyboard shortcut and isn't user-bindable

## The problem

The **File → New File** menu item opens a blank, file-backed code editor — effectively a scratchpad you can type into and save to disk. It's genuinely useful for jotting things down with far more room than the agent's rich input affords.

But unlike its siblings, **New File can't be triggered from the keyboard at all** — it has no default shortcut *and* it's the only "New …" item that doesn't appear in Settings → Keybindings, so you can't even assign one yourself:

| File menu item   | Default shortcut | Customizable in Settings? |
| ---------------- | ---------------- | ------------------------- |
| New Window       | ⌘N               | No                        |
| New Terminal Tab | ⌘T               | Yes                       |
| New Agent Tab    | —                | Yes                       |
| **New File**     | **—**            | **No**                    |

New Agent Tab shows the precedent: no default shortcut, but still user-bindable. New File is the only "New …" action with neither — so the only way to open the scratchpad is mouse → menu, every time.

## Who experiences it

Anyone who uses New File as a quick scratchpad/notes buffer (more room than rich input, and savable to a file). 
Using the mouse on every invocation breaks an otherwise keyboard-driven workflow.

## Current behavior and why it falls short

- New File is wired as a fixed binding with no default keystroke (`CustomAction::NewFile => None` in `app/src/util/bindings.rs`), and it's registered only as a `FixedBinding` (`app/src/workspace/mod.rs`), so it never surfaces in the editable keybindings settings.
- Its siblings (New Window, New Terminal Tab) get default shortcuts via the same mechanism, so the gap looks like an oversight rather than an intentional exclusion.

## Desired behavior

Be able to open New File from the keyboard. **Preferred:** make it a **user-customizable keybinding** — show "New File" in Settings → Keybindings (unassigned by default) so each user can pick a combo that doesn't collide with their setup. A sensible built-in default would be a bonus but isn't required.

(Equally happy if the team prefers a fixed default shortcut instead — flagging the customizable route to avoid having to pick a default, but obviously not a requirement.)

## Constraints / notes

- ⌘N and ⌘T are already taken in the File menu; New File and New Agent Tab are the two without defaults, so any default needs an un-taken combo and cross-platform care (the codebase already has `cmd_or_ctrl_shift` / PTY-compliance helpers for this).
- Scope is small and self-contained: the action and its handler already exist; this is purely about exposing a trigger.

SkyNotSilent:
## Description

Closes #12681.

This moves the existing **New File** action from a fixed-only binding to an editable keybinding so it appears in Settings -> Keybindings and can be assigned by the user. The binding remains unassigned by default, and it keeps the existing `CustomAction::NewFile` trigger so the File menu item and custom shortcut dispatch stay aligned.

Implementation notes:
- add a stable `workspace:new_file` binding name
- register New File via `EditableBinding` instead of `FixedBinding`
- keep the existing `WorkspaceAction::NewCodeFile` handler unchanged
- add a regression test that verifies New File is editable, has no default shortcut, and is still backed by `CustomAction::NewFile`

## Linked Issue

- [x] Screenshots/video are included below because this is visible in Settings -> Keybindings.

## Visual Evidence

Searching Settings -> Keyboard Shortcuts for `New File` now shows the entry, and selecting it opens the editable shortcut capture state.

New File keybinding editable

## Testing

- [x] `cargo fmt -p warp`
- [x] `cargo test -p warp util::bindings::tests --lib`
- [x] `git diff --check`
- [x] Manually tested with `./script/run` by opening Settings -> Keyboard Shortcuts, searching `New File`, and selecting the row to verify the editable shortcut capture state.

## Agent Mode

CHANGELOG-IMPROVEMENT: New File can now be assigned a custom keyboard shortcut from Settings -> Keybindings.

### DIFF
--- a/app/src/workspace/mod.rs
+++ b/app/src/workspace/mod.rs
@@ -67,8 +67,8 @@ pub use toast_stack::ToastStack;
 use crate::workspace::view::{
     LEFT_PANEL_AGENT_CONVERSATIONS_BINDING_NAME, LEFT_PANEL_GLOBAL_SEARCH_BINDING_NAME,
     LEFT_PANEL_PROJECT_EXPLORER_BINDING_NAME, LEFT_PANEL_WARP_DRIVE_BINDING_NAME,
-    NEW_AGENT_TAB_BINDING_NAME, NEW_AMBIENT_AGENT_TAB_BINDING_NAME, NEW_TAB_BINDING_NAME,
-    NEW_TERMINAL_TAB_BINDING_NAME, OPEN_GLOBAL_SEARCH_BINDING_NAME,
+    NEW_AGENT_TAB_BINDING_NAME, NEW_AMBIENT_AGENT_TAB_BINDING_NAME, NEW_FILE_BINDING_NAME,
+    NEW_TAB_BINDING_NAME, NEW_TERMINAL_TAB_BINDING_NAME, OPEN_GLOBAL_SEARCH_BINDING_NAME,
     TOGGLE_CONVERSATION_LIST_VIEW_BINDING_NAME, TOGGLE_NOTIFICATION_MAILBOX_BINDING_NAME,
     TOGGLE_PROJECT_EXPLORER_BINDING_NAME, TOGGLE_RIGHT_PANEL_BINDING_NAME,
     TOGGLE_TAB_CONFIGS_MENU_BINDING_NAME, TOGGLE_VERTICAL_TABS_PANEL_BINDING_NAME,
@@ -316,13 +316,15 @@ pub fn init(app: &mut AppContext) {
             id!("Workspace"),
         )
         .with_enabled(|| ContextFlag::CreateNewSession.is_enabled()),
-        FixedBinding::custom(
-            CustomAction::NewFile,
-            WorkspaceAction::NewCodeFile,
-            "New File",
-            id!("Workspace") & !id!("Workspace_ViewOnlySharedSession"),
-        ),
     ]);
+
+    app.register_editable_bindings([EditableBinding::new(
+        NEW_FILE_BINDING_NAME,
+        BindingDescription::new("New File"),
+        WorkspaceAction::NewCodeFile,
+    )
+    .with_custom_action(CustomAction::NewFile)
+    .with_context_predicate(id!("Workspace") & !id!("Workspace_ViewOnlySharedSession"))]);
 
     if FeatureFlag::UIZoom.is_enabled() {
         app.register_fixed_bindings([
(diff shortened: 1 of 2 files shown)


SkyNotSilent:
Visual evidence for the Settings -> Keyboard Shortcuts path: searching for `New File` now shows the row, and selecting it opens the editable shortcut capture state with `Default`, `Cancel`, and `Save`.

New File keybinding editable

app/src/util/bindings_tests.rs
```diff
@@ -73,6 +76,31 @@ fn test_keybinding_name_to_display_string() {
...
+                &Trigger::Custom(CustomAction::NewFile.into())
+            );
+            assert_eq!(binding.original_trigger, None);
+            assert_eq!(
+                keybinding_name_to_display_string(NEW_FILE_BINDING_NAME, ctx),
+                None
+            );
+        });
+    });
+}
```
acarl005: I appreciate the desire to have test coverage. We already have tests for editable and fixed bindings so we don't need tests for each individual instantiation. Please delete this.

app/src/workspace/mod.rs
```diff
@@ -1391,6 +1387,16 @@ pub fn init(app: &mut AppContext) {
...
 
+pub(crate) fn register_new_file_binding(app: &mut AppContext) {
+    app.register_editable_bindings([EditableBinding::new(
+        NEW_FILE_BINDING_NAME,
+        BindingDescription::new("New File"),
+        WorkspaceAction::NewCodeFile,
+    )
+    .with_custom_action(CustomAction::NewFile)
+    .with_context_predicate(id!("Workspace") & !id!("Workspace_ViewOnlySharedSession"))]);
+}
```
acarl005: Since this is just one binding you can inline this function into `fn init`

acarl005 (changes requested):
Thanks for updating this. Just some nits to change and then afterwards I'll approve.

SkyNotSilent:
Addressed the two review nits from @acarl005:

- Removed the per-binding regression test from `app/src/util/bindings_tests.rs` since editable/fixed binding behavior already has coverage.
- Inlined the New File editable binding registration into `workspace::init` and removed the single-use helper.

Validation:
- `./script/format --check`
- `git diff --check`
- `cargo test -p warp test_keybinding_name_to_display_string --lib`
- `cargo test -p warp test_orchestration_cycle_bindings_are_editable --lib`
- `cargo clippy -p warp --lib --tests -- -D warnings`

/oz-review

acarl005 (approved):

PR #13002, opened 2026-06-24
Fix crdt input buffer inconsistencies with agent prompts

seemeroland:
## Description

The input buffer ID is determined by the active block ID. When making a bunch of agent prompts, the active block stays the same, so we keep the same input buffer ID even though submitting agent prompts clears the input.

When a viewer submitted a prompt, they would see a frozen input UI (this created a temporary buffer for this frozen input) until the server sends back a "prompt in flight" ack. On this ack, the viewer would unfreeze and clear the input.

However, this clear was done by reinitializing the buffer, which replaces it completely and does not emit a CRDT operation. The sharer on receiving the prompt request would also call unfreeze and clear input, which does not emit a CRDT operation. As a result a new viewer joining would still see the input populated, because there was no CRDT input update to clear the buffer.

The takeaway is that changes to the input buffer must by synchronized through crdt operations. The viewer now only unfreezes the input on the "prompt in flight" ack. When the sharer executes the prompt, it will ultimately go through `system_clear_buffer`, which will emit a crdt operation, and all viewers will apply that. This alone fixes the issue, but the viewer who sent the prompt would see the input become unfrozen and then get cleared a moment later. To smooth out this UI, the viewer sending the prompt optimistically clears the input by displaying an ephemeral empty buffer. When a crdt operation comes in (the clear), we replace the ephemeral buffer with the real one with the crdt operation applied, so that new operations can be applied.

### Alternative considered

Another way to fix the issue would have been to create a new buffer ID on every agent prompt submitted (similar to on every shell command submitted). But we don't have an existing mechanism to synchronize this buffer ID, which would be more complicated to add. Block IDs are automatically synchronized because they come from pty bytes that create the block.

## Testing

### Screenshots / Videos

https://www.loom.com/share/cf6461aacabb4075a4e804f11b96c876

## Agent Mode


app/src/editor/view/model/mod.rs
```diff
@@ -682,11 +727,20 @@ impl EditorModel {
...
-            let snapshot = self.as_snapshot(ctx);
-            self.buffer_and_display_map.deactivate_ephemeral_state();
-            Some(snapshot)
+            if self.buffer_and_display_map.ephemeral_is_display_only {
+                // Display-only ephemeral: discard the ephemeral content entirely and
+                // proceed directly on the regular buffer. Do NOT snapshot-and-restore,
+                // which would generate spurious CRDT delete ops for whatever the regular
+                // buffer currently contains (e.g. another viewer's concurrent edits).
+                self.buffer_and_display_map.deactivate_ephemeral_state();
+                None
```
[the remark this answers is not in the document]
seemeroland: I tested viewer typing immediately after sending a query. It's fine - there's a brief moment the inputs are appended to the prompt, but the correct prompt is sent, and the inputs end up in the new buffer (and are consistent across partaicipants)

app/src/editor/view/mod.rs
```diff
@@ -3235,6 +3235,23 @@ impl EditorView {
         ctx.emit(Event::BufferReinitialized);
     }
 
+    /// Exits the ephemeral loading state created by `set_buffer_text_ignoring_undo`
+    /// without touching the CRDT buffer or emitting any `UpdatePeers` operations.
+    /// The editor switches back to displaying the regular collaborative buffer.
+    pub fn exit_ephemeral_loading_state(&mut self, ctx: &mut ViewContext<Self>) {
```
harryalbert, line 3241: IIUC, we don't really have an 'loading' state anymore right (or at least a visual loading state)? I guess the ephemeral buffer is the loading state?

app/src/editor/view/model/mod.rs
```diff
@@ -388,6 +398,15 @@ impl BufferAndDisplayMaps {
...
         );
         self.ephemeral = Some((ephemeral_buffer, ephemeral_display_map));
+        self.ephemeral_is_display_only = false;
+    }
+
+    /// Activates a display-only ephemeral state. When this ephemeral is materialized
+    /// (exited by a non-ephemeral edit), its content is discarded rather than applied
+    /// to the regular buffer, preventing spurious CRDT operations.
+    fn activate_display_only_ephemeral_state(&mut self, ctx: &mut ModelContext<EditorModel>) {
+        self.activate_new_ephemeral_state(ctx);
```
harryalbert, line 408: what happens if a viewer submits a prompt while in the ephemeral state?

app/src/terminal/input.rs
```diff
@@ -7366,22 +7366,54 @@ impl Input {
...
+    /// (materializing the ephemeral), its empty content is **discarded** — no delete ops
+    /// are generated for the regular buffer's contents. The edit proceeds directly on
+    /// the regular buffer (which the sharer's delete ops will have cleared by then).
+    ///
+    /// When `optimistically_show_empty` is false (used for `AgentPromptRequestFailed`),
+    /// the regular buffer's original prompt text is shown instead, allowing the viewer
+    /// to retry.
+    pub fn unfreeze_agent_input(
+        &mut self,
+        optimistically_show_empty: bool,
```
harryalbert: could you maybe add a comment on this field or rename it? I read the comment above, but I'm still not entirely grokking what this field is for

harryalbert (approved):
Great find

app/src/editor/view/mod.rs
seemeroland: Yea we do still have the UI of making the input gray and adding a spinner between viewer submitting prompt and receiving the ack - that's the ephemeral buffer for loading state

app/src/editor/view/model/mod.rs
seemeroland: I tried and wasn't able to

app/src/terminal/input.rs
seemeroland: updated

PR #13016, opened 2026-06-24
Fix silently-dropped Oz viewer prompts after network disconnect

peicodes:
## Description
**Bug:** On an Oz (cloud) shared session, follow-up agent prompts were silently dropped after a network disconnect (laptop sleep, Wi-Fi change, NAT rebind) — no freeze, no error, the prompt just never reached the agent.

**Why:** The viewer's shared-session `Network` only reconnected when the websocket stream itself ended. If the TCP connection died silently (no FIN), nothing detected it: the session stayed marked as connected and follow-ups were written into a dead socket. Unlike `CloudObjects::Listener`, the viewer had no proactive liveness detection.

**Fix:** Mirror the `CloudObjects::Listener` pattern. The viewer now subscribes to OS connectivity signals (`SystemStats` sleep/wake, `NetworkStatus` online/offline): it tears down the live socket when the connection is about to drop, and proactively reconnects when it likely came back. A supporting change makes socket teardown deterministic so a dead/hung socket is replaced reliably (and a stale socket can't tear down a healthy replacement).

**Out of scope:** No message queuing/replay — a proactive reconnect keeps the dead window short enough to be acceptable. Sharer-side subscriptions are a possible follow-up.

## Linked Issue

## Testing
Added unit tests covering CPU-wake and offline→online reconnect triggers while joined. `format --check`, `clippy`, and `cargo test -p warp shared_session::viewer::network` all pass.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed shared session (Oz) viewer follow-up prompts being silently dropped after a network disconnection or laptop sleep by proactively detecting the dead connection and reconnecting.

_Conversation: https://app.warp.dev/conversation/e990098e-326b-4dd2-90fe-fef95e54a62f_
_Run: https://oz.warp.dev/runs/019efb3d-820e-791e-8945-c54fb43380da_

### DIFF
--- a/app/src/terminal/shared_session/viewer/network.rs
+++ b/app/src/terminal/shared_session/viewer/network.rs
@@ -8,7 +8,7 @@ use std::time::Duration;
 
 use anyhow::bail;
 use async_channel::Receiver;
-use futures_util::stream::AbortHandle;
+use futures_util::stream::{abortable, AbortHandle};
 use futures_util::{SinkExt, StreamExt};
 use instant::Instant;
 use parking_lot::FairMutex;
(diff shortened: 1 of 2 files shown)


peicodes:
Hmmm...I'm not totally convinced this is the right fix. I just tested the current version of warp by disconnecting my wifi and letting my laptop go to sleep, and I was able to send the agent follow ups

peicodes:
I believe I've found a more robust solution that covers a wider scope.

PR #13023, opened 2026-06-24
[QUALITY-884] Hide custom models from inline model picker in cloud agent panes

JaidenRatti:
## Description

Custom endpoint models (e.g. locally-configured custom LLM endpoints) were incorrectly appearing in the inline `/MODEL` picker when the user was in a cloud agent pane. Custom models cannot route through Warp's cloud inference infrastructure, so they should not be selectable there.

The root cause: `ModelSelectorDataSource::run_query` — which powers the inline `/MODEL` tabbed picker — had no awareness of whether its pane was an Oz cloud agent pane. It fetched all models including custom endpoints unconditionally.

**Fix:** Thread `ambient_agent_view_model` through `InlineModelSelectorView` and into `ModelSelectorDataSource`. In `run_query`, derive `is_cloud_pane` from whether the model handle is `Some`, and filter out any model where `custom_llm_info_for_id` returns `Some` when in a cloud pane. 

### Screenshots / Videos

Loom showing the fix in action: https://www.loom.com/share/2317394c09a4468b956da64a6dac7217

## Linked Issue

Loom showing the buggy behaviour is in here as well --> https://linear.app/warpdotdev/issue/QUALITY-884/custom-models-are-selectable-for-cloud-agent-requests-when-changing

## Testing

## Agent Mode

### DIFF
--- a/app/src/terminal/input.rs
+++ b/app/src/terminal/input.rs
@@ -3399,6 +3399,7 @@ impl Input {
         let inline_model_selector_view = ctx.add_view(|ctx| {
             InlineModelSelectorView::new(
                 terminal_view_id,
+                ambient_agent_view_model.clone(),
                 suggestions_mode_model.clone(),
                 agent_view_controller.clone(),
                 &buffer_model,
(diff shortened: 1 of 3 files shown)


danielpeng2 (approved):
thanks for fixing!

PR #13034, opened 2026-06-25
Add Markdown syntax highlighting in the editor

### LINKED ISSUE #9042: Markdown syntax highlighting
### Pre-submit Checks

Markdown should have syntax highlighting while in viewing mode in warp.

_No response_

### Additional context

_No response_

### Operating system (OS)

macOS

4

None

bradleyjames:
## Description

Adds Markdown to the languages the editor syntax-highlights. Opening a `.md` / `.markdown` file in Warp's code editor (Raw view) now highlights block-level structure instead of rendering as plain text.

Markdown is highlighted with arborium's bundled Markdown query — the same source every other language uses — plus a small extension to Warp's shared capture-name → color mapper so it understands the nvim-style capture names arborium's Markdown query emits.

### What's supported

Block-level Markdown constructs:
- Headings — ATX (`#`–`######`) and setext, markers and text
- Fenced and indented code blocks — fence delimiters and the code body
- Lists — bullets (`-` `*` `+`) and ordered (`1.` `3)`)
- Blockquotes (including multi-line continuation) and thematic breaks
- Link **reference definitions** — `[label]: url "title"`
- Backslash escapes

### What's not supported yet

**Awaiting upstream arborium query rules** — the grammar produces these nodes, but arborium's bundled Markdown query has no rule for them, so they render in the default color (to be added in arborium):
- YAML / TOML front matter
- Task markers `[x]` / `[ ]`
- Pipe tables
- HTML blocks
- Entity / numeric character references (`&amp;`)

**Inline** (styling *within* a line of text) — needs the separate Markdown *inline* grammar, which arborium does not yet package upstream:
- Emphasis — `**bold**`, `*italic*`, `~~strikethrough~~`
- `` `inline code` ``
- In-text links — `[title](url)`, autolinks `<url>`, reference *uses* `[text][label]`

## Implementation

- Enable arborium's `lang-markdown` grammar and register `markdown` in the `languages` crate (`.md` / `.markdown` extensions, `md` alias).
- Highlight Markdown with arborium's bundled Markdown query (`arborium::lang_markdown::HIGHLIGHTS_QUERY`), the same path every other language uses.
- Extend Warp's shared capture-name → color mapper (`convert_capture_name_to_color`) to resolve arborium's nvim-style capture names — `text.title`, `text.literal`, `text.uri`, `text.reference`, and `punctuation.*` — which previously fell through to no color. The mapper is shared across languages, so any grammar whose query uses these names now colors them (e.g. punctuation in Rust); the impact is minor and additive (see the cross-language screenshots below).
- Sync `ProgrammingLanguage::to_extension` so fenced ` ```markdown ` blocks in agent output highlight through the same path.

## Linked Issue

Closes #9042.

## Testing

- **Unit:** `crates/languages` — `markdown_extensions_resolve_to_markdown` and `all_supported_languages_load_successfully` (loads Markdown via arborium's query); `app` — `to_extension` covers the `markdown`/`md` fence aliases. Full presubmit (clippy + nextest) is green apart from pre-existing environment-only SSH integration tests.
- **Manual:** verified highlighting in the editor's Raw view against Markdown fixtures and representative Rust/Python files.

### Screenshots / Videos

Captured in the editor's **Raw** view (dark theme).

#### Markdown

**1. Headings** — `01-headings.md`

01-headings

**2. Lists, quotes & rules** — `02-lists-quotes-rules.md`

02-list-quotes

**3. Code blocks** — `03-code.md`

03-code

**4. Links & reference definitions** — `04-links-table-html-inline.md`

04-inline-tables

#### Cross-language impact of the shared mapper

Extending `convert_capture_name_to_color` colors `text.*` / `punctuation.*` in any language whose query uses those names. Before / after on a representative file:

**Python — before**

python-before

**Python — after**

python-after

**Rust — before**

rust-before

**Rust — after**

rust-after

## Agent Mode

---

CHANGELOG-IMPROVEMENT: Markdown files now have syntax highlighting in the editor.


crates/languages/grammars/markdown/highlights.scm
```diff
@@ -0,0 +1,72 @@
...
+;
+; arborium bundles an upstream Markdown highlights query that uses nvim-style capture
+; names (e.g. `text.title`, `punctuation.special`). Warp only maps a fixed set of capture
+; prefixes to theme colors (see `convert_capture_name_to_color` in crates/syntax_tree), so
+; this query re-targets the same grammar nodes onto the names Warp understands: keyword,
+; function, string, type, comment, property, and tag.
+;
+; This is the block-level grammar only. Inline emphasis and language-specific highlighting
+; inside fenced code blocks require tree-sitter injections, which the editor's highlighter
+; does not yet support.
```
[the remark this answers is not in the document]
bradleyjames: fixed

crates/syntax_tree/src/queries/highlight_query.rs
```diff
@@ -92,6 +93,110 @@ fn convert_capture_name_to_color(name: &str, color_map: &ColorMap) -> Option<Col
...
+
+        let source = source.get_or_insert_with(|| buffer.text().into_string());
+        let Ok(language_name) = language_node.utf8_text(source.as_bytes()) else {
+            continue;
+        };
+        let Some(language) = language_by_name(&language_name.trim().to_ascii_lowercase()) else {
+            continue;
+        };
+
+        let Ok(content) = content_node.utf8_text(source.as_bytes()) else {
```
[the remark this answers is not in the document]
bradleyjames: fixed

crates/languages/grammars/markdown/highlights.scm
kevinyang372, line 1: Why aren't we using arborium's packaged in highlights query?

crates/languages/src/lib.rs
```diff
@@ -312,17 +327,33 @@ fn load_language(lang: &str) -> Option<Language> {
         })
         .collect();
 
-    // Use arborium's bundled highlight query instead of loading from custom .scm files
-    let highlight_query_str = get_arborium_highlight_query(lang)?;
-    let highlight_query = Query::new(&grammar, highlight_query_str)
-        .expect("arborium highlight query should be valid");
+    // Prefer a Warp-authored highlight query bundled alongside the grammar, falling back to
```
kevinyang372: See comments above, I don't think we need to do this

What are the capture names that we don't recognize? We should consider supporting these capture names rather than rewriting a query from scratch

crates/syntax_tree/src/lib.rs
```diff
@@ -51,7 +51,7 @@ struct HighlightCache {
 struct HighlightCacheKey {
     version: BufferVersion,
     ranges: RangeSet<CharOffset>,
-    language_id: Option<arborium::tree_sitter::Language>,
+    language: Option<arborium::tree_sitter::Language>,
```
kevinyang372: Why the refactor here?

crates/syntax_tree/src/queries/highlight_query.rs
```diff
@@ -92,6 +93,110 @@ fn convert_capture_name_to_color(name: &str, color_map: &ColorMap) -> Option<Col
...
 }
 
+/// Highlight colors contributed by language injections (e.g. Markdown fenced code blocks).
+///
+/// For each injection in `tree` whose content overlaps `range`, the embedded language named by the
+/// fence is resolved, its content is parsed and highlighted with its own grammar, and the resulting
+/// colors are returned in the parent buffer's character coordinates. Injections that set a static
+/// language without a captured language node (e.g. HTML blocks) are skipped here — those stay
+/// colored by the host grammar's own highlight query.
+pub fn injected_highlights(
```
kevinyang372: Language injection is a whole different scope of a project imo (we need to think about performance / optimization for this + potential support for other languages)

I would separate this out into a different PR rather than mixing it with markdown support

kevinyang372 (commented):
Good progress here! Have a few feedback on this

crates/languages/src/lib.rs
bradleyjames: The reason I didn't touch `convert_capture_name_to_color(...)` is because the markdown highlight queries are a different structure.  It's your call but the root of the markdown queries are: `@text.<keyword>` and `@punctuation.<keyword>` where the existing languages supported by warp use `@keyword`, `@function`, etc.  The only query that worked out of the box was `@string.escape`.

Markdown highlights in arborium: https://github.com/bearcove/arborium/blob/c2121b480bd7a8ab537d1f2e10433614405b3a54/langs/group-willow/markdown/def/queries/highlights.scm

Your call, let me know.

My opinion is this a non standard case out of ~30 cases.  If it becomes standard it can be refactored into the core function but until then a separate file is satisfactory and doesn't muddy the core logic.

crates/syntax_tree/src/lib.rs
bradleyjames: mistake, good catch.

crates/syntax_tree/src/queries/highlight_query.rs
bradleyjames: I'll kill it for now and get back to it one day.

crates/languages/grammars/markdown/highlights.scm
bradleyjames: answered below

dor-tzur-lmnd:
This really needs to be fast tracked. Most anyone is doing this days is reading markdown files. I would have expected Warp team itself to ship this, it's an adoption blocker

crates/languages/src/lib.rs
kevinyang372: I would prefer us supporting `@text.*` and `@punctuation.*` keywords in our color mapping since this is a common convention across nvim-style names

specs/gh-9042/03-code.md
```diff
@@ -0,0 +1,20 @@
+# Code blocks
```
kevinyang372: What are these md files for?

bradleyjames: The md files were my test files.  Just needed to put them somewhere.

crates/languages/src/lib.rs
bradleyjames: Done.  I updated the screenshots in the description and added before and after for python and rust as this caused a slight change in those grammars with the addition of`@punctuation`.

specs/gh-9042/03-code.md
bradleyjames: removed from git

bradleyjames:
FYI, created https://github.com/bearcove/arborium/pull/208 to style more markdown blocks.

kevinyang372 (approved):

PR #13044, opened 2026-06-25
[VA] Add computer-use recording substrate

vkodithala:
## Description

This is the first PR in the computer-use video-recording stack.

Broadly, it adds the shared recording substrate in `crates/computer_use`: a recorder abstraction, recording metadata/configuration types, Linux X11 capture through supervised `ffmpeg`, and no-op recorder stubs for unsupported platforms.

Cloud computer-use sessions need a bounded client-side recording primitive before agent actions can start/stop recordings or upload artifacts. Wanted to keep this substrate seperate so that the recorder API, Linux process supervision, and resource limits could be reviewed independently, which should help some.

Tl;dr on the work done here: the Linux implementation shells out to `ffmpeg`/`x11grab`, applies duration and size bounds at the recorder layer, reports typed recording errors, and returns metadata through `RecordingHandle` so app-layer callers can surface the applied capture parameters. Unsupported platforms return a typed unsupported-platform error rather than pretending recording succeeded (recording utils are only implemented for Linux).

## Stack context

- This PR is based on `master` and intentionally stops at the recording substrate.
- #13227 (`varoon/va-video-recording-execution`) is stacked on top and wires this substrate into agent StartRecording/StopRecording execution, artifact upload, and lifecycle handling.
- Review this PR for the recorder API shape, Linux `ffmpeg` behavior, platform boundaries, and resource-limit enforcement; review #13227 for agent/API/action-model integration.

## Linked Issue

N/A

## Testing

Here's a [demo](https://www.loom.com/share/59ea36e89f9b42479a48d9ca0d1900b2) of this all working properly end-to-end!

- `./script/format`
- `cargo check -p computer_use`
- `cargo clippy -p computer_use --all-targets -- -D warnings`
- [x] I have manually tested my changes locally with `./script/run` (not applicable to this substrate-only PR; Linux X11 capture needs the computer-use environment)

### Screenshots / Videos

N/A; this PR has no user-visible UI.

## Agent Mode


crates/computer_use/src/lib.rs
```diff
@@ -57,6 +59,103 @@ pub trait Actor: Send + Sync + 'static {
...
+pub struct RecordingConfig {
+    /// Capture frame rate in frames per second.
+    pub frame_rate: u32,
+}
+
+impl Default for RecordingConfig {
+    fn default() -> Self {
+        Self { frame_rate: 15 }
+    }
+}
```
vkodithala, line 239: nit: We should probably add some context here around the choice of default frame rate. Why is 15 fps a good candidate? This'll probably be tweaked a bit while dogfooding, but justifying this decision might be nice.

crates/computer_use/src/linux/recording.rs
```diff
@@ -0,0 +1,196 @@
...
+            .map_err(|e| format!("Failed to spawn ffmpeg for recording: {e}"))?;
+
+        // Resolve once capture is confirmed live (the output file has grown,
+        // meaning ffmpeg opened the display and the muxer is writing).
+        if let Err(e) = wait_for_first_output(&path, &mut process).await {
+            let _ = process.start_kill();
+            let detail = ffmpeg_error_tail(&std::fs::read_to_string(&log_path).unwrap_or_default());
+            let _ = std::fs::remove_file(&path);
+            let _ = std::fs::remove_file(&log_path);
+            return Err(format!("ffmpeg failed to start recording: {e}{detail}"));
```
vkodithala: nit: I wonder if we should just define an error enum that classifies all of these states, for clarity? i.e. an error state for `FailedToStart`, or `InvalidDimensions`, etc. Is there precedent for this in the codebase? Would it be helpful?

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -0,0 +1,75 @@
+//! Process-global registry of in-progress video recordings.
+//!
+//! A recording's capture process must outlive the `StartRecording` tool call
+//! that launches it and survive until a later `StopRecording` call (possibly
+//! from a different subagent sharing the display), so the live handle lives here
+//! rather than in a per-call executor.
```
vkodithala: I'm not sure if this comment accurately communicates the intent of a process-global registry here. Since subagents don't share task context, a future subagent wouldn't actually be aware of a previous subagent's started recordings. Moreover, this probably isn't the type of behavior we'd want to support anyhow; each subagent should start and stop its own recording, unless the recording hits limits or the conversation fails somehow. Basically I just wonder if we can tighten this wording a bit.

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -0,0 +1,75 @@
...
+            sessions: HashMap::new(),
+            starting: false,
+        }
+    }
+
+    /// Reserves the single recording slot, failing if one is already active or
+    /// starting.
+    pub fn try_begin_start(&mut self) -> Result<(), String> {
+        if self.starting || !self.sessions.is_empty() {
+            return Err("A recording is already in progress on this display.".to_string());
```
vkodithala: Somewhat confused here, but this is probably just my misunderstanding/naivety. I see `RecordingController::new` is instantiated on app initialization, so I don't fully understand how `self.starting` indicates that a recording is already on progress on the current display. Are we assuming that the agent only has access to a single display? Does this sync up with our semantics of how Xvfb/computer use works more generally? If so, we should document this somewhere.

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        self.starting = true;
+        Ok(())
+    }
+
+    /// Registers a successfully started recording, releasing the start reservation.
+    pub fn finish_start(&mut self, recording_id: String, session: RecordingSession) {
+        self.starting = false;
```
vkodithala: It feels like, especially with the language "releasing" that we should just be using locks here anyhow. Maybe this means we can get rid of `self.starting` in favor of something a little more idiomatic. Any reason why we _aren't_ doing this?

app/src/ai/blocklist/action_model/execute/start_recording.rs
```diff
@@ -0,0 +1,110 @@
...
+use crate::ai::agent::AIAgentActionType;
+use crate::ai::blocklist::action_model::recording_controller::{
+    RecordingController, RecordingSession,
+};
+
+/// Reported maximum recording duration. Enforcement (auto-stop) is a follow-on;
+/// for now this is informational so the agent can plan around it.
+const MAX_RECORDING_DURATION: Duration = Duration::from_secs(10 * 60);
+/// Reported maximum recording size in bytes.
+const MAX_RECORDING_SIZE_BYTES: i64 = 1024 * 1024 * 1024;
```
vkodithala: This is 1GB and 10min. Not super important that we get this right from the beginning, but maybe we can document why we made this decision? i.e. by inspecting server code to analyze what startup memory normally is, where recordings are stored, and doing some napkin math around why we're limiting recording sizes in the first place/what we're preserving.

app/src/ai/blocklist/action_model/recording_controller.rs
vkodithala: Also, a ditto to one of my earlier comments where we should probably enumerate error types from the recording controller so this isn't ambiguous, and we have a documented understanding of failure modes. I generally don't like plain string errors.

app/src/ai/blocklist/action_model/execute/stop_recording.rs
```diff
@@ -0,0 +1,166 @@
...
+
+            // Pull the live session out now; an unknown id means the recording
+            // already stopped (stop is idempotent).
+            let session = RecordingController::handle(ctx)
+                .update(ctx, |controller, _| controller.take_session(&recording_id));
+            let Some(session) = session else {
+                return ActionExecution::<()>::Sync(AIAgentActionResultType::StopRecording(
+                    StopRecordingResult::Error(format!(
+                        "No active recording with id '{recording_id}'."
+                    )),
```
vkodithala: I'm being super redundant here and you probably already get the point, but we should enumerate the error types/failure modes here.

app/src/ai/blocklist/action_model/execute/stop_recording.rs
```diff
@@ -0,0 +1,166 @@
...
+                .and_then(|conversation| conversation.server_conversation_token())
+                .cloned();
+            let Some(server_conversation_token) = server_conversation_token else {
+                return ActionExecution::<()>::Sync(AIAgentActionResultType::StopRecording(
+                    StopRecordingResult::Error(
+                        "Current conversation has not been synced to the server yet".to_string(),
+                    ),
+                ))
+                .into();
+            };
```
vkodithala: In what cases would a conversation not have been synced to the server at this point? Not sure, but would we want to await/otherwise try to gracefully upload here vs. just failing outright? IIUC failing here means we never upload the artifact; there are probably better ways of going about this, or we want to mark as a KI/TODO to fix at the very least.

app/src/ai/blocklist/action_model/execute/stop_recording.rs
```diff
@@ -0,0 +1,166 @@
...
+
+                    match upload_result {
+                        Ok(upload) => {
+                            let completion_status = output.completion_status;
+                            let termination_reason = match completion_status {
+                                computer_use::RecordingCompletionStatus::Complete => {
+                                    "Stopped by agent".to_string()
+                                }
+                                computer_use::RecordingCompletionStatus::Incomplete => {
+                                    "Recording did not finalize cleanly".to_string()
```
vkodithala: nit: Does this map well to a 'termination reason' (as it's declared here)? Should it be something like "Recording interrupted; did not finalized cleanly" or something? Not opinionated on the wording, this is pretty rough, but the general idea applies.

app/src/ai/blocklist/action_model/execute/stop_recording.rs
```diff
@@ -0,0 +1,166 @@
...
+                    let uploader = FileArtifactUploader::new(ai_client, server_api);
+                    let request = FileArtifactUploadRequest {
+                        path: output.path,
+                        run_id: None,
+                        conversation_id: Some(server_conversation_token),
+                        description: None,
+                    };
+                    let upload_result = async {
+                        let association = uploader.resolve_upload_association(&request).await?;
+                        uploader.upload_with_association(request, association).await
```
vkodithala: Mostly a drive-by thought and not sure if this is necessary/advisable, but would we want to add some final checks (maybe a helper fn) to verify that the file itself is in the right format and wasn't visibly corrupted before upload?

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -0,0 +1,75 @@
...
+//!
+//! A recording's capture process must outlive the `StartRecording` tool call
+//! that launches it and survive until a later `StopRecording` call (possibly
+//! from a different subagent sharing the display), so the live handle lives here
+//! rather than in a per-call executor.
+
+use std::collections::HashMap;
+
+use warpui::{Entity, SingletonEntity};
+
```
vkodithala: This definitely seems better, but I'm worried that we're mixing in startup and teardown errors. i.e. `AlreadyInProgress` seems very much like a startup error, and `NoActiveRecording`/`ConversationNotSynced` seem scoped to teardown?

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -0,0 +1,75 @@
...
+        Self { handle }
+    }
+
+    pub fn into_handle(self) -> computer_use::RecordingHandle {
+        self.handle
+    }
+}
+
+/// Tracks recordings keyed by id and enforces a single active recording per
+/// display. Stop is idempotent: an unknown id resolves to `None`.
```
vkodithala: nit: Probably best not to say things like "one computer-use display for v0", a better way to enrich this would be to reference the fact that ambient agent environment setup uses one Xvfb instance, which provides a single display (with a todo/note that this could expand in the future perhaps?).

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// Releases the start reservation after a failed start.
+    pub fn abort_start(&mut self) {
+        self.starting = false;
+    }
+
+    /// Removes and returns the session for `recording_id`, if present.
+    pub fn take_session(&mut self, recording_id: &str) -> Option<RecordingSession> {
+        self.sessions.remove(recording_id)
+    }
```
vkodithala: nit: We probably don't need fn `take_session` anymore and can just inline it here, especially since this is the only caller?

app/src/ai/blocklist/action_model/execute/stop_recording.rs
```diff
@@ -0,0 +1,166 @@
...
+                    }
+                    .await;
```
vkodithala: Should mark as TODO(vkodithala) so it's explicitly tagged to me (do this for all todos in the future).

crates/computer_use/src/lib.rs
```diff
@@ -57,6 +59,103 @@ pub trait Actor: Send + Sync + 'static {
...
+/// Runtime-owned capture configuration for a recording.
+#[derive(Debug, Clone)]
+pub struct RecordingConfig {
+    /// Capture frame rate in frames per second.
+    pub frame_rate: u32,
+}
+
+impl Default for RecordingConfig {
+    fn default() -> Self {
+        Self { frame_rate: 15 }
```
vkodithala: Don't mention dogfood and prefix with "NOTE" so that the intent here is clear.

crates/computer_use/src/linux/recording.rs
```diff
@@ -0,0 +1,196 @@
...
+        command
+            .arg("-y")
+            .args(["-f", "x11grab"])
+            .args(["-framerate", &config.frame_rate.to_string()])
+            .args(["-video_size", &format!("{width}x{height}")])
+            .args(["-i", &display])
+            .args(["-c:v", "libx264"])
+            .args(["-preset", "ultrafast"])
+            .args(["-pix_fmt", "yuv420p"])
+            .args(["-movflags", "+faststart"])
```
vkodithala: Probably best to add some comments explaining that this enforces file time/size limits.

vkodithala:
* **#13227** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13044** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13044?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13044?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/computer_use/src/linux/recording.rs
```diff
@@ -0,0 +1,224 @@
...
+    async fn start(&self, config: RecordingConfig) -> Result<RecordingHandle, RecordingError> {
+        let display = std::env::var("DISPLAY").map_err(|_| RecordingError::MissingDisplay)?;
+
+        // libx264 with yuv420p requires even dimensions.
+        let (width, height) = query_display_dimensions()?;
+        let width = width & !1;
+        let height = height & !1;
+        if width == 0 || height == 0 {
+            return Err(RecordingError::InvalidDimensions { width, height });
+        }
```
danielpeng2, line 51: Would it be easy to support a recording that's 1 pixel off, rather than failing altogether if the dimensions are odd? It'd probably be fine if 1 pixel was cut off from each end.

crates/computer_use/src/linux/recording.rs
```diff
@@ -0,0 +1,224 @@
...
+            .args(["-movflags", "+faststart"]);
+        // Enforce capture limits in ffmpeg so abandoned recordings remain bounded.
+        if let Some(max_duration) = config.max_duration {
+            command
+                .arg("-t")
+                .arg(format!("{:.3}", max_duration.as_secs_f64()));
+        }
+        if let Some(max_size_bytes) = config.max_size_bytes {
+            command.arg("-fs").arg(max_size_bytes.to_string());
+        }
```
danielpeng2: nit: might be worth adding some (large) default bounds here, so if somehow these bounds aren't available in `config`, we still don't blow up the user's runner.

crates/computer_use/src/lib.rs
```diff
@@ -36,6 +39,48 @@ pub fn is_supported_on_current_platform() -> bool {
     }
 }
 
+#[derive(Debug, Error)]
+pub enum RecordingError {
```
danielpeng2, line 43: This is a lot of variants lol. Might be worth doing a bit of thinking through whether any of the cases can be collapsed together. A good litmus test might be, would a consumer using the `computer_use` crate need to branch differently depending on the specific error?

danielpeng2 (approved):

crates/computer_use/src/linux/recording.rs
vkodithala: So, we actually _do_ handle this gracefully in exactly the way you recommend here.

`{width/height} & !1` just tries to convert each dim to an even number, and then fails if dims go to zero on conversion (i.e. if width/height were 1, which shouldn't be hit).

crates/computer_use/src/linux/recording.rs
```diff
@@ -0,0 +1,220 @@
...
+                        completion_status = RecordingCompletionStatus::StoppedEarly;
+                    }
+                } else {
+                    completion_status = RecordingCompletionStatus::StoppedEarly;
+                }
+
+                match tokio::time::timeout(STOP_TIMEOUT, process.wait()).await {
+                    Ok(Ok(_)) => {}
+                    Ok(Err(_)) => completion_status = RecordingCompletionStatus::StoppedEarly,
+                    Err(_) => {
```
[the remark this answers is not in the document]
vkodithala: Handled!

crates/computer_use/src/linux/recording.rs
vkodithala: Agreed; we actually do set client-side defaults [here](https://github.com/warpdotdev/warp/blob/varoon/va-video-recording-execution/crates/computer_use/src/lib.rs#L146-L157) which are used by the executor in lieu of a config set by the server, [here](https://github.com/warpdotdev/warp/blob/varoon/va-video-recording-execution/app/src/ai/blocklist/action_model/execute/start_recording.rs#L75-L76). However `RecordingConfig::max_duration` and `max_size_bytes` were effecively modelled as optional even though they're unconditionally set with a fallback; fixed in latest.

crates/computer_use/src/lib.rs
vkodithala: Agreed, I collapsed these into 3 high-level modes: Environment, Start, and Finalize. 

TBH most of the errors here are actually due to VM setup/environment, so it's not actually LLM or argument error that can be fixed. IMO the main place where errors become important is telemetry, and think these modes break it down well.

PR #13046, opened 2026-06-25
Make authenticated Warp launch reliable: loud api-key warning + test-warp-ui hardening

vkodithala:
## Description
Authenticated Warp launches for computer-use agents were failing silently. `app/Cargo.toml` sets `default-run = "warp-oss"`, so a bare `cargo run` builds the OSS channel (`src/bin/oss.rs`), which drops `--api-key`/`WARP_API_KEY` and starts Warp logged out; only `cargo run --bin warp` (the Local/dogfood channel) honors the key. This PR makes that silent drop loud, and teaches the agent-facing skill to detect it.

Two changes:
- **`app/src/lib.rs`**: After API-key resolution (immediately before `AuthState::initialize`), if an `App` launch was given a key but it was dropped — non-dogfood channel, or API-key auth disabled — print a one-line warning to **stderr** via `eprintln!`, including the active channel via `ChannelState::channel()`. Scoped to the App launch path only; CommandLine mode is unaffected.
  - **Why `eprintln!` and nothing else:** this message exists purely to show up in the command output a human or agent reads. `log::*` is the wrong tool here — when the process's stdout is not a TTY (e.g. an agent capturing `cargo run` through a pipe), the logger routes records to the on-disk log file instead of the captured stream, so the message wouldn't be seen. `log::error!` would additionally be captured as a Sentry event and carries false "something failed" semantics. `eprintln!` always lands on stderr (which the skill greps for `provided but IGNORED`) and keeps Sentry out of it.
- **`resources/channel-gated-skills/dogfood/test-warp-ui/SKILL.md`**: Adds a "Verify the launch is authenticated" step — after launching, confirm Warp opened authenticated (not the logged-out onboarding/sign-in screen) and that the `cargo run` stderr does not contain `provided but IGNORED`; if it does (or the app is logged out), relaunch with `cargo run --bin warp`.

## Linked Issue

## Testing
`cargo check -p warp --lib` from the worktree root succeeds with no errors or warnings. The skill change is markdown-only.

## Agent Mode

Conversation: https://staging.warp.dev/conversation/b217489b-c8d2-4d7b-be84-839df2b1f167
Run: https://oz.staging.warp.dev/runs/019effad-34c2-7a02-a246-d8df3a3c521e

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -1259,7 +1259,7 @@ pub(crate) fn initialize_app(
     {
         let channel = ChannelState::channel();
         let warning = format!(
-            "WARNING: --api-key/WARP_API_KEY was provided but IGNORED on the '{channel}' channel — Warp is starting LOGGED OUT. API-key auth requires a dogfood build; relaunch with `cargo run --bin warp`."
+            "WARNING: --api-key/WARP_API_KEY was provided but IGNORED on the '{channel}' channel — Warp is starting LOGGED OUT. API-key auth is only available on internal (dogfood) builds."
         );
         eprintln!("{warning}");
     }
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -1246,6 +1246,24 @@ pub(crate) fn initialize_app(
         None
     };
 
+    // A key supplied to an App launch but dropped here means Warp will start logged out
+    // (non-dogfood channel or feature disabled). Surface this loudly so it isn't silent.
+    if api_key.is_none()
+        && matches!(
+            launch_mode,
+            LaunchMode::App {
+                api_key: Some(_),
+                ..
+            }
+        )
+    {
+        let channel = ChannelState::channel();
+        let warning = format!(
+            "WARNING: --api-key/WARP_API_KEY was provided but IGNORED on the '{channel}' channel — Warp is starting LOGGED OUT. API-key auth requires a dogfood build; relaunch with `cargo run --bin warp`."
+        );
+        eprintln!("{warning}");
+    }
+
     let auth_state = Arc::new(AuthState::initialize(ctx, api_key));
     timer.mark_interval_end("AUTH_MANAGER_SET_USER");
 
--- a/resources/channel-gated-skills/dogfood/test-warp-ui/SKILL.md
+++ b/resources/channel-gated-skills/dogfood/test-warp-ui/SKILL.md
@@ -33,6 +33,15 @@ Authenticating this way starts the app directly without interactive login prompt
 
 Initial builds may take several minutes; subsequent incremental builds are faster.
 
+### Verify the launch is authenticated
+
+After launching, confirm both of the following before testing:
+
+- Warp is **authenticated** — it opens straight to the terminal, NOT the logged-out onboarding/sign-in screen.
+- The `cargo run` stderr/terminal output does **not** contain the substring `provided but IGNORED`.
+
+If that warning appears (or the app is logged out), the wrong binary/channel was launched — stop and relaunch with `cargo run --bin warp`.
+
 ## Testing Workflow
 
 ### 1. Hardcode or Mock Data (When Needed)


app/src/lib.rs
```diff
@@ -1246,6 +1246,17 @@ pub(crate) fn initialize_app(
...
 
+    // A key supplied to an App launch but dropped here means Warp will start logged out
+    // (non-dogfood channel or feature disabled). Surface this loudly so it isn't silent.
+    if api_key.is_none() && matches!(launch_mode, LaunchMode::App { api_key: Some(_), .. }) {
+        let channel = ChannelState::channel();
+        let warning = format!(
+            "WARNING: --api-key/WARP_API_KEY was provided but IGNORED on the '{channel}' channel — Warp is starting LOGGED OUT. API-key auth requires a dogfood build; relaunch with `cargo run --bin warp`."
+        );
+        eprintln!("{warning}");
+        log::error!("{warning}");
```
vkodithala: agent: Is there a reason we use `eprintln`/`log::error!` here vs. `log::info!`? What's the difference?

vkodithala: They serve two different audiences, and `info!` would be wrong for both:

- `eprintln!` writes straight to the process stderr, unconditionally — no log-level filtering, no log-file redirection. That's deliberate: the whole verification flow (and the test-warp-ui / factory skills) tells the agent to grep the `cargo run` terminal output for the substring `provided but IGNORED`. `eprintln!` guarantees it actually shows up on the console regardless of `RUST_LOG` or how logging happens to be wired.
- `log::error!` routes the same message into Warp's structured/persisted logging (log files, telemetry). Using `error` (vs `info`) means it survives a level filter set to warn/error-only and is recorded with the right severity, so a later "why did this run start logged out?" investigation finds it instead of it being buried in info noise.
- `log::info!` would risk being filtered out entirely, and semantically this isn't normal informational flow — the user explicitly passed a key expecting auth and we silently dropped it, which is an error-worthy misconfiguration.

So: `eprintln!` = guaranteed live console signal (greppable), `log::error!` = filter-resilient persisted record. Belt-and-suspenders on purpose.

vkodithala:
* **#13046** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13046?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13046?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/lib.rs
```diff
@@ -1246,6 +1246,24 @@ pub(crate) fn initialize_app(
...
+            launch_mode,
+            LaunchMode::App {
+                api_key: Some(_),
+                ..
+            }
+        )
+    {
+        let channel = ChannelState::channel();
+        let warning = format!(
+            "WARNING: --api-key/WARP_API_KEY was provided but IGNORED on the '{channel}' channel — Warp is starting LOGGED OUT. API-key auth requires a dogfood build; relaunch with `cargo run --bin warp`."
```
danielpeng2: mega nit: do we want/need this line `API-key auth requires a dogfood build; relaunch with cargo run --bin warp.`? Only reason I ask is because it might be weird if an external user runs into this but can't actually run `cargo run --bin warp`. Not a huge deal though.

app/src/lib.rs
```diff
@@ -1246,6 +1246,24 @@ pub(crate) fn initialize_app(
...
+        );
+        eprintln!("{warning}");
```
danielpeng2, line 1264: might make sense to use a `log::warn` instead of `eprintln`?

danielpeng2 (approved):

app/src/lib.rs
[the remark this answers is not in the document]
vkodithala: Eh, this change should be mostly read-only. Probably don't need this, resolving.

app/src/lib.rs
vkodithala: Yeah this is a good q - I thought about this, but have seen agents frequently pipe outputs from build commands via something like `| -tail 30`. IIUC in our logger [here](https://github.com/warpdotdev/warp/blob/master/crates/warp_logging/src/native.rs#L567-L571), we write outputs to a logfile when we detect redirection like this, which means agents wouldn't see this warning in `readShellCommandOutput` tool call responses. There's also a precedent in this file for using `eprintln` for error logging, which I believe is for this reason, lmk if that makes sense/I'm misunderstanding anything. Pushing as-is for now!

app/src/lib.rs
vkodithala: Fixed! Avoided directly mentioning `cargo run --bin warp` since ext contributors won't be able to run this.

app/src/lib.rs
danielpeng2: I'm good with that!

danielpeng2 (approved):

PR #13056, opened 2026-06-25
Extend pane dragging to support tab groups - vertical tabs

johnturcoo:
## Description

This PR fixes several bugs associated with pane dragging (shown in demo below) and now handles pane dragging interactions with tab groups + pinned items. 

This PR reworks vertical tab pane dragging to:
- Decide the drop position from the cursor instead of overlapping tab drop targets (mirroring the pattern for pane dragging in htabs)
- Make pane dragging group aware (by using the cursor position to 'hit test' groups, similar to tab dragging patterns)
- Change rendering of tab highlights and add a `catch all` drop target to the whole tab panel to prevent flickering bugs

I suggest watching the demo of bugs for more context, but existing bugs: 
- dragging below all tabs in vtab bar causes flickering 
- dragging on the border of a "Over tab" / "Above tab" boundary causes flickering
- dragging over tab groups cause flickering and it is not possible to drop into groups

NOTE: This has no interaction with cross window dragging, as pane dragging is a completely different system than tab dragging.

## How it works

The main idea is to route the cursor position to the workspace, where the tabs and tab groups live. A pane header and the per-tab drop targets only know their own local geometry, they have no access to where tab groups are laid out. So the header emits the raw drag-cursor rect (`drag_position`) and forwards it to the workspace. The workspace takes that `drag_position` and the hovered tab and, using each group's laid-out rect, resolves a `BeforeTab { index, group }`  which captures both the index the new tab should land at and the group it's a part of (so we can differentiate dropping into a group's first/last slot vs. right before/after the group itself)

Something to understand here is the meaning of the `TabBarDropTargetData` that is on our drop targets: 
- `TabBarLocation::TabIndex(index)` means that the resolver will test the cursor against the tab and tab group geometry to determine where this pane should land
- `TabBarLocation::AfterTabIndex(index)` means that the resolver will place the pane right at tab index without any hit testing (we know this pane can not land in a group)

What is added in this PR and why:
- **Group hit-testing** — `insertion_group()` compares the cursor against each group's saved container rect to tell "inside a group" from "between groups". `refine_hovered_tab_index()` uses this helper function and also clamps past the pinned region.
- **Group-header drop target** — during a drag the group header becomes a drop target so you can drop above a group or into it as the first member. Expanded groups use a splittable target (the cursor position determines if the pane will land in the group). A collapsed group uses a plain target that always resolves to "before the group".
-  **Catch-all drop target** — the whole panel is a fallback target so the gaps between/below groups aren't dead zones. Without it, dragging into empty space hit no target and reflowed the pane in its last dragged-over tab (the end-of-bar flicker).
- **Tab-bar axis** — the drag carries a `TabBarAxis` so the header knows whether we are dealing with vertical or horizontal tabs
- **New indicators for tab groups** — tab groups now render an above indicator, allowing panes to be dropped above collapsed/expanded groups
- **Framework change** — added `Container::with_foreground_border`  this paints a border on top of a container without changing the layout. This prevents the existing flickering bug when dragging over a boundary, since a moving layout can cause mouse position to repeatedly swap between two states. Visually this keeps the same desired border highlight without the flickering bugs.

 

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4724/update-tab-groups-to-support-dragging-a-pane-into-them

## Testing
- [x] Verified that pane dragging within a window remains unchanged
- [x] Verified pane dragging into tab groups behaves as expected

### Screenshots / Videos

[Demo of bugs](https://www.loom.com/share/6d8fecde93d6469498c7a83157902666)

[Demo of fix](https://www.loom.com/share/ef9be076a6cb4991a220d353079c1d12)

### DIFF
--- a/app/src/code/view.rs
+++ b/app/src/code/view.rs
@@ -55,7 +55,7 @@ use crate::pane_group::pane::view::header::render_pane_header_draggable;
 use crate::pane_group::pane::{view, ActionOrigin, PaneHeaderAction};
 use crate::pane_group::{
     BackingView, CodePane, PaneConfiguration, PaneConfigurationEvent, PaneDragDropLocation,
-    PaneEvent,
+    PaneEvent, TabBarAxis,
 };
 use crate::quit_warning::UnsavedStateSummary;
 use crate::search::files::icon::icon_from_file_path;
@@ -1614,7 +1614,7 @@ impl CodeView {
                         origin: ActionOrigin::EditorTab(index),
                         drag_location: PaneDragDropLocation::TabBar(data.tab_bar_location),
                         drag_position,
-                        precomputed_tab_hover_index: None,
+                        tab_bar_axis: Some(TabBarAxis::Horizontal),
                     },
                 );
             } else {
--- a/app/src/pane_group/pane/view/header/mod.rs
+++ b/app/src/pane_group/pane/view/header/mod.rs
@@ -324,78 +324,61 @@ impl<P: BackingView> PaneHeader<P> {
         self.is_visible_in_pane_group
     }
 
-    /// Based on the drag position and tab bar location, returns whether or not the given drag
-    /// is over a tab, or between two tabs. This is done by splitting the tabs into quadrants and seeing
-    /// what quadrant the center of the dragged element lives.
+    /// Based on the drag position and tab bar location, returns whether the drag
+    /// is over a tab or between two tabs, by splitting the hovered tab into
+    /// quarters along the active axis (X for the horizontal bar, Y for the
+    /// vertical panel): the leading quarter inserts before, the middle half
+    /// merges onto the tab, and the trailing quarter inserts after.
     fn calculate_tab_focus_hover_index(
         drag_position: &RectF,
         tab_bar_location: &TabBarLocation,
+        axis: TabBarAxis,
         ctx: &ViewContext<Self>,
     ) -> TabBarHoverIndex {
+        let is_vertical = matches!(axis, TabBarAxis::Vertical);
         match tab_bar_location {
             TabBarLocation::TabIndex(idx) => {
-                if let Some(tab_rect) = ctx.element_position_by_id(tab_position_id(*idx)) {
-                    let tab_center_x = tab_rect.center().x();
-                    let tab_quarter_x = (tab_center_x + tab_rect.lower_left().x()) / 2.;
-                    let tab_three_quarters_x = (tab_center_x + tab_rect.lower_right().x()) / 2.;
-                    if drag_position.center().x() < tab_quarter_x {
+                let Some(tab_rect) = ctx.element_position_by_id(tab_position_id(*idx)) else {
+                    // If for some reason we can't retrieve the tab position, fall
+                    // back to the per-axis default: the vertical panel inserts
+                    // before the tab, the horizontal bar merges onto it.
+                    return if is_vertical {
                         TabBarHoverIndex::BeforeTab {
                             index: *idx,
                             group: None,
                         }
-                    } else if drag_position.center().x() < tab_three_quarters_x {
-                        TabBarHoverIndex::OverTab(*idx)
                     } else {
-                        TabBarHoverIndex::BeforeTab {
-                            index: *idx + 1,
-                            group: None,
-                        }
-                    }
-                } else {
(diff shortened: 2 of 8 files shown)


app/src/pane_group/pane/view/header/mod.rs
```diff
@@ -334,18 +339,71 @@ impl<P: BackingView> PaneHeader<P> {
...
+                index: *tab_count,
+                group: None,
+            },
+        }
+    }
+
+    /// Vertical-tabs variant of [`Self::calculate_tab_focus_hover_index`]:
+    /// splits the hovered tab row into quarters along the Y axis (top quarter
+    /// inserts before, middle half merges onto the tab, bottom quarter inserts
+    /// after).
```
vkodithala: nit: if this is the vtab variant, maybe we should just rename `calculate_tab_focus_hover_index` to `calculate_horizontal_tab_focus_hover_index` or smth? 

it also looks like there's a bunch of shared logic, maybe we could just parametrize the type of tab (vertical/horizontal) and extract shared logic into a helper? (idk if this second part is feasible and defer to you)

either way, we should disambiguate so it's clear that `calculate_tab_focus_hover_index` is specifically for htabs

app/src/code/view.rs
```diff
@@ -1614,7 +1614,7 @@ impl CodeView {
                         origin: ActionOrigin::EditorTab(index),
                         drag_location: PaneDragDropLocation::TabBar(data.tab_bar_location),
                         drag_position,
-                        precomputed_tab_hover_index: None,
+                        tab_bar_axis: Some(TabBarAxis::Horizontal),
```
vkodithala, line 1617: you're likely already aware of this, but just to be sure, this means pane-drag over vertical tabs/groups isn't supported for editor tabs?

not critical that we get this in here, but if it's a no-op for vtabs it's probably right that we support this as a fast-follow

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -1900,24 +1924,22 @@ fn render_groups(
...
                 ));
                 i += run_len;
             }
             None => {
                 let insert_before_index = tab_index;
-                let insert_after_index = (i == total_visible - 1).then_some(tab_index + 1);
+                // Gaps between tabs are covered by the next tab's before-indicator,
+                // and the area after the last tab by the trailing indicator below,
+                // so an ungrouped row doesn't render its own "after" indicator.
+                let insert_after_index = None;
```
vkodithala, line 1961: nice!

app/src/workspace/view/vertical_tabs.rs
```diff
@@ -2941,18 +2980,47 @@ fn render_grouped_tab_container(
...
+        // - Collapsed: the members are hidden, so there's nothing to drop *into*.
+        //   The only sensible result is to land the pane just above the whole
+        //   group, so `AfterTabIndex(first)` always inserts at the group's first
+        //   slot (directly above it), no matter where in the header the cursor is.
+        //   Dropping just below a collapsed group is the next tab/group's job, so
+        //   we don't handle it here.
+        // - Expanded: the first member is visible, so the drop should follow the
+        //   cursor: above the group near the top of the header, or into the group
+        //   as its new first member lower down. `TabIndex(first)` enables that by
+        //   testing the cursor against the first member's row.
```
vkodithala, line 3053: added a comment about this on your Loom, and very much non-critical, but i wonder if in the future we should consider auto-expanding collapsed tab groups upon hover?

vkodithala (approved):
great PR, left a few comments - nothing blocking. LGTM :rocket:

app/src/pane_group/pane/view/header/mod.rs
johnturcoo: I agree here, good callout

app/src/code/view.rs
johnturcoo: Yes, editor tab pane dragging already was not supported for vtabs. This can be a follow up and is non blocking IMO

app/src/workspace/view/vertical_tabs.rs
johnturcoo: Left a reply on the loom, but for viz will put it here:

I aligned on the behavior for tab dragging, and we specifically don't expand the group when dragging over one. I think it should be consistent here. Other tab groups like chrome don't expand when dragging another item over it.

PR #13081, opened 2026-06-26
Hide "New group with tab" on tabs that are already in a group

### LINKED ISSUE #13073: Vertical tabs: "New group with tab" is offered on a tab that is already in a group
_(Searched open issues — no existing report for this.)_

### Describe the bug

With the `GroupedTabs` feature enabled, right-clicking a tab that is **already a member of a tab group** still shows the **"New group with tab"** context-menu item. Selecting it spins the tab into a confusing one-member "New Group" (which reads as overwriting/duplicating the current tab). A grouped tab shouldn't offer to start a brand-new group from its menu.

### To reproduce

1. Enable the `GroupedTabs` feature flag (with vertical tabs).
2. Create a tab group containing a tab (or add a tab to a group).
3. Right-click that grouped tab.
4. Observe **"New group with tab"** is present in the context menu.

### Expected behavior

"New group with tab" should only appear for an **ungrouped** tab — mirroring the sibling **"Remove from group"** item, which is already gated on the tab being in a group (`self.group_id.is_some()`).

### Root cause (OSS codebase)

In `app/src/tab.rs`, `tab_group_menu_items` pushes the "New group with tab" item **unconditionally**; it is missing the `self.group_id.is_none()` guard that its sibling "Remove from group" uses a few lines below. Gating it the same way fixes the menu; regrouping a grouped tab stays reachable via drag / multi-select.

### Environment

- Built from the OSS repository (`master`); behaviour is behind the `GroupedTabs` feature flag.
- OS: macOS.
- Regression: No — present since the flag-gated grouping UI.
- Blocks daily use: No.
- Warp-only: Yes (Warp-specific UI).

I have a fix ready and would like to open a PR for this.

2nd1st:
## Description

The vertical-tab context menu offered **"New group with tab"** unconditionally. On a tab that is **already in a group**, selecting it spun the tab into a confusing one-member "New Group" that read as overwriting the current tab. This gates the menu item on `self.group_id.is_none()`, mirroring the sibling **"Remove from group"** item which is already gated on the tab being in a group. Regrouping a grouped tab stays reachable via drag / multi-select.

(The grouped-tabs UI is behind the `GroupedTabs` feature flag.)

## Linked Issue

Fixes #13073

- [x] Screenshots are included below.

## Testing

Added a regression test — `app/src/tab_tests.rs` (`tab::tests::new_group_with_tab_hidden_when_tab_already_grouped`): with `GroupedTabs` enabled it asserts `tab_group_menu_items` omits "New group with tab" for a grouped tab and still offers it for an ungrouped tab. The grouped-tab assertion fails against the old unconditional code.

### Screenshots / Videos

Right-click context menu with `GroupedTabs` enabled:

**Tab already in a group** — the menu no longer offers "New group with tab" (only "Remove from group", etc.):

Screenshot 2026-06-26 at 6 32 18 PM

**Ungrouped tab (for contrast)** — still offers "New group with tab":

Screenshot 2026-06-26 at 6 32 35 PM

_(Before this fix, the grouped-tab menu also offered "New group with tab" — see #13073.)_

## Agent Mode

### DIFF
--- a/app/src/tab.rs
+++ b/app/src/tab.rs
@@ -595,9 +595,18 @@ impl TabData {
         if !FeatureFlag::GroupedTabs.is_enabled() {
             return vec![];
         }
-        let mut menu_items = vec![MenuItemFields::new("New group with tab")
-            .with_on_select_action(WorkspaceAction::NewTabGroupFromTab(index))
-            .into_item()];
+        let mut menu_items: Vec<MenuItem<WorkspaceAction>> = vec![];
+        // Only offer "New group with tab" for an ungrouped tab. On a tab that is
+        // already in a group this created a confusing one-member "New Group" header
+        // (the user reads it as overwriting the current tab). Regrouping a grouped
+        // tab is still reachable via drag / multi-select.
+        if self.group_id.is_none() {
+            menu_items.push(
+                MenuItemFields::new("New group with tab")
+                    .with_on_select_action(WorkspaceAction::NewTabGroupFromTab(index))
+                    .into_item(),
+            );
+        }
         let has_other_groups = tab_groups.keys().any(|gid| Some(*gid) != self.group_id);
         if has_other_groups {
             menu_items.push(MenuItemFields::new_submenu(MOVE_TO_GROUP_LABEL).into_item());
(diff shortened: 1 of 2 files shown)


2nd1st:
@cla-bot check

PR #13057, opened 2026-06-25
Conversation streaming for the TUI

harryalbert:
## Description

This PR changes it so that the main cluster of conversation models (i.e. the `BlocklistAIContextModel`, `BlocklistAIInputModel`, `BlocklistAIController`, and `BlocklistAIHistoryModel`) necessary for conversation streaming and basic conversation management can be re-used by the `TUI` w/ no `TUI` or `GUI` specific code within said models (i.e. the models are not concerned w/ whether they are associated with a `TUI` or `GUI` surface).

Luckily enough, there's actually not much that needs to be changed within these models to get them working for the TUI and GUI in a surface-agnostic way. The major things that change are:
* the models stop referencing a `terminal_view_id: EntityId`, replaced with `terminal_surface_id: EntityId`. This is where the majority of the LOC changed come from in this PR, and it should be a behavior no-op. Just want to make sure we're clear on our naming here.
* the shared models no longer hold an `AgentViewController`, as this controller is GUI specific. What makes this a little tricky is that the `AgentViewController` was doing several jobs: it was telling the models whether an agent view was open or not, telling the models what conversation was currently selected, and emitting presentation lifecycle events consumed by the shared models. To handle these jobs, this PR adds an object-safe `ConversationSelection` abstraction that each surface implements and passes to the shared models behind a `ConversationSelectionHandle`. For `GUI` surfaces, `AgentViewConversationSelection` delegates presentation operations to `AgentViewController` and translates Agent View lifecycle events into events consumed by shared models. For `TUI` surfaces, `TuiConversationSelection` owns the pending conversation selection directly.

The history model also now distinguishes between transferring a conversation to a new terminal surface and simply marking an already-owned conversation as the active streaming target. This ensures automatic follow-ups don't accidentally transfer conversations between surfaces, while explicit transfers still clean up the previous renderer and selection.

Finally, this PR adds a one-shot `--prompt` runtime path to `warp_tui`, which streams a response directly back as output. I would just delete this as "testing code", but I actually think it's worthwhile to keep given our long-term plan of replacing the portion of the Oz CLI responsible for handling prompt submission with the TUI. The reusable conversation coordination lives in `TuiConversationModel`, while the one-shot stdout behavior lives in `PromptStreamSurface`.

This PR also adds a `--conversation-id` flag so that you can continue an existing conversation. This flag accepts a server conversation token, allowing the TUI to restore the corresponding conversation and test that follow-ups are WAI.

[...]

PR #13104, opened 2026-06-26
Wire up tui input, transcript view, into full TUI app

harryalbert:
## Description

This PR replaces the old TUI prompt-stream/stdout path with the first production-shaped TUI transcript session. The ownership split is clear: `RootTuiView` only gates login, `TuiTerminalSessionView` owns the authenticated terminal surface + prompt path, and `TuiTranscriptView` adapts the canonical `TerminalModel::BlockList` into TUI viewport content.

The most important model invariant is that TUI terminal blocks and TUI agent blocks are still ordered by the shared `TerminalModel::BlockList`, not a TUI-only transcript list.

### What to review

* `crates/warp_tui/src/terminal_session_view.rs` is the root renderer and handler for the transcript and input. It also handles sending prompts (was previously handled by the now deleted tui conversation model, but that model was just a thin wrapper around prompt submission and is now unnecessary)
* `crates/warp_tui/src/transcript_view.rs` is effectively the blocklist element, using the viewported list to render the blocklist. This view also owns history event handling, which I think is reasonable but curious on your take.
* `crates/warp_tui/src/tui_block_list_viewport_source.rs`is used by the transcript view to adapt the block list into viewport items, specifically terminal_blocks (`crates/warp_tui/src/terminal_block.rs`) and agent_blocks (`crates/warp_tui/src/agent_block.rs`)
    * terminal block rendering is pretty clear and can be skimmed IMO, but agent block rendering is worth a pass, specifically on how it does layout calculations for element heights (see the new `content` field)
* can skim over `crates/warpui_core/src/elements/tui/container.rs` — this PR adds per-axis padding but this is pretty schematic

Note: test-util flags were added in a bunch of places because `warp_tui` tests live in a separate crate, where warp’s `#[cfg(test)]` helpers are not compiled, so the feature explicitly exposes those test-only terminal helpers to downstream tests without making them part of normal builds. Apparently,  cfg(test) is only enabled for the crate whose tests Cargo is currently compiling, not for that crate’s dependencies.

## Testing

## Agent Mode


harryalbert:
* **#13291** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13104** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13104?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13104?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13177** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/warp_tui/src/tui_block_list_viewport_source.rs
```diff
@@ -0,0 +1,317 @@
...
+}
+
+impl TuiViewportedElement for TuiBlockListViewportSource {
+    fn visible_items(
+        &self,
+        window: TuiViewportWindow,
+        available_width: u16,
+        app: &AppContext,
+    ) -> TuiViewportContent {
+        let dirty_rich_content_items = self
```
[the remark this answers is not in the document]
harryalbert: good catch, I'm fixing this now

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,201 @@
...
+
+use warp::tui_export::{AIAgentTextSection, AIBlockModel};
+use warpui_core::elements::tui::{
+    Color, Modifier, TuiBuffer, TuiColumn, TuiConstraint, TuiContainer, TuiElement,
+    TuiLayoutContext, TuiParentElement, TuiRect, TuiSize, TuiStyle, TuiText,
+};
+use warpui_core::{AppContext, Entity, EntityIdMap, TuiView};
+
+const INPUT_PREFIX: &str = "≫ ";
+const INPUT_TEXT_COLOR: Color = Color::from_u32(0xffffff);
```
kevinyang372: We should avoid these hard-coded colors -- these should be derived from appearance

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,201 @@
...
+const OUTPUT_COLOR: Color = Color::from_u32(0xf1f1f1);
+
+/// Renderable pieces of an agent block; this will grow as we add tool calls and other sub-elements.
+#[derive(Clone, Debug, Eq, PartialEq)]
+enum TuiAgentBlockSection {
+    Input(String),
+    PlainText(String),
+}
+
+struct TuiAgentBlockElement {
```
kevinyang372: I think we need to have a strong motivation to justify creating a custom element instead of using a generic one

What's the main reason we need this?

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -0,0 +1,211 @@
...
+
+use crate::conversation_selection::TuiConversationSelection;
+use crate::input::{TuiInputView, TuiInputViewEvent};
+use crate::transcript_view::TuiTranscriptView;
+
+/// Width used before the first layout pass pushes the real terminal width into the editor.
+const INITIAL_INPUT_WIDTH: u16 = 80;
+const MAX_INPUT_TEXT_ROWS: u16 = 6;
+const BORDER_ROWS: u16 = 2;
+const INPUT_BORDER_COLOR: Color = Color::from_u32(0xd0d1fe);
```
kevinyang372: ditto appearance

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -0,0 +1,211 @@
...
+const SESSION_PADDING: u16 = 2;
+
+/// This surface emits no PTY intents; commands are driven only by the spawned shell.
```
kevinyang372: This comment is not true yeah? Commands should be driven by the terminal model

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -0,0 +1,211 @@
...
+            TuiTranscriptView::new(terminal_surface_id, model.clone(), ctx)
+        });
+        let input_model = ctx.add_model(|ctx| CodeEditorModel::new_tui(INITIAL_INPUT_WIDTH, ctx));
+        let input_view =
+            ctx.add_typed_action_tui_view(move |ctx| TuiInputView::new(input_model, ctx));
+        ctx.subscribe_to_view(&input_view, |view, _, event, ctx| {
+            view.handle_submitted_prompt(event, ctx);
+        });
+
+        ctx.subscribe_to_model(&model_events, |_, _, _, ctx| ctx.notify());
```
kevinyang372: Are we supposed to refresh the view on any update from model_events?

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// Sends a prompt to the selected conversation, creating one if needed.
+    fn send_prompt(&mut self, prompt: String, ctx: &mut ViewContext<Self>) {
+        let conversation_id = match self
+            .conversation_selection
+            .as_ref(ctx)
+            .selected_conversation_id(ctx)
+        {
+            Some(conversation_id) => conversation_id,
+            None => match self.conversation_selection.update(ctx, |selection, ctx| {
+                selection.try_start_new_conversation(AgentViewEntryOrigin::Cli, ctx)
```
kevinyang372: Should we create a new AgentViewEntryOrigin enum here?

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// Renderable pieces of an agent block; this will grow as we add tool calls and other sub-elements.
+#[derive(Clone, Debug, Eq, PartialEq)]
+enum TuiAgentBlockSection {
+    Input(String),
+    PlainText(String),
+}
+
+/// A simple TUI block backed by one agent exchange.
+pub(super) struct TuiAgentBlockView {
```
kevinyang372: Does this need to be a separate view? Or should it really be just a struct exposing some rendering methods

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,167 @@
...
+        // Match the GUI block prompt text: displayable input queries separated by newlines.
+        let input = self
+            .model
+            .inputs_to_render(app)
+            .iter()
+            .filter_map(|input| input.display_query())
+            .collect::<Vec<_>>()
+            .join("\n");
+        if !input.is_empty() {
+            sections.push(TuiAgentBlockSection::Input(input));
```
kevinyang372: Why doesn't TuiAgentBlockSection itself support multi-lined inputs?

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,167 @@
...
+                .with_padding_bottom(u16::from(!sections.is_empty()) * BLOCK_BOTTOM_PADDING_ROWS),
+        )
+    }
+}
+
+/// Converts one logical section into a renderable TUI element.
+impl TuiAgentBlockSection {
+    fn render_element(&self, top_padding: u16, app: &AppContext) -> Box<dyn TuiElement> {
+        let theme = Appearance::as_ref(app).theme();
+        let text_color: Color = GuiFill::from(theme.main_text_color(theme.surface_1())).into();
```
kevinyang372: Kind of weird that we are using GuiFill here -- are these truly Gui-only Fills or this is not accurately named

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,167 @@
...
+                )
+                .with_padding_top(top_padding),
+            ),
+        }
+    }
+}
+
+#[cfg(test)]
+#[path = "agent_block_tests.rs"]
+mod tests;
```
kevinyang372: uber nit: tests should be at the very bottom

crates/warp_tui/src/session.rs
```diff
@@ -34,52 +33,76 @@ impl Entity for TuiSession {
...
-    let (window_id, root) = ctx.add_tui_window(
-        AddWindowOptions {
-            window_style: WindowStyle::NotStealFocus,
-            ..Default::default()
+/// Creates the transcript root surface and starts the headless draw + input
+/// driver. Registered as a singleton so the session lives for the app lifetime.
+fn init(ctx: &mut AppContext) {
+    let banner = ctx.add_model(|_| BannerState::default());
+    let mut root_for_driver = None;
+    let manager = LocalTtyTerminalManager::<TuiTerminalSessionView>::create_tui_model(
```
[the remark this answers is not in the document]
kevinyang372: I think defering creating these until logged in makes sense to me

crates/warp_tui/src/session.rs
```diff
@@ -1,31 +1,31 @@
...
+use crate::root_view::RootTuiView;
+use crate::terminal_session_view::TuiTerminalSessionView;
 
-/// Holds the live TUI session for the app's lifetime; dropping it on app
-/// teardown restores the terminal.
+/// Holds the live TUI session for the app's lifetime.
 struct TuiSession {
-    _handle: TuiDriverHandle,
+    _driver: TuiDriverHandle,
+    _manager: ModelHandle<Box<dyn TerminalManagerTrait>>,
```
kevinyang372: Is this never expected to be used? If not, we should remove the `_` prefix and add `expect(dead_code)` for now

crates/warp_tui/src/terminal_grid_snapshot.rs
```diff
@@ -0,0 +1,141 @@
...
+use warpui_core::AppContext;
+
+#[derive(Clone)]
+struct TerminalCellSnapshot {
+    symbol: String,
+    style: TuiStyle,
+}
+
+/// Renderable terminal-grid rows copied out of the terminal-model lock.
+pub(super) struct TerminalGridSnapshot {
```
kevinyang372: I don't really understand what this is for

crates/warp_tui/src/theme.rs
kevinyang372, line 1: What is this?

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,171 @@
...
+        Self { model }
+    }
+
+    /// Replaces the exchange model backing this block.
+    pub(super) fn set_model(&mut self, model: Rc<dyn AIBlockModel<View = Self>>) {
+        self.model = model;
+    }
+
+    /// Returns this block's wrapped height at the given width.
+    pub(super) fn desired_height(&self, width: u16, app: &AppContext) -> usize {
```
kevinyang372, line 75: Just to make sure this is efficient to do on the main thread?

kevinyang372: Also are we applying any optimizations to avoid repeated height estimation on every render pass? (e.g. I think in GUI we only compute an agent block's desired height when it's marked as dirty) Does it actually matter for TUI?

crates/warp_tui/src/tui_block_list_viewport_source.rs
```diff
@@ -0,0 +1,350 @@
...
+            .scroll_top
+            .saturating_add(usize::from(window.viewport_height));
+        let mut descriptors = Vec::new();
+        let mut content_height = 0usize;
+        let mut cursor = block_list
+            .block_heights()
+            .cursor::<TotalIndex, BlockHeightSummary>();
+        cursor.seek(&TotalIndex(0), SeekBias::Right);
+
+        while let Some(item) = cursor.item() {
```
kevinyang372, line 165: Is this efficient? We are essentially iterating through every item in the blocklist which doesn't feel like we are utilizing the power of the sumtree

I would expect us to slice to the right index with the scroll_top dimension

crates/warp_tui/src/tui_block_list_viewport_source.rs
```diff
@@ -0,0 +1,350 @@
...
+impl TuiBlockListVisibleItemDescriptor {
+    fn render(
+        self,
+        model: &Arc<FairMutex<TerminalModel>>,
+        window: TuiViewportWindow,
+        available_width: u16,
+        app: &AppContext,
+    ) -> TuiVisibleViewportItem {
+        let visible_rows = self.visible_rows(window);
+        let origin_y = if matches!(&self.item, TuiBlockListVisibleItem::TerminalBlock { .. }) {
```
kevinyang372: Do we have different origin_y coordinate system for terminal block vs agent blocks? Why?

crates/warpui_core/src/core/app.rs
```diff
@@ -1103,10 +1103,14 @@ impl AppContext {
...
-    /// Subscribes to a [`ViewHandle`] for changes, calling `callback` with the emitted event whenever the view is invalidated.
+    /// Subscribes to a GUI or TUI [`ViewHandle`] for emitted events.
+    ///
+    /// This is bounded by [`Entity`] rather than [`View`](crate::View) because
+    /// TUI views implement [`TuiView`](crate::TuiView), not GUI `View`. The
+    /// [`ViewHandle`] parameter still limits callers to actual view instances.
     pub fn subscribe_to_view<S, F>(&mut self, handle: &ViewHandle<S>, mut callback: F)
     where
-        S: View,
+        S: Entity,
```
kevinyang372, line 1114: Entity feels a tad to broad here -- anyway we could represent S could be either View or TuiView here?

crates/warp_core/src/ui/theme/color.rs
```diff
@@ -356,6 +356,38 @@ impl WarpTheme {
     pub fn tooltip_background(&self) -> ColorU {
         internal_colors::neutral_6(self)
     }
+
+    /// Background for the dark-mode TUI transcript terminal surface.
+    pub fn tui_transcript_background(&self) -> Fill {
```
kevinyang372: Instead of creating these tui specific ui surface , I think each caller should just get call `self.surface_1()` / `self.foreground()` -- I don't think these are very shareable across the tui surface

harryalbert: fair point

crates/warp_tui/src/agent_block.rs
harryalbert: yeah, the GUI also does this work on the main thread so I don't think it's too concerning.

For the repeated-height optimization, the GUI actually does re-compute all visible heights for every render pass (recomputing the dirty view heights first and then nearby visible heights after), but it avoids the double layout call for both height calculating and then rendering. 

The TUI path is doing an extra pre-layout desired_height() pass for visible agent blocks, then laying them out again in TuiViewportedList. I’ll change the TUI viewport path so visible agent block heights are reported from the normal viewport layout result back into the block-list height cache, then we can keep the existing dirty pre-pass so viewport slicing starts from up-to-date heights.

harryalbert: had the agent create a flow diagram for old vs. new call path to make this clearer.

What we're doing now:
```
TuiBlockListViewportSource::visible_items()
  -> measured_visible_agent_heights()
       -> agent_block.desired_height()
            -> render_element()
            -> element.layout()          // layout #1, only to get height

  -> descriptor.render()
       -> returns element

TuiViewportedList::layout_visible_elements()
  -> element.layout()                    // layout #2, actual viewport layout
  -> stores visible clipped element
  ```
  
  What I'll update it to:
  ```
TuiViewportedList::layout_visible_elements()
└─ viewport_content()
   └─ TuiBlockListViewportSource::visible_items()          [PASS 1 query]
      ├─ take_dirty_rich_content_items()
      ├─ measured_dirty_agent_heights(): desired_height()  // layout, DIRTY items only
      │     └─ write_line_heights() → tree cache updated
      ├─ visible_items_in_window()                         // seek tree by CACHED heights
      └─ item.render() per visible item
            → TuiVisibleViewportItem { origin_y, element,
                                       measured_height_id }  // Some only for AgentBlock
└─ layout_viewport_content()                               [PASS 1 layout]
   ├─ element.layout(loose, width × u16::MAX) → full_size  // lays out ALL visible items
   └─ record (item_id, full_size.height)                   // only where measured_height_id = Some
└─ source.update_visible_item_heights(measured_heights)     // reconcile → bool
   └─ if true: viewport_content() + layout_viewport_content() again  [PASS 2]
  ```

File: crates/warp_tui/src/agent_block.rs (hunk 9 above)
Comment (harryalbert) [reply crates/warp_tui/src/agent_block.rs]: It should, I'll update

File: crates/warp_tui/src/agent_block.rs (hunk 10 above)
Comment (harryalbert) [reply crates/warp_tui/src/agent_block.rs]: just not accurately named. I'll update

File: crates/warp_tui/src/session.rs (hunk 13 above)
Comment (harryalbert) [reply crates/warp_tui/src/session.rs]: yeah not expected to be used. I'll update

File: crates/warp_tui/src/session.rs (hunk 12 above)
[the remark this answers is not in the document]
Comment (harryalbert) [reply crates/warp_tui/src/session.rs:112]: yeah agreed. I'll update

File: crates/warp_tui/src/terminal_grid_snapshot.rs (hunk 14 above)
Comment (harryalbert) [reply crates/warp_tui/src/terminal_grid_snapshot.rs]: it was so that we didn't have to hold a terminal model lock during  paint to figure out which rows to render, but I think we can just hold the lock, it seems like we do it all over the place in other render fns

File: crates/warp_tui/src/theme.rs
Comment (harryalbert) [reply crates/warp_tui/src/theme.rs:1]: ope, meant to fully delete, good catch

File: crates/warp_tui/src/terminal_session_view.rs (hunk 6 above)
Comment (harryalbert) [reply crates/warp_tui/src/terminal_session_view.rs]: no good point, we can be more careful here

File: crates/warp_tui/src/terminal_session_view.rs (hunk 7 above)
Comment (harryalbert) [reply crates/warp_tui/src/terminal_session_view.rs]: yeah probably, good point

File: crates/warp_tui/src/tui_block_list_viewport_source.rs (hunk 16 above)
Comment (harryalbert) [reply crates/warp_tui/src/tui_block_list_viewport_source.rs:165]: yeah you're right, oversight on my part.

File: crates/warp_tui/src/tui_block_list_viewport_source.rs (hunk 17 above)
Comment (harryalbert) [reply crates/warp_tui/src/tui_block_list_viewport_source.rs]: no good reason, I'll fix

File: crates/warpui_core/src/core/app.rs (hunk 18 above)
Comment (harryalbert) [reply crates/warpui_core/src/core/app.rs:1114]: hmm, I investigated w/ the agent a bit and I'm not sure there's a great way to do this that's low lift (and rust doesn't just allow us to be like TuiView || View unfortunately).

Comment (harryalbert) [reply crates/warpui_core/src/core/app.rs:1114]: unfortunately we can'd do something like:
```
pub trait AnyViewKind: Entity {}

impl<T: View> AnyViewKind for T {}
impl<T: TuiView> AnyViewKind for T {}
```

because a type could theoretically implement both View and TUIView, so rust would fail the type

File: crates/warp_tui/src/terminal_session_view.rs (hunk 5 above)
Comment (harryalbert) [reply crates/warp_tui/src/terminal_session_view.rs]: yeah you're right, not accurate. Updated

File: crates/warp_tui/src/agent_block.rs (hunk 8 above)
Comment (harryalbert) [reply crates/warp_tui/src/agent_block.rs]: yeah, just because the TUI block-list rich-content path stores/render-controls agent blocks by ViewHandle<TuiAgentBlockView> / view id.

we could make the view a thin adapater and then have a seaprate struct exposing some rendering methods, but that feels like more indirection than necessary imo.

Comment (harryalbert) [reply crates/warp_tui/src/agent_block.rs]: We could also try to make the blocklist infrastructure key off of something else besides view id, but that's a bigger refactor across the GUi and TUI

Comment (kevinyang372) [reply crates/warp_tui/src/agent_block.rs]: Ah I see -- can we just name this TuiAIBlock then? This maps directly to the GUI AIBlock concept

File: crates/warp_tui/src/agent_block.rs [hunk 20]
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Extracts this exchange's visible input/output into logical render sections.
+    fn sections(&self, app: &AppContext) -> Vec<TuiAgentBlockSection> {
+        let mut sections = Vec::new();
+        let input_lines = self
+            .model
+            .inputs_to_render(app)
+            .iter()
+            .filter_map(|input| input.display_query())
+            .flat_map(|query| query.split('\n').map(str::to_owned).collect::<Vec<_>>())
```
kevinyang372: Ah I was thinking we don't even need to store Vec<String> for input

We can just store a String which it could contain linebreaks for the multi-line case?

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,212 @@
...
+            .collect::<Vec<_>>();
+        if input_lines.iter().any(|line| !line.is_empty()) {
+            sections.push(TuiAgentBlockSection::Input(input_lines));
```
kevinyang372: This will collapse empty linebreaks from user inputs which we might not want to do?

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,212 @@
...
+            // Output is many sections (one per text section), so top padding is
+            // applied only to the section right after the input, giving a single
+            // gap at the input→output boundary rather than before every line.
+            let follows_input = index
+                .checked_sub(1)
+                .is_some_and(|prev| matches!(sections[prev], TuiAgentBlockSection::Input(_)));
+            column = column.with_child(section.render_element(u16::from(follows_input), app));
+        }
+
+        Box::new(
```
kevinyang372: nit: Why we are calling Box::new instead of TuiElement::finish?

crates/warp_tui/src/root_view.rs
```diff
@@ -1,64 +1,42 @@
...
-/// Rows the box's border occupies (top + bottom).
-const BORDER_ROWS: u16 = 2;
-
-/// The root view: a login placeholder until the user is authenticated, then a
-/// single bordered input box docked at the bottom of the screen. Owns the
-/// editor-backed [`TuiInputView`] as a child view.
+/// The app-level TUI shell. It gates the authenticated terminal session on login state.
 pub struct RootTuiView {
-    input: ViewHandle<TuiInputView>,
+    terminal_session: Option<ViewHandle<TuiTerminalSessionView>>,
```
kevinyang372: Similar to AuthOnboardingState in the GUI root view, I think it's clearer to have explicit enums for each state (e.g. `Auth`, `Terminal(ViewHandle<TuiTerminalSessionView>)`)

crates/warp_tui/src/agent_block.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// Converts one logical section into a renderable TUI element.
+impl TuiAgentBlockSection {
+    fn render_element(&self, top_padding: u16, app: &AppContext) -> Box<dyn TuiElement> {
+        let theme = Appearance::as_ref(app).theme();
+        match self {
+            Self::Input(lines) => {
+                let text_color = Fill::from(theme.foreground()).into();
+                let accent = ThemeFill::from(theme.terminal_colors().normal.cyan);
+                let background = Fill::from(
```
kevinyang372, line 155: Is it common for CLIs to set background colors or they are expected to respect whatever color the terminal is rendering for background?

kevinyang372 (approved):

crates/warp_tui/src/agent_block.rs
harryalbert: yeah for sure, good point

crates/warp_tui/src/agent_block.rs
harryalbert: yeah true, I'll fix

crates/warp_tui/src/agent_block.rs
harryalbert: because TuiElement doesn't have a finish fn lol, but it probably should. I'll add one in

crates/warp_tui/src/agent_block.rs
harryalbert: depends on the CLI — claude code doesn't but opencode does. I can get rid of the background painting for now, seems easy to add back in if we decide we want it

crates/warp_tui/src/root_view.rs
harryalbert: yeah good call

crates/warp_tui/src/agent_block.rs
harryalbert: yeah can do

PR #13134, opened 2026-06-28
fix(markdown_parser): prevent unclosed <u> tags from pairing via emphasis algorithm

### LINKED ISSUE #12863: markdown_parser: unclosed `<u>` markers pair via the emphasis algorithm and delete surrounding text
### Summary

`parse_markdown` silently **deletes text** when a `<u>` underline marker is left unclosed and another `<u>` appears later in the same line. The two `<u>` markers are matched against each other as an emphasis pair and consumed, taking the text around them with them.

### Reproduction (against the real parser)

| Input | Expected (literal round-trip, like an unclosed `*`) | Actual |
|---|---|---|
| `<u><u>` | `<u><u>` | **empty line — all text gone** |
| `<u>a<u>` | `<u>a<u>` | `a` (both `<u>` deleted) |
| `a <u>word<u> b` | `a <u>word<u> b` | `a word b` (both `<u>` deleted) |

```rust
assert_eq!(parse_markdown("<u><u>").unwrap().raw_text(), "<u><u>\n"); // fails: left is "\n"
```

### Root cause

`<u>` is tokenized as an `UnderlineStart` delimiter (`crates/markdown_parser/src/markdown_parser.rs`, the `InlineToken::Delimiter` arm) and pushed onto the shared delimiter stack with `can_close` set from the right-flanking rules (`Delimiter::new`). As a result `process_emphasis` sees two `<u>` markers as a matchable opener/closer pair — `can_open_for` passes because they share `kind == DelimiterKind::UnderlineStart` — and the `opener.kind == UnderlineStart` branch in `process_emphasis` consumes both markers and removes their text nodes.

Underline is only ever supposed to be closed by an explicit `</u>`, which is handled separately by `parse_underline` (it never goes through `can_open_for`). So pairing two `<u>` openers through the emphasis path is always wrong; an unclosed `<u>` should round-trip to literal text, exactly like an unclosed `*`.

### Impact

`parse_markdown` is the main renderer for agent output and markdown-viewer content, so a partial/streamed `<u>` or two `<u>` tags silently drop the literal text. This is data loss in rendered output, not a crash.

### Notes

- Crate-scoped and reproducible with a pure unit test (no UI needed).
- I have a small fix ready (make `can_open_for` reject `UnderlineStart`; ~8 lines) with regression tests for the cases above, and all existing `markdown_parser` tests still pass. Happy to open a PR once this is labeled `ready-to-implement`.

### Operating system

All (platform-independent parser logic)

Jason-Shen2:
## Summary

When two `<u>` underline start tags appear without closing `</u>` tags, they were incorrectly paired as an opener/closer pair through the emphasis algorithm, causing the literal `<u>` markers and surrounding text to be silently deleted.

## Root Cause

`<u>` was tokenized as `UnderlineStart` with `can_close` set from right-flanking rules, allowing `process_emphasis` to match two `<u>` openers as a pair. Underline is only meant to be closed by an explicit `</u>` tag (handled by `parse_underline`), never by another `<u>`.

## Fix

1. Set `can_close = false` for `UnderlineStart` delimiters so `<u>` is never treated as a closing delimiter by `process_emphasis`
2. Add a guard in `can_open_for` to reject `UnderlineStart` pairs (defense in depth)

## Regression Tests

Added tests covering the cases from the issue:
- `<u><u>` round-trips as literal text (was: empty line, all text deleted)
- `<u>a<u>` preserves both `<u>` markers (was: `a`, both markers deleted)
- `a <u>word<u> b` does not delete text (was: `a word b`, markers deleted)

Fixes #12863

### DIFF
--- a/crates/markdown_parser/src/markdown_parser.rs
+++ b/crates/markdown_parser/src/markdown_parser.rs
@@ -1758,7 +1758,7 @@ impl Delimiter {
                 right_flanking && (!left_flanking || followed_by_punctuation)
             }
             DelimiterKind::Strikethrough => right_flanking,
-            DelimiterKind::UnderlineStart => right_flanking,
+            DelimiterKind::UnderlineStart => false,
         };
 
         Self {
@@ -1781,6 +1781,13 @@ impl Delimiter {
     fn can_open_for(&self, other: &Delimiter) -> bool {
         // Base rules that apply to all styling.
         if !self.can_open || self.kind != other.kind {
+            return false;
+        }
+
+        // Underline start (`<u>`) is only ever closed by an explicit `</u>` tag,
+        // handled separately by `parse_underline`. It must never be paired with
+        // another `<u>` through the emphasis algorithm.
+        if self.kind == DelimiterKind::UnderlineStart {
             return false;
         }
 
--- a/crates/markdown_parser/src/markdown_parser_tests.rs
+++ b/crates/markdown_parser/src/markdown_parser_tests.rs
@@ -1918,6 +1918,24 @@ fn test_parse_empty_underline() {
 }
 
 #[test]
+fn test_unclosed_underline_roundtrips_literal() {
+    assert_eq!(
+        parse_all("<u><u>", parse_inline),
+        vec![FormattedTextFragment::plain_text("<u><u>")]
+    );
+    assert_eq!(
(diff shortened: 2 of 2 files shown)


Jason-Shen2:
@cla-bot check

kevinyang372 (approved):

PR #13151, opened 2026-06-28
Fix watcher rebuild storm: skip re-indexing gitignored directories

### LINKED ISSUE #13153: Windows: repo watcher saturates the background thread pool re-indexing gitignored dirs (e.g. `target/`), starving other background work
## Summary
On Windows, heavy churn in a large **gitignored** directory — e.g. a Rust `target/` during a `cargo build`, or while rust-analyzer is running — makes WarpDev's repository watcher repeatedly re-index that ignored subtree. Each re-walk is slow, blocking filesystem I/O, and the walks run directly on the background-executor's worker threads, so they **occupy the entire background thread pool**. While that happens, other background work is **starved**: most visibly, opening new terminal sessions bootstraps far more slowly because their startup tasks sit in the runtime's queue behind the walks. The pool stays saturated until the filesystem churn stops; restarting WarpDev clears it until the next build. Reproduced repeatedly and confirmed via process-dump symbolication.

> CPU caveat: in Task Manager this looks like only ~10–20% CPU, which *understates* the problem. The work is I/O-bound (threads spend most of their time blocked in filesystem syscalls), so a low CPU percentage hides the fact that every worker thread is tied up. The real symptom is pool saturation and starvation of other background tasks, not CPU burn.

## Environment
- OS: Windows (NTFS; Windows Defender real-time protection enabled)
- CPU: Intel Core Ultra 9 285H — 16 logical cores (so the background pool is 16 worker threads)
- Trigger context: a git repo indexed in Warp with an active `cargo build` and/or rust-analyzer writing into `target/`. In the original report, neovim (which starts rust-analyzer) and Warp sessions were both open in the same large Rust repo.

## Symptoms
- **New terminal sessions bootstrap much more slowly** while the churn is active — their background startup work is queued behind the subtree walks.
- More generally, anything scheduled on the shared background executor is delayed until a worker frees up; the pool does not drain until the churn stops.
- Sustained background disk I/O, and CPU sits around ~10–20% in Task Manager (low only because the work is I/O-bound — see caveat), never returning to idle.
- Closing neovim/rust-analyzer does not help while churn continues; it recurs on the next build. A restart returns things to idle until the next build.

## Root cause
The incremental repository-metadata watcher re-walked entire gitignored subtrees on every debounced filesystem event:

1. Git repos are registered with a single **recursive** OS watch plus a descend filter (`repo_watch_filter` / `should_watch_repo_directory` in `crates/repo_metadata/src/entry.rs`) that is meant to prune gitignored dirs like `target/`.
2. That descend filter is only honored by the **Linux/inotify** backend. On **macOS (FSEvents)** and **Windows (ReadDirectoryChangesW)** the OS does native recursive watching and ignores the per-directory descend predicate, so events for `target/**` are still delivered. (See the `cfg!(target_os = "linux")` branch and comment in `crates/repo_metadata/src/local_model.rs`.)
3. For each newly-created directory, `compute_file_tree_mutations` (`crates/repo_metadata/src/local_model.rs`) called `Entry::build_tree_with_standing_queries`. The gitignore check inside the builder (`evaluate_entry`) is **not ancestor-aware**, so a deep ignored path like `target/debug/.fingerprint` was not recognized as ignored and its entire subtree (tens of thousands of files) was walked — on every event.

## Dump analysis
We captured a process dump while WarpDev was busy and symbolicated it (PDB pulled from Sentry):
- The background-executor is a tokio runtime with **one worker thread per logical core** (`num_cpus`). On this 16-core machine that is 16 workers, and the dump showed **all 16** parked in `repo_metadata::entry::Entry::build_tree_with_force_included_paths_and_ancestor` — i.e. the **entire pool was saturated**, leaving no worker to service other background tasks. This is the direct cause of the slow session bootstrapping.
- Their leaf frames were synchronous Windows filesystem syscalls: `NtCreateFile`, `ZwQueryDirectoryFileEx`, `ZwClose`. Because the walks are synchronous, blocking I/O run directly on the async workers (not offloaded via `spawn_blocking`), a blocked worker cannot poll any other task — and the process still showed only ~10–20% CPU because those threads were mostly *waiting* on disk, not computing.
- The triggering call chain was `LocalRepoMetadataModel::handle_watcher_event` → `compute_file_tree_mutations` → `Entry::build_tree_with_standing_queries` → `Entry::build_tree_with_force_included_paths_and_ancestor`.
- A second dump captured while the app was **idle** showed all `background-executor` threads parked with **zero** `build_tree` frames, confirming the activity correlates with watcher work.
- Temporary diagnostic logging captured lines such as:
  `rebuilt subtree dir=...\target\debug\.fingerprint files=27245 elapsed_ms=53021`
  — ~53s of wall-clock for a single subtree walk (~27k files, mostly spent waiting on I/O), with walks queued back-to-back so the pool stayed continuously busy and the backlog grew.

## Why this is worse on Windows
The buggy logic is **not** Windows-specific — macOS receives the same `target/**` events and runs the same re-walks. The difference is **cost**, which is what turns wasted work into pool saturation:
- On **NTFS**, every open/enumerate traverses the I/O filter-driver stack, and **Windows Defender** intercepts file opens. A `target/` full of tiny files (`.fingerprint`, `deps`, `incremental`) is the worst case, so a single walk takes ~50–90s and ties up a worker that whole time.
- On **macOS/APFS** with a warm cache the same walk is ~1–3s and isn't AV-gated by default, so workers free up quickly and the pool never stays saturated (a brief blip you'd attribute to cargo/rust-analyzer).
- **Linux** is effectively immune: the inotify descend filter prevents watches on `target/` entirely (events never arrive unless the user explicitly expands the directory).

Note: the `target/` layout is identical across platforms and the gitignore-matching flaw is platform-independent. It is not the directory structure — it is the cost of walking it on NTFS+Defender, compounded by running rust-analyzer + Warp + live builds against the same large repo.

## Reproduction
1. Open a large Rust git repo in WarpDev on Windows (so it gets indexed/watched).
2. Keep Warp sessions in the repo; optionally run rust-analyzer (e.g. via neovim) in the same repo.
3. Trigger heavy `target/` churn, e.g. `cargo build` or `cargo clean && cargo build`.
4. Observe new sessions bootstrap slowly and background updates lag while the build runs, with elevated disk I/O, not returning to normal until the churn stops.

## Fix
PR #13151 makes the incremental watcher path represent newly-created gitignored directories as **unloaded placeholders** (matching the initial index's `IgnoredPathStrategy::IncludeLazy` behavior) instead of recursively walking them, using an ancestor-aware ignore check. Force-included paths (e.g. skill-provider directories) are still materialized. After the fix, a full `cargo build` churns thousands of `target/` paths with **zero** subtree rebuilds, so the worker pool stays free and other background work (including new-session bootstrapping) is no longer starved.

_Filed from a Warp Agent Mode investigation._

acarl005:
## Description
Fixes a bug where the repository file watcher re-indexed entire **gitignored** directories on every filesystem change. On Windows each re-walk is slow, blocking I/O that runs on the background-executor's worker threads, so the walks **saturate the entire background thread pool and starve other background work** - most visibly, new terminal sessions bootstrap far more slowly because their startup tasks queue behind the walks.

On a git repository, an incremental watcher update (`compute_file_tree_mutations` in `crates/repo_metadata/src/local_model.rs`) called `Entry::build_tree_with_standing_queries` for every newly-created directory, even gitignored ones. Inside the tree walk, `evaluate_entry` checks gitignore against the literal path only (`check_ancestors = false`), so a *deep* ignored path such as `target/debug/.fingerprint` was not recognized as ignored and its whole subtree was walked. During a `cargo` build this re-walked ~27,000 files per event (50-90s each), fanned across the background-executor pool.

The background-executor is a tokio runtime with one worker thread per logical core (`num_cpus`). On the repro machine (16 logical cores) the dump showed all 16 workers parked in the walk, so the entire runtime was busy and nothing else scheduled on it could run until a walk finished.

### Why this hits Windows hardest
Two platform differences compound:
- **Linux never sees the churn.** Git repos use a "recursive" watch on every platform, but on Linux that is implemented as per-directory inotify watches registered only on directories that pass the descend filter (`should_watch_repo_directory`), which prunes gitignored subtrees. `target/` is therefore never watched and its events never reach `compute_file_tree_mutations`. macOS and Windows instead get a single recursive OS watch (FSEvents / `ReadDirectoryChangesW`) that ignores the descend filter and delivers events for the whole subtree, gitignored dirs included.
- **Each re-walk is far more expensive on Windows.** The walk is blocking, per-file filesystem I/O (`NtCreateFile` / `NtQueryDirectoryFileEx`), and those syscalls are much slower on Windows — amplified by the NTFS filter-driver stack and real-time antivirus/Defender scanning — than the same traversal on ext4/APFS. So even where macOS receives the same events, on Windows a single `target/debug/.fingerprint` change can take 50-90s to re-walk ~27k files: long enough to pin every worker thread and starve everything else.

The fix: in `compute_file_tree_mutations`, when an added directory is gitignored (using the ancestor-aware `path_is_ignored`), insert an unloaded placeholder instead of walking it, exactly how the initial index represents ignored directories (`IgnoredPathStrategy::IncludeLazy`). Force-included paths (e.g. skill-provider dirs) are exempted so they still materialize.

Root cause was found by symbolicating a Windows process dump (all 16 background-executor workers were in `repo_metadata::entry::Entry::build_tree_with_force_included_paths_and_ancestor` doing filesystem syscalls, starving other background tasks) and confirmed with temporary watcher instrumentation that is removed in this PR.

Created with Warp Agent Mode: https://app.warp.dev/conversation/3ac9dd74-82ae-4401-aea2-ae0e8bfc8b0d

## Linked Issue
Closes #13153

## Testing
Ran the app locally (OSS build via `cargo run`) with repositories open and triggered heavy `target/` churn with a full `cargo build`:
- Before the fix: a single change under `target/debug/.fingerprint` produced full-subtree rebuilds (`files=27245 elapsed_ms=53021`-style, 50-90s each) that occupied every worker thread; new sessions were slow to bootstrap and background updates lagged.
- After the fix: the same build churned thousands of `target/` paths with **zero** subtree rebuilds; the pool stayed free and sessions/background work were unaffected.

Confirmed with temporary `[watcher-diag]` logging that was used to capture the before/after and then removed in this PR. `cargo check` and `cargo clippy` pass on `repo_metadata`.

Also added some unit tests.

- [x] I have manually tested my changes locally with `./script/run` (via `cargo run`)

### Screenshots / Videos
N/A - no UI changes.

CHANGELOG-BUG-FIX: [Windows] Fixed background work (including new terminal sessions) being starved and slow to start, plus sustained disk usage, caused by repeatedly re-indexing gitignored directories (e.g. build output like `target/`) when many files changed at once.

### DIFF
--- a/crates/repo_metadata/src/entry.rs
+++ b/crates/repo_metadata/src/entry.rs
@@ -152,12 +152,14 @@ impl Entry {
         )
     }
     /// Builds the materialized tree and standing results during the same filesystem traversal.
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn build_tree_with_standing_queries(
         path: impl Into<PathBuf>,
         files: &mut Vec<FileMetadata>,
         gitignores: &mut Vec<Gitignore>,
         remaining_file_quota: Option<&mut usize>,
         options: BuildTreeOptions<'_>,
+        ancestor_is_ignored: bool,
         standing_results: &mut StandingQueryResults,
         definitions: &StandingQueryDefinitions,
     ) -> Result<Self, BuildTreeError> {
@@ -171,7 +173,7 @@ impl Entry {
             gitignores,
             remaining_file_quota,
             options,
-            false,
+            ancestor_is_ignored,
             Some(&mut standing_queries),
         )
     }
--- a/crates/repo_metadata/src/entry.rs
+++ b/crates/repo_metadata/src/entry.rs
@@ -691,7 +691,7 @@ pub fn is_git_internal_path(path: &Path) -> bool {
 /// `force_included_paths`. Each force-included path is a relative component
 /// sequence (e.g. `.agents/skills`) matched against the tail of `path`, so a
 /// match also holds for the ancestor prefixes leading to it.
-fn matches_force_included_path(path: &Path, force_included_paths: &[PathBuf]) -> bool {
+pub(crate) fn matches_force_included_path(path: &Path, force_included_paths: &[PathBuf]) -> bool {
     let path_components: Vec<_> = path
         .components()
         .filter_map(|component| match component {
--- a/crates/repo_metadata/src/entry_tests.rs
+++ b/crates/repo_metadata/src/entry_tests.rs
@@ -394,6 +394,7 @@ fn standing_queries_report_skills_below_an_ignored_directory() {
                 force_included_paths: &[std::path::PathBuf::from(".agents/skills")],
                 budget_exceeded_behavior: super::BudgetExceededBehavior::StopAndLazyLoad,
             },
+            false,
             &mut results,
             &definitions,
         )
@@ -448,6 +449,7 @@ fn standing_queries_report_symlinked_skills_without_materializing_symlinked_dire
                     force_included_paths: &[],
                     budget_exceeded_behavior: super::BudgetExceededBehavior::StopAndLazyLoad,
                 },
+                false,
                 &mut results,
                 &definitions,
             )
@@ -491,6 +493,7 @@ fn standing_queries_do_not_report_rules_below_an_unloaded_shallow_directory() {
                 force_included_paths: &[],
                 budget_exceeded_behavior: super::BudgetExceededBehavior::StopAndLazyLoad,
             },
+            false,
             &mut results,
             &StandingQueryDefinitions::default(),
         )
@@ -543,6 +546,7 @@ fn shallow_tree_expands_force_included_skill_branch_only() {
                 force_included_paths: &[std::path::PathBuf::from(".agents/skills")],
                 budget_exceeded_behavior: super::BudgetExceededBehavior::StopAndLazyLoad,
             },
+            false,
             &mut results,
             &definitions,
         )
--- a/crates/repo_metadata/src/local_model.rs
+++ b/crates/repo_metadata/src/local_model.rs
@@ -39,8 +39,8 @@ pub struct RepoContents<'a> {
 use warp_util::standardized_path::StandardizedPath;
 
 use crate::entry::{
-    BudgetExceededBehavior, BuildTreeError, BuildTreeOptions, Entry, FileId, IgnoredPathStrategy,
-    matches_force_included_path,
+    matches_force_included_path, BudgetExceededBehavior, BuildTreeError, BuildTreeOptions, Entry,
+    FileId, IgnoredPathStrategy,
 };
 use crate::repository::Repository;
 use crate::standing_queries::{
--- a/crates/repo_metadata/src/local_model.rs
+++ b/crates/repo_metadata/src/local_model.rs
@@ -543,24 +543,6 @@ impl LocalRepoMetadataModel {
         event: &BulkFilesystemWatcherEvent,
         ctx: &mut ModelContext<Self>,
     ) {
-        // Diagnostics: surface the volume and a sample of the changed paths that
-        // triggered this rebuild, so the source of watcher churn can be traced.
-        let added_count = event.added.len();
-        let modified_count = event.modified.len();
-        let deleted_count = event.deleted.len();
-        let moved_count = event.moved.len();
-        let total_count = added_count + modified_count + deleted_count + moved_count;
-        let sample = event
-            .added_or_updated_iter()
-            .chain(event.deleted.iter())
-            .take(20)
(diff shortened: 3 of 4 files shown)


kevinyang372 (approved):
Thanks for fixing!

PR #13162, opened 2026-06-29
Fix cloud agent follow-ups creating local session

peicodes:
After an oz cloud agent completes, there are two possibilities for follow ups:
- The runner is still alive (for 10 min): follow-ups are sent to `session-sharing-server` and forwarded to the agent
- The runner is torn down (after 10 min): follow-ups are sent to `warp-server` and a new instance is spun up

The way this is determined is based on whether there is an active session: `active_execution_session_id`. This value should be reset when the agent teardown happens.

There are three ways a session can end:
1. `SessionEnded`
2. `ViewerRemoved`
3. `FailedToReconnect`

When we receive `SessionEnded`, we reset `active_execution_session_id`. This is the happy path.

Hypothesis: There are cases where a `SessionEnd` is not received by the client, such as network switching or computer sleeping. These cases can result in 2. `ViewerRemoved` or 3. `FailedToReconnect`.

In these last two cases, we do not reset `active_execution_session_id`. The input falls through to the regular AI input case, which then checks if we're composing a cloud agent.

This check also fails (`matches!(self.status, Status::Composing)`) because the status after a cloud agent finishes is `Status::AgentRunning` and remains that way. That causes us to submit the query as a regular AI input and produces the bug.

What this PR changes:
- Consolidates code for the three ways a session can end
- Ensures all session-end paths reset `active_execution_session_id`


seemeroland (commented):
Refactoring so all these paths share the same teardown logic makes sense to me, but how does this address the original issue? You said if `active_execution_session_id` is not reset we'd be sending to session-sharing-server, but what was causing us to continue the conversation locally?

peicodes:
@seemeroland updated the description - I got confused while writing it

seemeroland (approved):
Thanks for updating!

peicodes:
Tested again manually:
- follow up while runner is live
- follow up after runner is dead

PR #13170, opened 2026-06-29
Prevent system sleep while a terminal command is running

### LINKED ISSUE #9056: The warp terminal does not seem to prevent macOS to put my system on sleep while a task is still running (PreventUserIdleSystemSleep)
### Pre-submit Checks

### Describe the bug

Mac MacBook Pro M5 Pro running macOS 26.4.1 goes into sleep mode while a long running command in the Warp Terminal is still running. I tried the same command in the native Terminal app and there is does not seem to happen, so I assume `PreventUserIdleSystemSleep` is not correctly obsolved.

### To reproduce

Start long running task in Terminal.

### Expected behavior

_No response_

_No response_

### Operating system (OS)

macOS

26.4.1

### Shell Version

_No response_

### Current Warp version

### Regression

### Recent working Warp date

Not sure, but it might be related to a macOS update as well

### Additional context

_No response_

No

None

amriksingh0786:
## Description

The terminal would let macOS idle into sleep mid-command (e.g. `aws configure sso`, a long build, anything that left the user away from the keyboard) because nothing in the terminal lifecycle was holding the user-initiated activity assertion. Agent Mode already does this via `prevent_sleep::prevent_sleep("Agent Mode request in-progress")` in `warp_multi_agent_client`, so the pattern is established.

`TerminalModel` now holds an `Option<prevent_sleep::Guard>` acquired in `start_command_execution` and dropped in the `command_finished` ANSI handler. RAII releases the assertion on drop, so short commands acquire and release within milliseconds without blocking sleep in practice; long-running commands hold the assertion for their full lifetime. The `Guard` type is `Send + Sync`, and on non-mac/non-windows targets it no-ops via the existing `crates/prevent_sleep/src/noop.rs` impl.

Fixes #9056.

## Testing

Added two unit tests in `app/src/terminal/model/terminal_model_tests.rs`:

- `prevent_sleep_guard_is_acquired_on_start_and_released_on_finish` covers the happy-path state transition: clear at construction, `Some` after `start_command_execution`, back to `None` after `command_finished`.
- `prevent_sleep_guard_is_reset_when_start_called_twice` covers a defensive case where the shell ever signals start without an intervening finish: the previous guard is replaced (not leaked), and a single `command_finished` drops the active one.

`cargo fmt`, `cargo clippy -p warp --tests --all-targets -- -D warnings`, and `cargo nextest run -p warp 'terminal::model::terminal_model_tests::prevent_sleep'` all pass cleanly.

Manually verified on macOS: started `sleep 600` in the patched dev build, then from a separate shell:

```
$ pmset -g assertions | grep -c "Terminal command in-progress"
1

$ pmset -g assertions | grep -B 1 -A 2 "Terminal command in-progress"
Listed by owning process:
   pid 88270(warp-oss): [0x00082252000197b7] 00:01:14 PreventUserIdleSystemSleep named: "Terminal command in-progress"
```

Confirms the correct assertion type (`PreventUserIdleSystemSleep`, exactly what the issue title names) is held by the `warp-oss` process for the lifetime of the command. Once the command finishes, the count drops back to 0.

## Agent Mode

- [ ] Warp Agent Mode — This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: The terminal now keeps macOS awake while a foreground command is running, so long jobs like \`aws configure sso\` or slow builds no longer idle the system to sleep mid-execution.

### DIFF
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -168,6 +168,7 @@ persistence.workspace = true
 pin-project.workspace = true
 plist = "1"
 pprof = { workspace = true, optional = true, features = ["protobuf-codec"] }
+prevent_sleep.workspace = true
 prost.workspace = true
 prost-types.workspace = true
 qrcode.workspace = true
--- a/app/src/terminal/model/terminal_model.rs
+++ b/app/src/terminal/model/terminal_model.rs
@@ -520,6 +520,14 @@ pub struct TerminalModel {
     /// bootstrap init scripts. Used to validate DCS hook integrity: hooks
     /// carrying an unrecognized session_id are rejected.
     registered_session_ids: HashSet<SessionId>,
+
+    /// Held while a foreground command is executing in this terminal. Acquired in
+    /// `start_command_execution` and dropped in `command_finished`, which keeps the
+    /// system awake for the lifetime of every command without needing a long-running
+    /// threshold (short commands acquire and release within milliseconds and never
+    /// block sleep in practice). Mirrors the Agent Mode pattern in
+    /// `warp_multi_agent_client`. See issue #9056.
+    running_command_sleep_guard: Option<prevent_sleep::Guard>,
 }
 
 #[derive(Clone, Debug)]
@@ -1089,6 +1097,7 @@ impl TerminalModel {
             // Start mid-way through the u32 range to avoid collisions
             next_kitty_image_id: 2147483647,
             registered_session_ids: HashSet::new(),
+            running_command_sleep_guard: None,
         }
     }
 
@@ -1638,6 +1647,10 @@ impl TerminalModel {
     /// the user's behalf, we consider the active block started.
     pub fn start_command_execution(&mut self) {
         self.block_list.start_active_block();
+        // Keep the system awake for the duration of the command. Replaces any
+        // previous guard (start without a matching finish would otherwise stack).
+        self.running_command_sleep_guard =
+            Some(prevent_sleep::prevent_sleep("Terminal command in-progress"));
     }
 
     pub fn start_command_execution_from_env_var_collection(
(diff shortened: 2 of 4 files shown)


amriksingh0786:
Thanks for catching that. Pushed `3f3c9fe6` on top of `71257831`:

- Reverted the in-function guard acquire from `start_command_execution`. That path now just starts the active block as before.
- Added a new `on_command_started` method on `TerminalModel` that holds the sleep guard.
- The PTY controller calls `on_command_started` only after the real-command match arms (User / AI / SharedSession / EnvVarCollection). `write_end_of_transmission_char` continues to call bare `start_command_execution` for its block-starting side effect and never holds an assertion.
- Added a regression test `prevent_sleep_guard_not_acquired_by_start_command_execution_alone` that locks in the EOT-style invariant, and renamed the prior two tests to point at `on_command_started`.

/oz-review

vorporeal:
Hey @amriksingh0786 - a couple thoughts here:
1. I don't think we should do this unconditionally or by default.  This should be controlled by a setting, which defaults to off, and the user can adjust in the UI.
2. This shouldn't apply to all types of command executions.  For example, for in-band commands (which are things we run in the background to power some of the chips in the input), we don't want those to suppress device sleep.

Can you update the PR accordingly?

vorporeal (commented):
(marking this as reviewed)

amriksingh0786:
Thanks @vorporeal, both make sense. Before I start on the setting + gating, two questions so I aim right the first time:

1. In-band commands: which entry point do those actually flow through? Looking at [`app/src/terminal/writeable_pty/pty_controller.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/terminal/writeable_pty/pty_controller.rs), the four `CommandExecutionSource` variants (User / AI / SharedSession / EnvVarCollection) all funnel through the match block where I added `on_command_started`, so I'd need to know which one (or which upstream caller) corresponds to the chip-generator background commands. Is `EnvVarCollection` the one to skip, or do those come through a different path entirely? A pointer to the code that submits them would save me some spelunking.

2. On the setting shape, would a persistent user setting (Preferences UI + settings crate + default off) be required here, or would a `FeatureFlag` gate be acceptable as a first step? The former is the right long-term shape but adds meaningful UI plumbing; if you'd prefer a feature-flag rollout first so the change lands quickly and you can promote it to a user setting later, I'm happy with either. Let me know the preferred shape.

Once I have direction on those two, I'll push the updated PR.

amriksingh0786:
Gentle nudge @vorporeal, just wanted to check in on the two questions from a few weeks back:

1. Which entry point corresponds to in-band commands (chip generators)? Is it `EnvVarCollection`, or a different path entirely?
2. Would a `FeatureFlag` gate be acceptable as a first step, or does this need a persistent user setting in Preferences up front?

Happy to keep this moving. If you'd rather I just take a shot with `FeatureFlag` + best-guess in-band gating, I can push that as a proposal and iterate on review. Let me know either way.

PR #13177, opened 2026-06-29
 add generalized virtualized viewport element

harryalbert:
## Description

This PR adds a generalized viewport element that takes caller-owned ordered-height index/cursor, and handles rendering only the visible rows in an efficient way (and keeping scroll position stable through resizes). This viewport will serve as the backing element to render the blocklist, without being blocklist-specific (any other caller could provide a sum tree and set of items and get the same efficient rendering for free). This PR also adds some smaller helpers to allow for generalized item clipping and scrollable behavior, and handling for scroll mouse events.

The big things to read in this PR are:
* crates/warpui_core/src/elements/tui/viewported_list.rs
    * New generic virtualized list / viewport.
* crates/warpui_core/src/elements/tui/scrollable.rs
    * New wheel-scroll wrapper.
* crates/warpui_core/src/elements/tui/clipped.rs
    * New vertical row clipping wrapper.
* crates/warpui_core/src/runtime/event_conversion.rs
    * Adds crossterm mouse/wheel conversion into Warp’s shared Event.
* crates/warpui_core/src/elements/tui/event.rs
    * Changes TUI event context from queued “app updates” to view notifications.
* crates/warpui_core/examples/tui_viewport_demo.rs
    * Manual harness demonstrating the pieces together.

## Testing

### Screenshots / Videos

Running the demo added in this PR:
https://www.loom.com/share/dd626bbe4c2e470da975b44c5d80ab15

## Agent Mode


harryalbert:
* **#13291** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13104** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13104?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13177** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13177?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13177?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warpui_core/src/runtime/mod.rs
```diff
@@ -229,10 +229,14 @@ where
...
 
-        for update in event_ctx.take_updates() {
-            update(app);
+        let notified = event_ctx.take_notified();
+        if !notified.is_empty() {
+            let window_id = self.window_id;
+            app.update(|ctx| ctx.notify_tui_event_views(window_id, notified));
         }
-        for action in event_ctx.take_typed_actions() {
+        // Drain reversed to match the GUI window event path, which dispatches
```
kevinyang372: Is this true? Looks like we are draining action in FIFO order https://github.com/warpdotdev/warp/blob/master/crates/warpui_core/src/core/app.rs#L3425

crates/warpui_core/src/core/app/tui.rs
```diff
@@ -10,6 +10,17 @@ use super::{
 use crate::{EntityId, WindowId};
 
 impl AppContext {
+    /// Marks TUI views notified during event dispatch as dirty.
+    pub(crate) fn notify_tui_event_views(
```
kevinyang372: What is this for? It also shouldn't be specific to tui right

crates/warpui_core/src/elements/tui/clipped.rs
```diff
@@ -0,0 +1,127 @@
...
+
+use super::{
+    TuiBuffer, TuiConstraint, TuiElement, TuiEventContext, TuiLayoutContext,
+    TuiPresentationContext, TuiRect, TuiRectExt, TuiSize,
+};
+use crate::{AppContext, Event};
+
+/// A single-child wrapper that paints from `vertical_offset` rows into the child.
+pub struct TuiClipped<E> {
+    child: E,
```
kevinyang372: Shouldn't this just be a `Box<dyn TuiElement>`

crates/warpui_core/src/elements/tui/clipped.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Wraps `child` with no initial offset.
+    pub fn new(child: E) -> Self {
+        Self {
+            child,
+            vertical_offset: 0,
+        }
+    }
+
+    /// Starts painting at `rows` logical rows into the child.
+    pub fn with_vertical_offset(mut self, rows: usize) -> Self {
```
kevinyang372: Why do we have a specific with_vertical_offset argument here? 

Isn't this equivalent to --
```
Container {
      padding_top: X rows,
      child: Clipped { ... }
}
```

crates/warpui_core/src/elements/tui/event.rs
```diff
@@ -3,14 +3,16 @@
 //! The TUI runtime (`crate::runtime`) converts raw crossterm events into the
 //! shared [`Event`](crate::Event) vocabulary (so element/view dispatch is
 //! identical to the GUI), then walks the rendered element tree handing each
-//! element the event plus a [`TuiEventContext`] it can use to queue app
```
kevinyang372, line 6: Hmmm why are we changing the generic app updates signature to be specifically for view notifications?

crates/warpui_core/src/elements/tui/geometry.rs
```diff
@@ -88,6 +90,9 @@ pub trait TuiRectExt: Sized {
 
     /// Splits off the left `width` columns, returning `(left, remainder)`.
     fn split_left(self, width: u16) -> (Self, Self);
+
+    /// Returns whether `position` is inside the rect's half-open cell bounds.
+    fn contains_point(self, position: Vector2F) -> bool;
```
kevinyang372: Why does this take a Vector2F? The unit is wrong

It should be a point with tuple of u16s?

crates/warpui_core/src/elements/tui/scrollable.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// `viewport_height` rows. Returns whether the scroll position changed.
+    fn scroll_by_rows(&mut self, rows: isize, viewport_height: usize) -> bool;
+}
+
+/// Wraps a [`TuiScrollableElement`], capturing wheel events over the child's
+/// area and translating them into scroll requests. Layout, render, cursor, and
+/// inner event dispatch are transparent — only the wheel is intercepted, and
+/// only when the child did not already handle the event.
+pub struct TuiScrollable<E> {
+    child: E,
```
kevinyang372: ditto: we can just make this `Box<dyn TuiScrollableElement>` to avoid the type parameter on the struct?

crates/warpui_core/src/elements/tui/viewported_list.rs
```diff
@@ -0,0 +1,614 @@
...
+    TuiBuffer, TuiConstraint, TuiElement, TuiEventContext, TuiLayoutContext,
+    TuiPresentationContext, TuiRect, TuiScrollableElement, TuiSize,
+};
+use crate::{AppContext, Event};
+
+/// A stable item-relative scroll anchor.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TuiViewportAnchor<ItemId> {
+    pub item_id: ItemId,
+    pub row_offset: usize,
```
kevinyang372: uber nit: Should this be u16 to be consistent with the rest of the app?

crates/warpui_core/src/runtime/event_conversion.rs
```diff
@@ -28,6 +27,83 @@ pub fn crossterm_event_to_warp_event(event: CrosstermEvent) -> Option<Event> {
     }
 }
 
+/// Converts TUI mouse input into cell-coordinate mouse events.
+fn mouse_event_to_warp_event(event: MouseEvent) -> Option<Event> {
```
kevinyang372: Can we implement TryFrom<MouseEvent> for Event here instead?

crates/warpui_core/src/runtime/event_conversion.rs
```diff
@@ -28,6 +27,83 @@ pub fn crossterm_event_to_warp_event(event: CrosstermEvent) -> Option<Event> {
...
+    let position = Vector2F::new(f32::from(event.column), f32::from(event.row));
```
kevinyang372: It's kinda nasty that we have to convert from a u16 based event position to vector2F and then handle that in each element's level

What's the scope if we would to make a TUIEvent with its unique typing instead?

crates/warpui_core/src/elements/tui/viewported_list.rs
```diff
@@ -0,0 +1,614 @@
...
+
+use super::{
+    TuiBuffer, TuiConstraint, TuiElement, TuiEventContext, TuiLayoutContext,
+    TuiPresentationContext, TuiRect, TuiScrollableElement, TuiSize,
+};
+use crate::{AppContext, Event};
+
+/// A stable item-relative scroll anchor.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TuiViewportAnchor<ItemId> {
```
kevinyang372: Why do we have a generic trait of ItemId here? Should this element define a ViewportItemId type and just have the callers use that instead (it can be monotonically increasing to guarantee uniqueness)?

crates/warpui_core/src/elements/tui/viewported_list.rs
```diff
@@ -0,0 +1,614 @@
...
+/// Adapts a caller-owned ordered-height index for [`TuiViewportedList`].
+pub trait TuiViewportIndex {
+    type ItemId: Clone + Eq;
+    type Item;
+
+    /// Opens a scoped cursor and releases all borrowed backing state when `f` returns.
+    fn with_cursor<R>(
+        &self,
+        position: TuiViewportIndexPosition<'_, Self::ItemId>,
+        f: impl FnOnce(&mut dyn TuiViewportCursor<ItemId = Self::ItemId, Item = Self::Item>) -> R,
```
kevinyang372: What is the semantics of `with_cursor`?

It's not clear to me what it is used for with the generic `f` parameter and `R` return type

crates/warpui_core/src/core/app/tui.rs
harryalbert: Chatted about this offline, but this is so that we can ctx.notify from event ctx (specifically from scroll wheel events). Doesn't need to be TUI specific though

crates/warpui_core/src/elements/tui/clipped.rs
harryalbert: yeah good call

crates/warpui_core/src/elements/tui/clipped.rs
harryalbert: misspoke when we were talking in person – it's actually the opposite. I renamed the field to `hidden_top_rows` to make it clearer, but what this field does is say like we should only render from rows x onwards for the given clipped child. 

I also think it makes more sense for this state to live on the clipped element itself instead of on the viewport, so I made that change as well

crates/warpui_core/src/elements/tui/geometry.rs
harryalbert: you're right, good call

crates/warpui_core/src/runtime/event_conversion.rs
harryalbert: yessir

crates/warpui_core/src/runtime/event_conversion.rs
harryalbert: I actually don't think it's too bad. Having an agent give it a shot but seems reasonable

harryalbert: yeah converting to TUI event was pretty well scoped. I'll push

crates/warpui_core/src/elements/tui/event.rs
harryalbert: yeah good call, we shouldn't

crates/warpui_core/src/runtime/mod.rs
harryalbert: hmm, I'm seeing that we are for non keybound events specifically:
https://github.com/warpdotdev/warp/blob/99c98bffd4f2c70db6dd6a1603961d9ac84b087c/crates/warpui_core/src/core/app.rs#L3876

crates/warpui_core/src/elements/tui/viewported_list.rs
```diff
@@ -0,0 +1,339 @@
...
+        matches!(*self.0.borrow(), TuiViewportPosition::End)
+    }
+}
+
+/// A content-space viewport window.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct TuiViewportWindow {
+    pub scroll_top: usize,
+    pub viewport_height: usize,
+    pub viewport_width: u16,
```
kevinyang372: Why do we need width here? (I thought the viewport is vertical only)

crates/warpui_core/src/elements/tui/viewported_list.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// A child element visible in a content-space viewport.
+pub struct TuiVisibleViewportItem {
+    pub origin_y: usize,
```
kevinyang372, line 70: Should we use u16 consistently? I am seeing both u16 and usize across this file

crates/warpui_core/src/runtime/mod.rs
```diff
@@ -140,10 +135,12 @@ impl<T: TuiView, R: TuiTerminal> TuiScreen<T, R> {
...
 
-        for update in event_ctx.take_updates() {
-            update(ctx);
+        let app_updates = event_ctx.take_app_updates();
+        for view_id in app_updates {
+            ctx.notify_view_observers(self.window_id, view_id);
         }
-        for action in event_ctx.take_typed_actions() {
+
+        for action in event_ctx.take_typed_actions().into_iter().rev() {
```
kevinyang372: Looking through the code -- I don't think rev actually matters here

crates/warpui_core/src/elements/tui/clipped.rs
```diff
@@ -0,0 +1,134 @@
...
+            child,
+            hidden_top_rows: 0,
+        }
+    }
+
+    /// Hides `rows` logical rows above the clipped viewport.
+    ///
+    /// This clips child rows above the viewport; it is not equivalent
+    /// to adding top padding, which would move the child down instead.
+    pub fn with_hidden_top_rows(mut self, rows: usize) -> Self {
```
kevinyang372: I still don't quite understand this API

crates/warpui_core/src/elements/tui/event.rs
```diff
@@ -29,11 +110,9 @@ pub struct TuiEventDispatchResult {
     pub handled: bool,
 }
 
-type TuiAppUpdate = Box<dyn FnOnce(&mut AppContext)>;
-
 #[derive(Default)]
 pub struct TuiEventContext {
-    updates: Vec<TuiAppUpdate>,
+    app_updates: HashSet<EntityId>,
```
kevinyang372: I would just name this `notified` similar to GUI

kevinyang372 (approved):

crates/warpui_core/src/elements/tui/clipped.rs
harryalbert: chatted about this offline, but renamed this field to `viewport_origin_y` and added a diagram to explain what it means

crates/warpui_core/src/elements/tui/viewported_list.rs
harryalbert: we don't need width as a field. We do need to pass width to the backing sum-tree so that it knows how tall each element will be (w/ things like text overflow) but that can just be through a fn param

crates/warpui_core/src/elements/tui/viewported_list.rs
harryalbert: chatted offline, but some of the fields that are not constrained by the size of the tui need to use usize, but we can change a few of the fields back to u16

PR #13185, opened 2026-06-29
Render a proper error state for web search/fetch inline actions

advait-m:
Related server PR: https://github.com/warpdotdev/warp-server/pull/12188

## Description
Renders a proper error state for the web search / web fetch inline actions.

**What:** When a web search (`exa_web_search`) or web page fetch (`fetch_web_pages`) fails, the inline action previously rendered identically to an empty success — "Searched the web / 0 URLs / No URLs found" (and "Fetched 0 web pages / No URLs fetched"). This made a genuine failure indistinguishable from a search/fetch that legitimately returned nothing (there was a `TODO(advait)` placeholder for it).

**Why:** Paired with a server-side fix (warp-server #12188) that stops leaving failed Exa tool calls dangling and instead marks the web search/fetch message as `error`. The client now needs to actually render that error state.

**How:**
- `web_search.rs` / `web_fetch.rs`: the `Error` status now renders the house failure style — the red `failed_icon` plus a single-line status header ("Web search failed" / "Web page fetch failed") instead of a fake empty-success.
- `WebSearchView::set_status` preserves the query from the prior `Searching`/`Success` state (the proto `error` variant carries no query), so a live failure can show "Web search failed for "{query}"". On history reload the query isn't available, so it falls back to the generic text.
- Renamed the shared `render_loading_header` → `render_status_header` (it now backs loading, cancelled, and error states; the icon conveys the status). Updated all call sites (`search_results_common`, `search_codebase`, `web_search`, `web_fetch`). The unrelated `AgentManagementView::render_loading_header` is left as-is.

## Testing
- `cargo fmt` and `cargo clippy -p warp --lib --features local_fs -- -D warnings` pass.
- Manually exercised by forcing the Exa failure path locally and confirming the inline action renders the error header instead of "0 URLs / No URLs found".

### Screenshots / Videos
image

CHANGELOG-BUG-FIX: Web search and web page fetch now show a clear failed state instead of appearing as an empty "0 results" when the search/fetch errors.

### DIFF
--- a/app/src/ai/blocklist/inline_action/web_search.rs
+++ b/app/src/ai/blocklist/inline_action/web_search.rs
@@ -5,7 +5,7 @@ use warpui::{AppContext, Entity, SingletonEntity, TypedActionView, View, ViewCon
 use super::search_results_common::{
     render_collapsible_search_results, CollapsibleSearchResultsState,
 };
-use crate::ai::agent::icons::yellow_running_icon;
+use crate::ai::agent::icons::{failed_icon, yellow_running_icon};
 use crate::ai::agent::WebSearchStatus;
 use crate::ai::blocklist::block::view_impl::WithContentItemSpacing;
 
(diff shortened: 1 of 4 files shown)


app/src/ai/blocklist/inline_action/web_search.rs
```diff
@@ -32,7 +32,21 @@ impl WebSearchView {
     }
 
     pub fn set_status(&mut self, status: &WebSearchStatus) {
-        self.status = status.clone();
+        // The proto error status carries no query; preserve the prior one for the header.
+        self.status = match status {
```
captainsafia, line 36: What are the implications of this? Do we get a less detailed error in the UI? I suppose that's OK if the actual query doesn't impact the success/failure rate.

captainsafia (approved):

app/src/ai/blocklist/inline_action/web_search.rs
advait-m: yep the actual query doesn't impact success/failure

didn't want to change protos for now, we basically just fall back to an error state without the query, which should be fine

PR #13186, opened 2026-06-29
Hook up input view with TUI app

kevinyang372:
## Description

Hooks the editor-backed input view into the headless `warp-tui` front-end with an invalidation-driven render/input driver, gated behind login. This is rebased onto master's TUI work: it keeps the device-login flow but replaces the earlier one-shot "prompt streaming to stdout" example with the interactive input box.

### Behavior
- `warp-tui` boots the real headless app (full `initialize_app`, so auth / `Appearance` / settings exist).
- While the user is signing in, the TUI shows a centered **"Sign in to continue"** placeholder, including the device-login verification URL + code (the alternate screen hides stdout, and the browser still auto-opens). Keystrokes are gated off in this state.
- Once authenticated, it renders a single bordered input box docked at the bottom that grows with newlines up to six text rows, then scrolls.

### Core infra
- **Invalidation-based redraw loop (`spawn_tui_driver` + `TuiDriverHandle`)** in `crates/warpui_core/src/runtime/mod.rs`: rendering is driven by an `on_window_invalidated` callback inside the core's normal `flush_effects` pass, so any `notify()` (input handler, model/async update, resize) schedules a repaint — mirroring the GUI's invalidate→redraw flow. Input is read on a background thread and only *dispatched* on the foreground executor; `Ctrl-C` quits. The blocking `TuiRuntime::run_until` remains as a dev/test-only harness.
- **`TuiTerminalGuard`** — an RAII guard owning raw mode + the alternate screen, split out of `CrosstermTerminal` so the terminal is deterministically restored on every exit path (panic, error, teardown). Held by `TuiDriverHandle`.
- **Shared `TuiScreen`** — presenter + renderer + host terminal, shared by both the driver and the dev/test harness; behind `Rc<RefCell<_>>` in the driver so the repaint never collides with a borrow held by input dispatch.
- **Core plumbing** — `AppContext::weak_app()` / `WeakApp`, and the `TuiEventContext` update closure is now `FnOnce(&mut AppContext)`.

### App wiring + auth/login gate
- `warp::run_tui(mount: TuiMountFn)` + `LaunchMode::Tui { mount }`; after `initialize_app`, `app/src/tui/mod.rs::init` runs the device-login flow and mounts the TUI immediately.
- `TuiLoginModel` / `TuiLoginPhase` (`AwaitingLogin { verification_uri, user_code } | Failed { message } | LoggedIn`), exposed from the `warp` crate root; `RootTuiView` (in `warp_tui`) observes it to render the placeholder vs. the input box.
- `warp_tui::run()` dispatches worker re-execs via `run_tui_worker_if_requested()` before starting the TUI, preventing a fork bomb when the binary is re-exec'd as a worker (e.g. the terminal server).

### Supporting UI
- New TUI layout elements: `TuiColumn` flex (`flex_child` / `child`) and `TuiConstrainedBox` (caps height to N rows), used by `RootTuiView`.
- Fixed `TuiContainer::cursor_position` to offset the reported cursor by the border/padding inset.
- Input box renders every visible row uniformly (no active-line dimming).

### Removed
- Master's non-interactive prompt-stream example: `prompt_stream`, `conversation_model`, `conversation_selection`, `args` (+ their tests) and `app/src/tui_export.rs`, plus the demo-only deps (`parking_lot`, `pathfinder_geometry`).

## Linked Issue

N/A — internal infrastructure for the not-yet-shipped headless TUI front-end.

## Testing

- `cargo check` green for both `warp` (default / non-tui) and `warp_tui` (tui); `cargo fmt` + `cargo clippy -D warnings` clean.
- Tests: `warp_tui` 16/16 pass (incl. the worker-dispatch regression test); `warpui_core --features tui` 365/365 pass.
- Live validation (terminal-use over a real PTY):
  - Logged-in path: input box renders; typing, multi-line growth + scroll past six rows, submit-clears, and a clean `Ctrl-C` exit (terminal restored).
  - Logged-out path (temporarily forced): centered "Sign in to continue" placeholder with the verification URL/code, no input box, keystrokes gated off, clean `Ctrl-C` exit.
  - Note: the real end-to-end device login (browser approval → `AuthComplete` → input box) is covered by code review + the forced-state check rather than a full live login.

## Agent Mode

CHANGELOG-NONE


kevinyang372:
* **#13186** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13186?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13186?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/lib.rs
```diff
@@ -849,7 +854,19 @@ pub fn run_tui_worker_if_requested() -> Option<Result<()>> {
...
-/// Runs the app (or CLI / daemon).
+/// The headless TUI front-end's mount callback, carried by [`LaunchMode::Tui`].
+/// Supplied to [`run_tui`] by the `warp_tui` crate; it runs after
+/// `initialize_app` to build the root TUI view and start the TUI driver.
+pub type TuiMountFn = Box<dyn FnOnce(&mut warpui::AppContext)>;
+
+/// The TUI login-gate state, exposed so the `warp_tui` root view can observe it
+/// and render the login placeholder vs. the input UI.
+#[cfg(feature = "tui")]
+pub use crate::tui::{TuiLoginModel, TuiLoginPhase};
```
harryalbert: nit: this import should be at the top of the file

crates/warp_tui/src/args.rs
harryalbert, line 1: didn't we decide that we probably want to use the TUI for oz prompting (instead of the oz cli)? Assuming we do, it might make sense to leave this in. I also think generally the TUI will likely need some arg-acceptance capability

crates/warp_tui/src/input/view.rs
```diff
@@ -667,15 +667,11 @@ impl TuiInputElement {
             }
         }
 
-        // Assemble the column of visible rows.
-        let dim = TuiStyle::default().add_modifier(Modifier::DIM);
+        // Assemble the column of visible rows. Every row uses the default style:
+        // there is no active-line highlighting, so the cursor row is not
```
harryalbert: mega nit: this comment seems unnecessary

crates/warp_tui/src/root_view.rs
```diff
@@ -0,0 +1,185 @@
...
+    pub fn new(ctx: &mut ViewContext<Self>) -> Self {
+        let model = ctx.add_model(|ctx| CodeEditorModel::new_tui(INITIAL_TERMINAL_WIDTH, ctx));
+        let input = ctx.add_typed_action_tui_view(move |ctx| TuiInputView::new(model, ctx));
+        ctx.subscribe_to_view(&input, Self::handle_input_event);
+        Self { input }
+    }
+
+    fn handle_input_event(
+        &mut self,
+        _input: ViewHandle<TuiInputView>,
```
harryalbert: nit: if these args are unused, are they necessary? it also seems like this fn doesn't really do anything right now

crates/warp_tui/src/root_view.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+impl TypedActionView for RootTuiView {
+    // The root handles no typed actions itself; editing actions are handled by
+    // the input view, and Ctrl-C quit is handled by the TUI driver.
+    type Action = ();
+}
+
+/// A centered block, vertically docked to the middle via flex spacers.
+fn centered(content: TuiColumn) -> Box<dyn TuiElement> {
```
harryalbert: feels like this should probably live in some TUI ui helper mod

crates/warpui_core/src/runtime/mod.rs
```diff
@@ -286,6 +317,151 @@ impl TuiTerminal for CrosstermTerminal {
...
+    pub fn enter() -> io::Result<Self> {
+        Ok(Self(RawModeGuard::enter(CrosstermModeControl)?))
+    }
+}
+
+/// Keeps a headless TUI session alive. Dropping it tears the session down: it
+/// restores the terminal (via the guard) and ends the draw loop + input reader.
+/// Store it for the lifetime of the app (e.g. in a singleton model) so the
+/// session lives as long as the app does.
+pub struct TuiDriverHandle {
```
harryalbert, line 347: is this following prior art from the GUI? it seems like a bit of a smell to me but I'm not very familiar w/ the session initialization code so defer to you

harryalbert (commented):
The main question I have is, do we actually want to remove args and the `--prompt` capability? I feel like the TUI will likely need to accept some optional args at some point and might also need prompt streaming outside of the main app

harryalbert (approved):
Spoke offline, disregard my args comment

crates/warpui_core/src/runtime/mod.rs
kevinyang372: This is a common design pattern -- we need to keep hold to these handles and they have logic to auto-clean up resources on deallocation (aka app termination)

PR #13191, opened 2026-06-29
Fix CliAgentUserQuery / LRC snapshot race: queue user query until snapshot fires

cephalonaut:
## What

When a user sends a query while an agent-requested `run_shell_command` action is still pending (snapshot not yet fired), the client was sending a `CliAgentUserQuery`. The server both auto-cancelled the `run_shell_command` AND spawned a CLI subagent. When the CLI subagent completed it added a synthetic `run_shell_command` result, producing two `tool_result` blocks for the same `tool_use` ID. Providers reject this causing retries and potential fatal errors.

Fix: `maybe_queue_input_for_in_progress_conversation` in `Input` detects the pending action via `BlocklistAIActionModel::is_shell_command_action_pending` and queues the query as `PendingLrcAutoQueue` instead of sending immediately. This goes through the normal queuing path so the input buffer is cleared and attachments are captured correctly.

`PendingLrcAutoQueue` rows are locked (send-now button disabled) until the snapshot fires. When `FinishedAction` fires for the completed action, `QueuedQueryModel::unlock_pending_lrc_rows` transitions the row to `LrcAutoQueue`, enabling auto-fire when the command eventually completes — identical to the behavior of queries submitted after the snapshot fires. On cancellation, `remove_pending_lrc_rows` cleans up locked rows.

New additions:
- `QueuedQueryOrigin::PendingLrcAutoQueue`: locked queue origin for pre-snapshot queries
- `QueuedQueryModel::unlock_pending_lrc_rows` / `remove_pending_lrc_rows`: lifecycle management for pending rows
- `BlocklistAIActionModel::is_shell_command_action_pending`: predicate used by `Input` to detect the pre-snapshot window

Demo: https://www.loom.com/share/e59301a9039b4f07bea98b44afbf648c

## Why

Fixes the bug: "Found multiple tool_result blocks with id" 400 error from Vertex/Anthropic, and the associated output flicker. Also ensures consistent behavior — a query submitted just before or just after the 2s snapshot threshold is handled identically (queued until command completes).

## Agent Mode

## Testing

Manually increased MAX_WAIT_DURATION to 30s to create a reproducible pre-snapshot window. Verified:
- Query submitted during the window is queued as PendingLrcAutoQueue with send-now disabled
- After 30s the snapshot fires, row transitions to LrcAutoQueue and fires when the command completes
- Queries submitted after the snapshot continue to use the CLI subagent path normally

---

_Conversation: https://staging.warp.dev/conversation/7a6d98f6-beef-4bad-b315-673800590208_
_Run: https://oz.staging.warp.dev/runs/019f13eb-35a7-7946-92e1-56742405408b_

### DIFF
--- a/app/src/ai/blocklist/action_model/execute/shell_command.rs
+++ b/app/src/ai/blocklist/action_model/execute/shell_command.rs
@@ -49,7 +49,7 @@ pub struct ShellCommandExecutor {
 }
 
 impl ShellCommandExecutor {
-    pub const MAX_WAIT_DURATION: Duration = Duration::from_secs(2);
+    pub const MAX_WAIT_DURATION: Duration = Duration::from_secs(30); // TEMP: bumped for manual testing, revert before merge
     /// Maximum delay we will honor for any agent-requested wait. Applies both  
     /// to finite `ShellCommandDelay::Duration` requests and to  
     /// `ShellCommandDelay::OnCompletion`, which would otherwise wait indefinitely.  
--- a/app/src/ai/blocklist/action_model/execute/shell_command.rs
+++ b/app/src/ai/blocklist/action_model/execute/shell_command.rs
@@ -49,7 +49,7 @@ pub struct ShellCommandExecutor {
 }
 
 impl ShellCommandExecutor {
-    pub const MAX_WAIT_DURATION: Duration = Duration::from_secs(30); // TEMP: bumped for manual testing, revert before merge
+    pub const MAX_WAIT_DURATION: Duration = Duration::from_secs(2);
     /// Maximum delay we will honor for any agent-requested wait. Applies both  
     /// to finite `ShellCommandDelay::Duration` requests and to  
     /// `ShellCommandDelay::OnCompletion`, which would otherwise wait indefinitely.  
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -545,6 +545,12 @@ impl BlocklistAIController {
                 return;
             }
             me.send_follow_up_for_conversation(*conversation_id, ctx);
+            // Unlock any query queued during the pre-snapshot window now that the snapshot has
+            // been sent. The row becomes a regular LrcAutoQueue row and fires when the command
+            // completes via the existing send_lrc_queued_prompts path.
+            QueuedQueryModel::handle(ctx).update(ctx, |model, ctx| {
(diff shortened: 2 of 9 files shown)


app/src/ai/blocklist/controller.rs
```diff
@@ -1167,6 +1167,35 @@ impl BlocklistAIController {
...
+            if let Some(action_id) = pending_action_id {
+                if self
+                    .action_model
+                    .as_ref(ctx)
+                    .is_shell_command_action_pending(&action_id, conversation_id)
+                {
+                    QueuedQueryModel::handle(ctx).update(ctx, |model, ctx| {
+                        model.append(
+                            conversation_id,
+                            QueuedQuery::new(query, QueuedQueryOrigin::LrcAutoQueue),
```
MaggieShan: I believe this skips the existing process of clearing the input buffer after queuing and sending the prompt + skips attaching any pending attachments

MaggieShan (commented):
Seeing a couple issues here: 
1. the input is not cleared after queuing the prompt
2. send now causes the input to disappear and gets you into a weird state 

https://www.loom.com/share/1efb08265c9e48bd869e3b9c94e688d5

app/src/ai/blocklist/controller.rs
cephalonaut: Yeah, sorry - started changing the implementation of this after talking to Suraj. Needs more work.

cephalonaut: I think I've already fixed the send issue, but will take a look at clearing and attachments.

app/src/ai/blocklist/action_model/execute/shell_command.rs
```diff
@@ -49,7 +49,7 @@ pub struct ShellCommandExecutor {
 }
 
 impl ShellCommandExecutor {
-    pub const MAX_WAIT_DURATION: Duration = Duration::from_secs(2);
+    pub const MAX_WAIT_DURATION: Duration = Duration::from_secs(30); // TEMP: bumped for manual testing, revert before merge
```
MaggieShan: temp?

MaggieShan (approved):
lgtm!

app/src/ai/blocklist/action_model/execute/shell_command.rs
cephalonaut: yes, will revert

PR #13227, opened 2026-06-30
[VA] Wire up computer-use video recording

vkodithala:
## Description

This is the second PR in the computer-use video-recording stack.

It wires the recording substrate from #13044 into agent execution: StartRecording/StopRecording API conversions, conversation YAML/redaction support, action/result model plumbing, blocklist executors, a process-local `RecordingController`, artifact upload, cancellation cleanup, and dogfood feature-flag enablement.

For context, the downstack substrate PR makes recording possible, but the agent runtime still needs a way to request bounded recordings, preserve the active recording session, stop/finalize it, and surface the resulting MP4 artifact back to the conversation and persist to GCS.

Tl;dr on the implementation: StartRecording creates a bounded recording through the singleton controller and records the applied limits/metadata in the action result. StopRecording validates upload preconditions before consuming the active session, finalizes the recorder, uploads the MP4 as an agent artifact, and reports whether the recording completed or stopped early. Cancellation handling aborts an in-flight start reservation so the controller does not get stuck.

## Stack context

- This PR is stacked on #13044 (`varoon/va-video-recording-client`), which adds only the `crates/computer_use` recorder substrate and Linux `ffmpeg` implementation.
- This PR owns app-layer/agent execution wiring and should be reviewed after or alongside #13044.
- CI/validation for this PR depends on generated `warp_multi_agent_api` types that include StartRecording/StopRecording. Locally, validation used the matching `warp-proto-apis` branch via an uncommitted top-level `Cargo.toml` patch.

## Linked Issue

N/A

## Testing

Manually tested end-to-end, here's a [demo](https://www.loom.com/share/59ea36e89f9b42479a48d9ca0d1900b2).

- `./script/format`
- `cargo check -p computer_use -p ai -p warp`
- `cargo clippy -p computer_use -p ai -p warp --all-targets -- -D warnings`

Validation note: these commands currently require the local/generated `warp_multi_agent_api` proto patch that includes StartRecording/StopRecording types. This execution layer does not compile against the current `master`\-generated API crate until that proto update lands.

- [x] I have manually tested my changes locally with `./script/run` (not yet; local macOS cannot exercise the Linux X11 recorder path end to end)

## Known issues / follow-ons before launch

These are the concrete v0 follow-ons from the video-recording launch review. They are not all intended to land in this PR, but they should be tracked before launching the feature broadly.

- Similarly to what we allow for screenshots, add video output scaling, likely with a runtime-controlled `RecordingConfig` target such as `max_long_edge_px` or `target_height_px`, and use an `ffmpeg` scale filter so output is roughly 720p. Return captured/applied dimensions in recording metadata.
- Add MP4 structural validation after stop/finalization (probably just parsed from `ffmpeg` outputs on SIGINT: parseable container, video stream present, non-zero duration, non-zero dimensions, and expected-enough codec/container for playback. Similar to what Cursor does. Mentioned in my doc.
- Improve termination reasons beyond the current coarse early-stop status: distinguish agent stop, duration limit, size limit, unexpected `ffmpeg` exit, and finalization timeout/force-kill.
- Add dedicated StartRecording/StopRecording UI cards that show recording parameters, limits, duration/size, completion state, and artifact link rather than opaque generic tool-call output.
- Capture computer-use action logs with monotonic timecodes while recording is active, including mouse move/click/wheel, key presses, and redacted/summarized typed text according to existing policy.
- Add basic overlay rendering for captured action logs, such as key/action captions, click indicators, and timestamps.
- Add lifecycle finalization/upload in a separate PR for agent-run end/cancel/timeout; explicit StopRecording after lifecycle finalization should be idempotent or report the already-finalized artifact.

## Agent Mode

CHANGELOG-NONE

Co-Authored-By: Oz [oz-agent@warp.dev](mailto:oz-agent@warp.dev) d

### DIFF
--- a/app/src/ai/blocklist/action_model/execute/stop_recording.rs
+++ b/app/src/ai/blocklist/action_model/execute/stop_recording.rs
@@ -2,7 +2,9 @@
 use ai::agent::action_result::{RecordingStopped, StopRecordingResult};
 use futures::future::BoxFuture;
 use futures::FutureExt;
-use warpui::{Entity, ModelContext, SingletonEntity};
+#[cfg(not(target_family = "wasm"))]
+use warpui::SingletonEntity;
+use warpui::{Entity, ModelContext};
 
 use super::{ActionExecution, AnyActionExecution, ExecuteActionInput, PreprocessActionInput};
 use crate::ai::agent::AIAgentActionType;
--- a/app/src/ai/blocklist/action_model/execute/stop_recording.rs
+++ b/app/src/ai/blocklist/action_model/execute/stop_recording.rs
@@ -1,0 +1,168 @@
+#[cfg(not(target_family = "wasm"))]
+use ai::agent::action_result::{RecordingStopped, StopRecordingResult};
+use futures::future::BoxFuture;
+use futures::FutureExt;
+use warpui::{Entity, ModelContext, SingletonEntity};
+
+use super::{ActionExecution, AnyActionExecution, ExecuteActionInput, PreprocessActionInput};
+use crate::ai::agent::AIAgentActionType;
+#[cfg(not(target_family = "wasm"))]
+use crate::{
+    ai::{
+        agent::AIAgentActionResultType,
+        agent_sdk::artifact_upload::{FileArtifactUploadRequest, FileArtifactUploader},
+        blocklist::action_model::recording_controller::{
+            RecordingController, StopRecordingControllerError,
+        },
+        blocklist::BlocklistAIHistoryModel,
+    },
+    server::server_api::ServerApiProvider,
+};
+
+#[cfg(not(target_family = "wasm"))]
+fn format_stop_recording_error(err: &anyhow::Error) -> String {
+    let error_chain = format!("{err:#}");
+    if error_chain != err.to_string() {
+        format!("Recording upload failed: {error_chain}")
+    } else {
+        error_chain
+    }
+}
+
+pub struct StopRecordingExecutor;
+
+impl StopRecordingExecutor {
+    pub fn new() -> Self {
+        Self
+    }
+
+    #[cfg_attr(target_family = "wasm", allow(unused_variables))]
+    pub(super) fn should_autoexecute(
+        &self,
+        input: ExecuteActionInput,
+        _ctx: &mut ModelContext<Self>,
+    ) -> bool {
+        let ExecuteActionInput { action, .. } = input;
+        matches!(action.action, AIAgentActionType::StopRecording { .. })
+            && warp_core::features::FeatureFlag::VideoRecording.is_enabled()
+    }
+
+    #[cfg_attr(target_family = "wasm", allow(unused_variables))]
+    pub(super) fn execute(
+        &mut self,
+        input: ExecuteActionInput,
+        ctx: &mut ModelContext<Self>,
+    ) -> AnyActionExecution {
+        #[cfg(target_family = "wasm")]
+        {
+            ActionExecution::<()>::InvalidAction.into()
+        }
+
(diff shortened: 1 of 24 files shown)


vkodithala:
* **#13227** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13227?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13227?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13044** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13044?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+/// Tracks recordings keyed by id and enforces one active recording per client runtime.
+///
+/// NOTE: Ambient agent environment setup currently provides one Xvfb instance
+/// per runtime, which exposes a single display to the recorder. If that changes,
+/// this controller should key active recordings by display.
+pub struct RecordingController {
+    sessions: HashMap<String, RecordingSession>,
```
danielpeng2: Why use a HashMap here if we can only have one active recording, would an `Option` not suffice?

app/src/ai/blocklist/action_model/execute/stop_recording.rs
```diff
@@ -0,0 +1,168 @@
...
+        #[cfg(not(target_family = "wasm"))]
+        {
+            let ExecuteActionInput {
+                action,
+                conversation_id,
+            } = input;
+            let AIAgentActionType::StopRecording { recording_id } = &action.action else {
+                return ActionExecution::<()>::InvalidAction.into();
+            };
+            let recording_id = recording_id.clone();
```
danielpeng2: nit: is this clone necessary? I only see us use the recording id in one spot.

crates/ai/src/agent/action_result/mod.rs
```diff
@@ -1197,6 +1213,71 @@ impl Display for RequestComputerUseResult {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum StopRecordingResult {
+    Success(RecordingStopped),
+    Error(String),
+    Cancelled,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RecordingStopped {
+    pub artifact_uid: String,
```
danielpeng2, line 1260: ooc what do we use this `artifact_uid` for on the server? we should make sure we've thought through the security implications of having a client-generated uid here. I'd be curious if other artifacts follow this same pattern of generating a uid in the client then sending it back up to the server.

danielpeng2 (approved):

app/src/ai/blocklist/action_model/execute/stop_recording.rs
vkodithala: Yup, you're right. Removed in latest.

crates/ai/src/agent/action_result/mod.rs
vkodithala: So, my understanding here is that this _is_ a server-generated uid. When a recording finishes, the client calls `upload_with_association`, which is resolved on the server, and returns uploaded artifact metadata, including a uid.  See [here](https://github.com/warpdotdev/warp/blob/varoon/va-video-recording-execution/app/src/ai/agent_sdk/artifact_upload.rs#L105-L132). This seems to be the general way artifacts are uploaded, and the uid should be accurate.

Lmk if I'm missing anything big here.

PR #13232, opened 2026-06-30
Add warp://settings deeplink entrypoints (open, search, scroll-to-widget)

szgupta:
Demo: https://www.loom.com/share/d26cdf004faa4aee9c128f77dc0b2af8

## Description
Adds a `warp://settings` deep link family so settings can be opened directly from a URL:
- `warp://settings` - opens a settings tab on the default page.
- `warp://settings?q=<query>` - opens settings with the search bar pre-filled and the page list filtered.
- `warp://settings?widget=<widget_id>` - opens settings scrolled to (and highlighting) a specific widget.

**Why:** lets external surfaces (docs, in-app links, support flows) deep-link users straight to the relevant setting.

**How:** the existing `UriHost::Settings` handler now treats an empty trailing path segment as "no sub-page" and routes the bare host plus the `q`/`widget` query params (precedence: `widget` > `q` > existing simple sub-pages > bare default). A new `OpenSettingsArgs` enum maps to the existing workspace actions (`ShowSettings`, `ShowSettingsPageWithSearch`, `ScrollToSettingsWidget`) via two new `root_view` actions. `?widget=` resolves through an allowlist (`settings_widget_deeplink_target`) of stable slugs to `(SettingsSection, &'static str)` so internal Rust type names aren't exposed as the public URL contract; seeded with `global_hotkey`, `cli_agents`, and `custom_router`. Also fixed `open_settings_pane` so a `?q=` search applies even when a settings tab is already open.

## Linked Issue
N/A - no linked issue.

## Testing
Added unit tests in `app/src/uri/uri_tests.rs`:
- `parse_settings_search_query` - present / empty / URL-encoded / missing `q`
- `settings_widget_deeplink_target` - known slugs (incl. `custom_router`) and unknown/empty
- `settings_section_for_simple_subpage` - regression for existing sub-pages

Ran `./script/format` and `cargo clippy -p warp --all-targets --tests -- -D warnings` (clean), plus the new unit tests.

> Not manually run via `./script/run`; covered by unit tests. Routing-only change with no new UI layout, so no screenshots.

## Agent Mode

## Warp artifacts
- Conversation: https://app.warp.dev/conversation/3d208974-67e1-44da-8157-12a7e7e2ef55
- Plan: https://app.warp.dev/drive/notebook/1soy0YOddbTETZH4fg3IV3

CHANGELOG-IMPROVEMENT: Added `warp://settings` deep links - open settings, pre-fill the search bar with `?q=`, or jump to a specific setting with `?widget=`.

### DIFF
--- a/app/src/root_view.rs
+++ b/app/src/root_view.rs
@@ -2842,9 +2842,6 @@ impl RootView {
         true
     }
 
-    /// Handles a `warp://settings` deeplink (bare, `?q=`, or `?widget=`) in an
-    /// existing window, mapping the [`OpenSettingsArgs`] variant to its
-    /// workspace action.
     pub fn open_settings_in_existing_window(
         &mut self,
         args: &OpenSettingsArgs,
--- a/app/src/root_view.rs
+++ b/app/src/root_view.rs
@@ -1004,8 +1004,6 @@ fn workspace_action_for_open_settings(args: &OpenSettingsArgs) -> WorkspaceActio
     }
 }
 
-/// Opens a new window for a `warp://settings` deeplink (bare, `?q=`, or
-/// `?widget=`), mapping the [`OpenSettingsArgs`] variant to its workspace action.
 fn open_settings_in_new_window(args: &OpenSettingsArgs, ctx: &mut AppContext) {
     let action = workspace_action_for_open_settings(args);
     let root_handle = open_new_window_get_handles(None, ctx).1;
--- a/app/src/root_view.rs
+++ b/app/src/root_view.rs
@@ -86,7 +86,7 @@ use crate::terminal::shell::ShellType;
 use crate::terminal::view::{cell_size_and_padding, TerminalAction};
 use crate::themes::onboarding_theme_picker_themes;
 use crate::themes::theme::{AnsiColorIdentifier, Blend, Fill, ThemeKind, WarpThemeConfig};
-use crate::uri::OpenMCPSettingsArgs;
+use crate::uri::{OpenMCPSettingsArgs, OpenSettingsArgs};
 use crate::util::bindings::{self, is_binding_pty_compliant};
 use crate::util::traffic_lights::{traffic_light_data, TrafficLightData, TrafficLightMouseStates};
 use crate::view_components::DismissibleToast;
@@ -385,6 +385,15 @@ pub fn init(app: &mut AppContext) {
     app.add_action(
         "root_view:open_settings_page_in_existing_window",
         RootView::open_settings_page_in_existing_window,
+    );
+
+    app.add_global_action(
+        "root_view:open_settings_in_new_window",
+        open_settings_in_new_window,
+    );
+    app.add_action(
+        "root_view:open_settings_in_existing_window",
+        RootView::open_settings_in_existing_window,
     );
 
     app.add_global_action(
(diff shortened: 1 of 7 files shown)


app/src/uri/mod.rs
```diff
@@ -337,87 +353,128 @@ impl UriHost {
...
+                        );
+                    }
+                    // No special sub-page: route the bare host, the `q` (search) and
+                    // `widget` (scroll-to) query params, and the simple section
+                    // sub-pages (e.g. billing_and_usage, platform, appearance,
+                    // warp_agent) resolved via `settings_section_for_simple_subpage`.
+                    maybe_simple_subpage => {
+                        let simple_section =
+                            maybe_simple_subpage.and_then(settings_section_for_simple_subpage);
+                        let search_query = parse_settings_search_query(url);
```
acarl005: Kind of a nit but `parse_settings_search_query` doesn't really need to be a `fn`. You've already parsed into pairs and you only need to pull out the `q` key and this function is redundantly parsing pairs again. Realistically no significant perf impact though. However, it does seem like this was put into a function for the sole purpose of producing a unit test and having the appearance of test coverage. The test for that isn't really helpful though IMO. Really we should be testing either `UriHost::handle` or break up `UriHost::handle` into sub-methods for each variant and be testing each of those. However, we don't have a way of stubbing out action dispatching and it's not really worth setting all that up just for this.

Overall I would just inline the logic to pull out the `q` param and delete the test.

app/src/root_view.rs
```diff
@@ -2803,6 +2842,25 @@ impl RootView {
         true
     }
 
+    /// Handles a `warp://settings` deeplink (bare, `?q=`, or `?widget=`) in an
+    /// existing window, mapping the [`OpenSettingsArgs`] variant to its
+    /// workspace action.
```
acarl005:
```suggestion
```

Uber nit: technically this is not coupled to URI handling. it handles the global action `root_view:open_settings_in_existing_window`, not specifically the URI. I would just drop the doc comment

app/src/root_view.rs
```diff
@@ -980,6 +989,36 @@ fn open_settings_page_in_new_window(section: &SettingsSection, ctx: &mut AppCont
...
+        },
+        OpenSettingsArgs::Widget { page, widget_id } => WorkspaceAction::ScrollToSettingsWidget {
+            page: *page,
+            widget_id,
+        },
+    }
+}
+
+/// Opens a new window for a `warp://settings` deeplink (bare, `?q=`, or
+/// `?widget=`), mapping the [`OpenSettingsArgs`] variant to its workspace action.
```
acarl005:
```suggestion
```

Uber nit: technically this is not coupled to URI handling. it handles the global action `root_view:open_settings_in_new_window`, not specifically the URI. I would just drop the doc comment

acarl005 (approved):
Some comments but nothing blocking! Up to you if you want to change or not.

app/src/uri/mod.rs
szgupta: agent being dumb here - will fix

szgupta: Good call — inlined the `q` extraction to reuse the already-parsed `query_string` map, and removed `parse_settings_search_query` along with its test. Pushed in ab0d6440d.

PR #13240, opened 2026-06-30
Optimize settings schema generation during release process.

vorporeal:
### Problem

The release pipeline generated `settings_schema.json` with a separate `generate_settings_schema` Cargo binary. PR #12884 gave that binary the application feature set so that it generated an accurate schema.

On macOS, the feature set enabled dynamic Sentry framework linkage. The release job ran the generator outside the application bundle. The binary had no runtime search path for the bundled framework, so `dyld` could not load `Sentry.framework`.

A framework search path would fix that immediate error, but it would not fix the release design. Packaging-only and cross-target jobs would still compile and link a separate Warp binary only to generate the schema. That work adds release time and can generate a schema with the wrong channel or feature state.

### Approach

This PR moves schema generation into the main Warp executable:

- `warp dump-settings-schema [output-path]` generates the schema with the executable's initialized channel and feature flags.
- The command writes files atomically. It writes to standard output when the output path is not present.
- Native application builds run the command with the production executable.
- Cross-target and packaging-only jobs consume a workflow-scoped canonical schema artifact for the same operating system and channel.
- macOS and Linux jobs generate schemas for each native architecture and compare them byte for byte before they select the canonical schema.
- Windows uses the x64 schema for both architectures. The current Windows ARM64 binary is cross-compiled on an x64 runner, so the workflow cannot execute it for the same comparison.
- TUI binaries expose `dump-settings-schema` for local development. Packaged TUI jobs still consume the canonical application-generated schema. Each public setting has an `x-warp-surfaces` annotation that identifies its GUI and TUI surfaces.
- Resource preparation requires exactly one input: `SETTINGS_SCHEMA_EXECUTABLE` or `SETTINGS_SCHEMA_SOURCE`.
- Linux keeps the reusable schema outside the package output directory. This prevents the schema from becoming an unintended top-level release asset.
- Nix and remote deployment paths run the target executable when that executable is available on the target system.

The standalone generator binary and its separate Cargo target are removed.

#### Set `DYLD_FRAMEWORK_PATH` for the standalone generator

This change would fix the immediate macOS launch error. It would keep the separate generator build and would not help cross-target or packaging-only jobs.

#### Remove `cocoa_sentry` or use a reduced `standalone` feature set

This change would avoid the dynamic framework dependency. It would also create a different Cargo unit and could generate a schema that does not match the shipped application feature set.

#### Revert the feature-set parity change

This change would restore the old release behavior, but it would also restore the schema mismatch that PR #12884 tried to remove.

#### Move the generator to a separate crate

A separate crate does not help if it still links the Warp dependency graph. It would still require another compilation and link step.

#### Build the generator with the initial Cargo command

This change would remove the later Cargo command, but the generator would remain a separate binary target. It would also not solve the cross-target packaging case.

#### Add dedicated schema producer jobs

A dedicated job would reduce dependencies between packaging jobs. It would also build Warp only to generate a schema. This PR instead reuses the native production executables that the release already builds.

#### Add metadata and hash sidecar files

Sidecar files could improve diagnostics and support artifact reuse across workflow runs. This workflow uses same-run artifacts, explicit job dependencies, fail-closed resource inputs, and byte-for-byte architecture comparisons. Those controls are sufficient for this change. Sidecar files can be added if the workflow later reuses schemas across runs.

## Linked Issue

There is no linked issue. This PR addresses a release failure introduced after PR #12884.

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. Not applicable.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below. Not applicable because this is a release pipeline change.

## Testing

- [x] `cargo nextest run -p warp -E 'test(schema_generation)'`
- [x] `cargo check -p warp --bin warp-oss --features release_bundle,gui`
- [x] `cargo check -p warp_tui --bin warp-tui-oss --features release_bundle,standalone`
- [x] `script/test_prepare_bundled_resources`
- [x] Bash syntax checks for the changed bundle, resource, and deployment scripts
- [x] PowerShell parsing for the changed Windows scripts
- [x] `actionlint` for `.github/workflows/create_release.yml`
- [x] YAML parsing for `.github/actions/bundle_arch_package/action.yml`
- [ ] I have manually tested my changes locally with `./script/run`. This check is not applicable because this PR does not change interactive application behavior.

An end-to-end release workflow run is still required to validate package publication and cross-job artifact transfer on all release platforms.

## Agent Mode

CHANGELOG-NONE


vorporeal:
* **#13240** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13240?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13240?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #13244, opened 2026-06-30
Update new tab placement for groups

johnturcoo:
## Description

New tabs now follow the New Tab Placement setting relative to tab groups, so you can always open a top-level tab — even when every tab is grouped.

•  After current tab → opens after the active tab and joins its group.
•  After all tabs → opens at the end of the bar, outside any group.

"New tab in group" (group More options menu) still always adds to that group. 

**Demo video** has more context on my thoughts and before/after behavior.

## Linked Issue

https://linear.app/warpdotdev/issue/APP-4802/fix-new-tab-placement-for-tab-groups

## Testing

### Screenshots / Videos
 
[Demo video](https://www.loom.com/share/a68c63a0e7934a149cb118aad718f766)

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -7248,17 +7248,18 @@ impl Workspace {
         ctx.notify();
     }
 
-    /// An active member reuses the normal new-tab inheritance + placement;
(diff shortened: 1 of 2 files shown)


peicodes (approved):
I think this is a good compromise

PR #13256, opened 2026-06-30
Add mouse event handling to TUI input

kevinyang372:
## Description
Fixes CODE-1804

Adds mouse support to the TUI prompt input (`crates/warp_tui`), bringing it close to the GUI editor's mouse interactions:

- **Click** to place the cursor, **shift-click** to extend, **drag** to select (auto-scrolling when the drag leaves the viewport), **double-click** to select a word, **triple-click** to select a line.
- **Mouse wheel** to scroll the input viewport.

Most of this is wiring rather than new machinery. The TUI framework already models mouse events, and the editor model already exposes the GUI's selection API (`select_at` / `select_word_at` / `select_line_at` / `update_pending_selection` / `end_selection`), so mouse input flows through the same `element → TuiInputAction → view.handle_action → model` path that keyboard input already uses. Selection highlighting and the cell→offset reverse map (`softwrap_point_to_offset`) already existed.

Multi-cursor (alt-click) is intentionally **out of scope** — it isn't a native TUI text-input affordance and the terminal exposes a single hardware cursor; the input stays single-selection.

### Click-event synthesis — where and why

crossterm reports only raw button **press/release**; it does **not** report click counts, so double/triple-click can't come from the event stream directly. We synthesize the count in the **runtime**, not in the view:

- A `ClickTracker` lives on `TuiScreen` (`crates/warpui_core/src/runtime/`). The crossterm→`TuiEvent` conversion stays pure; `TuiScreen::convert_event` runs the tracker immediately after conversion and before dispatch, filling the `click_count` field that already exists on the `*MouseDown` events. Both dispatch paths (the blocking harness and the production `spawn_tui_driver`) go through it.
- It is **per-button**: consecutive presses of the *same* button within ~400ms and within one cell escalate `1 → 2 → 3` (then wrap to 1); a different button, a slower press, or a press elsewhere resets to 1.
- Doing this once at the event layer mirrors the GUI (where the OS supplies `click_count` for every button) and means every TUI consumer gets correct click counts for free, instead of each view re-deriving them.

### Why not `TuiScrollable` for wheel scrolling

`TuiScrollable` is designed for an element that **persistently owns its scroll position** (e.g. a virtualized list), and its `scroll_by_rows(&mut self, …)` hook has no `AppContext`/event context. That doesn't fit the input:

- `TuiInputView` is the persistent entity; `TuiInputElement` is a **per-frame snapshot** rebuilt every render from the view's `scroll_offset`. A mutation inside `scroll_by_rows` would be discarded on the very next render — which the wheel's own `notify()` triggers.
- `scroll_by_rows` can't reach the view (to persist the new offset) or the model (to compute the total visual-line count needed for clamping).
- Input scrolling is coupled to the cursor on the view (`scroll_to_cursor`); routing the wheel through the same `handle_action` path keeps a single coherent scroll model.

So the wheel is handled like every other input event: a `TuiInputAction::Scroll { rows }` adjusts the view's `scroll_offset` (clamped to the content), and — unlike other actions — returns early so it does **not** snap back to the cursor. A `cursor_visible` flag suppresses the block cursor once it's scrolled out of view. Adopting `TuiScrollable` would require refactoring scroll state into a shared handle and is the right move only once the TUI grows a general clipped-scroll adapter (which `TuiScrollable`'s own docs call out as future work).

## Linked Issue

N/A — internal TUI input framework work (not yet wired into the shipping app).

## Testing

Unit tests were added across the changed crates; all pass, and `./script/format` + `cargo clippy -- -D warnings` are clean.

- Runtime click-count synthesis: escalate/wrap, slow-press reset, distance reset, one-cell jitter tolerance, per-button reset, and non-button events left untouched.
- TUI input mouse: single-click placement, shift-click extend, drag selection, double-click word, triple-click line, click outside the input ignored, drag-past-last-row auto-scroll, wheel up/down clamping, wheel not moving the cursor, and wheel outside the input ignored.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/crates/warp_tui/src/input/view.rs
+++ b/crates/warp_tui/src/input/view.rs
@@ -775,27 +775,43 @@ impl TuiInputElement {
         self.selected_spans = selected_spans;
     }
 
-    /// Maps a terminal cell `position` to a 1-based buffer [`CharOffset`].
+    /// Maps a terminal cell `position` to the 1-based buffer [`CharOffset`] under
+    /// it (the gap the cursor should move to for a click/drag at that cell).
     ///
-    /// Points outside the input's vertical bounds are intentionally *not*
-    /// clamped to the viewport: a point above the input maps toward row 0 and a
-    /// point below it maps past the last visible row (bounded by the buffer's
-    /// last visual row), so a drag that leaves the input drives auto-scroll.
+    /// The mouse reports an absolute terminal cell, so getting to a buffer offset
+    /// crosses three coordinate spaces:
+    ///   1. screen cell (`position`) -> row/col relative to the input's `area`,
+    ///   2. + `scroll_offset` -> the buffer's *visual* row (undoes scrolling),
+    ///   3. (visual row, col) -> char offset via the char-cell soft-wrap map.
+    ///
+    /// Points outside the input's vertical bounds are intentionally *not* clamped
+    /// to the viewport: a point above the input maps toward row 0 and a point
+    /// below it maps past the last visible row (bounded by the buffer's last
+    /// visual row), so a drag that leaves the input drives auto-scroll.
     fn offset_at(&self, position: TuiPoint, area: TuiRect, app: &AppContext) -> CharOffset {
         let inner = self.model.as_ref(app);
         let render = inner.render_state().as_ref(app);
 
-        // Buffer-relative visual row. Signed math keeps a point above `area`
-        // (smaller `y`) mapping above the current scroll offset.
+        // Step 1: row of the pointer within the input, where 0 is the input's top
+        // row. Signed so a point *above* the input (`position.y < area.y`) stays
+        // negative instead of wrapping around at 0.
         let row_in_view = i64::from(position.y) - i64::from(area.y);
+        // Step 2: the input shows buffer visual rows starting at `scroll_offset`,
+        // so the buffer row under the pointer is `scroll_offset + row_in_view`
+        // (floored at the first row).
         let visual_row = (i64::from(self.scroll_offset) + row_in_view).max(0) as u32;
+        // ...and capped at the last real visual row, so a drag below the text
+        // resolves to the buffer's end rather than past it.
         let last_row = render.max_line().as_u32().max(1).saturating_sub(1);
         let visual_row = visual_row.min(last_row);
 
+        // Column within that row, in display cells (0 is the input's left edge).
         let col = position.x.saturating_sub(area.x);
+
+        // Step 3: resolve (visual_row, col) to a char offset. The soft-wrap map
+        // clamps the column to the row's end and is 0-based, while the buffer
+        // uses 1-based gap offsets (see the kill-range helpers), so re-add 1.
         let point = SoftWrapPoint::new(visual_row, ColumnUnit::Chars(col));
-        // The char-cell soft-wrap API is 0-based; the buffer uses 1-based gap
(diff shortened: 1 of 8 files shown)


kevinyang372:
* **#13256** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13256?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13256?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_tui/src/input/view.rs
```diff
@@ -683,8 +771,87 @@ impl TuiInputElement {
...
         self.selected_spans = selected_spans;
     }
+
+    /// Maps a terminal cell `position` to a 1-based buffer [`CharOffset`].
+    ///
+    /// Points outside the input's vertical bounds are intentionally *not*
+    /// clamped to the viewport: a point above the input maps toward row 0 and a
+    /// point below it maps past the last visible row (bounded by the buffer's
+    /// last visual row), so a drag that leaves the input drives auto-scroll.
+    fn offset_at(&self, position: TuiPoint, area: TuiRect, app: &AppContext) -> CharOffset {
```
harryalbert, line 791: I'm not really understanding what's going on here, but if you're confident in it I trust you

crates/warpui_core/src/runtime/event_conversion.rs
```diff
@@ -166,6 +169,94 @@ fn key_without_modifiers(code: KeyCode) -> Option<String> {
...
+/// Maximum delay between consecutive presses of the same button for them to
+/// count as part of the same multi-click (double/triple). Roughly the standard
+/// desktop double-click window.
+const MULTI_CLICK_INTERVAL: Duration = Duration::from_millis(400);
+
+/// The pointer button a [`ClickTracker`] is tracking a multi-click run for.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum ClickButton {
+    Left,
+    Middle,
```
harryalbert, line 181: is multi click a thing for middle and right mouse clicks?

crates/warpui_core/src/runtime/event_conversion.rs
```diff
@@ -166,6 +169,94 @@ fn key_without_modifiers(code: KeyCode) -> Option<String> {
...
+            position,
+            count,
+        });
+        count
+    }
+}
+
+/// Whether two cells are the same or immediate neighbours (Chebyshev distance
+/// <= 1), tolerating slight pointer jitter between the clicks of a multi-click.
+fn is_adjacent(a: TuiPoint, b: TuiPoint) -> bool {
```
harryalbert: I wonder if this should live in the geometry file. Feels like it might be useful generally

harryalbert (approved):
very cool

crates/warpui_core/src/runtime/event_conversion.rs
kevinyang372: Yeah this gets us to parity with GUI

crates/warp_tui/src/input/view.rs
kevinyang372: Basically what we are trying to do here is converting from viewport coordinate (row / column in the render coordinate space) into the buffer coordinate (char offset) so we could update the selection

PR #13291, opened 2026-07-01
thinking blocks for TUI

harryalbert:
## Description

The overarching goal of this PR is to add collapsible thinking blocks to the TUI. This PR does that, and along the way adds some TUI primitives and elements to allow for collapsible thinking blocks.

Specifically, this PR adds:
- **New TUI primitives:** `TuiHoverable` + `TuiMouseStateHandle` in `crates/warpui_core/src/elements/tui/hoverable.rs` (hover tracking and click handling, mirroring the GUI's `Hoverable`/`MouseStateHandle` split) and `tui_collapsible` in `crates/warpui_core/src/elements/tui/collapsible.rs`, a stateless compose function — the thinking block is a composition of primitives, not a bespoke element.
- **API convention change:** This is kind of unrelated to this PR, but while working on the PR I noticed that the convention in the TUI and GUI was different for accepting child constructors. Specifically, the TUI was taking `Element impl TuiElement`, where the convention (set by the GUI) is to take `Box<dyn TuiElement>`, with callers terminating trees via `.finish()`. You can definitely skim or skip all these form changes.
- **Agent block refactor:** per-section render functions moved from `crates/warp_tui/src/agent_block.rs` to the new `crates/warp_tui/src/agent_block_sections.rs`; `TuiAIBlock::sections` now preserves message order so reasoning interleaves with text/tool calls, and gains a `TuiAIBlockSection::Thinking` variant.

## Testing

### Screenshots / Videos

https://www.loom.com/share/45b49c0c9e1e4add90c256ce2e6f5989

## Agent Mode


harryalbert:
* **#13291** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13291?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13291?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13104** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13104?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13177** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13177?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/warp_tui/src/agent_block.rs
```diff
@@ -27,6 +32,63 @@ enum TuiAIBlockSection {
...
+    /// permanently.
+    collapse_override: Option<bool>,
+    /// Hover state for the thinking header. Owned here (not created inline
+    /// during render) so it survives element-tree rebuilds, mirroring the
+    /// GUI's `MouseStateHandle` pattern.
+    hover_state: TuiMouseStateHandle,
+}
+
+impl ThinkingBlockStates {
+    /// Whether the thinking block for `message_id` is collapsed: the manual
```
kevinyang372, line 65: Why are we auto-collapsing after finished? Is that the same behavior as GUI?

crates/warp_tui/src/agent_block.rs
```diff
@@ -112,23 +178,48 @@ impl TuiAIBlock {
...
+                                .iter()
+                                .filter_map(|section| match section {
+                                    AIAgentTextSection::PlainText { text } => Some(text.text()),
+                                    // The TUI can't render these section kinds yet.
+                                    AIAgentTextSection::Code { .. }
+                                    | AIAgentTextSection::Table { .. }
+                                    | AIAgentTextSection::Image { .. }
+                                    | AIAgentTextSection::MermaidDiagram { .. } => None,
+                                })
+                                .collect::<Vec<_>>()
```
kevinyang372: Why do we need to collect first? (I think if you import itertools we could join directly)

crates/warpui_core/src/elements/tui/collapsible.rs
```diff
@@ -0,0 +1,65 @@
...
+/// Disclosure glyph shown when the section is expanded.
+const CHEVRON_EXPANDED: &str = "▾";
+
+/// Composes a collapsible section: a clickable `label` header (suffixed with a
+/// state chevron) over `body`, which is included only when `collapsed` is
+/// `false`. `on_toggle` runs when the header is clicked. The header is styled
+/// with `header_hover_style` while `mouse_state` reports it hovered and
+/// `header_style` otherwise; hover transitions are recorded on `mouse_state`,
+/// which the caller owns so it survives re-renders.
+pub fn tui_collapsible(
```
kevinyang372, line 27: Should we have a similar `TuiUiBuilder` (corresponding to GUI's UiBuilder)?

crates/warpui_core/src/elements/tui/hoverable.rs
```diff
@@ -0,0 +1,138 @@
...
+    TuiPresentationContext, TuiRect, TuiRectExt, TuiSize,
+};
+use crate::AppContext;
+
+type ClickCallback = Box<dyn FnMut(&mut TuiEventContext, &AppContext)>;
+
+/// Shared hover state for one hoverable region, owned by the composing view so
+/// it survives element-tree rebuilds.
+#[derive(Clone, Default)]
+pub struct TuiMouseStateHandle {
```
kevinyang372: Any reason we are not reusing MouseStateHandle? It feels like this should be shared

crates/warp_tui/src/agent_block.rs
harryalbert: yeah the default setting in the GUI is "Show & collapse", which is this behavior

crates/warp_tui/src/agent_block.rs
harryalbert: good call

crates/warpui_core/src/elements/tui/collapsible.rs
harryalbert: yeah that makes sense to me

crates/warpui_core/src/elements/tui/hoverable.rs
harryalbert: `MouseState` has a good amount of GUI-specific stuff that the TUI doesn't need, like:
* having `is_mouse_over_element` vs `is_hovered` being two different fields (dependent on hover delays)
* `last_event_is_synthetic_hover`, which has no meaning to the TUI
* `hover_in_timer` and `hover_out_timer` (these might be useful in the future, but I'd rather wait on adding them until we actually need them)

MouseState is generally pretty thin an doesn't do a bunch of implementation that we really want/need to share IMO, so I lean towards keeping them separate

kevinyang372 (approved):

PR #13294, opened 2026-07-01
Initial settings refactor for TUI

kevinyang372:
## Description

This PR lays the groundwork for **file-based settings in the TUI front-end** (`warp_tui`), reusing the GUI's existing `settings.toml` pipeline instead of building a parallel config system. The core idea is to teach the settings infrastructure about the **surface** it is running in (GUI vs. TUI) so a single settings stack can serve both front-ends without the two clobbering each other's local config or shared cloud state.

All new behavior is gated behind `FeatureFlag::SettingsFile`, and the TUI is not yet wired to consume these settings end-to-end — this is the enabling refactor.

Design doc: [File-based settings for the Warp TUI](https://staging.warp.dev/drive/notebook/nIIfu2BEGik3y6EMW7tjJv)

### Core settings refactors

- **Runtime settings surface (`SettingsMode`)** — Added a process-wide `SettingsMode` (`Gui` / `Tui`) to the `settings` crate, established once at startup from `LaunchMode`. Downstream settings logic consults it to vary per-surface behavior. It intentionally has no `Default`; the single accessor falls back to `Gui` explicitly.

- **Per-setting surface scoping (`SettingSurfaces`)** — Every setting now declares a **required** `surface:` attribute (`GUI`, `TUI`, or `ALL`) via the `define_settings_group!` macros. Like `feature_flag`, it is consumed at schema / default-file generation time to decide which settings belong to a given surface, so each front-end only sees the keys that apply to it. Making the attribute required forces authors to consciously scope every new setting.

- **Per-surface behavior divergence** — Behavior that must differ between surfaces is centralized on `SettingsMode`:
  - **Cloud sync** is gated (`should_sync_to_cloud`): the TUI keeps its config local, so `CloudPreferencesSyncer` is still registered but inert (no cloud reads/writes) rather than clobbering shared Drive state.
  - **Native → TOML migration** is gated (`should_migrate_native_settings`): only the GUI has legacy native-store settings to migrate; other surfaces start fresh.
  - **Config file location**: both surfaces use the same `settings.toml` filename but live in separate directories (GUI `warp`, TUI `warp_cli`) via a new `warp_core::paths::tui_config_local_dir()`. The managed-paths watcher was extended to create and watch the TUI directory so hot-reload works there too.

- **First TUI setting (`agents.model`)** — Added a TUI-scoped `agents.model` setting (default `auto`) as the initial TUI-only key exercising the new surface plumbing.

- **Resilience & diagnostics** — Settings parsing stays lenient (per-key fallback to defaults on malformed input), and `settings::init` logs load failures (parse errors and invalid keys) for both surfaces — the baseline signal for headless surfaces that have no error UI.

### Out of scope / follow-ups
- Wiring the TUI to actually consume these settings end-to-end (e.g. agent harness).
- Project-level (per-repo) config.

## Testing

Validated end-to-end by building `warp-tui-dev` under a PTY: startup parses file values (incl. `agents.model`), live edits hot-reload, malformed files log errors and fall back to defaults, and the default-settings generator emits the correct per-surface key sets (GUI excludes `agents.model`; TUI only includes TUI/ALL keys). `cargo check` and `cargo clippy -D warnings` pass for the changed crates.

## Agent Mode


kevinyang372:
* **#13294** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13294?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13294?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/examples/generate_default_settings.rs
```diff
@@ -88,6 +96,17 @@ fn main() {
 
     let active_flags = active_flags_for_channel(channel);
 
+    // Only emit settings that apply to the target surface, so e.g. the TUI
+    // file excludes GUI-only keys. Defaults to the GUI surface.
+    let surface_mode = match surface {
+        "gui" => SettingsMode::Gui,
+        "tui" => SettingsMode::Tui,
+        other => {
+            eprintln!("Unknown surface '{other}', defaulting to gui");
```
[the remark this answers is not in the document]
harryalbert: this seems reasonable

app/src/settings/ai.rs
```diff
@@ -1020,10 +1046,26 @@ define_settings_group!(AISettings, settings: [
...
+    // The default model the TUI agent uses, as a file-backed setting.
+    //
+    // TUI-only (`surface: Tui`): the GUI selects its model via execution
+    // profiles, so this key only appears in (and is read from) the TUI's
+    // settings file. `"auto"` defers to Warp's automatic model selection.
+    agent_model: TuiAgentModel {
+        type: String,
+        default: "auto".to_string(),
+        supported_platforms: SupportedPlatforms::ALL,
+        sync_to_cloud: SyncToCloud::Never,
```
harryalbert, line 1063: is there a way that we could avoid this field for TUI focused settings, if we don't plan on doing cloud syncing for TUI?

app/src/settings/init.rs
```diff
@@ -150,6 +150,23 @@ pub fn init(
         None
     };
 
+    // Always log a settings-load failure. The GUI additionally surfaces this
+    // via a banner/footer, but headless surfaces (e.g. the TUI) have no such
+    // UI, so the log is the baseline signal. Final user-facing UX is TBD.
```
harryalbert, line 155: could be worth creating a ticket for this, not sure how we want to surface

app/src/settings/mod.rs
```diff
@@ -592,7 +592,16 @@ pub fn user_preferences_file_path() -> PathBuf {
...
+///
+/// Both surfaces use the same `settings.toml` file name but live in different
+/// config directories (the GUI under [`warp_core::paths::config_local_dir`], the
+/// TUI under [`warp_core::paths::cli_config_local_dir`]) so an installed GUI and
+/// TUI never share (and clobber) one file.
 pub fn user_preferences_toml_file_path() -> PathBuf {
-    warp_core::paths::config_local_dir().join("settings.toml")
+    let config_dir = match settings::settings_mode() {
+        settings::SettingsMode::Gui => warp_core::paths::config_local_dir(),
+        settings::SettingsMode::Tui => warp_core::paths::cli_config_local_dir(),
```
harryalbert: nit: given that we're using tui verbiage elsewhere, `tui_config_local_dir` might be a more legible fn name

app/src/tui/mod.rs
```diff
@@ -119,3 +128,60 @@ fn set_login_phase(ctx: &mut AppContext, phase: TuiLoginPhase) {
         ctx.notify();
     });
 }
+
+/// Sets up the temporary end-to-end validation of the settings-file flow.
+///
+/// Logs a snapshot of the parsed settings on startup, then re-logs it on every
+/// settings-file reload (and logs any parse/validation errors). This is purely
+/// a validation aid — it does not change agent behavior.
+fn init_settings_validation_logging(ctx: &mut AppContext) {
```
harryalbert: should we merge this, or was this just for testing?

harryalbert (approved):
very cool. Just had a few small nits/questions.

I also didn't read every decision between GUI/TUI/ALL for the settings, but lmk if there are any that would be helpful for me to TAL at

app/src/settings/ai.rs
kevinyang372: This is a top level field for both TUI and GUI settings, so we could not totally avoid this field

I also don't want to paint ourselves into a corner where we cannot do cloud sync for TUI settings. This is just my strawman V0 behavior

PR #13301, opened 2026-07-02
Variable-width terminal cells: fix Indic (Telugu) script rendering

### LINKED ISSUE #13719: Telugu (and other Indic-script) terminal output renders with wrong cell widths — squeezed/overlapping/torn text
## Problem

Telugu (and other Indic-script) terminal output currently renders with fixed 1-cell-per-character allocation, which is wrong for a proportional/conjunct script. This causes squeezed, overlapping, or torn-apart-looking text — Indic grapheme clusters (which can render at 1-8x the width of a normal cell) get clipped or visually merged into neighboring characters instead of getting their real shaped width.

This is the terminal-rendering counterpart to the existing CJK "wide char = 2 cells" handling, but Indic scripts need variable width (not a fixed 2), driven by actual shaped glyph width rather than a fixed multiplier.

## Expected

Each Indic grapheme cluster should occupy cells sized to its real Core Text-measured width, with correct cursor movement, selection, resize/reflow, copy/paste, and rendering — matching how a proportional text renderer (e.g. the app's own rich-text input bar) draws the same string.

## Proposed fix

A fix is already implemented and proposed in #13301, which generalizes the terminal's wide-char mechanism into a variable-span (1-8 cell) model driven by a Core Text cluster-width measurer, and updates every consumer (cursor, selection, resize/reflow, copy/paste, rendering) to use real span instead of an assumed fixed pair.

sumanthmukkala:
Closes #13719

## Summary

Telugu (and other Indic-script) terminal output currently renders with fixed 1-cell-per-character allocation, which is wrong for a proportional/conjunct script — this causes squeezed, overlapping, or torn-apart-looking text. This PR generalizes the terminal's existing CJK "wide char = 2 cells" mechanism into "any script = N cells," giving each Indic grapheme cluster a real Core Text-measured cell width (1-8 cells).

## What changed

- `Cell::span()`/`set_span()` (3 spare bits in `Flags`) generalize the wide-char-pair mechanism to spans 1-8, keeping the legacy `WIDE_CHAR` flag in sync for backward compatibility.
- A `ClusterWidthMeasurer` trait measures a completed Indic grapheme cluster's real shaped width at input time (cluster-completion, not every repaint) via a dedicated Core Text `FontDB` instance.
- `ansi_handler.rs` buffers Indic-extending characters and writes the full cluster + measured span when the cluster completes.
- Every consumer that previously assumed a fixed 2-cell pair now uses the real span: cursor movement, selection, resize/reflow, copy/paste, line-to-string, rendering.
- The render layer detects each maximal contiguous run of Indic cells (plus absorbed spaces/attached punctuation) in a row and shapes it as ONE Core Text call, so inter-cluster and inter-word spacing flows at natural advances instead of being centered (and gapped) per cluster — matching how a proportional text renderer draws the same string. Per-cluster/per-cell foreground color is preserved via one style run per cluster (Core Text splits `CTRun`s at style-attribute boundaries).
- Cluster width allocation uses `ceil()` (never undersized) — an earlier attempt at `round()` to shrink allocation slack was reverted after it caused a conjunct glyph to visually overflow into and merge with the next word.

## Test plan

- [x] ~700 automated tests across the touched surface (cell storage, ANSI input handling, cursor/selection, resize/reflow, rendering) plus the full pre-existing suite — all pass
- [x] Visually verified on real Telugu sentences and paragraphs against Core Text's own proportional rendering (the app's rich-text input bar), across multiple sessions/restarts
- [ ] Would appreciate testing against other Indic scripts (Devanagari, Kannada, Tamil, etc.) and mixed Indic+CJK+ASCII content, since this was developed and tested primarily against Telugu

## Known limitations

- Very narrow terminals (under ~4-8 columns) resizing against scrollback content with wide clusters degrade gracefully (clamped) rather than being fully correct — no crash, but not pixel-perfect either.
- The ligature-rendering code path (`render_grid_with_ligatures`, off by default) is not yet Indic-aware; this PR only touches the default (non-ligature) render path.

### DIFF
--- a/app/src/editor/view/snapshot.rs
+++ b/app/src/editor/view/snapshot.rs
@@ -515,13 +515,20 @@ impl ViewSnapshot {
             .map(|(cur_line_idx, stylized_chars)| {
                 let mut line = String::with_capacity(stylized_chars.len());
                 let mut style_runs = Vec::with_capacity(stylized_chars.len());
-                for (idx, stylized_char) in stylized_chars.into_iter().enumerate() {
+                let mut byte_offset = 0usize;
+                for stylized_char in stylized_chars.into_iter() {
                     // Accumulate both style information and the characters to lay out.
+                    // Use byte offsets (not char indices) so that multi-byte Unicode
+                    // codepoints (Telugu, CJK, emoji, etc.) don't split a char boundary
+                    // and cause a panic in the text layout engine.
+                    let c = stylized_char.char();
+                    let char_byte_len = c.len_utf8();
                     style_runs.push((
-                        idx..idx + 1,
+                        byte_offset..byte_offset + char_byte_len,
                         StyleAndFont::new(family_id, properties, stylized_char.style()),
                     ));
-                    line.push(stylized_char.char());
+                    line.push(c);
+                    byte_offset += char_byte_len;
                 }
 
                 layout_cache.layout_text(
--- a/app/src/terminal/alt_screen/alt_screen_element.rs
+++ b/app/src/terminal/alt_screen/alt_screen_element.rs
@@ -773,7 +773,7 @@ impl Element for AltScreenElement {
             grid_renderer::render_cursor(
                 &self.grid_render_params,
                 grid.cursor_render_point(),
-                grid.is_cursor_on_wide_char(),
+                grid.cursor_cell_span(),
                 model.alt_screen().cursor_style(),
                 padding_x,
                 adjusted_grid_origin,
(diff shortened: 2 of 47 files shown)


sumanthmukkala:
@cla-bot check

sumanthmukkala:
This was auto-closed by the stale-bot after 14 days of no activity, not because of an unresolved technical issue. The original Oz review requested changes only asking that the PR be linked to a same-repo issue via `Closes #123` — that was already satisfied (`Closes #13719` is in the PR description), so there was nothing further to push. Requesting a maintainer reopen this so review can continue on the merits.

PR #13302, opened 2026-07-02
GUI/TUI agnostic file execution

harryalbert:
## Description

`RequestFileEdits` was only executable on the GUI: the shared executor required a registered `CodeDiffView`, and that view was the sole thing that could save files and assemble a `RequestFileEditsResult`. This PR makes file-edit execution surface-agnostic while keeping master's surface-owned persistence flow: the shared parts of that flow (save-completion tracking and result assembly) are expressed as the `DiffStorage` / `DiffStorageHelper` trait pair, so the GUI `CodeDiffView` and the up-stack TUI storage all run the same code.

The executor knows surfaces only through `RegisteredDiffStorage`, a small object-safe handle trait (GUI views and models share no common handle type). A surface registers its storage when the action arrives — the same register-before-preprocess contract master had with the GUI view, now stated for every surface; the executor seeds it with resolved diffs when preprocess completes and calls `accept_and_save` on it at execute time, getting back a `BoxFuture<RequestFileEditsResult>`. Every terminal outcome funnels through the action model's result choke point, which drops the executor's per-action state so prepared file contents never outlive their action.

This unblocks the TUI file-edits surface up-stack (`harry/tui-file-edits-surface`), which registers a headless storage of its own (base + deltas → `FileModel` writes) and renders a summary over it.

## Call flow

Before (GUI-only):

```
PHASE 1 — preprocess
    apply_diffs ──> on_diffs_applied ──> diff_view.set_candidate_diffs
                                          (diffs live inside the GUI view's editor buffers)

    ╌╌╌╌╌╌╌╌ GAP: user reviews / edits / clicks Accept ╌╌╌╌╌╌╌╌

PHASE 2 — execute
    diff_views[id] ──> accept_and_save
        ──> view reads its editor buffers
        ──> FileModel::save per file
        ──> SavedAcceptedDiffs ──> result (assembled by the view)
```
After (surface-agnostic):

```
registration (at action arrival, before preprocess resolves)
    surface ──> register_requested_edits(id, Box<dyn RegisteredDiffStorage>)
        GUI:  WeakViewHandle<CodeDiffView>
        TUI:  its own headless storage (up-stack)

PHASE 1 — preprocess
    apply_diffs ──> on_diffs_applied ──> storage.set_candidate_diffs(diffs)

    ╌╌╌╌╌╌╌╌ GAP: user reviews / edits in place (GUI) ╌╌╌╌╌╌╌╌

PHASE 2 — execute
    storage.accept_and_save() ──> BoxFuture<RequestFileEditsResult>
        shared DiffStorageHelper flow:
            start_saving (surface-specific write kickoff)
                GUI:            editor-buffer saves via InlineDiffView (master's flow)
                TUI (up-stack): base + deltas ──> FileModel write/rename/delete
            handle_file_saved / handle_diff_computed per file
            all complete ──> assemble result ──> resolve the future
```
## what to read

- `specs/surface-agnostic-file-edit-execution/TECH.md` — the design doc; worth reading first.
- `app/src/ai/blocklist/diff_storage.rs` — the new module. `DiffStorage` is the trait every surface implements: all methods are required — state accessors (traits can't hold fields) plus `start_saving`. `DiffStorageHelper` is the shared save-completion flow (`accept_and_save`, `handle_file_saved`, `handle_diff_computed`), blanket-implemented so surfaces can't override it — master's `SavingDiffs`/`try_emit_diffs_saved` machinery relocated. Result delivery is a oneshot stored in `DiffSaveState` instead of the `SavedAcceptedDiffs` event + executor subscription.
- `app/src/ai/blocklist/diff_types.rs` — plain diff data types (`FileDiff`, `DiffSessionType`, changed-line derivation) moved out of the GUI view module so shared code can name them without depending on any view.
- `app/src/ai/blocklist/action_model/execute/request_file_edits.rs` — the executor: a `Box<dyn RegisteredDiffStorage>` registry replaces the view map; `execute` just calls `accept_and_save` on the registered storage. Per-action state is dropped at the action model's terminal-result choke point.
- `app/src/ai/blocklist/inline_action/code_diff_view.rs` — `CodeDiffView` implements `DiffStorage`; its editor-buffer save flow is master's, with completions forwarded into the shared trait methods. `WeakViewHandle<CodeDiffView>` is the registered handle, so the executor never keeps a dead review view alive; a dead view at execute time fails recoverably.

## Testing

### Screenshots / Videos

code diff tests pass and code diffs still work.

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/4cdd3733-0e92-4383-b2c2-3e708ce2b7c7_
_Run: https://oz.staging.warp.dev/runs/019f2518-9260-72f2-8fbf-f35fc5184114_
_Plans:_
- _[File-edit persistence via DiffStorageView trait](https://staging.warp.dev/drive/notebook/rkKsLXQAekrqQgqJXi0kOF)_

### DIFF
--- a/app/src/ai/blocklist/action_model.rs
+++ b/app/src/ai/blocklist/action_model.rs
@@ -1259,6 +1259,13 @@ impl BlocklistAIActionModel {
 
         let action_id = action_result.id.clone();
 
+        // Every terminal outcome (success, failure, cancellation — from any
+        // path) funnels through here, so this is the one place executor-held
+        // per-action state is released.
+        self.executor.update(ctx, |executor, ctx| {
+            executor.discard_action_state(&action_id, ctx);
+        });
+
         // If a command action entered long-running mode (returned a snapshot), cancel all other
         // pending RequestCommandOutput actions. Only one command can be active at a time, and the
         // server can only spawn one CLI subagent. We don't cancel other actions because those
--- a/app/src/ai/blocklist/diff_storage.rs
+++ b/app/src/ai/blocklist/diff_storage.rs
@@ -3,12 +3,12 @@
 //!
 //! Every surface that stores pending diffs — the GUI `CodeDiffView` and the
 //! up-stack TUI diff storage — implements [`DiffStorage`], a required-methods-
-//! only contract: state accessors plus the surface-specific write kickoff.
-//! The shared save-completion flow is [`DiffStorageHelper`], blanket-implemented
-//! for every `DiffStorage` so no surface can override it: it tracks per-file
-//! progress in a [`DiffSaveState`] and assembles the final
-//! [`RequestFileEditsResult`], so every surface produces results through the
-//! same code.
+//! only contract: an accept-time snapshot of per-file state plus the
+//! surface-specific write kickoff. The shared save-completion flow is
+//! [`DiffStorageHelper`], blanket-implemented for every `DiffStorage` so no
+//! surface can override it: it joins the per-file save futures, computes each
+//! file's result diff, and assembles the final [`RequestFileEditsResult`], so
+//! every surface produces results through the same code.
 //!
 //! The executor knows surfaces only through [`RegisteredDiffStorage`], a small
 //! object-safe handle trait, because GUI `ViewHandle`s and model `ModelHandle`s
@@ -19,12 +19,12 @@
 //! storage.
 use std::collections::HashMap;
 use std::ops::Range;
-use std::rc::Rc;
+use std::sync::Arc;
 
-use futures::channel::oneshot;
-use futures::future::BoxFuture;
+use futures::future::{join_all, BoxFuture};
 use futures::FutureExt;
 use itertools::Itertools;
+use warp_editor::multiline::{AnyMultilineString, MultilineString, LF};
 use warp_util::file::FileSaveError;
 use warpui::AppContext;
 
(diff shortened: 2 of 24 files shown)


harryalbert:
* **#13418** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13332** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13332?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13302** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13302?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13302?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/blocklist/inline_action/code_diff_view.rs
```diff
@@ -441,20 +330,46 @@ enum AcceptSelection {
...
+/// The GUI review surface registers as a weak handle so the executor never
+/// keeps a dead review view alive; a dead view at execute time fails
+/// recoverably.
+impl RegisteredDiffStorage for WeakViewHandle<CodeDiffView> {
+    fn set_candidate_diffs(
+        &self,
+        diffs: Vec<FileDiff>,
+        session_type: DiffSessionType,
+        app: &mut AppContext,
+    ) {
```
kevinyang372, line 348: Should this return an explicit error if the viewhandle can not be upgraded?

app/src/ai/blocklist/inline_action/code_diff_view.rs
```diff
@@ -714,6 +631,12 @@ impl CodeDiffView {
                             } else if status.is_cancelled() {
                                 me.state = CodeDiffState::Rejected;
                                 me.should_expand_when_complete = false;
+                            } else if status.is_success() {
```
kevinyang372: Was this a miss from before? It's a bit unclear to me how this fits into the overall refactor

kevinyang372: Should we call `mark_accepted_for_save` here?

app/src/ai/blocklist/inline_action/code_diff_view.rs
```diff
@@ -1082,9 +988,29 @@ impl CodeDiffView {
         }
 
         ctx.emit(CodeDiffViewEvent::TryAccept);
+
+        // Persistence and result assembly run through the shared
+        // [`DiffStorageHelper`] flow once the executor (or the passive
+        // handler) calls `accept_and_save`. Optimistically mark the diff
+        // accepted and minimize the review UI.
+        self.state = CodeDiffState::Accepted;
```
kevinyang372: ditto: should we call mark_accepted_for_save here?

app/src/ai/blocklist/action_model.rs
```diff
@@ -1259,6 +1259,13 @@ impl BlocklistAIActionModel {
 
         let action_id = action_result.id.clone();
 
+        // Every terminal outcome (success, failure, cancellation — from any
```
kevinyang372, line 1262: Hmmm what's the motivation of this change?

app/src/ai/blocklist/diff_storage.rs
```diff
@@ -0,0 +1,429 @@
...
+    pub changed_lines: Vec<Range<usize>>,
+    /// Final file content.
+    pub final_content: String,
+    /// Whether the user hand-edited the content during review.
+    pub was_edited: bool,
+}
+
+/// Combines per-file progress and report state into one
+/// [`RequestFileEditsResult`], failing the whole edit if any file failed to save.
+fn assemble_result(saving: SavingDiffs, files: Vec<PendingFileState>) -> RequestFileEditsResult {
```
kevinyang372: These and below are existing logic yeah?

app/src/ai/blocklist/diff_storage.rs
```diff
@@ -0,0 +1,429 @@
...
+        self.start_saving(app);
+        // Zero-file accepts are already complete.
+        if self.save_state_mut().is_complete() {
+            let files = self.pending_file_state(app);
+            self.save_state_mut().finish(files);
+        }
+        result
+    }
+
+    fn handle_file_saved(
```
kevinyang372: Instead of having this awkward save state so we could return a future -- I wonder if file model should return a save future on top of emitting an event, this ways we could reuse that future directly

kevinyang372 (approved):
Nice

app/src/ai/blocklist/action_model.rs
harryalbert: It's a bit of a drive by fix, but, pre this change, the executor held diff_views (`HashMap<AIAgentActionId, ViewHandle<CodeDiffView>>`) but never actually removed entries. So entries accumulated for the lifetime of the executor, and the strong `ViewHandle<CodeDiffView>` entries kept dead review views alive indefinitely, which I'm pretty confident is wrong.

So, paired w/ changing what the executor stored, I also added a cleanup so that the executor doesn't just compound diff storage references forever (and keep those references alive forever) but instead cleans them up when the action gets a terminal result and they're no longer needed

app/src/ai/blocklist/diff_storage.rs
harryalbert: good call, refactored to make this change.

For context, the thing that made the original design look necessary was that an accept actually waits on two async completions per file: the save (FileModel events) and the result-diff computation (round-tripped through the editor's UnifiedDiffComputed event). A save future alone only covered half, but it turns out the second half didn't need the editor pipeline at all — the diff computation is a pure function over (base, final content), and the surface already has both. so:
•  FileModel::save now returns a completion future per request (still emits FileSaved/FailedToSave for everyone else)
•  the result diff is computed inline in the shared flow via the hoisted compute_unified_diff — no editor event round-trip
•  per-file report state is snapshotted at accept time instead of after completion, which is also just more correct: we report exactly the content that was handed to the save, instead of whatever the editor holds when the save finishes

with those three, accept_and_save collapses into snapshot → kick off saves → join_all → assemble, all in one async block.

app/src/ai/blocklist/diff_storage.rs
harryalbert: yeah, just moved to a unified location

app/src/ai/blocklist/inline_action/code_diff_view.rs
harryalbert: I think it's reasonable to log an error, but imo returning an error doesn't make as much sense. 

If the view is dead at seeding time, it's because the whole block got torn down mid-preprocess (tab closed, conversation cleared, etc), and in that case the action is getting cancelled anyway. And if the action somehow still reaches execute with a dead view, accept_and_save right below already returns an explicit DiffApplicationFailed that gets reported to the LLM.

app/src/ai/blocklist/inline_action/code_diff_view.rs
harryalbert: ah this was lingering from the earlier refactor. All acceptances should flow through the view directly (as the view is the thing that controls diff application), so this is not necessary anymore. I'll delete.

PR #13329, opened 2026-07-02
TUI autoupdate

kevinyang372:
## Description

Adds a background auto-updater to the headless `warp-tui` front-end so managed installs stay on the channel's latest nightly build. Modeled on peer CLI updaters (e.g. Claude Code's native installer): updates are downloaded and staged silently in the background and take effect on the next launch — the running session is never touched.

**Why:** the TUI is installed once via the `install.sh` one-liner and never updates, while the dev channel ships a new build nightly.

**How it works** (`crates/warp_tui/src/autoupdate.rs`):
- A `TuiAutoupdater` singleton is registered at TUI session init (always registered; eligibility is stored on the model). The polling loop only runs for release builds (baked-in `GIT_RELEASE_TAG`) running from the managed versioned install layout (`~/.warp/tui/versions/<version>/` + `current` symlink — see the companion warp-server PR below). `WARP_TUI_DISABLE_AUTOUPDATE` opts out.
- Polls every 10 minutes, matching the GUI autoupdater's cadence; each poll is a single lightweight `/client_version` request (with the GCS channel-versions JSON as fallback), and only ever moves strictly forward (no downgrades). Once an update is staged and activated, subsequent polls resolve to a `PendingRestart` no-op until relaunch.
- Downloads the tarball from `/download/tui` (streamed to disk), extracts + validates it in a staging dir next to `versions/`, renames it into `versions/<version>`, and atomically retargets the `current` symlink — all via async fs/IO on the background executor. The updater never deletes the version directory the running process is executing from, and a lock file serializes installs across concurrent TUI processes. Old versions are pruned (keeping the new install, the running version, and the `current` target).
- Emits `TUI Autoupdate Check Completed` / `Check Failed` telemetry only when the check outcome changes, so the frequent poll doesn't spam repeated events.

Companion warp-server PR (versioned install layout in `tui_install.sh`): warpdotdev/warp-server#12340

Deliberately out of scope for this PR: a manual `warp-tui update` command and an in-TUI "restart to apply" notice (re-running the install script remains the manual escape hatch).

## Linked Issue

N/A — internal, dev-channel-only tooling.

## Testing

- Unit tests for layout detection and the install lock: `cargo nextest run -p warp_tui -E 'test(autoupdate)'`; scoped `cargo clippy -D warnings` and `./script/format` clean.
- Manual E2E on macOS against real staging artifacts: built `warp-tui-dev` with an old `GIT_RELEASE_TAG`, placed it into the managed layout, and verified the updater downloaded the real latest dev tarball, staged it, atomically flipped `current`, preserved the running version, and left no staging/lock leftovers. Also verified: skip-download re-flip when the version is already staged, the background check running at TUI mount, the concurrent-process lock path, and launching through the `bin → current → versions/<v>` symlink chain (no Gatekeeper prompt; resources resolve via canonicalized `current_exe`).

## Agent Mode

CHANGELOG-NONE


kevinyang372:
* **#13329** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13329?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13329?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_tui/src/autoupdate.rs
```diff
@@ -0,0 +1,666 @@
...
+    Disabled {
+        /// Why updates are disabled, for logging/debugging.
+        reason: &'static str,
+    },
+}
+
+impl AutoupdateEligibility {
+    /// Determines whether this process should run the background update loop.
+    fn determine() -> Self {
+        if std::env::var_os(DISABLE_ENV_VAR).is_some() {
```
harryalbert, line 176: we should probably also have a setting for this

crates/warp_tui/src/autoupdate.rs
```diff
@@ -0,0 +1,666 @@
...
+                    // network errors (e.g. waking from sleep) are common here.
+                    Err(error) => log::warn!("TUI autoupdate check failed: {error:#}"),
+                }
+                me.report_outcome(&result, ctx);
+                me.schedule_next_check(layout_for_next, ctx);
+            },
+        );
+    }
+
+    fn schedule_next_check(&mut self, layout: InstallLayout, ctx: &mut ModelContext<Self>) {
```
harryalbert: nit: could just inline this

harryalbert (approved):
just had one thought on settings

PR #13332, opened 2026-07-02
Render and execute file edits in the TUI surface

harryalbert:
## Description

The base PR made file-edit execution surface-agnostic, but the TUI still rendered `RequestFileEdits` tool calls as a dead stub — no storage was ever registered, so the action could never execute there. This PR gives the TUI a real file-edits surface on top of that pipeline: `TuiDiffStorage` implements the shared `DiffStorage` contract, and `TuiFileEditsView` registers one per action with the shared executor and renders a compact summary row (`Edited 3 files (+42 −7)`) in the transcript.

The TUI has no review UI or editor buffers, which shapes the whole storage: final content is derived by applying each diff's deltas to its base and written straight through `FileModel` (the GUI reads final content out of its editor buffers instead), execution applies the deltas unmodified since there's no in-place editing yet, and result diffs are computed with `similar` because no editor computes them.

## Call flow

```
agent output gains a RequestFileEdits action
    TuiAIBlock::sync_action_views ──> add_tui_view(TuiFileEditsView)

TuiFileEditsView::new
    ├─> storage = TuiDiffStorage (owned by the view)
    └─> executor.register_requested_edits(id, TuiDiffStorageHandle(storage))

PHASE 1 — preprocess
    executor ──> storage.set_candidate_diffs(diffs)
        ──> storage emits ──> view re-renders the summary

PHASE 2 — execute
    executor ──> storage.accept_and_save() ──> BoxFuture<RequestFileEditsResult>
        shared DiffStorageHelper flow, per file:
            base + deltas ──> final content
            PersistAction (write / rename / delete) ──> FileModel dispatch
            similar-based result diff ──> handle_diff_computed
            FileModel save event ──> handle_file_saved
        all complete ──> assemble result ──> resolve the future
```

## what to read

- `crates/warp_tui/src/tui_diff_storage.rs` — the core of the PR. `TuiDiffStorage` implements `DiffStorage` for a surface with no buffers: `start_saving` derives each file's final content from base + deltas and dispatches writes through `FileModel`, routing save completions back into the shared trait flow. `PersistAction` is the single source of truth for what write actually happens (write / rename / delete) — both the `FileModel` dispatch and the reported outcome derive from it, so they can't drift; remote sessions have no rename primitive, so remote renames fall back to an in-place write and are reported at the original path. `TuiDiffStorageHandle` is the `RegisteredDiffStorage` newtype over the model handle (both types are foreign to the crate, so the orphan rule forbids implementing the trait on the handle directly).
- `crates/warp_tui/src/tui_file_edits_view.rs` — the per-action view: owns the storage, registers it with the executor at construction (before preprocess resolves, per the executor's contract), and renders `Preparing edits…` / `Edited N files (+a −r)` over the stored diffs. Purely presentational — no persistence logic lives here.
- `crates/warp_tui/src/agent_block.rs` — new machinery for *stateful* tool calls. Until now every tool-call row was a pure render fn; `TuiAIBlock` now keeps a `TuiToolCallView` registry keyed by action id, synced from the exchange output at construction and on every `on_updated_output` (`sync_action_views`), and reports the children via `child_view_ids`. File edits are the only variant so far.
- `crates/warp_tui/src/test_fixtures.rs` — tests now construct a real `BlocklistAIActionModel` over minimal session state (`add_test_action_model`), since agent blocks need one at construction. `Sessions::new_for_test` is exposed under `test-util` for this.
- Everything else is plumbing: threading `action_model` through `TerminalSessionView` → `TuiTranscriptView` → `TuiAIBlock`, new `tui_export` re-exports for the diff-storage types, a `fail_saving` hook on `DiffStorage` (the TUI is the only surface that can fail an accept wholesale, e.g. out-of-bounds deltas), and test updates to register blocks in a real TUI window now that blocks create child views.

## Testing

### Screenshots / Videos

demo: https://www.loom.com/share/216f39453f4645a39e464c7d977ee9b9

## Agent Mode

_Conversation: https://staging.warp.dev/conversation/4cdd3733-0e92-4383-b2c2-3e708ce2b7c7_
_Run: https://oz.staging.warp.dev/runs/019f2518-9260-72f2-8fbf-f35fc5184114_


harryalbert:
* **#13418** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13332** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13332?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13332?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13302** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/warp_tui/src/tui_diff_storage.rs
```diff
@@ -0,0 +1,400 @@
...
+        let start = delta.replacement_line_range.start.saturating_sub(1);
+        let end = delta.replacement_line_range.end.saturating_sub(1);
+        if start > lines.len() || end > lines.len() || start > end {
+            return Err(format!(
+                "Diff range {:?} is out of bounds for file with {} lines",
+                delta.replacement_line_range,
+                lines.len()
+            ));
+        }
+        let replacement = split_lines_preserving_newlines(&delta.insertion);
```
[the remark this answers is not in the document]
kevinyang372: Is this real?

crates/warp_tui/src/tui_diff_storage.rs
```diff
@@ -0,0 +1,311 @@
...
+fn final_content_from_op(base_content: &str, op: &DiffType) -> Result<String, String> {
+    match op {
+        DiffType::Create { delta } => Ok(delta.insertion.clone()),
+        DiffType::Update { deltas, .. } => apply_deltas_to_content(base_content, deltas),
+        DiffType::Delete { .. } => Ok(String::new()),
+    }
+}
+
+/// Applies line-range replacement deltas to `content`, producing the new content.
+fn apply_deltas_to_content(content: &str, deltas: &[DiffDelta]) -> Result<String, String> {
```
kevinyang372, line 39: Should this be an async method? Are we worried about the performance of this?

crates/warp_tui/src/tui_diff_storage.rs
```diff
@@ -0,0 +1,311 @@
...
+/// Registers `path` with [`FileModel`] using the session's local/remote backend.
+fn register_file(
+    file_model: &mut FileModel,
+    session_type: &DiffSessionType,
+    path: &str,
+    ctx: &mut ModelContext<FileModel>,
+) -> Result<FileId, FileSaveError> {
+    match session_type {
+        DiffSessionType::Local => Ok(file_model.register_file_path(Path::new(path), false, ctx)),
+        DiffSessionType::Remote(host_id) => {
```
kevinyang372, line 114: Will we even have remote diff sessions for TUI? What does that mean?

crates/warp_tui/src/tui_diff_storage.rs
```diff
@@ -0,0 +1,311 @@
...
+    fn set_candidate_diffs(
+        &self,
+        diffs: Vec<FileDiff>,
+        session_type: DiffSessionType,
+        app: &mut AppContext,
+    ) {
+        self.storage.update(app, |model, ctx| {
+            model.set_candidate_diffs(diffs, session_type);
+            // Wake subscribers (e.g. the TUI summary view) now that diffs exist.
+            ctx.emit(());
```
kevinyang372: Erm what is this for?

kevinyang372 (approved):

crates/warp_tui/src/tui_diff_storage.rs
harryalbert: I'm not too worried about performance, mostly because the GUI already does equivalent synchronous work on the same path. The GUI materializes the full buffer content synchronously at accept time (`CodeDiffView::snapshot_pending_files` calls `diff_view.editor().as_ref(app).text(app).into_string()` for every file, at `code_diff_view.rs:2936`).

We _could_ make it async and offload it, but the derived content is needed at write-dispatch time, which requires the main-thread model context. So offloading would mean restructuring the dispatch into a background-compute + main-thread-hop, which doesn't seem worth it given the GUI does the equivalent synchronous work at accept time anyway.

crates/warp_tui/src/tui_diff_storage.rs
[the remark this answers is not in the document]
harryalbert: yeah I think so. I'll fix

crates/warp_tui/src/tui_diff_storage.rs
harryalbert: I actually think we can, just given the fact that the terminal model + core stack is being re-used for the TUI. Like a user could realistically ssh into a remote box in the TUI, and then have the agent continue working.

I haven't actually tested if that flow works yet (I'm guessing it'll need some work), but given we're using the same core editing/diff flow I don't see any reason to prevent it on the TUI.

crates/warp_tui/src/tui_diff_storage.rs
harryalbert: to notify the tui file edits view when candidate diffs are set (so that it can update w/ the correct diff info). But the weird () thing isn't necessary — I'll give it an actual event type so it's clear what's going on

crates/warp_tui/src/tui_file_edits_view.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    fn render(&self, app: &AppContext) -> Box<dyn TuiElement> {
+        let theme = Appearance::as_ref(app).theme();
+        let text_color = Fill::from(ThemeFill::from(theme.terminal_colors().bright.black)).into();
+        let label = match self.summary_stats() {
+            Some((files, added, removed)) => {
+                let files_label = if files == 1 { "file" } else { "files" };
+                format!("Edited {files} {files_label} (+{added} −{removed})")
+            }
+            None => "Preparing edits…".to_string(),
```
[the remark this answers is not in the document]
harryalbert: I think this is reasonable — I'll fix

PR #13338, opened 2026-07-02
[Stable CP] Fix client-side fallback when selected models are disabled (#13331)

danielpeng2:
When a team admin disables all hosted models (e.g. direct API access is turned off), the client kept falling back to the server-provided *default* model even when that default was admin-disabled. Most visibly, users with no explicit CLI agent selection sat on `cli-agent-auto`, which the server now marks admin-disabled when no model in its fallback chain is available — so every request failed with:

```
Invalid request: the requested CLI agent model (cli-agent-auto) is not allowed for your account
```

This PR makes the no-explicit-selection fallback disable-aware:
- New `LLMPreferences::fallback_llm_info`: usable server default → first usable server choice → first custom-endpoint (team) model → raw server default as a last resort. Used by the active base, coding, and CLI agent model getters (computer use uses the same logic minus the custom-endpoint step, since custom models aren't offered there).
- The `get_default_*` getters now return the same effective default, so the profile editor dropdowns, settings profile card, context-window logic/telemetry, onboarding default, and the "(default)" model-menu label all match what requests actually send.
- The CLI agent model picker now hides admin-disabled models, matching the base/coding pickers.

N/A — fixes a regression surfaced by warpdotdev/warp-server#11498 (disabling auto models when their fallback chain has no available model).

- New unit tests in `llms_tests.rs`: base/coding fall back to the first usable hosted choice when the default is admin-disabled; CLI agent falls back to a custom endpoint when no hosted choice is usable; defaults are unchanged when usable.
- `cargo nextest run -p warp` for the llms, execution_profiles, and onboarding suites — all pass.
- [x] I have manually tested my changes locally with `./script/run` (with direct API disabled and a team custom model configured, requests now succeed with the CLI agent routed to the custom endpoint)

[Conversation](https://staging.warp.dev/conversation/0367d782-add6-481f-829d-d7e71ded8ce9)

## Justification
Why do we need this cherry-pick? Does it align with our [guidelines](https://www.notion.so/warpdev/How-We-Work-Releases-8e7d7cb4f2ca44b880fa1d0ae4876473?pvs=4)?

## Testing
How did you test this change? What automated tests did you add? If you didn't add any new tests, what's your justification for not adding any?

## Validation
- [ ] I've cut a new WarpDev to validate this change
- [ ] If this change is Wednesday or later, there's a dedicated bug bash scheduled for either this change or the release

### DIFF
--- a/app/src/ai/execution_profiles/editor/mod.rs
+++ b/app/src/ai/execution_profiles/editor/mod.rs
@@ -755,7 +755,7 @@ impl ExecutionProfileEditorView {
                         current_permissions.base_model.clone(),
                         |prefs, app| prefs.get_base_llm_choices_for_agent_mode(app).collect_vec(),
                         |id| ExecutionProfileEditorViewAction::SetBaseModel { id },
-                        |prefs| prefs.get_default_base_model().id.clone(),
+                        |prefs, app| prefs.get_default_base_model(app).id.clone(),
                         &me.upgrade_footer_mouse_state,
                         ctx,
                     );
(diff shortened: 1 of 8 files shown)


acarl005 (approved):

acarl005 (changes requested):
wrong base branch

PR #13342, opened 2026-07-02
fix(build): fall back to Cargo PROFILE env var when CARGO_FULL_PROFILE is unset

### LINKED ISSUE #11315: cargo build --release panics on Windows when CARGO_FULL_PROFILE is unset
### Pre-submit Checks

- [x] I have searched Warp bugs and there are no duplicates.
- [x] I have searched Warp known issues page and my issue is not there.

### Describe the bug

`cargo build --release` panics on Windows because `app/build.rs` reads a custom `CARGO_FULL_PROFILE` env var and defaults to `"debug"` when unset. With only a release build done, `target/debug/` doesn't exist, and the build script's Windows-asset copy step panics trying to write `conpty.dll` to a nonexistent directory.

The faulty fallback is at `app/build.rs:132`:

```rust
let cargo_full_profile = env::var("CARGO_FULL_PROFILE").unwrap_or(String::from("debug"));
let target_dir = app_target_dir(&cargo_full_profile).expect("Could not get app target directory");
copy_windows_assets(&target_dir);
```

This silently fails for any Windows contributor who follows the obvious path (`cargo build --release` after bootstrap). It accidentally works for `cargo run` because cargo's default profile is debug, so the hardcoded fallback happens to match.

`CARGO_FULL_PROFILE` is not set by `script/windows/bootstrap.ps1`, not mentioned in `CONTRIBUTING.md`, and not in the `cargo run` snippet in `README.md`. A first-time Windows contributor has no signal that they need to set it.

### To reproduce

On a Windows 11 machine with all bootstrap deps installed (including the manual fixes from #11301):

1. `cargo build --release` from a fresh shell with `CARGO_FULL_PROFILE` unset.
2. Build progresses through every dependency crate.
3. When cargo reaches the `app` crate's build script, it panics:

```
error: failed to run custom build command for `warp v0.1.0 (...\app)`

Caused by:
  process didn't exit successfully: `...\target\release\build\warp-...\build-script-build` (exit code: 101)
  --- stderr
  thread 'main' (64820) panicked at app\build.rs:436:9:
  Could not copy conpty.dll from "assets\\windows\\x64" to "...\\target\\debug": The system cannot find the path specified. (os error 3)
```

### Expected behavior

`cargo build --release` should succeed without requiring any env var the user wasn't told about. The build script should derive the profile from cargo's standard `PROFILE` env var (which IS set for build scripts), falling back to `CARGO_FULL_PROFILE` only if a non-standard custom profile name is needed.

A minimal fix:

```rust
let cargo_full_profile = env::var("CARGO_FULL_PROFILE")
    .or_else(|_| env::var("PROFILE"))
    .unwrap_or_else(|_| "debug".to_string());
```

Alternatively, the `cargo build --release` invocation could be documented (in `CONTRIBUTING.md` and/or `README.md`) as requiring `CARGO_FULL_PROFILE=release`, but that puts the burden on every Windows contributor instead of fixing the script.

### Workaround

Set the env var explicitly before the cargo invocation:

```bash
export CARGO_FULL_PROFILE=release
cargo build --release
```

```powershell
$env:CARGO_FULL_PROFILE = "release"
cargo build --release
```

### Operating system (OS)

Windows

Windows 11 Pro, 10.0.26200 (build 26200)

### Current Warp version

N/A — building from source. Repository state: `b376889` on `master`.

### Regression

No, this bug or issue has existed throughout my experience using Warp. This is the first time I've attempted a `cargo build --release` locally. The `app/build.rs` fallback to `"debug"` has been there at least as far back as `b376889`.

No — this affects local development and contribution, not daily use of the shipped Warp binary.

Yes — this is a defect in Warp's own `app/build.rs`.

### Additional context

The same build-script comment block at `app/build.rs:119-131` already acknowledges the limitation of cargo's `PROFILE` env var:

> We need to pass this information manually through an environment variable. Of the built-in variables set by Cargo: `OUT_DIR` is only a temporary directory, and `PROFILE` can only be `debug` or `release`.

But "only `debug` or `release`" is fine for the common case — the script could fall through to `PROFILE` as a sensible default. The current `unwrap_or("debug")` is the wrong fallback because it gives the *opposite* of what the developer is asking for on a release build.

---

🤖 *Generated by Claude Code on behalf of @cbeaulieu-gt*

Jason-Shen2:
## Summary

Fixes #11315

When running `cargo build --release` directly on Windows (without Warp's bundle scripts), the build script at `app/build.rs:130` defaulted `CARGO_FULL_PROFILE` to `"debug"`, causing `copy_windows_assets` to panic because `target/debug/` doesn't exist for release-only builds.

## Root Cause

`CARGO_FULL_PROFILE` is a custom env var set by `script/windows/bundle.ps1` to support custom profiles (e.g. `release-lto`). When contributors run `cargo build --release` directly, this var is unset and the build script fell back to a hardcoded `"debug"` instead of using Cargo's built-in `PROFILE` environment variable (which Cargo always sets to `"debug"` or `"release"` for build scripts).

## Fix

Chain `env::var("CARGO_FULL_PROFILE")` with `.or_else(|_| env::var("PROFILE"))` before the ultimate `"debug"` default. This:
- Preserves custom profile support for bundle scripts
- Correctly resolves to `"release"` for direct `cargo build --release` invocations
- Correctly resolves to `"debug"` for `cargo build`/`cargo run` (matching the previous default)

### DIFF
--- a/app/build.rs
+++ b/app/build.rs
@@ -117,17 +117,16 @@ fn main() -> Result<()> {
         // Retrieve the Cargo profile name so that we can put a copy of ConPTY in
         // the correct target subdirectory.
         //
-        // We need to pass this information manually through an environment variable.
-        // Of the built-in variables set by Cargo: `OUT_DIR` is only a temporary
-        // directory, and `PROFILE` can only be `debug` or `release`.
-        // See https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
-        // for more on Cargo environment variables.
+        // `CARGO_FULL_PROFILE` is set by bundle scripts for custom profiles (e.g.
+        // release-lto). Fall back to Cargo's built-in `PROFILE` ("debug"/"release")
+        // for direct `cargo build` invocations. See also:
+        // https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
         //
         // Ideally we could access `CARGO_TARGET_DIR` but this doesn't exist at build time.
         // See https://github.com/rust-lang/cargo/issues/9661.
-        //
-        // Cargo defaults to the `debug` profile.
-        let cargo_full_profile = env::var("CARGO_FULL_PROFILE").unwrap_or(String::from("debug"));
+        let cargo_full_profile = env::var("CARGO_FULL_PROFILE")
+            .or_else(|_| env::var("PROFILE"))
+            .unwrap_or_else(|_| String::from("debug"));
         let target_dir =
             app_target_dir(&cargo_full_profile).expect("Could not get app target directory");
         copy_windows_assets(&target_dir);


acarl005 (approved):
FYI for others who find this PR, you can file `PROFILE` documented here: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

PR #10462, opened 2026-05-08
spec: <details>/<summary> in markdown rendering (#10259)

lonexreb:
Spec-only PR. References #10259.

## Summary
Adds support for standard HTML5 `<details>` and `<summary>` tags in Warp's markdown renderer, so collapsible disclosure widgets work in agent output, READMEs, and other rendered markdown surfaces. Today these tags are stripped or rendered as plain text.

## Why
`<details>` / `<summary>` is part of GFM's de facto surface and standard HTML5. It is heavily used in long agent outputs and open-source READMEs to support progressive disclosure. Aligning Warp's renderer with this convention removes friction from a common rendering case.

## Spec
See `specs/GH10259/SPEC.md` for the full behavior contract: collapsed-by-default unless `open`, click + keyboard toggle, nested-to-arbitrary-depth, inline-only `<summary>` content, full markdown in the body, sanitizer attribute allowlist (`details:[open]`, `summary:[]`), accessibility (native or ARIA `role=group` + `aria-expanded` + `aria-controls`), and tests covering each.

## Notes
- No new settings, no new public API.
- Sanitizer change is minimal and explicit — only these two tags and one attribute are added.
- Persistence of open/closed state across re-renders is deferred to V1.5.

## Labels
`ready-to-spec`

PR #13345, opened 2026-07-03
spec: <details>/<summary> support in markdown rendering (#10259)

### LINKED ISSUE #10259: Support <details>/<summary> HTML tags in markdown rendering
## Feature Request

Support the `<details>` and `<summary>` HTML tags in Warp's markdown renderer to allow collapsible sections.

## Motivation

These tags are standard HTML5 and widely supported in GitHub-flavored markdown (GFM). They are commonly used in:
- AI agent output (long code blocks, verbose logs, step-by-step reasoning)
- README-style content rendered in terminal
- Structured output that benefits from progressive disclosure

## Expected Behavior

The following markdown/HTML should render as a collapsible section:

```html
<details>
<summary>Click to expand</summary>

Content hidden by default, revealed on click.

</details>
```

## Current Behavior

The `<details>` and `<summary>` tags are either stripped or rendered as plain text, with no collapsible behavior.

## Additional Context

This would be particularly useful in agent mode where responses can be long and users benefit from being able to collapse sections they don't need.

seobew:
Closes #10259

## What

Product + tech spec for #10259 (`ready-to-spec`), under `specs/GH10259/` in the `product.md` + `tech.md` format from CONTRIBUTING.md.

- **product.md** — 12 testable behavior invariants: collapsible rendering with GFM semantics, deterministic malformed-input degradation (missing/duplicate/unclosed `<summary>`), hard resource limits (nesting depth 8, 512 widgets/doc — constants, not heuristics), collapsed-state-as-view-state (copy/`raw_text` always full content), accessibility contract with renderer-generated IDs, round-trip serialization, and streaming behavior.
- **tech.md** — grounded in the current parser layout: a `FormattedTextLine::Details` container variant following the `FormattedTable` precedent (single line variant, flat document model preserved), a `parse_details` branch in `parse_markdown`'s block chain, explicit handling in the imported-HTML path (`html_parser.rs`), editor `BufferBlockStyle` mapping, and a per-invariant test matrix. Tradeoffs (start/end marker lines, tree-shaped model) are discussed and rejected with reasons.

## Relation to prior work

#10462 proposed a spec for the same issue; Oz's reviews there flagged non-deterministic resource caps, undefined malformed-`<summary>` handling, implementation pointers not matching the checkout, and unscoped preview surfaces. This spec addresses those points directly and follows the two-file spec format. Happy to converge with the earlier author if they're still active.

## Open questions for maintainers

Called out in tech.md: whether the disclosure widget reuses the existing block-folding machinery or gets a dedicated component, and the confirmed enumeration of non-interactive preview call sites for invariant 11.


seobew:
@cla-bot check

seobew:
Addressed all three review points in 2d2b79f:

1. **Malformed `<summary>`** — product invariant 8 now covers unclosed `<summary>` (consumes remainder of the details body; body empty; content preserved), stray `</summary>` inside a body (plain text), and attribute/inline-only rules (line breaks collapse, block constructs inside summary render as literal inline text). Test matrix rows 8d-8f added.
2. **Buffer boundaries/nesting** — tech spec §4 now specifies the `FormattedTable` mechanism verbatim: a `DETAILS_BLOCK_MARKDOWN_LANG = "warp-markdown-details"` internal code block (per `TABLE_BLOCK_MARKDOWN_LANG`, markdown_parser.rs:39) whose fence delimits the region; nested details survive as literal markdown inside the parent block and re-materialize on re-parse under `MAX_DETAILS_DEPTH`. Round-trip test added to the matrix.
3. **Surface enumeration** — invariant 11 is now self-contained two-tier rendering (interactive widget in the agent conversation path via `parse_markdown_into_text_and_code_sections`, util.rs:35/189; deterministic static fallback everywhere else), and tech spec §5 enumerates every current `parse_markdown` call site per tier. The only remaining open question is the component choice (reuse block-folding vs dedicated disclosure), which is intentionally left to maintainers.

/oz-review

seobew:
Addressed all three points in a74d6ae:

1. **Tag scanning vs literal regions** — new product invariant 8(g): tags inside fenced code blocks are code content and never open/close a container; recognition happens only at block start outside fences. Tech spec now specifies matching as a fence-aware line walk (reusing the fence recognition `parse_code_block` performs) instead of a raw text scan, with test row 8g locking the behavior.
2. **Round-trip contract vs data model** — `FormattedDetails` now carries `body_source: String` (the verbatim input slice) alongside the parsed `body`. Serialization emits `body_source` directly, so the `warp-markdown-details` round-trip is byte-exact and implementable without reconstructing markdown from parsed `FormattedText`. Cost and equality semantics are stated.
3. **Consumer coverage** — the fallback is no longer an enumerated per-surface obligation. It is anchored structurally at the two sinks all consumers render through: `FormattedTextElement` (warpui_core, the shared layout element — fallback there, widget as an opt-in builder used only by the agent path) and the editor buffer conversion (§4 mapping). The enum-variant addition is also a compile-time forcing function across the 57 non-test files referencing `FormattedTextLine`. The complete 18-file `parse_markdown` consumer list is retained as informative reference.

/oz-review

vorporeal:
hey @bnavetta - can you take a look at this?  you have much more context on markdown rendering than i do.

specs/GH10259/product.md
```diff
@@ -0,0 +1,46 @@
...
+- Supporting `<details>` in non-markdown terminal output (raw ANSI streams).
+- Persisting open/collapsed state across app restarts.
+- CSS/attribute styling of details blocks (`class`, `style`, etc. are ignored).
+
+## Behavior
+
+1. A block-level `<details>` element whose content is markdown renders as a collapsible section: a summary line with a disclosure indicator (▸/▾), followed by the body when expanded.
+2. The section is collapsed by default. If the `<details>` tag carries the standard `open` attribute, it renders expanded initially.
+3. Clicking the summary line (or pressing Enter/Space while it is focused) toggles the section. The toggle target is keyboard-focusable and exposed to accessibility APIs as an activatable disclosure control. Any accessibility IDs are renderer-generated; they are never derived from input attributes.
+4. The first `<summary>` child provides the summary line, rendered with inline markdown styling. If no `<summary>` is present, the summary line is the localized literal "Details". Any additional `<summary>` siblings are treated as ordinary body content.
```
bnavetta: By "ordinary body content", do you mean that the text would be shown as-is, with the `<summary>` tags, or would the tags be omitted?

specs/GH10259/product.md
```diff
@@ -0,0 +1,46 @@
...
+5. The body content between `<summary>` (or `<details>` when no summary exists) and `</details>` is parsed as normal markdown: paragraphs, code blocks, lists, tables, and nested `<details>` all render as they would at top level.
+6. Nesting is supported up to a fixed depth of 8. This limit is deterministic: a `<details>` opening at depth 9 or greater is not rendered as a widget — its tags are rendered as plain text and its content rendered inline (current behavior), and this fallback applies consistently regardless of input size or timing.
+7. A document renders at most 512 details widgets. From the 513th onward, the same deterministic plain-text fallback as (6) applies. Both limits exist to bound resource use against untrusted input; they are constants, not heuristics.
```
bnavetta: Could you elaborate on this a bit? The cost of another level of `<details>` nesting or an additional `<details>` block should be fairly comparable to the cost of tracking the same content as plain text

specs/GH10259/product.md
```diff
@@ -0,0 +1,46 @@
...
+9. Collapsed state is view state, not content: copying the block, `raw_text()` extraction, and buffer conversion always include summary and full body regardless of the current collapsed state.
+10. Round-trip: converting the rendered block to markdown (e.g. copying an agent message as markdown) emits `<details>`/`<summary>` markup, preserving the `open` attribute only if the block was expanded by default in the source.
+11. Rendering is two-tier. The interactive disclosure widget (invariants 1-3) is implemented in the agent conversation block renderer — the surface the issue targets. Every other markdown surface uses a generic static fallback: the block renders expanded, in source order (summary line first, then body), with no disclosure indicator and no interaction. The fallback is the default for any renderer that has not opted into the widget, so no surface can render details in an undefined state. The concrete call sites for both tiers are enumerated in the tech spec.
```
bnavetta: I think it's worth speccing out support in the notebook/plan renderer as well (`RichTextEditorView`). For plans in particular, support for collapsed content would be useful.

specs/GH10259/tech.md
```diff
@@ -0,0 +1,104 @@
...
+pub struct FormattedDetails {
+    pub summary: FormattedTextInline,     // parsed inline markdown; "Details" literal if absent
+    pub body: FormattedText,              // recursively parsed markdown (rendering model)
+    pub body_source: String,              // verbatim markdown source of the body (serialization model)
+    pub default_open: bool,               // `open` attribute present
+}
+// FormattedTextLine::Details(FormattedDetails)
+```
+
+`body_source` is the verbatim input slice between the end of the summary and the closing tag. It exists so the buffer/markdown round-trip (§4) serializes the body exactly, without reconstructing markdown from parsed `FormattedText`; `body` is derived from it (`parse_markdown(body_source)`) and used only for rendering. Cost: one extra copy of the body text per details block, bounded by input size. `PartialEq` derives over all fields; equal `body_source` implies equal `body`, so streaming delta comparisons stay consistent.
```
bnavetta: Generally, Warp's rich-text pipeline doesn't attempt to guarantee exact preservation of the source Markdown; I think it's fine to continue that here

specs/GH10259/tech.md
```diff
@@ -0,0 +1,104 @@
...
+
+The whole `<details>…</details>` region becomes a single `FormattedTextLine`, like `Table`. Rationale over start/end marker lines: markers would let intervening edits produce unbalanced documents, and every renderer/serializer consumer would need to track container state; a single variant keeps the flat model's invariants intact.
```
bnavetta: The approach taken by `Table` is very much an intermediate implementation to keep our options open for eventually supporting table editing. We shouldn't try to replicate it here - details regions should be modeled directly in the buffer. Among other things, that approach makes text selection much more complicated (which is true of tables regardless).

Given that you're supporting nested details items, I'd look at lists for inspiration instead (specifically ordered lists). Details items are _kind_ of like ordered list markers, where the block marker metadata is the summary text rather than the list number. That approach should also make it fairly doable to track nested details depth in the `SumTree`

bnavetta (changes requested):

specs/GH10259/product.md
seobew: Good catch — the spec did not actually say. Clarified in 2632ed67: the tags render as visible plain text and the inner content as normal markdown, consistent with 8(b)/8(e) and with the principle that markup the parser did not consume is never silently dropped.

specs/GH10259/product.md
seobew: You are right that the content cost is comparable — the limits are not about content size. What each widget adds *beyond* its content is (a) a level of parser recursion (the depth limit) and (b) a keyboard-focusable element with an accessibility-tree node and event bookkeeping (the count limit). Invariant 7 now states this explicitly. Happy to adjust the constants if you would prefer different bounds.

specs/GH10259/product.md
seobew: Agreed — and the buffer-native model from your §4 comment makes this nearly free: details regions now live directly in editor buffers with `HiddenLinesModel` folding, so `RichTextEditorView` becomes the second interactive-tier surface rather than a static-fallback consumer. Product invariant 11 and tech §5 are updated to commit to it (2632ed67).

specs/GH10259/tech.md
seobew: Makes sense. Dropped `body_source` entirely; round-trip is now canonical re-serialization from parsed structure, and product invariant 10 is reworded to match the pipeline-wide guarantee (structure and content preserved, exact source bytes not). This also removes the duplicated body storage and its equality subtleties. (2632ed67)

specs/GH10259/tech.md
seobew: Thank you — this reshaped tech §4. A details region is now modeled directly in the buffer on the ordered-list precedent:

- A one-line `DetailsSummary` block whose marker carries `{ depth, default_open }` and whose **line text is the summary itself** (editable/selectable, like ordered-list item text); the disclosure triangle is a render-time margin decoration like the list number, never in the char stream.
- Body blocks follow as ordinary blocks, closed by a zero-content `DetailsEnd` sentinel.
- Depth is a `StyleSummary`-style +1/−1 counter over the begin/end pair, so depth-at-offset is an O(log n) `SumTree` seek — the "track nested depth in the SumTree" you pointed at.
- Both markers read as a single `\n` in the byte iterator, so selection/copy stay trivial (the property the table mechanism loses).
- Unbalanced pairs after edits get deterministic rules mirroring product 8(a)/8(b); collapse reuses `HiddenLinesModel` (per-view, outside undo history), which makes invariant 9 hold by construction.

The table `to_internal_format` path is dropped. One judgment call worth flagging: I put the summary in the line's **text content** rather than in the marker metadata — reading your note strictly ("block marker metadata is the summary text") you may have meant the summary string lives on the marker itself. I chose line text because it keeps the summary editable and inline-selectable with zero new offset machinery, but it is an easy flip if you would rather keep it out of the char stream. (2632ed67)

seobew:
@bnavetta thanks for the review — all five points addressed in 2632ed67, replies inline on each thread. Summary:

1. **Duplicate `<summary>`** (product): tags render as visible plain text, inner content as normal markdown — consistent with 8(b)/8(e).
2. **Depth/count limits** (product): reworded — the bound is on what each *widget* adds (parser recursion; focusable element + a11y node + event bookkeeping), not content size.
3. **`RichTextEditorView`** (product): promoted to the second interactive-tier surface; the buffer-native model makes native folding for plans nearly free.
4. **Byte-exact round-trip** (tech): dropped `body_source`; round-trip is now canonical re-serialization, matching the rest of the pipeline.
5. **Buffer model** (tech §4): rewritten off the table mechanism onto the ordered-list precedent — `DetailsSummary`/`DetailsEnd` block markers, summary as line text, a `SumTree` counter dimension for depth, `HiddenLinesModel` for collapse, and deterministic edit-rebalancing. Flagged one judgment call on that thread (summary as line text vs. marker metadata) — easy to flip if you prefer.

Ready for another look when you have a moment.

specs/GH10259/tech.md
```diff
@@ -0,0 +1,119 @@
...
+    pub default_open: bool,               // `open` attribute present
+}
+// FormattedTextLine::Details(FormattedDetails)
+```
+
+There is no verbatim-source field: serialization is canonical re-serialization from parsed structure (product invariant 10), consistent with the rest of the rich-text pipeline, so `body` is the only body representation. `PartialEq` derives over all fields, so streaming delta comparisons compare parsed structure.
+
+In the parser IR (the read-only rendering model), the whole `<details>…</details>` region is a single `FormattedTextLine`. Start/end marker lines are wrong *at this layer*: `FormattedText` consumers are stateless per-line renderers and `compute_formatted_text_delta` diffs lines independently, so nothing maintains marker balance here. The editor buffer is the opposite case — it has edit machinery that already maintains paired markers — and gets a paired representation instead (§4).
+
+- `LineCount::num_lines` returns `1 + body line count` (content lines, independent of collapsed state — collapsed is view state, spec invariant 9).
```
bnavetta: Just to be clear, the `1` is for the summary line?

specs/GH10259/tech.md
```diff
@@ -0,0 +1,119 @@
...
+- **Region extent**: the body is the run of ordinary blocks following the summary line, terminated by a zero-content end sentinel block `BufferBlockStyle::DetailsEnd { depth: u8 }`. Begin/end markers form pairs the same way paired inline-style markers do (`StyleSummary`'s +1/−1 counters, text.rs:1551): a counter field added to the buffer summary increments on `DetailsSummary` and decrements on `DetailsEnd`, so nesting depth at any offset is an O(log n) `SumTree` summary seek, and the `depth` field on each marker is a cached copy of that structural depth (as `indent_level` is for lists).
```
bnavetta: The fact that this is zero-content makes me wonder if details start/end markers should be new top-level variants of `BufferText`, rather than nested in `BufferBlockStyle`. We assume that there's only one `BufferBlockStyle` active for a given character, so we'll need a way to express that something is a code block within a details section, for example - that seems easier if details are not themselves block styles.

When rendering a given block, we'd then be able to consult the `SumTree` to check both whether or not it's part of a details section and what its specific style is. We can then adjust styling (colors, fonts, etc.) on a per-block basis.

This seems sort of equivalent to how link start/end markers _don't_ use quite the same modeling as data-less markers for other inline styles like bold and italic

specs/GH10259/tech.md
```diff
@@ -0,0 +1,119 @@
...
+- **Collapse = existing folding**: collapsed state reuses `HiddenLinesModel` (hidden_lines_model.rs:20) — the body range is an anchor-pair hidden region, per-editor and outside buffer content, initialized from `default_open`. This satisfies product invariant 9 by construction: hiding is view state; copy, `raw_text()`, and serialization walk the buffer, which always contains the full body. Toggling does not edit the buffer, so it cannot dirty deltas or undo history.
```
bnavetta: This is something to check during implementation, but copying _might_ also respect `HiddenLinesModel`

bnavetta (changes requested):

bnavetta:
Thanks for the second pass! I'd like the spec to clarify how details sections interact with other block-level styling, but this is generally headed in the right direction.

specs/GH10259/tech.md
seobew: Yes — the `1` is the summary/header line the container renders above its body, present regardless of collapse state. Reworded in ecd6ccdc to say so explicitly.

specs/GH10259/tech.md
seobew: This is the right call — adopted in ecd6ccdc, and it also settles the summary-placement question I flagged last round. §4 now models details as a top-level `BufferText::Details(DetailsMarker::Start { summary, default_open } / End)` zero-width span marker, directly on the `Link` precedent rather than as a `BufferBlockStyle`:

- Because it is a span, not a style, it is orthogonal to per-character block style: a code block/list inside a details region keeps its own `BufferBlockStyle`, and a renderer consults the `SumTree` for both "inside details, at what depth" and "this block's own style" — exactly the per-block styling adjustment you described. (You are right that this parallels how link start/end markers differ from the data-less bold/italic markers.)
- Depth is the balanced counter, not a stored field: I add `details_counter`/`total_details_marker` to `StyleSummary` and a `DetailsDepth` dimension mirroring `link_counter`/`LinkCount` (text.rs:1550/1743), so depth-at-offset is an O(log n) seek.
- The marker is zero-width like `Link` (added to the zero-width arm of `Item::summary()`, text.rs:1817, and skipped by the `Bytes` iterator, text.rs:301), so it adds no `\n` and selection stays trivial.
- Summary now lives on the `Start` payload (your "block marker metadata is the summary text"), so editing it is a marker-data affordance like editing a link URL, not inline line-text editing.

specs/GH10259/tech.md
seobew: Checked the copy path against the current tree: it does **not** respect `HiddenLinesModel`. `selected_text_as_plain_text` → `clipboard_text_in_range` (buffer.rs:1039/2374) walks the selection's raw char-offset range, and the selection anchors resolve to real buffer offsets (selection_model.rs:128) with no `HiddenLinesModel` reference on that path — hidden ranges are only collapsed in the rendered tree (render/model/mod.rs:2691) and used for cursor navigation (selection.rs:414-459). So selecting across a collapsed details span copies the full hidden body, which is what makes invariant 9 hold today. I have written §4 to state this as an invariant the implementation must preserve (if copy is ever changed to respect folds, details would need to opt out), and added a regression test guarding it. Agree it is worth re-confirming during implementation.

seobew:
@bnavetta thanks for the second pass — all three addressed in ecd6ccdc, replies inline.

Headline: **details are now a zero-width span marker, not a `BufferBlockStyle`.** Your point about per-character block-style exclusivity was correct, so §4 follows the `Link` precedent — a top-level `BufferText::Details(DetailsMarker::Start { summary, default_open } / End)` that is orthogonal to block style. Body blocks keep their own styles; a renderer queries the `SumTree` for details-depth and block-style independently; depth is the balanced counter (`LinkCount` precedent), not a stored field; the marker is zero-width so selection stays trivial; and the summary lives on the `Start` payload (settling the summary-placement question from last round).

Also: I verified against the current tree that the copy path does not consult `HiddenLinesModel` (buffer.rs:1039/2374), so a collapsed span still copies its full body — invariant 9 holds, now written as an invariant to preserve with a regression test. And confirmed the `1` in `num_lines` is the summary/header line.

Ready for another look.

specs/GH10259/tech.md
```diff
@@ -0,0 +1,138 @@
...
+
+### 4. Editor buffer: details as a span marker, not a block style (link-marker precedent)
+
+Details start/end are a **new top-level `BufferText` variant modeled on `Link`**, not a `BufferBlockStyle`. This is the correction from review: a `BufferBlockStyle` is per-character-exclusive (one active block style per position), so a code block *inside* a details region could not be both "a code block" and "inside details" if details were a block style. Links already solve exactly this — a link span coexists with whatever inline styles its text carries — by being an independent zero-width marker layer. Details replicate that layer at block granularity.
+
+- **The marker**: add `BufferText::Details(DetailsMarker)` (text.rs:541), mirroring `Link(LinkMarker)`:
+
+  ```rust
+  pub enum DetailsMarker {
+      Start { summary: String, default_open: bool }, // summary as inline-markdown source, like LinkMarker::Start(url)
```
bnavetta: Ah sorry, thinking about this more, since  the summary is also formatted text, we'll need a way to track that formatting. We probably want 3 markers, so the buffer structure is something like:

```
<details:start default_open=true>This is the summary<details:end-summary>This is the body<details:end>
```

I think we can do that without significantly changing the SumTree setup - we'd still count start/end markers to determine depth, and when extracting the summary, we can seek or linearly scan to the nearest `EndSummary` marker.

With the current structure, we'd have to parse the inline summary markdown again to display it

bnavetta (commented):
Sorry for the delay! One last suggestion, but otherwise ready to implement

seobew:
@bnavetta thanks — good call, the summary needs its own formatting. Pushed in 869f7478 as the 3-marker buffer model you sketched:

```
<details:start default_open=true> summary text <details:end-summary> body blocks <details:end>
```

- **`DetailsMarker`** drops the `summary: String` from `Start` (now just `default_open`) and gains `EndSummary`. The summary is the ordinary styled buffer text between `Start` and `EndSummary`, so its formatting (bold, inline code, links) rides on the buffer's normal inline-style + link-marker machinery like body text — no separate re-parse, and editing the summary is plain inline character editing.
- **SumTree unchanged, as you noted**: only `Start` (+1) / `End` (−1) feed `details_counter`; `EndSummary` is depth-neutral. Summary extraction is a seek/linear scan from a `Start` to its matching `EndSummary`.
- **Rebalancing** got the third-marker rules: a `Start` with no `EndSummary` before its `End` → empty summary + literal `Details` label (the invariant-4 missing-summary rule); a stray `EndSummary` is inert and skipped on serialization.
- Round-trip (`to_markdown`/`to_formatted_text`/`core.rs` flatten), render model, and selection/copy updated for the summary sub-range; added a summary-formatting round-trip test (bold/code/link survive) plus `EndSummary` rebalancing rows to the editor test matrix.

seobew:
@bnavetta the 3-marker model from your last suggestion is pushed in `869f7478` (summary tracked via an `EndSummary` marker between `Start` and `End`, so its formatting rides on the normal inline-style/link machinery; SumTree counter unchanged as you noted — only `Start`/`End` feed depth). I believe that closes the last open item — whenever you have a moment for another look, it should be ready to implement. Thanks!

PR #13389, opened 2026-07-05
[GH6156] Windows Jump List for Tab Configs

### LINKED ISSUE #6156: Windows Jump List
### Pre-submit Checks

Windows Terminal has this great feature: if placed into the app bar, when right clicked it shows all the possible shells. In warp, instead, I have to open Warp (with the default shell) and then select another one.
Which is very annoying. Could it be possible to develop a similar behaviour for Warp too?

Image

_No response_

### Additional context

_No response_

### Operating system (OS)

Windows

4

None

kevinnio:
Closes #6156

## Description

Implements a Windows taskbar jump list for Tab Configs. The user's saved Tab Configs are now pinned to Warp's taskbar jump list under a "Tab Configs" category, with one static "New Window" task above it. Clicking an entry opens the config via the existing `<scheme>://tab_config/<stem>` deeplink; clicking "New Window" uses the existing `<scheme>://action/new_window?path=~` deeplink.

The jump list is populated by a new `JumpListManager` singleton entity that subscribes to `WarpConfigUpdateEvent::TabConfigs`. A feature flag `WindowsJumpList` gates the code; when disabled after a previous enabled run, any previously committed custom destinations/tasks are cleared from the jump list so stale entries don't persist.

## Testing

Manual testing was performed on Windows with the flag enabled:

- Saved a Tab Config, pinned Warp to the taskbar, right-clicked, and confirmed "New Window" and the "Tab Configs" category/entry appear; clicked each and verified behavior.
- Quit Warp and clicked each entry from the jump list; confirmed cold-start opens the correct window with exactly one tab (the config).
- Added/deleted a `*.toml` while Warp ran; confirmed the jump list updates without restart.
- With the flag disabled after a previous enabled run, confirmed any previously committed custom tasks/categories are cleared from the jump list.
- For a parametrized Tab Config, confirmed the params modal appears on jump list click (matching the `+` menu behavior).
- Removed a Tab Config from the jump list via the Windows UI, then triggered a refresh; confirmed the removed entry stays removed.
- Verified the first window paints before the jump list commit finishes; a failed refresh does not block startup or surface an error UI.

### Videos

1. **Cold-start via jump list entry.** Record clicking a Tab Config entry from the taskbar jump list while Warp is fully closed; confirm exactly one window opens with exactly one tab (the config).

   https://github.com/user-attachments/assets/4261382f-769c-409a-9744-ebb466400609

2. **Live refresh on save/delete.** While Warp runs, save a new `*.toml` in `~/.warp/tab_configs/`, then right-click the taskbar and confirm the entry appears without restart. Then delete it and confirm it disappears from the jump list after a few seconds.

   https://github.com/user-attachments/assets/183ac365-017a-428a-812b-0efdd06c1829

3. **Flag-disabled cleanup.** Disable the flag, relaunch Warp; confirm the custom destinations/tasks are cleared from the jump list.

   https://github.com/user-attachments/assets/19f7d911-1948-48a6-b998-f8ab9f194e18

---

## Caveats and deviations from the spec

The implementation follows the tech spec (`specs/GH6156/tech.md`) for the most part, but some deviations were required to achieve the desired behavior. The following changes are **not** in the spec — they were added during implementation to fix correctness issues discovered while exercising the new cold-start path:

### `launch()` skip-restore logic (`is_launch_intent_url`)

The existing `launch()` function always ran session restore before handling startup URLs. When a jump list entry (or any other startup deeplink) was clicked while Warp had prior windows saved, restore reopened those windows and the URL then opened another window on top, yielding duplicate windows. The same pattern produced 2 tabs when clicking a Tab Config entry with no existing window: session restore created a default workspace with a homepage tab, then `handle_tab_config_uri` added the config as a second tab via `add_tab_with_pane_layout`. My agent flagged this as a bug, but this rather looks like a new use case that wasn't covered yet.

**Fix:** Added `is_launch_intent_url(url: &Url) -> bool`, which returns true for any startup URL that already carries a full "open this" intent, specifically `<scheme>://action/new_window`, `<scheme>://tab_config/<stem>`, and the existing cloud-agent-web-home pattern. When any is present, `app_state` is set to `None` so session restore is skipped; the URL becomes the sole driver of window creation. This matches the precedent already in the codebase for cloud-agent web-home launches and patches the root cause at the shared launch path. All callers of those deeplinks benefit, not just jump list entries.

I'm willing to discuss removing/changing this if maintainers consider this overboard or against product design.

---

## Future work

- While working on this I noticed that Tab Configs don't include Active Shells (like powershell, bash, zsh, wsl, etc), which is probably something we also want to cover. Implementation is similar but I didn't want to include it in this PR since I'm already deviating from the agreed spec, so I'm willing to work on it in a future one.
- Launch configs (`~/.warp/launch_configs/`) open whole window layouts and could be a natural fit for the jump list as well. I left them out to keep this PR scoped to Tab Configs.

---

CHANGELOG-NEW-FEATURE: Windows taskbar jump list for Tab Configs and "New Window"
CHANGELOG-BUG-FIX: skip session restore when a startup deeplink (tab_config, action/new_window) is present so it drives window creation instead of duplicating windows


kevinnio:
@cla-bot check

specs/GH6156/tech.md
```diff
@@ -0,0 +1,50 @@
...
+   - `pub fn refresh_jump_list(configs: &[TabConfigEntry])` that calls `ICustomDestinationList::BeginList`, adds a "New Window" task, adds a "Tab Configs" category when configs are non-empty, and commits. On COM failure, log and return; jump list population must never break startup. The function is safe to call from a background thread.
+   - A helper that builds `IShellLinkW` shortcuts targeting the current executable with the deeplink URI as arguments. Set the visible label via `IPropertyStore::SetValue(PKEY_Title, ...)` and the tooltip via `IShellLinkW::SetDescription`.
+   - Initialize COM (`CoInitializeEx(COINIT_APARTMENTTHREADED)`) at entry and pair with `CoUninitialize`, because `IShellLinkW`/`ICustomDestinationList` are apartment-threaded and must be created and used on the same thread.
+4. **Tab Config to jumplist item mapper.** Add `tab_configs_to_jump_entries(configs: &[TabConfig]) -> Vec<TabConfigEntry>` in the jump list module. Derive the file stem from `config.source_path`, use `config.name` as the label, and produce `warp://tab_config/<stem>` deeplinks. Sort alphabetically by stem.
```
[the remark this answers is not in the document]
kevinnio: Adding it.

specs/GH6156/product.md
```diff
@@ -0,0 +1,46 @@
...
+
+10. This feature is Windows-only. macOS and Linux are unaffected; no jump list entries, categories, or AUMID changes are introduced there.
+
+11. When the `WindowsJumpList` feature flag is disabled, no jump list entries are populated and no AUMID is set by this feature; existing behavior is unchanged.
+
+12. When two Tab Config files have the same display `name`, both appear in the jump list with identical labels, but each entry resolves via its unique file stem. This matches the `+` menu behavior.
+
+13. Renaming a Tab Config file removes the old jump list entry and adds the new one after the next directory-watcher refresh.
+
+14. Windows usually allows up to 12 items per category in jump lists. When the number of saved Tab Configs exceeds that, only the capped subset appears. The subset is deterministic, ordered alphabetically by file stem to match the `+` menu source order.
```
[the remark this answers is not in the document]
kevinnio: Got it. I'd rewrite it as display name order.

acarl005:
@kevinnio nice! I'll review later today.

specs/GH6156/tech.md
```diff
@@ -0,0 +1,51 @@
...
+4. **Tab Config to jumplist item mapper.** Add `tab_configs_to_jump_entries(configs: &[TabConfig]) -> Vec<TabConfigEntry>` in the jump list module. Derive the file stem from `config.source_path`, use `config.name` as the label, and produce `<scheme>://tab_config/<stem>` deeplinks using `ChannelState::url_scheme()`. Sort alphabetically by display name. The jump-list refresh function then truncates the sorted entries to the slot count returned by `ICustomDestinationList::BeginList` (`pcMinSlots`, default 10) so the capped subset is deterministic.
+5. **Trigger refresh asynchronously from the existing reload path.** In `app/src/user_config/native.rs`, inside the `ctx.spawn` callback for tab configs (both startup and `WarpConfigUpdateEvent::TabConfigs`), map the parsed configs to entries and fire a fire-and-forget background task:
+   ```rust
+   ctx.background_executor()
+       .spawn(async move {
+           tokio::task::spawn_blocking(move || refresh_jump_list(&entries)).await
+       })
+       .detach();
+   ```
```
acarl005, line 51: I don't think `refresh_jump_list` belongs in `app/src/user_config/native.rs`. We should have a separate `SingletonEntity` that subscribes to `WarpConfig` events and performs the update. `WarpConfig` shouldn't need to have any coupling to jump list stuff.

As for the `background_executor().spawn()` here.. that is redundant. Since we don't care about the result there is no need to await it in the background executor. I think you can just write

```
let _ = tokio::task::spawn_blocking(move || refresh_jump_list(&entries));
```

acarl005 (changes requested):
Overall looks good! Requesting some minor changes to the spec.

specs/GH6156/tech.md
kevinnio: Fair enough. I'll update the spec.

kevinnio: Spec updated.

acarl005:
@kevinnio Moved this into `ready-to-implement` so feel free to start coding. Thanks!

kevinnio:
Reopening request: both concerns from the last review were addressed on Aug 19. I just forgot to ask Oz for a re-review.

- `?new_window=true` behavior restored in `handle_tab_config_uri` (`175e1ab9`, `dd5cf324`), and the doc comment updated to match the implementation.
- The `NewWorkspaceSource::TabConfig` doc comment no longer narrates its caller (`aa8af3b3`).

The PR was auto-closed before a re-review could pick these up. @oss-maintainers could you reopen it?

PR #13406, opened 2026-07-06
Add shell command input mode + execution to TUI

harryalbert:
## Description

Just like in the GUI, you should be able to manually enter bash input mode (using `!` w/ the same semantics to enter and exit the shell mode) and, when in that mode, execute a shell command. We already have shell command execution and rendering _mostly_ implemented in the TUI for free (given our re-use of the blocklist and terminal model) but we need to add in the bash input mode (rendering, entry, etc) and shell execution entry point.

* `crates/warp_tui/src/terminal_session_view.rs` — core new behavior: shell submit routing, PTY-busy rejection, conversation cancellation, footer hint priority.
* `crates/warp_tui/src/input/view.rs` — shell-mode entry/exit (!/backspace/esc)

## Testing

### Screenshots / Videos

https://www.loom.com/share/97a50459c6764668a3c9aced781621e8

## Agent Mode


harryalbert:
* **#13442** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13406** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13406?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13406?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13417** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13417?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

app/src/terminal/terminal_manager.rs
```diff
@@ -43,10 +43,42 @@ impl warpui::Entity for Box<dyn TerminalManager> {
...
+    /// terminal-spacing style, the rendered Warp prompt's height, and the
+    /// debug memory-stats footer toggle.
+    pub(super) fn from_settings(ctx: &AppContext) -> Self {
+        let appearance = Appearance::as_ref(ctx);
+        let terminal_spacing =
+            TerminalSettings::as_ref(ctx).terminal_spacing(appearance.line_height_ratio(), ctx);
+        Self {
+            block_padding: terminal_spacing.block_padding,
+            warp_prompt_height_lines: WARP_PROMPT_HEIGHT_LINES,
+            show_memory_stats: DebugSettings::as_ref(ctx).should_show_memory_stats(),
```
kevinyang372, line 71: Should we actually show memory stats? I think these are only really relevant for GUI

crates/warp_tui/src/input/view.rs
```diff
@@ -537,6 +558,9 @@ pub enum TuiInputAction {
 pub struct TuiInputView {
     /// The backing code editor in char-cell (terminal) mode.
     model: ModelHandle<CodeEditorModel>,
+    /// Shared input-mode state; `None` disables shell-mode handling entirely
+    /// (e.g. in the standalone input demo).
+    input_mode: Option<ModelHandle<BlocklistAIInputModel>>,
```
kevinyang372: I thought I already removed this -- can we remove the Option here? Shell-mode should always exist

crates/warp_tui/src/input/view.rs
```diff
@@ -557,23 +584,43 @@ impl TuiInputView {
...
+        model: ModelHandle<CodeEditorModel>,
+        input_mode: Option<ModelHandle<BlocklistAIInputModel>>,
+        ctx: &mut ViewContext<Self>,
+    ) -> Self {
         ctx.subscribe_to_model(&model, |_, _, event, ctx| {
             if matches!(event, CodeEditorModelEvent::ContentChanged { .. }) {
                 ctx.notify();
             }
         });
+        if let Some(input_mode) = &input_mode {
```
kevinyang372: Do we need to notify on every input mode event?

crates/warp_tui/src/input/view.rs
```diff
@@ -621,6 +668,29 @@ impl TuiInputView {
...
+            self.prefix_mouse_state.clone(),
+            TuiContainer::new(TuiText::new("!").with_style(prefix_style).finish())
+                .with_padding_right(1)
+                .finish(),
+        )
+        .on_click(|event_ctx, _| {
+            event_ctx.dispatch_typed_action(TuiInputAction::SelectionStartAt {
+                offset: CharOffset::from(1),
+            });
+        });
```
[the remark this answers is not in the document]
kevinyang372: This seems valid -- wonder if we should have a `SetCursor` TuiInputAction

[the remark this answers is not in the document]
kevinyang372: Also is it true that TuiHoverable::on_click fires on mouse-down only? If so we need to fix it so it properly fires on mouse down + mouse up like GUI

crates/warp_tui/src/input/view.rs
```diff
@@ -621,6 +668,29 @@ impl TuiInputView {
...
+        TuiFlex::row()
+            .child(prefix.finish())
+            .flex_child(Box::new(self.render_element(ctx)))
```
kevinyang372: Can we rename this to `self.render_input`? Also should it return `Box<dyn TuiElement>` directly rather than relying the parent to add the Box wrapper

crates/warp_tui/src/agent_block.rs
```diff
@@ -257,11 +260,19 @@ impl TuiAIBlock {
...
+            } else {
+                column.add_child(element);
+            }
         }
-        column.finish()
+        // Blocks space themselves with blank rows on top — the same
+        // `BLOCK_TOP_PADDING_ROWS` baked into terminal block heights — so
+        // every adjacent block pair (terminal or agent) is separated by
+        // exactly that many rows.
+        Box::new(TuiContainer::new(column.finish()).with_padding_top(BLOCK_TOP_PADDING_ROWS))
```
kevinyang372: .finish()?

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -358,6 +396,100 @@ impl TuiTerminalSessionView {
...
+    /// Whether the input is in `!` shell mode (locked shell input).
+    fn is_shell_mode(&self, ctx: &AppContext) -> bool {
+        input_mode_policy::is_shell_mode(self.ai_input_model.as_ref(ctx))
+    }
+
+    /// Routes a submission to shell execution or the agent conversation based
+    /// on the input mode.
+    fn handle_submitted(&mut self, text: String, ctx: &mut ViewContext<Self>) {
+        if self.is_shell_mode(ctx) {
+            self.execute_user_command(text.trim(), ctx);
```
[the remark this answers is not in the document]
kevinyang372: seems valid

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -252,9 +264,27 @@ impl TuiTerminalSessionView {
...
             exit_confirmation: ExitConfirmation::default(),
+            ai_input_model,
+            terminal_model: model,
+            transient_hint: TransientHint::default(),
         }
     }
 
+    /// Displays `text` in the footer's hint slot for
+    /// [`TRANSIENT_HINT_DURATION`], then reverts to the persistent content.
+    fn show_transient_hint(&mut self, text: String, ctx: &mut ViewContext<Self>) {
```
kevinyang372, line 280: Instead of spawning N background threads of N transient hints, should the latter hints abort the timer thread for the earlier hints + start a new timer? This ways we could enforce 1 thread for all the transient hints

This will also allow us to get rid of the generation parameter

kevinyang372 (approved):

app/src/terminal/terminal_manager.rs
harryalbert: the TUI doesn't show memory stats (see `TRANSCRIPT_BLOCK_SPACING` in `crates/warp_tui/src/transcript_view.rs`). This is just serving as the default block spacing that the GUI uses

harryalbert: I can give the fn a better name so that's clearer

crates/warp_tui/src/input/view.rs
harryalbert: can do

crates/warp_tui/src/input/view.rs
harryalbert: yeah I think so. The only input mode events are `InputTypeChanged` and `LockChanged`, both of which are worth refreshing on (`LockChanged` shouldn't really be emitted for the TUI, but there's no reason _not_ to refresh on it given that).

The input model itself also has guards to only emit when things actually changed.

crates/warp_tui/src/input/view.rs
harryalbert: yeah that makes sense to me. I'll update

crates/warp_tui/src/input/view.rs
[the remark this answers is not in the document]
harryalbert: yeah TUI hoverable fires on mouse down – will update it so it follows the GUI semantics.

and agree w/ the `SetCursor` action, that feels like a clean fix. I'll add both

crates/warp_tui/src/terminal_session_view.rs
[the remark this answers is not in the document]
harryalbert: agreed, I'll fix

crates/warp_tui/src/terminal_session_view.rs
harryalbert: yeah good call

PR #13418, opened 2026-07-07
Add per-tool-call labels for TUI tool calls

harryalbert:
## Description

Before this PR, we were just rendering a generic "tool call executed" row whenever the agent executed a tool call. For some tool calls we'll want specific views with more detailed state (i.e. the ask questions tool, shell command execution, diffs, etc.), but for many others (like grep) we'll just want some stateful text like "grepping for ..." -> "grepped for ... w/ ... results".

This PR adds in that stateful, differentiated text for each tool call (besides file edits, which already have a special separate rendering path). As we add more specific rendering for some tool calls, we can remove them from the text match arm, but for now our text match covers all tool calls besides file edits.

## Files to look at
* `specs/CODE-1815/TECH.md`: gives a general overview of the PR's structure
* `crates/warp_tui/src/agent_block.rs`: the main thing to notice here is that, specifically for shell commands, we're referencing the command block state itself for the shell command header, instead of referencing the action model to get the action state. This is because agent-run shell commands are actually composed of multiple different tool calls + results (i.e. a shell command request, which is responded to with an LRC snapshot, and then a request when the shell command is finished to get the contents of the command). This is following GUI prior art
* `crates/warp_tui/src/tool_call_labels.rs`: this is very skim-able, but it's worth reading over the fns offered (and then skimming the giant match arm where we assign different labels for each (tool call, state) combination).

## Testing

### Screenshots / Videos

https://www.loom.com/share/48bdbf10bce44417aa1e21c38f3231cd

## Agent Mode

### DIFF
--- a/crates/warp_tui/src/agent_block_sections.rs
+++ b/crates/warp_tui/src/agent_block_sections.rs
@@ -7,12 +7,13 @@
 
 use std::time::Duration;
 
-use warp::tui_export::{format_elapsed_seconds, MessageId};
+use warp::tui_export::{format_elapsed_seconds, AIActionStatus, AIAgentAction, MessageId};
 use warpui_core::elements::tui::{TuiContainer, TuiElement, TuiFlex, TuiText};
 use warpui_core::elements::CrossAxisAlignment;
 use warpui_core::AppContext;
 
 use crate::agent_block::ThinkingBlockStates;
+use crate::tool_call_labels::{tool_call_display_state, tool_call_label, ToolCallDisplayState};
 use crate::tui_builder::TuiUiBuilder;
 
 const INPUT_PREFIX: &str = "≫ ";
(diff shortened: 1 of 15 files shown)


harryalbert:
* **#13418** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13418?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13418?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13332** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13332?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13302** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13302?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/warp_tui/src/agent_block_sections.rs
```diff
@@ -48,12 +51,37 @@ pub(crate) fn render_plain_text_section(text: &str, app: &AppContext) -> Box<dyn
...
-    TuiText::new("executed a tool call")
-        .with_style(TuiUiBuilder::from_app(app).dim_text_style())
-        .finish()
+/// Renders a one-line status row for an agent tool call: per-tool, per-state
+/// text with de-emphasized styling for non-final states and cancellations,
+/// and error styling for failures. `output_streaming` marks tool calls whose
+/// arguments are still streaming in (see `ToolCallDisplayState::Constructing`);
+/// `block_state` carries the terminal block's ground truth for shell-command
+/// tool calls (see `CommandBlockState`).
+pub(crate) fn render_tool_call_section(
```
kevinyang372: Can we actually try to get closer to the mock / our GUI parity here? https://www.figma.com/design/yg5nbPZuGoAszHS3Rhvehu/TUI?node-id=323-17553&t=ZINACTiPr1Rk74Dn-0

(check mark for success, yellow dot for running, grey block for cancelled, red x for failed...)

kevinyang372: We should also makes it clear in the naming that this is a fallback / plain text tool call status instead of the richer view based action status

kevinyang372 (approved):

crates/warp_tui/src/agent_block_sections.rs
harryalbert: yeah for sure. I'll update

PR #13433, opened 2026-07-07
Exclude directories from Command Palette file search so files aren't starved from the result cap

### LINKED ISSUE #12391: Bug: Files might be missing from command palette when too many directories match the search query

maxmilian:
## Description

Command Palette file search (the file filter / `Cmd+O`) could return **no files** when the query matched many directories. The repo-metadata traversal caps results at a fixed budget, and directory entries matching the query were consuming those slots before any files were reached — so a query like a common folder name surfaced only directories (or nothing useful), never the files the user was searching for.

This routes the Command Palette file filter through a new **file-only** contents path that excludes directories from the traversal, so the result cap is spent on file matches. The fix is deliberately scoped to that one surface:

- `FileSearchModel::get_repo_file_contents()` is a new sibling of `get_repo_contents()` that traverses with `include_folders = false`. Both delegate to a shared `get_repo_contents_with_options(query, include_folders, app)`.
- `contents_args()` applies `GetContentsArgs::exclude_folders()` when `include_folders` is false, so folders are dropped *during* traversal (before the cap), not filtered out afterward.
- Only the Command Palette file `data_source` query path is rerouted to `get_repo_file_contents()`. Callers that legitimately surface directories (e.g. the AI context menu, and the zero-state git-status listing) keep using `get_repo_contents()` unchanged — a blanket `.exclude_folders()` would have regressed them.
- The file-only path is uncached (like the existing query path), since the shared per-repo cache holds the directory-inclusive contents.
- WASM (`not(feature = "local_fs")`) gets a matching stub.

Building on the diagnosis and approach from the earlier PRs #12440 (@amlndz — root cause) and #12894 (@SkyNotSilent — the file-only design), both of which were auto-closed for inactivity; this relands that approach against current `main` and adds the before/after recording that was the one outstanding review request.

## Linked Issue

Closes #12391

## Testing
- [x] `cargo fmt -- --check`
- [x] `cargo check -p warp`
- [x] `cargo clippy -p warp --tests -- -D warnings` (clean)
- [x] `cargo nextest run -p warp file_search_model_tests` — 16 passed

<sub>Note: `--all-features` clippy surfaces some pre-existing `dead_code`/`unused_imports` warnings in unrelated SSH/session-sharing modules that are present on `main` independent of this change; the changed files are clippy-clean.</sub>

### Screenshots / Videos

Reproduced in a repo whose matching entries are a deeply-nested chain of directories all named `widget` (every one matches the query), with three files at the bottom of the chain (`widget_alpha_FOUND.txt`, `widget_beta_FOUND.txt`, `widget_gamma_FOUND.txt`). Searching `widget` in the Command Palette file filter:

**Before** — the matching directories consume the result cap during traversal, so the three files at the bottom of the chain are never reached and the search returns nothing:

https://github.com/user-attachments/assets/e91d2623-35f9-481e-adc6-19cf6e9ca189

**After** — directories are excluded from the file-search traversal, so it descends to the files and all three are returned:

https://github.com/user-attachments/assets/c2bd868e-db27-4d77-9333-35488a602212

## Agent Mode

### DIFF
--- a/app/src/search/command_palette/files/data_source.rs
+++ b/app/src/search/command_palette/files/data_source.rs
@@ -113,9 +113,6 @@ impl FileDataSource {
         match &self.mode {
             FileDataSourceMode::Repo => {
                 let file_search_model = FileSearchModel::as_ref(app);
-                // The Command Palette file filter only opens files, so exclude
-                // directories: otherwise directories matching the query consume
-                // the repo-metadata result cap and starve file matches (#12391).
                 file_search_model.get_repo_file_contents(query, app)
             }
             FileDataSourceMode::CurrentFolder { cached_contents } => {
--- a/app/src/search/command_palette/files/data_source.rs
+++ b/app/src/search/command_palette/files/data_source.rs
@@ -113,7 +113,10 @@ impl FileDataSource {
         match &self.mode {
             FileDataSourceMode::Repo => {
                 let file_search_model = FileSearchModel::as_ref(app);
-                file_search_model.get_repo_contents(query, app)
+                // The Command Palette file filter only opens files, so exclude
+                // directories: otherwise directories matching the query consume
+                // the repo-metadata result cap and starve file matches (#12391).
+                file_search_model.get_repo_file_contents(query, app)
             }
             FileDataSourceMode::CurrentFolder { cached_contents } => {
                 Arc::new(cached_contents.clone())
--- a/app/src/search/files/model.rs
+++ b/app/src/search/files/model.rs
@@ -201,14 +201,6 @@ impl FileSearchModel {
         self.get_repo_contents_with_options(query, true, app)
     }
 
-    /// Gets repository files (no directories) for the current working directory.
-    ///
-    /// Intended for search surfaces that only ever open files and cannot act on
-    /// directory entries, such as the Command Palette file filter. Excluding
-    /// directories keeps them from consuming the repo-metadata result cap, which
-    /// otherwise starves file matches when many directories match the query.
-    /// Kept separate from [`Self::get_repo_contents`] so callers
-    /// that do surface directories (e.g. the AI context menu) are unaffected.
     #[cfg(feature = "local_fs")]
     pub fn get_repo_file_contents(
         &self,
--- a/app/src/search/files/model.rs
+++ b/app/src/search/files/model.rs
@@ -198,21 +198,50 @@ impl FileSearchModel {
     /// root location, invalidated when the file tree changes.
     #[cfg(feature = "local_fs")]
     pub fn get_repo_contents(&self, query: &str, app: &AppContext) -> Arc<Vec<FileSearchResult>> {
+        self.get_repo_contents_with_options(query, true, app)
+    }
+
+    /// Gets repository files (no directories) for the current working directory.
+    ///
+    /// Intended for search surfaces that only ever open files and cannot act on
+    /// directory entries, such as the Command Palette file filter. Excluding
+    /// directories keeps them from consuming the repo-metadata result cap, which
+    /// otherwise starves file matches when many directories match the query
+    /// (see #12391). Kept separate from [`Self::get_repo_contents`] so callers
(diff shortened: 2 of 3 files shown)


app/src/search/files/model.rs
```diff
@@ -198,21 +198,50 @@ impl FileSearchModel {
...
+    }
+
+    /// Gets repository files (no directories) for the current working directory.
+    ///
+    /// Intended for search surfaces that only ever open files and cannot act on
+    /// directory entries, such as the Command Palette file filter. Excluding
+    /// directories keeps them from consuming the repo-metadata result cap, which
+    /// otherwise starves file matches when many directories match the query
+    /// (see #12391). Kept separate from [`Self::get_repo_contents`] so callers
+    /// that do surface directories (e.g. the AI context menu) are unaffected.
```
acarl005:
```suggestion
    /// otherwise starves file matches when many directories match the query.
```

Comments only need to reflect the current state of the code.

app/src/search/command_palette/files/data_source.rs
```diff
@@ -113,7 +113,10 @@ impl FileDataSource {
         match &self.mode {
             FileDataSourceMode::Repo => {
                 let file_search_model = FileSearchModel::as_ref(app);
-                file_search_model.get_repo_contents(query, app)
+                // The Command Palette file filter only opens files, so exclude
+                // directories: otherwise directories matching the query consume
+                // the repo-metadata result cap and starve file matches (#12391).
```
acarl005:
```suggestion
```

Unnecessary comment.

app/src/search/files/model_tests.rs
```diff
@@ -152,6 +152,30 @@ mod file_search_model_tests {
...
+
+        let empty = FileSearchModel::contents_args("", true, |_| Some("src/main.rs".into()));
+        assert!(empty.include_folders);
+    }
+
+    #[cfg(feature = "local_fs")]
+    #[test]
+    fn test_contents_args_excludes_folders_for_file_only_search() {
+        // The Command Palette file filter opts out of folders so directory
+        // matches don't consume the result cap and starve files (#12391).
```
acarl005:
```suggestion
```

acarl005 (changes requested):
Thanks for the fix. Requesting cleanup of unnecessary comments with inline suggestions. Once those are addressed I'll approve.

acarl005:
@oz-agent Please apply the suggested edits in my PR review

app/src/search/files/model.rs
```diff
@@ -198,21 +198,50 @@ impl FileSearchModel {
...
+    }
+
+    /// Gets repository files (no directories) for the current working directory.
+    ///
+    /// Intended for search surfaces that only ever open files and cannot act on
+    /// directory entries, such as the Command Palette file filter. Excluding
+    /// directories keeps them from consuming the repo-metadata result cap, which
+    /// otherwise starves file matches when many directories match the query.
+    /// Kept separate from [`Self::get_repo_contents`] so callers
+    /// that do surface directories (e.g. the AI context menu) are unaffected.
```
acarl005:
```suggestion
```

Delete this

acarl005 (changes requested):

acarl005:
@oz-agent Perform the edit I requested in my most recent review

acarl005:
@oz-agent I only asked you to delete the last 2 lines of the doc comment but you deleted the entire doc comment.

acarl005 (approved):

acarl005:
@cla-bot check

PR #13435, opened 2026-07-07
Add auto save setting for the Warp text editor

szgupta:
Demo: https://www.loom.com/share/e61d6e666caa41deb9171c1cbbcc07c3

## Description
Adds an opt-in **Auto save** setting for the Warp text editor. When enabled, the editor saves changes automatically so you don't have to press Cmd-S.

Behavior mirrors VS Code:
- **While typing** — debounced ~1s after you stop typing (VS Code's `files.autoSave: afterDelay`). Saves **without** running the LSP formatter, so formatting never disrupts you mid-edit.
- **On focus loss** — when the editor loses focus to another view in the app (`onFocusChange`) or its window is navigated away from / the app is deactivated (`onWindowChange`). These **honor** the existing **Format on save** setting, exactly like a manual Cmd-S.

Scope & guards:
- Applies to normal file editing **and the code review panel's editable diffs** (both are `LocalCodeEditorView` with `diff_type = None`).
- **Never** auto-saves the agent "edit-file" accept/reject diffs (a separate `InlineDiffView` type; plus a `diff_type.is_some()` guard).
- Skips untitled files (no path), disconnected remotes, and buffers with no unsaved changes.
- Defaults to **off**; cloud-synced like other editor settings.

Unobtrusive behavior when auto-save is on:
- The **"File saved." toast is suppressed** for auto-saves (still shown for manual Cmd-S), via an `auto_saved` flag on `LocalCodeEditorEvent::FileSaved`.
- The **unsaved-changes dot is hidden** at every surface (editor tab bar, single-tab pane header, workspace tab strip, and code review diff headers), so it no longer flickers as you type.
- **Closing never blocks on an "unsaved changes" dialog for files that can be auto-saved.** Instead the pending edits are flushed silently and the close proceeds. This covers the file/editor tab, the code pane, the workspace tab(s), the window, app quit, and the code review pane. The close/quit dialog still appears for things auto-save can't handle — running processes, shared sessions, or unsaved buffers with **no backing file** (e.g. untitled files) — so those are never silently discarded.

The setting lives at **Settings → Code → Code Editor and Review → Auto save** (`code.editor.auto_save`).

### How
- New `auto_save` bool in `CodeSettings` (`app/src/settings/code.rs`), mirroring `format_on_save`.
- Toggle widget + `ToggleAutoSave` action in `app/src/settings_view/code_page.rs`.
- Auto-save logic in `app/src/code/local_code_editor.rs`: a debounced channel (afterDelay, no formatting), `View::on_blur` (focus loss), and a `WindowManager` subscription (window navigation). Both focus paths respect format-on-save.
- Indicator/toast suppression in `app/src/code/view.rs`, `app/src/tab.rs`, and `app/src/code_review/code_review_view.rs`.
- Close/quit handling centralized in `UnsavedStateSummary::save_unsaved_code_and_should_warn` (`app/src/quit_warning/mod.rs`), backed by `CodeView::auto_save_all_unsaved_tabs` and `CodeReviewView::auto_save_all_unsaved_files`. Wired into the window/app close (`app/src/lib.rs`), pane close (`app/src/pane_group/mod.rs`), tab close (`app/src/workspace/view.rs`), the editor tab close (`app/src/code/view.rs`), and the code review close (`app/src/code_review/code_review_view.rs`).

## Linked Issue
_None._

## Testing

Validated with `cargo fmt` and `cargo clippy -p warp --lib -- -D warnings` (both pass) after merging latest `master`. No automated tests were added: auto-save is driven by debounce timers plus focus/window/close lifecycle events that are awkward to unit test, and the underlying save path (`save_local` / `perform_save`) is unchanged and already exercised. Manual verification is recommended before merge.

### Screenshots / Videos
_N/A — settings toggle plus editor behavior. Happy to add a screen recording._

## Agent Mode

CHANGELOG-IMPROVEMENT: Added an opt-in "Auto save" setting for the Warp text editor (Settings > Code > Code Editor and Review) that saves changes as you type and when the editor loses focus, and flushes edits automatically when closing instead of prompting.

---
[Warp conversation](https://staging.warp.dev/conversation/d9f632e9-8c0c-4f27-8c0c-a7fa791ad0f8)


app/src/code/local_code_editor.rs
```diff
@@ -334,6 +348,17 @@ impl LocalCodeEditorView {
...
                     ctx.emit(LocalCodeEditorEvent::UserEdited);
+
+                    // Queue a debounced auto-save while the user types. The
+                    // debounced path saves without running the LSP formatter,
+                    // matching VS Code's `files.autoSave: afterDelay` behavior.
+                    // A `Some` `diff_type` means this editor is showing a
+                    // pending accept/reject diff (e.g. an agent "edit-file"
+                    // proposal), which must never auto-save. Editable
+                    // code-review diffs use `diff_type = None` and stay eligible.
+                    if me.diff_type.is_none() && *CodeSettings::as_ref(ctx).auto_save {
```
kevinyang372, line 362: very much a nit -- we should queue a debounced auto-save after user manually saves as well?

app/src/code/local_code_editor.rs
```diff
@@ -287,6 +295,12 @@ pub struct LocalCodeEditorView {
     context_menu_state: ContextMenuState,
     /// Channel for debouncing hover requests.
     hover_debounce_tx: async_channel::Sender<CharOffset>,
+    /// Channel for debouncing auto-save requests while the user types.
+    auto_save_debounce_tx: async_channel::Sender<()>,
+    /// Set while an auto-save (debounced or focus-change) is in flight so the
+    /// resulting `FileSaved` event can be marked as an auto-save and suppress
+    /// the success toast.
+    auto_save_in_flight: bool,
```
kevinyang372, line 306: Should we have a general save state? Something like
```
enum SaveState {
      Saving { is_auto_save: bool },
      None
}
```

This way we could also do things like early return on save attempts when the editor is already saving

app/src/code/view.rs
```diff
@@ -985,6 +990,44 @@ impl CodeView {
...
+    /// flicker on and off as the user types. In that case we hide it entirely.
+    fn show_unsaved_indicator(tab: &TabData, app: &AppContext) -> bool {
+        !*CodeSettings::as_ref(app).auto_save && Self::has_unsaved_changes(tab, app)
+    }
+
+    /// Flush-saves every unsaved tab that has a backing file, marking each save
+    /// as an auto-save so it stays silent (no "File saved." toast). Returns
+    /// `true` if any unsaved tab could NOT be auto-saved (e.g. an untitled
+    /// buffer with no path), so callers can still warn before discarding those.
+    pub fn auto_save_all_unsaved_tabs(&mut self, ctx: &mut ViewContext<Self>) -> bool {
```
kevinyang372, line 1019: CodeView wraps LocalCodeEditorView right? In that case, do we need an ad-hoc autosave logic for the parent view (I assume the child editors will handle auto-save themselves)?

app/src/code_review/code_review_view.rs
```diff
@@ -6257,6 +6257,28 @@ impl CodeReviewView {
         }
     }
 
+    /// Flush-saves every unsaved file in the review, marking each save as an
```
kevinyang372, line 6296: ditto -- I assume the child editor views will handle auto-save here?

app/src/code/local_code_editor.rs
szgupta: [Warp Agent] Queuing a debounced auto-save right after a manual save would end up a no-op: `auto_save_after_delay` early-returns when `has_unsaved_changes` is false, and any edit made after the manual save re-queues the debounce via the `ContentChanged` handler anyway. So the current wiring already covers this — leaving as-is.

app/src/code/local_code_editor.rs
szgupta: [Warp Agent] Considered this — the subtlety is that `auto_save_in_flight` really means "label the *next* save as an auto-save", not "a save is in flight": the close flows set it via `mark_next_save_as_auto_save()` before separately invoking the save, so `SaveState::Saving { is_auto_save }` would conflate labeling with in-flight tracking. The early-return-on-saving idea is also a behavior change: saves complete asynchronously via `GlobalBufferModel` events, so a manual cmd-s during an in-flight auto-save would get dropped, and a stuck `Saving` state would silently swallow future saves (re-saving today is idempotent/harmless). Deferring the refactor to a follow-up rather than bundling it here — happy to pick it up if you feel strongly.

app/src/code/view.rs
szgupta: [Warp Agent] The child editors auto-save only on the typing debounce (1s) and focus change; `auto_save_all_unsaved_tabs` is the close/quit-time flush (called from `UnsavedStateSummary::save_unsaved_code_and_should_warn`). At close, a pending debounce hasn't fired yet and no blur may occur before teardown, so edits from the last second would be lost without it. The parent also needs to aggregate across all tabs to decide whether anything remains that auto-save *can't* persist (untitled buffers, disconnected remotes) and still warrant the warning dialog. The actual save is still delegated to the child editor — this is just tab orchestration.

app/src/code_review/code_review_view.rs
szgupta: [Warp Agent] Same rationale as `CodeView`: this is the close-time flush (called from `CodeReviewView::close`). The child editors can't know the review pane is closing, and the parent owns the unsaved-file aggregation and the close-vs-prompt decision.

szgupta:
[Warp Agent] Pushed 348b260ae fixing a bug in the unsaved-indicator suppression: with auto-save enabled, the unsaved-changes indicator (editor tab dot, workspace tab indicator, and code review per-file dot) was hidden unconditionally — including for untitled buffers and disconnected remotes that auto-save can't actually persist.

The fix adds `LocalCodeEditorView::can_auto_save()` (backing `file_id` present + remote host still connected, reused by the close-time flush in `auto_save_all_unsaved_tabs`) and makes all three indicator sites auto-save-aware: the dot is now hidden only for changes auto-save can persist, so unsaveable changes stay visible.

kevinyang372 (approved):

PR #13436, opened 2026-07-07
[VA] Add blocklist UI for CU video recordings

vkodithala:
## Description

Adds a dedicated UI for computer-use video recordings produced by `start_recording` and `stop_recording` tool calls. This is rendered in the blocklist wherever video artifacts are enabled (for now, this is Cloud Mode/transcripts from Oz runs with CU enabled).

### What changed

- Renders a compact `Recording started` card when the agent starts a recording.
  - The card uses a static record icon and a short summary line derived from the conversation title for now. The upstack [PR](https://github.com/warpdotdev/warp/pull/13437) switches this to the LLM-provided recording summary, so the conversation title is a stopgap.
  - Capture dimensions and redundant status text are intentionally hidden because they are not meaningful to users.
- Derives recording spans from the conversation transcript and action results (`AIConversation::recording_spans_by_action_id`).
  - Single ordered walk with buffer-and-flush semantics: a successful `StartRecordingResult` opens a span; the start action and any `UseComputer` actions inside the open span are buffered; a matching successful `StopRecordingResult` closes the span (attaching the published artifact UID) and flushes the buffer into the map; a failed or cancelled stop drops the buffer, so actions are never shown as captured when no recording was saved; a span still open at the end of the scan is flushed as open so live recordings decorate their rows.
  - Action results are resolved transcript-first: results embedded in exchange inputs (the source of truth for restored/cloud conversations) are collected once up front, and `BlocklistAIActionModel` is only a fallback for live results that haven't been drained into a follow-up request's inputs yet.
- Decorates recorded computer-use interaction rows with `Captured in recording` (or `Recording active` while the span is open).
  - Screenshot-only `UseComputer` rows are excluded; they are captures, not user-visible interactions. Detection is semantic (`Action::is_no_op`): agents that only want a screenshot emit a zero-duration wait plus screenshot params rather than an empty action list, so rows whose actions are all no-op waits are not decorated, while real waits and interactions are.
- Renders a compact `Recording saved` card when the recording is published, showing the video duration in `m:ss` format and a single `Open recording` button. Partial recordings render as `Partial recording • m:ss`. Raw capture dimensions, size, completion status, and internal termination reason strings are intentionally hidden.
- Wires `Open recording` through the artifact download endpoint to mint a fresh signed URL and open it.
- Adds an env-var-gated mock recorder to `computer_use` (macOS only): `WARP_MOCK_RECORDER=1` routes `create_recorder()` to a mock whose `stop` publishes the MP4 at `WARP_MOCK_RECORDING_FIXTURE` (`WARP_MOCK_RECORDING_STOPPED_EARLY=1` exercises the partial-recording copy). This lets the recording UI be exercised on macOS, where real capture is unsupported. Decided to commit this here since it might be useful to test out future UI changes and is effectively a no-op when the envvar is unset; lmk if you think we should remove this from the PR, it's OOS and don't care _too_ much that this gets in!

### Mechanics

- Recording card state is driven from transcript data rather than live `RecordingController` state so restored/cloud conversations render consistently.
- Span computation is gated per block: `AIBlock::handle_updated_output` computes `has_recording_related_actions` whenever the output changes (including at construction for restored/complete blocks), and rendering only derives the conversation-wide span map for blocks whose output contains `StartRecording`/`StopRecording`/`UseComputer` actions. Non-recording blocks pay one bool check per render.

### Known issues / caveats

- **Per-render span recomputation for recording blocks** (`TODO(vkodithala)` at the computation site in `output.rs`): blocks that do contain recording-related actions still recompute the conversation-wide span map on every render. The cost is bounded — derivation is a single pass over the transcript (the per-lookup exchange re-scan is gone) and non-recording blocks skip it entirely via the gate bit — but the intended fix is caching spans on `BlocklistAIActionModel` keyed by conversation and refreshing on action/result mutations. Deferred because the cache's inputs span two models (transcript ordering from `BlocklistAIHistoryModel`, live results from the action model), restore hydrates the action model before conversations land in the history model, and transcript rewrites (rewind/revert, fork, summarization message moves, remote-child transcript merges) each need dirty hooks — any missed hook is a silent stale-footer bug, so it deserves its own PR with mock-recorder verification.
- **Global action-ID keying in `BlocklistAIActionModel`**: `get_action_result` searches all conversations' results, and some providers only guarantee tool-call IDs unique within a conversation. Mitigated here by resolving results transcript-first (conversation-scoped by construction); the action model is only consulted for live, not-yet-persisted results.
- **Failed turns lose recording decorations**: `FinishedAIAgentOutput::output()` returns `None` for `Error`, so a failed turn's partial output is excluded from span derivation even though the block may still render it. Accepted: the turn itself errored and the decoration carries little value there; documented on `recording_spans_by_action_id`.
- **Transient state while a stop is pending**: between the `stop_recording` call and its result, prior CU rows still show `Recording active` while the stop card shows `Saving recording`. Self-corrects when the result arrives.
- **Stop failures show generic copy**: `Recording could not be saved` intentionally hides raw error/termination strings; `TODO(vkodithala)` to switch to typed, user-facing termination copy once finalization emits structured reasons.

## Testing

Automated validation run locally:

- `./script/format` / `git --no-pager diff --check`
- `MACOSX_DEPLOYMENT_TARGET=10.14 cargo nextest run -E 'test(recording_span_) | test(stop_recording_card_text_) | test(use_computer_decoration_)'` (7 passed)
- `MACOSX_DEPLOYMENT_TARGET=10.14 cargo clippy -p computer_use -p warp --all-targets --tests -- -D warnings`
- `MACOSX_DEPLOYMENT_TARGET=10.14 cargo check -p computer_use --features test-util`

Tests cover:

- Recording span derivation for successful, open, failed-start, mismatched-stop, and failed-stop flows.
- Excluding screenshot-only `UseComputer` rows from recording decoration.
- Partial recording duration copy without surfacing raw termination reasons.

Manual verification on macOS via the mock recorder: run the client with `./script/run --features with_local_server,with_local_session_sharing_server --host-id local-dev` against `warp-server/script/oz-local`, passing `-e WARP_MOCK_RECORDER=1 -e WARP_MOCK_RECORDING_FIXTURE=<mp4>` so the direct-backend worker uses the mock recorder.

### Screenshots / Videos

Here's a demo of how this works end-to-end: https://www.loom.com/share/51cd1f3419274f7eabd5db652df6b051.

## Agent Mode

---

Conversation: https://staging.warp.dev/conversation/2ff5857d-5a45-4034-8eef-c9aa21cd674f
Run: https://oz.staging.warp.dev/runs/019f3978-70f1-7342-af74-24ec3c784f9f


vkodithala:
* **#13437** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13437?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13436** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13436?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13436?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/agent/conversation.rs
```diff
@@ -1703,6 +1718,116 @@ impl AIConversation {
...
+                                    stopped_span.artifact_uid = Some(stopped.artifact_uid.clone());
+                                    true
+                                }
+                                Some(AIAgentActionResultType::StopRecording(
+                                    StopRecordingResult::Error(_),
+                                ))
+                                | Some(AIAgentActionResultType::StopRecording(
+                                    StopRecordingResult::Cancelled,
+                                )) => {
+                                    stopped_span.stop_action_id = Some(action.id.clone());
```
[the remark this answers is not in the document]
vkodithala: Resolved in latest; we now only decorate CU actions from in-progress or successful recordings with a "Captured in recording" subtext in UI.

app/src/ai/agent/conversation.rs
```diff
@@ -90,6 +91,20 @@ impl TodoStatus {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RecordingSpanInfo {
+    pub recording_id: String,
+    pub start_action_id: AIAgentActionId,
+    pub stop_action_id: Option<AIAgentActionId>,
+    pub artifact_uid: Option<String>,
+}
```
danielpeng2, line 98: non-blocking question: do we need all of these fields in the span? my mental model is the span is basically used to determine whether each action is included in a finished recording.

it's relatively cheap to store this data though, so not a huge deal

danielpeng2 (approved):

app/src/ai/agent/conversation.rs
vkodithala: Yeah, agreed here and this is good pushback. We don't actually use `start/stop` action IDs anywhere, or the artifact UID (this is resolved separately for blocklist rendering, directly from the StopRecording tcr). Reduced this in latest to a recording ID and status (in-progress/completed).

PR #13417, opened 2026-07-06
Implement `CrossAxisAlignment` options for TUI flex

harryalbert:
## Description

This PR was originally motivated by wanting to turn the input into a flex row w/ `[!][input]` (when shell mode was enabled). To do this with the existing flex API (where the flex reports its cross axis as the largest cross axis offered instead of auto-constraining the height), we would have had to manually constrain the row with a calculated height, which we could only know at layout time.

The solution was to update the TUI flex api to be aligned w/ the GUI flex API, where the flex now reports its cross axis as the largest child's cross extent, clamped to that constraint. You can still achieve the original full-constraint-reporting by using the `CrossAxisAlignment::Stretch` (using the same enum as the GUI w/ the same semantics), but you can also use the other cross axis alignment options offered to the GUI flex (i.e. start  — the default — center, or end). With this change, the TUI flex's cross axis behavior reaches feature/behavior parity w/ the GUI flex.

The PR above this one uses this new API to implement the `[!][input]` flex row for the input.

## Testing

- [x] I have manually tested my changes locally with `./script/run`

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

PR #13442, opened 2026-07-07
TUI warping indicator

harryalbert:
## Description

The TUI should show a `⋮ Warping (Ns)` indicator while the selected conversation is in progress ([Figma](https://www.figma.com/design/yg5nbPZuGoAszHS3Rhvehu/TUI?node-id=323-17216&m=dev)): a rotating spinner, a shimmering "Warping" label, and a ticking elapsed counter, rendered between the transcript and the input box. We already had the shimmer _math_ for free (the GUI's `ShimmeringTextElement`), but the TUI had no animation machinery at all — redraws are purely invalidation-driven — so the bulk of this PR is building the native repaint-scheduling mechanism (mirroring the GUI's `PaintContext::repaint_after`) that the indicator then sits on top of. Animations run as paint-only redraws over the cached element tree, so nothing re-renders views or relayouts the transcript, and the TUI stays fully idle when nothing animates.

* `crates/warpui_core/src/elements/tui/mod.rs` — the new `TuiPaintContext` with `repaint_after` + earliest-deadline-wins coalescing. This split (render/cursor now take `TuiPaintContext` instead of `TuiLayoutContext`) is what causes the mechanical signature churn across every TUI element.
* `crates/warpui_core/src/runtime/mod.rs` — the driver side: each draw surfaces the frame's earliest requested repaint deadline and schedules a single one-shot timer for it, self-sustaining while anything animates.
* `crates/warpui_core/src/elements/tui/animated.rs` — `TuiAnimated`, the reusable animation element: a repaint cadence plus a closure that rebuilds the current frame each pass (drives the spinner and counter).
* `crates/warpui_core/src/elements/tui/shimmering_text.rs` + `crates/warpui_core/src/elements/shimmer_math.rs` — the TUI shimmering text and the band math extracted from the GUI element so both backends share it.
* `crates/warp_tui/src/warping_indicator.rs` — the indicator row itself: the spinner's keyframe choreography, the shimmer label, and the elapsed counter, all phase-locked to one `AnimationClock`.
* `crates/warp_tui/src/terminal_session_view.rs` — visibility wiring: renders the row while the selected conversation is in progress, subscribed to history/selection events.

## Testing

### Screenshots / Videos

https://www.loom.com/share/60f0cc8358c3431f8233ba95380292fe

## Agent Mode


harryalbert:
* **#13442** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13417** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13417?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/warpui_core/src/presenter/tui.rs
```diff
@@ -141,18 +155,20 @@ impl TuiPresenter {
...
         // Element resolution order:
         //   1. Fresh from rendered_views (populated by invalidate() this frame).
-        //   2. Cached last_element — ONLY when rendered_views is non-empty,
-        //      meaning invalidate() was called and this view was not changed.
-        //      If rendered_views is empty (no invalidate() was called), skip
-        //      last_element: the root may be stale (e.g. view called notify()
-        //      but the caller drives the presenter standalone without the
-        //      runtime's invalidate() step).
-        //   3. Direct render fallback for callers that skip invalidate().
+        //   2. Cached last_element — ONLY when invalidate() ran this frame, so
```
kevinyang372, line 158: This is a genuine behavior change yeah? This means if a view doesn't have animation AND no view was changed, we used to not repaint and reuse the last element tree. Now we are going to redraw for that case

kevinyang372: Is there anyway we could capture if this redraw is triggered by an animation?

kevinyang372 (approved):
Nice -- had one question

crates/warpui_core/src/presenter/tui.rs
harryalbert: It is a real behavior change, but I think in the opposite direction from what you're saying. The old condition was just a bad proxy for what we actually wanted to check.

Before, step 2 reused `last_element` when `rendered_views` was non-empty. That was meant to mean "invalidate ran, so the cache is trustworthy", but it's the wrong proxy: `rendered_views` holds cached child-view elements, and step 1 just popped the root out. So:
* If the root has no child views, the map is empty at that point even though invalidate ran. Step 2 gets skipped and we fall to step 3, a full fresh render of the root view on every draw, pointlessly.
* If the root has child views, the map stays non-empty across frames (cached child elements get re-inserted after every use), so the check couldn't actually detect "invalidate didn't run", which is the case its own comment said it was guarding. In practice every caller pairs invalidate + present so that path is unreachable, but the proxy was measuring map state, not the thing we cared about.

The new flag checks the intended condition directly: reuse the cache iff `invalidate()` ran this frame.

The reason this came up while building animation was that an animation repaint is a timer-triggered draw where nothing changed. The draw still runs `invalidate()`, but with an empty set, so nothing gets re-rendered and the root is never in `rendered_views`. Step 2 is the only thing that lets us repaint the cached tree; under the old emptiness check, a childless root would have re-run its whole `render()` at the spinner's cadence, which is bad lol.

harryalbert: and per the animation redraw capture, that should already be encoded. An animation-triggered draw is exactly a draw whose invalidation set is empty, so invalidate() re-renders nothing and present() reuses the cached tree. That reuse is the animation fast path. We could thread an explicit "draw reason" from the runtime, but it would just duplicate what the invalidation set already tells us.

PR #13444, opened 2026-07-07
TUI inline diff rendering for agent file-edit tool calls (CODE-1800)

harryalbert:
## Description

`RequestFileEdits` tool calls in the TUI now render a per-file inline diff instead of the old one-line `Edited N files (+a −r)` summary: a clickable header (`✓ Updated name +a −r ▾`, verb from the diff type, renames shown as `old → new`) over a read-only body with line numbers, green added/changed lines, red removed "ghost" lines, and `… N lines` separators eliding unchanged context beyond ±3 lines around each hunk. The view is deliberately thin — the TUI analogue of the GUI's `EditorWrapper`: for each resolved `FileDiff` it seeds one char-cell `CodeEditorModel` (`reset_content(base)` → `apply_diffs(deltas)` → `hide_lines_outside_of_active_diff(3)` → `expand_diffs()`), and everything diff-shaped — hunk computation, removed-line ghosts, hidden context ranges — flows through the existing model pipeline into the core editor element from #13493. The view never walks hunks, computes hidden ranges, or builds rows. This turned out to be the first consumer combining delta-seeded content with model-side hunk-context hiding, which surfaced two shared-model fixes: a one-shot hidden-lines recalculation for diffs that compute without a content replacement, and an indexing fix in `calculate_hidden_lines` that had every GUI code-review context window skewed by one line (context+1 above, context−1 below each hunk).

- `crates/warp_tui/src/tui_file_edits_view.rs` — the diff wrapper: per-file model seeding at storage-seed time (fire-and-forget; the model completes everything itself when the async diff lands), header chrome whose `+a −r` counts read from the same computed diff that colors the body, per-file collapse with a persistent hover state, and the fallback label for actions whose storage was never seeded (failed/cancelled/restored).
- `app/src/code/editor/model.rs` — the two shared-model fixes: `pending_hidden_lines_recalculation` (version-gated so an in-flight compute of stale content can't consume it early; one-shot so per-keystroke GUI behavior is untouched; a no-op in existing GUI flows) and the `modified_lines()` 0-vs-1-indexing fix, both with regression tests.
- `crates/warp_tui/src/transcript_view.rs` + `tui_block_list_viewport_source.rs` + `crates/warpui_core/src/elements/tui/viewported_list.rs` — transcript integration so per-file bodies measure and refresh as ordinary content: agent-block roots resolve through the presenter cache as `TuiChildView` nodes in both measurement and render paths (unchanged blocks aren't re-rendered per frame), measurement threads the live layout context so child views report real heights, and exchange-output / action-status invalidations route to the owning block view.
- `crates/warp_tui/src/tui_builder.rs` — the `diff_added_style` / `diff_removed_style` theme styles.
- `specs/CODE-1800/PRODUCT.md` + `TECH.md` — the behavior contract and the design notes for the model pipeline and the shared-model fixes.

## Testing

### Screenshots / Videos
https://www.loom.com/share/55321dd145f7473a882413f4d14acf92

## Agent Mode


harryalbert:
* **#13541** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13541?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13444** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13444?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13444?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13493** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13493?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

specs/CODE-1800/PRODUCT.md
```diff
@@ -0,0 +1,71 @@
...
+## Non-goals
+
+- Editing the diff, accepting/rejecting changes, or any permissions/approval UI (the TUI has no permissions model).
+- Live token-by-token diff growth while the agent is still generating the tool call arguments.
+- Syntax highlighting and intra-line (word-level) diff highlights.
+- Keyboard navigation or keyboard toggling within the transcript.
+
+## Behavior
+
+### Per-file sections
```
kevinyang372, line 20: GUI currently keeps a title summarizing the edit's content. Should we keep that?

Image

specs/CODE-1800/PRODUCT.md
```diff
@@ -0,0 +1,71 @@
...
+10. Coloring is foreground-only (no background fills, no `+`/`-` sign column):
+    - Context lines, including their line numbers, render dim (theme bright-black).
+    - Removed lines render entirely in theme red.
+    - Added lines, including their line numbers, render entirely in theme green.
+
+11. Added and context lines show their new-file line numbers. Removed lines show no line number — their gutter renders blank — consistent with the GUI's diff rendering. (This intentionally diverges from the mock, which shows an old-file number on the removed line.)
+
+12. Created files render their full content as added (green) lines with new-file numbers. Deleted files render their full former content as removed (red) lines with blank gutters.
+
+13. Lines longer than the available width soft-wrap; there is no horizontal scrolling. Continuation rows render a blank gutter with content aligned to the content column, and keep the color of the line they continue.
```
kevinyang372, line 53: This means effectively we need to build softwrapping support into our editor? Is this the same behavior for other editors too?

specs/CODE-1800/TECH.md
```diff
@@ -0,0 +1,135 @@
...
+         ──> shared TUI row-builder ──> styled TuiText rows in TuiFileEditsView
+```
+
+## Proposed changes
+
+### Phase 1: crates/editor — implement the TODO(TUI-diff) arm
+
+In `handle_layout_action`'s `LayoutTemporaryBlock` arm (mod.rs:2641), stop no-op'ing in CharCell mode: store `temporary_blocks: Vec<CharCellTemporaryBlock>` on `CharCellState` per the TODO's own sketch (`{ content: String, insert_before: LineCount, line_decoration: Option<ColorU>, inline_decorations: Vec<(Range<usize>, ColorU)> }`), converted from incoming `TemporaryBlock`s, with a getter. Clear/replace per `add_temporary_blocks`/`reset_temporary_block` cycle mirroring the Pixels path (mod.rs:2846, mod.rs:2851). GUI path untouched; no SumTree or font-shaping changes.
+
+### Phase 2: shared TUI row-builder (crates/warp_tui)
```
kevinyang372: Why not just have a TUI re-implementation of EditorWrapper and RichTextElement?

specs/CODE-1800/PRODUCT.md
harryalbert: discussed offline

specs/CODE-1800/TECH.md
harryalbert: discussed offline

specs/CODE-1800/PRODUCT.md
harryalbert: I'll start a thread w/ erica, but IMO it's not worth blocking this PR on (very easy to change after merging)

crates/warpui_core/examples/tui_viewport_demo.rs
kevinyang372, line 1: Should we remove this demo

app/src/code/editor/model.rs
```diff
@@ -305,6 +305,13 @@ pub struct CodeEditorModel {
     hovered_symbol_range: Option<HoverableLink>,
     /// Automatically hide lines outside of the active diff with X context lines.
     hide_lines_outside_of_active_diff: Option<usize>,
+    /// One-shot request to recalculate hidden lines once a diff computes for
```
kevinyang372: I don't quite understand why we need this specifically for TUI

I think the mechanism we have for GUI is we trigger hidden lines calculation _after_ diff finished computing

crates/warp_tui/src/transcript_view.rs
```diff
@@ -71,32 +72,28 @@ impl TuiTranscriptView {
...
-                }
+                view.notify_action_owner(event.action_id(), ctx);
             },
         );
 
+        // Command tool-call rows read their backing terminal block's
+        // ground-truth state (running/finished, exit code, executed command)
+        // at render time. The presenter caches agent-block elements, so a
+        // block starting or finishing must re-render the owning agent block.
+        ctx.subscribe_to_model(model_events, |view, _, event, ctx| {
```
kevinyang372: Erm why do we need this for file edit tools?

crates/warp_tui/src/tui_block_list_viewport_source.rs
```diff
@@ -106,21 +107,31 @@ impl TuiBlockListViewportSource {
...
         width: u16,
+        ctx: &mut TuiLayoutContext,
         app: &AppContext,
     ) -> HashMap<EntityId, BlockHeight> {
         let agent_blocks = self.agent_blocks.borrow();
         view_ids
             .into_iter()
             .filter_map(|view_id| {
                 let view = agent_blocks.get(&view_id)?;
+                let height = TuiChildView::new(view)
```
kevinyang372: Why this change from `TuiAIBlock::desired_height` (which I also thought lays out the element to get its height)

crates/warp_tui/src/tui_file_edits_view.rs
```diff
@@ -1,37 +1,156 @@
...
 
+use crate::editor_element::{TuiEditorElement, TuiEditorStyles};
+use crate::tool_call_labels::{tool_call_display_state, tool_call_glyph, ToolCallDisplayState};
+use crate::tui_builder::TuiUiBuilder;
 use crate::tui_diff_storage::{TuiDiffStorage, TuiDiffStorageEvent, TuiDiffStorageHandle};
 
+/// Unchanged context lines rendered on each side of a hunk.
+const CONTEXT_LINES: usize = 3;
+/// Chevron shown on an expanded file header.
+const CHEVRON_EXPANDED: &str = "▾";
```
kevinyang372: Should this be a common constant since it will be used for a bunch of tool calls?

kevinyang372 (commented):
Have a couple of questions -- high-level looks good!

app/src/code/editor/model.rs
harryalbert: This isn’t specifically needed because the view is a TUI. It’s needed because this file-edit view combines delta-applied content with hidden-line elision, which none of the existing GUI flows do.

The GUI file-edit view also seeds the original content and applies deltas, but it displays the full file and doesn’t calculate hidden lines. GUI code review does calculate hidden lines, but its current content arrives later through ContentReplaced, which requests recalculation when it starts the diff.

Here, apply_diffs starts the first real diff through ContentChanged. That path only requests hidden-line recalculation when an edit intersects existing hidden ranges, but no ranges exist yet on the first diff. The one-shot request makes that in-flight diff initialize them when it finishes. The version check prevents an earlier diff for the seeded, unchanged content from consuming the request.

crates/warp_tui/src/transcript_view.rs
harryalbert: This subscription isn’t for file-edit events, it's just a consequence of changing agent-block measurement for the file-edit child view, rather than handling a file-edit event directly.

`TuiFileEditsView` is embedded as a `TuiChildView` inside `TuiAIBlock` (`agent_block.rs:390-395`). The old height path called TuiAIBlock::desired_height, which directly rendered the block using a new, empty layout context. That context could not resolve the nested file-edit child from the presenter, so the file-edit body was missing from the measured height.

The viewport now measures and renders the outer agent block through `TuiChildView` (tui_block_list_viewport_source.rs:125 and :350), using the presenter’s real child-view context. This also means the outer block’s element tree is cached until the block view is invalidated (which feels overall correct semantically).

Command rows inside that cached tree read running/completed state directly from terminal blocks, not solely from action status. AfterBlockStarted and BlockCompleted can therefore change a command row without invalidating TuiAIBlock. This subscription routes those two events to the owning agent block so command rows do not become stale after the child-view change.

harryalbert: I do think though that the transcript view being the one that like does this notifying feels weird. The GUI has it so that the ai block itself subscribes to these events, so I'll update the TUI to do that as well

crates/warp_tui/src/tui_block_list_viewport_source.rs
harryalbert: Yeah `desired_height` did lay out the block, but it created a new layout context with an empty rendered_views map. That was sufficient while TuiAIBlock contained only ordinary elements.

The file-edit body is a nested TuiChildView. Its layout resolves the child’s pre-rendered element from the live TuiLayoutContext; with the empty context used by desired_height, it measures as zero and the file-edit body is omitted from the block height.

The viewport now lays out the outer TuiAIBlock through TuiChildView using the presenter’s live context. That allows the nested file-edit child to resolve and measures the same cached element tree that will be rendered, rather than creating a separate temporary tree solely for measurement.

harryalbert: For context, the old like desired height code was:
```Rust
pub(super) fn desired_height(&self, width: u16, app: &AppContext) -> usize {
    let mut rendered_views = EntityIdMap::default();
    let mut ctx = TuiLayoutContext {
        rendered_views: &mut rendered_views,
    };
    let mut element = self.render_element(app);
    usize::from(
        element
            .layout(
                TuiConstraint::loose(TuiSize::new(width, u16::MAX)),
                &mut ctx,
                app,
            )
            .height,
    )
}
```

crates/warp_tui/src/tui_file_edits_view.rs
harryalbert: yeah for sure, I'll update

crates/warpui_core/examples/tui_viewport_demo.rs
harryalbert: yessir

kevinyang372 (approved):
https://staging.warp.dev/conversation/6b43718a-4279-486a-91d6-83eaae5b3294

PR #13445, opened 2026-07-07
fix hoverable hit box calculation

harryalbert:
## Description
Fix TUI hoverable hit-testing so hover and click interactions are bounded to the wrapped child's laid-out footprint instead of the parent's full slot. This prevents collapsible thinking headers from activating over trailing blank space to the right of their label and chevron.

This also centralizes child-slot calculation in `TuiFlex` and `TuiViewportedList`, ensuring rendering, cursor placement, and event dispatch use the same geometry (no behavior change, just centralizes some duplicated code).

## Testing

### Screenshots / Videos

https://www.loom.com/share/185afb9f6a1548fa9e02f0d6e4c72535

## Agent Mode

CHANGELOG-BUG-FIX: Fixed TUI hover and click targets extending into trailing blank space.

### DIFF
--- a/crates/warpui_core/src/elements/tui/hoverable.rs
+++ b/crates/warpui_core/src/elements/tui/hoverable.rs
@@ -13,33 +13,39 @@
 //! are transparent — they delegate to the wrapped child.
 //!
 //! # Dispatch policy
-//! On [`MouseMoved`](TuiEvent::MouseMoved) the pointer position is compared
-//! against this element's area; a hover transition is recorded on the handle
-//! and queues a notification so the owning view re-renders. Mouse moves are
-//! never consumed, so sibling hoverables observe their own transitions from
-//! the same event. Other events are offered to the child first; clicks use the
-//! GUI's press-then-release pairing: an unconsumed
-//! [`LeftMouseDown`](TuiEvent::LeftMouseDown) inside the area arms a pending
-//! click (recorded on the shared state, so [`MouseState::is_clicked`] styling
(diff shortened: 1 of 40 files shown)


harryalbert:
* **#13445** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13445?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13445?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warpui_core/src/elements/tui/hoverable.rs
```diff
@@ -40,6 +44,9 @@ pub struct TuiHoverable {
     child: Box<dyn TuiElement>,
     state: MouseStateHandle,
     on_click: Option<ClickCallback>,
+    /// Size from the most recent `layout`, bounding hover/click hit-testing
+    /// to the child's content rather than the parent's whole slot.
+    laid_out: Option<TuiSize>,
```
kevinyang372: Should we consider adding a TuiElement::size and TuiElement::origin trait similar to GUI?

kevinyang372 (commented):
I am a bit worried we are building on the wrong foundation here

The current approach from ZB's vibe code is we are passing in areas from root down to every element in the tree, which effectively re-performs a layout pass. We should think about following GUI's approach instead where each element is expected to cache and provide their canonical laid out area at dispatch time through Element::size and Element::origin APIs

PR #13446, opened 2026-07-07
[VA] Robust recording finalization + upload drain

vkodithala:
## Description
Centralizes computer-use video recording finalization so a recording that goes live always reaches exactly one terminal outcome (`Published`/`Discarded`/`Failed`), regardless of how the run ends. Adds a `RecordingController` state machine and one idempotent `finalize_recording` (stop → burn-in hook placeholder → upload → unconditional temp cleanup), wired into every exit:
- `stop_recording` — `StoppedByAgent`
- the cancellation funnel — `Cancelled`, for an active recording not tracked as an async action
- an ffmpeg-exit watcher — `LimitReached`/`FfmpegExited`, poll-based
- the driver run-tail — `AgentFinished`

Also adds a process-global `ArtifactUploadState` that the run tail drains (time-boxed) before teardown, generalized to recordings + `upload_artifact`.

**Companion (not in this PR):** a server cancellation-resumption prompt bullet lives uncommitted in the `warp-server` worktree; it'll be a separate warp-server PR.

**Follow-ons / known gaps (all TODOs in code):**
1. Proactive limit-notification to the agent mid-turn (v1: the agent learns on its next stop/finish) — would reuse `OrchestrationEventService`'s per-conversation input channel; recording is single-per-runtime so it may not need full multiplexing.
2. Streaming/segmented upload during capture — the only defense against a hard `MaxInstanceRuntime` TTL kill mid-active-recording.
3. Retain-and-retry the local file on upload failure.

## Linked Issue
No linked GitHub issue (internal VA recording workstream).

## Testing
- `computer_use` builds (with and without `test-util`); `cargo check -p warp --tests`
- `./script/format`; `cargo clippy -p warp -p computer_use --all-targets --tests -- -D warnings`
- Finalization `nextest`: outcome mapping, temp cleanup on `Published`/`Failed`, discard-on-cancel, upload count, drain + timeout
- Real ffmpeg capture (Linux) via `oz-local` + a custom Docker image. The linux-only recorder change needs CI/Linux to exercise real capture.
- [ ] I have manually tested my changes locally with `./script/run` (covered by automated tests + oz-local; the local app run does not exercise this cloud-agent/server path)

## Agent Mode

CHANGELOG-NONE


vkodithala:
* **#13447** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13446** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13446?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13446?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/agent_sdk/driver.rs
```diff
@@ -162,6 +169,9 @@ pub(crate) const WARP_DRIVE_SYNC_TIMEOUT: Duration = Duration::from_secs(60);
 /// If no follow-up status arrives within this window, the driver terminates with the
 /// original error so the CLI does not hang indefinitely.
 const AUTO_RESUME_TIMEOUT: Duration = Duration::from_secs(120);
+/// Time budget for draining in-flight artifact uploads (recordings, screenshots,
+/// files) during the run tail before the caller triggers process termination.
+const RECORDING_UPLOAD_DRAIN_TIMEOUT: Duration = Duration::from_secs(120);
```
vkodithala, line 174: q: Do we think 2 minutes is enough time to drain in-flight uploads? How long do we think uploads to GCS take? At the very least, we should probably document the thinking behind a 120s timeout in comments here (maybe this is just to be adherant to existing timeout semantics, or to avoid unnecessary stalling driver exit?).

app/src/ai/agent_sdk/driver.rs
```diff
@@ -818,6 +828,25 @@ impl AgentDriver {
         unregister_agent_event_consumer(conversation_id, ctx.model_id(), ctx);
     }
 
+    /// Finalizes any recording still active at the end of the run (the agent
+    /// finished without stopping it), spawning the stop+upload so the run-tail
+    /// drain can await it before the process is torn down.
+    fn finalize_active_recordings(&self, ctx: &mut ModelContext<Self>) {
+        let active =
+            RecordingController::handle(ctx).update(ctx, |controller, _| controller.take_active());
+        if let Some((_recording_id, conversation_id, handle)) = active {
```
vkodithala, line 837: nit: Is there a reason why `take_active` takes the active recording but only returns a subset of its fields? I feel like this might subtly break the semantics of this fn, i.e. future callers may expect all of `ActiveRecording` to be returned, and we can handle the full object gracefully in this fn by just ignoring the extraneous fields (seems like we already do this for `recording_id`?), right?

app/src/ai/agent_sdk/driver.rs
```diff
@@ -892,6 +921,16 @@ impl AgentDriver {
...
                 // mid-flight. Provider cleanup is just local temp-file teardown, so it's safe
                 // to run after the send.
+                // Finalize any still-active recording (agent finished without
+                // stopping it) and drain in-flight artifact uploads before signaling
+                // the caller — process termination on receipt abandons async work.
+                let _ = foreground
+                    .spawn(|me, ctx| me.finalize_active_recordings(ctx))
+                    .await;
+                ArtifactUploadState::global()
+                    .drain(RECORDING_UPLOAD_DRAIN_TIMEOUT)
```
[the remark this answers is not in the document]
vkodithala: Agreed; I think at the very least we should surface a warning in logs, via some log::info/warn, that mentions there were recording uploads in flight at the time of exit. There's some prior art in MCP startup within `AgentDriver::run`; we wait for some window, and then log an error that MCPs failed to startup within it, proceeding gracefully.

app/src/ai/blocklist/action_model/execute/stop_recording.rs
```diff
@@ -97,56 +121,25 @@ impl StopRecordingExecutor {
...
-                                width_px: output.width as i32,
-                                height_px: output.height as i32,
-                                size_bytes: output.size_bytes as i64,
-                                completion_status,
-                                termination_reason,
-                            })
-                        }
-                        Err(err) => StopRecordingResult::Error(format_stop_recording_error(&err)),
-                    }
+                    stop_recording_result_from_outcome(outcome)
```
vkodithala, line 142: nit: Why do we return a separate `RecordingTerminalOutcome` object from `finalize_recording`, and then coerce it into a `StopRecording` result? Since they map 1:1, is there a downside to returning `StopRecordingResult` directly? Might help cut down on some cruft here if feasible.

app/src/ai/blocklist/action_model/execute/upload_artifact.rs
```diff
@@ -135,9 +136,13 @@ impl UploadArtifactExecutor {
             let ai_client = ServerApiProvider::as_ref(ctx).get_ai_client();
             let server_api = ServerApiProvider::as_ref(ctx).get();
             let description = request.description.clone();
+            // Register the upload as in-flight so the driver run-tail drain waits
+            // for it before the process is torn down.
+            let upload_guard = ArtifactUploadState::global().begin();
 
             ActionExecution::new_async(
                 async move {
+                    let _upload_guard = upload_guard;
```
vkodithala, line 145: nit: My understanding here is that we pass this RAII guard into the async execution so that it's dropped when the upload resolves, and we update `in_flight`. My main question is whether there's precedent for this in the codebase already, i.e. do we already have RAII guards elsewhere? We should probably also document this via comments since it's somewhat confusing.

app/src/ai/blocklist/action_model/execute.rs
```diff
@@ -954,6 +959,30 @@ impl BlocklistAIActionExecutor {
...
+            let (token, ai_client, server_api) = recording_finalize_deps(ctx, conversation_id);
+            spawn_detached_finalize(
+                ctx,
+                handle,
+                FinalizeReason::Cancelled,
+                token,
+                ai_client,
+                server_api,
+            );
+        }
```
vkodithala, line 985: Just to clarify, the reason we don't place unresolved `StartRecording`s in async actions is because *technically*, the original `StartRecording` tc already resolved, so we're not explicitly waiting on an action result, just the corresponding `StopRecording`? OOC, how's this handled for in-progress orchestrations during conversation cancellation? Do we have any special handling there?

vkodithala: Also, as a drive-by, I wonder if we should actually be coupling finalizing active recordings with `cancel_all_running_async_actions`? Maybe we're better off defining a separate method on the executor and then calling it from `cancel_all_pending_actions`?

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -87,6 +148,81 @@ impl RecordingController {
...
+        match self.active.take() {
+            Some(active) if active.id == recording_id => {
+                Some((active.conversation_id, active.handle))
+            }
+            other => {
+                self.active = other;
+                None
+            }
+        }
+    }
```
vkodithala, line 169: nit: I wonder if we actually need this function? Even though the ffmpeg-exit watcher is scoped to a specific recording, we've enforced earlier on the controller that only 1 active recording can exist at a time. And erroring like we do in `take_handle_or_err` when this active recording doesn't match that which we want to take seems like a fine approach? Why not just collapse into `take_handle_or_err`?

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -87,6 +148,81 @@ impl RecordingController {
...
+        match self.active.take() {
+            Some(active) if active.conversation_id == conversation_id => {
+                Some((active.id, active.handle))
+            }
+            other => {
+                self.active = other;
+                None
+            }
+        }
+    }
```
vkodithala, line 189: Ditto to this, I feel like there's probably a way we could just call `take_handle_or_err`? Why is it important that we preserve `self.active` in a mismatched case when that mismatch should *never* exist since at most 1 recording can be active at a time?

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -87,6 +148,81 @@ impl RecordingController {
...
+    /// Removes and returns whatever recording is active, used by the driver
+    /// run-tail drain to finalize a still-running recording as the run ends.
+    #[cfg_attr(target_family = "wasm", allow(dead_code))]
+    pub fn take_active(
+        &mut self,
+    ) -> Option<(String, AIConversationId, computer_use::RecordingHandle)> {
+        self.active
+            .take()
+            .map(|active| (active.id, active.conversation_id, active.handle))
+    }
```
vkodithala, line 200: Same with this, I feel like we probably don't need to have this as a special function when we can just consume the handle via and clone what fields we need in caller or smt? And then can reuse `take_handle_or_err` again?

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -87,6 +148,81 @@ impl RecordingController {
...
+
+    /// Polls the active recording (if it matches `recording_id`) for an early
+    /// exit of the capture process. Cheap and non-blocking; the watcher calls
+    /// this on an interval to detect a duration/size cap hit or a crash.
+    #[cfg_attr(target_family = "wasm", allow(dead_code))]
+    pub fn poll_active_exit(&mut self, recording_id: &str) -> Option<RecordingExitKind> {
+        match self.active.as_mut() {
+            Some(active) if active.id == recording_id => active.handle.poll_exit(),
+            _ => None,
+        }
```
vkodithala, line 224: Again from a safety perspective it's probably fine to pass `recording_id` here, but I feel like we don't strictly need to since there's only 1 active recording that *can* exist?

app/src/ai/blocklist/action_model/recording_finalize.rs
```diff
@@ -0,0 +1,282 @@
...
+/// Uploads a finalized recording file as a conversation artifact. Abstracted so
+/// tests can inject a mock without a live server or HTTP stack.
+#[cfg_attr(target_family = "wasm", allow(dead_code))]
+#[async_trait]
+pub trait RecordingUploader: Send + Sync {
+    async fn upload(
+        &self,
+        request: FileArtifactUploadRequest,
+    ) -> anyhow::Result<CompletedFileArtifactUpload>;
+}
```
vkodithala, line 83: nit: Naming the uploader object `RealRecordingUploader` feels a little wrong. I wonder if we can just call it `RecordingUploader` throughout, and then add a test constructor, like `#[cfg(test)] new_for_test() {}` for something?

app/src/ai/blocklist/action_model/recording_finalize.rs
```diff
@@ -0,0 +1,282 @@
...
+            };
+            outcome.log(reason);
+            return outcome;
+        }
+    };
+
+    let local_path = output.path.clone();
+    let log_path = local_path.with_extension("log");
+
+    // TODO(recording burn-in, #3): composite click/keyboard overlays onto
```
vkodithala, line 156: Denote this as TODO(vkodithala) so it's directly tagged to me + I get to it.

app/src/ai/blocklist/action_model/recording_finalize.rs
```diff
@@ -0,0 +1,282 @@
...
+    let request = FileArtifactUploadRequest {
+        path: output.path,
+        run_id: None,
+        conversation_id: server_conversation_token,
+        description: None,
+    };
+    let upload_result = uploader.upload(request).await;
+
+    // Invariant: remove the temp mp4 + sidecar log on every terminal branch.
+    // TODO(recording upload retry, follow-on): on upload failure, retain the
```
vkodithala, line 168: Ditto here!

app/src/ai/blocklist/action_model/recording_finalize.rs
```diff
@@ -0,0 +1,282 @@
...
+        },
+    };
+    outcome.log(reason);
+    outcome
+}
+
+/// Resolves the finalize dependencies (server conversation token + server
+/// clients) from a concrete app context. Split out from [`spawn_detached_finalize`]
+/// because `SingletonEntity::as_ref` takes `&AppContext`, which only coerces from a
+/// concrete `ModelContext`, not a generic one.
```
vkodithala, line 194: Hm, can you explain this?

app/src/ai/blocklist/action_model/recording_finalize.rs
```diff
@@ -0,0 +1,282 @@
...
+    recording_id: String,
+    ctx: &mut ModelContext<RecordingController>,
+) {
+    ctx.spawn(
+        async move {
+            Timer::after(EXIT_POLL_INTERVAL).await;
+        },
+        move |controller, _output, ctx| match controller.poll_active_exit(&recording_id) {
+            Some(kind) => {
+                if let Some((conversation_id, handle)) = controller.take_by_id(&recording_id) {
```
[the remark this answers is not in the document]
vkodithala: I think this is valid. We should find an easy way to notify future stop/finish callers that a `StartRecording`'s been cancelled, as this is a deviation from spec. What's an elegant way to do this?

crates/computer_use/src/lib.rs
```diff
@@ -6,9 +6,14 @@ mod imp;
 mod noop;
 #[cfg(any(macos, linux, windows))]
 mod screenshot_utils;
+/// In-memory recorder for tests; off-Linux only, where the real capture fields
+/// on [`RecordingHandle`] are absent.
+#[cfg(all(feature = "test-util", not(linux)))]
+pub mod testing;
```
vkodithala, line 12: Do we still need this, and the corresponding `test-util` inclusion in `Cargo.toml`? What purpose does this serve?

crates/computer_use/src/lib.rs
```diff
@@ -238,12 +243,29 @@ impl Default for RecordingConfig {
...
 /// An opaque handle to an in-progress recording, returned by [`Recorder::start`]
 /// and consumed by [`Recorder::stop`]. It owns the live capture process and the
 /// metadata needed to report the applied capture settings.
 pub struct RecordingHandle {
     width: u32,
     height: u32,
+    /// Set once the capture process is observed to have exited on its own (cap or
+    /// crash). Shared so a watcher can poll for early exit without owning the
+    /// process; the real recorder also updates it from `try_wait` in `poll_exit`.
+    exit_state: RecordingExitState,
```
vkodithala, line 268: Small structural point: the current approach seems to be polling for a process exit every 1s, and then processing a valid exit, but I wonder if we can do something similar to what we do for LRCs, where we await on a channel, and then once a result is received on that channel, process it? Did we consider this approach instead of an explicit poll? Poll vs. wakeup basically.

PR #13449, opened 2026-07-07
Speedup TUI startup time

kevinyang372:
## Description

`warp-tui` took ~3s before first paint. The TUI reuses the full headless app bootstrap and only mounts after `initialize_app` completes, and nearly all of that time was spent inside `persistence::initialize` loading state the TUI never uses.

### 1. What's causing the slow startup

On a 580 MB `warp.sqlite` (dev-channel dogfood DB), the stall breaks down as:

- **`agent_tasks` is the dominant cost (304 MB, 1,730 rows).** Each row is a protobuf-encoded full task transcript (every message, tool call, and tool result). Startup loaded and prost-decoded *all* of them (~1.3s of raw blob I/O, roughly doubled by decode/allocation), then converted them **twice more**: once into `RestoredAgentConversations` (GUI pane restoration — unused by the TUI) and once in `initialize_historical_conversations`, which walked every task's messages just to derive list metadata (initial query, title, restorability, passive-diff filter).
- **GUI-only session-restore payloads**: windows/tabs/panes, restored blocks, 10k command-history rows, user profiles, pending object actions — all loaded for a launch mode that never reads them.
- All of this runs synchronously on the main thread before the TUI mounts.

For contrast: reading just the `agent_conversations` metadata rows (≤200, capped) takes ~7ms.

### 2. Options considered

- **Denormalized write-time summary (chosen):** compute list metadata from the task snapshot at persist time in the SQLite writer (single choke point covers live updates, forks, sync), store it in a new nullable `summary` column on `agent_conversations`, and make startup metadata-only with lazy per-conversation hydration (`read_agent_conversation_by_id`, which already existed). Pre-existing rows are derived once at read time and backfilled.
  - *Why a JSON column instead of structured columns:* the summary is only ever read whole for ≤200 rows (never filtered/sorted in SQL), evolves via `#[serde(default)]` without migrations, and follows the existing `conversation_data` precedent. A dedicated column (vs. embedding in `conversation_data`) keeps the backfill a trivially safe `UPDATE ... WHERE summary IS NULL` with no clobber risk against concurrent writes.
- **Store the summary inside `conversation_data`:** no migration, but backfill becomes read-modify-write of a user-authored blob (clobber risk) and touches ~50 construction sites. Rejected.
- **Background decode, no schema change:** hydrate the list off-thread after paint. Fixes perceived latency but every startup still pays the full 304 MB decode; list pops in late. Rejected as the primary fix.
- **Mount the TUI before `initialize_app` (paint-first):** still viable as a follow-up, but measurements show it's no longer necessary for the TUI's startup goal.
- **Normalize messages into per-row tables / move transcripts out of the DB:** the "proper" fix for both read and write costs, but a much larger migration touching the whole restore path. Deferred (see below) — this change is a compatible stepping stone.

Additionally, a `PersistedDataScope` (Full / TuiFrontend / CodebaseIndicesOnly) now gates what each launch mode reads: the TUI skips session restoration, command history, user profiles, and object actions; the remote server daemon reads only codebase indices (replacing its previous load-then-discard block).

### 3. Long-term vision

This is a low-risk stepping stone toward restructuring conversation persistence:

- **Structured `conversation_metadata` table** — one row per conversation (title, initial query, working dir, flags, credits) that supersedes both the `summary` blob and `ai_queries`' display role, unlocking SQL-side search/sort/pagination if retention grows past the current 200-row cap.
- **Incremental task persistence** — today every persist rewrites the *entire* task snapshot (replace semantics), so a growing transcript is O(n²) total bytes written over its life; this is why `agent_tasks` is 304 MB inside a 580 MB file with a 60 MB WAL. Dirty-task tracking and eventually append-only message rows fix the write amplification at the source and enable partial transcript loading.
- **Remaining TUI-scope trims** — `ai_queries` (36 MB, feeds cross-session up-arrow history the TUI doesn't surface yet) and notebooks (37 MB, Drive UI + local document-tool cache) are the next candidates once the document tools' server-fetch fallback is verified.

### Migration / rollout notes

- Additive O(1) `ALTER TABLE`; old builds ignore the new column (rollback-safe, `down.sql` provided).
- First launch on the new build derives + backfills summaries for existing rows (one startup at roughly the old cost); idempotent if interrupted; steady state from the second launch.
- Known soft spot: an old-build writer (GUI/TUI share the DB across a version skew window) updates `conversation_data`/tasks without refreshing `summary`, leaving stale list metadata until any new-build write refreshes it. Cosmetic only (transcripts always load fresh) and self-healing.

## Testing

- New unit tests: summary derivation (`crates/persistence/src/model_tests.rs`), in-memory-DB round trips for metadata read + legacy backfill + stale-backfill no-clobber (`app/src/persistence/agent_tests.rs`), summary-driven history list and skip rules (`history_model_tests.rs`).
- Full suites for touched areas pass (`persistence`, history model, pane group, sqlite tests); `./script/format` + `cargo clippy` clean.
- Manual validation against a copy of a real 580 MB dev DB: one-time backfill of all 200 conversations, second startup metadata-only, orchestration children hydrate lazily, and the conversation-list log output matches the pre-change build exactly.
- Measured with `./script/run-tui`: first launch logs `Backfilling N conversation summaries`; steady-state SQLite window drops from ~3s (dev DB, release build, old code) to sub-second.

## Agent Mode

CHANGELOG-IMPROVEMENT: Faster app startup: persisted agent conversations are now loaded lazily instead of being fully decoded at launch.

### DIFF
--- a/app/src/ai/restored_conversations.rs
+++ b/app/src/ai/restored_conversations.rs
@@ -1,26 +1,62 @@
-//! A singleton model for storing conversations by ID to enable restoration across terminal views.
+//! A singleton model for restoring conversations by ID across terminal views.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+#[cfg(feature = "local_fs")]
+use std::sync::{Arc, Mutex};
 
+#[cfg(feature = "local_fs")]
+use diesel::SqliteConnection;
 use warpui::{Entity, SingletonEntity};
 
 use crate::ai::agent::conversation::{AIConversation, AIConversationId};
+#[cfg_attr(not(any(feature = "local_fs", test)), allow(unused_imports))]
 use crate::ai::blocklist::history_model::convert_persisted_conversation_to_ai_conversation_with_metadata;
+#[cfg(test)]
 use crate::persistence::model::AgentConversation;
+#[cfg(feature = "local_fs")]
+use crate::persistence::{database_file_path_for_scope, establish_ro_connection, PersistenceScope};
 
-/// Singleton model that holds restored agent conversations on app startup.
+/// Singleton model that restores agent conversations on demand.
 ///
-/// Loading restored conversations into this model is a means of propagating restored data from
-/// sqlite (read at startup) to arbitrary consuming locations in the view/model hierarchy without
-/// piping it all the way from the root view to the terminal view(s) that require it.
-#[derive(Default)]
+/// Startup only loads conversation metadata, so the full task payloads are
+/// loaded lazily from the local database the first time a consumer (e.g. pane
+/// restoration) asks for a conversation. Consuming restored data this way
+/// avoids piping it from the root view down to the terminal view(s) that
+/// require it.
 pub struct RestoredAgentConversations {
-    /// All conversations stored by their ID, available for restoration
+    /// Conversations already loaded (or test-seeded) but not yet taken.
     conversations: HashMap<AIConversationId, AIConversation>,
+    /// IDs already handed out via `take_conversation(s)`. Preserves the
+    /// historical take-once semantics now that the backing database can
+    /// otherwise serve the same conversation repeatedly.
+    taken: HashSet<AIConversationId>,
+    #[cfg(feature = "local_fs")]
+    db_connection: Option<Arc<Mutex<SqliteConnection>>>,
 }
 
 impl RestoredAgentConversations {
-    pub fn new(conversations: Vec<AgentConversation>) -> Self {
+    pub fn new() -> Self {
+        #[cfg(feature = "local_fs")]
+        let db_connection = database_file_path_for_scope(&PersistenceScope::App)
+            .to_str()
+            .and_then(|db_url| {
+                establish_ro_connection(db_url)
+                    .ok()
+                    .map(|conn| Arc::new(Mutex::new(conn)))
+            });
+
+        Self {
+            conversations: HashMap::new(),
+            taken: HashSet::new(),
+            #[cfg(feature = "local_fs")]
+            db_connection,
+        }
+    }
+
+    /// Seeds the store with already-loaded conversations instead of a backing
+    /// database. Only used by tests.
+    #[cfg(test)]
+    pub fn new_seeded(conversations: Vec<AgentConversation>) -> Self {
         let mut conversations_by_id = HashMap::new();
         for conversation in conversations.into_iter() {
             let conversation_id = conversation.conversation.conversation_id.clone();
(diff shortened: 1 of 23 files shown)


kevinyang372:
* **#13449** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13449?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13449?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/restored_conversations.rs
```diff
@@ -1,26 +1,62 @@
...
+    taken: HashSet<AIConversationId>,
+    #[cfg(feature = "local_fs")]
+    db_connection: Option<Arc<Mutex<SqliteConnection>>>,
 }
 
 impl RestoredAgentConversations {
-    pub fn new(conversations: Vec<AgentConversation>) -> Self {
+    pub fn new() -> Self {
+        #[cfg(feature = "local_fs")]
+        let db_connection = database_file_path_for_scope(&PersistenceScope::App)
```
harryalbert, line 40: is there a reason we store this db connection on a field instead of acquiring it on demand when we need it? Not sure what the cost is of maintaining a long-lived db connection

app/src/persistence/sqlite.rs
```diff
@@ -2413,198 +2438,240 @@ fn box_persisted_generic_string_object(
...
-                    agent_management_filters: window
-                        .agent_management_filters
-                        .and_then(|s| serde_json::from_str(&s).ok()),
-                    tab_groups: tab_groups_snapshots,
-                }
-            },
-        )
-        .collect();
+        // Load active MCP servers from database
+        let running_mcp_servers = load_active_mcp_servers(conn)?;
```
harryalbert, line 2664: should the TUI be skipping active mcp servers? I'm not sure I get why this would be GUI specific

app/src/persistence/sqlite.rs
```diff
@@ -2717,25 +2784,39 @@ fn read_sqlite_data(
...
-    let commands = schema::commands::dsl::commands
-        // Ensure the commands come into memory sorted chronologically.
-        .order(schema::commands::columns::id.desc())
-        .load_iter::<model::Command, DefaultLoadingMode>(conn)?
-        .filter_map(|command| command.ok())
-        .map(PersistedCommand::from)
-        .collect();
+    // Command history, user profiles, and pending object actions are only
+    // consumed by the GUI; headless launch modes skip loading them.
+    let commands = if data_scope.gui_history() {
```
harryalbert, line 2789: is the idea that we'll re-add these when the TUI does have up-arrow command history? Or will the TUI have a separate command history from the GUI (or no command history at all)

crates/persistence/src/model.rs
```diff
@@ -950,48 +956,182 @@ pub struct AgentConversation {
...
+            .unwrap_or_default();
+
+        // The title is the root task description, falling back to
+        // `initial_query` when the description is empty.
+        let title = root_task
+            .map(|task| task.description.clone())
+            .filter(|desc| !desc.is_empty())
+            .unwrap_or_else(|| initial_query.clone());
+
+        // TODO: search tasks in correct order once we've implemented task ordering.
```
harryalbert: I think this todo is very stale

harryalbert (approved):
nice! just had a few questions

app/src/persistence/sqlite.rs
kevinyang372: This is actually legacy mcp servers and only kept for a one-time migration. I don't think we need to use this in TUI

app/src/ai/restored_conversations.rs
kevinyang372: Holding a db connection is pretty minimal cost. We also have a prior art of this in BlocklistAIHistoryModel

app/src/persistence/sqlite.rs
kevinyang372: Yeah we can re-add the read trivially when we do need it

PR #13111, opened 2026-06-26
perf(ai): O(1) exchange_with_id via exchange_id_index

acarl005:
## Description

Part of a stack of performance fixes for [#9799](https://github.com/warpdotdev/warp/issues/9799): opening a large Oz agent conversation (3000+ exchanges) caused a beachball / multi-second load due to several O(n²) hotspots in the restore path.

**This PR:** `AIConversation::exchange_with_id` previously performed a linear scan through every exchange on every call. During conversation restoration this function is called once per exchange to look up command-block metadata, making the restore path O(n²) overall.

Adds `exchange_id_index: HashMap<AIAgentExchangeId, usize>` to `TaskStore`. The index is rebuilt on all structural mutations (append, insert, remove, exchange removal). `exchange_with_id` and the new `exchange_by_id` now return in O(1).

Also fixes `remove()`: previously it retained entries in `linearized_refs` but left stale entries in `exchange_id_index` for the removed task; now it does a full index rebuild.

## Linked Issue

Fixes https://github.com/warpdotdev/warp/issues/9799

## Testing

Manually loaded a conversation with 3014 AI exchanges and confirmed the restore time was significantly reduced.

- [x] I have manually tested my changes locally with `./script/run`

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

PR #13463, opened 2026-07-08
refactor away extra hashmap in TaskStore

acarl005:
## Description

This PR is a follow-up to #13111.

We already had 2 fields on the `TaskStore` that needed to be kept in sync, `tasks` and `linearized_refs`. In #13111 I added yet-another field that needed to be kept in sync, `exchange_id_index`. There were also a couple of obvious optimizations for re-building the new `exchange_id_index` which I hadn't done.

This was not a good way to do this. Instead, I propose combining `linearized_refs` and `exchange_id_index` into a single field `exchanges: IndexMap<AIAgentExchangeId, ExchangeRef>`. This reduces the number of different data structures that need to be maintained into a single one that supports both ordering as in a `Vec` and constant-time lookup via the key as in a `HashMap`.

I've also implemented `TaskStore::remove` and `append_exchange` in a way that avoids rebuilding the index.

## Testing

Added a few unit tests for a method that wasn't tested.

### DIFF
--- a/app/src/ai/agent/task_store.rs
+++ b/app/src/ai/agent/task_store.rs
@@ -93,7 +93,7 @@ impl TaskStore {
         let is_at_dfs_tail = self
             .exchanges
             .last()
-            .map_or(true, |(_, r)| r.task_id == *task_id);
+            .is_none_or(|(_, r)| r.task_id == *task_id);
         // If we know this exchange is going at the end, it's faster to just append it.
         if is_at_dfs_tail && !has_subagent_output {
             self.exchanges.insert(
@@ -303,8 +303,7 @@ impl TaskStore {
 
     pub fn remove(&mut self, task_id: &TaskId) -> Option<Task> {
         let task = self.tasks.remove(task_id)?;
-        self.exchanges
-            .retain(|_, r| self.tasks.contains_key(&r.task_id));
+        self.rebuild_exchange_index();
         Some(task)
     }
 
--- a/app/src/ai/agent/task_store.rs
+++ b/app/src/ai/agent/task_store.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet};
 
+use indexmap::IndexMap;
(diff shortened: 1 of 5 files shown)


app/src/ai/agent/task_store_tests.rs
```diff
@@ -699,6 +697,129 @@ fn test_exchange_by_id_after_remove_task_exchange_index_shift() {
     );
 }
 
+#[test]
+fn test_prune_unreachable_subtasks_noop_when_no_subtasks() {
```
acarl005, line 701: I didn't add this `fn prune_unreachable_subtasks` but I'm adding tests for it since it was untested.

app/src/ai/agent/task_store.rs
```diff
@@ -296,12 +298,13 @@ impl TaskStore {
...
         self.tasks.insert(task.id().clone(), task);
-        self.rebuild_linearized_refs_index();
+        self.rebuild_exchange_index();
     }
 
     pub fn remove(&mut self, task_id: &TaskId) -> Option<Task> {
         let task = self.tasks.remove(task_id)?;
-        self.rebuild_linearized_refs_index();
+        self.exchanges
+            .retain(|_, r| self.tasks.contains_key(&r.task_id));
```
acarl005: perf optimization as you suggested.

app/src/ai/agent/task_store.rs
```diff
@@ -76,14 +74,39 @@ impl TaskStore {
...
+                exchange_id,
+                ExchangeRef {
+                    task_id: task_id.clone(),
+                    exchange_index,
+                },
+            );
+        } else {
+            // Otherwise, we need to rebuild the whole index.
+            self.rebuild_exchange_index();
+        }
```
acarl005, line 109: `append_exchange` is the common case so this is optimized too

acarl005:
@oz-agent Fix the lint errors in CI

harryalbert (approved):
nice

PR #13483, opened 2026-07-08
Migrate most `log::error!` calls to`report_error!`

acarl005:
### Motivation

The goal is to reduce the *quantity* and *variety* of error events we accidentally send to sentry. There are 2 problems:

1. The quantity is costing us money, [see here](https://warpdev.slack.com/archives/C06L77SHTRD/p1781542746417089).
2. The variety is making monitoring real errors difficult due to floods of duplicated issues. I've gotten down on some folks during my oncall shifts in the past for this. I don't want to keep doing it every shift.

The 2 problems are caused by the same issue: the way the team is using the `log::error!` macro. Currently, that is set up so that every error-level log sends an event to Sentry. I don't think everyone on the team realizes/remembers this, certainly not the agents people are spinning up nor our review agent. People invoke `log::error!` when they just want an emphatic error message in the log file unaware that it _also_ sends an event to Sentry. That's an understandable mistake. One would not assume `log::error!` behaves so differently from `log::warn!` or `log::info!`. It's unintuitive, and that's why I'm not just fixing this by adding skills. It needs to be easier to get right by default.

### Changes

I separated by commit to make it easier to see salient changes.

1. [ed2419d](https://github.com/warpdotdev/warp/pull/13483/changes/ed2419dace77d3374f6058d8c1480d06d74fa84f) - Changes to macros
    1. Move `report_error!` into its own crate, a leaf crate called `warp_errors`. It was moved from `warp_core` so that all call sites of `log::error!` are able to be migrated without dependency cycles, e.g. `warpui_core` could not depend on `warp_core`. Arguably `warpui_core` shouldn't actually be calling `report_error!`, it should be returning errors so that the caller can report it. However, I'll leave those changes for a follow-up to limit scope. This is already a big change.
    2. Add a couple of overloads to `report_error!`.
        1. `report_error!("literal string")` for very simple constant-message errors.
        2. `report_error!("some failure", extra: { key: ?value })` for when highly variable data needs to be attached as context (and logged to the log file).
        3. Analogous changes to `report_if_error!`
    3. `log::error!` no longer sends Sentry error events.
2. [4db69fc](https://github.com/warpdotdev/warp/pull/13483/changes/4db69fc78ff19a4e8d40e1514eb3851ce53a9597) - migrate the `log::error!` call sites to `report_error!`. Note that the interface for `report_error!` isn't exactly the same as `log::error!`. This is by design. For example, `report_error!("literal string")` is added but NOT a full format-string, i.e. not `report_error!("some context: {:#}", err")`. This is to discourage the types of practices that caused the original problem. Instead, prefer `report_error!(e.context("some context"))`. The other overload is that `report_error!` expects an error that `impl ErrorExt`. This includes anyhow but also encourages people to `impl ErrorExt` on custom error enums. This change is huge and I wouldn't advise reading line-for-line. Rather, I would skim it and validate the general pattern. I used agents to carry out these edits and have scanned these but there may be a couple of weird ones. I think that's ok because our error reporting is a mess to begin with. I'm confident this is an improvement.
3. [569a0f6](https://github.com/warpdotdev/warp/pull/13483/changes/569a0f6990b860adfeb173dd264170ad88185b9c) Add a skill for how to do logging/error reporting correctly. I'll have the review bot read this too.

### Follow-up

Once this lands I will have our review bot consult this skill when reviewing PRs with any error logging. I will also be posting in #engineering to publicize the change once this is in. 

I mostly just ported all `log::error!` calls to `report_error!` in order to preserve the current issues in Sentry. Note that this is not the end. I will next go through and audit each one and demote to `log::error!`, `log::warn!`, `safe_error!` etc where it makes sense.

## Linked Issue

None

## Testing

This call:

```rust
    report_error!(
        "Dummy startup error to demo report_error! extra formatting",
        extra: {
            "string_field" => %"hello",
            "int_field" => %42,
            "debug_field" => ?("tuple", 7, true),
        }
    );
```

Gets printed to the log file as such:

```
13:49:03.090 [ERROR] [errors::report_error] Dummy startup error to demo report_error! extra formatting [string_field=hello, int_field=42, debug_field=("tuple", 7, true)]
```

This call:

```rust
        report_error!(
            anyhow::anyhow!("intentional test error")
                .context("Dummy startup error to verify Sentry reporting"),
            extra: {
                "purpose" => "sentry-verification",
                "channel" => ?ChannelState::channel(),
                "app_version" => ?ChannelState::app_version()
            }
        );
```

Shows up in Sentry [as such](https://warpdotdev.sentry.io/issues/7600902064/?project=5701177)

Screenshot 2026-07-08 at 6 01 19 PM
Screenshot 2026-07-08 at 6 01 45 PM


app/src/auth/auth_view_modal.rs
```diff
@@ -289,7 +290,7 @@ impl AuthView {
             }
             AuthManagerEvent::AuthFailed(err) => {
                 if err.is_actionable() {
-                    log::error!("Failed to log in user: {err:#}");
+                    report_error!(err);
```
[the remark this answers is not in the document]
acarl005: Wrong.

app/src/root_view.rs
```diff
@@ -3013,8 +3019,8 @@ impl RootView {
                         }
                     }
                 }
-                UserAuthenticationError::Unexpected(err) => {
-                    log::error!("Encountered unexpected error when trying to fetch user: {err:#}");
+                UserAuthenticationError::Unexpected(_) => {
+                    report_error!(err);
```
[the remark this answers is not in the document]
acarl005: This is an intentional choice. We want to report the outer error that wraps it.

crates/warp_logging/src/native.rs
```diff
@@ -497,7 +497,12 @@ fn sentry_log_filter(md: &log::Metadata) -> sentry_log::LogFilter {
         // anything in the process of forwarding logs to Sentry.
         t if t.starts_with("warp::crash_reporting::") => sentry_log::LogFilter::Ignore,
 
-        _ => sentry_log::default_filter(md),
+        _ => match md.level() {
+            log::Level::Error | log::Level::Warn | log::Level::Info => {
+                sentry_log::LogFilter::Breadcrumb
+            }
+            log::Level::Debug | log::Level::Trace => sentry_log::LogFilter::Ignore,
+        },
```
acarl005, line 506: This prevents the error-level logs from becoming events.

crates/warp_core/src/errors.rs
vorporeal, line 1: i'd appreciate a follow-up at some point that removes the re-export; we're starting to collect re-exports (a lot of this is my doing) and ideally, given the relatively low cost of asking an agent to do it, we pay down that technical debt.

crates/warp_errors/src/errors_tests.rs
```diff
@@ -15,6 +15,9 @@ struct TestLogger;
 
 static LOGGER: TestLogger = TestLogger;
 static LOGS: OnceLock<Mutex<Vec<LogEntry>>> = OnceLock::new();
+// Serializes tests that share the global logger + LOGS buffer (relevant under `cargo test`, where
+// tests share a process; nextest isolates per-process).
+static LOG_TEST_LOCK: Mutex<()> = Mutex::new(());
```
vorporeal: i feel like this sets a nasty precedent of trying to support both raw `cargo test` and `cargo nextest` - we historically haven't bothered with this, and i think i'd rather us simply force agents to use nextest?

crates/warp_errors/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+mod websocket;
+
+// Re-export for macro use. The `register_error!` macro itself is available at the crate root via
+// `#[macro_export]`; here we only re-export the supporting types it references. Re-export anyhow
+// so the `report_error!` macro's string-literal form can build an `anyhow::Error` without callers
+// needing `anyhow` in scope.
+#[doc(hidden)]
+pub use ::anyhow as __anyhow;
+#[doc(hidden)]
+pub use inventory::submit;
```
vorporeal, line 31: what model were you using that added the `#[doc(hidden)]` annotations?  i haven't yet seen them used elsewhere in our codebase.  (not saying we shouldn't have them, just curious)

crates/warp_errors/src/reqwest.rs
```diff
@@ -1,15 +1,14 @@
...
-        // Outside of timeouts, there's nothing we can do about errors
-        // that occur prior to the successful receipt of an HTTP
-        // response.
+        // Outside of timeouts, there's nothing we can do about errors that occur prior to the
+        // successful receipt of an HTTP response.
 
-        // There's no way to check for connection errors via web APIs, so
-        // `is_connect` can only be called on native platforms.
+        // There's no way to check for connection errors via web APIs, so `is_connect` can only be
+        // called on native platforms.
```
vorporeal, line 11: curious what tool you were using that reflowed comments

vorporeal (approved):
tbqh there's too much here for me to really review; i did a quick pass over the first commit and going to just approve

crates/warp_core/src/errors.rs
acarl005: you have my word

crates/warp_errors/src/lib.rs
acarl005: just default auto(genius)

crates/warp_errors/src/reqwest.rs
acarl005: Ah it was one of my NeoVim plugins

acarl005:
> tbqh there's too much here for me to really review; i did a quick pass over the first commit and going to just approve

I swear to keep an eye on Sentry for the next few weeks and make sure this goes 💯

crates/warp_errors/src/errors_tests.rs
acarl005: I did hate that hack. removed it

PR #13486, opened 2026-07-08
Add TUI slash command query state

moirahuang:
## Description

Adds the non-visual TUI slash command query-state layer on top of the data-source split from the parent PR.

### Query state

- Adds `TuiSlashCommandModel`, which observes the TUI input editor and opens when slash command composition begins.
- Runs `slash_command_query` through the shared `SlashCommandMixer` and snapshots results into TUI-owned rows containing a title, optional description, and acceptance action.
- Refreshes the query when active commands change and refreshes rows when mixer results change.
- Resets mixer results and closes the state when the input no longer represents slash command composition.
- Extracts `slash_command_composition_filter` from the GUI model so GUI and TUI use the same composition rules.

### TUI wiring

- Constructs `TuiSlashCommandDataSource` with `TuiSlashCommandDataSourceArgs`; it does not use the GUI source or optional GUI handles.
- Constructs `TuiZeroStateDataSource` explicitly and passes it to the generic mixer builder.
- Registers the new `slash_commands` module and wires the model into `TuiTerminalSessionView` for later rendering and dispatch layers.

This PR intentionally does not render the menu or execute selections. Those behaviors are added by the next two PRs in the stack.

## Linked Issue

N/A

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Testing

- `cargo check -p warp_tui`
- `./script/format`
- `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`
- `cargo clippy -p warp_completer --all-targets --tests -- -D warnings`
- [ ] I have manually tested my changes locally with `./script/run`

## Agent Mode

- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-NONE

Co-Authored-By: Oz [oz-agent@warp.dev](mailto:oz-agent@warp.dev)

PR #13485, opened 2026-07-08
Extract shared slash command mixer

moirahuang:
## Description

Refactors slash command data sources so GUI and TUI share only surface-neutral state and behavior. Instead of making the TUI construct a GUI-oriented data source or adding TUI flags and optional GUI handles to one source, each surface wraps the same concrete shared component and owns its own dependencies.

There is no intended change to GUI slash command menu behavior.

### Data source split

- **`SlashCommandDataSource` (shared core)** owns the active session, CLI controller, terminal ID, active command set, and repository root. It implements shared availability bits and gates, command and skill matching, and zero-state ordering.
- **`GuiSlashCommandDataSource`** wraps the core and owns `AgentViewController`, `AmbientAgentViewModel`, and cloud-mode-v2 state. GUI-only availability, cloud/harness gates, subscriptions, and compact rendering stay here.
- **`TuiSlashCommandDataSource`** wraps the core without constructing or importing GUI-only dependencies. It supplies TUI availability and query behavior.
- **Zero state is surface-specific**: `GuiZeroStateDataSource` preserves GUI/cloud-mode skills and saved prompts, while `TuiZeroStateDataSource` uses the shared command ordering needed by TUI.
- **Mixer construction is generic**: `build_slash_command_mixer` accepts the surface's primary and zero-state sources, allowing both surfaces to reuse composition without sharing one concrete source type.

### Integration changes

- Renames GUI call sites to use the explicit GUI source and zero-state types.
- Exports only the TUI-facing source, zero-state source, mixer, and result types through `tui_export`.
- Derives `Default` for `EphemeralMessageModel` as a small clippy cleanup required by the refactored stack.

## Linked Issue

N/A

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [x] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Testing

- `cargo check -p warp`
- `cargo check -p warp_tui`
- `cargo test -p warp terminal::input::slash_commands::data_source::core::tests --lib` — 6 passed
- `./script/format`
- `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`
- `cargo clippy -p warp_completer --all-targets --tests -- -D warnings`
- [ ] I have manually tested my changes locally with `./script/run`

## Agent Mode

- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-NONE

[...]

PR #13487, opened 2026-07-08
Render TUI slash command menu

moirahuang:
## Description

Adds the visual and keyboard-interaction layer for the TUI slash command menu on top of the query state from the parent PR.

### Menu rendering

- Renders slash command results above the TUI input with selected, unselected, loading, and no-results states.
- Shows up to eight result rows and keeps the selected row inside the visible window while navigating.
- Preserves the selected action across result refreshes when possible and selects the highest-priority zero-state result initially.
- Constrains the rendered menu to ten terminal rows in the session layout.

### Keyboard interaction

- Routes up/down navigation to the slash command model while the menu is open.
- Routes Enter through an `AcceptedSlashCommand` input event without executing the action in this PR.
- Adds Escape dismissal for the open menu.
- Keeps normal input behavior unchanged when the slash command menu is closed.

This PR intentionally stops at rendering and selection. Command parsing and execution are added by the final PR in the stack.

## Linked Issue

N/A

## Testing

- `cargo check -p warp_tui`
- `./script/format`

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/crates/warp_search_core/src/inline_menu.rs
+++ b/crates/warp_search_core/src/inline_menu.rs
@@ -1,0 +1,109 @@
+//! Headless selection state shared by inline-menu frontends.
+
+/// Which selectable result should be chosen when resetting a menu.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InitialSelection {
+    First,
+    Last,
+}
+
+/// Logical selection over an ordered list of inline-menu results.
+///
+/// This type intentionally knows nothing about rendering direction, scrolling,
+/// result payloads, or menu lifecycle. Frontends supply the current item count
+/// and a predicate that excludes disabled rows and separators.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct InlineMenuSelection {
+    selected_index: Option<usize>,
+}
+
+impl InlineMenuSelection {
+    pub fn selected_index(&self) -> Option<usize> {
+        self.selected_index
+    }
+
+    pub fn clear(&mut self) {
+        self.selected_index = None;
+    }
+
+    pub fn reset(
+        &mut self,
+        item_count: usize,
+        initial_selection: InitialSelection,
+        mut is_enabled: impl FnMut(usize) -> bool,
+    ) -> Option<usize> {
+        self.selected_index = match initial_selection {
+            InitialSelection::First => (0..item_count).find(|&index| is_enabled(index)),
+            InitialSelection::Last => (0..item_count).rev().find(|&index| is_enabled(index)),
+        };
+        self.selected_index
+    }
+
+    pub fn select(
+        &mut self,
+        index: usize,
+        item_count: usize,
+        mut is_enabled: impl FnMut(usize) -> bool,
+    ) -> Option<usize> {
+        if index >= item_count || !is_enabled(index) {
+            return None;
+        }
+        self.selected_index = Some(index);
+        self.selected_index
+    }
+
+    pub fn select_next(
+        &mut self,
+        item_count: usize,
+        is_enabled: impl FnMut(usize) -> bool,
+    ) -> Option<usize> {
+        let start = match self.selected_index {
+            Some(index) if index < item_count.saturating_sub(1) => index + 1,
+            Some(_) | None => 0,
+        };
+        self.select_from(start, item_count, ScanDirection::Forward, is_enabled)
+    }
+
+    pub fn select_previous(
+        &mut self,
+        item_count: usize,
+        is_enabled: impl FnMut(usize) -> bool,
+    ) -> Option<usize> {
+        let start = match self.selected_index {
+            Some(index) if index > 0 && index < item_count => index - 1,
+            Some(_) | None => item_count.saturating_sub(1),
+        };
+        self.select_from(start, item_count, ScanDirection::Backward, is_enabled)
+    }
+
+    fn select_from(
+        &mut self,
+        start: usize,
+        item_count: usize,
+        direction: ScanDirection,
+        mut is_enabled: impl FnMut(usize) -> bool,
+    ) -> Option<usize> {
+        if item_count == 0 {
+            self.clear();
+            return None;
+        }
+
+        self.selected_index = (0..item_count)
+            .map(|offset| match direction {
+                ScanDirection::Forward => (start + offset) % item_count,
+                ScanDirection::Backward => (start + item_count - offset) % item_count,
+            })
+            .find(|&index| is_enabled(index));
+        self.selected_index
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+enum ScanDirection {
+    Forward,
+    Backward,
+}
+
+#[cfg(test)]
+#[path = "inline_menu_tests.rs"]
+mod tests;
--- a/crates/warp_tui/src/inline_menu.rs
+++ b/crates/warp_tui/src/inline_menu.rs
@@ -1,0 +1,320 @@
+//! Reusable active-menu routing and character-cell presentation for TUI inline menus.
+
+use warp::tui_export::AcceptSlashCommandOrSavedPrompt;
+use warpui_core::elements::tui::{
+    TuiBuffer, TuiConstraint, TuiContainer, TuiElement, TuiFlex, TuiLayoutContext, TuiPaintContext,
+    TuiRect, TuiSize, TuiText,
+};
+use warpui_core::elements::CrossAxisAlignment;
+use warpui_core::{AppContext, ModelAsRef, ModelHandle, UpdateModel};
+
+use crate::slash_commands::TuiSlashCommandModel;
+use crate::tui_builder::TuiUiBuilder;
+
+/// A presentation-only row in a TUI inline menu.
+#[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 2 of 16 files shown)


moirahuang:
* **#13488** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13488?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13487** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13487?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13487?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13486** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13486?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13485** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13485?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -236,6 +243,9 @@ impl TuiTerminalSessionView {
                     ctx.notify();
                 }
             }
+            TuiInputViewEvent::AcceptedSlashCommand(action) => {
+                log::debug!("Accepted TUI slash command menu item: {action:?}");
```
moirahuang, line 283: doesn't actually submit for now, that's in the PR next in the stack

crates/warp_tui/src/input/view.rs
```diff
@@ -94,6 +96,14 @@ pub fn init(app: &mut AppContext) {
         .with_context_predicate(id!("TuiInputView"))
         .with_group(TUI_BINDING_GROUP)
         .with_key_binding("alt-enter"),
+        EditableBinding::new(
```
harryalbert, line 105: does this interact well w/ the esc to clear `!` input mode keybinding? I think it probably makes sense to have some keybinding priority order (and not show the `esc` to clear shell mode if the slash command menu takes priority, which it probably should imo)

crates/warp_tui/src/input/view.rs
```diff
@@ -560,13 +576,22 @@ impl TuiInputView {
     /// Subscribes to [`CodeEditorModelEvent::ContentChanged`] to trigger re-renders
     /// whenever the buffer changes from outside `handle_action`.
     pub fn new(model: ModelHandle<CodeEditorModel>, ctx: &mut ViewContext<Self>) -> Self {
+        Self::new_with_slash_commands(model, None, ctx)
+    }
+
+    pub(crate) fn new_with_slash_commands(
```
harryalbert: what's the point of this separate new fn? Could this just be the new fn?

crates/warp_tui/src/slash_commands.rs
```diff
@@ -177,3 +307,71 @@ fn row_from_result(
         action: result.accept_result(),
     })
 }
+fn keep_selected_visible(rows_len: usize, selected_index: usize, scroll_offset: &mut usize) {
```
harryalbert: could you add a doc comment explaining this fn? I'm not sure I understand what this does

harryalbert (commented):
nice! this looks great

harryalbert (approved):
sorry meant to approve

crates/warp_tui/src/slash_commands.rs
```diff
@@ -91,6 +99,101 @@ impl TuiSlashCommandModel {
...
+        *selected_index = (*selected_index + 1) % rows.len();
+        keep_selected_visible(rows.len(), *selected_index, scroll_offset);
+        ctx.emit(TuiSlashCommandModelEvent);
+    }
+
+    pub(crate) fn dismiss(&mut self, ctx: &mut ModelContext<Self>) {
+        self.close(ctx);
+    }
+
+    pub(crate) fn render_menu(&self, app: &AppContext) -> Option<Box<dyn TuiElement>> {
```
harryalbert: In an ideal world, it would be great to extract out a shared menu element that can be used here and elsewhere. I'm thinking about the conversation list, for ex, and it would be nice if that could just re-use the same like inline menu component

harryalbert: if this is high lift I think it's ok to mark as todo for later, but ideally it would be good to do this

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -504,19 +514,21 @@ impl TuiView for TuiTerminalSessionView {
...
-                .flex_child(TuiChildView::new(&self.transcript).finish())
-                .child(input_box.finish())
-                .child(
-                    TuiConstrainedBox::new(self.render_footer(ctx).finish())
-                        .with_max_rows(1)
-                        .finish(),
-                )
+        let mut content =
+            TuiFlex::column().flex_child(TuiChildView::new(&self.transcript).finish());
+        if let Some(menu) = slash_command_menu {
```
harryalbert: I wonder if we should have a more generalized like "some inline menu is open" concept. Feels like we'll be re-using this in the future and it might be worth making it more extensible.

I think this pairs w/ my comment in the PR below about some generalized menu component. It might be worth referencing how the GUI's inline menu model works and trying to do something similar here

crates/warp_tui/src/input/view.rs
moirahuang: Yep — I consolidated this into one contextual Escape binding with an explicit priority order: dismiss an open inline menu first, otherwise exit `!` shell mode. There’s also a regression test that puts both states together and verifies the first Escape closes the menu while shell mode remains active. In the current production flow the slash menu and shell mode are mutually exclusive, so the shell-mode footer hint doesn’t conflict.

crates/warp_tui/src/input/view.rs
moirahuang: Agreed — I removed the separate constructor. `TuiInputView::new` is now the single crate-internal constructor and accepts an optional generalized `TuiInlineMenu`, so menu and non-menu initialization can’t diverge.

crates/warp_tui/src/slash_commands.rs
moirahuang: Added documentation and moved this into the generalized inline-menu module. It now explicitly takes the number of visible result rows, clamps stale scroll offsets, and moves the viewport only as far as needed to keep the selected row visible. The renderer derives that capacity from its actual allocated height.

crates/warp_tui/src/slash_commands.rs
moirahuang: Implemented this. The TUI now has a reusable, domain-neutral inline-menu presentation layer with rows, optional headers/tabs, loading and empty states, height-aware scrolling, and shared rendering. The slash-command model only supplies a snapshot and retains the domain-specific accepted action, so another menu such as conversation selection can reuse the same component without copying slash-specific UI code.

crates/warp_tui/src/terminal_session_view.rs
moirahuang: Implemented this as `TuiInlineMenu`. `TuiInputView` and `TuiTerminalSessionView` now interact with the generalized adapter for `is_open`, rendering, navigation, acceptance, and dismissal rather than depending directly on slash-command state. Slash commands are currently the first variant, while future menu kinds can retain their own query/action models behind the same boundary.

crates/warp_search_core/src/inline_menu.rs
```diff
@@ -0,0 +1,109 @@
...
+    Last,
+}
+
+/// Logical selection over an ordered list of inline-menu results.
+///
+/// This type intentionally knows nothing about rendering direction, scrolling,
+/// result payloads, or menu lifecycle. Frontends supply the current item count
+/// and a predicate that excludes disabled rows and separators.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct InlineMenuSelection {
```
moirahuang, line 16: This is the headless piece shared by GUI and TUI. Both frontends now use `InlineMenuSelection` for reset and wraparound navigation while supplying their own enabled-item predicate. Rendering, menu lifecycle, scrolling, and result payloads intentionally remain frontend-specific. The GUI migration also clears its selected payload when a refresh contains no enabled item.

crates/warp_tui/src/inline_menu.rs
```diff
@@ -0,0 +1,320 @@
...
+    SlashCommand(AcceptSlashCommandOrSavedPrompt),
+}
+
+/// The active TUI inline menu.
+///
+/// This is the input and placement boundary shared by menu kinds. Each variant
+/// retains its own query state and accepted action, while all variants expose
+/// the same navigation, dismissal, snapshot, and rendering behavior.
+#[derive(Clone)]
+pub(crate) enum TuiInlineMenu {
```
moirahuang, line 66: This is the reusable TUI-specific boundary added in response to the review feedback. Menu kinds own their domain query state and accepted actions, but input routing and character-cell rendering operate on domain-neutral snapshots. A future conversation-list menu should therefore add an adapter/variant here rather than duplicating the slash-command renderer.

moirahuang:
### Merge activity

* **Jul 10, 6:35 PM UTC**: [Graphite](https://app.graphite.com/github/pr/warpdotdev/warp/13487) couldn't merge this PR because it had merge conflicts.

PR #13493, opened 2026-07-08
Extract core TUI editor element and migrate the input onto it

harryalbert:
## Description

The TUI needs to render editor content in more than one surface: the prompt input today, and the inline diff viewer for agent file edits next (CODE-1800, stacked on top as #13444) — which needs everything the input has *plus* structural overlays: a line-number gutter, interleaved removed-line "ghost" rows, and hidden-range elision. Before this PR the input assembled its own rows inside a bespoke `TuiInputElement`, re-implementing wrap math the editor crate already owns; a second surface would have meant a second copy. This PR extracts the shared core — the TUI analogue of the GUI's `RichTextElement` — as two layers: a display-row projection (`DisplayLattice`) computed by the char-cell render state, so what's painted on row N and what a click on row N resolves to are projections of one computation and can never disagree; and a paint/interact-only element (`TuiEditorElement`) that consumers configure with builder methods and that knows nothing about diffs, keybindings, or kill/yank. Editor session state also moves to where the GUI keeps it: viewport scroll lives on the char-cell render state, drag gating derives from the selection model's pending selection, and visual-row kill edits become model APIs — leaving the input view as pure input policy (the readline binding table, the kill buffer, submit, `!` shell mode; net −650 lines). TUI input behavior is unchanged

- `specs/tui-editor-element/TECH.md` — the design rationale: why structural overlays are core-element mechanisms in char-cell space (they change the row lattice and wrap width), and the buffer-visual-row vs display-row coordinate split.
- `crates/editor/src/render/model/char_cell_display.rs` — the display-row projection: `DisplayRow` (`Buffer` | `Ghost` | `Gap`, carrying char *ranges*, never text or styles) and `DisplayLattice`, which owns the row list plus the offset↔point queries answered against those same rows (including the deferred-wrap "phantom row" the cursor occupies when a line exactly fills the width). The actual code is pretty skimable IMO, but worth reading through the APIs that are presented.
- `crates/editor/src/render/model/mod.rs` — `CharCellState` grows the overlay mechanisms: ghost-block storage (`CharCellTemporaryBlock`, flattened from the GUI's `TemporaryBlock`; the previously no-op'd `LayoutTemporaryBlock` char-cell arm now stores them, also fixing a latent outstanding-layouts counter leak), `with_display_lattice` as the single projection entry point, and the scroll offset with its clamp/follow-cursor policy (the char-cell mirror of the GUI's `RenderState`-owned scroll). `RenderState::hidden_line_ranges` projects the `HiddenLinesModel` into the line-range shape the lattice consumes.
- `crates/warp_tui/src/editor_element.rs` — `TuiEditorElement`: snapshots the model per frame, opens one lattice scope at layout to build rows/cursor/selection spans, hit-tests mouse events in a fresh lattice scope at event time (elements are cached across frames), and emits editor-generic `TuiEditorAction`s to the owning view. Gutter, styles, trailing-newline elision, and viewport windowing are builder config; read-only is simply not wiring `.editable()` / `.on_action()`.
- `crates/warp_tui/src/input/view.rs` — the migration: `TuiInputElement` and its duplicated wrap/hit-test/scroll/kill machinery deleted; the view renders the core element and keeps only the `tui:input:*` binding table, the kill buffer, submit/clear, and shell mode.
- `app/src/code/editor/model.rs` — `kill_to_visual_row_end/_start`: buffer mutation is model semantics, so the raw `BufferEditAction` plumbing lives behind the model, which returns the killed text for the view's kill buffer.
- `crates/editor/src/selection.rs` — `has_pending_selection()`: drag gating is derived from the model's pending selection instead of a mirrored `is_selecting` flag on the view.

## Testing

Confirmed TUI input works the same as before

## Agent Mode


harryalbert:
* **#13541** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13541?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13493** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13493?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13493?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/code/editor/model.rs
```diff
@@ -1920,6 +1926,88 @@ impl CodeEditorModel {
         self.backspace(ctx);
     }
 
+    // ── Char-cell (TUI) visual-row kill ──────────────────────────────────────
+
+    /// Deletes from the primary cursor to the end of its soft-wrapped visual
+    /// row, returning the deleted text; `None` when the cursor is already at
+    /// the row end. Char-cell (TUI) mode only — visual rows follow the
+    /// terminal-width wrap math.
+    pub fn kill_to_visual_row_end(&mut self, ctx: &mut ModelContext<Self>) -> Option<String> {
```
kevinyang372: Anyway we could 1) make it clear in the naming that this TUI mode only 2) better enforce this can only be called when the editor is in TUI mode in code

crates/editor/src/render/model/char_cell_display.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+/// A position in display-row space: a 0-based display row and a 0-based
+/// display column in terminal cells. The display-space analogue of
+/// [`SoftWrapPoint`](super::SoftWrapPoint) with
+/// [`ColumnUnit::Chars`](super::ColumnUnit) columns; display space is
+/// char-cell only, so the column is a bare `u16`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DisplayPoint {
+    pub row: u32,
```
kevinyang372, line 82: nit: I don't quite understand why we have different units for rows vs columns

crates/editor/src/render/model/char_cell_display.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// Offsets inside hidden lines resolve to their gap row; when the hidden
+    /// run produced no gap (edge runs emit no rows), leading runs resolve to
+    /// the first display row and trailing runs to the last. A deferred-wrap
+    /// cursor at the end of a line that exactly fills the width lands on the
+    /// next *buffer* row, or one past the entire display when no buffer row
+    /// follows — never on an interleaved ghost or gap row, which holds no
+    /// buffer gap for a cursor. Callers sizing a viewport must accommodate
+    /// that phantom row.
+    pub fn offset_to_display_point(&self, char_idx: usize) -> DisplayPoint {
```
kevinyang372: What is the main use case for offset_to_display_point and display_point_to_offset?

Mostly asking because if this is used for things like cursor display, we should consider returning Option<DisplayPoint> if the char_idx happens to be on a hidden range, for example

crates/editor/src/render/model/mod.rs
```diff
@@ -456,6 +493,126 @@ impl CharCellState {
...
+    /// The projection is computed once per call and every query `f` makes is
+    /// answered against those same rows, so painting and interaction geometry
+    /// derived in one closure can never disagree. `hidden_line_ranges` is a
+    /// parameter (not internal state) so consumers can append structural
+    /// extras (e.g. a diff view eliding the buffer's final empty line) to the
+    /// model-derived set from [`RenderState::hidden_line_ranges`].
+    pub fn with_display_lattice<R>(
+        &self,
+        hidden_line_ranges: &[Range<usize>],
+        f: impl FnOnce(&DisplayLattice) -> R,
```
kevinyang372: The semantics of this method is weird...Typically, `with_` is a setter on the struct itself but we are not really setting new fields. Instead here we are constructing a DisplayLattice and passing it through an arbitrary function -- I would consider just having a API (`display_lattice) that returns a DisplayLattice and the caller is responsible for calling f

crates/editor/src/render/model/mod.rs
```diff
@@ -456,6 +493,126 @@ impl CharCellState {
...
+        f(&lattice)
+    }
+
+    /// The 0-based char range of the soft-wrapped visual row containing the
+    /// gap at 0-based char index `char_idx`, excluding any trailing newline.
+    ///
+    /// Buffer visual-row space (no ghosts/hidden ranges); the row boundaries
+    /// follow the same display-width wrapping as everything else in this
+    /// state, so e.g. kill-to-visual-line-end ranges match the rendered rows.
+    pub fn visual_row_char_range(&self, char_idx: usize) -> Range<usize> {
```
kevinyang372: For character indices, we should try to use CharOffset instead of usize

crates/editor/src/render/model/mod.rs
```diff
@@ -2054,6 +2210,36 @@ impl RenderState {
         }
     }
 
+    /// Char-cell only: the hidden line ranges from the attached
```
kevinyang372: From the naming it's not clear this is only for char cell mode

Also similar to above, we should consider how to make this so it's only valid in char. cell mode

kevinyang372 (approved):

crates/editor/src/render/model/char_cell_display.rs
harryalbert: b/c rows is relative to the entire like content of the document (not just what's visible), while cols is relative to the terminal width, so it's necessarily constrained

crates/editor/src/render/model/char_cell_display.rs
harryalbert: `offset_to_display_point` is mainly used for cursor placement, and then `display_point_to_offset` is used for mouse hit testing. So I think you're right that both should just return options so that if they target a buffer that's hidden they just return None.

crates/editor/src/render/model/mod.rs
harryalbert: yeah good point, I like your api suggestion. I'll update

crates/editor/src/render/model/mod.rs
harryalbert: yeah, I think we could maybe just move these fns directly into CharCellState so that it's clear that they're only for TUI

app/src/code/editor/model.rs
harryalbert: yeah, it's tough because I don't really want to make a model specific to the tui, and for the case where there's no char cell we just return None. I think we can make the naming better and make the return earlier

PR #13266, opened 2026-07-01
fix(macOS): stop registering Warp as the default handler for Markdown (.md) files

AaronReboot:
## Problem

On macOS, Warp registers itself as the default application for Markdown (`.md`) files and reclaims that association after updates, overriding the user's chosen editor (e.g. VS Code). Reported in #9076 (see comments — Warp re-registering for `.md`/`.go`/`.svg` after relaunch; *"Even after selecting Always open with VS Code, Warp remains the default"*) and #5732.

## Root cause

`script/update_plist` injects a `Markdown File` `CFBundleDocumentTypes` entry declaring Warp an **Editor** for the markdown UTIs:

```xml
<string>net.daringfireball.markdown</string>
<string>com.unknown.md</string>
<string>net.ia.markdown</string>
<string>public.markdown</string>
```

`.md` resolves to the UTI `net.daringfireball.markdown`. Most editors (e.g. VS Code) register markdown only by file **extension**, not by this UTI. macOS prefers a UTI-level handler over an extension-only one, so Warp wins the default-app resolution for `.md` **even at `LSHandlerRank Alternate`**, and re-wins on every auto-update's re-registration. Because this is an OS-level Launch Services claim, disabling the in-app Markdown viewer (`prefer_markdown_viewer = false`) does not prevent it. (`com.unknown.md` also appears to be a placeholder/invalid UTI.)

## Fix

Remove the `Markdown File` document type so a terminal no longer registers itself as a Markdown editor — this also drops the invalid `com.unknown.md` UTI. The `Plain Text File` and `Source Code File` types are intentionally left in place.

Closes #13268

Related: #9076, #5732.

PR #13513, opened 2026-07-09
fix(macOS): stop registering Warp as an editor for text/source/all-document types

### LINKED ISSUE #9076: Stop hijacking file links
### Pre-submit Checks

My .go files used to open in Vscode when Command-clicking. Now they open in warp which is useless. Searching for `open` in settings does not reveal any setting.

Please stop enshittifying warp by adding features that keep breaking daily working routine.

### Additional context

_No response_

### Operating system (OS)

Select an OS

1 (Not too important)

None

andig:
## Problem

On macOS, Warp's `Info.plist` registers itself as an `Editor` for a very broad set of document types via `script/update_plist`: Markdown, Plain Text, Source Code (both the `public.source-code` UTI and an explicit ~150-entry extension list including `.go`, `.svg`, `.py`, etc.), and `All Documents` (`public.data`, the base UTI for every file type). This causes macOS Launch Services to keep selecting Warp as the default handler over the user's actual chosen editor (e.g. VS Code), and it re-claims that default after every relaunch/update.

#13121 previously set `LSHandlerRank=Alternate` on these entries, but as the analysis on #9076 and #13266 established, a UTI/broad-registration still outranks an extension-only registration (like VS Code's) regardless of `LSHandlerRank` — so `Alternate` alone does not stop the hijack.

## Fix

Remove the `Markdown File`, `Plain Text File`, `Source Code File`, `Source Code`, and `All Documents` `CFBundleDocumentTypes` entries from `script/update_plist`. Only `Folder`, `Terminal shell script`, and `Unix Executable File` remain — these are the types Warp actually needs to function (opening folders as new tabs, executing shell scripts/binaries); nothing else in the codebase depends on the removed registrations. Warp's own external-editor open logic (`app/src/util/file/external_editor/`) resolves editors via its own settings, independent of this plist registration.

Fixes #9076

## Testing

- `bash -n script/update_plist` passes.
- Parsed the generated `CFBundleDocumentTypes` XML with `plistlib`, confirmed it's well-formed with exactly 3 entries (Folder, Terminal shell script, Unix Executable File).
- This is a macOS-only packaging/plist change with no Rust code path to exercise; a maintainer on macOS should confirm Warp no longer becomes the default handler for `.go`/`.md`/`.svg`/etc. after bundling.


andig:
I honestly don't care for the review comments. Warp is broken, this fixes it. Feel free to leave it broken if usage on MacOS doesn't matter for you.

andig:
I‘ve given up on Warp.

PR #13523, opened 2026-07-09
import warp_errors directly instead of chain of re-exports

acarl005:
Follow up on #13483 for [this comment](https://github.com/warpdotdev/warp/pull/13483#discussion_r3548174877) to remove the re-exporting.


crates/warp_core/src/lib.rs
```diff
@@ -26,15 +25,12 @@ pub mod telemetry;
...
-// an `anyhow::Error` without callers needing `anyhow` in scope.
 #[doc(hidden)]
 pub use anyhow as __anyhow;
 pub use app_id::AppId;
 pub use session_id::SessionId;
-// The error-reporting macros now live in the `warp_errors` leaf crate; re-export
-// them at the crate root so `report_error!` etc. keep working.
-pub use warp_errors::{register_error, report_error, report_if_error};
+#[doc(hidden)]
+pub use warp_errors as __warp_errors;
```
acarl005, line 33: This is apparently still needed so that `macro_rules! safe_assert` works without the caller importing `warp_errors`

vorporeal (approved):

PR #13566, opened 2026-07-10
Add selection to tui transcript view (through new generic selectable trait)

harryalbert:
## Description

The TUI captures mouse input globally, so before this PR transcript content could not be selected at either layer: host-terminal selection never saw the drag, and the transcript itself only implemented scrolling. This PR adds terminal-style selection directly to the transcript — single-click drag selects rendered cells linearly, double-click uses the same smart-select rules as the GUI terminal, triple-click selects full visual rows, edge dragging autoscrolls, and mouse-up copies exactly what is highlighted through OSC 52 (including tmux passthrough, so the clipboard write reaches the local terminal over SSH). The selection remains visible after mouse-up and across scrolling and transcript appends; rich-content height changes rebase absolute row anchors when the selected glyphs survive, while horizontal reflow, removed selected rows, or changed selected glyphs invalidate it.

The implementation is split at the layer that owns each fact. `TuiSelectable<Child>` is only interaction plumbing: existing child interactions get first refusal, active drags capture mouse-move/up, and start/copy callbacks escape to the owning view. `TuiViewportedList` owns selection geometry because it already owns the resolved content-row ↔ screen-row mapping; it handles hit-testing, word/line units, edge autoscroll, reverse-video painting, off-screen extraction, and row-resize rebasing. Selection state lives beside scroll state in `TuiViewportedListState`, so an `Rc`-backed `TuiSelectionHandle` survives the element tree being rebuilt every frame. `TuiBlockListViewportSource` adds a read-only row-materialization path for copying content that has scrolled off-screen, plus ordered rich-content resize deltas, without remeasuring or mutating the transcript during selection. Warp-specific word semantics stay above `warpui_core` as an injected resolver, and `TuiTerminalSessionView` owns the cross-surface invariant and side effects: starting selection in one surface clears the other; completing a transcript selection writes OSC 52 and shows the success-colored `copied to clipboard` hint. Existing transcript interactions and input-editor behavior are unchanged apart from that mutual exclusion.

```text
TuiTerminalSessionView
│
├── TuiInputView
│   └── CodeEditorModel selection
│       └── SelectionChanged ─────────────────────────────┐
│                                                        │ clears the
└── TuiTranscriptView                                    │ other surface
    └── TuiScrollable                                    │
        └── TuiSelectable[child]                          │
            ├── on_selection_start ───────────────────────┘
            ├── on_copy(text) ──→ session ──→ OSC 52 + transient hint
            │
            └── TuiViewportedList[TuiBlockListViewportSource]
                ├── TuiViewportedListState
                │   ├── scroll position + resolved viewport
                │   └── TuiSelectionHandle
                │       └── anchor/focus spans + glyph snapshot
                ├── TuiSelectionConfig
                │   └── word_span ──→ SemanticSelection::smart_search
                └── TuiBlockListViewportSource
                    ├── selection_content()       (read-only off-screen rows)
                    └── row-resize deltas         (selection rebasing)
```

- product/tech spec
- `crates/warpui_core/src/elements/tui/selectable.rs` — the generic wrapper and `TuiSelectableElement` protocol; this deliberately contains no viewport geometry or Warp-specific selection semantics (so API skim is sufficient).
- `crates/warpui_core/src/elements/tui/selectable/state.rs` — persistent anchor/focus state, selected-glyph snapshots, width invalidation, and ordered row-resize rebasing.
- `crates/warpui_core/src/elements/tui/viewported_list.rs` — the core implementation: resolved viewport geometry, hit-testing, selection-unit resolution, edge autoscroll, highlight painting, and chunked off-screen text extraction.
- `crates/warp_tui/src/tui_block_list_viewport_source.rs` — read-only selection content plus rich-content resize reporting, kept separate from the normal measurement/mutation path.
- `crates/warp_tui/src/transcript_view.rs` and `crates/warp_tui/src/transcript_word_selection.rs` — transcript composition, callback wiring, lifecycle invalidation, and GUI-compatible smart word selection.
- `crates/warp_tui/src/terminal_session_view.rs` — input/transcript mutual exclusion, copy dispatch, and success-colored footer feedback.
- `crates/warp_tui/src/clipboard.rs` — UTF-8 OSC 52 writes for clipboard and Linux PRIMARY, including tmux DCS passthrough.

## Testing

### Screenshots / Videos

https://www.loom.com/share/bc8d1243c138483ab2a3c9fa721e943a

## Agent Mode


harryalbert:
* **#13566** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13566?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -92,6 +94,10 @@ pub(crate) enum TuiTerminalSessionAction {
     /// Click on the footer's usage entry: flips the persisted credits⇄cost
     /// display-mode setting.
     ToggleUsageDisplay,
+    /// A transcript drag began; clear the input editor's selection.
```
kevinyang372: Should this action belong to Terminal session view? I feel like it really belong to transcript view and should be bubbled up to session view

crates/warp_tui/src/transcript_view.rs
```diff
@@ -81,6 +84,19 @@ impl TuiTranscriptView {
         }
     }
 
+    fn block_rows(&self, view_id: EntityId) -> Option<Range<usize>> {
```
kevinyang372: Not sure what this is

crates/warp_tui/src/transcript_view.rs
```diff
@@ -333,12 +352,20 @@ impl TuiView for TuiTranscriptView {
...
-        TuiScrollable::new(
-            TuiViewportedList::new(self.viewport.clone(), source)
-                .with_vertical_alignment(TuiViewportVerticalAlignment::GrowFromBottom)
-                .finish_scrollable(),
-        )
-        .finish()
+        let viewport = TuiViewportedList::new(self.viewport.clone(), source)
+            .with_vertical_alignment(TuiViewportVerticalAlignment::GrowFromBottom)
+            .with_selection(TuiSelectionConfig::new(Rc::new(word_span)));
+        let selectable = TuiSelectable::new(viewport)
```
kevinyang372: A little surprised we don't need a selection state handle here?

crates/warp_tui/src/transcript_word_selection.rs
```diff
@@ -0,0 +1,69 @@
...
+//! word span using the same configurable smart-selection rules as the GUI
+//! terminal, falling back to word-boundary scanning when no rule matches.
+use string_offset::ByteOffset;
+use warp_core::semantic_selection::SemanticSelection;
+use warpui::SingletonEntity;
+use warpui_core::elements::tui::{point_after_col, TuiContentPoint, TuiRowGlyph, TuiSelectionSpan};
+use warpui_core::AppContext;
+
+/// Resolves transcript word selection using configured smart-selection rules.
+pub(super) fn word_span(
```
kevinyang372: Can we use the WordBoundariesPolicy here? I feel like this should largely be shared logic with our GUI semantic selection

crates/warp_tui/src/tui_block_list_viewport_source.rs
```diff
@@ -47,12 +47,14 @@ enum TuiBlockListVisibleItemKind {
     TerminalBlock(BlockId),
     AgentBlock(ViewHandle<TuiAIBlock>),
 }
+/// One original row range and its newly measured height.
+type TuiBlockListHeightChange = (Range<usize>, usize);
```
kevinyang372: nit: can this be a struct

crates/warp_tui/src/tui_block_list_viewport_source.rs
```diff
@@ -137,12 +140,78 @@ impl TuiBlockListViewportSource {
...
             .update_rich_content_heights_in_lines(line_heights);
     }
 
+    /// Collects one `(old row range, new height)` pair per re-measured
+    /// rich-content item, in canonical block-list order, computed against the
+    /// cached heights *before* the new ones are written back. The viewport
+    /// drains these pairs via `take_selection_row_resizes` and rebases any
+    /// active selection around them, so selected rows stay anchored to the
+    /// same content when blocks grow or shrink above or inside the selection.
+    fn rich_content_row_resizes(
```
kevinyang372, line 154: I am a little worried about the performance of these resize mappings

I wonder if we actually need to eagerly recompute these selection updates on resize (e.g. what's claude code / codex's behavior here?)

crates/warpui_core/src/elements/tui/selectable/cells.rs
```diff
@@ -0,0 +1,94 @@
+use std::ops::Range;
+
+use ratatui::buffer::CellWidth;
+
+use super::super::TuiBuffer;
+
+/// An absolute cell position in selectable content.
```
kevinyang372: I feel like I've seen a few variations of this type -- would it make sense to combine / pull them into one type?

crates/warpui_core/src/elements/tui/selectable.rs
```diff
@@ -0,0 +1,190 @@
...
+type CopyCallback = Box<dyn FnMut(String, &mut TuiEventContext, &AppContext)>;
+
+/// Resolves a semantic word unit from rendered row glyphs.
+pub type TuiWordSelectionResolver =
+    Rc<dyn Fn(TuiContentPoint, u16, &[TuiRowGlyph], &AppContext) -> Option<TuiSelectionSpan>>;
+
+/// Semantic-word policy for a selectable viewport.
+#[derive(Clone)]
+pub struct TuiSelectionConfig {
+    pub(crate) word_resolver: TuiWordSelectionResolver,
```
kevinyang372: Similar to GUI, this should take word boundary policy and embed the resolving logic internally with the given policy imo

crates/warpui_core/src/elements/tui/selectable.rs
```diff
@@ -0,0 +1,190 @@
...
+        ctx: &mut TuiLayoutContext,
+        app: &AppContext,
+    ) -> TuiSelectionEventResult;
+
+    /// Paints persistent selection state over normal child rendering.
+    fn render_selection(&self, area: TuiRect, buffer: &mut TuiBuffer, ctx: &mut TuiPaintContext);
+}
+
+/// Delegates selection interaction to a child and owns only external callbacks.
+pub struct TuiSelectable<Child> {
```
kevinyang372, line 88: As mentioned above, I think Selectable should hold the selection handle

crates/warpui_core/src/elements/tui/viewported_list.rs
```diff
@@ -59,7 +89,212 @@ impl TuiViewportedListState {
 
     /// Returns whether the requested viewport position follows the content end.
     pub fn is_at_end(&self) -> bool {
-        matches!(*self.0.borrow(), TuiViewportPosition::End)
+        matches!(self.0.borrow().position, TuiViewportPosition::End)
+    }
+    /// Clears the viewport selection, returning whether state existed.
+    pub fn clear_selection(&self) -> bool {
```
kevinyang372: These methods should belong to selectable not viewport list imo

crates/warpui_core/src/elements/tui/viewported_list.rs
kevinyang372, line 1: On a high-level, here is how I would divide the responsibility of Selectable wrapper and viewported list

Selectable is responsible for:
1. Keeping the selection state
2. Updating the wrapped selectable element when the selection range changed
3. Notify the parent view on selection changed events

Viewported list is responsible or:
1. On a selection changed event from selectable, resolve the raw screen coordinate space to actual selected text + ranges to store in the selection state
2. Render the selection based on the passed down selection state

harryalbert: yeah this makes a lot of sense. I updated it so that:
•  TuiSelectable owns:
   ◦  active gesture capture;
   ◦  mouse-down/drag/mouse-up transitions;
   ◦  anchor and focus updates;
   ◦  persistent selection state;
   ◦  resize rebasing;
   ◦  selection-start and copy callbacks;
   ◦  repaint notifications.
•  TuiViewportedList implements a narrower adapter contract:
   ◦  screen position → content-space selection span;
   ◦  selection span → copied text;
   ◦  viewport-specific edge autoscroll;
   ◦  row-resize reporting;
   ◦  rendering from passed-down selection state.

crates/warpui_core/src/elements/tui/selectable.rs
harryalbert: makes sense

crates/warp_tui/src/transcript_view.rs
harryalbert: addressed w/ the main refactor

crates/warpui_core/src/elements/tui/selectable.rs
harryalbert: yeah that makes sense, I'll update

crates/warp_tui/src/transcript_word_selection.rs
harryalbert: for sure

crates/warp_tui/src/terminal_session_view.rs
harryalbert: That makes sense. I moved selection interactions onto TuiTranscriptView and exposed them as SelectionStarted / SelectionEnded view events. TuiTerminalSessionView now subscribes to those events for its input-clearing, clipboard, and footer behavior.

crates/warp_tui/src/tui_block_list_viewport_source.rs
harryalbert: codex doesn't do any highlight work, so it's not super applicable, but claude code _does not_ do any anchoring (highlights just stay where they are no matter how content streams). I think this behavior is kind of bleh though and we have the opportunity to make highlighting feel pretty native w/o too much extra work.

But I definitely think you're right about performance concerns. I updated it so that we:
* Only generate resize mappings for blocks whose measured row height actually changed.
* Resolve each block’s old row range through the block list’s existing view-ID index instead of scanning the entire block-height tree.
* Clear width-invalid selections before child layout, so we don’t rebase coordinates that are about to be discarded.
* Skip selection rebasing when no selection exists or when the changed content is entirely below the selected rows.

so the big O profile is now more like (according to the agent):
O(M log N + K log K + A × S log S), instead of adding an O(N) full-tree scan on every layout. Here, M is the small set of measured rich-content blocks, K is the subset whose row height actually changed, A is the subset of those changes that can affect the selection, and S is the snapshotted selection size.

In the normal stable or streaming-without-reflow case, K = 0. For ordinary streaming reflow, it’s generally K = 1, and we only touch the snapshot if that block is above or intersects the selection.

crates/warp_tui/src/transcript_view.rs
harryalbert: This was retrieving the block’s absolute row range before removal so an existing selection could be cleared or shifted around the deleted row, but it's not necessary (esp. not on the transcript view).

was removed anyways w/ the performance refactor from below

crates/warpui_core/src/elements/tui/selectable/cells.rs
harryalbert: yeah I'll make a shared `TuiGridPoint`

crates/warp_tui/src/tui_block_list_viewport_source.rs
harryalbert: yessir

kevinyang372 (approved):
Nice work!

Did look into the detailed viewporting / autoscrolling logic (I trust 5.6 sol can cook these) but the high-level design makes sense to me

PR #13570, opened 2026-07-10
[CODE-1829] Render agent task lists in the TUI transcript

ianhodge:
## Description
- Render agent `UpdateTodos` and `MarkAsCompleted` operations inline in the TUI transcript with status glyphs, collapse state, and semantic theme styling.
- Derive task status from canonical conversation todo state so active, stopped, completed, cancelled, and superseded lists stay consistent with Agent Mode.
- Extend the generic TUI collapsible header to support independently styled spans, keeping hover underline on the task label without underlining the leading glyph or chevron.
- Refresh affected transcript blocks when conversation-wide todo or status projections change, preventing cached older task lists from showing stale state.
- Document the implementation and redraw model in `specs/CODE-1829/TECH.md`.

## Linked Issue
[CODE-1829: Agent task list](https://linear.app/warpdotdev/issue/CODE-1829/agent-task-list)

## Testing
- `./script/format --check`
- `./script/check_no_inline_test_modules`
- `cargo nextest run -p warp_tui --no-fail-fast` (113 passed)
- `cargo nextest run -p warp history_model` (71 passed)
- `cargo nextest run -p warpui_core --features tui collapsible` (2 passed)
- `cargo build -p warp_tui --bin warp-tui-oss`
- Manually verified expanded/collapsed task rendering, status styling, and label-only hover underline in the macOS TUI using tmux capture.

## Agent Mode
- [Conversation](https://staging.warp.dev/conversation/2ed8235b-00a7-4d25-a377-f503987542fc)
- [Run](https://oz.staging.warp.dev/runs/019f489c-78ce-7639-ac2b-1dcacc92c04d)

CHANGELOG-IMPROVEMENT: Agent task lists now render inline in the terminal UI transcript.

### DIFF
--- a/app/src/ai/blocklist/history_model.rs
+++ b/app/src/ai/blocklist/history_model.rs
@@ -948,9 +948,8 @@ impl BlocklistAIHistoryModel {
     }
 
     /// Returns the render status of one todo item in the conversation's todo
-    /// history (see [`AIConversation::todo_status`]). A narrow projection so
-    /// TUI consumers don't need the whole `AIConversation` exported through
-    /// `tui_export`.
+    /// history (see [`AIConversation::todo_status`]) — a narrow projection
+    /// for consumers that don't need the whole `AIConversation`.
     pub fn todo_status(
         &self,
         conversation_id: &AIConversationId,
@@ -959,8 +958,8 @@ impl BlocklistAIHistoryModel {
         self.conversation(conversation_id)?.todo_status(todo_id)
     }
 
-    /// Returns the conversation's active (most recent) todo list, if any. A
-    /// narrow projection for TUI consumers (see [`Self::todo_status`]).
+    /// Returns the conversation's active (most recent) todo list, if any — a
+    /// narrow projection (see [`Self::todo_status`]).
     pub fn active_todo_list(&self, conversation_id: &AIConversationId) -> Option<&AIAgentTodoList> {
         self.conversation(conversation_id)?.active_todo_list()
     }
--- a/app/src/ai/blocklist/history_model.rs
+++ b/app/src/ai/blocklist/history_model.rs
@@ -26,15 +26,16 @@ use super::persistence::{PersistedAIInput, PersistedAIInputType};
 use super::RequestInput;
 use crate::ai::agent::api::ServerConversationToken;
 use crate::ai::agent::conversation::{
-    AIConversation, AIConversationId, ConversationStatus, ServerAIConversationMetadata,
+    AIConversation, AIConversationId, ConversationStatus, ServerAIConversationMetadata, TodoStatus,
     UpdateConversationError,
 };
 use crate::ai::agent::task::helper::{MessageExt, ToolCallExt};
 use crate::ai::agent::task::TaskId;
+use crate::ai::agent::todos::AIAgentTodoList;
 use crate::ai::agent::{
     AIAgentActionId, AIAgentExchange, AIAgentExchangeId, AIAgentInput, AIAgentOutputStatus,
-    CancellationReason, FinishedAIAgentOutput, MessageId, RenderableAIError, RequestCost,
-    Suggestions,
+    AIAgentTodoId, CancellationReason, FinishedAIAgentOutput, MessageId, RenderableAIError,
+    RequestCost, Suggestions,
 };
 use crate::ai::artifacts::Artifact;
 use crate::ai::document::ai_document_model::AIDocumentModel;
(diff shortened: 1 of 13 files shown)


crates/warp_tui/src/agent_block.rs
```diff
@@ -47,37 +49,47 @@ enum TuiAIBlockSection {
...
 }
 
-/// UI state for a single thinking block.
+/// UI state for a single collapsible section.
 #[derive(Default)]
-struct ThinkingBlockState {
-    /// Manual collapse override. `None` means the default: collapsed iff
-    /// reasoning has finished, so a block streams expanded and auto-collapses
-    /// on finish unless the user has toggled it — a recorded override wins
+struct CollapsibleSectionState {
```
ianhodge, line 81: Felt it made sense to combine these into a more generic struct but lmk if that's not ideal here

crates/warp_tui/src/transcript_view.rs
```diff
@@ -248,6 +261,53 @@ impl TuiTranscriptView {
         }
     }
 
+    /// Marks every agent block on this terminal surface dirty. Todo-list
+    /// updates are conversation-wide: a newly active list can restyle rows in
+    /// any older exchange as cancelled.
+    fn mark_all_agent_blocks_dirty(&mut self, ctx: &mut ViewContext<Self>) {
```
ianhodge: I don't think this and the function below will cause downstream perf, it should only happen when a task state is invalidated

needing to invalidate the whole convo felt weird to me, but agent mode seemed to think it was necessary

crates/warpui_core/src/elements/tui/collapsible.rs
```diff
@@ -26,13 +26,38 @@ pub fn tui_disclosure_chevron(collapsed: bool) -> &'static str {
...
+/// the header is hovered. Spans let a header mix hover treatments — e.g.
+/// underline only the label while a decorative leading glyph (whose own
+/// strokes would clash with an underline) keeps its resting style.
+pub struct TuiCollapsibleHeaderSpan {
+    text: String,
+    style: TuiStyle,
+    hover_style: TuiStyle,
+}
+
+impl TuiCollapsibleHeaderSpan {
```
ianhodge: This allowed the hover state of the task to only be over the title and not the hamburger menu (which looked weird IMO)

app/src/ai/blocklist/history_model.rs
```diff
@@ -946,6 +947,24 @@ impl BlocklistAIHistoryModel {
             .map(|conversation| conversation.status())
     }
 
+    /// Returns the render status of one todo item in the conversation's todo
+    /// history (see [`AIConversation::todo_status`]). A narrow projection so
+    /// TUI consumers don't need the whole `AIConversation` exported through
```
kevinyang372: uber nit: I don't think we need to specifically mention TUI here -- this is a generic API

crates/warp_tui/src/agent_block.rs
kevinyang372: Yeah this makes sense to me

crates/warp_tui/src/agent_block.rs
```diff
@@ -462,12 +493,40 @@ impl TuiAIBlock {
...
                     app,
                 ),
+                TuiAIBlockSection::TodoList { message_id, todos } => {
+                    // Statuses resolve against the conversation's todo
+                    // history at render time, so superseded lists restyle
+                    // without needing a dedicated invalidation. Items the
+                    // conversation no longer knows belong to a superseded
+                    // list (matching the GUI's fallback).
+                    let history = BlocklistAIHistoryModel::as_ref(app);
+                    let rows: Vec<(String, TodoStatus)> = todos
```
kevinyang372, line 607: Just confirming this matches GUI's default behavior? I am a little worried about doing a linear scan on each conversation for TODO list status but that might be fine if we are also doing this in GUI

crates/warp_tui/src/transcript_view.rs
kevinyang372: Hmmm I think it should be pretty trivial to only invalidate blocks that contain a todo list message right? If so, we should do that here

I understand this is the same behavior as in GUI, but I am worried invalidating all blocks might be more expensive in TUI because we eagerly layout blocks in the cell grid for viewporting

kevinyang372 (approved):

crates/warp_tui/src/transcript_view.rs
ianhodge: that's a good point, let me see if we can be more efficient here

crates/warp_tui/src/agent_block.rs
ianhodge: yep, I agree it's a bit iffy but this matches what we do in the GUI

crates/warp_tui/src/transcript_view.rs
ianhodge: Okay fixed this by having TuiAIBlock track whether its output contains a todo message and both the todo-list and conversation-status handlers only dirty those blocks

Issue #11091, opened 2026-05-16
IME regression on macOS: duplicate characters, double-press Backspace, and Tab inserts literal \t with Boshiamy IME (broken between Stable 0.2026.04.29 and 0.2026.05.13)


ethan-tsai-tsai:
### Pre-submit Checks

- [x] I have searched Warp bugs and there are no duplicates
- [x] I have searched Warp known issues page and my issue is not there

### Describe the bug

Between Warp Stable `0.2026.04.29.08.57.01` and `0.2026.05.13.09.15.03`, IME handling on macOS regressed. With the Boshiamy (無蝦米) input method active, three distinct symptoms appear that did not exist in the prior Stable build:

1. **Duplicate characters** — committed characters appear twice (committed char + the raw latin keystrokes used during composition leak into the terminal).
2. **Backspace requires two presses** to delete a single character (first press is consumed by the IME composition buffer, second press reaches the shell).
3. **Tab inserts a literal `\t`** instead of triggering shell completion (Tab is sent as a raw character before the IME finishes committing).

All three symptoms are facets of the same underlying issue: IME composition keystrokes are leaking into the terminal instead of being fully consumed via `NSTextInputClient`, contrary to standard macOS `NSResponder` behavior.

### To reproduce

1. macOS with the Boshiamy (無蝦米) input method selected as the active input source.
2. Open Warp Stable `0.2026.05.13.09.15.03`.
3. Type any Chinese character via Boshiamy. Example: keys `v o r` → expected `項`.
4. **Observed**: `項項` or `項vor` appears in the prompt instead of just `項`.
5. Press Backspace once → nothing visibly deleted. Press a second time → one character is deleted.
6. Press Tab during or immediately after composition → a literal `\t` is inserted; shell completion is not triggered.

Reproduces 100% of the time on the affected build. Does not reproduce at all on Warp Stable `0.2026.04.29.08.57.01` on a second Mac with the same OS family and the same Boshiamy install.

### Expected behavior

- IME composition keystrokes should be fully consumed by the input method during composition. Only the committed text should reach the terminal.
- Backspace during/after composition should behave atomically — one press deletes one character.
- Tab should trigger shell completion, not insert a literal tab character, regardless of whether an IME is active.
- Behavior should match `0.2026.04.29.08.57.01`, Terminal.app, and iTerm2.

### Screenshots, videos, and logs

[...]

PR #13606, opened 2026-07-11
Fix stale classic completions after user edits

### LINKED ISSUE #13605: Classic completions remain stale after Tab then Backspace
### Summary

### Steps to reproduce

   ```
   cd Do
   ```

### Observed

### Expected

### Environment

- macOS (Apple Silicon).

maxmilian:
## Description

With Classic Completions enabled, pressing <kbd>Tab</kbd> and then <kbd>Backspace</kbd> back past the point completion was anchored to left the old suggestion list on screen (stale), and backspacing to an empty prefix re-showed the entire original result set. This is the `TODO` in `update_tab_completion_menu` ("if you hit tab and backspace, the result set won't go away (stale)").

Non-classic completions already close the menu when the text up to the cursor no longer starts with the buffer text that triggered completion. Classic completions were exempted from that check so that Tab **cycling** (which rewrites the buffer to each candidate, diverging from the original query) keeps working. The problem is that the exemption applied to *all* edits, so genuine user edits were also allowed to keep a stale result set.

### Fix

Distinguish edits the completion system applied itself from edits the user made, using the editor's recorded last action:

```rust
let is_user_edit = !matches!(
    self.editor.as_ref(ctx).get_last_action(ctx),
    Some(
        PlainTextEditorViewAction::AcceptCompletionSuggestion
            | PlainTextEditorViewAction::CycleCompletionSuggestion
    )
);
```

The classic exemption now only applies to system-applied edits (accept / cycle). A user edit (typing, backspacing, pasting) still runs the original-query validity check and invalidates + closes the menu when the input retreats past the completion boundary or becomes incompatible. `SelectionChanged` (a cursor move, not a buffer edit) passes `is_user_edit = false`.

This reads first-class recorded editor state (`get_last_action`), set on the undo stack *before* `EditorEvent::Edited` is emitted — not a heuristic. Cycling stays exempt in both fuzzy and non-fuzzy modes because both record `CycleCompletionSuggestion`.

## Linked Issue

Closes #13605

## Testing

- [x] I have manually tested my changes locally with `./script/run` (WarpOss built with the classic-completions feature).

Added two model tests in `app/src/terminal/input_tests.rs`:

- `test_classic_tab_completions_close_after_user_backspace` — **red before this change** (menu stayed `CompletionSuggestions` / re-showed the full result set after backspacing to `cd `), **green after** (mode becomes `Closed`).
- `test_classic_tab_completions_keep_menu_open_while_cycling` — passes before and after; guards against a cycling regression.

Validation on the affected package:

- `cargo nextest run -p warp classic_tab_completions` → 2 passed
- `cargo nextest run -p warp tab_completion` → 24 passed
- `cargo check -p warp` → ok
- `cargo clippy -p warp --tests -- -D warnings` → no issues
- `cargo fmt --all -- --check` → clean

### Screenshots / Videos

Manually verified in the real app (WarpOss built with the classic-completions feature). Same evidence is on the linked issue (#13605).

**1. `cd Do` + <kbd>Tab</kbd>** — the classic completion menu opens (Documents/Downloads).

menu open

**2. After <kbd>Backspace</kbd> back past the completion boundary** — the menu closes (the fix; previously it stayed open with stale candidates, and an empty prefix re-showed the whole original result set).

menu closed

**3. `cd Do` + <kbd>Tab</kbd> × 3** — the menu stays open and the selection advances, so Tab cycling is unaffected.

cycling preserved

### Out of scope

IME marked-text composition and undo/redo are `skip_undo` edits that don't record a fresh action, so they keep the pre-existing classic exemption behavior (unchanged by this PR). IME completion interaction is tracked separately (#11091).

### DIFF
--- a/app/src/terminal/input.rs
+++ b/app/src/terminal/input.rs
@@ -10531,6 +10531,19 @@ impl Input {
                 // e.g., we don't want to sync EditorEvent::CmdUpOnFirstRow.
                 self.send_input_sync_event(edit_origin, ctx);
 
+                // Distinguish edits the completion system applied itself (accepting or
+                // cycling through a candidate) from edits the user made (typing,
+                // backspacing, pasting). System-applied edits are allowed to diverge from
+                // the original completion query so that classic cycling keeps working;
+                // user edits still have to be revalidated against that query.
+                let is_user_edit = !matches!(
+                    self.editor.as_ref(ctx).get_last_action(ctx),
+                    Some(
+                        PlainTextEditorViewAction::AcceptCompletionSuggestion
+                            | PlainTextEditorViewAction::CycleCompletionSuggestion
+                    )
+                );
+
                 let mode = self.suggestions_mode_model.as_ref(ctx).mode().clone();
                 match &mode {
                     InputSuggestionsMode::CompletionSuggestions {
@@ -10604,6 +10617,7 @@ impl Input {
                                 replacement_start,
                                 buffer_text_original.as_str(),
                                 &completion_results,
+                                is_user_edit,
                                 ctx,
                             );
                             if should_close {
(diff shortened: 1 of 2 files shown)


zachbai:
@seemeroland can you take a look at this?

seemeroland (approved):
Nice fix, thank you! I verified your branch locally as well

PR #13624, opened 2026-07-13
Mint IAP tokens for Oz runners via WIF (bootstrapped by injected OIDC JWT)

IsaiahWitzke:
## What
Sandboxed Oz runners authenticate to staging through IAP by self-minting an IAP-valid ID token via Workload Identity Federation, replacing the server-injected IAP bearer token.

## Why
After IAP was enabled on staging, runners 403 at the proxy — the injected `WARP_IAP_TOKEN` was minted as warp-server's own SA identity (not an authorized IAP member) and never refreshed (REMOTE-1370). The runner image has no `gcloud`/ADC, so the existing gcloud path can't run there.

## How
Mint chain (native HTTP, no gcloud):
1. **Warp OIDC JWT** (audience = the `oz-agents` WIF provider). Leg 1 prefers a server-injected **bootstrap JWT** (`WARP_STAGING_IAP_BOOTSTRAP_JWT`) so a cold runner needn't call the IAP-gated identity-token endpoint (chicken-and-egg: minting the JWT via `issue_task_identity_token` itself goes through IAP-gated warp-server). Once the injected JWT expires, refreshes mint a fresh one via `issue_task_identity_token` — reachable through IAP by then because a valid IAP token exists.
2. **GCP STS** token exchange → federated access token.
3. **IAM `generateIdToken`** impersonating `iap-access@` (audience = the IAP OAuth client ID), an authorized IAP member.

Flows through the existing `IapManager` lifecycle (proactive + reactive refresh), closing the no-refresh gap.

## Notes
- The **WIF provider audience is derived from the bootstrap JWT's `aud` claim** — no client config needed. The server already mints that JWT with `aud` = the provider resource name (GCP STS requires the subject token's `aud` to match the provider), so the value is already delivered to the runner. The env var persists for the process lifetime, so its `aud` stays readable even after the token itself expires (needed for refresh-time minting). This keeps a single source of truth (terraform → warp-server config) with no third copy in channel config.
- Gated on runner context (ambient-agent run id); local staging clients keep the gcloud impersonation path.
- Old `WARP_IAP_TOKEN` handling removed.
- Minting abstracted behind `IapIdentityTokenMinter` (impl in the app crate) so `warp_server_client` doesn't depend on the managed-secrets stack.

## Dependencies (must land together)
- warp-terraform: `serviceAccountTokenCreator` for the staging Oz Warp-team principalSet on `iap-access@`.
- warp-server: inject `WARP_STAGING_IAP_BOOTSTRAP_JWT` (a task-identity OIDC JWT whose `aud` is the WIF-provider resource name) into runner containers — separate PR. This JWT's `aud` is now the sole source of the federation audience, so it must be set correctly.

## Testing
- `cargo nextest run -p warp_server_client iap` — 21 passing (incl. new `aud`-parsing cases).
- `cargo check -p warp` (lib.rs wiring) passes; `cargo clippy -p warp_server_client -p warp_core --all-targets --tests -- -D warnings` clean.
- Not exercised end-to-end (needs the grant + injection live).

[Conversation](https://staging.warp.dev/conversation/89ccf173-dc13-461c-b90f-6bee7612f989)

### DIFF
--- a/app/src/server/iap_identity_minter.rs
+++ b/app/src/server/iap_identity_minter.rs
@@ -1,0 +1,43 @@
+use std::sync::Arc;
+use std::time::Duration;
+
+use futures::FutureExt as _;
+use vec1::vec1;
+use warp_managed_secrets::client::{IdentityTokenOptions, ManagedSecretsClient};
+use warp_server_client::iap::IapIdentityTokenMinter;
+use warpui::r#async::BoxFuture;
+
+/// Mints Warp-signed OIDC identity tokens for the runner-context IAP Workload
+/// Identity Federation flow, backed by the managed-secrets client. Lives in the
+/// app crate so `warp_server_client` need not depend on the managed-secrets
+/// stack.
+pub struct ManagedSecretsIapMinter {
+    client: Arc<dyn ManagedSecretsClient>,
+}
+
+impl ManagedSecretsIapMinter {
+    pub fn new(client: Arc<dyn ManagedSecretsClient>) -> Self {
+        Self { client }
+    }
+}
+
+impl IapIdentityTokenMinter for ManagedSecretsIapMinter {
+    fn mint_identity_token(
+        &self,
+        audience: String,
+        requested_duration: Duration,
+    ) -> BoxFuture<'static, anyhow::Result<String>> {
+        let client = self.client.clone();
+        async move {
+            let token = client
+                .issue_task_identity_token(IdentityTokenOptions {
+                    audience,
+                    requested_duration,
+                    subject_template: vec1!["principal".to_string()],
+                })
+                .await?;
+            Ok(token.token)
+        }
+        .boxed()
+    }
+}
--- a/crates/warp_server_client/src/iap.rs
+++ b/crates/warp_server_client/src/iap.rs
@@ -36,8 +36,7 @@ const WIF_MINT_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 /// Env var carrying a Warp OIDC task-identity JWT (audience = the WIF provider
 /// resource name), injected by warp-server so a cold sandboxed runner can
 /// bootstrap its first IAP mint without calling the IAP-gated identity-token
-/// endpoint. This is NOT the IAP bearer token — it is the subject token for the
-/// STS exchange.
+/// endpoint.
 const STAGING_IAP_BOOTSTRAP_TOKEN_ENV_VAR: &str = "WARP_STAGING_IAP_BOOTSTRAP_JWT";
 
 pub type PathResolver = Box<dyn Fn(&mut AppContext) -> BoxFuture<'static, Option<String>>>;
--- a/crates/warp_server_client/src/iap.rs
+++ b/crates/warp_server_client/src/iap.rs
@@ -38,7 +38,7 @@ const WIF_MINT_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
 /// bootstrap its first IAP mint without calling the IAP-gated identity-token
 /// endpoint. This is NOT the IAP bearer token — it is the subject token for the
 /// STS exchange.
-const INJECTED_OIDC_JWT_ENV_VAR: &str = "WARP_IAP_OIDC_JWT";
+const INJECTED_OIDC_JWT_ENV_VAR: &str = "WARP_STAGING_IAP_BOOTSTRAP_JWT";
 
 pub type PathResolver = Box<dyn Fn(&mut AppContext) -> BoxFuture<'static, Option<String>>>;
 
--- a/crates/warp_server_client/src/iap.rs
+++ b/crates/warp_server_client/src/iap.rs
@@ -5,6 +5,7 @@ use anyhow::Result;
 use base64::Engine;
 use blocking::unblock;
 use instant::Instant;
+use serde::{Deserialize, Serialize};
 use warp_core::channel::IapConfig;
 use warpui_core::r#async::{BoxFuture, FutureExt as _, Timer};
 use warpui_core::{AppContext, Entity, ModelContext, SingletonEntity};
@@ -12,7 +13,6 @@ use warpui_core::{AppContext, Entity, ModelContext, SingletonEntity};
 use websocket::connect_error_http_response;
 
 const PROACTIVE_REFRESH_BUFFER: Duration = Duration::from_secs(5 * 60);
-const INJECTED_TOKEN_ENV_VAR: &str = "WARP_IAP_TOKEN";
 
 const BASE_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(30);
 const MAX_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(5 * 60);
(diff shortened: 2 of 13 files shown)


app/src/server/iap_identity_minter.rs
```diff
@@ -0,0 +1,43 @@
...
+            let token = client
+                .issue_task_identity_token(IdentityTokenOptions {
+                    audience,
+                    requested_duration,
+                    subject_template: vec1!["principal".to_string()],
+                })
+                .await?;
+            Ok(token.token)
+        }
+        .boxed()
```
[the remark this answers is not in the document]
IsaiahWitzke: fixed

crates/warp_server_client/src/iap.rs
```diff
@@ -372,6 +445,122 @@ impl Entity for IapManager {
...
+            subject_token_type: SUBJECT_TOKEN_TYPE_ID_TOKEN,
+        })
+        .timeout(WIF_MINT_REQUEST_TIMEOUT)
+        .send()
+        .await
+        .map_err(|err| anyhow::anyhow!("STS token exchange request failed: {err:#}"))?;
+    let status = response.status();
+    if !status.is_success() {
+        let body = response.text().await.unwrap_or_default();
+        anyhow::bail!("STS token exchange failed (status {status}): {body}");
```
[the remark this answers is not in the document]
IsaiahWitzke: i kinda want token/principal details to be visible to me when staging IAP fails though... :/

crates/warp_server_client/src/iap.rs
```diff
@@ -22,8 +22,49 @@ const MAX_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(5 * 60);
...
+const SUBJECT_TOKEN_TYPE_ID_TOKEN: &str = "urn:ietf:params:oauth:token-type:id_token";
+const REQUESTED_TOKEN_TYPE_ACCESS_TOKEN: &str = "urn:ietf:params:oauth:token-type:access_token";
+const WIF_IDENTITY_TOKEN_DURATION: Duration = Duration::from_secs(60 * 60);
+const WIF_MINT_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Env var carrying a Warp OIDC task-identity JWT (audience = the WIF provider
+/// resource name), injected by warp-server so a cold sandboxed runner can
+/// bootstrap its first IAP mint without calling the IAP-gated identity-token
+/// endpoint.
+const STAGING_IAP_BOOTSTRAP_TOKEN_ENV_VAR: &str = "WARP_STAGING_IAP_BOOTSTRAP_JWT";
```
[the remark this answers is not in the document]
IsaiahWitzke: ive changed it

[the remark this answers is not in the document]
IsaiahWitzke: https://github.com/warpdotdev/warp-server/pull/12600/changes#diff-c8c83c5a20b814ed763f5707142fd3c811aba2f7a67575bb4b159c1b677dff73R18

IsaiahWitzke:
Im not totally sure if it's even possible to test e2e on local ( think i would need to give some sort of warp local SA (or my personal GCP account) the perms to create WIF OIDC JWTs with the aud claim set as the new id provider i set up?)  -- it's def is _**easier**_ on staging, so my plan is just to test e2e there.

crates/warp_server_client/src/iap_tests.rs
```diff
@@ -118,3 +119,28 @@ fn get_cached_failed_uses_valid_previous_token() {
...
+        serde_json::from_str(r#"{"token": "an-id-token"}"#).unwrap();
+    assert_eq!(parsed.token, "an-id-token");
+}
+
+#[test]
+fn sts_response_parses_and_ignores_extra_fields() {
+    let parsed: StsTokenExchangeResponse =
+        serde_json::from_str(r#"{"access_token": "federated", "expires_in": 3600}"#).unwrap();
+    assert_eq!(parsed.access_token, "federated");
+}
```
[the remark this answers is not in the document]
IsaiahWitzke: done

crates/warp_server_client/src/iap.rs
```diff
@@ -468,13 +731,116 @@ fn get_expires_at(token: &str) -> Result<Instant> {
...
+        write_owner_only(&tmp, token.as_bytes())?;
+        std::fs::rename(&tmp, &path)
+    }
+
+    #[cfg(unix)]
+    fn write_owner_only(path: &std::path::Path, bytes: &[u8]) -> std::io::Result<()> {
+        use std::io::Write as _;
+        use std::os::unix::fs::OpenOptionsExt as _;
+
+        let mut file = std::fs::OpenOptions::new()
```
IsaiahWitzke: chose to just write to a file over secret store since most of our runners are just on linux boxes, and IIUC, a secret store on linux is more just like security via obfuscation ?

crates/warp_server_client/src/iap.rs
```diff
@@ -206,17 +246,33 @@ impl IapManager {
...
-            );
+    /// Returns `true` when a currently-valid (unexpired) IAP token is available.
+    /// Used to gate warp-server requests until IAP access is established.
+    pub fn has_valid_token(&self) -> bool {
+        self.state
+            .as_ref()
+            .is_some_and(|s| s.get_cached().is_some())
+    }
+
+    /// Establishes IAP access for a blocking caller: triggers a refresh
```
bnavetta: nit: what should callers wait on for success?

crates/warp_server_client/src/iap.rs
```diff
@@ -272,38 +333,105 @@ impl IapManager {
...
+        ctx.emit(IapManagerEvent::StateChanged);
+        ctx.notify();
+
+        let minter = mint.minter.clone();
+        let iap_audience = state.audiences().to_string();
+        let service_account_email = state.service_account_email().to_string();
+
+        ctx.spawn(
+            async move {
+                // The env var persists for the process lifetime, so its `aud` stays
```
bnavetta, line 375: Ah clever

crates/warp_server_client/src/iap.rs
```diff
@@ -372,6 +500,141 @@ impl Entity for IapManager {
...
+    // separate client config.
+    let federation_audience = parse_aud_from_jwt(&injected_jwt)
+        .ok_or_else(|| anyhow::anyhow!("injected OIDC JWT has no readable `aud` claim"))?;
+
+    // Leg 1: obtain the Warp-signed OIDC JWT used as the STS subject token.
+    let identity_token =
+        resolve_wif_identity_token(injected_jwt, &federation_audience, &minter).await?;
+
+    // Leg 2: exchange the JWT at GCP STS for a federated access token.
+    let response = http_client::Client::new()
```
bnavetta, line 584: Is it worth trying to use a single client for the process lifetime? This is probably infrequent enough that a connection pool doesn't help, but the client might cache other state

crates/warp_server_client/src/iap.rs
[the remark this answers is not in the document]
bnavetta: Also presumably if the exchange failed then there isn't a token from STS to leak 🤔

crates/warp_server_client/src/iap.rs
```diff
@@ -468,13 +731,116 @@ fn get_expires_at(token: &str) -> Result<Instant> {
     Ok(Instant::now() + Duration::from_secs(secs_remaining))
 }
 
-fn parse_exp_from_jwt(token: &str) -> Option<u64> {
+fn decode_jwt_payload(token: &str) -> Option<serde_json::Value> {
```
bnavetta, line 739: Probably not necessary here since we act as the client presenting the JWT, but wondering if we should verify the JWT at all when decoding

crates/warp_server_client/src/iap.rs
```diff
@@ -468,13 +731,116 @@ fn get_expires_at(token: &str) -> Result<Instant> {
...
+    use instant::Instant;
+
+    use super::{CachedToken, get_expires_at};
+
+    /// Discard a cached token with less than this remaining, so we never hand back
+    /// a near-dead credential the mint path should refresh instead.
+    const CACHE_SKEW: Duration = Duration::from_secs(30);
+
+    fn cache_path() -> std::path::PathBuf {
+        warp_core::paths::state_dir()
```
bnavetta: Not that we have a secure state dir on Linux, but may as well use `secure_state_dir().unwrap_or_else(state_dir)` in case we add one

crates/warp_server_client/src/iap.rs
```diff
@@ -468,13 +731,116 @@ fn get_expires_at(token: &str) -> Result<Instant> {
...
+            log::warn!("Failed to write IAP token cache: {err}");
+        }
+    }
+
+    fn write_atomic(token: &str) -> std::io::Result<()> {
+        let path = cache_path();
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent)?;
+        }
+        // Write to a sibling temp file then rename so a reader never observes a
```
bnavetta: Maybe overkill, but we could use a tempfile and [`persist`](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.persist) it, so that if this errors, the temporary token file is cleaned up

crates/warp_server_client/src/iap.rs
bnavetta: Pretty much, yeah. Writing temporary JWTs to files under `$HOME` seems pretty common for CLIs generally

bnavetta (approved):

crates/warp_server_client/src/iap.rs
```diff
@@ -22,8 +22,55 @@ const MAX_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(5 * 60);
...
+const SUBJECT_TOKEN_TYPE_ID_TOKEN: &str = "urn:ietf:params:oauth:token-type:id_token";
+const REQUESTED_TOKEN_TYPE_ACCESS_TOKEN: &str = "urn:ietf:params:oauth:token-type:access_token";
+const WIF_IDENTITY_TOKEN_DURATION: Duration = Duration::from_secs(60 * 60);
+const WIF_MINT_REQUEST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Env var carrying a Warp OIDC task-identity JWT (audience = the WIF provider
+/// resource name), injected by warp-server so a cold sandboxed runner can
+/// bootstrap its first IAP mint without calling the IAP-gated identity-token
+/// endpoint.
+const STAGING_IAP_BOOTSTRAP_TOKEN_ENV_VAR: &str = "WARP_STAGING_IAP_BOOTSTRAP_JWT";
```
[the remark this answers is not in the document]
IsaiahWitzke: renamed the env var [here](https://github.com/warpdotdev/warp-server/blob/214334fccf431076e515ca44dd45121090743023/logic/ai/ambient_agents/dispatch/dispatcher.go#L18)

crates/warp_server_client/src/iap.rs
IsaiahWitzke: yeah... i dont think a (non-trivial) JWT verification addition here actually buys us any security... we're just  decoding the JWT here just to extract aud and exp claims for later STS exchange (security is enforced there) / refresh scheduling.

crates/warp_server_client/src/iap.rs
IsaiahWitzke: AM:

Not worth it, and your instinct is right. A few concrete reasons:
Pooling buys nothing at this cadence. The mint runs at startup, then proactively ~once per token lifetime (tokens ≈1h, refreshed 5min early → roughly hourly), plus the rare reactive challenge. reqwest's idle keep-alive connections get dropped long before the next mint (default idle timeout is ~90s), so a persistent client would be re-dialing every time anyway. Same for any DNS cache — expired well before reuse.

> default idle timeout is ~90s

this seems like the killer... I guess i could overwrite this to something like 10hr if we really wanted to reuse the same client 9x more, but im not totally sure of the a) code implications b) if the caching of client data at that point is generally valid anymore

PR #13644, opened 2026-07-13
fix synthetic mouse build scene pass

harryalbert:
## Description

Fix stale TUI hover state when layout moves beneath a stationary pointer. After each presentation, replay the last pointer position and rebuild if hover changes invalidate the view, using the same three-pass safety cap and consecutive synthetic-flip guard as the GUI.

This fixes the issue where, if you clicked on a thinking head and it subsequently moved out from under your mouse (because the row expanded), the header would stay hovered.

## Testing

- [x] confirmed the fix by manually running ./script/run-tui
https://www.loom.com/share/86efb7d1a5a948bfb9e2f8fbcf740faf

## Agent Mode

### DIFF
--- a/crates/warpui_core/src/runtime/mod.rs
+++ b/crates/warpui_core/src/runtime/mod.rs
@@ -119,27 +119,24 @@ impl<T: TuiView, R: TuiTerminal> TuiScreen<T, R> {
         let size = self.terminal.size()?;
         let area = TuiRect::new(0, 0, size.width, size.height);
 
-        let invalidation = ctx.take_all_invalidations_for_window(self.window_id);
-        self.presenter
-            .invalidate(&invalidation, ctx, self.window_id);
-        let mut frame = self.presenter.present(ctx, &self.root_view, area);
-        self.replay_mouse_position(ctx);
-
         // Mirrors the GUI's `build_scene` loop: pointer replay can invalidate
         // hover-dependent layout, requiring another presentation and replay.
-        // Cap at three total presentations so a hover/layout feedback loop
-        // cannot hang the redraw.
-        for _ in 2..=3 {
+        // The first iteration always presents; later ones only run if the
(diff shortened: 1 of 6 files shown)


harryalbert:
* **#13644** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13644?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13644?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warpui_core/src/runtime/mod.rs
```diff
@@ -99,24 +105,65 @@ impl<T: TuiView, R: TuiTerminal> TuiScreen<T, R> {
...
             .invalidate(&invalidation, ctx, self.window_id);
-        let frame = self.presenter.present(ctx, &self.root_view, area);
+        let mut frame = self.presenter.present(ctx, &self.root_view, area);
+        self.replay_mouse_position(ctx);
+
+        // Mirrors the GUI's `build_scene` loop: pointer replay can invalidate
+        // hover-dependent layout, requiring another presentation and replay.
+        // Cap at three total presentations so a hover/layout feedback loop
+        // cannot hang the redraw.
+        for _ in 2..=3 {
```
kevinyang372: Erm why we don't move everything into this loop? Weird to see one iteration outside and a 2..=3 condition for looping

kevinyang372 (approved):

crates/warpui_core/src/runtime/mod.rs
harryalbert: that was to avoid a .expect around variables defined in the loop existing, but agreed it looks a bit weird. I'll make it a 1-3 loop and just add the expect back in

PR #13659, opened 2026-07-13
Remove OrchestrationViewerStreamer and OwnerOrchestrationAncestorStreamer flags

cephalonaut:
## Description

Removes two feature flags that have been in the default feature set (enabled in all builds) since they stabilized.

## What

**OrchestrationViewerStreamer**: Removes the legacy REST polling path from OrchestrationViewerModel (fetch_children, apply_children_fetch, schedule_next_poll, maybe_kick_polling and the associated struct fields/constants). The SSE-driven path via OrchestrationEventStreamer is now unconditional. Also drops the flag guard in stop_orchestration_polling in terminal_manager.rs.

**OwnerOrchestrationAncestorStreamer**: Parents now always use AncestorRunId { include_self: true } in desired_sse_filter. The UnsupportedRunIdCount variant, its report_error! handler, and MAX_RUN_ID_STREAM_FILTER are removed — the RunIds parent fallback is unreachable because a parent always has self_run_id by the time children exist (the server must assign the parent a run before any child can be created). The RunIds fallback is retained only for non-parent (child-only) conversations.

## Why

Both flags were always on in production. Removing them eliminates ~750 lines of dead code and simplifies the SSE filter logic.

## Testing

- cargo clippy -p warp --tests -- -D warnings passes
- 94 tests across the three affected modules pass (orchestration_viewer_model, orchestration_event_streamer, terminal_manager)
- Legacy polling-path tests deleted; flag override guards removed from remaining tests; two now-invalid without-flag streamer tests deleted

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -683,8 +683,6 @@ default = [
     "oz_handoff",
     "handoff_local_cloud",
     "run_agents_tool",
-    "orchestration_viewer_streamer",
-    "owner_orchestration_ancestor_streamer",
     "pending_user_query_indicator",
     "queue_slash_command",
     "queued_prompts_v2",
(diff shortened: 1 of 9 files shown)


advait-m (approved):
sweet!

Issue #13652, opened 2026-07-13
Markdown viewer: support common raw-HTML subset (img sizing, tables, details, anchors)


fbartho:
### Pre-submit Checks

- [x] I have searched Warp feature requests and there are no duplicates
- [x] I have searched Warp docs and my feature is not there

### Describe the solution you'd like?

Warp's Markdown Viewer should support the common subset of raw inline HTML that real-world repo markdown relies on (this is standard practice under GitHub Flavored Markdown, and is what GitHub, GitLab, and most other markdown renderers already support). Today the viewer renders plain Markdown syntax reasonably well, but silently drops or ignores raw HTML tags, which breaks a lot of READMEs, docs, and notes that use HTML for layout control that Markdown syntax alone can't express.

Prioritized list of tags, most important first:

1. **`<img>` with `width` / `height` / `align`** — sizing control for images. This is the one that bit me directly (see repro below).
2. **`<a href="…">` and `<a id="…">`** — links, plus named anchors used as deep-link targets for hand-built tables of contents.
3. **Tables: `<table>` / `<thead>` / `<tbody>` / `<tr>` / `<th>` / `<td>`** — needed for cells that require a line break (`<br>`) or a `colspan`/`rowspan`, which Markdown table syntax cannot express at all.
4. **`<details>` / `<summary>`** — collapsible sections, extremely common in READMEs, changelogs, and bug-report templates (including this repo's own issue templates elsewhere on GitHub).
5. **`<br>`** — explicit line breaks, most often needed inside table cells.
6. **`<kbd>`** — keyboard-shortcut styling. Feels especially on-brand for a terminal app's own markdown viewer.
7. **`<sub>` / `<sup>`** — subscript/superscript.
8. **`<p align="center">` / `<div align="center">`** — centered "hero" headers, very common at the top of README files.
9. **`<picture>` / `<source media="(prefers-color-scheme: …)">`** — light/dark logo variant swapping, standard pattern for README badges/logos.

I'd be happy to see this land as "the common subset above," even if full arbitrary-HTML rendering (with all its sanitization complexity) is out of scope.

### Is your feature request related to a problem? Please describe.

I'm always frustrated when I write locally-generated SVG dashboards into my Markdown notes and Warp's viewer won't let me size them.

Concretely, today in Warp's Markdown Viewer (verified on this build):

Works, but not usefully — renders the image at a small fixed width that ignores the SVG's intrinsic `width`/`height`, and does not scale when the pane or document is resized:

[...]

PR #13660, opened 2026-07-13
docs(specs): markdown viewer <details>/<summary> support (GH13652)

fbartho:
## What

Spec (product.md + tech.md, per CONTRIBUTING's spec-first workflow for feature requests) for teaching the Markdown viewer to render `<details>/<summary>` collapsible sections — a summary row with a disclosure caret that shows/hides its body on click.

Part of #13652 — this is tier-zero tag #2 from that issue's prioritized list, following the `` sizing spec in #13656. Deferring the remaining raw-HTML subset (tables next, then anchors, `<kbd>`, `<picture>`, …) to keep each slice reviewable.

## Why

`<details>/<summary>` is extremely common in READMEs, changelogs, and issue templates. Warp's viewer currently drops the raw HTML, so those sections render as literal tag text or lost content — one of the everyday cases the issue calls out.

## Highlights

- Recognize a block-level `<details>` region with an optional `<summary>`; render the summary as a clickable caret row (▶ collapsed / ▼ expanded) and show/hide the body on click, re-laying-out the document.
- Body renders as ordinary Markdown blocks (paragraphs, lists, code, block images per #13656), so it reuses the existing block renderers.
- Session-persistent per-section expand state, keyed by buffer anchor so it survives edits — mirroring the existing mermaid render-toggle mechanism.
- Default collapsed unless `<details open>`; only the `open` attribute is honored, everything else ignored; no script execution or navigation.
- Graceful degradation: unterminated/malformed `<details>` falls back to literal text without swallowing the rest of the document.

## Honest scope note

This is the architecturally heaviest tier-zero tag. Per-block toggle state and the caret affordance have clean precedents in the editor (mermaid render-toggle, hidden-section), but **the editor has no block that owns child blocks today**. The tech spec therefore recommends an MVP that models the `<details>` body as sibling top-level blocks gated by a hidden-line range (Option A) rather than building a true nested-block container (Option B). That bounds the MVP: nested `<details>` and arbitrary raw HTML inside the body are explicit out-of-scope limitations, not silent failures. If you'd prefer the nested-container model up front, this becomes a design-only spec and the implementation is materially larger — flagging that as the key decision for review.

## Process note

#13652 is currently `enhancement, triaged` (not yet `ready-to-spec`) — opening this ahead of the label per the linked-issue allowance in CONTRIBUTING; happy to sequence differently. Nested under `specs/GH13652/` following the `specs/APP-4319/<slug>/` precedent for one issue split across a chain of PRs.

### DIFF
--- a/specs/GH13652/details-summary/product.md
+++ b/specs/GH13652/details-summary/product.md
@@ -1,0 +1,117 @@
+# PRODUCT.md — Markdown viewer: `<details>/<summary>` collapsible sections
+
+Issue: https://github.com/warpdotdev/warp/issues/13652
+Preceded by: `<img>` sizing spec (`specs/GH13652/`, PR #13656)
+
+## Summary
+
+Warp's Markdown viewer drops raw inline HTML, so `<details>/<summary>` collapsible
+sections — extremely common in READMEs, changelogs, and issue templates — render as
+either literal tag text or dropped content. This spec covers teaching the Markdown
+viewer to recognize a `<details>` block with an optional `<summary>`, render the summary
+as a clickable disclosure row with a caret, and show/hide the body on click.
+
+This is tier-zero tag #2 of issue #13652 (after `<img>` sizing). It is a **larger and
+more architecturally involved** change than `<img>` sizing: unlike an image (a single
+leaf block), a `<details>` owns an arbitrary run of body content, and the collapsed
+state must persist across re-layout. The tech spec documents both a pragmatic MVP that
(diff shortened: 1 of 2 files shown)


fbartho:
@cla-bot check

fbartho:
Pausing this spec in favor of #13345, which targets the same issue (#10259) and is already converging through review — no need for a third competing draft (see also #10462). Happy to reopen this one if that effort stalls. My up-to-date findings on the current rendering behavior are on the issue: https://github.com/warpdotdev/warp/issues/10259#issuecomment-4974120224

PR #13662, opened 2026-07-13
Add TUI conversation management

harryalbert:
## Description

Before this PR, the Warp TUI could resume a conversation only when launched with a server token. There was no way to discover or switch to another local or cloud conversation from the running TUI, and startup restoration could not be cancelled into the provisional new session.

This PR adds a searchable `/conversations` inline menu backed by the shared normalized conversation model and query policy. It lists eligible personal Oz conversations from local and cloud history, preserves selection across live updates, and revalidates command, conversation, target, and restore state before switching. Successful loads replace the sole TUI transcript through the existing GUI restoration path; failed or cancelled loads leave the current conversation intact. Cloud metadata failures fall back to searchable local history with an explicit warning, and Escape or Ctrl-C can cancel both list and startup restoration without allowing stale completion to replace the active session.

### Review guide

**Read**
- `app/src/ai/agent_conversations_model/query.rs` — shared GUI/TUI recency and fuzzy-ranking policy.
- `app/src/ai/agent_conversations_model.rs` — partial cloud-metadata failure state exposed to the TUI.
- `crates/warp_tui/src/conversation_menu.rs` — menu lifecycle, live refresh, query projection, and stable selection.
- `crates/warp_tui/src/terminal_session_view.rs` — acceptance revalidation, loading/cancellation, and conversation replacement.

**Skim**

- `crates/warp_tui/src/inline_menu.rs` and `input/view.rs` — conversation-specific adapter and accepted-action routing onto the infrastructure from PR 1.
- `crates/warp_tui/src/session.rs`, `app/src/tui_export.rs`, and `app/src/terminal/input/slash_commands/mod.rs` — startup target and command/export wiring.

**Can skip**

The remaining files are dependency bookkeeping, copy/layout adjustments, or focused coverage for the behavior above.

## Linked Issue

[CODE-1821 — Conversation management](https://linear.app/warpdotdev/issue/CODE-1821/conversation-management)

## Testing

### Screenshots / Videos

[Loom: TUI conversation management](https://www.loom.com/share/4bf696091ef24d0a8a8eacb1f855830b)

## Agent Mode

CHANGELOG-IMPROVEMENT: Added a searchable `/conversations` menu to the Warp TUI.


harryalbert:
* **#13690** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13690?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13662** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13662?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13662?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13669** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13669?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13668** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13668?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

specs/CODE-1821/TECH.md
```diff
@@ -0,0 +1,181 @@
...
+- A menu-only query sourced from the shared input editor.
+- Loading and empty snapshots using the existing TUI inline-menu presentation.
+
+Opening the menu registers its model ID with `AgentConversationsModel`; closing unregisters it. This activates existing polling and deferred RTC refresh behavior only while the list is visible.
+
+The model subscribes to input changes and `AgentConversationsModelEvent`. Every refresh reruns the current query, preserves selection by stable entry ID, and falls back to the nearest valid index when the selected row disappears.
+
+### Query and order eligible entries
+
+Request personal Oz entries from `AgentConversationsModel::get_entries` and classify each through the TUI’s existing conversation-selection policy. Pass eligible entries to the same pure `query_conversation_entries` helper used by the GUI inline menu. The helper owns the 50-entry recent cap, case-insensitive fuzzy matching, score threshold, score/recency ordering, and 500-result search cap. GUI-only directory filtering and frontend-specific row construction remain outside the helper.
```
moirahuang: i know we talked about it in the other PR but `Oz` here again tripped me up

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -1176,6 +1372,66 @@ impl TuiTerminalSessionView {
         ctx.notify();
     }
 
+    fn handle_accepted_conversation(
+        &mut self,
+        entry_id: AgentConversationEntryId,
+        ctx: &mut ViewContext<Self>,
+    ) {
+        if self.is_conversation_restore_loading() {
```
moirahuang, line 1414: when would we want to respect the older conversation restore? instead of always loading the most recent conversation restore request?

moirahuang (commented):
[Screen Recording 2026-07-13 at 6.12.15 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/26b42df5-f9c3-4ad7-a46a-b5a7814123f0.mov)i'm seeing some thinking blocks not restore while others do restore

harryalbert:
> [Screen Recording 2026-07-13 at 6.12.15 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/26b42df5-f9c3-4ad7-a46a-b5a7814123f0.mov)i'm seeing some thinking blocks not restore while others do restore

​Good catch. Apparently, the GUI specifically hides these empty thinking blocks as they are expected from some model providers. We should do the same from the TUI (updated the PR to do so).

Asked AM to dig in a bit about why these empty thinking blocks exist at all, and it said:
```
The server confirms these records are intentional:

•  OpenAI reasoning models use reasoning.summary = auto, so some turns produce no user-visible reasoning summary.
•  They still return encrypted reasoning metadata needed for conversation continuity.
•  The server therefore persists an AgentReasoning message with:
◦  empty reasoning
◦  populated finished_duration
◦  populated opaque server_message_data
•  ../warp-server/logic/ai/multi_agent/builders/agent_reasoning.go emits the message because duration and metadata are independently valid fields.
•  server_message_data contains provider-specific reasoning IDs/signatures—not displayable thought text.

So the GUI is correct to hide these. The TUI bug is that crates/warp_tui/src/agent_block.rs creates a collapsible section for every Reasoning message without checking whether it has renderable content.
```

crates/warp_tui/src/terminal_session_view.rs
harryalbert: This is mostly a defensive guard, but it's basically protecting against the case where the user:
* opens the conversation menu and selects a conversation
* _while that conversation is loading_ (which should only take a few seconds), they re-open the menu and select another conversation

If that specific edge case happens, we just show a quick warning that says "Another conversation is already loading". They can restore another convo once the other conversation is done loading and is restored, or they can cancel that restoration w/ ctrl c.

It would be more complex to cancel the currently loading conversation and then load the new one instead (and the user can always cancel restoration if it's in progress w/ ctrl-c manually), so I figured this was reasonable behavior (and didn't think it was worth spending time on the more complex solution given that this is solving an edge case in a transient loading state on the order of seconds)

harryalbert: lmk if you disagree though

specs/CODE-1821/TECH.md
harryalbert: Yeah, for now this is just how we refer to them throughout the codebase, but agreed the language is confusing. I'll update the spec

harryalbert:
Here's a demo restoring the same convo in the GUI and TUI and showing that thinking blocks are restored correctly:
https://www.loom.com/share/cada3c94b5324fbb929897f1c0e27629

specs/CODE-1821/TECH.md
moirahuang: 🙇 and mind tracking a PR to refactor / rename this in the codebase please?

crates/warp_tui/src/terminal_session_view.rs
moirahuang: should we disable input when we're loading a conversation? i'm not sure what the GUI prior art is but that feels semantically correct

harryalbert: we disable sending any prompts while the conversation is loading. I think disabling any input would also be reasonable, but I'd kind of rather get it on dogfood and see if folks find the experience confusing, just because it's such a transient state and I'd rather not spend too much time on it yet

moirahuang (approved):
niceee

moirahuang:
### Merge activity

* **Jul 14, 5:29 PM UTC**: @moirahuang merged this pull request with [Graphite](https://app.graphite.com/github/pr/warpdotdev/warp/13662).

PR #13696, opened 2026-07-14
TUI alt screen mouse interactions

harryalbert:
## Description

Stacked on #13626, which adds alt-screen rendering and keyboard forwarding to the Warp TUI.

This PR adds mouse interaction for full-screen terminal applications in the TUI, and unifies the alt-screen mouse-reporting policy and wheel encoding across the GUI and TUI:

**TUI mouse forwarding (`crates/warp_tui/src/alt_screen_view.rs`)**

- Converts crossterm click, release, drag, motion, and wheel events into terminal input using pane-relative cell coordinates (single `cell_point` conversion helper).
- Gating is driven by a small `MouseReportPolicy` (buttons / motion / scroll) derived from the *same* policy functions the GUI uses — `should_intercept_mouse` / `should_intercept_scroll` — now re-exported via `tui_export`. This means the TUI honors the `AltScreenReporting` settings toggles and shared-session reader state identically to the GUI, with no duplicated mode logic.
- Keeps Shift-click available for terminal-native selection instead of forwarding it to the application; synthetic motion and out-of-bounds events are never forwarded.
- Routes the resulting bytes through the TUI's existing raw `ForwardToPty` path.

**Shared wheel encoding (`crates/warp_terminal/src/model/escape_sequences.rs`)**

- `alt_screen_scroll_to_pty_bytes` encodes alt-screen wheel movement for both front-ends:
  - Applications using SGR mouse reporting receive mouse-wheel escape sequences.
  - Applications without mouse reporting, including default Vim, receive the existing SS3 arrow-key fallback.

**GUI (behavior-preserving refactor)**

- `TerminalAction::AltScroll` now carries only event data (`{ delta, point }`); the report-vs-intercept decision moved out of the element into the `alt_scroll` handler, which calls `should_intercept_scroll` and the shared encoder. Bytes written to the PTY are unchanged; zero-delta scrolls now skip the PTY write entirely instead of writing an empty sequence.

## Testing

Automated:

- `cargo nextest run -p warp_terminal -E 'test(test_alt_screen_scroll_to_pty_bytes)'`
- `cargo nextest run -p warp_tui -E 'test(alt_screen_view)'` (policy-driven tests: byte-exact SGR coordinates, per-category gating, SGR-vs-arrow scroll fallback, rejection cases)
- `cargo nextest run -p warp -E 'test(test_alt_screen_select_with_sgr_mouse)'`
- `cargo clippy -p warp --lib -- -D warnings`
- `cargo clippy -p warp_tui --all-targets -- -D warnings`

Manual:

- Verified default Vim scrolls through the shared SS3 fallback.
- Verified Vim with SGR mouse reporting moves its cursor to the clicked terminal cell.
- Verified Vim exits back to the normal TUI surface.

### Screenshots / Videos

[Video demo (Loom)](https://www.loom.com/share/3172496bde19417088526b336e90ef2b)

## Agent Mode

### DIFF
--- a/app/src/terminal/alt_screen/alt_screen_element.rs
+++ b/app/src/terminal/alt_screen/alt_screen_element.rs
@@ -472,20 +472,13 @@ impl AltScreenElement {
             .alt_screen_mut()
             .accumulate_lines_to_scroll(delta);
 
-        if should_intercept_scroll(&self.model.lock(), app) {
-            ctx.dispatch_typed_action(TerminalAction::AltScroll { delta });
-        } else {
-            let point = self.coord_to_point(local_position);
-
-            ctx.dispatch_typed_action(TerminalAction::AltMouseAction(
-                MouseState::new(
-                    MouseButton::Wheel,
-                    MouseAction::Scrolled { delta },
-                    Default::default(),
-                )
-                .set_point(point),
-            ));
-        }
+        let point = self.coord_to_point(local_position);
+        let report_mouse = !should_intercept_scroll(&self.model.lock(), app);
+        ctx.dispatch_typed_action(TerminalAction::AltScroll {
+            delta,
+            point,
+            report_mouse,
+        });
         true
     }
 
--- a/app/src/terminal/alt_screen/alt_screen_element.rs
+++ b/app/src/terminal/alt_screen/alt_screen_element.rs
@@ -22,7 +22,7 @@ use warpui::{
     SizeConstraint,
 };
 
-use super::{should_intercept_mouse, should_intercept_scroll};
+use super::should_intercept_mouse;
 use crate::appearance::Appearance;
 use crate::pane_group::SplitPaneState;
 use crate::settings::EnforceMinimumContrast;
@@ -445,7 +445,6 @@ impl AltScreenElement {
         delta: Vector2F,
         precise: bool,
         ctx: &mut EventContext,
-        app: &AppContext,
     ) -> bool {
         let cell_height = self.grid_render_params.size_info.cell_height_px;
         let delta = if precise {
@@ -473,12 +472,7 @@ impl AltScreenElement {
             .accumulate_lines_to_scroll(delta);
 
         let point = self.coord_to_point(local_position);
-        let report_mouse = !should_intercept_scroll(&self.model.lock(), app);
-        ctx.dispatch_typed_action(TerminalAction::AltScroll {
-            delta,
-            point,
-            report_mouse,
-        });
+        ctx.dispatch_typed_action(TerminalAction::AltScroll { delta, point });
         true
     }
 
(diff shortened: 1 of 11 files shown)


harryalbert:
* **#13696** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13696?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13696?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13626** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13626?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

app/src/terminal/alt_screen/alt_screen_element.rs
```diff
@@ -472,20 +471,8 @@ impl AltScreenElement {
             .alt_screen_mut()
             .accumulate_lines_to_scroll(delta);
 
-        if should_intercept_scroll(&self.model.lock(), app) {
```
kevinyang372, line 475: ooc why the change here?

crates/warp_terminal/src/model/escape_sequences.rs
```diff
@@ -351,6 +352,37 @@ impl<T: ModeProvider> ToEscapeSequence<T> for MouseState {
     }
 }
 
+/// Encodes alt-screen wheel movement as mouse reports or SS3 arrow keys.
+pub fn alt_screen_scroll_to_pty_bytes<T: ModeProvider>(
```
kevinyang372, line 356: This is not specific to alt screen right? (e.g. could it be used for long-running command scrolling as well?)

crates/warp_tui/src/alt_screen_view.rs
```diff
@@ -48,6 +57,86 @@ impl AltScreenElement {
     }
 }
 
+/// Which alt-screen mouse reports the active app and user settings allow.
+#[derive(Clone, Copy)]
+pub(crate) struct MouseReportPolicy {
```
kevinyang372: Should this live in the terminal content wrapper now?

kevinyang372 (approved):

app/src/terminal/alt_screen/alt_screen_element.rs
harryalbert: This is just a code re-org. The `should_intercept_scroll` check moved into `TerminalView::alt_scroll`. The element now passes the scroll delta and cursor point through one action, and the shared encoder chooses between an SGR mouse report and the existing SS3 arrow-key fallback. This lets the GUI and TUI share the same encoding logic.

crates/warp_terminal/src/model/escape_sequences.rs
harryalbert: The SGR wheel encoding is reusable, and the long-running path already gets that through `MouseState::to_escape_sequence`. The combined helper is alt-screen-specific because its non-mouse-reporting fallback sends SS3 arrow keys. For a long-running command, the equivalent fallback is to scroll Warp’s transcript rather than write arrows to the PTY, so I don’t think the full helper should be shared there.

crates/warp_tui/src/alt_screen_view.rs
harryalbert: good call, updated

PR #13707, opened 2026-07-14
extract orchestration edit state into a frontend-neutral module

harryalbert:
## Description
Extracts the orchestration run-config editing domain out of the GUI-only `orchestration_controls.rs` into a frontend-neutral `app/src/ai/orchestration/` module, so the TUI orchestration card (upstack in this stack) can share the exact same state, transitions, and validation as the GUI confirmation card and plan-card config block. No behavior change intended.

The module is split into:
- `config_state.rs` — `OrchestrationConfigState` (the run-wide model/harness/execution-mode/auth fields; renamed from the old `OrchestrationEditState`) and `AuthSecretSelection`.
- `edit_state.rs` — the new `OrchestrationEditState` (config state + per-harness model memory) and the `apply_execution_mode_change` / `apply_harness_change` / `revalidate_after_catalog_change` transitions, consolidated from loose GUI helper functions into state methods. Each has a `_core` variant with injected catalog callbacks so transitions are unit-testable without app singletons.
- `validation.rs` — Accept gating and picker-visibility predicates, plus `harness_is_selectable` (extracted from inline picker filtering so both frontends filter identically).
- `providers.rs` — `AppContext`-backed catalog lookups, default host/environment/auth-secret resolution, and persistence helpers.

## Review guide
**New code:**
- `app/src/ai/orchestration/edit_state.rs` — the only substantially new code: the consolidated transitions and the config-state/model-memory split. Previously this logic was interleaved with picker repopulation in `apply_harness_change` / `apply_execution_mode_change` / `repopulate_all_pickers` in `orchestration_controls.rs`; behavior should be identical.
- `app/src/ai/blocklist/inline_action/run_agents_card_view.rs` — the card's `RunAgentsEditState` is split into run-wide config vs card-only fields, and per-harness model memory moves off the view onto `OrchestrationEditState`. Mostly mechanical, but this is where a behavior regression would hide.

**Skim (moved code, bodies unchanged):**
- `app/src/ai/orchestration/config_state.rs`, `providers.rs`, and most of `validation.rs` — cut/paste from `orchestration_controls.rs` (and `can_execute_with_auth_secret` / `populate_default_auth_secret_for_execution` from the `run_agents.rs` executor), with GUI-typed `ViewContext` params relaxed to `AppContext`.
- `app/src/ai/blocklist/inline_action/orchestration_controls.rs` — almost entirely deletions plus re-exports of the moved items; what remains is GUI rendering/picker population only.

**Skim (mechanical caller churn):**
[...]

PR #13714, opened 2026-07-14
add option snapshots and adapt GUI orchestration pickers onto them

harryalbert:
## Description
Adds `app/src/ai/orchestration/snapshots.rs`: one plain-data `OptionSnapshot` builder per orchestration configuration field (location, harness, model, API key, host, environment). A snapshot carries rows, ordering, badges, disabled reasons, load status, footer affordances, and the current selection — with no GUI types — so the GUI pickers and the TUI configuration pages (upstack) render the exact same option lists and cannot drift.

The GUI's picker-population helpers in `orchestration_controls.rs` are rewritten to consume the snapshots instead of querying catalogs inline: `populate_harness_picker`, `populate_model_picker_for_harness`, `populate_environment_picker`, `populate_host_picker`, `populate_auth_secret_picker_for_harness`, and `sync_picker_selections` now map snapshot rows to menu items and only keep genuinely GUI concerns (icons, brand colors, rich Oz model rows via `available_model_menu_items`, dropdown trigger labels). No behavior change intended.

## Review guide
**New Code**
- `app/src/ai/orchestration/snapshots.rs` — the snapshot types and the per-field builders. Each `*_snapshot(state, ctx)` reads the live catalogs and delegates to a pure `build_*_snapshot` core (inputs injected for testability). The row inclusion/ordering/disabled-reason policy here was lifted from the old picker-population code as is, so no need to read too carefully.
- `app/src/ai/blocklist/inline_action/orchestration_controls.rs` — the picker rewrites. Again, pretty much pure moves, so no need to read super carefully.

## Testing
Again, no behavior change

## Agent Mode

### DIFF
--- a/app/src/ai/orchestration/snapshots.rs
+++ b/app/src/ai/orchestration/snapshots.rs
@@ -1,0 +1,564 @@
+//! Plain-data option lists both frontends render from. Builders mirror
+//! the GUI picker population logic exactly; each is a thin
+//! `AppContext`-reading wrapper over a pure core so the mirroring logic
+//! is unit-testable without app singletons.
+
+use ai::agent::action::RunAgentsExecutionMode;
+use warp_cli::agent::Harness;
+use warpui::{AppContext, SingletonEntity};
+
+use super::config_state::{AuthSecretSelection, OrchestrationConfigState};
+use super::providers::{
+    get_base_model_choices, resolve_default_host_slug, resolve_recent_host_slug,
+    ORCHESTRATION_ENV_NONE_LABEL, ORCHESTRATION_WARP_WORKER_HOST,
+};
+use crate::ai::auth_secret_types::auth_secret_types_for_harness;
+use crate::ai::cloud_environments::CloudAmbientAgentEnvironment;
+use crate::ai::connected_self_hosted_workers::ConnectedSelfHostedWorkersModel;
+use crate::ai::harness_availability::{AuthSecretFetchState, HarnessAvailabilityModel};
+use crate::ai::harness_display;
+use crate::ai::local_harness_setup::{
+    local_harness_is_product_enabled, local_harness_setup_state, LocalHarnessSetupState,
+};
+use crate::cloud_object::CloudObjectLookup as _;
+use crate::LLMPreferences;
+
+const DEFAULT_MODEL_LABEL: &str = "Default model";
+/// Label shown in the auth secret picker when no secret is selected
+/// (the child agent will inherit credentials from its environment).
+pub(crate) const AUTH_SECRET_INHERIT_LABEL: &str = "Skip (advanced)";
+const CUSTOM_HOST_LABEL: &str = "Custom host…";
+const AUTH_SECRETS_LOAD_FAILED_MESSAGE: &str = "Unable to load secrets";
+
+/// Row id for the Cloud location option.
+#[cfg_attr(not(feature = "tui"), allow(dead_code))]
+pub const LOCATION_CLOUD_ID: &str = "cloud";
+/// Row id for the Local location option.
+#[cfg_attr(not(feature = "tui"), allow(dead_code))]
+pub const LOCATION_LOCAL_ID: &str = "local";
+
+/// One selectable row in an option snapshot. Carries no GUI types.
+#[derive(Debug, Clone, PartialEq)]
+pub struct OptionRow {
+    pub id: String,
+    pub label: String,
+    /// Harness identifier for rows representing harnesses; each frontend
+    /// maps it to its own icon (GUI `Icon`) or glyph/color (TUI).
+    pub harness: Option<Harness>,
+    pub badge: Option<OptionBadge>,
+    pub disabled_reason: Option<String>,
+}
+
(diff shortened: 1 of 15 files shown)


harryalbert:
* **#13716** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13716?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#13717](https://github.com/warpdotdev/warp/pull/13717) <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13717?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#13830](https://github.com/warpdotdev/warp/pull/13830) <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13830?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#13817** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13817?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13771** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13771?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#13714** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13714?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13714?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#13707** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13707?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

app/src/ai/orchestration/snapshots.rs
```diff
@@ -0,0 +1,566 @@
...
+    Failed { message: String },
+    Empty { message: String },
+}
+
+/// Trailing affordance below the option list.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum OptionFooter {
+    /// Free-form text entry (e.g. custom host slug).
+    CustomText { label: String },
+    /// "New API key…" affordance. Emitted for GUI parity; the TUI ignores
```
moirahuang: do we need to emit it if we don't actually use it? even though it's for GUI parity it seems like it might just be unnecessary code

app/src/ai/orchestration/snapshots.rs
```diff
@@ -0,0 +1,566 @@
...
+                get_base_model_choices(llm_prefs, ctx, is_local)
+                    .partition(|llm| llm.id.as_str().starts_with("auto"));
+            let (custom_models, other_models): (Vec<_>, Vec<_>) = rest
+                .into_iter()
+                .partition(|llm| llm_prefs.custom_llm_info_for_id(&llm.id).is_some());
+            let choices: Vec<ModelChoiceInput> = auto_models
+                .into_iter()
+                .chain(custom_models)
+                .chain(other_models)
+                .map(|llm| ModelChoiceInput {
```
moirahuang, line 305: i know this wasn't code you wrote but this came up in my readout review. should we care about other fields from `LLMInfo` like `disable_reason`?

app/src/ai/orchestration/snapshots.rs
```diff
@@ -0,0 +1,566 @@
...
+            continue;
+        }
+        rows.push(OptionRow {
+            badge: Some(OptionBadge::Connected),
+            ..OptionRow::new(slug.clone(), slug.clone())
+        });
+        known_slugs.push(slug);
+    }
+    if let Some(slug) = recent_host.filter(|s| !s.trim().is_empty()) {
+        if !known_slugs
```
moirahuang: i don't fully understand this logic here. it seems like we're saying if the slug isn't known we label it as recent?

moirahuang (approved):
i know not all of the comments are bc of your newly added code! but they came up when i used `/readout` so i figured i'd flag. going to approve to unblock though

app/src/ai/orchestration/snapshots.rs
harryalbert: This is used by the GUI to decide whether to render the “New API key…” item and synchronize its trigger label. Removing it would require the GUI to independently derive that capability again. The comment is misleading, though—I’ll reword it to clarify that the GUI consumes the footer while the TUI intentionally ignores it because resource creation isn’t supported there

app/src/ai/orchestration/snapshots.rs
harryalbert: yeah good call, we should. I'll fix

app/src/ai/orchestration/snapshots.rs
harryalbert: `recent_host` is already sourced from the persisted `last_selected_host`; this condition isn’t inferring recency from an unknown slug. It only checks whether that recent slug was already added as the default, warp, or a connected host. If so, the earlier classification wins; otherwise we add the separate recent row. 

Agreed though that known_slugs makes this unclear — I'll  rename to `added_slugs` and add a short comment before this block.

PR #13744, opened 2026-07-14
Add V0 MCP implementation

kevinyang372:
### What

Adds Phase 1 support for independently configured global MCP servers in the headless Warp TUI:

- Uses a TUI-owned, channel-specific `.mcp.json` rather than reading GUI MCP installations or credentials.
- Discovers and hot-reloads global stdio and HTTP/SSE servers, with last-known-good behavior for malformed or partially written config files.
- Starts configured servers after the TUI session is ready and gives stdio servers the `warp-tui` process environment/PATH.
- Adds OAuth authorization through an ephemeral `127.0.0.1` loopback callback, while preserving the GUI's existing custom-scheme flow.
- Stores TUI OAuth credentials separately in secure storage and supports logging out per server.
- Adds an `/mcp` inline menu for server status and lifecycle actions, plus an MCP summary in zero state.
- Gates the feature behind the dogfood `TuiMcpServers` feature flag.

### Why

The TUI previously had no way to configure or inspect MCP servers. Reusing GUI state would couple two independently installed clients and make config, credentials, callbacks, and lifecycle behavior difficult to reason about. This establishes a global-only TUI foundation while keeping GUI and TUI configuration and secrets physically separate.

### How

- Adds a TUI-specific managed path and file watcher with typed config diagnostics, stale-event protection, and atomic-replacement handling.
- Reuses the existing file-based/templatable MCP runtime through a TUI-facing model that exposes only global servers.
- Refactors OAuth callback delivery behind custom-scheme and loopback modes. The loopback receiver binds to `127.0.0.1:0`, validates callback path/state/peer, limits request size, and times out after five minutes.
- Requires a separately registered loopback-capable TUI client for statically configured OAuth providers; GUI client registrations are never inferred or reused.
- Integrates MCP into the TUI's shared input-suggestions lifecycle so `/mcp`, model, skills, conversation, and slash-command menus remain mutually exclusive.

Implementation plan: https://staging.warp.dev/drive/notebook/z91AX6nD49vuSFVwbXQWqa

## Linked Issue

N/A — scoped as Phase 1 of TUI MCP support.

## Testing

Automated validation:

- `./script/format`
- `cargo nextest run -p mcp` — 23/23 passed
- `cargo nextest run -p warp_tui` — 224/224 passed after rebasing
- Targeted Warp MCP manager, watcher, and slash-command tests — 110/110 passed
- `cargo clippy -p mcp -p warp_tui --all-targets --tests -- -D warnings`
- `git diff --check`

Manual validation with `./script/run-tui` under tmux:

- Verified missing-config zero state and the channel-specific config path shown by `/mcp`.
- Hot-loaded a temporary stdio MCP through the inherited PATH and discovered its tool.
- Confirmed invalid JSON surfaces a diagnostic while retaining the running last-known-good server.
- Confirmed deleting the config stops the removed server.
- Completed independent TUI OAuth for Linear and Sentry and discovered their tools.

Known provider limitation: Figma advertises dynamic client registration but returns HTTP 403 for clients outside the Figma MCP Catalog. Supporting Figma's remote server requires a separately approved, loopback-capable Warp TUI client registration; this implementation intentionally does not reuse the GUI registration. Figma Desktop's local MCP endpoint remains usable without this remote OAuth registration.

## Agent Mode

CHANGELOG-NONE


kevinyang372:
* **#13744** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

app/src/ai/execution_profiles/profiles.rs
```diff
@@ -1474,6 +1474,8 @@ impl AIExecutionProfilesModel {
             }
             TemplatableMCPServerManagerEvent::LegacyServerConverted
             | TemplatableMCPServerManagerEvent::StateChanged { uuid: _, state: _ }
+            | TemplatableMCPServerManagerEvent::AuthenticationRequired { uuid: _ }
```
peicodes, line 1477: Could we name these to be clear that they're for TUI and not for GUI? (Unless that's not true)

peicodes (approved):
This looks reasonable to me, but I'd like @vkodithala to take a look at the changes specific to file-based MCP to audit nothing breaks there

app/src/ai/mcp/file_based_manager.rs
```diff
@@ -48,24 +59,38 @@ impl FileBasedMCPManager {
...
         match event {
             FileMCPWatcherEvent::ConfigParsed {
+                config_path,
                 root_path,
                 provider,
                 servers,
             } => {
+                self.config_diagnostics_by_path.remove(config_path);
                 self.apply_parsed_servers(root_path.clone(), *provider, servers.clone(), ctx);
+                ctx.emit(FileBasedMCPManagerEvent::ServersChanged);
```
vkodithala: question: should this event be emitted from `apply_parsed_servers`? and ditto for the emission from `ConfigRemoved` below (we may want to emit from `remove_servers_for_root_provider`)

IIUC reaching `handle_watcher_event` just means that there's been some change to a file-based MCP config, not that the actual set of servers has changed (the config change could be a no-op in servers parsed, which doesn't map to the semantic definition of an event like `ServersChanged`). it's a thin edge case in any case so i don't care all that much

app/src/ai/mcp/file_based_manager.rs
```diff
@@ -28,10 +29,20 @@ pub struct FileBasedMCPManager {
...
     pending_scan_auto_started_servers_by_root:
         HashMap<PathBuf, HashMap<MCPProvider, HashSet<Uuid>>>,
+    /// Latest read or parse diagnostic for each config path.
+    ///
+    /// This lives beside the parsed server snapshot rather than in a frontend:
+    /// an invalid update preserves the last-known-good servers while any
+    /// consumer can query the current config health. A successful parse or
+    /// removal clears the diagnostic for that path.
+    config_diagnostics_by_path: HashMap<PathBuf, FileMCPConfigDiagnostic>,
+    defer_global_warp_autostart: bool,
```
vkodithala, line 43: nit: could we add some comments that explain _why_ we defer global Warp autostart for the TUI? might be helpful, was a little hard for me to grok. OOS and idrc much though

app/src/ai/mcp/file_based_manager_tests.rs
```diff
@@ -73,6 +74,8 @@ fn subscribe_events(
                     .extend(installation_uuids.iter().copied());
             }
             FileBasedMCPManagerEvent::PurgeCredentials { .. } => {}
+            FileBasedMCPManagerEvent::ServersChanged
+            | FileBasedMCPManagerEvent::ConfigDiagnosticChanged => {}
```
vkodithala: nit: can we combine these 3 events into the same block?

app/src/ai/mcp/file_mcp_watcher.rs
```diff
@@ -528,8 +547,13 @@ impl FileMCPWatcher {
         was_added: bool,
         ctx: &mut ModelContext<Self>,
     ) {
-        if was_deleted {
+        // Atomic replacements can be reported as a delete and add in the same
+        // update. Parse the replacement without transiently removing the
+        // last-known-good servers.
+        if was_deleted && !was_added {
```
vkodithala, line 553: nice!

vkodithala (approved):
this roughly makes sense, few nits but nothing blocking. thanks @kevinyang372!

PR #13762, opened 2026-07-15
docs(specs): tab config boolean parameters (GH11134)

### LINKED ISSUE #11134: Feature request: boolean/checkbox parameter type for tab configs
## Summary

Tab configs currently only support `text`, `branch`, and `repo` parameter types. When a user wants to offer a simple on/off toggle (e.g., "set upstream automatically"), they have to use a `text` parameter and ask the user to type `yes` or `no`. This is clunky and error-prone. A native checkbox or boolean parameter type would make these workflows much cleaner.

## Problem

In tab config TOML files, the `[params.<name>]` section supports only `type = "text"`, `type = "branch"`, and `type = "repo"`. There is no way to present a simple true/false or yes/no choice as a checkbox or toggle in the parameter fill-in modal.

For example, a user might want a worktree tab config with an option like "Set upstream to base branch?" that is checked by default. Today the only workaround is:

```toml
[params.set_upstream]
type = "text"
description = "Set upstream to base branch? (yes/no)"
default = "yes"
```

This requires the user to type a magic string (`yes`/`no`) and relies on the shell author to write a conditional that checks for an exact string match. It also prevents the UI from rendering a compact, intuitive control.

1. Create or edit a tab config TOML file.
2. Add a parameter that should be a boolean choice.
3. Try to declare `type = "checkbox"` or `type = "boolean"` under `[params.<name>]`.
4. Observe that the type is rejected or ignored because it is not in the supported set (`text`, `branch`, `repo`).

### Desired behavior

A new parameter type — e.g., `type = "boolean"` or `type = "checkbox"` — that renders a checkbox in the fill-in modal and produces a true/false value that can be interpolated into commands.

Example usage:

```toml
[params.set_upstream]
type = "boolean"
description = "Set upstream to base branch"
default = true
```

In `commands`, the parameter could be referenced as `{{set_upstream}}` and evaluate to `"true"` / `"false"` (or `1` / `0`), making it easy to use in shell conditionals without magic strings.

## Artifacts

None attached

## Warp version

Unknown

## Operating system

macOS

rasitakyol:
## Description

Adds product and technical specs for boolean parameters in tab configs.

The proposal uses `type = "boolean"`, a native TOML boolean default, a keyboard- and screen-reader-accessible checkbox in the parameter form, and lowercase `true` / `false` interpolation across titles, directories, and commands. Existing string parameters and existing tab configs remain unchanged.

The technical spec grounds the change in the current parser, form, rendering, and test paths at `9e19f0741e3224c1bf8311c0223fd5f4d4a2e260` and maps all product invariants to automated and manual validation.

## Linked Issue

Closes #11134.

- [x] Screenshots are not applicable to this docs-only spec PR.

## Testing

Docs-only validation:

- `git diff --check origin/master...HEAD`
- Confirmed the branch changes only `specs/GH11134/product.md` and `specs/GH11134/tech.md`

### Screenshots / Videos

Not applicable; this PR defines behavior and implementation plans without changing runtime UI.

## Agent Mode


rasitakyol:
@cla-bot check

specs/GH11134/tech.md
```diff
@@ -0,0 +1,259 @@
...
+- Declare the module from `app/src/tab_configs/mod.rs`. Register its fixed
+  bindings from the existing `params_modal::init` entry point so app startup
+  does not gain a second tab-config initialization call.
+- Store `checked: bool`, the parameter name/optional description, a persistent
+  `MouseStateHandle`, and whether the child currently has keyboard focus.
+- Render the existing themed checkbox primitive with a clickable visible name.
+  Apply the modal's normal label/description typography and a visible focus
+  treatment; the params modal must not render a second copy of those labels for
+  this variant.
+- Handle mouse activation and a typed `Toggle` action. Register `Space` for that
```
moirahuang, line 105: can we check if we already handle `Space` as part of the existing checkbox component?

moirahuang (approved):
overall this looks good to me!

specs/GH11134/tech.md
rasitakyol: Checked: the existing `Checkbox` component does not register keyboard activation or a `Space` binding; it only builds the visual `Hoverable`. Its current tab-config usage wires mouse activation externally with `.on_click(...)`. Therefore, the typed `Toggle` action and `Space` binding proposed for the focused child view provide the missing keyboard behavior rather than duplicate existing component behavior. I can clarify this explicitly in the spec if preferred.

moirahuang: we should directly update the existing checkbox component to have keyboard activation and the space binding instead of only implementing it here. if we do so, we should broading the testing to make it more comprehensive to make sure it doesn't regress other existing usage of the checkbox

PR #13792, opened 2026-07-15
[MCP] Add well-known MCP type

Legoben:
## Description

Adds a `MCPSpec::WellKnown` variant to support non-UUID managed MCP identifiers (e.g. `"linear"`). Previously, `warp_id` fields were required to be UUIDs; now any non-empty, non-UUID string is treated as a well-known id that the server resolves. This allows new integrations to be introduced server-side without requiring a client update.

**Parsing behavior:**

- CLI: bare identifiers matching `[A-Za-z0-9_-]+` that are not UUIDs are parsed as `MCPSpec::WellKnown` instead of falling through to inline JSON.
- Config file: `warp_id` values that are not UUIDs become `MCPSpec::WellKnown`; empty `warp_id` values are rejected with an error.
- Validation no longer requires `warp_id` to be a UUID — only that it is non-empty.

**Resolution behavior:**

- `MCPSpec::WellKnown` resolves via the same `create_managed_mcp_client_config` GraphQL mutation as UUID specs, passing the id as an opaque `ID!`.
- Resolution is best-effort: if the server does not recognize the id or the integration is disconnected, the server is skipped with a warning rather than failing run setup.
- The `ManagedMcpClient::create_managed_mcp_client_config` signature is updated from `uid: Uuid` to `uid: String` to accommodate both UUIDs and well-known ids.

**Config serialization:**

- `MCPSpec::WellKnown` round-trips through the MCP servers map as `{ "warp_id": "<id>" }`, consistent with how UUID specs are represented.

## Testing

New unit tests cover:

- Well-known ids parsed from CLI and config file become `MCPSpec::WellKnown`.
- Well-known ids resolve via the managed MCP client and produce ephemeral installations.
- Resolution failures for well-known ids skip the server without failing run setup or dropping other specs.
- Empty `warp_id` values are rejected at both the config file and validation layers.
- Non-identifier, non-JSON strings fall through to the inline-JSON path.

## Agent Mode

### DIFF
--- a/app/src/ai/agent_sdk/config_file.rs
+++ b/app/src/ai/agent_sdk/config_file.rs
@@ -101,12 +101,15 @@ fn supported_keys_context() -> String {
 /// Convert an unwrapped `mcp_servers` map into runtime MCP specs for AgentDriver.
 ///
 /// Behavior:
-/// - Entries with `warp_id` become `MCPSpec::Uuid`.
+/// - Entries with a UUID `warp_id` become `MCPSpec::Uuid`.
+/// - Entries with any other non-empty `warp_id` (e.g. `"linear"`) become `MCPSpec::WellKnown`;
+///   the server owns the set of recognized ids and unknown ids are skipped at resolution.
 /// - Entries with `command`/`url` remain as inline JSON (`MCPSpec::Json`) containing the unwrapped server map.
 pub fn mcp_specs_from_mcp_servers(
     mcp_servers: &Map<String, Value>,
 ) -> anyhow::Result<Vec<MCPSpec>> {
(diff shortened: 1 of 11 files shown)


Legoben:
* **#13792** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13792?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13792?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

dagmfactory (approved):

MaggieShan (approved):

PR #13808, opened 2026-07-16
fix(user-config): shell-quote worktree paths with spaces in generated commands (#11144)

### LINKED ISSUE #11144: New Worktree Config fails with spaces in the path

lonexreb:
Resolves #11144.

## Problem

Creating a New Worktree Config for a repository whose absolute path contains a space (e.g. `~/Developer/2026-05 Site da Jô`) generates broken commands. The shell splits the unquoted path on whitespace, so `git worktree add` receives the wrong number of positional arguments and prints its `usage:` help instead of creating the worktree; the follow-up `cd` misbehaves the same way.

## Root cause

`materialize_default_worktree_config` bakes the computed worktree path into the generated tab config's shell `commands` as **literal text**, not as a param. Param values are shell-quoted at render time by `render_tab_config` (via `shell_words::quote`), but this literal path bypasses that quoting entirely — so a space in the repo/workspace name splits the command.

## Fix

Shell-quote the literal path prefix in `materialize_default_worktree_config` using the same `shell_words::quote` the runtime path uses. Only the prefix is quoted: the trailing `{{autogenerated_branch_name}}` placeholder is resolved (and independently shell-quoted) later at render, so it stays outside the quotes — adjacent shell quotes concatenate into a single argument. Quoting is a no-op for space-free paths, so existing behavior and tests are unchanged.

## Tests

Adds `test_materialized_worktree_config_shell_quotes_paths_with_spaces`, which materializes the default template for `/tmp/my project`, renders it, and asserts via `shell_words::split` that both generated commands parse with the spaced worktree path as exactly **one** shell argument:

- `git worktree add -b my-feature <path with space>` → 6 args
- `cd <path with space>` → 2 args

Existing materialize tests pass unchanged (quoting no-op on space-free paths). `cargo fmt`, `cargo clippy --tests`, and unit tests are clean locally.

## Validation

This is a non-visual logic fix in config generation — behavior is fully captured by the deterministic unit test above (the exact generated command strings before/after are asserted).


lonexreb:
Addressed in [`7711fc60`](https://github.com/warpdotdev/warp/pull/13808/commits/7711fc60): extracted the prefix-quoting into a shared `shell_quoted_generated_worktree_path_string` helper in `tab_config.rs` and applied it to `build_worktree_config_toml` as well (both the autogenerated and manual-name cases). `materialize_default_worktree_config` now calls the same helper instead of its inline copy. Added shell-splitting regression tests for both `build_worktree_config_toml` cases with a spaced repo path — all 35 worktree tests pass, fmt/clippy clean.

/oz-review

app/src/tab_configs/tab_config.rs
```diff
@@ -58,6 +58,23 @@ pub(crate) fn generated_worktree_path_string(repo_path: &Path, worktree_name: &s
         .to_string()
 }
 
+/// Like [`generated_worktree_path_string`], but shell-quotes the literal path prefix so the
+/// generated `git worktree add` / `cd` commands survive shell parsing when the repo or data
+/// directory name contains spaces (#11144). Only the literal prefix is quoted: the trailing
```
moirahuang: no need to reference the issue number here

app/src/user_config/mod.rs
```diff
@@ -278,7 +278,10 @@ pub(crate) fn materialize_default_worktree_config(
     repo_path: &str,
     pane_type: &str,
 ) -> Result<(String, TabConfig), String> {
-    let worktree_path = crate::tab_configs::tab_config::generated_worktree_path_string(
+    // The worktree path is baked as literal text into shell `commands`, so shell-quote it to
+    // keep paths containing spaces from splitting `git worktree add` / `cd` into extra
+    // arguments (#11144).
+    let worktree_path = crate::tab_configs::tab_config::shell_quoted_generated_worktree_path_string(
```
moirahuang: i know this was the existing code but let's import at the top instead of importing inline

app/src/tab_configs/tab_config.rs
```diff
@@ -58,6 +58,23 @@ pub(crate) fn generated_worktree_path_string(repo_path: &Path, worktree_name: &s
...
+/// `worktree_name` is typically a handlebars placeholder that is resolved (and independently
+/// shell-quoted) at render time, so it must stay outside the quotes — adjacent shell quotes
+/// concatenate into a single argument.
+pub(crate) fn shell_quoted_generated_worktree_path_string(
```
moirahuang: a general comment, since we're strictly replacing `generated_worktree_path_string` with `shell_quoted_generated_worktree_path_string`, let's just directly update the logic in `generated_worktree_path_string` instead of having `shell_quoted_generated_worktree_path_string` call ``generated_worktree_path_string``

lonexreb:
Thanks @moirahuang — all three addressed in [`668c9183`](https://github.com/warpdotdev/warp/pull/13808/commits/668c9183):

1. Folded the quoting directly into `generated_worktree_path_string` and removed the wrapper — every generated-command caller wants the quoted form. Nice side effect: the `session_config.rs` worktree command builders (which still interpolated the unquoted path) now get the fix too.
2. Dropped the issue reference from the doc comment.
3. Moved the inline path in `user_config/mod.rs` to a top-of-file import.

All 35 worktree tests pass; fmt/clippy clean.

acarl005:
Thanks for submitting this fix. Closing this out as `shell_words::quote` is intended for "Unix shells" as it says in the docs. The solution needs to account for PowerShell. I'm reviving #11392 as this is the complete solution.

PR #13809, opened 2026-07-16
fix(macos): declare Reminders + full-access EventKit usage keys in Info.plist (#13558)

### LINKED ISSUE #13558: icalBuddy fails with "error: No calendars.": Info.plist is missing NSRemindersUsageDescription, so macOS silently denies the Reminders prompt (works in Terminal.app)
### Describe the bug

Running `icalBuddy` inside Warp fails with `error: No calendars.` even when Warp has been granted Calendar access in System Settings > Privacy & Security > Calendars. The exact same binary works fine in Terminal.app on the same machine.

This was previously reported in https://github.com/warpdotdev/Warp/issues/5639 and auto-closed as needs-info. I debugged it to the root cause, which comes down to missing keys in Warp's `Info.plist`, so this should be a small fix.

### Root cause

`icalBuddy` opens both the EventKit event store and the reminders store at startup, regardless of whether the query only concerns calendar events. For a CLI tool spawned from a terminal, macOS attributes both permission requests to the terminal app (the TCC "responsible process").

macOS only shows a consent prompt if the responsible app declares the matching usage description key in its `Info.plist`. If the key is missing, the request is denied silently: no prompt, and no deny entry in System Settings that the user could flip.

Warp declares the Calendar key but none of the Reminders keys (and also not the "full access" variants that macOS 14 introduced):

```
$ for key in NSCalendarsUsageDescription NSCalendarsFullAccessUsageDescription NSRemindersUsageDescription NSRemindersFullAccessUsageDescription; do printf '%s: ' "$key"; defaults read /Applications/Warp.app/Contents/Info.plist "$key" 2>/dev/null || echo "(missing)"; done
NSCalendarsUsageDescription: A program in Warp wants to use your calendar.
NSCalendarsFullAccessUsageDescription: (missing)
NSRemindersUsageDescription: (missing)
NSRemindersFullAccessUsageDescription: (missing)
```

So the calendar half of icalBuddy's startup succeeds (Warp can prompt and be granted), the reminders half is auto-denied without any prompt, and icalBuddy bails out with `error: No calendars.` before it ever queries events.

The TCC database confirms it. After running icalBuddy once in Terminal.app (which prompts twice: Calendar, then Reminders) and once in Warp (which can only prompt for Calendar):

```
$ sqlite3 "$HOME/Library/Application Support/com.apple.TCC/TCC.db" \
  "SELECT service, client, auth_value FROM access WHERE client IN ('com.apple.Terminal','dev.warp.Warp-Stable') AND service IN ('kTCCServiceCalendar','kTCCServiceReminders');"
kTCCServiceCalendar|com.apple.Terminal|2
kTCCServiceReminders|com.apple.Terminal|2
kTCCServiceCalendar|dev.warp.Warp-Stable|2
```

There is no `kTCCServiceReminders` row for Warp at all, not even a denied one, because the prompt could never be shown.

Proof that this is the only missing piece: manually inserting a `kTCCServiceReminders` grant for `dev.warp.Warp-Stable` into the user TCC database makes icalBuddy work perfectly inside Warp, with no other change.

### To reproduce

1. `brew install ical-buddy`
2. In Warp: `icalBuddy calendars` (grant the Calendar permission if prompted, or enable Warp under System Settings > Privacy & Security > Calendars)
3. Output: `error: No calendars.` Note that no Reminders prompt ever appears.
4. In Terminal.app: same command. Two permission prompts appear (Calendar, then Reminders), and the command lists all calendars.

### Expected behavior

Running icalBuddy (or any EventKit-based CLI) inside Warp triggers the Reminders consent prompt the same way Terminal.app does, and the tool works after the user grants access.

### Suggested fix

Add the missing usage description keys to Warp's `Info.plist`:

- `NSRemindersUsageDescription`
- `NSRemindersFullAccessUsageDescription`
- `NSCalendarsFullAccessUsageDescription` (for completeness; macOS 14 split calendar access into "add only" and "full access", and tools linked against newer SDKs request the full-access variant)

Terminal.app is the reference behavior here: Apple's own terminal handles this correctly out of the box, so any workflow built on local calendar access works there but silently breaks in Warp. (As an Apple platform app, Terminal.app does not need the plist keys itself; third-party terminals do.)

### Additional context

Why this matters beyond one CLI tool: icalBuddy is probably the easiest and fastest way for AI agents and scripts running inside Warp to read the local macOS calendar (everything Calendar.app syncs, including Exchange and CalDAV accounts that are not reachable via web APIs). With agents increasingly running inside terminals, "the terminal can read the user's calendar after consent" is becoming a real capability gap between Warp and Terminal.app.

Workaround for affected users until this is fixed (run from any app with Full Disk Access; reversible with a corresponding DELETE):

```
sqlite3 "$HOME/Library/Application Support/com.apple.TCC/TCC.db" \
  "INSERT OR REPLACE INTO access (service, client, client_type, auth_value, auth_reason, auth_version, csreq, indirect_object_identifier, flags, boot_uuid, last_reminded) \
   SELECT 'kTCCServiceReminders', client, client_type, auth_value, auth_reason, auth_version, csreq, indirect_object_identifier, flags, boot_uuid, 0 \
   FROM access WHERE client='dev.warp.Warp-Stable' AND service='kTCCServiceCalendar';"
```

### Environment

- Warp: v0.2026.07.01.09.21.01
- macOS: Tahoe 26.5 (also reported on Sequoia 15.2 in https://github.com/warpdotdev/Warp/issues/5639)
- Shell: zsh 5.9
- icalBuddy: 1.10.1 (Homebrew)
- Regression: no, longstanding behavior
- Does this block daily Warp use: no, but it silently breaks calendar-based workflows that work in Terminal.app

lonexreb:
Resolves #13558.

## Problem

Running `icalBuddy` (or any EventKit-based CLI) inside Warp fails with `error: No calendars.` even when Warp has Calendar access — while the exact same binary works in Terminal.app.

## Root cause

As diagnosed in the issue: `icalBuddy` opens both the EventKit **events** store and the **reminders** store at startup. macOS attributes both permission requests to the terminal app as the TCC responsible process, and only shows a consent prompt if the responsible app declares the matching usage-description key in its `Info.plist`. Missing key ⇒ silent denial: no prompt, and no entry in System Settings the user could flip.

Warp's `script/update_plist` declares `NSCalendarsUsageDescription` but:
- none of the **Reminders** keys, and
- neither of the **FullAccess** variants that macOS 14+ checks instead.

## Fix

Add the three missing keys alongside the existing permission descriptions, matching the established message style:

- `NSRemindersUsageDescription`
- `NSCalendarsFullAccessUsageDescription` (macOS 14+)
- `NSRemindersFullAccessUsageDescription` (macOS 14+)

## Validation

`bash -n script/update_plist` passes. The new `plutil -insert` invocations were run verbatim against a scratch plist on macOS 15 (Darwin 25.2.0):

```
$ plutil -lint Test.plist
Test.plist: OK
$ for key in NSCalendarsUsageDescription NSRemindersUsageDescription NSCalendarsFullAccessUsageDescription NSRemindersFullAccessUsageDescription; do
>   printf "%s = " "$key"; /usr/libexec/PlistBuddy -c "Print :$key" Test.plist
> done
NSCalendarsUsageDescription = A program in Warp wants to use your calendar.
NSRemindersUsageDescription = A program in Warp wants to use your reminders.
NSCalendarsFullAccessUsageDescription = A program in Warp wants full access to your calendar.
NSRemindersFullAccessUsageDescription = A program in Warp wants full access to your reminders.
```

The reporter's issue includes the TCC-level trace confirming the missing-key → silent-denial mechanism; end-to-end verification requires a signed rebuilt bundle granting the new prompts (happy to record that flow if desired).

### DIFF
--- a/script/update_plist
+++ b/script/update_plist
@@ -263,6 +263,11 @@ plutil -insert NSCameraUsageDescription -string "A program in Warp wants to use 
 plutil -insert NSMicrophoneUsageDescription -string "A program in Warp wants to use your microphone." "$WARP_PLIST_PATH"
 plutil -insert NSContactsUsageDescription -string "A program in Warp wants to use your contacts." "$WARP_PLIST_PATH"
 plutil -insert NSCalendarsUsageDescription -string "A program in Warp wants to use your calendar." "$WARP_PLIST_PATH"
+plutil -insert NSRemindersUsageDescription -string "A program in Warp wants to use your reminders." "$WARP_PLIST_PATH"
+# macOS 14+ checks the FullAccess variants instead; without them EventKit requests from
+# terminal subprocesses (e.g. icalBuddy) are silently denied with no consent prompt (#13558).
+plutil -insert NSCalendarsFullAccessUsageDescription -string "A program in Warp wants full access to your calendar." "$WARP_PLIST_PATH"
+plutil -insert NSRemindersFullAccessUsageDescription -string "A program in Warp wants full access to your reminders." "$WARP_PLIST_PATH"
 plutil -insert NSLocationUsageDescription -string "A program in Warp wants to use your location information." "$WARP_PLIST_PATH"
 plutil -insert NSPhotoLibraryUsageDescription -string "A program in Warp wants to use your photo library." "$WARP_PLIST_PATH"
 


lonexreb:
Closing — this was fixed upstream in #13608 (merged July 13, before this PR was opened), which adds the same three keys. Apologies for the noise.

Issue #8741, opened 2026-02-22
CLI hooks for Warp's built-in tools (markdown viewer, editor, split panes)


mrlerch:
[###]Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Warp has excellent built-in tools — the markdown viewer, the file editor, split pane management — but they're only accessible through GUI interactions (clicking file paths, keyboard shortcuts). There's no way to trigger them programmatically from the terminal.

A warp CLI tool (or extending the existing one) that exposes hooks for these internal tools would make Warp truly scriptable and open up powerful automation workflows.

Proposed CLI hooks:

Markdown viewer/editor:
•  warp open --viewer file.md — open a file in Warp's rendered markdown viewer panel
•  warp open --editor file.md — open a file in a visual markdown editor (WYSIWYG, no markup knowledge required)
•  warp open --viewer --split right --size 33% file.md — open viewer in a right split at a specific width

Split pane management:
•  warp split right — programmatically open a right split
•  warp split right --size 33% — open a split at a specific width ratio
•  warp split right --run "command" — open a split and run a command in it
•  warp split close — close the current split pane

File tools:
•  warp open file.txt — open any file in Warp's built-in viewer/editor
•  warp open --diff file1.txt file2.txt — open a diff view

Use cases:
•  Custom cheat / help aliases that open a personal reference sheet in the markdown viewer alongside work
•  CI/CD scripts that display formatted reports in Warp's viewer
•  Workflow automation: one command that opens a split pane with logs/docs while you work
•  Editing markdown files (READMEs, docs, notes) visually without knowing markdown syntax — especially valuable for users who aren't developers but use the terminal
•  AppleScript/Automator currently being the only (fragile) workaround for any of this

Why this matters:
Warp already has these tools built in — they're polished and they work great. Exposing them via CLI hooks would be low-friction for the Warp team while dramatically increasing what users and AI agents (Warp AI, Claude Code, etc.) can do. It would make Warp the first terminal that's fully scriptable as an IDE-like environment, not just a shell.

Also, if there is any way to make a clickable TOC inside a .md   maybe specifically in a Warp .md, that would be totally awesome :)

### Is your feature request related to a problem? Please describe.

[...]

PR #13810, opened 2026-07-16
docs(warpctrl): specify file open wait behavior

salmonumbrella:
## Description

Adds focused product and technical specs for `warpctrl file open PATH --wait`. The proposal waits for the exact file view opened or focused by the command while keeping the default command nonblocking. It adds no public view handles or separate wait action.

This PR is spec-only. Implementation would follow after maintainer approval.

## Linked Issue

Contributes to #8741

- [x] Screenshots are not applicable to this spec-only PR.

## Testing

- `git diff --check`
- `./script/format --check`
- Runtime testing is not applicable because this PR changes documentation only.

## Agent Mode

- [ ] Warp Agent Mode - This PR was created via Warp AI Agent Mode

### DIFF
--- a/specs/GH8741/tech.md
+++ b/specs/GH8741/tech.md
@@ -1,0 +1,285 @@
+# Tech Spec: Wait for a file view opened by `warpctrl`
+
+Issue: https://github.com/warpdotdev/Warp/issues/8741
+Product spec: `specs/GH8741/product.md`
+
+## Context
+
+Warp Control already exposes one typed file action:
+
+- `crates/warp_cli/src/local_control/mod.rs` defines
+  `FileCommand::Open(FileOpenArgs)`. `FileOpenArgs` contains the path, optional
+  line and column, `new_tab`, and the standard target selectors.
+- `crates/warp_cli/src/local_control/commands.rs::run_file_command` maps that
+  command to the existing `ActionKind::FileOpen` request.
+- `crates/local_control/src/protocol.rs::FileOpenParams` is the strict wire
+  parameter type. `crates/local_control/src/catalog.rs` declares `file.open`
+  with an acknowledgement result.
+- `crates/local_control/src/client.rs::send_request` performs one blocking HTTP
+  request and reads one response. It has no client-side request timeout, so the
+  existing transport can carry a long-lived wait without polling or streaming.
+- `app/src/local_control/mod.rs::handle_control_request` authenticates the
+  request, dispatches it to the main-thread `LocalControlBridge`, awaits the
+  bridge result, and serializes one response.
+- `app/src/local_control/bridge.rs::LocalControlBridge::handle_request` is
+  synchronous on the WarpUI model thread and currently returns a
+  `ResponseEnvelope` immediately.
+- `app/src/local_control/handlers/app_state.rs::file_open` validates the
+  parameters and target, calls `WorkspaceView::open_file_with_target`, and
+  immediately returns the acknowledgement.
+
+The selected in-Warp file surface depends on existing settings and state:
+
+- `app/src/util/openable_file_type.rs::resolve_file_target_to_open_in_warp`
+  guarantees that local-control `file.open` selects either Warp's code editor
+  or its Markdown viewer, even if the user's general editor setting points to
+  an external application.
+- `app/src/workspace/view.rs::open_file_notebook` focuses an existing Markdown
+  pane for the same path or creates a new `FilePane`.
+- `app/src/workspace/view.rs::open_code` and
+  `app/src/code/view.rs::CodeView::open_or_focus_existing` focus an existing
+  code-editor tab or create a new one.
+- A logical code-editor file is a `TabData` inside `CodeView`, not necessarily
+  the containing pane. A logical Markdown file view is a `FileNotebookView`
+  owned by `FilePane`.
+- `app/src/pane_group/pane/mod.rs::DetachType` distinguishes a real close
+  (`HiddenForClose` or `Closed`) from `Moved`. The undo-close path hides a pane
+  before destroying it.
+- Code-tab movement is not uniform. Reordering and most merges carry `TabData`,
+  while `CodeView::remove_tab_for_move` currently reconstructs a new
+  `CodePane` from its path. Waiting therefore cannot rely only on a view handle,
+  pane ID, tab index, or path.
+- `crates/warp_util/src/sync.rs::Condition` is a cloneable, set-once async
+  condition. It records completion before notifying, supports multiple and late
+  waiters, and does not require a per-waiter registry.
+
+## Proposed changes
+
+### 1. Extend the existing CLI and wire parameter
+
+In `crates/warp_cli/src/local_control/mod.rs`:
+
+- Add a Clap `--wait` boolean to `FileOpenArgs`.
+
+In `crates/local_control/src/protocol.rs`:
+
+- Add `wait: bool` to `FileOpenParams` with a Serde default.
+- Omit `wait` when false so requests from the default command retain their
+  current serialized shape.
+
+In `crates/warp_cli/src/local_control/commands.rs`:
+
+- Pass the parsed flag through the existing `ActionKind::FileOpen` request.
+- Keep `run_action_with_params`, instance selection, target selection, output
+  formatting, and the success payload unchanged.
+
+Do not add an `ActionKind`, parameter-spec variant, result-spec variant,
+endpoint, or protocol version. A new CLI using `--wait` against an older strict
+server will receive `InvalidParams` rather than silently behaving as
+nonblocking. Normal packaged usage invokes the channel-matched Warp binary, and
+an old CLI request without the field remains accepted by the new server.
+
+Update `resources/bundled/skills/warpctrl/SKILL.md` with a `file open --wait`
+example and the rule that Ctrl+C cancels the caller without closing the view.
+
+### 2. Add one private logical file-view lifecycle
+
+Create `app/src/file_view_lifecycle.rs` as the single owner of the private wait
+contract. It defines:
+
+- `FileViewLifecycle`, a cloneable wrapper around
+  `warp_util::sync::Condition`.
+- `FileViewLifecycle::close()`, an idempotent transition to closed.
+- `FileViewLifecycle::wait_until_closed()`, which completes immediately for a
+  late waiter or asynchronously for an open view.
+- `FileOpenReceipt`, which carries the lifecycle selected by an open/focus
+  operation back to local control.
+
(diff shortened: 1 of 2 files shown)


salmonumbrella:
@cla-bot check

specs/GH8741/tech.md
```diff
@@ -0,0 +1,135 @@
...
+- Document `file open --wait` and Ctrl+C behavior in
+  `resources/bundled/skills/warpctrl/SKILL.md`.
+
+Do not add an action, endpoint, result type, or protocol version. Default
+requests retain their wire shape. An older strict server rejects `wait: true`
+with `InvalidParams` instead of silently ignoring it.
+
+### 2. Represent one private logical lifetime
+
+Add `app/src/file_view_lifecycle.rs` with:
```
bnavetta: This should probably be part of the `pane_group` module, since essentially what it does is wait on pane backing views

specs/GH8741/tech.md
```diff
@@ -0,0 +1,135 @@
...
+- Existing non-control callers may ignore the receipt. A missing receipt is an
+  internal error for `file.open`, whose target resolver guarantees an in-Warp
+  code or Markdown view.
+
+Selection and receipt retrieval occur in the same model-thread update, avoiding
+an open-then-search race and path ambiguity.
+
+### 4. Carry code-tab lifecycles through ownership changes
+
+Add a `FileViewLifecycle` to `TabData`.
```
bnavetta: This seems like it should be scoped to the pane rather than the tab - a tab might have multiple code panes, or both a code pane and terminal pane

specs/GH8741/tech.md
```diff
@@ -0,0 +1,135 @@
...
+- Actual removal after unchanged close, Save, or Discard calls `close()`; Cancel
+  does not.
+- Detaching a containing `CodePane` as `HiddenForClose` or `Closed` closes all
+  tab lifecycles; `Moved` preserves them.
+- If merge deduplication removes the source tab in favor of an existing
+  same-path destination, close the source lifecycle.
+
+### 5. Apply the same contract to Markdown panes
+
+Add a `FileViewLifecycle` to `FileNotebookView` and expose it through
```
bnavetta: It might be easier to make this work for _any_ pane, by storing it in `PaneView`

specs/GH8741/tech.md
```diff
@@ -0,0 +1,135 @@
...
+
+- Reorders and transfers preserve it.
+- `remove_tab_for_move` extracts and passes it into the rebuilt pane instead of
+  using the close-oriented removal path.
+- Actual removal after unchanged close, Save, or Discard calls `close()`; Cancel
+  does not.
+- Detaching a containing `CodePane` as `HiddenForClose` or `Closed` closes all
+  tab lifecycles; `Moved` preserves them.
+- If merge deduplication removes the source tab in favor of an existing
+  same-path destination, close the source lifecycle.
```
bnavetta: Would it make more sense to use the existing destination's lifecycle?

specs/GH8741/tech.md
```diff
@@ -0,0 +1,135 @@
...
+
+### 6. Defer only waiting responses
+
+Let `LocalControlBridge::handle_request` return either:
+
+- `Ready(ResponseEnvelope)` for existing actions and non-waiting opens; or
+- a private deferred result containing the request ID, acknowledgement data,
+  and selected lifecycle.
+
+The bridge still validates and mutates UI state synchronously. The Axum request
```
bnavetta, line 130: Since the local control protocol runs over HTTP, I'm not sure keeping the request open until the pane is closed will work. I don't think the client or server set explicit per-request timeouts, but we'd need to confirm if there are defaults

bnavetta (changes requested):
A few things to follow up on, but generally this seems useful! Could you remove the `Closes https://github.com/warpdotdev/warp/issues/8741` tag though? This contributes, but there's some other functionality in that issue which we should still track.

salmonumbrella:
Thanks — this makes the design better. I'll reword the tag to `Contributes to #8741` so the rest of the issue stays open.

**Module placement.** Agreed, it belongs in `pane_group` next to `DetachType` (`app/src/pane_group/pane/mod.rs:541`), which is the signal it consumes.

**Pane vs. tab.** My spec was ambiguous — there are two `TabData` types: the workspace tab (`app/src/tab.rs:170`, holds a `PaneGroup`, so one tab → many panes) and the editor tab (`app/src/code/view.rs:207`, one file inside a `CodePane`). Section 4 meant the second, so the code-pane-plus-terminal-pane case doesn't apply. I'll qualify every mention.

That does cut against pane-scoping, though: `CodeView` holds `tab_group: Vec<TabData>` (`app/src/code/view.rs:235`), so with `prefer_tabbed_editor_view` on, one `CodePane` has many files open. A pane-only lifecycle means `warpctrl file open --wait a.rs` wouldn't return when the user closes the `a.rs` tab — it'd block until the whole pane went away.

**`PaneView`.** I'd take this as the base layer. It's a generic struct (`app/src/pane_group/pane/view/mod.rs:84`), so a lifecycle field covers all 19 `BackingView` impls for free and makes "wait on any pane" fall out later. For `FilePane` it's exactly right — one pane, one file — so the Markdown section collapses into it and I can delete it. Revised model:

1. **`PaneView` owns a lifecycle** — closes on `detach` for `HiddenForClose`/`Closed`, preserved on `Moved`. Covers every pane kind; whole story for Markdown.
2. **Editor `TabData` owns one too**, closing when the tab is removed or its pane's lifecycle closes.

Level 2 exists only because `CodeView` multiplexes. If you'd rather ship level 1 and accept pane granularity for grouped tabs, that's smaller and I'll do it — I just don't think it satisfies the use case with tabbed editor view on.

Either way, `PaneGroup` stores `Box<dyn AnyPaneContent>` (`app/src/pane_group/mod.rs:893`), so generic access needs an accessor on the `PaneContent` trait. `detach` is already that kind of seam, but say if you'd rather not widen it.

**Merge dedup.** You're right — the file is still on screen, so completing the wait is wrong. This means it can't be a bare set-once `Condition`; source waiters have to be re-pointed at the survivor, so it needs an adopt/redirect layer. Sites are `CodeView::merge_tabs` (`app/src/code/view.rs:2089`) and the `DroppedOnTabBar`/`OverTab` path (`app/src/workspace/view.rs:16557`). I'll rewrite product invariant 4 accordingly. Separately, `remove_tab_for_move` (`app/src/code/view.rs:1194`) rebuilds the editor from a `CodeSource::Link` rather than carrying the handle, so identity needs threading through there regardless.

**HTTP.** You're right that it breaks, though not where I'd have guessed. The server is fine: bare Axum, no tower layers (`app/src/local_control/mod.rs:252`), and hyper's 30s `h1_header_read_timeout` default is inert because `axum::serve` never installs a timer — and it bounds the request head, not handler duration. It'll hold a request open indefinitely.

The client is the problem:

```rust
// crates/local_control/src/client.rs:54
let client = reqwest::blocking::Client::new();
```

reqwest's *async* client defaults to no timeout, but the blocking wrapper overrides it to **30 seconds** (`reqwest-0.13.3/src/blocking/client.rs:1499-1504`), per-operation rather than as one deadline. So `--wait` would fail deterministically at 30s with `TransportUnavailable`. Fix is `.timeout(None)` scoped to the wait path, leaving the default for one-shot actions where it's a feature. The spec will say so explicitly, since "we don't set a timeout" currently reads as "there is no timeout" and that's false.

Two things I'd want your read on:

- `--wait` would be the first local control action to hold a connection open — all ~84 `ActionKind`s today are one-shot, no SSE/websocket anywhere. An idle loopback connection held for hours with no keepalive is untested; I'd verify sleep/wake on macOS before calling it done. If you'd rather not hold a connection, the alternative is long-poll or SSE with a wait token, but that means a second action/endpoint the current design avoids.
- Ctrl+C: no SIGINT handler in `warp_cli`, and `warpctrl` is the app binary re-invoked via `--warpctrl` (`app/src/lib.rs:701`) through a bundled wrapper at `Contents/Resources/bin/warpctrl` that isn't in the repo tree. If it `exec`s, default disposition gives invariant 8 for free; if it forks, signals hit the wrapper. Do you know which offhand?

If the two-level lifecycle and the client-timeout change sound right, I'll push the revised specs.

salmonumbrella:
@bnavetta hey is there any uipdate here?

bnavetta:
Hey sorry I missed this! Could you push spec updates in line with your earlier comment?

To answer your questions:

> Editor TabData owns one too

Ah, I confused this with the workspace-level tab data. Storing it per editor tab seems reasonable.

> --wait would be the first local control action to hold a connection open — all ~84 ActionKinds today are one-shot, no SSE/websocket anywhere. An idle loopback connection held for hours with no keepalive is untested; I'd verify sleep/wake on macOS before calling it done. If you'd rather not hold a connection, the alternative is long-poll or SSE with a wait token, but that means a second action/endpoint the current design avoids.

Yeah I'd test the simpler long-lived connection approach before considering something more complicated.

> Ctrl+C: no SIGINT handler in warp_cli, and warpctrl is the app binary re-invoked via --warpctrl (app/src/lib.rs:701) through a bundled wrapper at Contents/Resources/bin/warpctrl that isn't in the repo tree. If it execs, default disposition gives invariant 8 for free; if it forks, signals hit the wrapper. Do you know which offhand?

The wrapper is created by https://github.com/warpdotdev/warp/blob/master/script/macos/create_warpctrl_wrapper, so it does use `exec`

salmonumbrella:
Thanks! Pushed the revised specs in 33ceda6, matching the thread:

- **Two-level lifecycle, now in `pane_group`.** `PaneView` owns a `ViewLifecycle` that `detach` closes on `HiddenForClose`/`Closed` and preserves on `Moved` — that covers every `BackingView` including Markdown, so the separate Markdown section is gone. The editor `TabData` owns the second level for the `CodeView` multiplexing case, and every `TabData` mention is now qualified as the editor tab.
- **Merge dedup redirects instead of completing.** The lifecycle grew a `redirect(target)` so `merge_tabs` / `DroppedOnTabBar` re-point source waiters at the surviving tab. Product invariant 4 now says the wait transfers and completes when the survivor closes.
- **Client timeout.** The spec states the blocking client's 30s default explicitly and scopes `.timeout(None)` to waiting requests only.
- **Held connection.** Kept the single held request per your preference; manual validation now includes an idle wait across macOS sleep/wake, with long-poll/SSE noted as the fallback if that proves unreliable.
- **Ctrl+C.** With the wrapper confirmed to `exec` (thanks for the pointer), the spec records that default SIGINT disposition gives invariant 8 with no handler needed.

salmonumbrella:
@bnavetta can you please check?

salmonumbrella:
@bnavetta I made the changes, anything else you're looking for?

PR #13825, opened 2026-07-16
fix(completions): don't treat attached-value tokens as short-flag bundles (#9820)

### LINKED ISSUE #9820: Random highlight of cmake parameters
### Pre-submit Checks

### Describe the bug

Editing this [KeePassXC build command](https://github.com/keepassxreboot/keepassxc/wiki/Building-KeePassXC#build-options) displays some parameters in brown, others in white, with no obvious pattern.

### To reproduce

1. Edit this:

   cmake -DWITH_XC_YUBIKEY=ON -DWITH_XC_BROWSER=ON -DWITH_XC_NETWORKING=OFF -DWITH_XC_SSHAGENT=ON -DWITH_XC_FDOSECRETS=ON -DWITH_XC_KEESHARE=ON -DWITH_XC_UPDATECHECK=OFF -DWITH_TESTS=OFF -DWITH_GUI_TESTS=OFF -DCMAKE_BUILD_TYPE=Release ..

### Expected behavior

_No response_

Image

### Operating system (OS)

Linux

Ubuntu 26.04

### Shell Version

Irrelevant

### Current Warp version

Latest

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

lonexreb:
Resolves #9820.

## Problem

Typing a CMake command like `cmake -DWITH_TESTS=OFF -DWITH_DOCS=ON ...` colors some `-D...` definitions as options and leaves others plain — apparently at random.

## Root cause

Token descriptions split `--flag=value` tokens at `=`, so `-DWITH_TESTS=OFF` is described as the flag token `-DWITH_TESTS` plus the value `OFF`. The flag engines (v2 and legacy) then interpret any single-dash multi-character token as a *bundle* of short-hand flags (the `ssh -Xv` pattern) and produce a current-token exact match whenever the **last character** happens to be a known short flag. For CMake, `-DWITH_TESTS` ends in `S` — a known short option (`-S`) — so the whole token matched and got option coloring; `-DWITH_DOCS`… ends in `S` too, but e.g. `-DWITH_FOO` doesn't, so it stayed plain. Hence the "randomness": it depends on the final letter.

## Fix

A single-dash token is only treated as a short-flag bundle when **every** character is a known short flag of the command — mirrored in both `flag/v2.rs` and `flag/legacy.rs`. `-DWITH_TESTS` (W/I/T/H/_/… are not all cmake short flags) no longer exact-matches, so all `-D...=...` definitions render consistently. Genuine bundles like `ssh -Xv` (X and v both known) keep their current-token match, preserving Describe behavior.

## Tests

Two regression tests in `flag/v2_tests.rs`:
- `attached_value_token_is_not_treated_as_short_flag_bundle` — `-DWITH_TESTS` against a cmake-like signature (`-S`, `-D`, `-G`) must produce no exact `-DWITH_TESTS` match. **Verified this fails without the fix** (RED→GREEN).
- `genuine_short_flag_bundle_still_matches_current_token` — `-Xv` against an ssh-like signature keeps its match.

Full `warp_completer` suite: 119 passed / 0 failed. `cargo fmt` + `clippy --tests` clean.

### DIFF
--- a/crates/warp_completer/src/completer/describe.rs
+++ b/crates/warp_completer/src/completer/describe.rs
@@ -200,11 +200,18 @@ pub async fn describe_given_token<T: CompletionContext>(
                     // A flag token with an attached value (`-DWITH_TESTS` from
                     // `-DWITH_TESTS=OFF`) is not a bundle of short-hand flags. Skip
                     // bundle-style exact matches — Option suggestions whose display (e.g.
-                    // CMake's `-S`) differs from the token while their replacement equals
-                    // it — so such tokens aren't misdescribed/colored as that option (#9820).
+                    // CMake's `-S`) doesn't itself exact-match the token (the exact match
+                    // came from the bundle replacement) — so such tokens aren't
+                    // misdescribed/colored as that option (#9820). The display is compared
+                    // with the suggestion's own matcher so legitimately case-insensitive
+                    // options (`-force` for `-Force`) keep their description.
+                    let display_exactly_matches_token = matches!(
+                        matcher.get_match_type(trimmed_token_item, suggestion.display()),
+                        Some(Match::Exact { .. })
+                    );
                     if is_flag_with_attached_value
                         && matches!(suggestion.suggestion_type(), SuggestionType::Option(..))
-                        && suggestion.display() != trimmed_token_item
+                        && !display_exactly_matches_token
                     {
                         continue;
                     }
--- a/crates/warp_completer/src/completer/describe.rs
+++ b/crates/warp_completer/src/completer/describe.rs
@@ -102,8 +102,16 @@ pub async fn describe<T: CompletionContext>(
                 Some(token) => {
                     // For --flag=value tokens, split based on cursor position so hovering the flag
                     // part describes the flag and hovering the value part describes the value.
-                    let token = split_flag_eq_token_at_cursor(token, pos);
-                    describe_given_token(line, &command.span(), token, context).await
+                    let (token, is_flag_with_attached_value) =
+                        split_flag_eq_token_at_cursor(token, pos);
+                    describe_given_token(
+                        line,
+                        &command.span(),
+                        token,
+                        is_flag_with_attached_value,
+                        context,
+                    )
+                    .await
                 }
                 None => None,
             }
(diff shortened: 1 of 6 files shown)


lonexreb:
Both concerns addressed in [`98a851f8`](https://github.com/warpdotdev/warp/pull/13825/commits/98a851f8):

**`-DS=OFF` edge case** — correct catch. The `=` context is known at the two places that split `flag=value` tokens (`get_token_descriptions` for typed-token coloring, `describe` for hover) but was discarded before reaching the engines. It's now threaded through: `split_flag_eq_token_at_cursor` reports whether it returned the flag part of an attached-value token, `describe_given_token` takes that as a parameter, and its exact-match arm skips bundle-style matches (Option suggestions whose display differs from the token) when set — covering both the v2 and legacy engines at the surface where the bug manifests.

**Coverage** — new end-to-end describe regression test against the real cmake signature from the default registry: hovering the flag part of `-DWITH_TESTS=OFF` **and** `-DS=OFF` must describe neither as an option. Verified RED→GREEN: without the describe-level filter, `-DS` describes as *"Path to root directory of the CMake project to build"* (the `-S` option). The genuine-bundle path is guarded by the pre-existing `test_xray_describe_with_flags` (`git commit -am`). Both feature configs green: 158 passed (default) / 119 passed (v2).

On visual evidence: the behavior here is fully deterministic token-description output (asserted end-to-end above, including the exact before/after description text). Happy to add a screenshot of the corrected highlighting if a maintainer would like one.

/oz-review

lonexreb:
Addressed in [`f7b5feaf`](https://github.com/warpdotdev/warp/pull/13825/commits/f7b5feaf0129b560a1162c02979b0ac3ff3fc5c1): the filter now judges the display against the typed token **using the suggestion's own matcher** rather than string inequality — a display that exact-matches under its own case-sensitivity rules (e.g. `-encoding` for `-Encoding`, CaseInsensitive) is a genuine option match and keeps its description; only matches whose exactness came from the bundle *replacement* (display doesn't match the token at all, e.g. `-S` for `-DS`) are skipped.

New regression test `test_describe_attached_value_keeps_case_insensitive_option_match` asserts `Add-Content -encoding=UTF8` keeps `-Encoding`'s description — verified it fails under the previous string-inequality filter. Full suites green in both configs: 159 (default) / 119 (v2).

/oz-review

PR #13919, opened 2026-07-17
feat(markdown_parser): support HTML <br> hard line break tag

### LINKED ISSUE #13732: Markdown viewer: support `<br>` for explicit line breaks
### Pre-submit Checks

Warp's Markdown Viewer should convert raw `<br>` tags into an actual line break, both standalone mid-paragraph and inside GFM pipe-table cells (the most common place `<br>` shows up, since pipe-table syntax has no other way to express a multi-line cell).

**Expected behavior:** `<br>` produces a real line break wherever it appears, including inside table cells.

Screenshot showing the break element rendering as code

`<br>` renders as literal text in both positions — standalone mid-paragraph, and inside a GFM pipe-table cell — it is not converted to a line break in either case.

### Additional context

**Test case:**

```markdown
# `<br>` (standalone and inside a GFM pipe-table cell)

Should render: two lines separated by a real line break where the standalone `<br>` sits, and a table cell with wrapped multi-line content where the `<br>` sits inside a pipe-table cell. Currently: expect `<br>` to render as literal text in both positions — it is not converted to a line break in either the plain-paragraph case or inside GFM table cells.

Standalone, mid-paragraph:

First line of the paragraph.<br>
Second line, after a raw `<br>`.

Inside a GFM pipe-table cell:

| Feature | Notes                          |
| ------- | ------------------------------- |
| Export  | Supports CSV.<br>Also supports JSON. |
```

- [`markdown_parser.rs`](https://github.com/warpdotdev/warp/blob/master/crates/markdown_parser/src/markdown_parser.rs)'s `InlineToken` enum has no `LineBreak` variant, so `<br>` isn't recognized anywhere in the `.md` render path — including inside GFM table cells, which route through the same `parse_inline`/`parse_cell_content` machinery as regular paragraph text.
- There's already reuse precedent for exactly this conversion: the paste-path parser ([`html_parser.rs`](https://github.com/warpdotdev/warp/blob/13902088affaf958188f99d7357cfeb7379463c4/crates/markdown_parser/src/html_parser.rs#L335)) maps `<br>` → `FormattedTextLine::LineBreak` when converting pasted HTML. Wiring the same mapping into the `.md` parser's inline tokenizer looks like the natural approach rather than inventing a new mechanism.
- CommonMark's own hard-break syntax (trailing double-space or backslash before a newline) also looks unhandled in the inline tokenizer — flagging it here since it's adjacent, though it's not the primary ask of this issue.

**Related:** this splits out one tag from #13652 (bulk raw-HTML-subset request). Sibling issues from the same split: #13721 (`` sizing), #13725 (anchor links), #13726 (raw HTML tables), #10259 (`<details>`/`<summary>` — pre-existing request), #13733 (`<kbd>`), #13734 (`<sub>`/`<sup>`), #13735 (`align` on `<p>`/`<div>`), #13736 (`<picture>`/`<source>`)

### Operating system (OS)

macOS

4

_No response_

Jason-Shen2:
Closes #13732

## Summary

This PR adds support for the HTML `<br>`, `<br/>`, and `<br />` tags (case-insensitive) as inline hard line breaks in Markdown parsing, fixing #13732.

## Changes

- Added `InlineToken::HardLineBreak` variant to represent `<br>` tags
- Added `parse_inline_token_br()` parser that matches `<br>`, `<br/>`, `<br />` case-insensitively
- Hard line breaks insert a `\n` character into fragment text, which `TuiText` natively splits into multiple visual rows
- Added `strip_trailing_hard_breaks()` to prevent spurious blank lines at the end of paragraphs from trailing `<br>` tags
- Tested with all cases: plain text, inline formatting, emphasis, lists, and edge cases

### DIFF
--- a/crates/markdown_parser/src/lib.rs
+++ b/crates/markdown_parser/src/lib.rs
@@ -285,11 +285,11 @@ impl LineCount for FormattedTextLine {
     fn num_lines(&self) -> usize {
         match self {
             Self::CodeBlock(text) => text.code.matches('\n').count(),
-            Self::Heading(_) => 1,
-            Self::Line(_) => 1,
-            Self::OrderedList(_) => 1,
-            Self::UnorderedList(_) => 1,
-            Self::TaskList(_) => 1,
+            Self::Heading(header) => count_inline_lines(&header.text),
+            Self::Line(texts) => count_inline_lines(texts),
+            Self::OrderedList(list) => count_inline_lines(&list.indented_text.text),
+            Self::UnorderedList(list) => count_inline_lines(&list.text),
+            Self::TaskList(list) => count_inline_lines(&list.text),
             Self::LineBreak => 0,
             Self::HorizontalRule => 0,
             Self::Embedded(_) => 1,
@@ -297,6 +297,13 @@ impl LineCount for FormattedTextLine {
             Self::Table(table) => 1 + table.rows.len(), // Header + data rows (separator not counted as a line)
         }
     }
+}
+
+fn count_inline_lines(fragments: &[FormattedTextFragment]) -> usize {
+    1 + fragments
+        .iter()
+        .map(|f| f.text.matches('\n').count())
+        .sum::<usize>()
 }
 
 #[derive(Debug, Clone, Eq, PartialEq)]
--- a/crates/markdown_parser/src/markdown_parser.rs
+++ b/crates/markdown_parser/src/markdown_parser.rs
@@ -1063,16 +1063,42 @@ fn parse_inline<'a, E: ContextError<&'a str> + ParseError<&'a str>>(
 /// This is a no-op if the last fragment does not end with a newline from a
 /// hard line break.
 fn strip_trailing_hard_break(fragments: &mut Vec<FormattedTextFragment>) {
-    let Some(last) = fragments.last_mut() else {
-        return;
-    };
-    let trimmed_ws = last
-        .text
-        .trim_end_matches(|c: char| c.is_whitespace() && c != '\n');
-    if !trimmed_ws.ends_with('\n') {
+    let mut trailing_ws_end = fragments.len();
+    for i in (0..fragments.len()).rev() {
+        let frag = &fragments[i];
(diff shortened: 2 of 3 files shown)


Jason-Shen2:
Fixed the `<br>` hard-line-break bug at [markdown_parser.rs:1038](https://github.com/warpdotdev/warp/pull/13919#discussion_r3609328316) in commit 001a8be99.

**Root cause:** The `ends_with_hard_break` boolean was being reset to `false` by non-rendered tokens (`Delimiter`, `LinkEnd`, `UnderlineEnd`) after a `<br>` `HardLineBreak` was emitted, which meant the trailing newline from an end-of-line `<br>` would not be stripped when followed by a closing markup tag like `</u>`.

**Fix:** Removed the fragile `ends_with_hard_break` flag entirely. Instead, `strip_trailing_hard_break` is now always called unconditionally after `consolidate_fragments`, and the function itself is a safe no-op when there is no trailing newline (returns early if `!trimmed_ws.ends_with("\n")`, rather than trimming trailing whitespace). Added a regression test for `<u>line<br></u>` → `underline("line")` and `<u>line<br><br></u>` → `underline("line\n")`.

All 146 markdown_parser tests pass; `cargo clippy -p markdown_parser -- -D warnings` is clean; `cargo fmt` applied.

Jason-Shen2:
Fixed in commit 7b9b24b53:

1. **Delimiter flanking after `<br>`**: Fixed the preceding_char calculation for delimiters to skip trailing newlines from `<br>` tags. Previously, `**line<br>**` was not parsed correctly because the hard newline from `<br>` was treated as whitespace, preventing the closing `**` from being recognized as a right-flanking delimiter. Added regression tests for bold (`**`), italic (`*`), and strikethrough (`~~`) with `<br>` before closing markup.

2. **Line counting**: Verified that `count_inline_lines` correctly counts embedded newlines from `<br>` for all inline line types (Line, Heading, OrderedList, UnorderedList, TaskList). Added an explicit test for this.

All 147 markdown_parser tests pass.

/oz-review

PR #13925, opened 2026-07-18
docs(warpctrl): specify file open cold start

### LINKED ISSUE #13630: Add a `warp <path>` desktop launcher for directories and files
### Pre-submit Checks

- [x] I searched Warp feature requests and did not find an issue for this focused desktop-launcher workflow.
- [x] I searched the Warp documentation and did not find a supported `warp <path>` command.

Add a small, editor-style desktop launcher command for opening local paths in Warp:

```shell
warp .
warp AGENTS.md
```

Expected behavior:

1. `warp <directory>` launches or focuses the stable Warp desktop app and opens a terminal tab whose working directory is that directory.
2. `warp <file>` launches or focuses Warp and opens the file in Warp's built-in editor/Markdown viewer.
3. Relative paths are resolved from the caller's current working directory.
4. `warp` with no path launches or focuses the desktop app.
5. The Preview channel can use a channel-specific name such as `warp-preview`, consistent with `oz-preview` and `warpctrl-preview`.

This should be a thin desktop-launcher surface, not a replacement for either existing CLI:

- `oz` remains the agent CLI.
- `warpctrl` remains the richer API for controlling a running Warp instance.

The underlying app capabilities already exist:

- directories can be opened with `warp://action/new_tab?path=...` or the OS file-open event;
- files can be opened with `warp://action/open_file_editor?path=...` (implemented by #10233 for #9561);
- `warpctrl file open <path>` can open a file when a controllable app instance is already running.

The requested command would make those capabilities easy to discover and use from shells and IDE integrated terminals, while also handling app startup and channel routing.

Developer tools commonly expose a memorable path launcher such as `code .`. Warp currently requires platform-specific commands or a user-maintained wrapper:

```shell
open -a Warp .
open "warp://action/open_file_editor?path=/absolute/path/to/AGENTS.md"
```

That is harder to remember, differs by operating system, and requires callers to choose different mechanisms for files and directories. The CLI portion of #9561 was deliberately left out when its URI support was implemented, and a later comment on that issue specifically requested `warp .` from an IDE integrated terminal. This issue tracks that narrower follow-up rather than reopening the file URI design.

Related but distinct:

- #9561 / #10233: external file opening via URI, now complete
- #11545: broad `warpctrl` scripting and live-instance control
- #2933: duplicate-window bug in the current macOS `open -a Warp .` workaround

### Additional context

Locally verified on Warp `0.2026.07.08.17.54.02` for macOS. A shell wrapper around the existing OS open events works for both examples, which demonstrates the user flow, but shipping the launcher with Warp would make it portable, documented, channel-aware, and supportable.

### Operating system (OS)

Cross-platform

4

psh4607:
## Description

Adds product and technical specs for the narrowed first step of #13630: allow `warpctrl file open <path>` to start the matching Warp app when no controllable same-channel instance is running.

The specs define:

- auto-start only for `file open`, without adding a top-level `warp <path>` command or new directory behavior
- caller-relative paths resolved before discovery so warm and cold app flows target the same file
- startup of the exact app bundle/package that supplied the wrapper, followed by the existing authenticated `file.open` request
- unchanged behavior for explicit `--instance` / `--pid` selectors and every other `warpctrl` command
- bounded readiness polling, concurrent-start coordination, and fail-closed Scripting behavior

This is a spec-only PR and does not change runtime behavior.

## Linked Issue

Closes #13630

- [x] Where appropriate, screenshots or a short video of the implementation are included below (not applicable to this spec-only change).

## Testing

- `./script/format`
- No automated or manual app tests were added because this PR changes documentation only.

- [ ] I have manually tested my changes locally with `./script/run` (not applicable to this spec-only change)

### Screenshots / Videos

Not applicable to this spec-only change.

## Agent Mode

CHANGELOG-NONE


psh4607:
Hi @bnavetta — following up on this spec PR after Oz approved it and requested your review. Would you mind taking a look when you have a chance? I’ll hold off on implementation until the spec is approved. Thanks!

specs/GH13630/tech.md
```diff
@@ -0,0 +1,168 @@
...
+
+The launcher uses this intent only to ask the OS or channel executable to start/reuse the matching app:
+
+- On macOS, invoke `/usr/bin/open -a "$WARPCTRL_APP_BUNDLE" <channel-scheme>://action/warpctrl_startup`. Supplying the exact bundle path prevents Launch Services from selecting another Stable, Preview, Dev, or Local installation that registered the same URL scheme. Do not fall back to opening the bare URL when the bundle contract is unavailable or invalid.
+- On Linux/FreeBSD package builds, spawn `WARPCTRL_APP_EXECUTABLE` with the startup URI as its only app argument. Existing startup forwarding treats the non-empty URI list as an app-open event rather than synthesizing a new-window request.
+- Windows remains unsupported until Warp Control's authenticated discovery and broker requirements are implemented.
+
+The startup intent matters in the race where Warp is already running but Scripting is disabled or a record has not yet been published. It can be forwarded to that process without opening an extra window. A genuinely new app still performs normal restoration and opens a default window if restoration and URI handling leave it with none.
```
bnavetta: Could you elaborate on this? The idea is that if a Warp instance is launched with the CLI startup intent, but scripting is disabled, it will exit?

specs/GH13630/tech.md
```diff
@@ -0,0 +1,168 @@
...
+request matching app startup once and poll authenticated discovery every 100 ms
+record the generation's ready, launch-failed, or timed-out outcome
+release the lock and return the discovered records or recorded error
+```
+
+The generation snapshot comes before initial discovery so an attempt that completes between discovery and lock acquisition is still recognized as overlapping. The second discovery under the lock closes the common race in which another command publishes an instance between the first scan and launch. Put the lock and a separate small attempt-state record in the existing owner-only local-control discovery directory and use an OS-released advisory lock (`flock` on the currently supported POSIX platforms), following the existing remote-server startup-lock pattern. The state stores only a monotonically increasing completed generation, its outcome, and the shared deadline; it never stores file paths, selectors, credentials, or other request parameters. Write the completed state to a temporary owner-only file and atomically rename it into place before releasing the lock, so commands can take a consistent pre-lock snapshot without reading a partial update.
```
bnavetta: Above, it sounded like we’re relying on Launch Services (on macOS) and the Linux single-instance logic to ensure that only one instance of Warp is opened.

Given that, do we need this lock as well?

specs/GH13630/tech.md
```diff
@@ -0,0 +1,168 @@
...
+### 5. Specialize `run_file_command` without changing generic commands
+
+Refactor the generic request function just enough to allow `run_file_command` to obtain records from the cold-start helper before calling the existing selector/request/output path. Do not put auto-start inside `run_action_with_params`, because that would silently change all 84 actions.
+
+The `file open` sequence becomes:
+
+1. Resolve the supplied path from the CLI caller's current working directory as described above.
+2. Parse target selectors and construct `FileOpenParams` with the resolved path and the existing line, column, and new-tab values.
+3. Discover or cold-start only when there is no explicit `--instance`/`--pid`.
+4. Call the existing `select_instance` with the resulting records and original selector.
```
bnavetta, line 91: Assuming we may want to support auto-open on other commands, would it make sense for `select_instance` to own triggering auto-start? It could still be configurable per command based on the arguments to `select_instance`

bnavetta (commented):
Sorry for the delay, thanks for putting this together!

Generally seems like the right direction, but had some questions about how this ensures only one instance starts up

specs/GH13630/tech.md
psh4607: No—the launched app remains running. The startup intent itself is a no-op, and when Scripting is disabled Warp simply does not publish an actionable discovery record. Only `warpctrl` times out after 10 seconds and returns `no_instance`; it does not terminate the app.

I updated this paragraph in `9afa92fd` to make that lifecycle explicit.

specs/GH13630/tech.md
psh4607: Launch Services and the existing Linux single-instance routing help ensure that only one lasting app instance is created, but they do not coordinate concurrent `warpctrl` processes. Without the lock, multiple commands can all observe empty discovery and independently issue startup requests. The lock and generation state make those commands share one startup request, deadline, and outcome rather than relying on the app layer to deduplicate repeated launches.

I updated this section in `9afa92fd` to clarify that distinction.

specs/GH13630/tech.md
psh4607: That makes sense from a future extensibility perspective. My hesitation with putting startup directly in `select_instance` is that it currently lives in `local_control` as a pure selector over already-discovered records; auto-start would couple that shared layer to wrapper-specific launch validation, filesystem coordination, and readiness polling.

I updated the spec in `9afa92fd` to make the discovery-or-start helper reusable and explicitly opt-in, so other commands can adopt it later while `select_instance` remains focused on selection.

psh4607:
Hi @oss-maintainers — following up on this spec PR. I addressed @bnavetta’s three questions in 9afa92fd and replied to each review thread on August 4. Oz has approved the spec, and there hasn’t been any further feedback since then.  
Could someone confirm whether the spec is ready for implementation, or let me know if any further changes are needed? I’ll continue to hold off on implementation until the spec is approved.

PR #13937, opened 2026-07-18
fix(sessions-launch-config): persist tab groups in launch configurations (#13898)

### LINKED ISSUE #13898: Save and Launch Configuration Template doesn't store Tab groups, all tabs are flat and they are not grouped
### Pre-submit Checks

### Describe the bug

I created a tab group, then I saved it as a configuration template, but when I launched the template, it loaded all tabs flat and they didn't group in the group.

### To reproduce

I created a tab group, then I saved it as a configuration template, but when I launched the template, it loaded all tabs flat and they didn't group in the group.

### Expected behavior

Save and Launch Configuration Template store Tab groups

_No response_

### Operating system (OS)

macOS

macos, linux

### Shell Version

latest

### Current Warp version

latest

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

lonexreb:
Resolves #13898.

## Problem

Saving a window with grouped tabs via **Save and Launch Configuration Template** loses the groups — reopening the config restores all tabs flat.

## Root cause

As pinpointed in triage: `WindowSnapshot.tab_groups` / `TabSnapshot.group_id` model grouped-tab state and normal session restoration persists them, but the launch-config path drops the information at both ends — `WindowTemplate` / `TabTemplate` have no group metadata fields, and `open_launch_config_window` adds each tab individually with no point where `TabGroup` records are recreated or tabs reattached.

## Fix

**Save path** — `WindowTemplate` gains `tab_groups` (name / color / collapsed / pinned) and `TabTemplate` gains `group`, an index into that list. `From<WindowSnapshot>` maps each group's runtime id to its serialized index and stamps member tabs. Runtime `TabGroupId`s are per-session UUIDs and deliberately **not** serialized — fresh ids are minted on open.

**Open path** — `open_launch_config_window` recreates the groups before adding tabs (gated on `FeatureFlag::GroupedTabs`; `pinned` additionally on `PinnedTabs`), mirroring grouped-tab session restore in `configure_new_workspace`, then reattaches each tab by its template group index. Unresolvable indices (hand-edited YAML, flag off) are dropped, matching restore's dangling-reference handling.

**Compatibility** — both new fields are optional/defaulted in serde: launch configs written before tab groups existed parse unchanged, and configs without groups serialize without the new keys.

Example of the new YAML shape:

```yaml
windows:
  - tab_groups:
      - name: Backend
        color: blue
        collapsed: true
    tabs:
      - layout: { cwd: /repo/api }
        group: 0
      - layout: { cwd: /notes }   # ungrouped
```

## Tests

- `test_config_from_snapshot_maps_tab_groups` — two groups + grouped/ungrouped/grouped tabs: templates and per-tab indices map correctly.
- `test_launch_config_yaml_round_trips_tab_groups` — serialize → parse preserves groups and membership.
- `test_launch_config_yaml_without_groups_parses` — pre-existing YAML without groups still parses (empty groups, no membership).

All 14 launch-config tests pass; `cargo fmt` + `clippy --tests` clean.

### DIFF
--- a/app/src/launch_configs/launch_config.rs
+++ b/app/src/launch_configs/launch_config.rs
@@ -6,6 +6,7 @@ use crate::app_state::{
     AppState, LeafContents, PaneNodeSnapshot, SplitDirection as StateSplitDirection, TabSnapshot,
     WindowSnapshot,
 };
+use crate::tab::SelectedTabColor;
 use crate::themes::theme::AnsiColorIdentifier;
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


lonexreb:
Addressed in [`39bf6775`](https://github.com/warpdotdev/warp/pull/13937/commits/39bf677501ed76a01b2d42aeb327be6d91b514d2): correct catch — a config can open into a window that already has unpinned tabs, so restoring a pinned group there would break the pinned-region ordering invariant.

Launch configs deliberately don't persist pinned state for *tabs*; group pinned persistence was scope creep on my part. Dropped the `pinned` field from `TabGroupTemplate` entirely — recreated groups always open unpinned, which removes the invariant hazard by construction and matches the existing per-tab behavior. All 14 launch-config tests green.

On visual evidence: the save→YAML→open mapping is covered end-to-end by the three deterministic tests (group metadata, membership indices, round trip, legacy compat); happy to add a screen recording of the grouped-tab flow if a maintainer would like one.

/oz-review

PR #13942, opened 2026-07-18
fix: avoid rendering share modal without a session

### LINKED ISSUE #13757: Tried to render share modal without a model fires in a ~10-min render loop (up to 84/sec)
### Describe the bug

Warp logs `Tried to render share modal without a model` in a tight render loop. In `~/Library/Logs/warp.log` I see **9,362 occurrences in ~25 hours** of normal use, with bursts of **up to 84 in a single second** on a steady **~10-minute cadence** (e.g. 07:00, 07:52, 08:14, 08:24, 08:35, 09:10, 09:18, 09:29, 09:51, 10:02). The regular spacing indicates a periodic background timer triggering a repaint storm while the modal view has no model.

### To reproduce

1. Use Warp normally (I had TUI apps / agent sessions running in panes).
2. Inspect `~/Library/Logs/warp.log` for the line:
   ```text
   [WARN] Tried to render share modal without a model
   ```
3. Observe it repeating in bursts (~84/sec) roughly every ~10 minutes.

Reproduces on a **fresh boot / clean log** (not a function of long uptime), in stable `v0.2026.07.08.17.54.stable_02`, and the warn is still present on current `master`.

### Expected behavior

The `ShareBlockModal` view should not remain mounted and re-rendering once its `model` is `None`; at minimum it should not emit a `warn!` on every repaint.

### Root cause / additional context

`app/src/terminal/share_block_modal.rs`, in `render_modal()` (present unchanged on `master`):

```rust
let single_block = match &self.model {
    Some(model) => { /* ... */ }
    None => {
        log::warn!("Tried to render share modal without a model");
        Empty::new().finish()
    }
};
```

The `ShareBlockModal` view lives at the `PaneGroup` level and its `model` is `Option` because terminal models are session-specific. Nothing removes the view from the tree when the model is absent, so it re-renders `Empty` and warns on every repaint. I do **not** use Warp Drive or session sharing, yet the loop still fires.

Suggested fix: don't mount/retain the modal view when `model` is `None`, or short-circuit rendering without the per-frame `warn` (rate-limit / move to `debug!`).

Log signature (counts from one ~25h rotation): 9,362 hits; top per-second bursts all ~84. Happy to attach a redacted log excerpt on request.

### Operating system (OS)

macOS

macOS 15.7.7 (build 24G720); Apple M3 Pro / 36 GB

### Shell Version

zsh 5.9

### Current Warp version

v0.2026.07.08.17.54.stable_02

### Regression

No (but it causes noticeable UI stalls during each burst)

### Pre-submit Checks

- [x] I searched Warp bugs and there are no duplicates
- [x] I searched the Warp known issues page and this is not there

KRRT7:
## Description
Fixes a share-modal lifecycle bug where the pane group could mark the modal open before confirming the terminal session existed. That left a modal with no model mounted, causing repeated render-time warnings and unnecessary repaint/logging work.

## Linked Issue
- Fixes #13757

## Testing
- `cargo nextest run -p warp -E "test(share_block_modal)"`
- `cargo clippy -p warp --lib --tests -- -D warnings`
- `cargo fmt --all`
- `git diff --check`

This is a lifecycle/performance fix with no visual UI change, so screenshots are not applicable.

### Changelog
CHANGELOG-BUG-FIX: Avoid repeated share-modal rendering when no terminal session is available.


app/src/terminal/share_block_modal.rs
```diff
@@ -1023,7 +1023,9 @@ impl ShareBlockModal {
                 .finish()
             }
             None => {
-                log::warn!("Tried to render share modal without a model");
+                // The pane group normally prevents this state, but keep the
+                // fallback quiet if a stale close/open event races a render.
+                log::debug!("Tried to render share modal without a model");
```
seemeroland: Can we keep this as a warning? I don't believe it's expected to hit this, and reducing to debug will hide the issue

seemeroland (commented):
This change looks good, but I would be surprised if this resolves the problem. I suspect there's some other case we're rendering the modal when we shouldn't

KRRT7:
@seemeroland agreed, there's other issues as well (check my other PRs) but I started small since my PRs typically stall here.

seemeroland (approved):
Sorry for the delay, I was on vacation the past couple weeks. Thanks!

KRRT7:
thank you @seemeroland, I'll follow up with a deeper fix if you're good with it

PR #13944, opened 2026-07-18
fix: close stale classic completion menus

### LINKED ISSUE #13605: Classic completions remain stale after Tab then Backspace

KRRT7:
## Description

Classic Completions could leave a stale suggestion menu open after the user backspaced past the query that originally produced the results. The short prefix could then re-filter to the entire original result set.

The completion menu now remains valid while the original replacement prefix is intact, but closes when editing removes part of that prefix.

## Linked Issue

Fixes #13605

## Testing

- [x] `cargo fmt --all`
- [x] `cargo nextest run -p warp -E 'test(classic_completions)'`
- [x] `cargo clippy -p warp --lib --tests -- -D warnings`
- [x] `git diff --check`
- [x] Added regression coverage for Classic Completions after backspacing past the original query.

### Screenshots / Videos

Not included: this is a terminal input behavior change covered by automated tests; no GUI visual layout changed.

## Agent Mode

CHANGELOG-BUG-FIX: Close stale Classic Completions menus after editing past the original query.

### DIFF
--- a/app/src/terminal/input.rs
+++ b/app/src/terminal/input.rs
@@ -11661,8 +11661,18 @@ impl Input {
                     // If we're in classic completion mode and the selected item is equal
                     // to the current word, then we should keep the menu open; the user is cycling.
                     // We early-return because we don't want to filter the menu based on the
-                    // selected item.
+                    // selected item. This must happen before validating the original prefix so
+                    // fuzzy matches that do not start with the original query can still cycle.
                     return false;
+                }
+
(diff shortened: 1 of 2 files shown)


KRRT7:
https://github.com/user-attachments/assets/11f95925-6487-425c-8df5-662e3cd317c1

Issue #13734, opened 2026-07-14
Markdown viewer: support `<sub>`/`<sup>` subscript and superscript


fbartho:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3Aenhancement) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Warp's Markdown Viewer should render `<sub>` and `<sup>` with a vertical baseline offset (and typically reduced font size), matching the standard subscript/superscript treatment. Markdown has no native syntax for either, so raw HTML is the only way authors express them.

**Expected behavior:** `<sub>2</sub>` renders as a subscripted "2"; `<sup>1</sup>` renders as a superscripted "1", both visually offset from the surrounding baseline text.

### Is your feature request related to a problem? Please describe.

Subscript shows up in chemical formulas and variable notation; superscript shows up constantly for footnote markers and exponents. Both are common enough in technical docs that their absence is noticeable.

Concretely, today in Warp's Markdown Viewer (verified on this build):

<img width="816" height="1179" alt="Screenshot showing that sub/sup elements render as raw code" src="https://github.com/user-attachments/assets/398cb884-3c37-4ef9-b7d9-18b4eabf4802" />

The raw `<sub>`/`<sup>` tags and their contents render as plain literal text at baseline, with no vertical offset.

### Additional context

**Test case:**

```markdown
# `<sub>`/`<sup>`

Should render: a chemical formula with a properly subscripted number, and a footnote-style superscript marker. Currently: expect the raw tags and their contents to render as plain literal text at baseline, with no vertical offset.

Water is H<sub>2</sub>O, and carbon dioxide is CO<sub>2</sub>.

This claim needs a citation<sup>1</sup>.

<sup>1</sup> Source: internal benchmark, 2026.
```

**Code pointers.** As far as I can tell:

- Neither tag appears in `PHRASING_ELEMENT_TAGS` ([`crates/markdown_parser/src/html_parser.rs`](https://github.com/warpdotdev/warp/blob/master/crates/markdown_parser/src/html_parser.rs)), and the paste-path styling match arm no-ops on unknown tags — there's an adjacent TODO in that same match arm noting more phrasing styling is needed generally, so this gap looks known rather than deliberately excluded.
[...]

PR #13952, opened 2026-07-18
Strip all font weights when clearing all text styles (#13949)

### LINKED ISSUE #13949: Strip-all-styles path only clears Bold — other font weights survive conversion into no-formatting blocks
## Summary

When styled text is converted into a block that disallows formatting (e.g. a code block), the strip-all-styles path only removes the font weight it happens to have baked into its mask: `TextStyles::all()` ([`crates/editor/src/content/text.rs:1391-1402`](https://github.com/warpdotdev/warp/blob/1dcbfdf46/crates/editor/src/content/text.rs#L1391-L1402)) carries `Some(CustomWeight::Bold)`, but `all::<BufferTextStyle>()` yields 8 distinct `Weight(_)` payload variants. The per-variant gate in `unstyle_internal_editor_actions` ([`crates/editor/src/content/buffer.rs`](https://github.com/warpdotdev/warp/blob/1dcbfdf46/crates/editor/src/content/buffer.rs), reached from [`block_style_range` at :3153](https://github.com/warpdotdev/warp/blob/1dcbfdf46/crates/editor/src/content/buffer.rs#L3153)) checks `text_style.exact_match_style(&style)` — which is false for every non-Bold weight — so text styled with e.g. `CustomWeight::Black` keeps its weight marker inside the code block.

## Evidence

This is the same mechanism as the vertical-align variant of the bug just fixed in #13948 ([commit 6fb1a0b88](https://github.com/warpdotdev/warp/pull/13948/commits/6fb1a0b88)), where it was empirically reproduced: a superscript run converted into a code block came out as `<code:Shell><sup_s>te<sup_e>st<text>` instead of `<code:Shell>test<text>`. The weight case is the identical `exact_match_style`-vs-mask gate with a payload-carrying variant; it predates that PR (not caused by it, deliberately left out of its scope).

The sub/sup fix pattern (`has_any_vertical_align()` special-case in the strip loop) suggests the analogous shape here: strip-all should treat *any* `Weight(_)` marker as a match rather than only the mask's literal value.

## Reproduction sketch

Adapted from the verified regression test added for the vertical-align twin ([`test_block_style_strips_vertical_align`](https://github.com/warpdotdev/warp/pull/13948/commits/6fb1a0b88) in `crates/editor/src/content/buffer_tests.rs`) — same harness, weight instead of vertical-align:

<details><summary>Test case (adapted from the verified sub/sup twin)</summary>

```rust
#[test]
fn test_block_style_strips_nonbold_weight() {
    // Style a run with a non-Bold weight, then convert the range into a
    // code block (formatting disallowed). The strip-all mask carries only
    // CustomWeight::Bold, so exact_match_style() never matches Black and
    // the weight marker survives into the code block.
    let mut buffer = buffer_from_markdown("test");
    let mut weight_style = TextStyles::default();
    weight_style.weight = Some(CustomWeight::Black);
    buffer.style_range(0..2, weight_style); // "te" -> Black

    buffer.block_style_range(.., BufferBlockStyle::CodeBlock { language: shell() });

    // Expected: "<code:Shell>test<text>"
    // Actual (bug): "<code:Shell><w_black_s>te<w_black_e>st<text>"
    assert_eq!(buffer.to_internal_format(), "<code:Shell>test<text>");
}
```

</details>

## Expected

Converting a range into a no-formatting block clears **all** inline style markers regardless of their payload value — every `Weight(_)`, matching how #13948's fix now handles both vertical-align markers.

Related: #13948 / #13734 (vertical-align twin + fix [commit 6fb1a0b88](https://github.com/warpdotdev/warp/pull/13948/commits/6fb1a0b88)).

fbartho:
Closes #13949

`TextStyles::all()` can only carry one `CustomWeight` (it hardcodes `Some(CustomWeight::Bold)`), so the strip-all-styles path used when converting styled text into a block that disallows formatting (e.g. a code block) only unstyled text that was exactly Bold, leaving other weights (e.g. Black, Light) to keep rendering inside the disallowed block.

Adds `TextStyles::has_any_weight()` and special-cases `Weight(_)` in `unstyle_internal_editor_actions`'s gate, mirroring the `has_any_vertical_align` fix for Sub/Superscript in #13734/#13948 (same file, same location, additive match arm — shaped to merge cleanly alongside that fix if both land). Because a run's concrete weight can vary within the "strip all" mask, the inner per-run loop now resolves and tracks the actual `BufferTextStyle::Weight` present at each position rather than assuming it's Bold.

Includes regression tests for a non-Bold weight (Black), a second non-Bold weight (Light) to confirm the fix isn't Black-specific, and a Bold guard case.

<details>
<summary>Manual verification</summary>

This bug has no user-facing visual surface, so verification is by regression test rather than a screenshot.

The bug only manifests when a buffer run carries a non-Bold `CustomWeight` (e.g. `Black`, `Light`) and that run is converted into a formatting-disallowed block (a code block). On current `master` there is no user-reachable way to get a non-Bold weight into a buffer run: HTML/rich paste collapses `font-weight` to a boolean (`crates/markdown_parser/src/html_parser.rs:79-89` sets `self.bold = weight_value > 400`) and only ever emits `Some(CustomWeight::Bold)` or `None` (`html_parser.rs:463-468`); markdown and `**bold**` produce `Bold` only; and header `Semibold` is applied at render time (`crates/editor/src/render/model/mod.rs`), never baked into a `BufferTextStyle::Weight` run. The non-Bold weight can only be introduced through the internal buffer API (`style_internal` with `TextStyles::set_weight`), which is exactly what the regression tests do.

The regression tests in `crates/editor/src/content/buffer_tests.rs`:

- `test_block_style_strips_non_bold_weight` (Black): styles a run `Weight::Black`, asserts `content.debug()` is `<text><b_s>te<b_e>st`, converts the range to a code block, and asserts the result is `<code:Shell>test<text>` with a `TextStylesWithMetadata::default()` run — i.e. no surviving weight. Before the fix this run kept its `Black` weight inside the code block.
- `test_block_style_strips_light_weight` (Light): same flow with `Weight::Light`, confirming the fix isn't specific to `Black`.
- `test_block_style_strips_bold_weight` (Bold guard): the `Bold` case already passed before the fix (it's the one weight `TextStyles::all()` names), kept as a guard so the non-Bold cases can't silently regress to Bold-only behavior.

The fix (`crates/editor/src/content/buffer.rs`) replaces the exact-weight match with `has_any_weight()` for the strip-all mask and tracks each run's actual weight, so any active weight is cleared instead of only the one the mask happens to carry.

</details>


fbartho:
The current implementation was a pothole waiting for something to fall into. No consumer was currently misusing it, but it would have blown up at a future date. As a result, I wouldn't be able to show you an interesting screenshot. It would just look like one of the font weights was applied somewhere, and you'd have to know that it was supposed to be unstyled in a particular use case. I've added a "Manual Verification" section to the PR body that discusses the unit tests that I've added and what they're for.

fbartho:
Closing in favor of #14433.

Issue #13721, opened 2026-07-14
Markdown viewer: support `<img>` sizing attributes (width/height/align)


fbartho:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3Aenhancement) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Warp's Markdown Viewer should honor sizing attributes (`width`, `height`, `align`) on raw HTML `<img>` tags, the way GitHub's and GitLab's renderers do. Markdown's `![alt](src)` syntax has no way to express image size or alignment at all, so authors reach for raw `<img>` whenever they need control over how large an image renders — this is the one piece of raw HTML that shows up constantly in ordinary READMEs, not just edge cases.

**Expected behavior:** the `<img>` tag renders as an actual image sized per its `width`/`height`/`align` attributes, matching the same subset GitHub supports.

### Is your feature request related to a problem? Please describe.

I generate SVG dashboards locally (charts, funnels, etc.) and embed them in Markdown notes for later reference. Warp's Markdown-syntax image renderer ignores the SVG's own intrinsic dimensions and doesn't scale with the pane, so the HTML `width` attribute is the only author-controllable sizing signal available — and Warp currently drops it, leaving no way to control image size at all short of re-encoding the SVG.

Concretely, today in Warp's Markdown Viewer (verified on this build):

<img width="816" height="1179" alt="Screenshot of the img tag rendering as raw code" src="https://github.com/user-attachments/assets/b29de93f-a279-4370-900d-33642ce64860" />

The raw `<img width="50%">` tag renders as literal text (the URL gets auto-linked inside the tag), while the markdown-native `![alt](src)` image right below it renders fine — just with no sizing control.

### Additional context

**Test case:**

```markdown
# `<img>` sizing (width/height/align)

Should render: a 50%-width image, and beside it a same-size markdown image for contrast. Currently: the HTML block should render as literal text (or vanish, if the tag is stripped) while the markdown image below renders normally.

<img src="https://placehold.co/600x400/png" width="50%" alt="HTML image, 50% width" align="center">

Markdown-native equivalent (should render as an image, full size, no sizing control):

![Markdown image, no sizing control](https://placehold.co/600x400/png)
```

**Code pointers.** As far as I can tell:

[...]

PR #13961, opened 2026-07-19
Align <div>/<p> blocks in the Markdown viewer (#13735)

### LINKED ISSUE #13735: Markdown viewer: support the `align` attribute on `<p>`/`<div>` blocks (center/right/left)
### Pre-submit Checks

Warp's Markdown Viewer should honor the `align` attribute on `<p>` and `<div>` blocks — `center`, but also `right` and `left`. Centering is the standard way READMEs build a "hero" header (logo, title, badges), since markdown itself has no alignment syntax outside table columns. Right-alignment matters too: GitHub renders the popular pattern of anchoring a project image to the right edge of the README so it sits interleaved alongside the opening prose (via `align="right"` — on a wrapping block, or on the `` itself, which the `` sizing sibling issue covers).

**Expected behavior:** content inside a `<div align="…">` or `<p align="…">` block renders horizontally aligned per the attribute (center/right/left), with nested markdown continuing to render normally (as it already does today). (Minor fidelity note: GitHub's own rendering has a known quirk where `<div align="center">` can break inline flow while `<p align="center">` doesn't — matching the common-case behavior matters more than replicating that quirk.)

Centered header blocks at the top of a README (logo + title + badge row, often wrapped in a single `<div align="center">`) are extremely common, and Warp currently renders them left-aligned with the alignment attribute silently dropped.

Screenshot showing alignment attributes don't appear to be applied (seem dropped)

Both the `<div align="center">` and `<p align="center">` blocks render left-aligned, with `align` ignored entirely. The nested markdown heading inside the `<div>` still renders as a real heading (left-aligned), and the nested `` inside it renders as literal text — consistent with `` being unsupported on its own (tracked separately).

### Additional context

**Test case:**

```markdown
# Centered blocks (`<p align="center">` / `<div align="center">`)

Should render: a centered heading-style line and a centered image, both horizontally centered in the pane. Currently: expect both blocks to render left-aligned as plain literal text/markdown, with the `align` attribute ignored.

<div align="center">

# Centered Title

<img src="https://placehold.co/300x150/png" alt="Centered image">

</div>

<p align="center">A centered caption line.</p>
```

- [`markdown_parser.rs`](https://github.com/warpdotdev/warp/blob/master/crates/markdown_parser/src/markdown_parser.rs) has no `div`/`p`/`center` matching at all.
- The paste path doesn't preserve the attribute either: `div` is in `TOP_LEVEL_ELEMENT_TAGS_TO_SKIP` ([`crates/markdown_parser/src/html_parser.rs`](https://github.com/warpdotdev/warp/blob/master/crates/markdown_parser/src/html_parser.rs), unwrapped, attributes discarded), and `p` falls to the generic catch-all arm — `align=` is never read in either case.
- There's no alignment concept in `FormattedTextStyles`, so this needs a content-model addition, similar in shape to the existing `TableAlignment` field (used only for GFM table columns today) — that's the closest in-repo precedent for "alignment as part of the content model" I could find, alongside the generic (but currently unwired-to-markdown) [`crates/warpui_core/src/elements/gui/align.rs`](https://github.com/warpdotdev/warp/blob/master/crates/warpui_core/src/elements/gui/align.rs) widget primitive.

**Related:** this splits out one tag from #13652 (bulk raw-HTML-subset request). Sibling issues from the same split: #13721 (`` sizing), #13725 (anchor links), #13726 (raw HTML tables), #10259 (`<details>`/`<summary>` — pre-existing request), #13732 (`<br>`), #13733 (`<kbd>`), #13734 (`<sub>`/`<sup>`), #13736 (`<picture>`/`<source>`)

### Operating system (OS)

macOS

2

_No response_

fbartho:
Closes #13735

Adds horizontal alignment for `<div align>`/`<p align>` blocks in the Markdown viewer — shipping enabled, like the sibling raw-HTML features. This PR contains the approved spec (merged from #13906's branch, `specs/GH13735/align/`) plus the full implementation, per the spec→impl workflow; the spec PR #13906 is superseded by this one.

Recognizes `align="left|center|right"` and `style="text-align:…"` on own-line `<div>`/`<p>` blocks and single-line `<p>` captions, through one shared attribute-matching contract (case-insensitive names, quoting variants, style-wins conflict, a small text-align style subset, unrelated-attrs-ignored, malformed→literal, unrecognized-value→unaligned). Alignment is modeled as paired content-model markers (the ordered-list precedent from the #13345 review), orthogonal to block style, so "aligned + code block" composes and copy/export round-trips canonically. A feature flag was considered and intentionally not added: flags guard genuinely complex/risky subsystems (like GFM tables), and this is a bounded parser+paint change with deterministic fallbacks at every boundary.

Rendering: the GUI paints aligned regions via per-line horizontal offsets at paint time (the same lever tables use for cell alignment) — the offset reads the already-shaped line width and never touches the platform shaper's attribute round-trip. The TUI reads the same parser boundary stream and renders best-effort within terminal width with a left-aligned fallback.

Tested: parser (22 detector/contract cases), buffer lowering + round-trip, TUI arms, paint-offset math, and layout threading — markdown_parser 157, ipynb_parser 24, warp_editor 485, warp_tui 400; clippy `-D warnings` + fmt clean across all six touched crates.

Known limitations, deliberate: selection-highlight/caret rects inside aligned regions are not offset yet (visual-only in the read-only viewer; follow-up), and a partial re-layout starting mid-region degrades to unaligned until the next full rebuild (doesn't arise in the read-only viewer's whole-document rebuilds).

<details><summary>Verification document used for the screenshot</summary>

```markdown
# `align` blocks verification

Each labeled case below. (Captured during development with the then-flag forced on; the feature now ships enabled.)

HERO (issue's case — centered div, heading + literal img):

<div align="center">

## Project Hero

<img src="assets/logo.svg" alt="logo">

</div>

CAPTION (single-line centered p):

<p align="center">A centered caption under a figure.</p>

RIGHT-ALIGNED block:

<div align="right">

This paragraph should hug the right edge.

</div>

STYLE VARIANT (text-align via style):

<p style="text-align: center">Centered via style attribute.</p>

MALFORMED (broken quote — must stay literal):

<p align="center>broken attribute quoting stays literal</p>

CONTROL (no alignment):

Plain left-aligned paragraph for comparison.
```

</details>

Screenshot of basic block alignment

### DIFF
--- a/crates/editor/src/content/buffer.rs
+++ b/crates/editor/src/content/buffer.rs
@@ -10,7 +10,7 @@ use line_ending::LineEnding;
 use markdown_parser::{
     CodeBlockText, FormattedIndentTextInline, FormattedTable, FormattedTaskList, FormattedText,
     FormattedTextFragment, FormattedTextHeader, FormattedTextLine, FormattedTextStyles,
-    MarkdownParseOptions, OrderedFormattedIndentTextInline, parse_markdown_with_options,
+    OrderedFormattedIndentTextInline, parse_markdown, parse_markdown_with_gfm_tables,
 };
 use num_traits::SaturatingSub;
 use pathfinder_color::ColorU;
@@ -847,11 +847,12 @@ impl Buffer {
         selection_model: ModelHandle<BufferSelectionModel>,
         ctx: &mut ModelContext<Self>,
     ) -> Self {
-        let options = MarkdownParseOptions {
-            gfm_tables: warp_core::features::FeatureFlag::MarkdownTables.is_enabled(),
-            block_align: warp_core::features::FeatureFlag::MarkdownBlockAlign.is_enabled(),
+        let parse_fn = if warp_core::features::FeatureFlag::MarkdownTables.is_enabled() {
+            parse_markdown_with_gfm_tables
+        } else {
+            parse_markdown
         };
-        let parsed_formatted_text = match parse_markdown_with_options(markdown, options) {
+        let parsed_formatted_text = match parse_fn(markdown) {
             Ok(parsed) => parsed,
             Err(e) => {
                 safe_error! {
@@ -886,11 +887,8 @@ impl Buffer {
         selection_model: ModelHandle<BufferSelectionModel>,
         ctx: &mut ModelContext<Self>,
     ) -> Result<Self, ipynb_parser::IpynbError> {
-        let options = MarkdownParseOptions {
-            gfm_tables: warp_core::features::FeatureFlag::MarkdownTables.is_enabled(),
-            block_align: warp_core::features::FeatureFlag::MarkdownBlockAlign.is_enabled(),
-        };
-        let formatted_text = ipynb_parser::ipynb_to_formatted_text(ipynb, options)?;
+        let gfm_tables = warp_core::features::FeatureFlag::MarkdownTables.is_enabled();
+        let formatted_text = ipynb_parser::ipynb_to_formatted_text(ipynb, gfm_tables)?;
         Ok(Self::from_formatted_text(
             formatted_text,
             embedded_item_conversion,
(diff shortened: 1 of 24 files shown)


fbartho:
Pausing alignment until the img tag feature #13721 ships, as that was a core object I wanted to align.

Issue #7897, opened 2025-10-25
Preserve visual context when switching Markdown view modes + search/replace support


RaisenOps:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3AFEATURE) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Currently, when opening a .md file in Warp, it can be viewed in two modes:

- **Raw Markdown**
- **Rendered Markdown**

However, when working with large files, switching between these modes **causes the view to reset to the beginning of the document**. This disrupts the reading or editing flow and makes navigation difficult—especially when focusing on a specific section.

**Suggested Improvements**:

**Preserve scroll position**: When toggling between Markdown view modes, it would be ideal for the scroll position to remain consistent rather than resetting to the top. This would help maintain visual **context across both modes**, particularly in lengthy documents.

**Universal search and replace functionality**: It would be extremely useful to have a built-in search and/or replace feature when viewing any file—Markdown or otherwise. This should be accessible via a keyboard shortcut 
(e.g., Ctrl + Shift + F, **similar to the terminal search**), allowing users to quickly locate and jump to specific lines without manual scrolling. While this need is most apparent in Markdown workflows, the feature would benefit any long-form file opened in Warp.

These enhancements would significantly improve the user experience when working with large files, especially in technical editing, documentation, and review workflows.

### Operating system (OS)

Select an OS

### How important is this feature to you?

5 (Can't work without it!)

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

PR #13967, opened 2026-07-19
Preserve scroll position when toggling markdown Rendered/Raw views

### LINKED ISSUE #14062: Markdown viewer: preserve scroll position when toggling Rendered/Raw views
### Pre-submit Checks

When viewing a Markdown file, toggling between **Rendered** and **Raw** view modes should preserve the reading position instead of jumping back to the top of the document.

**Expected behavior:** scroll partway through a long Markdown file, toggle Rendered ⇄ Raw, and the new view opens at the equivalent position rather than at the top.

Today, toggling the view mode always resets the viewport to the top. The toggle replaces the entire pane (`FileNotebookView` ⇄ `CodeView`), and the new pane starts with a fresh scroll state, so any reading position is lost. On a long document this means re-scrolling from the top every time you switch modes to, e.g., check the raw source of a table or a link.

Because the two modes render **different documents** (rendered blocks vs. raw source lines), an exact line mapping isn't available; a proportional position (fraction of the scrollable range) is the practical target.

### Additional context

This is the scroll-preservation half of #7897, which bundles two separate asks (scroll preservation **and** find/replace in the Markdown viewer). Filing this as a focused, single-scope issue to match the repo's `Markdown viewer: …` issue convention and to cleanly link an implementation.

PR #13967 implements this: it captures the outgoing pane's vertical scroll as a fraction of its scrollable range and re-applies it to the incoming pane after layout. The find/replace half of #7897 is intentionally out of scope here and can stay tracked on #7897 (or its own focused issue).

slash9494:
## Description

Toggling a markdown file between **Rendered** and **Raw** view modes always reset the view to the top of the document, because the toggle replaces the entire pane (`FileNotebookView` ⇄ `CodeView`) and the new pane starts with a fresh scroll state.

This PR preserves the reading position across the toggle:

- The outgoing pane's vertical scroll is captured as a **fraction of its scrollable range** (`scroll_top / max_scroll`, clamped `0..=1`).
- The fraction rides on `PaneEvent::ReplaceWithCodePane` / `ReplaceWithFilePane` (as `Option<OrderedFloat<f32>>` so `PaneEvent` keeps `Eq`).
- The incoming pane re-applies it **after its content is laid out**:
  - Raw side: a new `ScrollPosition::Fraction` variant on the existing pending-scroll (`ScrollTrigger`) path.
  - Rendered side: a new `LayoutAction::ScrollToFraction`, stored as a pending fraction in `RenderState` and applied in `apply_element_update`, gated on the viewport having been laid out (`viewport.height() > 0`) and the reset content's buffer version. Applying immediately would read a stale (~0) content height on a freshly created pane and clamp to the top.

A fraction is used instead of a char-offset snapshot (`ScrollPositionSnapshot`) because the two modes render **different documents**, so a char offset cannot map between them. The landing position is therefore proportional rather than line-exact; documents whose blocks resize late (e.g. large tables) may land slightly off, which seems like an acceptable trade-off for the common case.

All pre-existing callers of the pane constructors pass `None` and keep the current scroll-to-top behavior.

## Linked Issue

Closes #14062

#14062 is the focused, single-scope issue for scroll preservation (following the repo's `Markdown viewer: …` issue convention). It was split out of #7897, which bundles two separate asks — scroll preservation (this PR) and find/replace in the Markdown viewer (still tracked on #7897).

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. — **Note:** #7897 currently has neither label; happy to go through the spec process first if the team prefers.

## Testing

- Added a unit test for the fraction ⇄ scroll-top math (`crates/editor/src/render/model/viewport_tests.rs::test_scroll_fraction`): midpoint mapping, out-of-range clamping at both ends, and content-fits-in-viewport (no scrollable range → 0).
- `cargo nextest`-compatible: full `warp_editor` suite passes locally; `cargo check -p warp` clean; `./script/format --check` clean.
- No UI integration test: the integration framework cannot simulate clicking the Rendered/Raw toggle (noted in `crates/integration/src/test/notebooks.rs`).

Manually verified end-to-end in a local build, both directions:
- Rendered → Raw: scrolled to ~40% in Rendered, toggled, Raw opens at the equivalent position.
- Raw → Rendered: scrolled to ~55% in Raw, toggled, Rendered opens at the equivalent position.
- Verified via debug logging that the applied `scroll_top` matches `fraction × (content_height − viewport_height)` exactly in both directions, and that panes opened without a toggle (file tree, session restore) still start at the top.

### Screenshots / Videos

**Before** (master build): Raw scrolled to the bottom of the document (§37–40 + table) → toggling to Rendered resets to the top:

| Raw, scrolled to bottom | → Rendered: reset to top (bug) |
|---|---|
| Raw scrolled to bottom | Rendered reset to top |

**After** (this branch): the same action lands at the equivalent position (§40 + table still in view):

| Raw, scrolled to bottom | → Rendered: position preserved |
|---|---|
| Raw scrolled to bottom | Rendered position preserved |

## Agent Mode

CHANGELOG-IMPROVEMENT: The markdown viewer now preserves your scroll position when switching between Rendered and Raw view modes.


slash9494:
Hi @bnavetta — gentle nudge on this one. Oz approved it (0 concerns) and requested your review on 07-22, and the linked issue #14062 is marked `ready-to-implement`, so I believe it's ready whenever you have a moment. Happy to rebase or make any changes if something's needed — just let me know. Thanks!

slash9494:
@oss-maintainers — flagging this one per the CONTRIBUTING escalation note.

Oz approved this PR on 07-22 with 0 concerns and requested human review from
@bnavetta, and the linked issue #14062 is marked `ready-to-implement`. It's been
about two weeks without a human review, so I wanted to raise it here rather than
keep pinging.

I've just merged the latest `master` in — the branch is up to date with no
conflicts, `cargo check` passes for both `warp_editor` and `warp`, and the
viewport tests pass.

Could someone on the team take a look, or route it to another reviewer if
@bnavetta is stretched thin right now? Happy to rebase or make any changes
needed. Thanks!

app/src/code/view.rs
```diff
@@ -781,6 +781,26 @@ impl CodeView {
...
+        let Some(tab) = self.tab_group.get(self.active_tab_index) else {
+            return;
+        };
+        tab.editor_view.update(ctx, |editor, ctx| {
+            editor.set_pending_scroll(ScrollPosition::Fraction(fraction), ctx);
+        });
+    }
+
+    /// The current vertical scroll fraction of the active tab's editor, in `0..=1`.
+    #[cfg(feature = "local_fs")]
```
bnavetta: Does this need to be gated on `local_fs`? The functionality itself isn’t FS-specific; if this is dead code without a filesystem then we can add a conditional `expect(dead_code)` attribute

bnavetta (commented):
Sorry for the delay! One small question, but the code otherwise looks good

app/src/code/view.rs
slash9494: Good catch — you're right that it isn't FS-specific.

Digging in, `code/view.rs` turned out to be a special case: `code/mod.rs` path-swaps it to `wasm.rs` for wasm, so `local_fs` is always enabled wherever this file compiles and the gate never actually applied. I dropped it rather than swapping the attribute.

`notebooks/file/mod.rs` does compile for wasm, so its two helpers use the conditional `expect(dead_code)` you suggested, matching `terminal/recorder.rs`.

While there I also narrowed `set_pending_scroll_fraction` to `pub(crate)` since `pane_group` is the only caller — that turned out to surface a real `dead_code` error on wasm that the `pub` had been masking, now covered by the same attribute.

Verified with `cargo clippy -p warp -- -D warnings` on both the native target and `wasm32-unknown-unknown`, plus the viewport tests. Latest `master` is merged in as well.

bnavetta (approved):

slash9494:
@bnavetta thanks for the review and for turning on auto-merge!

It's stuck on a CI hiccup rather than anything on the branch: `Verify
compilation with release flags (wasm)` hit the job's 15-minute cap at 15m17s
(the run had also sat in the queue for about five hours first), so it ended as
cancelled and `Check CI results` went red. Everything else on that run passed,
including the tests on all three platforms and Clippy for wasm.

Nothing has changed on the branch since your approval, so a re-run of the
workflow should be enough to let auto-merge through — could you kick it off
when you have a moment? Happy to push an empty commit instead if you'd rather
not re-run, just say the word.

Issue #13736, opened 2026-07-14
Markdown viewer: support `<picture>`/`<source>` for theme-aware images


fbartho:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3Aenhancement) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Warp's Markdown Viewer should render `<picture>`/`<source media="(prefers-color-scheme: …)">` blocks, swapping between the light- and dark-mode image variant based on the current theme, falling back to the `<img>` inside `<picture>` when no `<source>` matches. This is the standard pattern for README badges and logos that need to stay legible in both light and dark rendering contexts.

**Expected behavior:** renders a single image that swaps between the light-mode and dark-mode `<source>` asset based on the active theme, falling back to the `<img>` if no source matches.

### Is your feature request related to a problem? Please describe.

Any README with a logo or badge that needs to look right against both a light and dark background uses this pattern — it's effectively the only portable way to do theme-aware images in Markdown/HTML.

Concretely, today in Warp's Markdown Viewer (verified on this build):

<img width="816" height="1179" alt="Screenshot demonstrating picture/source rendering as raw code" src="https://github.com/user-attachments/assets/426ea28a-bfa1-4ba8-ba50-e28b36bbc8de" />

The whole `<picture>` block, including the fallback `<img>`, renders as literal text — the fallback img is not implemented yet, see: #13721 

### Additional context

**Test case:**

```markdown
# `<picture>`/`<source>` (light/dark variants)

Should render: a single image that swaps automatically between a light-mode and dark-mode asset based on the current theme. Currently: expect the raw tags to render as literal text, or — if the fallback `<img>` is what survives — a single static image with no theme-reactive swapping.

<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://placehold.co/400x200/111827/FFFFFF/png?text=Dark+Mode">
  <source media="(prefers-color-scheme: light)" srcset="https://placehold.co/400x200/FFFFFF/111827/png?text=Light+Mode">
  <img src="https://placehold.co/400x200/FFFFFF/111827/png?text=Fallback" alt="Theme-aware logo">
</picture>
```

**Code pointers.** As far as I can tell, this needs both parser work and image-model work, not just a new tag handler:

[...]

Issue #13733, opened 2026-07-14
Markdown viewer: support `<kbd>` keyboard-shortcut styling


fbartho:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3Aenhancement) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

Warp's Markdown Viewer should render `<kbd>` tags with a small, monospace, keycap-styled badge — the standard treatment GitHub and most other renderers give it for documenting keyboard shortcuts.

**Expected behavior:** each `<kbd>…</kbd>` renders as a small keycap-styled badge (typically monospace font, subtle border/background, slightly raised appearance).

### Is your feature request related to a problem? Please describe.

`<kbd>` is the conventional way to document keyboard shortcuts in Markdown docs (there's no markdown-native equivalent), and it feels especially on-brand for a terminal app's own docs and READMEs to render it properly.

Concretely, today in Warp's Markdown Viewer (verified on this build):

<img width="816" height="1179" alt="Screenshot shows the kbd element rendering as raw code" src="https://github.com/user-attachments/assets/85d00cf5-e989-49d3-9cb3-5ad107621720" />

The raw `<kbd>` tags and their shortcut text render as plain literal text, with no keycap styling.

### Additional context

**Test case:**

```markdown
# `<kbd>`

Should render: a small, monospace, keycap-styled badge for a keyboard shortcut. Currently: expect the raw tags and shortcut text to render as plain literal text, with no keycap styling.

Press <kbd>Cmd</kbd>+<kbd>K</kbd> to open the command palette.
```

[...]

PR #13971, opened 2026-07-19
docs(specs): markdown viewer raw-HTML <img> support (rendering + sizing) (GH13721)

### LINKED ISSUE #13721: Markdown viewer: support `<img>` sizing attributes (width/height/align)

fbartho:
## What

Spec (product.md + tech.md, per CONTRIBUTING's spec-first workflow for feature requests) for teaching the Markdown viewer to recognize a block-level raw-HTML `` tag and honor its `width` / `height` / `align` attributes.

Closes #13721 — the focused ``-sizing issue split out of #13652 per maintainer guidance there. The other raw-HTML tags are tracked in their own sibling issues (#13725, #13726, #13732, #13733, #13734, #13735, #13736).

## Why

Markdown image syntax gives authors no sizing control, and the viewer's default layout ignores an image's intrinsic dimensions — so locally generated SVG charts embedded in Markdown notes render at a small fixed size with no author-side remedy. The HTML `width` attribute is the only sizing signal an author can express; today it is silently dropped.

## Highlights

- `` on its own line renders through the existing Markdown-image pipeline (`markdown_parser` → `BufferBlockItem::Image` → `RenderableImage`), following the field-threading precedent from `specs/GH849/`.
- Sizing: absolute px (`width="640"`/`"640px"`) and percentage (`width="90%"`), clamped to pane width; aspect ratio preserved when one dimension is given; `align="left|center|right"`.
- Sanitized by construction: only `src`/`alt`/`title`/`width`/`height`/`align` are read; `onerror`, `onload`, `style`, and all other attributes are ignored; no script execution or navigation. Malformed tags fall back to today's literal-text rendering — never a panic or a blank box.
- No behavior change for documents without raw `` tags; no change to default sizing when no attributes are present.

## Process note

#13652 was closed in favor of per-feature issues; this spec now targets #13721 and the files moved to `specs/GH13721/`. Keeping this draft open and updated while #13721 awaits triage / `ready-to-spec` — happy to adjust if you'd rather sequence it differently.

### DIFF
--- a/specs/GH13721/product.md
+++ b/specs/GH13721/product.md
@@ -5,12 +5,17 @@ Issue: https://github.com/warpdotdev/warp/issues/13721
 ## Summary
 
 Warp's Markdown Viewer renders Markdown image syntax (`![alt](src)`) but silently
-drops raw inline HTML, including the `<img>` tag. This is the single highest-priority
-tag in the issue's prioritized list, and the one that directly motivated the request:
-authors who embed locally-generated SVG dashboards in Markdown notes have no way to
-control the rendered size, because the Markdown-image renderer draws every image at a
-fixed default size that ignores the image's intrinsic dimensions, and the HTML `width`
-attribute — the only author-controllable sizing signal — is dropped entirely.
+drops raw inline HTML, so a raw-HTML `<img>` tag does not load or render **at all** —
+it disappears into literal, unrendered text. This is the single highest-priority tag in
+the issue's prioritized list, and the one that directly motivated the request. Making
+the tag render is the core fix; honoring its presentational attributes is the rest.
+Authors reach for HTML `<img>` over `![alt](src)` precisely *because* of those
+attributes — chiefly sizing: the Markdown-image renderer draws every image at a fixed
+default size that ignores the image's intrinsic dimensions, and the HTML `width`
+attribute — the only author-controllable sizing signal — is dropped along with the rest
+of the tag. So the motivating example (embedding a locally-generated SVG dashboard at a
+controlled width) needs both halves: the tag must render, and its `width`/`height`/`align`
+must be honored.
 
 This spec covers a narrow, self-contained slice split out of the original bulk
 raw-HTML-subset request (#13652): teach the Markdown viewer to recognize a
@@ -47,6 +52,12 @@ Out of scope (not changed by this spec):
 - All other HTML tags from the issue's list: `<a href/id>`, `<table>`/`<tr>`/`<td>`,
   `<details>`/`<summary>`, `<br>`, `<kbd>`, `<sub>`/`<sup>`, `<p align>`/`<div align>`,
   `<picture>`/`<source>`.
+- Responsive-image source selection via `srcset` (density/width candidate lists such as
+  `1x`/`2x`/`480w`). This is a refinement deferred to the `<picture>`/`<source>` feature
+  (#13736), which owns responsive selection; on a bare `<img>`, `srcset` is ignored and
+  only `src` is read — matching the `src`-fallback behavior of a `srcset`-unaware
+  browser. This is a **feature deferral, not a security exclusion** (contrast the
+  `onerror`/`onload`/`style` drops in invariant 13, which are security-motivated).
 - Honoring the intrinsic `width`/`height` of an SVG when **no** HTML sizing attribute
   is present. This spec makes author-specified sizing work; changing the no-attribute
   default sizing behavior is a separate follow-up.
@@ -61,7 +72,13 @@ Out of scope (not changed by this spec):
    surrounded by leading/trailing whitespace) renders as an image in the Markdown
    viewer, using the tag's `src` attribute as the image source, resolved with the
    same rules as a Markdown `![alt](src)` image (data URIs, `http(s)://` URLs,
-   absolute paths, and paths relative to the document's directory).
+   absolute paths, and paths relative to the document's directory). If that source
+   *fails to load* (missing file, unreadable, or a decode failure), the tag behaves
+   exactly as a Markdown `![alt](missing.png)` does today — it stays an image block at
+   its configured size but renders as an empty box (no placeholder, broken-image icon,
+   or alt text), never a panic and never a fallback to literal text. This load-failure
+   visual is inherited from the existing image pipeline and unchanged by this spec; see
+   the tech spec's failure-state note for the mechanism.
 
 2. The tag's `alt` attribute, when present, is used as the image's alt text (the same
    role `alt` plays for a Markdown image). When `alt` is absent, alt text is empty.
@@ -155,9 +172,11 @@ Out of scope (not changed by this spec):
 
 13. Malicious or unexpected attribute values do not cause code execution or navigation.
     Only `src`, `alt`, `title`, `width`, `height`, and `align` are read from the tag;
-    all other attributes (including `onerror`, `onload`, `style`, `srcset`, etc.) are
-    ignored. `src` values are resolved through the same asset-source resolver used for
-    Markdown images and are never executed.
+    all other attributes (including `onerror`, `onload`, `style`, etc.) are ignored.
+    `src` values are resolved through the same asset-source resolver used for Markdown
+    images and are never executed. (`srcset` is also not read, but for a different
+    reason — it is a deferred *feature*, not a security drop; see the responsive-image
+    non-goal above.)
 
 14. Copy / export round-trips of a document containing a sized `<img>` preserve enough
     information to reproduce the rendered image. (The exact serialized form — whether the
--- a/specs/GH13721/product.md
+++ b/specs/GH13721/product.md
@@ -78,7 +78,11 @@ Out of scope (not changed by this spec):
    — the document scrolls. A pixel height is subject only to a sanity cap for
    unreasonable/hostile values (see the tech spec), not to the pane's dimensions. This
    asymmetry follows the guiding principle *model any reasonable markdown file, not any
-   possible HTML file*: a reasonable author-specified height should be respected.
+   possible HTML file*: a reasonable author-specified height should be respected. (The
+   one case where a specified height yields is when the image has no specified width and
+   its aspect-derived width would overflow the pane — see invariant 6; horizontal space
+   remains the hard constraint, so the box scales down uniformly rather than distorting
+   or overflowing.)
 
 5. When the tag specifies `width` as a percentage (`width="90%"`), the image is laid
    out at that percentage of the pane's available content width. A percentage `height`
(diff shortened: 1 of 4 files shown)


specs/GH13721/tech.md
```diff
@@ -0,0 +1,619 @@
...
+Covers invariants 4–8:
+
+- Pixel width smaller than the pane → `ImageBlockConfig.width == px`.
+- Pixel width larger than the pane → clamped to `available_width`.
+- Percent width → `available_width * p / 100`.
+- No dimensions → identical `ImageBlockConfig` to today (regression against the
+  hardcoded default).
+- **`width="200%"`** → clamps to `available_width` (invariant 5, same result as
+  `width="100%"`), not an unclamped `2 * available_width` overflow.
+- **`width="0%"` / `width="-10%"`** → floors at the `clamp_to_bound` minimum of `1px`,
```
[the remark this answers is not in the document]
fbartho: Addressed in c09aed648 (contradiction resolved: negatives are ignored at parse per the WHATWG dimension-parsing rules, cited in 38cccd54f; the clamp floor applies only to valid non-negative percents).

specs/GH13721/tech.md
```diff
@@ -0,0 +1,640 @@
...
+- **Aspect ratio when exactly one dimension is set (invariant 6):** resolve the
+  specified axis per the rules above (already clamped), then derive the other axis
+  from the intrinsic ratio using the Mermaid mechanism verbatim. The invariant that
+  governs every sub-case below: **the author-specified dimension, once resolved and
+  clamped, is never altered again by fallback or fit logic in any load state** —
+  pre-decode and post-decode alike. Only the *derived* (unspecified) axis is ever
+  adjusted layout-to-layout.
+  - Call `AssetCache::as_ref(app).load_asset::<ImageType>(asset_source.clone())`
+    (the `asset_source` is already computed at this point via
+    `resolve_asset_source`, `edit.rs:886`, so this requires no new resolution step).
```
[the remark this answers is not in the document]
fbartho: Addressed in b1115801f — §4 now specifies the implementable ordering (source resolved in from_styled_block, the AppContext-bearing phase, before config construction; threaded through LayoutTask). The sweep found and fixed one sibling phase-crossing gap (align_available_width stored on the config to cross layout→paint).

specs/GH13721/tech.md
```diff
@@ -0,0 +1,640 @@
...
+    a zero/unreadable intrinsic size:** this is the state that needs its own explicit
+    contract, because a naive "derived axis gets a plain default box" description
+    (what earlier drafts of this spec said) leaves a gap — see "Why the pre-decode
+    fallback needs `stretch()`, not `contain()`" below. The specified axis keeps its
+    resolved value unchanged (per the invariant above); the derived axis uses today's
+    plain default for that axis (`available_width` for a derived width,
+    `default_height` for a derived height) as before. What changes is *how the element
+    renders that box*: for this one transient layout, `RenderableImage::layout()`
+    (§4 "Where the offset is applied" sibling section, `render/element/image.rs:39-51`)
+    must use `Image::new(...).stretch()` instead of `.contain()` for this block. A
```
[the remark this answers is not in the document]
fbartho: Addressed in b1115801f — invariant 6 now admits one bounded pre-decode frame of derived-axis-only stretch, self-correcting on decode, with the transition covered at the layout-math layer.

specs/GH13721/product.md
```diff
@@ -0,0 +1,125 @@
...
+   uses as a Markdown image title (round-trip serialization / export / tooltip where
+   applicable). An absent or empty `title` is treated as no title.
+
+4. When the tag specifies `width` as an integer number of pixels (`width="640"` or
+   `width="640px"`), the image is laid out at that width in pixels, clamped so it never
+   exceeds the available content width of the pane. `height` is treated the same way.
+
+5. When the tag specifies `width` as a percentage (`width="90%"`), the image is laid
+   out at that percentage of the pane's available content width. `height` as a
+   percentage is likewise relative to the available height budget.
```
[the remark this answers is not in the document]
fbartho: Addressed in b1115801f — invariant 5 names default_height as the percent-height reference. The same pass encodes a product decision: explicit pixel heights are now honored (sanity-capped at 8192px, the conservative GPU single-texture floor) instead of clamping to the default height budget.

specs/GH13721/tech.md
```diff
@@ -0,0 +1,716 @@
...
+    straight into this cache query for the intrinsic size and (ii) threaded onto the
+    `LayoutTask::Image` variant so `into_block_item` reuses it instead of re-resolving.
+    Net effect: the source is resolved exactly once, in the `AppContext`-bearing layout
+    phase, which is the only phase that can perform this query — no second resolution,
+    no `AppContext`-less call site trying to size the image.
+  - **Post-decode — `AssetState::Loaded { data }` with `data.image_size()` returning
+    `Some((intrinsic_w, intrinsic_h))` with both `> 0`:** derive the missing axis from
+    the specified axis's *resolved* (post-clamp) value — given `width`,
+    `height = width * intrinsic_h / intrinsic_w`; given `height`,
+    `width = min(height * intrinsic_w / intrinsic_h, available_width)`. The specified
```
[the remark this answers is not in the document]
fbartho: Addressed in 774f4ef87 — precedence is now documented as aspect ratio > pane-width bound > specified height: when the derived width would exceed available_width, the whole box scales down uniformly (width = pane, effective height proportionally below the specified value), never distorting and never overflowing. Mirrors the existing width-side rule; the deliberate divergence from browsers (which allow horizontal overflow) is noted with the pane-model rationale, and invariants 4/6 carry the exception explicitly.

fbartho:
(Just commenting as this is waiting for review).

fbartho:
@bnavetta This PR is still waiting for review. (Just trying to avoid the stale-bot)

specs/GH13721/product.md
```diff
@@ -0,0 +1,185 @@
...
+   unreasonable/hostile values (see the tech spec), not to the pane's dimensions. This
+   asymmetry follows the guiding principle *model any reasonable markdown file, not any
+   possible HTML file*: a reasonable author-specified height should be respected. (The
+   one case where a specified height yields is when the image has no specified width and
+   its aspect-derived width would overflow the pane — see invariant 6; horizontal space
+   remains the hard constraint, so the box scales down uniformly rather than distorting
+   or overflowing.)
+
+5. When the tag specifies `width` as a percentage (`width="90%"`), the image is laid
+   out at that percentage of the pane's available content width. A percentage `height`
```
bnavetta: Since the default image height is an implementation detail, I’m not sure this will be intuitive. I think omitting support for percentage-based height would be reasonable, but it could also be a percentage of the viewport height

specs/GH13721/tech.md
```diff
@@ -0,0 +1,791 @@
...
+```rust
+pub struct FormattedImage {
+    pub alt_text: String,
+    pub source: String,
+    pub title: Option<String>,
+    /// Author-specified width from a raw-HTML `<img width>` attribute.
+    /// `None` for Markdown `![alt](src)` images, which have no sizing syntax.
+    pub width: Option<ImageDimension>,
+    pub height: Option<ImageDimension>,
+    pub align: ImageAlign,   // defaults to Left
```
bnavetta: I believe Markdown images default to center - this should be preserved

specs/GH13721/tech.md
```diff
@@ -0,0 +1,791 @@
...
+    /// Percentage of the available content dimension, e.g. `width="90%"`.
+    Percent(f32),
+}
+
+#[derive(Default)]
+pub enum ImageAlign { #[default] Left, Center, Right }
+```
+
+`FormattedImage` stays `Clone + Debug + PartialEq + Eq`-compatible with the rest of the
+enum. Because `f32` is not `Eq`, either store the parsed dimension as an integer
```
bnavetta: Elsewhere we use `OrderedFloat`

specs/GH13721/tech.md
```diff
@@ -0,0 +1,791 @@
...
+or generally `ImageType::image_size()` at `warpui_core/src/image_cache.rs:472-484`,
+which also handles `StaticBitmap`/`AnimatedBitmap`) and computes
+`height = width * intrinsic_height / intrinsic_width` (`mermaid_diagram.rs:104-106`).
+When the asset is not yet `Loaded` (`Loading`/`FailedToLoad`/`Evicted`), it falls back
+to a height-multiplier default (`mermaid_diagram_config`, `:54-71`) — the same shape of
+fallback `BufferBlockItem::Image` already uses today, just parameterized instead of
+hardcoded. This is a real, shipped, layout-time re-derivation, not a speculative
+"generous cap": every time editor content layout re-runs (the same re-run that lets a
+`Loading` Mermaid diagram flip to a rendered `MermaidDiagram` block once its asset
+resolves — driven by the normal buffer/viewport invalidation path, not by the paint
```
bnavetta: Won’t this need to use `repaint_after_load`, as the image loading won’t otherwise trigger a new layout pass? I believe Mermaid diagrams have dedicated logic to request layout as part of their toggleable UI

bnavetta (commented):
Sorry for the delay!

specs/GH13721/product.md
fbartho: I agree that percentage height isn't likely to be relevant in markdown images. Dropped it in 1d4e85291

specs/GH13721/tech.md
fbartho: Got it! Applied in 1d4e85291

specs/GH13721/tech.md
fbartho: Got it. In 1d4e85291 I used the asset's `when_loaded` to trigger `invalidate_layout_for_range` + `add_pending_edit` in addition to also connecting to `repaint_after_load`

specs/GH13721/tech.md
fbartho: I agree that that was the previous default. I've implemented that centering in 1d4e85291.

@bnavetta: Should this be documented as explicit design/policy choice of Warp as a known variance from github/other markdown renderers?

PR #13983, opened 2026-07-19
fix(unix): check return values of libc calls in PTY setup

### LINKED ISSUE #12950: fix(unix): multiple safety/correctness bugs in getpwuid_r and PTY setup
### Pre-submit Checks

### Describe the bug

Several safety and correctness bugs exist in the Unix PTY/shell-spawn code path (`app/src/terminal/local_tty/unix.rs`):  
  
1. **Undefined behavior**: `entry.assume_init()` is called before checking whether `getpwuid_r` succeeded, reading uninitialized memory on failure.  
2. **Wrong error check**: The code checks `status < 0`, but `getpwuid_r` returns a *positive* errno on failure — it never returns -1. Failures like `ERANGE` (34) are silently ignored.  
3. **Panics instead of errors**: `get_pw_entry` calls `panic!("getpwuid_r failed")` and `panic!("pw not found")` rather than returning a `Result`, crashing the process instead of surfacing a recoverable error.  
4. **Unwrap on untrusted C strings**: `pw_name`, `pw_dir`, and `pw_shell` are converted with `.unwrap()`, panicking if the passwd entry contains non-UTF-8 bytes.  
5. **Unchecked `fcntl` return values**: Both `fcntl(FD_CLOEXEC)` calls in `make_pty` discard their return values, silently leaving fds without close-on-exec if the call fails.  
6. **No `ERANGE` retry**: `getpwuid_r` is called with a fixed 1024-byte buffer. If the passwd entry is larger (e.g. long home directory path), it returns `ERANGE` and the entry is never read.  

### To reproduce

1. Run Warp on a Unix system where the current user's passwd entry is missing or malformed (e.g. in a container, CI environment, or with an unusual NSS configuration).  
2. Alternatively, have a home directory path longer than ~900 characters to trigger the ERANGE buffer overflow.  
3. Observe Warp crash with a panic message such as "getpwuid_r failed" or "pw not found" instead of showing a graceful error.  

### Expected behavior

Warp should propagate passwd lookup failures as recoverable errors rather than panicking. The PTY spawn path should return an error to the caller, allowing Warp to display a meaningful message to the user instead of crashing.  

N/A — this is a code-level safety/correctness bug. A backtrace from a crash would show the panic originating in `get_pw_entry` in `app/src/terminal/local_tty/unix.rs`.  

### Operating system (OS)

macOS

Linux and/or macOS (Unix only — this code is #[cfg(unix)])

### Shell Version

N/A (crash occurs before shell is launched)

### Current Warp version

No : these bugs appear to have been present since the code was written.

### Regression

### Recent working Warp date

_No response_

### Additional context

Affected file: app/src/terminal/local_tty/unix.rs  
Affected functions: get_pw_entry, get_pw_shell, make_pty, build_host_shell_command, build_docker_sandbox_command  
  
Most likely to trigger in:  
- Docker/container environments where the user has no passwd entry  
- CI systems with minimal /etc/passwd  
- Systems using LDAP/NIS/SSSD where getpwuid_r may return ERANGE for large entries  

No

None

Jason-Shen2:
## Summary

Fixes #12950 (the remaining safety/correctness issues in the Unix PTY spawn path).

### Changes

The original bug report described six issues; five of the six (getpwuid_r undefined behavior, wrong errno check, panics on lookup failure, non-UTF-8 C string unwraps, ERANGE retry) were already resolved by a prior refactoring that replaced the manual get_pw_entry/get_pw_shell functions with nix's safe User::from_uid wrapper plus a three-tier fallback chain (in-process getpwuid_r, then getent passwd, then /etc/passwd).

This PR fixes the remaining issues that were still present:

1. Check fcntl(FD_CLOEXEC) return values in make_pty() - previously a failure to set close-on-exec would silently leak PTY file descriptors into spawned child processes
2. Check sigemptyset/sigprocmask return values in the pre_exec fork child using the existing cvt() helper
3. Replace panic!() in on_resize() TIOCSWINSZ failure with log::warn!() - a resize ioctl failure is not fatal
4. Replace assert_eq!() in set_nonblocking() with proper io::Result propagation

### Validation

- All 110 warp_terminal unit tests pass
- cargo clippy -p warp_terminal -- -D warnings passes
- cargo fmt applied

### DIFF
--- a/app/src/terminal/local_tty/unix.rs
+++ b/app/src/terminal/local_tty/unix.rs
@@ -57,11 +57,16 @@ fn make_pty(size: winsize) -> Result<(RawFd, RawFd)> {
     // will _not_ be closed when we exec the shell.
     unsafe {
         if libc::fcntl(ends.master, libc::F_SETFD, libc::FD_CLOEXEC) == -1 {
-            return Err(io::Error::last_os_error())
-                .context("fcntl FD_CLOEXEC on master pty failed");
+            let err = io::Error::last_os_error();
+            libc::close(ends.master);
+            libc::close(ends.slave);
+            return Err(err).context("fcntl FD_CLOEXEC on master pty failed");
         }
         if libc::fcntl(ends.slave, libc::F_SETFD, libc::FD_CLOEXEC) == -1 {
-            return Err(io::Error::last_os_error()).context("fcntl FD_CLOEXEC on slave pty failed");
+            let err = io::Error::last_os_error();
+            libc::close(ends.master);
+            libc::close(ends.slave);
+            return Err(err).context("fcntl FD_CLOEXEC on slave pty failed");
         }
     }
 
@@ -625,7 +630,12 @@ impl Pty {
         let fd = unsafe {
             // Maybe this should be done outside of this function so nonblocking
             // isn't forced upon consumers. Although maybe it should be?
-            set_nonblocking(leader_fd).context("failed to set leader fd non-blocking")?;
+            if let Err(err) = set_nonblocking(leader_fd) {
+                libc::close(leader_fd);
+                let mut pty_handle = pty_handle;
+                let _ = pty_handle.kill();
+                return Err(err).context("failed to set leader fd non-blocking");
+            }
 
             File::from_raw_fd(leader_fd)
         };
--- a/app/src/terminal/local_tty/unix.rs
+++ b/app/src/terminal/local_tty/unix.rs
@@ -14,9 +14,9 @@ use std::{io, ptr};
 use anyhow::{Context as _, Error, Result};
 use command::blocking::Command;
 use itertools::Itertools;
-use libc::{self, c_int, winsize, TIOCSCTTY};
-use mio::unix::SourceFd;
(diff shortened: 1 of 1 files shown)


Jason-Shen2:
Both fd-leak issues are addressed in the latest commits:

1. **CLOEXEC failure in `make_pty`**: Both master and slave fds are now explicitly closed with `libc::close` before returning an error if either `fcntl(FD_CLOEXEC)` call fails, preventing raw PTY fd leaks on the error path.

2. **Nonblocking setup failure after spawn**: `set_nonblocking` now returns `io::Result<()>` instead of asserting, and on failure:
   - The leader fd is closed with `libc::close`
   - The spawned child PTY process is killed via `pty_handle.kill()`
   - Error is propagated with context
   - `File::from_raw_fd` is only called after successful nonblocking setup

Additional fixes in these commits:
- Check return values of `sigemptyset`/`sigprocmask` in the pre_exec hook (previously ignored)
- Replaced `panic!` on `TIOCSWINSZ` ioctl failure with `log::warn!`
- All libc return values are checked via the `cvt()` helper instead of being silently ignored

/oz-review

PR #13984, opened 2026-07-20
Store the staging IAP cache in the Warp home directory

IsaiahWitzke:
see original slack thread here for context: https://warpdev.slack.com/archives/C0B9XLRE6UE/p1784502573576239

## Summary
- store the staging IAP token cache under `warp_core::paths::warp_home_config_dir()`
- use `~/.warp-dev/staging/iap_cache.jwt` consistently across macOS and Linux
- preserve development-profile isolation through the existing Warp home directory derivation

## Validation
- `cargo fmt --all -- --check` passed before the final path-only adjustment
- `git diff --check`

Focused `warp_server_client` IAP tests were started but interrupted before completion.

Companion: https://github.com/warpdotdev/factory-agents/pull/127
Conversation: https://staging.warp.dev/conversation/e54a361e-2d94-4106-aeb1-783aa75d3188

### DIFF
--- a/crates/warp_server_client/src/iap.rs
+++ b/crates/warp_server_client/src/iap.rs
@@ -763,9 +763,8 @@ fn parse_aud_from_jwt(token: &str) -> Option<String> {
 /// it instead of each doing their own STS/IAM mint. The file stores the raw IAP
 /// JWT; validity is derived from its own `exp` claim on read.
 ///
-/// This is a plain owner-only file rather than OS secure storage: it only exists
-/// in the staging Oz runner (headless Linux), where no keyring is available and
-/// the container already holds the bootstrap JWT / API key / GitHub token.
+/// This is a plain owner-only file rather than OS secure storage so short-lived
+/// child processes can read it directly.
 #[cfg(not(target_family = "wasm"))]
 mod cache {
     use std::time::Duration;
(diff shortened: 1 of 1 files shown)


crates/warp_server_client/src/iap.rs
```diff
@@ -779,7 +778,7 @@ mod cache {
     const CACHE_SKEW: Duration = Duration::from_secs(30);
 
     fn cache_path() -> std::path::PathBuf {
-        warp_core::paths::secure_state_dir()
+        warp_core::paths::warp_home_config_dir()
```
IsaiahWitzke, line 781: basically, i want [this pr](https://github.com/warpdotdev/factory-agents/pull/127) to be able to pull from a static iap cache file thats going to be the same on mac or linux runners.

I have a feeling that we're going to want to do a bit more raw curl-ing to the staging server with non oz-dev processes... so it'd be nice if there's a dedicated place for the iap cache to live

bnavetta: Fair enough - one thing we can do to harden this a little is to only write it in a cloud agent context (i.e. `ServerApi::ambient_agent_task_id` is set, or equivalently, the `OZ_RUN_ID` environment variable)

bnavetta (approved):

crates/warp_server_client/src/iap.rs
IsaiahWitzke: yep -- this is effectively already done. We only write to the cache file when [managed_mint is some](https://github.com/warpdotdev/warp/blob/569a51b22a440fc1798e0690148cd1168e4d8ddc/crates/warp_server_client/src/iap.rs#L414), which is only set [when ambient_agent_task_id is some](https://github.com/warpdotdev/warp/blob/569a51b22a440fc1798e0690148cd1168e4d8ddc/app/src/lib.rs#L2211)

PR #13986, opened 2026-07-20
Set up a real macOS main thread run loop for headless warpui backend.

vorporeal:
## Description

The macOS headless `warpui` backend currently blocks the process main thread waiting on its Rust event channel. Although work handled by that loop runs on the main thread, the loop does not service the native CFRunLoop or GCD main queue. A worker that calls `dispatch2::run_on_main` can therefore deadlock indefinitely.

This limitation led #11671 to replace main-thread dispatch in the Carbon keycode cache with mutex-based serialization. The mutex prevented concurrent Carbon calls, but did not satisfy Carbon's main-thread requirement; #13547 provides the immediate call-site fix for that regression.

This PR fixes the underlying headless event-loop behavior:

- On macOS, headless event senders enqueue the existing `AppEvent`, signal a `CFRunLoopSource`, and wake the process main thread's CFRunLoop.
- The run-loop source stops the current run-loop invocation so queued Warp events can be drained and handled in ordinary Rust code.
- The headless run loop verifies that it is initialized on the process main thread.
- Non-macOS headless backends retain the existing blocking channel behavior.
- The Carbon keycode cache can use `dispatch2::run_on_main` again, transparently running inline on the main thread or synchronously dispatching from a worker.

This preserves the existing Warp foreground executor and event dispatch model while allowing native main-queue work to make progress.

## Related changes

- #11671 introduced the mutex workaround because headless Warp did not service the GCD main queue.
- #13547 moves current computer-use actor construction onto the main thread as an immediate regression fix.

## Testing

Added a custom `harness = false` regression test that owns the process main thread, starts the headless backend, invokes `dispatch2::run_on_main` from a worker, and verifies that the callback executes on the primordial thread and returns without deadlocking.

- `MACOSX_DEPLOYMENT_TARGET=10.14 cargo check -p warpui -p computer_use --tests`
- `MACOSX_DEPLOYMENT_TARGET=10.14 cargo nextest run -p warpui` — 42 passed, 1 skipped.
- `MACOSX_DEPLOYMENT_TARGET=10.14 cargo nextest run -p warpui -E 'test(services_main_dispatch_queue)'`
- `cargo nextest run -p computer_use --no-tests=pass` — test targets compiled; the package currently has no tests.
- `MACOSX_DEPLOYMENT_TARGET=10.14 cargo clippy -p warpui -p computer_use --all-targets --no-deps -- -D warnings`
- `./script/format`
- `git diff --check`

A full dependency clippy run remains blocked by pre-existing `clippy::unnecessary_unwrap` findings in `warpui_core/src/elements/gui/hoverable.rs`; clippy passes for both modified packages.

- [x] I have manually tested a headless computer-use action locally.

### Screenshots / Videos

Not applicable; this PR has no user-visible UI changes.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/crates/warpui/src/platform/headless/event_loop.rs
+++ b/crates/warpui/src/platform/headless/event_loop.rs
@@ -7,11 +7,8 @@ use std::sync::mpsc::TryRecvError;
 use std::sync::mpsc::{self, Receiver, SendError};
 
 #[cfg(target_os = "macos")]
-use core_foundation::base::{TCFType, kCFAllocatorDefault};
-#[cfg(target_os = "macos")]
-use core_foundation::runloop::{
-    CFRunLoop, CFRunLoopSource, CFRunLoopSourceContext, CFRunLoopSourceCreate,
-    CFRunLoopSourceSignal, CFRunLoopWakeUp, kCFRunLoopDefaultMode,
+use objc2_core_foundation::{
+    CFRetained, CFRunLoop, CFRunLoopSource, CFRunLoopSourceContext, kCFRunLoopDefaultMode,
 };
 
 use crate::platform::app::{
@@ -82,8 +79,8 @@ pub(super) fn channel() -> (EventSender, EventReceiver) {
 
 #[cfg(target_os = "macos")]
 struct RunLoopSignal {
-    run_loop: CFRunLoop,
-    source: CFRunLoopSource,
+    run_loop: CFRetained<CFRunLoop>,
+    source: CFRetained<CFRunLoopSource>,
 }
 
 #[cfg(target_os = "macos")]
@@ -92,7 +89,7 @@ impl RunLoopSignal {
         objc2::MainThreadMarker::new()
             .expect("the macOS headless event loop must run on the process main thread");
 
-        let run_loop = CFRunLoop::get_current();
+        let run_loop = CFRunLoop::current().expect("the current thread must have a run loop");
         let mut context = CFRunLoopSourceContext {
             version: 0,
             info: std::ptr::null_mut(),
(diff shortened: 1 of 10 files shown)


vorporeal:
* **#13986** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13986?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/13986?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warpui/src/platform/headless/event_loop.rs
```diff
@@ -21,50 +34,155 @@ pub(super) enum AppEvent {
...
+}
+
+#[cfg(target_os = "macos")]
+impl RunLoopSignal {
+    fn new() -> Self {
+        objc2::MainThreadMarker::new()
+            .expect("the macOS headless event loop must run on the process main thread");
+
+        let run_loop = CFRunLoop::get_current();
+        let mut context = CFRunLoopSourceContext {
```
bnavetta, line 93: Is there a`core_foundation` API to construct this? Probably fine since it’s unlikely to change, but seems like fields we shouldn’t have to set

crates/warpui/tests/headless_main_thread.rs
```diff
@@ -0,0 +1,100 @@
+const TEST_NAME: &str = "services_main_dispatch_queue";
+
+fn main() {
```
bnavetta, line 3: Definitely not blocking, but we could add something like https://github.com/LukasKalbertodt/libtest-mimic to handle the test flags

bnavetta (approved):

crates/warpui/src/platform/headless/event_loop.rs
vorporeal: unfortunately not.  i realized though, that we should be using `objc2-core-foundation` for all of this stuff; updated to do so.

crates/warpui/tests/headless_main_thread.rs
vorporeal: oh that's a nice crate; done.

PR #13992, opened 2026-07-20
fix: use configured shell path for SSH bootstrap

### LINKED ISSUE #13934: SSH bootstrap fails when the remote login shell lives outside the default PATH (Nix/home-manager): bare-name `exec -l zsh` -> "command not found: zsh", session closes
### Bug description

`ssh` from Warp (with the SSH integration / Warpify active) fails to open a session on hosts where the user's login shell is installed **outside the system default `PATH`** — the canonical case being single-user Nix / home-manager setups, where zsh exists only at `~/.nix-profile/bin/zsh` and the passwd entry points at that absolute path.

Observed in a fresh Warp install (macOS, latest stable as of 2026-07-18) connecting to an Ubuntu 24.x VPS:

```text
ssh alex@tyrode.dev
zsh:58: command not found: zsh
Connection to tyrode.dev closed.
```

(The `zsh:58:` prefix is from the assembled inline bootstrap script streamed to the remote shell — line numbering is of the generated script, not of a source file.)

Plain `ssh` from any other terminal works, because sshd execs the login shell via its **absolute path** from `/etc/passwd`; no bare-name lookup is involved.

### Root cause

The zsh bootstrap ends by re-exec'ing the user shell **by bare name**:

https://github.com/warpdotdev/warp/blob/26b7c9cdbd011749480be8ace9b7e8d7d3d5da8c/app/assets/bundled/bootstrap/zsh_body.sh — final `case` arm:

```sh
TMPPREFIX="'$HOME/.zshtmp-'" WARP_SSH_RCFILES="'${ZDOTDIR:-$HOME}'" WARP_HONOR_PS1="'$WARP_HONOR_PS1'" ZDOTDIR="'$WARP_TMP_DIR'" exec -l zsh -g $TRACE_FLAG_IF_WARP_SHELL_DEBUG_MODE
```

At that point the environment is the non-interactive sshd default (`ENV_PATH`, e.g. `/usr/local/bin:/usr/bin:/bin:...`) — the same "system default $PATH" discipline the file's own header comment prescribes via `command -p` for external binaries. If the shell binary is not in those directories, `exec -l zsh` aborts the bootstrap and the connection closes.

### Reproduction (server side)

Any Linux host where zsh is Nix-managed:

```text
$ getent passwd alex | cut -d: -f7
/home/alex/.nix-profile/bin/zsh          # absolute Nix path; sshd execs this fine

$ ls /usr/bin/zsh /bin/zsh
ls: cannot access '/usr/bin/zsh': No such file or directory
ls: cannot access '/bin/zsh': No such file or directory

$ env PATH=/usr/local/bin:/usr/bin:/bin sh -c 'command -v zsh || echo NOT_FOUND'
NOT_FOUND
```

Then `ssh` to that host from Warp with the SSH integration enabled → error above; `~/.warp*` never gets created (the failure precedes remote-server install). Homebrew-only shells on Linux hosts and other prefix-managed installs hit the same wall.

### Expected behavior

The bootstrap re-execs the shell it already targeted. Warp necessarily knows *which* shell it is bootstrapping (it selected `zsh_body.sh`); the durable fixes are either:

1. Resolve the shell to an absolute path before exec — from the login shell sshd already started (`$SHELL` / `getent passwd "$USER" | cut -d: -f7`) — and `exec -l "$resolved_shell"`; or
2. Prepend the detected shell's directory to `PATH` before the bare-name exec.

The equivalent bare-name exec in the other shell bodies presumably has the same exposure (unverified beyond zsh).

### Workaround

`sudo ln -s ~/.nix-profile/bin/zsh /usr/local/bin/zsh` on the remote host (works because `ENV_PATH` includes `/usr/local/bin`).

### Related

#5137 ("Warp fails to bootstrap when RC files change the startup shell") is adjacent — both are bootstrap assumptions about the remote shell environment — but this one requires no RC-file tricks: a stock home-manager host with `chsh` to the Nix zsh is sufficient.

### Environment

- Client: Warp latest stable (fresh install 2026-07-18), macOS
- Server: Ubuntu (kernel 6.8, x86_64), OpenSSH, login shell `/home/alex/.nix-profile/bin/zsh` (home-manager); no zsh in `/usr/bin`, `/bin`, or `/usr/local/bin`

CINC0S:
## Description

Fixes SSH/Warpify bootstrap when the remote bash or zsh executable is only available at the absolute path configured by sshd in `$SHELL`, rather than by its bare name in the default SSH `PATH`.

The three embedded SSH wrapper payloads now re-exec `"$SHELL"` for bash and zsh while preserving the existing bash argv0 and zsh login-shell flags. A regression test covers every duplicated payload and rejects the former bare-name commands.

## Linked Issue

Fixes #13934

  - The issue is triaged with `repro:high`, but currently has neither ready label.
- [x] Screenshots or video are not applicable to this shell bootstrap change.

## Testing

Automated and manual verification performed:

- [x] `./script/format`
- [x] `git diff --check`
- [x] `bash -n app/assets/bundled/bootstrap/bash_body.sh`
- [x] Verified all three payloads use the configured `$SHELL` path and no longer contain the bare-name bash/zsh re-exec commands.
- [x] Simulated a restricted remote `PATH` with `PATH=/nonexistent` and confirmed an absolute bash `$SHELL` path still re-execs successfully.
- [x] Installed and verified the Debian 13 Clang 19 runtime and generic headers; the previously blocked `minimp4-sys` bindgen build now completes successfully.
- [ ] `cargo test -p warp --lib test_ssh_wrapper_reexecs_configured_login_shell -j 1` compiled through the Warp application crate, but the final test-binary link was stopped after the 7.5 GiB host entered heavy swapping; the test did not execute.
- [ ] `cargo clippy --workspace --exclude warp_completer --all-targets --tests -j 1 -- -D warnings` passed the former bindgen failure and checked the supporting workspace through the main `warp` crate without lint diagnostics, but the final application analysis was stopped under the same host memory constraint before the command completed.
- [ ] I have manually tested my changes locally with `./script/run` (not run; this requires a remote Nix/home-manager SSH setup and the full GUI toolchain).

## Maintainer review checklist

- [ ] Confirm `$SHELL` is the intended source of the executable path across supported sshd configurations.
- [ ] Confirm the existing bash argv0 and zsh login-shell semantics remain intact.
- [ ] Run the targeted Rust regression test and complete Clippy in CI or a development environment with sufficient memory.
- [ ] Verify an end-to-end SSH/Warpify session where zsh is installed outside sshd's default `PATH` (for example, through Nix/home-manager).

## Agent Mode

CHANGELOG-BUG-FIX: SSH sessions now bootstrap correctly when the remote login shell is installed outside the default PATH.

### DIFF
--- a/app/assets/bundled/bootstrap/bash_body.sh
+++ b/app/assets/bundled/bootstrap/bash_body.sh
@@ -1110,7 +1110,7 @@ fi
 
 case "'${SHELL##*/}'" in
   bash)
-    exec -a bash bash --rcfile <(echo '"'
+    exec -a bash "'$SHELL'" --rcfile <(echo '"'
       command -p stty raw
       HISTCONTROL=ignorespace
       HISTIGNORE=" *"
@@ -1137,7 +1137,7 @@ if [[ "'$?'" == 0 ]]; then
 else
   echo \"Failed to bootstrap warp. Continuing with a non-bootstrapped shell.\"
 fi
-TMPPREFIX="'$HOME/.zshtmp-'" WARP_SSH_RCFILES="'${ZDOTDIR:-$HOME}'" ZDOTDIR="'$WARP_TMP_DIR'" exec -l zsh -g $TRACE_FLAG_IF_WARP_SHELL_DEBUG_MODE
+TMPPREFIX="'$HOME/.zshtmp-'" WARP_SSH_RCFILES="'${ZDOTDIR:-$HOME}'" ZDOTDIR="'$WARP_TMP_DIR'" exec -l "'$SHELL'" -g $TRACE_FLAG_IF_WARP_SHELL_DEBUG_MODE
       ;;
 esac
 "
--- a/app/assets/bundled/bootstrap/fish.sh
+++ b/app/assets/bundled/bootstrap/fish.sh
@@ -727,7 +727,7 @@ fi
 
 case "'${SHELL##*/}'" in
 bash)
-    exec -a bash bash --rcfile <(echo '"'
+    exec -a bash "'$SHELL'" --rcfile <(echo '"'
       stty raw
       HISTCONTROL=ignorespace
       HISTIGNORE=" *"
@@ -755,7 +755,7 @@ if [[ "'$?'" == 0 ]]; then
 else
   echo \"Failed to bootstrap warp. Continuing with a non-bootstrapped shell.\"
 fi
-TMPPREFIX="'$HOME/.zshtmp-'" WARP_SSH_RCFILES="'${ZDOTDIR:-$HOME}'" ZDOTDIR="'$WARP_TMP_DIR'" exec -l zsh -g $TRACE_FLAG_IF_WARP_SHELL_DEBUG_MODE
+TMPPREFIX="'$HOME/.zshtmp-'" WARP_SSH_RCFILES="'${ZDOTDIR:-$HOME}'" ZDOTDIR="'$WARP_TMP_DIR'" exec -l "'$SHELL'" -g $TRACE_FLAG_IF_WARP_SHELL_DEBUG_MODE
     ;;
 esac
 "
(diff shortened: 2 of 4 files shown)


CINC0S:
@cla-bot check

PR #14009, opened 2026-07-20
fix(markdown): strip HTML comments instead of rendering them as text

### LINKED ISSUE #13977: Markdown viewer: strip HTML comments instead of rendering them as text
## Problem

The Markdown viewer renders HTML comments (`<!-- … -->`) as visible literal text. Comments are metadata, not content — every other markdown renderer (GitHub, VS Code preview, Obsidian) strips them from rendered output. Documents that use comments as section identifiers, editorial notes, or tooling markers (a common pattern in READMEs and docs maintained by scripts) display their internal plumbing to the reader.

## Reproduction

Render this in the Markdown viewer:

```markdown
<!-- section: quick-start -->
# Quick Start

Some intro text. <!-- TODO: expand this -->

<!--
multi-line
block comment
-->
Closing paragraph.
```

**Expected** (GitHub/CommonMark behavior): only the heading and the two paragraphs render; all three comments are invisible.

**Actual**: the comment text renders inline/as blocks alongside the content.

## Spec grounding

CommonMark classifies `<!-- … -->` as an HTML block (type 2 start condition `<!--`, end condition `-->`) and as inline raw HTML; renderers that pass raw HTML through emit it for the *browser* to hide. A viewer that renders raw HTML as literal text (as Warp's does for unsupported tags) inverts that: the correct degradation for comments specifically is to strip them, since their semantic content is "not content." Both block-level comments (own line) and inline comments (mid-paragraph) apply; an unterminated `<!--` should degrade deterministically to literal text rather than swallowing the rest of the document.

## Notes

- Part of the raw-HTML subset effort (#13652); independent of and smaller than the tag-rendering features — stripping needs no layout or style machinery, only tokenizer recognition of the comment span.
- Multi-line block comments need the block parser to consume until `-->` (analogous to fenced code block state); inline comments need an inline token that emits nothing.

maxmilian:
## Description

The Markdown viewer had no handling for HTML comments, so `<!-- ... -->` rendered as visible text. This fixes it in the shared parser rather than the view layer, as the triage note on #13977 suggested.

Root cause, at both levels of `crates/markdown_parser`:

- **Inline** — `parse_inline_token`'s `alt` had no comment branch, so `<`, `!` and `-` fell through to `unmatched_char` (the deliberate catch-all at the end of the chain) and became `InlineToken::Text` → visible `FormattedTextFragment`s.
- **Block** — `parse_markdown_internal`'s `alt` had no comment branch either, so a comment line fell through to `parse_paragraph` and rendered as an ordinary line.

The fix adds `parse_html_comment` (`<!--` … `-->`, multi-line) and consumes comments at both levels:

- Block-level comments are handled in the `parse_markdown_internal` loop rather than inside `alt`, because every `alt` branch must yield a `FormattedTextLine` and a comment must yield *nothing at all*. It runs before the `alt` since a comment can only begin at `<!--`, which no other block construct claims.
- Inline comments become an `InlineToken::Comment` that is discarded without emitting a fragment.

Two ordering decisions worth calling out, both pinned by tests:

- `code_span` is ordered **ahead of** the inline comment branch, so `` `<!-- x -->` `` still renders literally.
- An unterminated `<!--` makes the parser **fail** rather than consume to end-of-input, so it stays literal per CommonMark (HTML block type 2).

Scope is parser-only — no view-layer changes. This is deliberately narrower than #13652's broader raw-HTML work and does not overlap it.

## Linked Issue

Closes #13977

## Testing

- [x] `cargo fmt -- --check`
- [x] `cargo check -p markdown_parser`
- [x] `PATH="/tmp/warp-corepack-shims:$PATH" cargo clippy -p markdown_parser --all-targets --all-features --tests -- -D warnings`
- [x] `cargo nextest run --no-fail-fast -p markdown_parser` — **147 passed, 0 failed** (6 new + 141 existing regression tests)

Six tests were added to `markdown_parser_tests.rs`, covering the five cases from the triage comment plus a code-span guard. They were written first and confirmed red before the fix: the three "should be stripped" cases failed, while the three "should be preserved" cases (unterminated, fenced code block, code span) passed from the start and now serve as regression guards.

- `test_parse_html_comment_inline_is_stripped`
- `test_parse_html_comment_single_line_block_is_stripped`
- `test_parse_html_comment_multi_line_block_is_stripped`
- `test_parse_unterminated_html_comment_stays_literal`
- `test_parse_html_comment_inside_code_block_is_preserved`
- `test_parse_html_comment_inside_code_span_is_preserved`

### Screenshots / Videos

Captured from two local `./script/run` dev builds of the same worktree — the only difference is this PR's parser change (`crates/markdown_parser/src/markdown_parser.rs` reverted for the "before" build, restored for the "after" build). Both open the same file in the Markdown viewer at identical window geometry.

Before and after: HTML comments in the Markdown viewer

Source file used:

```markdown
# HTML comment handling

<!-- this whole line is a comment and should not render -->

Inline <!-- hidden note --> comment inside a paragraph.

<!--
a multi-line comment
spanning several lines
-->

Code span keeps it literal: `<!-- not a comment -->`

Unterminated stays literal: <!-- no closing tag here
```

**Before** — the standalone comment line, the inline comment and the whole multi-line comment all render as visible text.
**After** — all three are gone, while the code span and the unterminated `<!--` are still rendered literally, which is the intended behavior.

## Agent Mode

### DIFF
--- a/crates/markdown_parser/src/markdown_parser.rs
+++ b/crates/markdown_parser/src/markdown_parser.rs
@@ -9,7 +9,7 @@ use nom::bytes::complete::{
 use nom::character::complete::{char, one_of, satisfy, space0, space1};
 use nom::character::is_digit;
 use nom::combinator::{
-    all_consuming, consumed, eof, fail, flat_map, map, map_parser, recognize, value, verify,
+    all_consuming, consumed, eof, fail, flat_map, map, map_parser, opt, recognize, value, verify,
 };
 use nom::error::{ContextError, ErrorKind, ParseError, context, make_error};
 use nom::multi::{fold_many_m_n, fold_many1, many_m_n, many0};
@@ -193,6 +193,14 @@ fn parse_markdown_internal<'a, E: ContextError<&'a str> + ParseError<&'a str>>(
     let mut remaining = markdown;
     let mut lines = Vec::new();
     while !remaining.is_empty() {
+        // Block-level comments produce no line at all, so they are handled outside `block`, which
+        // must yield a `FormattedTextLine`. This runs first because a comment can only start at
+        // `<!--`, which no other block construct claims.
+        if let Ok((remaining_after_comment, _)) = parse_html_comment_block::<E>(remaining) {
+            remaining = remaining_after_comment;
+            continue;
+        }
+
         let (remaining_after_block, mut line) = block(remaining)?;
         remaining = remaining_after_block;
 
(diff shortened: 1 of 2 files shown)


fbartho:
I took a stab at addressing Oz's feedback here: https://github.com/maxmilian/warp/pull/5 feel free to ignore it if you have a different approach

maxmilian:
Thanks for the catch — you're right. `opt(parse_line_ending)` let the block-comment branch succeed even when non-whitespace followed the closing `-->` on the same line, so `<!-- hidden --> # Heading` dropped the comment and then reparsed `# Heading` as a fresh heading block.

Fixed in c2bfb425c: the trailing spaces after `-->` must now be terminated by a line ending or EOF (`alt((value((), parse_line_ending), value((), eof)))`). When other content follows on the same line the block parser fails, so the line falls through to inline parsing, where the comment is stripped inline and the remainder stays literal text.

Added a regression test (`test_parse_html_comment_with_trailing_same_line_content_is_not_block`): `<!-- hidden --> # Heading` now yields a single `Line([" # Heading"])` instead of a `Heading`. Full suite green (148/148), fmt/clippy clean.

kevinyang372 (approved):

PR #14014, opened 2026-07-20
Open OSC 8 hyperlinks via Cmd/Ctrl+click inside mouse-reporting TUIs

### LINKED ISSUE #13965: Cmd+click doesn't open OSC 8 hyperlinks inside full-screen TUIs (mouse reporting) on macOS
## Summary

On macOS, when a full-screen TUI enables mouse reporting, Warp forwards **all** mouse clicks to the app — there is no `Cmd`/`Ctrl` modifier bypass — so OSC 8 hyperlinks (including `file://`) that the TUI emits cannot be opened with Cmd+click. Only `Shift` currently bypasses mouse capture (for selection).

## Details

Warp already parses and stores OSC 8 hyperlinks as grid-cell attributes and can resolve the URI at a point (`crates/warp_tui/src/grid_handler.rs`, `hyperlink_at_point`), independent of the block-based path detection — so the link data exists even inside the alternate screen.

The gap is in click routing:

- `app/src/terminal/alt_screen/mod.rs::should_intercept_mouse(model, shift, ctx)` decides whether Warp keeps the click or forwards it to the PTY. It intercepts only when `is_reader() || shift`; otherwise, with `SGR_MOUSE` + mouse tracking + `mouse_reporting_enabled`, it forwards.
- The click is dispatched to `TerminalAction::ClickOnGrid` **with the full modifier state including `Cmd`** (`app/src/terminal/alt_screen/alt_screen_element.rs`), but the interception decision is passed only `shift` — so `Cmd`/`Ctrl` are discarded and Cmd+click behaves like a plain click that goes to the app.

## Impact

Hyperlinks emitted by TUIs (lazygit, `gh`, and others) aren't clickable inside the app even though Warp knows about them. iTerm2 and Kitty open modifier-click links even while mouse reporting is active.

## Proposed fix

Intercept `Cmd`(macOS)/`Ctrl`(other-platform) + click when a hyperlink exists at that cell (open it via the existing link-open path), otherwise forward the click to the app unchanged so plain Cmd+click behavior in TUIs isn't affected. The existing `Shift` interception is preserved.

This is the mouse-side cousin of the keyboard modifier-handling gap in #9159: in both, a macOS modifier that is present in the event is not consulted at the TUI-mode input gate.

## Out of scope

Making **plain-text** paths (not OSC-8-tagged) clickable inside a TUI — that remains the emitting app's responsibility (emit OSC 8). This issue is specifically the modifier bypass for hyperlinks that already exist.

## Environment

macOS (Apple Silicon), built from source at current `master`.

FreddieMcHeart:
# Open OSC 8 hyperlinks via Cmd/Ctrl+click inside mouse-reporting TUIs

Fixes #13965. Cousin of the keyboard-side modifier gap in #9159.

## Problem

When a full-screen TUI enables mouse reporting, Warp forwards every click to the PTY, so OSC 8
hyperlinks (including `file://`) the TUI emits cannot be opened with Cmd/Ctrl+click. Only `Shift`
bypasses mouse capture today. iTerm2 and Kitty open modifier+click links even under mouse reporting.

## Root cause

Two gaps combined:

1. The alt-screen click path never resolves or opens a link. `ClickOnGrid` routes to
   `TerminalView::click_on_grid` → `maybe_open_link`, which reads `self.highlighted_link` — a
   **main-screen** hover field the alt screen never populates (it keeps its own hover state). So an
   alt-screen Cmd+click found no link to open.
2. The alt-screen mouse-interception gate (`should_intercept_mouse`) only consults `shift`, so a
   Cmd+click is forwarded to the app like any other click.

## Changes

- **`app/src/terminal/view.rs`** (`maybe_open_link`): for a `WithinModel::AltScreen` position,
  resolve the OSC 8 hyperlink at the clicked point via `TerminalModel::hyperlink_at_point` and open
  it through the existing `open_hyperlink_uri` (which validates the URI before `ctx.open_url`).
- **`app/src/terminal/alt_screen/alt_screen_element.rs`** (`mouse_up`): when a Cmd (macOS) / Ctrl
  (other) + click will open a link, skip forwarding the click to the TUI so the app doesn't also
  swallow it. `Shift` selection and plain-click forwarding are unchanged; `should_intercept_mouse`'s
  signature is untouched.

Behavior table (mouse reporting active): plain click → forward; Shift+click → select; Cmd/Ctrl+click
on a link → **open, don't forward**; Cmd/Ctrl+click on a non-link → forward (apps that use
modifier+click keep working).

## Dependency: `OscHyperlinks` feature flag

OSC 8 hyperlink storage/lookup is gated behind the `OscHyperlinks` feature flag, which is currently
**dogfood-only** (`DOGFOOD_FLAGS`) — not in preview/stable. With the flag off, `hyperlink_at_point`
returns `None` and this path is inert (as is Warp's existing OSC 8 support). This fix presupposes
the flag is enabled; it may be worth graduating `OscHyperlinks` toward preview/stable.

## Testing

No fast unit harness exists for the alt-screen mouse element, so verification is compile + manual:

- Compiles clean (`cargo build --bin warp-oss --profile release-lto`), no new warnings.
- **Verified live in a from-source GUI build** (`WarpOss.app`, with `OscHyperlinks` enabled): inside
  a simulated full-screen TUI (alt screen + mouse reporting on) showing an OSC 8 `file://` link,
  Cmd+click opens the file while a plain click is still delivered to the app.

### DIFF
--- a/app/src/terminal/alt_screen/alt_screen_element.rs
+++ b/app/src/terminal/alt_screen/alt_screen_element.rs
@@ -385,20 +385,30 @@ impl AltScreenElement {
         }
 
         let point = self.coord_to_point(local_position);
+        let modifiers = *mouse_state.modifiers();
+        let position = WithinModel::AltScreen(Point {
+            col: point.col,
+            row: point.row,
+        });
+
+        // Cmd (macOS) / Ctrl (other) + click on an OSC 8 hyperlink opens it (handled by the
+        // ClickOnGrid action). When it will, don't also forward the click to the TUI even though
+        // mouse reporting is on — otherwise the app swallows the click alongside opening the link.
+        let opening_link = crate::terminal::links::should_directly_open_link(&modifiers)
+            && self.model.lock().hyperlink_at_point(&position).is_some();
 
         ctx.dispatch_typed_action(TerminalAction::ClickOnGrid {
-            position: WithinModel::AltScreen(Point {
-                col: point.col,
-                row: point.row,
-            }),
-            modifiers: *mouse_state.modifiers(),
+            position,
+            modifiers,
         });
 
         if self.is_terminal_selecting {
             ctx.dispatch_typed_action(TerminalAction::AltSelect(SelectAction::End));
         }
 
-        if !should_intercept_mouse(&self.model.lock(), mouse_state.modifiers().shift, app) {
+        if !opening_link
+            && !should_intercept_mouse(&self.model.lock(), modifiers.shift, app)
+        {
             ctx.dispatch_typed_action(TerminalAction::AltMouseAction(mouse_state.set_point(point)));
         }
 
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -18318,8 +18318,10 @@ impl TerminalView {
                 .map(|(_, uri)| uri);
             if let Some(uri) = uri {
                 self.open_hyperlink_uri(&uri, ctx);
+                return;
             }
-            return;
+            // No OSC 8 hyperlink at the clicked cell: fall through to the `highlighted_link`
+            // handler below so a regex-detected URL still opens on Cmd/Ctrl+click.
         }
 
         let Some(link) = self.highlighted_link.as_ref() else {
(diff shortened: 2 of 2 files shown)


FreddieMcHeart:
@cla-bot check

PR #14033, opened 2026-07-20
Force-close dangling prompt capture when a command starts

### LINKED ISSUE #13043: Warp shell integration broken by Intelligent Terminal
### Pre-submit Checks

### Describe the bug

After updating Warp, terminal output is no longer rendered.

Commands such as ls, pwd, and echo hello execute successfully, but nothing is displayed in the terminal.

The shell itself is still working. I can navigate directories, edit files with vim, and execute commands, but no command output is visible.

Disabling all AI features and restarting Warp did not resolve the issue.

### To reproduce

Open Warp.

Run any command, such as:

ls
pwd
echo hello

The command executes, but no output is displayed.

### Expected behavior

Command output should be rendered normally in the terminal.

The issue only occurs in Warp. The same shell works correctly in other terminal applications, which suggests this is a Warp rendering issue rather than a shell or configuration problem.

I also searched for previous reports and found (quite old) issues #3254, #3334, and #4162 describing very similar symptoms. However, I couldn't find any documented workaround, and the suggested fixes (restart/reinstall) did not resolve the problem in my case.

### Operating system (OS)

Windows

10.0.26200

### Shell Version

_No response_

### Current Warp version

v0.2026.06.17.09.49.stable_02

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

_No response_

acarl005:
## Human Remarks [Andy]

I verified locally that this works.

## Description
Fixes WSL/bash sessions rendering completely empty blocks (no command, no output) when a third-party shell integration leaves an unmatched OSC 133;A prompt marker.

Both reporters on #13043 have Microsoft Intelligent Terminal's bash shell integration installed in their WSL distro. That integration prints `OSC 133;D` + `OSC 133;A` from `PROMPT_COMMAND` and appends the matching `OSC 133;B` to `PS1`. Warp's bash bootstrap preserves the user's `PROMPT_COMMAND` (so the `133;A` keeps firing every precmd) but empties `PS1` on every precmd when the Warp prompt is active (`bash_body.sh`) — destroying the `133;B`. The unmatched `133;A` sets `HeaderGrid::receiving_chars_for_prompt = Some(Initial)`, which is only ever cleared by a `133;B`. Since the `delegate!` macro in `Block` consults that flag before block state, the command echo and the entire command output get routed into the invisible prompt grids, so every finished block renders as nothing — while DCS hooks (chips), typing, and the alt screen keep working.

The fix: when a command starts (`Block::start()`), defensively force-close prompt-capture mode. The prompt is necessarily done once a command starts, so any capture still open at that point is dangling. This is the same defensive pattern as `end_in_band_command_output` in `PtyController::write_command`. A `log::warn!` records when it fires, for diagnosability.

`Block::start()` is the single choke point all execution paths funnel through (user submit, preexec-without-start fallback, and completion-without-preexec via `ensure_started_for_preexec`), and it fires before the command echo arrives, so both the command text and output are recovered.

## Linked Issue
Fixes #13043

## Testing
Added `test_dangling_start_prompt_marker_is_cleared_at_command_start`, which reproduces the poisoned state (a `StartPrompt` marker with no matching `EndPrompt`), starts a command, and asserts that capture is force-closed, the command echo lands in the prompt-and-command grid, and the command output lands in the output grid.

```
cargo nextest run -p warp -E 'test(dangling_start_prompt_marker)'   # 1 passed
```

## Agent Mode

[Conversation](https://staging.warp.dev/conversation/a394e3fc-d483-40d6-9287-3a2753801536)

### DIFF
--- a/app/src/terminal/model/block.rs
+++ b/app/src/terminal/model/block.rs
@@ -1200,6 +1200,16 @@ impl Block {
             self.start_ts = Some(Local::now());
         }
 
+        // Defensively end prompt-capture mode if it is still active.
+        if let Some(kind) = self.header_grid.receiving_chars_for_prompt.take() {
+            log::warn!(
+                "Prompt capture was still active when command execution began; force-ending it \
+                 so command echo/output are not routed into the prompt grid. An OSC 133;A \
+                 start-prompt marker was likely never followed by a matching OSC 133;B end \
+                 marker. kind={kind:?}"
+            );
+        }
+
         self.header_grid.start_command_grid();
         self.wakeup_after_delay();
     }
--- a/app/src/terminal/model/block.rs
+++ b/app/src/terminal/model/block.rs
@@ -1200,16 +1200,6 @@ impl Block {
             self.start_ts = Some(Local::now());
         }
 
-        // Defensively end prompt-capture mode if it is still active.
-        if let Some(kind) = self.header_grid.receiving_chars_for_prompt.take() {
-            log::warn!(
-                "Prompt capture was still active when command execution began; force-ending it \
-                 so command echo/output are not routed into the prompt grid. An OSC 133;A \
-                 start-prompt marker was likely never followed by a matching OSC 133;B end \
-                 marker. kind={kind:?}"
-            );
-        }
-
         self.header_grid.start_command_grid();
         self.wakeup_after_delay();
     }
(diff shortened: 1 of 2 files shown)


app/src/terminal/model/block.rs
```diff
@@ -1200,6 +1200,16 @@ impl Block {
             self.start_ts = Some(Local::now());
         }
 
+        // Defensively end prompt-capture mode if it is still active.
```
acarl005: Prompt capture (`HeaderGrid::receiving_chars_for_prompt`) is turned on by an OSC `133;A` start-prompt marker and turned off by the matching OSC `133;B` end-prompt marker. Third-party shell integrations can emit a start marker whose end marker never reaches Warp — e.g. an integration that prints OSC `133;A` from `PROMPT_COMMAND` while appending the matching OSC `133;B` to `PS1`, which Warp's bash bootstrap empties on every precmd when the Warp prompt is active. If capture were left on, `delegate!` would route all subsequent PTY output (the command echo and the entire command output) into the invisible prompt grids. The prompt is necessarily done once a command starts, so we force-close capture here.

advait-m: hmm, re the last sentence here — `start()` fires at *write* time, not when the shell accepts the command, so I think on a laggy session capture can be open here and we'd force-close mid-prompt (tail leaks into the command region; the late `133;B` hits the end-without-start fallback in `BlockList::prompt_marker`, which mostly cleans it up). 

likely fine to force-close in this case - should be very rare

advait-m (approved):
nice fix! 🔥 

small non-blocking comment re a possible race, but I don't think we need to realistically worry about

PR #14039, opened 2026-07-20
Use scoped feature-flag overrides in view_tests instead of global set_enabled

### LINKED ISSUE #13972: terminal::view: 5 tests fail in full-suite runs due to leaked global FeatureFlag state (AgentView never reset)
## Summary

Five `terminal::view::tests` pass in isolation but fail whenever the module (or the full `warp` lib suite) runs, because other tests in `view_tests.rs` flip the process-global `FeatureFlag` state via `FeatureFlag::AgentView.set_enabled(true)` and never reset it. The five failing tests do no flag setup and implicitly rely on `AgentView` defaulting to `false`; once an earlier test leaves the global `true`, they take the wrong code branches and their assertions fail. These currently report as failures on every contributor's full-suite PR gate.

## Affected tests

All in `app/src/terminal/view_tests.rs`:

- `terminal::view::tests::test_insert`
- `terminal::view::tests::test_insert_into_input`
- `terminal::view::tests::test_viewport_iter_most_recent_at_bottom`
- `terminal::view::tests::test_reinput_blocks`
- `terminal::view::tests::test_scroll_position_doesnt_change_when_block_finished`

## Reproduction

Full suite (fails; the 5 appear among 12-14 failures depending on run):

```
cargo test -p warp --lib
```

The 5 tests pass when run together with no flag-setting test in the set:

```
cargo test -p warp --lib -- --exact \
  terminal::view::tests::test_insert \
  terminal::view::tests::test_insert_into_input \
  terminal::view::tests::test_viewport_iter_most_recent_at_bottom \
  terminal::view::tests::test_reinput_blocks \
  terminal::view::tests::test_scroll_position_doesnt_change_when_block_finished
# => test result: ok. 5 passed
```

Minimal deterministic reproduction — run a flag-setting test first, single-threaded, and the clean tests fail every time:

```
cargo test -p warp --lib -- --test-threads=1 --exact \
  terminal::view::tests::agent_view_lifecycle_updates_input_mode \
  terminal::view::tests::test_insert \
  terminal::view::tests::test_viewport_iter_most_recent_at_bottom \
  terminal::view::tests::test_scroll_position_doesnt_change_when_block_finished
# => test_insert FAILED, test_viewport_iter_most_recent_at_bottom FAILED,
#    test_scroll_position_doesnt_change_when_block_finished FAILED
```

`--test-threads=1` does not help: this is process-global state carried forward across sequential tests, not a data race.

Observed assertion failures in the full run:

- `test_insert` (view_tests.rs:2698): `left: "One", right: "None"` — a block selection that should have been cleared is retained.
- `test_insert_into_input` (view_tests.rs:2824): `Expected no selections in the blocklist but got Some(...)`.
- `test_viewport_iter_most_recent_at_bottom` (view_tests.rs:3220): `item 3` — a trailing zero-height viewport entry is missing.
- `test_reinput_blocks` (view_tests.rs:4072): `left: "pwd", right: "sudo pwd"`.
- `test_scroll_position_doesnt_change_when_block_finished` (view_tests.rs:5310): `ScrollTop(...)` vs `FollowsBottomOfMostRecentBlock`.

## Root cause

Feature-flag enabled-state is stored in a process-global array:

```rust
// crates/warp_features/src/lib.rs:918
static FLAG_STATES: [AtomicBool; cardinality::<FeatureFlag>()] =
    [const { AtomicBool::new(false) }; { cardinality::<FeatureFlag>() }];
```

`set_enabled` writes into it globally with no teardown, and its own guard documents that this is the wrong thing to do in a test:

```rust
// crates/warp_features/src/lib.rs:1038
pub fn set_enabled(self, enabled: bool) {
    if cfg!(test) && cfg!(not(feature = "integration_tests")) {
        panic!(
            "Tried to globally enable {self:?} in a test. Use FeatureFlag::{self:?}.override_enabled instead"
        );
    }
    FLAG_STATES[self as usize].store(enabled, Ordering::Relaxed);
}
```

`view_tests.rs` contains 22 `set_enabled(true)` calls (e.g. `AgentView` at lines 148, 737, 1476, 1535, 5141, 5322, 5383, 5443, 5506, 5562, 5608, 5690, 5817, 5865, 5923, 7439, 7485, 7565, 7877, 7944; `CloudMode` at 1477, 1536). None of them reset the flag afterward. Because these tests run under the `integration_tests` feature, the guard's panic is bypassed and the global write silently persists for the rest of the process.

The five failing tests read the flag and branch on it. With `AgentView` leaked `true`:

- `app/src/terminal/view.rs:19985` and `:20028` guard `clear_selected_blocks()` with `&& !FeatureFlag::AgentView.is_enabled()`. Leaked-on means the block selection is not cleared when typing in shell mode, so `test_insert` sees `"One"` selection instead of `"None"`, and `test_insert_into_input` sees a leftover `Some(...)` selection.
- `app/src/terminal/block_list_viewport.rs:2040`: `if !FeatureFlag::AgentView.is_enabled() || block_height.as_f64() > 0. { return next; }`. Leaked-on skips the trailing zero-height entry, so `test_viewport_iter_most_recent_at_bottom` panics on the missing `item 3` and `test_scroll_position_doesnt_change_when_block_finished` computes a different content height / most-recent-block and resolves the scroll position differently.
- `test_reinput_blocks` goes through `reinput_commands` → `with_non_hidden_selected_blocks` (view.rs:20954), whose filter uses the AgentView-dependent `transcript_scope`; the same non-clearing-and-scope divergence drops the `sudo` variant so the input box retains `"pwd"`.

That the same file already defensively uses `AgentView.override_enabled(false)` in places (view_tests.rs:439, 798) is direct evidence the leak is known.

## Fix direction

Replace the 22 global `set_enabled(true)` calls in `view_tests.rs` with the RAII thread-local override the module already uses in a couple of spots:

```rust
let _guard = FeatureFlag::AgentView.override_enabled(true); // reverts on drop
```

`override_enabled` (crates/warp_features/src/lib.rs:1067) writes a thread-local `FLAG_OVERRIDES` and returns an `OverrideGuard` (lib.rs:1192) that clears the override on drop, so no global write leaks across tests. This restores isolation without a per-test reset dance and matches the pattern the guard's own panic message prescribes.

## Related

Related: #13853 (warp_editor render::tests Once-poisoning — same class, different crate) and the sibling flake tickets filed alongside this one: #13973, #13974, #13975

fbartho:
Closes #13972

## Description

[`view_tests.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/terminal/view_tests.rs) had 22 call sites flipping the process-global `FeatureFlag::AgentView`/`FeatureFlag::CloudMode` state via `set_enabled(true)` with no reset. Under plain `cargo test`, every test in the `warp` lib target shares one process, so the flag state leaked into whichever tests ran next — five tests that assume `AgentView` defaults to `false` failed whenever they ran after an offender.

This converts all 22 sites to `override_enabled(true)`, the thread-local RAII guard the same module already uses elsewhere; the override reverts on drop at the end of each test. Thread-locality is sufficient: `App::test`'s executor is a single-threaded `LocalExecutor`, and none of the converted tests spawn OS threads (so no `get_overrides()`/`set_overrides()` propagation is needed). Two tests set both flags and bind both guards for the full test body. `override_enabled` itself is unchanged.

## Why CI didn't catch this

CI runs `cargo nextest`, which isolates each test in its own process — the leak needs a shared-process runner to manifest.

The `set_enabled` panic guard that should have flagged the misuse ([`warp_features/src/lib.rs`](https://github.com/warpdotdev/warp/blob/master/crates/warp_features/src/lib.rs): `if cfg!(test) && cfg!(not(feature = "integration_tests")) { panic!(…) }`) is dead code by construction: `cfg!(test)` resolves per-crate at the point it's written, and `warp_features` is a normal `[dependencies]` library that's never compiled as a test binary — so the check is unconditionally `false` for every caller.

<details><summary>Ruled out along the way: feature unification</summary>

The issue originally theorized that `crates/integration` enabling `integration_tests` defeats the guard via Cargo feature unification. Verified false: [`warp_core/Cargo.toml`](https://github.com/warpdotdev/warp/blob/master/crates/warp_core/Cargo.toml)'s `integration_tests` feature is `[]` and never forwards to `warp_features/integration_tests`, and `app` doesn't depend on `warp_features` directly — so that feature can't be reached from a normal `app`/`warp` build at all. The guard's deadness is independent of feature flags.

</details>

## Reproduction

Deterministic single-process repro — a flag-setting test followed by affected tests:

```
cargo test -p warp --lib -- --test-threads=1 --exact \
  terminal::view::tests::agent_view_lifecycle_updates_input_mode \
  terminal::view::tests::test_insert \
  terminal::view::tests::test_scroll_position_doesnt_change_when_block_finished \
  terminal::view::tests::test_viewport_iter_most_recent_at_bottom
```

Before: `FAILED. 1 passed; 3 failed` (assertions like ``left: "One" / right: "None"``, scroll-position and viewport-iter panics — matching the issue's reported failures). After: all originally-affected tests plus the flag-setter pass together, `ok. 6 passed; 0 failed`.

`cargo nextest run -p warp -E 'test(view_tests) or test(terminal::view)'`: 307 passed before and after (process isolation always masked this).

## Preventing recurrence

I looked at catching this class mechanically going forward. Verdicts:

| Option | Verdict | Why |
|---|---|---|
| Clippy `disallowed-methods` on `set_enabled` (repo already has `.clippy.toml` + `-D warnings` CI) | Not implemented | ~74 legitimate call sites (≈26 production, ≈48 `crates/integration` tests that want real global state) would each need `#[allow(clippy::disallowed_methods)]` — and every allow looks identical to a reintroduced leak, burying the signal |
| Repair the `cfg!(test)` panic guard | Not implemented — maintainer call | Dead by construction (above). A real fix needs a runtime "in test" signal or a test-only-constructible token (≈ re-deriving `override_enabled` for the global case) — design work, and it inherits the same production/integration scoping problem |
| Grep-style check in `./script/presubmit` | Noted only | Cruder duplicate of `disallowed-methods`; only worth it if the lint is rejected specifically on annotation-noise grounds |

## Testing

- [x] `cargo test -p warp --lib` targeted repro (before/after above)
- [x] `cargo nextest run -p warp -E 'test(view_tests) or test(terminal::view)'` — 307 passed
- [x] `./script/format --check` — clean
- [x] `cargo clippy -p warp --lib --tests` — zero warnings

## Agent Mode

### DIFF
--- a/app/src/terminal/view_tests.rs
+++ b/app/src/terminal/view_tests.rs
@@ -145,7 +145,7 @@ fn has_pending_user_query_block(view: &TerminalView) -> bool {
 fn agent_view_lifecycle_updates_input_mode() {
     App::test((), |mut app| async move {
         initialize_app_for_terminal_view(&mut app);
-        FeatureFlag::AgentView.set_enabled(true);
+        let _agent_view = FeatureFlag::AgentView.override_enabled(true);
         let terminal = add_window_with_terminal(&mut app, None);
 
         terminal.read(&app, |view, ctx| {
@@ -736,7 +736,7 @@ fn unregister_cli_agent_session_restores_unlocked_input_config() {
 fn clear_buffer_action_in_fullscreen_agent_view_starts_new_conversation() {
     App::test((), |mut app| async move {
         initialize_app_for_terminal_view(&mut app);
-        FeatureFlag::AgentView.set_enabled(true);
+        let _agent_view = FeatureFlag::AgentView.override_enabled(true);
 
         let terminal = add_window_with_terminal(&mut app, None);
 
@@ -1477,8 +1477,8 @@ fn root_cloud_mode_pane_sets_root_cloud_mode_context_key() {
     App::test((), |mut app| async move {
         initialize_app_for_terminal_view(&mut app);
         app.add_singleton_model(ImportedConfigModel::new);
-        FeatureFlag::AgentView.set_enabled(true);
-        FeatureFlag::CloudMode.set_enabled(true);
+        let _agent_view = FeatureFlag::AgentView.override_enabled(true);
+        let _cloud_mode = FeatureFlag::CloudMode.override_enabled(true);
 
         let terminal = add_window_with_cloud_mode_terminal(&mut app);
         let nested_terminal = add_window_with_cloud_mode_terminal(&mut app);
@@ -1540,8 +1540,8 @@ fn root_cloud_mode_pane_sets_root_cloud_mode_context_key() {
 fn set_input_mode_agent_does_not_enter_local_agent_from_root_cloud_mode_pane() {
     App::test((), |mut app| async move {
         initialize_app_for_terminal_view(&mut app);
-        FeatureFlag::AgentView.set_enabled(true);
-        FeatureFlag::CloudMode.set_enabled(true);
+        let _agent_view = FeatureFlag::AgentView.override_enabled(true);
+        let _cloud_mode = FeatureFlag::CloudMode.override_enabled(true);
 
         let terminal = add_window_with_cloud_mode_terminal(&mut app);
 
@@ -5154,7 +5154,7 @@ fn test_prompt_context_menu_items_for_agent_toolbelt_flag() {
 fn agent_footer_updates_chip_groups_when_side_assignment_changes() {
     App::test((), |mut app| async move {
         initialize_app_for_terminal_view(&mut app);
-        FeatureFlag::AgentView.set_enabled(true);
+        let _agent_view = FeatureFlag::AgentView.override_enabled(true);
 
         let terminal = add_window_with_terminal(&mut app, None);
 
@@ -5335,7 +5335,7 @@ fn test_scroll_position_doesnt_change_when_block_finished() {
 fn inline_agent_view_exits_when_tagged_in_long_running_command_is_tagged_out() {
     App::test((), |mut app| async move {
         initialize_app_for_terminal_view(&mut app);
-        FeatureFlag::AgentView.set_enabled(true);
+        let _agent_view = FeatureFlag::AgentView.override_enabled(true);
 
         let terminal = add_window_with_terminal(&mut app, None);
 
@@ -5397,7 +5397,7 @@ fn inline_agent_view_exits_when_tagged_in_long_running_command_is_tagged_out() {
 fn ctrl_c_after_stop_takeover_cancels_conversation() {
     App::test((), |mut app| async move {
         initialize_app_for_terminal_view(&mut app);
-        FeatureFlag::AgentView.set_enabled(true);
+        let _agent_view = FeatureFlag::AgentView.override_enabled(true);
 
         let terminal = add_window_with_terminal(&mut app, None);
         let pty_writes: Rc<RefCell<Vec<Vec<u8>>>> = Rc::new(RefCell::new(Vec::new()));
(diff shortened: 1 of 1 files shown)


fbartho:

fbartho:

fbartho:
This PR is associated with a ticket, but the ticket wasn’t in the right status yet. The ticket hasn’t been reviewed yet to mark it as `ready-for-spec` or `ready-to-implement` so this PR is stuck in limbo until that happens.

I’m not sure what to do here: @oss-maintainers

fbartho:
Temporarily closing the PR until the issue gets processed.

PR #14040, opened 2026-07-20
Isolate secret-redaction tests from the shared SECRETS_REGEX global

### LINKED ISSUE #13973: terminal secrets: redaction tests race on the shared process-global SECRETS_REGEX, failing nondeterministically in parallel runs
## Summary

The secret-redaction tests each install their own regex set into a single process-global `SECRETS_REGEX` and then assert on the resulting detections. Run in parallel (or sequentially without teardown), one test overwrites another's regexes before its assertion runs, so the loser detects zero secrets instead of one and panics with `left: 0, right: 1`. Which member fails rotates run-to-run — the classic parallel-isolation signature — and all members pass in isolation. These currently report as failures on every contributor's full-suite PR gate.

## Affected tests

Primary (`app/src/terminal/model/grid/secrets_tests.rs`):

- `terminal::model::grid::grid_handler::secrets::tests::test_secret_redacted_after_byte_processing` (:34)
- `terminal::model::grid::grid_handler::secrets::tests::test_secret_redacted_after_multiple_byte_processing`
- `terminal::model::grid::grid_handler::secrets::tests::test_secret_redaction_secret_remains_after_resize` (:249)
- `terminal::model::grid::grid_handler::secrets::tests::test_secret_redaction_unobfuscated_secret_remains_after_byte_processing`
- `terminal::model::grid::grid_handler::secrets::tests::test_bytes_processed_for_secrets_after_turning_redaction` and the `_on` / `_off_then_on` / `_on_then_off` variants (:270, :301, :327, :362)

Secondary consumer of the same global, observed failing under full-suite load:

- `ai::blocklist::block::secret_redaction::tests::test_detect_secrets_no_regexes_configured`

Other modules that also write the same global (relevant to serialization scope): `app/src/ai/blocklist/block/secret_redaction_tests.rs`, `app/src/ai/blocklist/action_model/execute/grep_tests.rs`, `app/src/server/telemetry/secret_redaction.rs`.

## Reproduction

Full suite — the secrets member(s) that fail vary between runs:

```
cargo test -p warp --lib
# run A: fails test_secret_redaction_secret_remains_after_resize
# run B: fails test_secret_redacted_after_byte_processing (+ ai::blocklist ... test_detect_secrets_no_regexes_configured)
# run C: no secrets failures
```

The module self-pollutes even alone (5 pass / 5 fail):

```
cargo test -p warp --lib terminal::model::grid::grid_handler::secrets
# => test result: FAILED. 5 passed; 5 failed
```

Every named test passes in isolation:

```
cargo test -p warp --lib -- --exact terminal::model::grid::grid_handler::secrets::tests::test_secret_redacted_after_byte_processing
# => ok. 1 passed
```

`--test-threads=1` does not fully fix it — the tests still share the same global with no teardown, so ordering alone can still overwrite regexes between sequential tests.

Observed panic (identical across members):

```
thread '…::test_secret_redacted_after_byte_processing' panicked at app/src/terminal/model/grid/secrets_tests.rs:50:5:
  left: 0
 right: 1
```

## Root cause

There is exactly one process-global regex registry for secret detection:

```rust
// app/src/terminal/model/secrets.rs:54
pub static ref SECRETS_REGEX: Mutex<Arc<SecretsRegex>> = Mutex::new(
    Arc::new(SecretsRegex {
        regex: regex_automata::meta::Regex::new_many(&[] as &[&str])...,
        ...
    })
);
```

Every test configures its own regexes into this single global via `set_user_and_enterprise_secret_regexes` (secrets.rs:351), e.g.:

```rust
// app/src/terminal/model/grid/secrets_tests.rs:35
crate::terminal::model::secrets::set_user_and_enterprise_secret_regexes(
    [&Regex::new("ABCD").expect("...")],
    std::iter::empty(),
);
...
assert_eq!(grid_handler.secrets.len(), 1);   // :50
```

The detection path reads the same global:

```rust
// app/src/ai/blocklist/block/secret_redaction.rs:26
let secrets_regex: Arc<SecretsRegex> = { SECRETS_REGEX.lock().clone() };
```

When two secret tests run concurrently (or sequentially without reset), test B's `set_user_and_enterprise_secret_regexes(...)` replaces test A's regexes — or resets toward a set that doesn't match A's input — before A's `assert_eq!(secrets.len(), 1)` runs. A then detects 0 secrets → `left: 0, right: 1`. Because the ordering that decides the loser varies, a different member fails each run, and sometimes none do.

Note on the `_turning_redaction*` family: those tests additionally call `grid_handler.obfuscate_secrets(...)` (e.g. secrets_tests.rs:291), but obfuscation mode is per-`GridHandler` local state, not the shared global. The regex registry is the sole cross-test culprit — the "redaction toggle" in the test name should not be read as implying a redaction-mode global; there isn't one.

## Fix direction

Serialize the secret tests into one group and give the global proper per-test scoping:

- Mark the secret-redaction tests (across `grid/secrets_tests.rs`, `ai/blocklist/block/secret_redaction_tests.rs`, `ai/blocklist/action_model/execute/grep_tests.rs`, `server/telemetry/secret_redaction.rs`) with a shared `#[serial]` group so they cannot interleave on the global.
- Additionally, either save and restore `SECRETS_REGEX` around each test (snapshot the `Arc` on entry, restore on drop via an RAII guard) or add a test-scoped override analogous to the feature-flag `override_enabled` pattern, so serialization is not the only line of defense.

Serialization alone reduces flakiness but leaves the tests coupled through the global; the save/restore (or scoped override) is what actually isolates them.

## Related

Related: #13853 (warp_editor render::tests Once-poisoning — same class, different crate) and the sibling flake tickets filed alongside this one: #13972, #13974, #13975

fbartho:
Closes #13973

## Description

The secret-redaction tests install their own regexes into the single process-global `SECRETS_REGEX` ([`secrets.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/terminal/model/secrets.rs)) with no isolation. Under plain `cargo test`, one test's regexes could overwrite another's before its assertion ran — the loser panics `left: 0, right: 1`, and which test loses rotates run to run, which is why this reported as a rotating flake rather than one consistent failure.

Fix: `test_set_secrets_regex()` returns a `#[must_use]` RAII guard (modeled on the existing `FeatureFlag::override_enabled` → `OverrideGuard` pattern) that snapshots the global and restores it on drop. Since the global is process-wide (not thread-local), the guard prevents cross-test leakage but not mid-test interleaving — so the affected tests are also `#[serial(secrets_regex)]`, a named group that avoids serializing against unrelated `#[serial]` tests. Converted all racing call sites in the three files from the issue: [`grid/secrets_tests.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/terminal/model/grid/secrets_tests.rs) (9 tests), [`block/secret_redaction_tests.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/ai/blocklist/block/secret_redaction_tests.rs), and [`execute/grep_tests.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/ai/blocklist/action_model/execute/grep_tests.rs).

## Linked Issue

  - _Not applicable: test-only change, no user-visible behavior._

## Testing

Repeated invocations of the affected modules under plain `cargo test` — the harness where the flake manifests (nextest's per-process isolation masks it):

- Before: `cargo test -p warp --lib terminal::model::grid::grid_handler::secrets` → `FAILED. 3 passed; 7 failed` (e.g. `assertion left == right failed: left: 0, right: 1`).
- After: same command ×3 → `ok. 10 passed` every run. The other two modules: `ai::blocklist::block::secret_redaction` → `ok. 19 passed` ×3; `…execute::grep` → `ok. 4 passed` ×3.
- Nextest (CI parity, was already green): `cargo nextest run -p warp --lib -E 'test(secrets) + test(secret_redaction) + test(grep_tests)'` → 72/72 passed.
- `cargo clippy -p warp --lib --tests` and `cargo check -p warp --lib --features test-util`: zero warnings.

### Screenshots / Videos

_Not applicable: test-only change._

## Agent Mode

### DIFF
--- a/app/src/ai/blocklist/action_model/execute/grep_tests.rs
+++ b/app/src/ai/blocklist/action_model/execute/grep_tests.rs
@@ -1,10 +1,13 @@
+use serial_test::serial;
+
 use super::*;
 use crate::terminal::model::secrets::regexes::FIREBASE_AUTH_DOMAIN;
 use crate::terminal::shell::ShellType;
 
 #[test]
+#[serial(secrets_regex)]
 fn test_create_redacted_grep_error_event() {
-    crate::terminal::model::set_user_and_enterprise_secret_regexes(
+    let _secrets_regex_guard = crate::terminal::model::secrets::test_set_secrets_regex(
         [&regex::Regex::new(FIREBASE_AUTH_DOMAIN).expect("Should be able to construct regex")],
         std::iter::empty(), // No enterprise secrets
     );
--- a/app/src/ai/blocklist/block/secret_redaction_tests.rs
+++ b/app/src/ai/blocklist/block/secret_redaction_tests.rs
@@ -307,21 +307,31 @@ fn test_merge_ranges_with_same_end() {
 // rather than system default regexes.
 
 #[test]
+#[serial(secrets_regex)]
 fn test_detect_secrets_no_regexes_configured() {
+    // Explicitly clear the global regex registry rather than relying on its initial state, since
+    // other tests in this serial group leave their own regexes installed until their guard drops.
+    let _secrets_regex_guard =
+        secrets::test_set_secrets_regex(std::iter::empty(), std::iter::empty());
+
     // With no regexes configured, no secrets should be detected
     let text = "foo warp-server-staging.firebaseapp.com bar";
     let detected_secrets = find_secrets_in_text(text);
     assert_eq!(detected_secrets, vec![]);
 }
 
-// #[serial] is used to ensure custom regexes state does not interfere with other tests,
-// as the custom regexes are global state.
+// #[serial(secrets_regex)] is used to ensure custom regexes state does not interfere with other
+// tests, as the custom regexes are global state. A named group is used (rather than the crate's
+// default `#[serial]` group) so these tests don't needlessly serialize against unrelated tests
+// elsewhere in the crate that also use `#[serial]`. The guard returned by `test_set_secrets_regex`
+// restores the previous regexes on drop, so tests are also isolated from state left behind by a
+// previous test — serialization alone only prevents concurrent interleaving, not leakage.
 
 #[test]
-#[serial]
+#[serial(secrets_regex)]
 fn test_detect_secrets_single_secret_custom() {
     // Set as user secret (enterprise secrets is empty)
-    secrets::set_user_and_enterprise_secret_regexes(
+    let _secrets_regex_guard = secrets::test_set_secrets_regex(
         [&Regex::new("ABCD").expect("Should be able to construct regex")],
         std::iter::empty(), // No enterprise secrets
     );
@@ -338,11 +348,11 @@ fn test_detect_secrets_single_secret_custom() {
 }
 
 #[test]
-#[serial]
+#[serial(secrets_regex)]
 fn test_detect_secrets_single_secret_custom_with_multibyte() {
     // Set a custom secret regex that matches a Chinese multibyte secret, e.g., "秘密"
     // Set as user secret (enterprise secrets is empty)
-    secrets::set_user_and_enterprise_secret_regexes(
+    let _secrets_regex_guard = secrets::test_set_secrets_regex(
         [&Regex::new("秘密").expect("Should be able to construct regex")],
         std::iter::empty(), // No enterprise secrets
     );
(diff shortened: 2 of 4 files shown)


fbartho:

fbartho:

fbartho:
This PR is associated with a ticket, but the ticket is stuck waiting for manual review, so the PR is in limbo. 

Please advise what you want me to do here, thanks @oss-maintainers?

fbartho:
Temporarily closing this until the linked issue #13973 gets processed.

Issue #13975, opened 2026-07-19
workspace::view: test_terminal_model_isnt_leaked fails under full-suite load — detached PTY reader thread outlives tab close


fbartho:
## Summary

`workspace::view::tests::test_terminal_model_isnt_leaked` asserts that a terminal model's `Arc` is fully dropped once its tab closes, checked via `Weak::upgrade()` returning `None`. It passes in isolation but fails under full-parallel suite load because the local-tty "PTY reader" OS thread holds a strong clone of the model `Arc` and is not always joined when the tab closes, so a strong reference lingers process-wide. This is the least deterministic of the flake set — it is a thread-lifecycle/join-ordering issue, not a single global to reset — and I am labeling it honestly as such. It currently reports as a failure on contributors' full-suite PR gate runs.

## Affected tests

- `workspace::view::tests::test_terminal_model_isnt_leaked` — `app/src/workspace/view_tests.rs:1987`

## Reproduction

Full suite (intermittent — fails under parallel load, most consistently in the full run):

```
cargo test -p warp --lib
```

Passes in isolation:

```
cargo test -p warp --lib -- --exact workspace::view::tests::test_terminal_model_isnt_leaked
# => ok. 1 passed
```

`--test-threads=1` does not reliably fix it: the leaking reference is held by a detached OS thread (a "PTY reader"), not a test task, so serializing test tasks does not guarantee the thread has dropped its `Arc` by assertion time.

Observed failure:

```
thread 'workspace::view::tests::test_terminal_model_isnt_leaked' panicked at app/src/workspace/view_tests.rs:1987:9:
The terminal model should not exist once the tab is closed.
```

Accompanying log lines seen in the same run (both relevant to the mechanism):

```
[WARN] warp::terminal::local_tty::terminal_manager: No event loop handle to join when dropping terminal manager.
[WARN] warp::terminal::view: Session abandoned before bootstrap for shell None on ssh false
```

## Root cause

The assertion is a pure `Arc::strong_count` check via `Weak::upgrade()` on the specific model `Arc` — there is no global registry:

```rust
// app/src/workspace/view_tests.rs:1974-1990
let model = workspace.get_active_session_terminal_model(ctx);   // the exact Arc
let terminal_model = Arc::downgrade(&model);
...
remove_tab(...);                                                 // close the tab
...
assert!(terminal_model.upgrade().is_none(), "The terminal model should not exist once the tab is closed.");
```

[...]

PR #14042, opened 2026-07-20
Replace fixed timer waits with condition polling in two flaky async tests

### LINKED ISSUE #13979: app tests: two async tests use fixed real-wall-clock `Timer::after` waits with no retry margin, and fail intermittently under full-parallel PR-gate load
## Summary

Two tests observed failing in full-parallel `cargo test -p warp --lib` PR-gate runs, both verified passing in isolation and in this investigation's two full local reruns (which instead reproduced the four already-tracked clusters, #13972-#13975). Unlike those four clusters, these two are not explained by any process-global static, cache, or leaked feature flag — I checked and neither test's code path touches `FLAG_STATES`, `SECRETS_REGEX`, `EXPERIMENT_LAYER_MAPPINGS`, the `TELEMETRY` queue, `BOOTSTRAP_CACHE`, or a detached PTY thread `Arc`. Both tests instead drive real-time async waits (`warpui::r#async::Timer::after`, backed by `async_io::Timer` — a genuine OS-reactor timer, not a virtual/mocked clock) as their sole synchronization mechanism, with no bounded poll-and-retry loop to absorb scheduler delay. Under full-suite parallel load (5765+ tests contending for 10 cores in this environment), that makes them structurally more exposed to false failures from thread/timer starvation than tests that poll.

I could not obtain a deterministic minimal reproduction (two full local runs of the full `--lib` suite did not reproduce either failure — see below), so this is filed on static-analysis mechanism evidence, not a captured failure, similar in spirit to the honesty caveat in #13975.

## Affected tests

- `server::cloud_objects::update_manager::tests::test_pending_metadata_update_with_polling` — `app/src/server/cloud_objects/update_manager_tests.rs:2601`
- `settings::cloud_preferences_syncer::tests::test_cloud_pref_not_synced_when_current_value_not_syncable` — `app/src/settings/cloud_preferences_syncer_tests.rs:1101`

## Reproduction

Both pass in isolation:

```
cargo test -p warp --lib -- --exact server::cloud_objects::update_manager::tests::test_pending_metadata_update_with_polling
# => ok. 1 passed; finished in 0.03s

cargo test -p warp --lib -- --exact settings::cloud_preferences_syncer::tests::test_cloud_pref_not_synced_when_current_value_not_syncable
# => ok. 1 passed; finished in 1.03s
```

Two full-suite local runs on this 10-core machine did **not** reproduce either failure; both instead reproduced exactly the four already-tracked clusters (12-13 failures matching #13972/#13973/#13974/#13975) plus two unrelated new failures out of this investigation's scope (`notebooks::file::tests::test_load_local`, `notebooks::file::tests::test_load_before_session`, `util::path::tests::test_resolve_command`):

```
cargo test -p warp --lib
# run 1: test result: FAILED. 5765 passed; 12 failed (the 4 tracked clusters' members)
# run 2: test result: FAILED. 5764 passed; 13 failed (same, +1 extra secrets-cluster member)
```

Neither `test_pending_metadata_update_with_polling` nor `test_cloud_pref_not_synced_when_current_value_not_syncable` appeared as a failure in either run. This is consistent with them being rarer, load-threshold-dependent flakes that need either more attempts or a higher core-count / more contended CI runner to trigger — the same category of difficulty documented for #13975's thread-join race, but via a different mechanism (see below).

## Root cause (mechanism, not captured failure)

Both tests rely on a fixed real-time sleep as their only synchronization step, without a bounded retry loop:

```rust
// app/src/server/cloud_objects/update_manager_tests.rs:2577-2581
// await long enough that all the trash object retries are exhausted
warpui::r#async::Timer::after(Duration::from_secs(10)).await;

assert_trashed_status_for_object(&mut app, &sync_id.uid(), false);
assert_pending_status_for_object(&mut app, &sync_id.uid(), false);
```

`Timer` resolves to a real OS-backed timer, not a mocked/virtual clock:

```rust
// crates/warpui_core/src/async/native/mod.rs:3
pub use async_io::{block_on, Timer};
```

The 10s wait is meant to cover the retry backoff for `ONLINE_ONLY_OPERATION_RETRY_STRATEGY` (500ms base, factor 2, 3 retries — worst case ~3.5s), so there is nominally a ~6.5s margin. That margin assumes the test's own async task gets scheduled promptly; under full-parallel load with thousands of concurrent single-threaded `App::test()` executors and OS reactor threads contending for 10 cores, the actual wall-clock time for the awaited work to complete (not just the timer to fire) can exceed the fixed budget, causing the post-wait assertions to observe not-yet-settled state.

By contrast, `update_manager_tests.rs` has no bounded poll/retry helper anywhere in the file — every synchronization point is either `ctx.await_spawned_future(...)` (driven directly by the future, not time-boxed) or a bare `Timer::after(...)`. This is a structurally more fragile pattern than the sibling module:

```rust
// app/src/settings/cloud_preferences_syncer_tests.rs:207-218
async fn wait_for_num_spawned_futures(app: &mut App, expected_num: usize, message: &str) {
    for _ in 0..50 {
        let num_spawned_futures =
            SyncQueue::handle(app).read(app, |sync_queue, _ctx| sync_queue.spawned_futures().len());
        if num_spawned_futures == expected_num {
            return;
        }
        warpui::r#async::Timer::after(Duration::from_millis(100)).await;
    }
    assert_num_spawned_futures(app, expected_num, message);
}
```

`test_cloud_pref_not_synced_when_current_value_not_syncable` (cloud_preferences_syncer_tests.rs:1148-1149) does use this bounded-retry helper (`await_spawned_futures`, which wraps `wait_for_num_spawned_futures`), giving it up to 5s of polled margin — more load-resistant than the update_manager test, but it is preceded by its own fixed `Timer::after(Duration::from_secs(1))` (line 1148) before the poll loop starts, which is itself a real-time wait that could, in principle, still leave assertions checking state that a starved executor hasn't caught up to, if the initial fixed second is not enough for `mock_initial_load` processing to even begin under contention.

Both mechanisms funnel to the same category: a fixed-duration real-time `Timer::after` used as a proxy for "enough time has passed for async work to settle," which is inherently margin-dependent and therefore load-sensitive, as opposed to a bounded poll-until-condition loop (robust) or a virtual/mocked clock (immune to real scheduling delay).

## Fix direction

- For `update_manager_tests.rs`: replace the bare `Timer::after(Duration::from_secs(10))` "wait long enough" pattern with a bounded poll-until-condition loop (analogous to `wait_for_num_spawned_futures` in `cloud_preferences_syncer_tests.rs`) that polls `assert_trashed_status_for_object`/`assert_pending_status_for_object`'s underlying state until it settles or a generous timeout elapses, rather than asserting immediately after a fixed sleep.
- For `cloud_preferences_syncer_tests.rs`: consider whether the leading fixed `Timer::after(Duration::from_secs(1))` before `await_spawned_futures` can be folded into the existing poll loop (e.g. by increasing its retry budget) rather than kept as a separate ungated wait.
- Longer term, consider whether `App::test()` should support a virtual/mocked time source for these test executors (many async runtimes offer this, e.g. tokio's `time::pause()`) so tests expressing "wait N seconds for backoff to exhaust" don't depend on real wall-clock scheduling at all. This would eliminate the whole class of load-sensitive timing flakes, of which this ticket's two tests may just be the currently-observed members.

## Related

Related: #13853 (warp_editor render::tests Once-poisoning — same class, different crate) and the sibling flake tickets #13972, #13973, #13974, #13975. Unlike those four (cumulative process-global state, deterministically reproducible via ordering), this ticket's mechanism is real-time scheduling margin under parallel load — closer in spirit to #13975's "load-dependent, not a single static to reset" caveat, but via fixed-sleep synchronization rather than a lingering thread `Arc`.

**Observation provenance:** both tests were observed failing (and verified pass-in-isolation) during full-parallel gate runs on a sibling feature branch earlier today, on the same 10-core hardware — the non-reproduction above is about the two dedicated attempts on master, not the original sightings.

fbartho:
Closes #13979

## Mechanism

Two tests used a fixed real-time `Timer::after` wait as their sole synchronization step before asserting on async state. `Timer` is a real OS-backed timer (`async_io::Timer`), not a virtual clock — under full-parallel suite load, the awaited async work can take longer than the fixed sleep budget even though the timer itself fires on schedule. That makes the failures load-sensitive and intermittent: they depend on contention at run time, not process-global state (unlike the sibling flake clusters #13972–#13975, which are all global-state races).

## Correction to the ticket

The ticket names `test_pending_metadata_update_with_polling` at line 2601 — on current master that function contains no `Timer::after`; line 2601 is where its signature *begins*. The quoted 10s wait lives at line 2578 inside the immediately-preceding test, `test_metadata_after_trash_item_failure`, whose code matches the ticket's described mechanism and snippet exactly. This PR fixes that test. (Also noted on the issue.)

## Changes

- **`test_metadata_after_trash_item_failure`** ([`update_manager_tests.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/server/cloud_objects/update_manager_tests.rs)): the fixed 10s "wait for retries to exhaust" sleep is now `wait_for_trashed_status_for_object()`: a condition poll on `metadata().trashed_ts` wrapped in the repo's own `with_timeout` primitive with a named 30s `Duration` ceiling — the deadline is stated directly rather than encoded as an iteration count. Polling is the reachable primitive here: `await_spawned_future` resolves a single spawned hop and cannot follow the retry cascade (each backoff re-spawns a fresh future whose handle is discarded), and the model-update event at the clear site is callback-based with no awaitable bridge; the sibling helper `wait_for_num_spawned_futures` waits the same way.
- **`test_cloud_pref_not_synced_when_current_value_not_syncable`** ([`cloud_preferences_syncer_tests.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/settings/cloud_preferences_syncer_tests.rs)): removed a fixed 1s sleep immediately before `await_spawned_futures(…)`, which already polls with a 5s ceiling; the awaited futures are spawned synchronously earlier in the test, so the extra fixed wait was redundant.

## Stability evidence

- `test_metadata_after_trash_item_failure`: 30/30 in a standalone loop (~3.6s each — the retry backoff itself still has to elapse; polling replaces only the fixed ceiling).
- `test_cloud_pref_not_synced_when_current_value_not_syncable`: 30/30, ~0.14–0.53s each (faster now that the redundant wait is gone).
- Full `cargo nextest run -p warp`: 5792/5792 passed.
- `cargo build -p warp --tests` + `cargo clippy -p warp --lib --tests`: zero warnings.

## Testing

## Agent Mode


fbartho:
Note: this is right on the bubble for me in terms of which approach is correct / best matches the style / concurrency practices in this repo. I don't love that it is a sleeping loop, but I think it's safer with this implementation.

I'd love to hear what a maintainer thinks about this approach.

fbartho:

seemeroland (approved):
Thank you!

fbartho:
@seemeroland the PR is approved, but I don't have a merge button, is that something you need to do? or what should I do to get this merged? Thanks!

fbartho:
@oss-maintainers — This was approved but not merged. Would you like anything else?

Issue #13974, opened 2026-07-19
app tests: three append-only process-global caches leak across tests (experiments layers, telemetry queue, bootstrap cache)


fbartho:
## Summary

Three tests pass in isolation but fail in full-suite runs because each depends on a process-global cache that some other test populates with real (non-test) data and never resets. They share one root-cause pattern — an append-only or first-writer-wins global that has no per-test teardown — so they are grouped here. These currently report as failures on every contributor's full-suite PR gate.

## Affected tests

- `experiments::tests::test_create_experiment_layer_mappings` — `app/src/experiments/mod_tests.rs:152`
- `server::telemetry::tests::test_persist_events_doesnt_include_ugc_events` — `app/src/server/telemetry/mod_tests.rs:46`
- `terminal::bootstrap::tests::test_trims_whitespace` — `app/src/terminal/bootstrap_tests.rs:46`

## Reproduction

Full suite (fails; these three appear among the 12-14 failures):

```
cargo test -p warp --lib
```

Each passes in isolation:

```
cargo test -p warp --lib -- --exact experiments::tests::test_create_experiment_layer_mappings
cargo test -p warp --lib -- --exact server::telemetry::tests::test_persist_events_doesnt_include_ugc_events
cargo test -p warp --lib -- --exact terminal::bootstrap::tests::test_trims_whitespace
# => each: ok. 1 passed
```

`--test-threads=1` does not fix them: these are cumulative process-global caches carried across sequential tests, not data races. A full single-threaded run still fails with the same set.

### 1. experiments — `EXPERIMENT_LAYER_MAPPINGS` never cleared

```rust
// app/src/experiments/mod.rs:77
static ref EXPERIMENT_LAYER_MAPPINGS: DashMap<&'static str, &'static Layer> = DashMap::new();
```

`create_experiment_layer_mappings` only inserts, never clears the map. The public `experiments::init` maps the four real layers into this same global. Any test that runs `app.update(experiments::init)` — e.g. `workspace/view_tests.rs:229`, `pane_group/mod_tests.rs:202`, `terminal/input_tests.rs:314`, and the shared helper `test_util/terminal.rs:186` — leaves the real layers behind. `test_create_experiment_layer_mappings` then asserts `EXPERIMENT_LAYER_MAPPINGS.len() == 2` (mod_tests.rs:152) and sees more than 2. In isolation the map only holds the 2 test experiments.

### 2. telemetry — global `TELEMETRY` event queue drained wholesale

```rust
[...]

PR #14043, opened 2026-07-20
Stop Workspace's chdir-to-home from leaking the process cwd across tests

### LINKED ISSUE #13980: app tests: unconditional `chdir` to home directory in RootView construction leaks the process cwd, breaking every test that resolves a relative path
## Summary

Three tests fail deterministically in full-parallel `cargo test -p warp --lib` runs — reproduced in 3 out of 3 full-suite attempts on this machine — because `RootView::new` (the production workspace-view constructor, `app/src/workspace/view.rs:2862`) unconditionally calls `terminal::platform::init()`, which on macOS calls `env::set_current_dir(dirs::home_dir().unwrap())` with no `#[cfg(not(test))]` guard and no restoration afterward. The first test in the process that constructs a `RootView` (there are dozens across `app/src/workspace/view_tests.rs`) permanently changes the whole test binary's working directory to the user's home directory for the remainder of the run. Every subsequent test in the same process that resolves a relative path against the crate-root cwd then fails. This is a first-writer-wins process-global cache pollution, the same shape as #13974's `BOOTSTRAP_CACHE`/`EXPERIMENT_LAYER_MAPPINGS`/`TELEMETRY` cluster, but via the OS process cwd rather than an in-process static.

## Affected tests

- `notebooks::file::tests::test_load_local` — `app/src/notebooks/file/mod_tests.rs:74`
- `notebooks::file::tests::test_load_before_session` — `app/src/notebooks/file/mod_tests.rs:233`
- `util::path::tests::test_resolve_command` — `app/src/util/path_tests.rs:21`

All three currently report as failures on every contributor's full-suite PR gate.

## Reproduction

All three pass in isolation:

```
cargo test -p warp --lib -- --exact notebooks::file::tests::test_load_local
cargo test -p warp --lib -- --exact notebooks::file::tests::test_load_before_session
cargo test -p warp --lib -- --exact util::path::tests::test_resolve_command
# => each: ok. 1 passed
```

Full suite fails deterministically (3/3 local runs on this 10-core machine reproduced all three, alongside the four already-tracked clusters #13972-#13975):

```
cargo test -p warp --lib
# run 1: 5765 passed; 12 failed, incl. all three of the above
# run 2: 5764 passed; 13 failed, incl. all three of the above
# run 3: 5764 passed; 13 failed, incl. all three of the above
```

Direct proof — instrumented `test_load_local`/`test_load_before_session` with `eprintln!("cwd={:?}", std::env::current_dir())` immediately on entry and reran the full suite:

```
DEBUG test_load_local cwd=Ok("/Users/fbarthelemy") readme_exists=false
DEBUG test_load_before_session cwd=Ok("/Users/fbarthelemy") readme_exists=false
```

`/Users/fbarthelemy` is the invoking user's home directory — exactly the value `dirs::home_dir()` returns — not the expected crate-root cwd (`app/`) that `cargo test` starts the process in. This is the smoking-gun match to the mutation below.

For `test_resolve_command`, a minimal two-target repro isolates the exact polluting sibling module and fails 5/5 times with the specific assertion that fires under a real cwd/PATH-adjacent collision from the same test file family (`pane_group::pane::local_harness_launch::tests`, which also does its own scoped `PATH`/`HOME` env-var overrides via an RAII `EnvVarGuard` — see note below on why that module is a secondary, narrower contributor, not the primary cause):

```
cargo test -p warp --lib -- --test-threads=4 util::path::tests::test_resolve_command pane_group::pane::local_harness_launch::tests
# => util::path::tests::test_resolve_command FAILED, 5/5 runs
#    panicked at app/src/util/path_tests.rs:43:5: assertion failed: resolve_executable("pwd").is_some()
```

Note: the `local_harness_launch::tests` pairing above is a second, independent trigger for `test_resolve_command` specifically (its `EnvVarGuard::set("PATH", fake_bin_dir)` calls are unsynchronized with `test_resolve_command`, which carries no `#[serial]` marker) — but it does not explain the two `notebooks::file` tests, which fail on a relative **path**, not `PATH`. The chdir-to-home mutation documented above is the common root cause across all three; the `local_harness_launch` PATH-guard interaction is a compounding, narrower issue affecting `test_resolve_command` alone (see Fix direction).

## Root cause

`RootView::new`, the constructor for the app's main workspace view, unconditionally initializes terminal platform state on every call:

```rust
// app/src/workspace/view.rs:2862
terminal::platform::init().expect("Terminal platform initialized");
```

On macOS, that function performs a real, permanent `chdir`:

```rust
// app/src/terminal/platform.rs:1-32
pub fn init() -> Result<()> {
    #[cfg(target_os = "macos")]
    mac::init()?;
    Ok(())
}

#[cfg(target_os = "macos")]
mod mac {
    pub fn init() -> Result<()> {
        set_locale_environment();

        // Switch to home directory.
        env::set_current_dir(dirs::home_dir().unwrap()).unwrap();

        Ok(())
    }
    ...
}
```

There is no `#[cfg(test)]`/`#[cfg(not(test))]` guard anywhere on this call path, and no code restores the original cwd afterward. `app/src/workspace/view_tests.rs` constructs `RootView` (via `app.add_window(...)` with the workspace view constructor) in dozens of tests, so the first such test to run in the process permanently relocates the whole test binary's working directory to `$HOME` for every test that runs after it — a first-writer-wins global side effect on the OS process, not an in-process Rust static, but the identical failure shape as the caches in #13974.

Every test that subsequently resolves a relative path against the crate-root cwd is affected:

- `test_load_local` / `test_load_before_session` call `file_notebook.open_local("../README.md", ...)` (mod_tests.rs:81, 244) and `dunce::canonicalize("../README.md")` (mod_tests.rs:265), both of which resolve against `std::env::current_dir()`. Once cwd is `$HOME`, `../README.md` means `$HOME/../README.md` (or wherever `$HOME`'s parent is), which does not exist — `dunce::canonicalize` returns `Os { code: 2, kind: NotFound }`, exactly the panic observed at mod_tests.rs:265:69, and `FileModel::open`'s `async_fs::read_to_string` on the same non-existent path logs "Unable to read local notebook file" before `test_load_local`'s `editor.markdown(ctx).len() > 4` assertion fails on empty content (mod_tests.rs:107:13).
- `test_resolve_command` calls `resolve_executable("../script/run")` (path_tests.rs:38, a relative path relying on cwd) and `resolve_executable("pwd")`/`resolve_executable("env")` (path_tests.rs:31, 43, PATH-based lookups). The relative-path assertion is directly cwd-sensitive; `pwd`/`env`, meanwhile, are additionally exposed to the separate `local_harness_launch::tests` PATH-guard race described above, since `resolve_executable` reads the process-global `env::var_os("PATH")` (`app/src/util/path.rs:84`) with no synchronization against concurrent test threads that hold `PATH` overridden.

## Fix direction

- Guard the `terminal::platform::init()` call in `RootView::new` (`app/src/workspace/view.rs:2862`) so it does not run under test — e.g. `#[cfg(not(test))]`, or route through a test-only mock/no-op `platform::init` the way other singletons in this constructor already use `_for_test`/`mock` variants (`ServerApiProvider::new_for_test()`, `AuthStateProvider::new_for_test()`, `AuthManager::new_for_test`, etc., all visible in the same constructor). This is the single highest-leverage fix: it eliminates the chdir-to-home side effect for the entire test binary, not just these three tests.
- Independently of the above, `resolve_executable`'s tests being sensitive to concurrent `PATH` mutation from `pane_group::pane::local_harness_launch::tests` is a real (if secondary) reliability gap: either add `#[serial_test::serial]` to `test_resolve_command` in the same serial group the `local_harness_launch` PATH/HOME-mutating tests already use, or have `test_resolve_command`'s PATH-based assertions (`env`, `pwd`) call `resolve_executable_in_path` with an explicitly captured `PATH` snapshot rather than relying on the ambient process env, consistent with the guidance in that function's own doc comment about callers needing a specific PATH.
- Longer term: any production code path that mutates true process-global OS state (cwd, env vars) on every construction of a commonly-tested view is a standing trap for the whole test suite. Consider whether `terminal::platform::init()` should be idempotent/guarded at the call site (e.g. `once_cell`-gated to run at most once per process) rather than relying on every caller to know not to invoke it in tests.

## Related

Related: #13853 (warp_editor render::tests Once-poisoning — same class, different crate) and the sibling flake tickets #13972, #13973, #13974, #13975, and #13979 (real-wall-clock `Timer::after` timing sensitivity — a different mechanism, filed from the same investigation). This ticket's mechanism (first-writer-wins mutation of true OS process state, not an in-process Rust static) is closest in shape to #13974's cumulative-cache cluster, but the "static" here is the kernel-tracked working directory itself.

fbartho:
Closes #13980

## Description

`Workspace::new` ([`view.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/workspace/view.rs)) unconditionally called `terminal::platform::init()`, which on macOS ([`platform.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/terminal/platform.rs)) does `env::set_current_dir(home_dir)` with no restoration. The first test in a shared process that constructs a `Workspace` permanently relocates the whole test binary's cwd to `$HOME` — a first-writer-wins process-global side effect (same shape as #13974's cache cluster, via the OS cwd instead of a static). Every later test resolving a crate-relative path then fails depending on scheduling: `test_load_local` / `test_load_before_session` (open `../README.md`), `test_resolve_command` (resolves `../script/run`).

### Fix

`#[cfg(not(test))]` on the `platform::init()` call site in `Workspace::new`, matching the crate's existing pattern for process-global side effects under test (`block.rs`, `context_chips/logging.rs`, `keyboard.rs`). `platform::init()` itself is untouched — its non-test callers keep it.

### Secondary fix

`test_resolve_command` also reads the ambient `PATH` with no synchronization against the already-`#[serial]` PATH-overriding tests in `pane_group::pane::local_harness_launch_tests`; added the same marker so it joins that serial group. Independent of the cwd leak, and would otherwise still flake.

## Linked Issue

  - _Not applicable: test-only change, no UI._

## Testing

Full crate suite under plain `cargo test -p warp --lib` (nextest's per-process isolation masks this class):

- Before: `FAILED. 5779 passed; 13 failed` — including all three ticketed tests, panic sites matching the ticket verbatim (`markdown len assertion`, `dunce::canonicalize` NotFound, `Option::unwrap()` on None).
- After: `FAILED. 5782 passed; 10 failed` — the three ticketed tests all pass; the remaining 10 are pre-existing and tracked in the sibling flake tickets (#13972–#13975).
- `cargo clippy -p warp --lib --tests`: zero warnings.

  - _Not applicable: the issue only applies to test builds (the guard only affects code compiled into test binaries)._

## Agent Mode

### DIFF
--- a/app/src/util/path_tests.rs
+++ b/app/src/util/path_tests.rs
@@ -18,6 +18,11 @@ fn test_trim_newline() {
 /// TODO(CORE-3626): write an equivalent test with Windows paths.
 #[cfg(not(windows))]
 #[test]
+// Reads the ambient process PATH; must not race tests elsewhere in the crate that
+// temporarily override PATH (e.g. `local_harness_launch_tests`'s `EnvVarGuard`). Those
+// tests are already `#[serial_test::serial]`, so joining the same (default) serial group
+// here is what actually synchronizes against them. See #13980.
+#[serial_test::serial]
 fn test_resolve_command() {
     use std::path::Path;
 
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -2864,6 +2864,12 @@ impl Workspace {
             model.insert(window_id, new_resizable_modal_sizes)
         });
 
+        // Skipped under test: on macOS this changes the process's current working
+        // directory to the user's home directory with no restoration afterward. Since it's
+        // never undone, the first test in the binary that constructs a `Workspace` would
+        // otherwise permanently relocate the whole test process's cwd, breaking every later
+        // test that resolves a relative path (see #13980).
+        #[cfg(not(test))]
         terminal::platform::init().expect("Terminal platform initialized");
 
         let tab_bar_overflow_menu = Self::build_tab_bar_overflow_menu(ctx);


fbartho:

fbartho:

fbartho:
This PR is still a valid fix. The upstream ticket is still awaiting maintainer review, please advise @oss-maintainers?

fbartho:
Temporarily closing in until the linked issue #13980 gets processed.

Issue #6204, opened 2025-03-28
Windows fullscreen window corners rounding


mckoda09:
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [x] I have [searched Warp known issues page](https://docs.warp.dev/help/known-issues) and my issue is not there
- [ ] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/help/sending-us-feedback#gathering-warp-logs)

### Describe the bug

![Image](https://github.com/user-attachments/assets/1670eee5-ac2e-4c26-912a-71a8f3dd4429)
Can you see those rounded corners on fullscreen? Usually desktop apps on windows don't have those rounded corners.

### To reproduce

1. Open Warp for Windows
2. Go fullscreen

### Operating system (OS)

Windows

### Operating system and version

Windows 11

### Regression

No, this bug or issue has existed throughout my experience using Warp

### Does this block you from using Warp daily?

No

### Is this an issue only in Warp?

Yes, I confirmed that this only happens in Warp, not other terminals.

### Warp Internal (ignore): linear-label:b9d78064-c89e-4973-b153-5178a31ee54e

None

mckoda09:
Also i noticed, that sometimes acryllic background texture becomes just white background

aadilayub:
This is a bug on Linux (GNOME/Wayland) as well. It always has rounded corners, even when tiled or fullscreen.

mckoda09:
warp was developed for macos, and i feel like they dont care about windows and linux users...

acarl005:
Related to #6745

MrBns:
this issue exist for such a long time,  I just noticed it today.

acarl005:
This is open for contributors to open a pull request

PR #14109, opened 2026-07-21
Don't round window corners in fullscreen

### LINKED ISSUE #3281: Full screen presentation on macOS

bloomingbridges

### Describe the bug

I'm hesitant to log this as a bug, but I've been using Warp for a couple months now and I feel like this only changed recently, so there is a chance it's unintentional.

Whenever I put Warp in full screen mode the rounded corners are still being drawn, so it basically looks like a blown up window rather than a full screen application.

### To reproduce

1. Go into fullscreen mode by either pressing the green button in the app's title bar or selecting "Window > Enter Full Screen"

### Expected behavior

Expecting the window chrome to disappear to reduce distracting elements on screen.

### Screenshots

Screenshot 2023-06-30 at 12 00 00

### Operating system

MacOS

13.4.1

### Shell Version

_No response_

### Current Warp version

v0.2023.06.27.19.34.stable_01

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

### Warp Internal (ignore): linear-label:b8107fdf-ba31-488d-b103-d271c89cac3e

None

s1ckoleg:
## Description

In fullscreen on macOS, the window corners look broken: the corner is drawn rounded, but a small square piece of the window sticks out past the rounded edge.

**Root cause:** `WindowManager::window_corner_radius()` only checks for tiling window managers and Windows — it never checks fullscreen state — so the root workspace container (and theme background image) in `WorkspaceView::render()` are always drawn with an 8px corner radius on macOS. Corner rounding in warpui is a per-quad rounded-rect SDF alpha effect, not a clip mask, so the rounded root container just leaves a transparent notch at each screen corner. The square, opaque terminal-background container sits only `WORKSPACE_PADDING` (1px) inside it (its own 8px radius was removed in #9474), while an 8px radius notch reaches ~2.3px deep along the diagonal — so the square corner pokes out through the notch. In windowed mode the native drop shadow hides this; in fullscreen it's clearly visible at the bare screen corner.

**Fix:** add `WindowManager::window_corner_radius_for_window(window_id)`, which returns square corners when that window is fullscreen (a fullscreen window occupies the entire screen, so square corners are also the native behavior), and use it for the root workspace container + background image. This is per-window rather than based on `is_active_window_fullscreen()` so that, on multi-display setups, a windowed Warp window keeps its rounded corners while another window is fullscreen. The ~15 modal scrim call sites intentionally keep the existing global method to avoid scope creep; if a scrim should also be square in fullscreen, that can follow up separately.

## Linked Issue

Closes #3281

Related: #6204 is the same complaint on Windows, but it is not closed by this PR — on Windows the in-app radius is already 0 and the rounding comes from OS APIs, which is a separate code path.

## Testing

- `cargo check -p warpui_core` passes.
- `cargo check -p warp` could not be run locally: the machine only has Command Line Tools, and `crates/warpui/build.rs` requires the Xcode Metal toolchain (`xcrun metal`). The app-side change is a one-line call-site swap on the same receiver chain (`app.windows()`), against the new method that compiles in `warpui_core` — relying on CI for the full check.
- Manual test plan: run `./script/run`, toggle native fullscreen, and confirm all four screen corners are square with no notch/square-piece artifact; exit fullscreen and confirm the 8px rounded corners return; with two displays, confirm a windowed Warp window keeps rounded corners while another Warp window is fullscreen.

Before:
Screenshot 2026-07-22 at 09 59 45

After:
Screenshot 2026-07-22 at 09 59 14

## Agent Mode

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -27495,7 +27495,9 @@ impl View for Workspace {
             );
         }
 
-        let window_corner_radius = app.windows().window_corner_radius();
+        let window_corner_radius = app
+            .windows()
+            .window_corner_radius_for_window(self.window_id);
         let workspace = Container::new(stack.finish()).with_corner_radius(window_corner_radius);
 
         let mut stack = Stack::new();
--- a/crates/warpui_core/src/windowing/state.rs
+++ b/crates/warpui_core/src/windowing/state.rs
@@ -198,6 +198,21 @@ impl WindowManager {
         CornerRadius::with_all(Radius::Pixels(radius))
     }
 
+    /// Like [`Self::window_corner_radius`], but square when the given window is fullscreen: a
+    /// fullscreen window occupies the entire screen, and rounding its corners leaves transparent
+    /// notches at the screen corners that square content behind can poke through.
+    pub fn window_corner_radius_for_window(&self, window_id: WindowId) -> CornerRadius {
+        let is_fullscreen = self
+            .platform_window(window_id)
+            .map(|window| window.fullscreen_state() == FullscreenState::Fullscreen)
+            .unwrap_or(false);
+        if is_fullscreen {
+            CornerRadius::with_all(Radius::Pixels(0.))
+        } else {
+            self.window_corner_radius()
+        }
+    }
+
     pub(crate) fn open_window(
         &mut self,
         window_id: WindowId,


s1ckoleg:
@alokedesai hi! could you please take a look

alokedesai (approved):

alokedesai:
Hi there! So sorry for the delay, this looks great. Just approved; thank you so much for contributing to Warp!

s1ckoleg:
@alokedesai Thank you! Could you please trigger a new Warp CI? Seems like I can't merge it until CI is green.

alokedesai (approved):

PR #14122, opened 2026-07-22
docs(specs): inline code chip ink-width background (GH13914)

### LINKED ISSUE #13914: Inline code chip background extends past glyph ink on the right (advance width vs ink width)
## Summary

The background chip behind inline monospace runs (inline code spans, and anything else flowing through the same paint path) is visibly asymmetric: blank space inside the background to the RIGHT of the text, while the glyphs sit nearly flush against the background's LEFT edge. Noticed while reviewing Markdown-viewer screenshots; the layout code involved predates current work.

## Root cause (traced)

`Line::paint` computes the chip's horizontal span at `crates/warpui_core/src/text_layout.rs:1547-1573`:

- `visible_right = max(glyph_x + glyph.width)` (:1572), where `Glyph.width` is documented (:685) as the glyph's **advance** — the reserved cursor-to-next-character distance, which includes the font's right-side bearing — not the inked extent.
- `visible_left` comes from the first glyph's origin, which for monospace fonts is near the actual left ink edge.

So the right edge reserves bearing whitespace the left edge never gets — an inherent advance-vs-ink asymmetry, not a padding bug: `paint_run_background` (:1207, sole call site :1578) applies `block_padding` symmetrically (:1258-1265) on top of the already-lopsided span.

## Affected surfaces

Everything using `paint_run_background`: inline code chips, hyperlink backgrounds, and the upcoming `<kbd>` keycaps (#13733) by shared code.

**Possibly related but separate:** fenced code blocks show a similar-looking right-side gap, but they render through a different path (block-level background in `crates/editor/src/render/element/`), so that would be an independent bug needing its own verification — not asserting shared root cause.

## Test note

`crates/warpui_core/src/text_layout_tests.rs:497-526` pins the current geometry (asserts the rect exactly equals `visible_left`/`visible_right`) — it validates the plumbing, not whether `visible_right` reflects ink; it will need updating, not just extending, with any fix.

## Expected

The background's left and right insets around the glyph ink should be visually symmetric (either measure ink extents for the trailing edge, or reserve equivalent bearing on both sides).

fbartho:
Closes #13914

Adds a product spec and tech spec for the inline-code chip background asymmetry: the chip's background/border is currently painted using each glyph's advance width (which bakes in right-side bearing/spacing), so the gap between the last character's ink and the chip's right edge is consistently larger than the equivalent left-side gap — and varies with whichever glyph ends the span.

The spec's proposed design draws the chip's edges from glyph **ink extents** plus a font-relative design padding: `[first_ink_left − pad, last_ink_right + pad]` where `pad` is computed per run at paint time as `font_size × ratio` — it scales automatically with the current font size, with interior glyph positions remaining advance-based. Per-glyph ink is available cross-platform through the existing `platform::FontDB::glyph_raster_bounds` (already cached; macOS font-kit raster bounds and Linux swash placement have matching true-ink semantics), so the correction is exact rather than a tuned constant — this also incidentally fixes the current right edge clipping ink overshoot from italic/overhang glyphs. The tech spec documents `glyph_typographic_bounds` as an examined dead end (its two backends measure different things), the device→paint scale conversion, a ~1px macOS AA-fudge residual, and a zero-ink (e.g. trailing space) advance-edge fallback; it confirms the TUI Markdown renderer does not reach this paint path and is unaffected.

Remaining open questions are design-taste rather than correctness: the dimensionless padding ratio (currently `1/12` of the font size — chosen to approximate a typical monospace left-side bearing so the chip's left edge stays where today's advance-based edge sits) and visual sign-off on the macOS AA residual.

<details><summary>Review-feedback accounting (continuation of #14026 — all 8 findings across 4 review rounds)</summary>

This PR replaces #14026, which is closed after exhausting its review-trigger budget. All eight `oz-for-oss` findings raised across four review rounds on #14026 are addressed as of this branch's spec tip; the table below is the full accounting.

| Round | Finding | File:Line | Resolving commit |
|---|---|---|---|
| 1 | Asymmetric `block_padding` interaction — legacy left-only shift undocumented against the new ink-edge design | `tech.md:27` | `74a882847` |
| 2 | Equal-padding promise conflicted with `tech.md`'s unchanged (left-only) `block_padding` residual | `product.md:22` | `0fcebacbd` |
| 2 | Sub-pixel left-edge-preservation overpromise from the fixed `font_size/12` padding | `product.md:63` | `0fcebacbd` |
| 2 | Blast-radius scope unresolved — implementation reaches every `background_color`/`border` consumer, not just chips | `product.md:73` | `0fcebacbd` |
| 3 | Arithmetic error: `2 × block_padding` magnitude cited as ~1.08px, actually `font_size/5` ≈ 2.6px at 13px | `tech.md:132` | `96620e117` |
| 3 | Zero-ink fallback conflict between product.md (`advance-edge ± pad`) and tech.md's testing section | `tech.md:387` | `96620e117` |
| 4 | Fallback-rules section still specified raw `visible_left`/`visible_right` for zero-ink glyphs, contradicting product.md and the implementation | `tech.md:254` | `bc345c70a` |
| 4 | Left-edge delta stated as an exact `2 × block_padding` figure, ignoring the first glyph's ink bearing and the new `pad` (glyph/font-dependent in reality) | `product.md:66` | `bc345c70a` |

All eight threads on #14026 have reply comments pointing at their resolving commit; the round-1–3 threads are marked resolved on GitHub. The round-4 threads (`tech.md:254`, `product.md:66`) are fixed by `bc345c70a` on this branch but not yet marked resolved on #14026, since #14026 is being closed in favor of this PR rather than receiving further replies.

A validation implementation exists locally (three commits, rebased onto this spec tip; not pushed, so no stable SHAs to cite — it will arrive as its own PR once this spec is approved, per the spec→implementation flow):
- rebalance inline code chip background (ink-edge geometry, `is_chip` predicate, `block_padding` compensation)
- equal ink-edge padding for chip-family runs
- fix `block_padding` magnitude in ink-padding doc comment

Gates: `cargo nextest run -p warpui_core -p warp_editor` (794 passed), `cargo clippy --all-targets --all-features --tests -- -D warnings` (clean), `cargo fmt -- --check` (clean).

</details>


fbartho:
(Just commenting as this is waiting for review).

fbartho:
@vorporeal This PR is still waiting for review. (Just trying to avoid the stale-bot)

PR #10357, opened 2026-05-07
fix: render Thai (and other complex-script) combining marks correctly

chatre7:
The terminal grid was rendering Thai vowel marks (ั, ี, …) at the wrong position or dropping them entirely. The bug was a chain of issues across font fallback, byte/char indexing, and HarfBuzz glyph positioning.

Five root causes, fixed end-to-end:

1. DirectWrite fallback locale on Windows was hardcoded to "en-US", which biased `IDWriteFontFallback::MapCharacters` toward Latin-script fonts even when asked for a Thai codepoint. Switched to "" (system locale).

2. Linux/FreeBSD font fallback used a hardcoded "en" instead of reading `LC_ALL` / `LC_CTYPE` / `LANG`. Now picks up the user's actual locale.

3. `CellGlyphCache::glyph_for_string` returned `None` whenever the layout produced more than one glyph, silently dropping every Thai cluster (base consonant + combining vowel = 2 glyphs). Renamed to `glyphs_for_string` returning a `Vec`, so all glyphs make it through.

4. `AttributedStringBuilder::character_index_to_cell_map` was filled one entry per char, but `paint_line` indexes it with `glyph.index` — which cosmic_text emits as a UTF-8 *byte* offset, not a char index. For ASCII the two coincide; for any 3-byte codepoint (Thai, CJK, emoji) the lookup landed on the wrong cell. Now fill one entry per byte.

5. (The actual visible bug.) `glyphs_for_string` collected only `(GlyphId, FontId)` and discarded `Glyph::position_along_baseline`. Every glyph then got drawn at the same cell origin, so combining marks stacked on top of the base instead of sitting where HarfBuzz placed them — visually they bled into the previous cell or disappeared. Now carry the position through and apply it in `render_cell_glyph`.

The ligature path (`render_grid_with_ligatures`) was also fixed for the same byte-indexing problem and gained a deferred-rendering pass for `CharOrStr::Str` cells, which use direct glyph_for_char + DirectWrite fallback (the cached `text_layout_cache` path can't load system Thai fonts on demand).

Tests:
- New `mixed_ascii_thai_records_missing_glyphs` exercises the text-layout missing-glyph path with mixed Latin + Thai.
- New `test_attributed_string_builder_byte_indexed_cell_map` locks in the byte-indexed (not char-indexed) cell map for multi-byte codepoints, with an ASCII regression check alongside it.
- Manually verified end-to-end on Windows with `python -c "print('สวัสดี')"` — output now matches PowerShell.

[...]

PR #14124, opened 2026-07-22
fix: render Thai/Lao complex-script combining marks (incl. SARA AM) correctly

### LINKED ISSUE #8357: Thai tone marks and upper vowels not rendering in Claude Code
### Pre-submit Checks

- [ ] I have an issue with AI and have included the debugging ID (Optional, but helps expedite the AI quality fix). [Debugging ID instructions](https://docs.warp.dev/agents/ai-faqs#gathering-ai-debugging-id)
- [ ] I have technical issue and have included the logs (optional, but helps expedite the bug fix). [Log instructions](https://docs.warp.dev/support-and-billing/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When using **Claude Code** (Anthropic's CLI tool) in Warp terminal, Thai tone marks (วรรณยุกต์) and upper vowels (สระบน) are not rendering properly. All diacritical marks above Thai consonants disappear.

This is related to #3584 which was closed as "not planned", but this issue significantly affects Thai language users who use Claude Code in Warp.

**Example:**
- Expected: ก่อนที่จะรายงาน (with tone marks)
- Actual: กอนทจะรายงาน (tone marks missing)

Characters affected:
- Tone marks: ่ ้ ๊ ๋
- Upper vowels: ิ ี ึ ื ั ็

### To reproduce

1. Open Warp terminal
2. Run Claude Code CLI (claude command)
3. Have Claude output any Thai text containing tone marks or upper vowels
4. Observe that all diacritical marks above consonants are missing

### Expected behavior

Thai text should display with all diacritical marks intact, including:
- Tone marks: ่ ้ ๊ ๋ (mai ek, mai tho, mai tri, mai chattawa)
- Upper vowels: ิ ี ึ ื ั ็ (sara i, sara ii, sara ue, sara uue, mai han-akat, mai tai khu)

Example: "สวัสดี" should show all characters, not "สวสด"

_No response_

### Operating system (OS)

Windows

Windows 11

### Shell Version

_No response_

### Current Warp version

_No response_

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

ultramcu:
## Description

Renders Thai/Lao (and other complex-script) combining marks correctly in the terminal grid, including the Thai/Lao **SARA AM** vowel (ำ / ຳ) — previously the grid split it from its base consonant so the nikhahit dot detached (e.g. `คำ`, `น้ำ`, `ทำ`, `สำคัญ`).

This **continues @chatre7's work from #10357**, which was auto-closed for inactivity. All of his commits are preserved here with their original authorship — thank you, @chatre7 🙏. The branch has been **rebased onto the latest `master`** (resolving the divergence since the original PR, including the Rust 2024 edition migration and the `grid_renderer` changes), re-formatted, and **rebuilt + run locally** to confirm it still renders correctly.

What it does:
- **Base fix:** route complex-script cells through the same HarfBuzz shaper with GPOS positions, and preserve styled (bold/italic/filter-match) fonts, so combining marks land on their base glyph instead of stacking at the cell origin.
- **SARA AM:** a look-ahead folds a consonant (or a `CharOrStr::Str` cell) and a following SARA AM into a single shaped cluster, so HarfBuzz places the nikhahit dot above the consonant via GPOS instead of drawing it standalone; a line-start SARA AM still falls back to its sara-aa tail.

Supersedes #10357.

## Linked Issue

Closes #8357

## Testing

Built and ran a local OSS bundle from this branch (rebased onto current `master`, Rust 2024 edition) on macOS: Thai text with SARA AM renders with the vowel attached to its consonant. `cargo check -p warp` and `cargo fmt --check` are clean. The existing `test_decompose_sara_am` unit test covers the decomposition helper.

### Screenshots / Videos

Thai text with SARA AM in a local build of this branch — **before** (stock `master`, the vowel detaches from its consonant) vs **after** (this PR, attached correctly):

| Before | After |
|---|---|
| before | after |

## Agent Mode

CHANGELOG-BUG-FIX: Fix rendering of Thai/Lao complex-script combining marks, including the SARA AM vowel attaching to its base consonant.

### DIFF
--- a/app/src/terminal/grid_renderer.rs
+++ b/app/src/terminal/grid_renderer.rs
@@ -1118,6 +1118,11 @@ fn render_grid_with_ligatures<'a>(
         let offset_row = start_row + offset;
         let mut string_builder =
             AttributedStringBuilder::new(font_family, font_family, grid.columns());
+        // Cells whose content is CharOrStr::Str (base char + zero-width combining chars, e.g.
+        // Thai "วั") are excluded from layout_line and rendered here via glyph_for_char, which
+        // has DirectWrite system-font fallback. layout_line uses cosmic_text/fontdb and cannot
+        // load system Thai fonts on demand, so it produces glyph_id=0 for those characters.
+        let mut deferred_str_cells: Vec<(usize, ColorU, String)> = Vec::new();
 
         let Some(row) = grid.row(row_idx) else {
             report_error!("grid_renderer should not try to render an out-of-bounds row");
@@ -1424,7 +1429,14 @@ fn render_grid_with_ligatures<'a>(
                     // attempt the same thing, so we replace it with a placeholder
                     string_builder.append_placeholder(col);
                 } else {
-                    string_builder.append_content(cell.content_for_display(), col);
+                    match cell.content_for_display() {
+                        CharOrStr::Str(s) => {
+                            string_builder.append_placeholder(col);
+                            deferred_str_cells
+                                .push((col, cell_colors.foreground_color, s.to_owned()));
+                        }
+                        other => string_builder.append_content(other, col),
+                    }
                 }
             }
         }
@@ -1453,6 +1465,30 @@ fn render_grid_with_ligatures<'a>(
             &string_data.character_index_to_cell_map,
             ctx.scene,
         );
+
+        // Render deferred CharOrStr::Str cells (base + combining chars, e.g. Thai "วั").
+        // Look up the base char with full DirectWrite fallback to find the script font, then
+        // reuse that font for every combining mark — DirectWrite cannot map combining marks
+        // standalone (no script context), but the script font that contains ว also contains ั.
+        for (col, foreground_color, content) in deferred_str_cells.drain(..) {
+            let origin = line_origin + vec2f(col as f32 * cell_size.x(), 0.);
+            let mut chars = content.chars();
+            let Some(first_char) = chars.next() else { continue };
+            let Some((first_glyph, base_font_id)) =
+                ctx.font_cache.glyph_for_char(default_font_id, first_char, true)
+            else {
+                continue;
+            };
+            ctx.scene
+                .draw_glyph(origin, first_glyph, base_font_id, font_size, foreground_color);
+            for c in chars {
+                if let Some((glyph_id, _)) = ctx.font_cache.glyph_for_char(base_font_id, c, false)
+                {
+                    ctx.scene
+                        .draw_glyph(origin, glyph_id, base_font_id, font_size, foreground_color);
+                }
+            }
+        }
 
         if grid.filter_has_context_lines() {
             maybe_render_dotted_lines(
@@ -1726,27 +1762,6 @@ fn render_cell_glyph(
     let font_id = font_id_cache.font_ids[font_style as usize]
         .get_or_insert_with(|| ctx.font_cache.select_font(font_family, properties));
 
-    let glyph_and_font = match cell_content {
-        // Special-case whitespace, which doesn't need rendering.  We
-        // explicitly check these two chars instead of using
-        // `char::is_whitespace` for performance reasons.
-        CharOrStr::Char(' ' | '\t') => None,
-        CharOrStr::Char(char) => glyphs.glyph_for_char(char, *font_id, ctx.font_cache),
-        // Certain zerowidth characters, such as emoji presentation selectors, can affect the underlying glyph and
-        // change the rendering. Hence, we need to layout/render the text as a combined string, instead of simply
-        // the single character. For example, in \0x2601\0xFE0F, the FE0F selector causes ☁️ to be changed from a
-        // 1-width char to a 2-width char.
-        CharOrStr::Str(content_with_zerowidth) => glyphs.glyph_for_string(
-            content_with_zerowidth,
-            *font_id,
-            ctx.font_cache,
-            font_family,
-            font_size,
-            properties,
-            ctx,
-        ),
-    };
-
     let origin = grid_origin + glyph_offset + baseline_position;
 
     // Handle special unicode characters that will look better with native
@@ -1759,16 +1774,49 @@ fn render_cell_glyph(
                 glyph_type,
             });
         }
-        None => {
-            // Add FontId as part of the hashkey since characters with different
-            // fonts will have different glyph ids.
-            if let Some((glyph_id, font_id)) = glyph_and_font {
-                // If we don't have special handling for the character, draw the
-                // glyph from the font.
-                ctx.scene
-                    .draw_glyph(origin, glyph_id, font_id, font_size, foreground_color);
+        None => match cell_content {
+            // Special-case whitespace, which doesn't need rendering.  We
+            // explicitly check these two chars instead of using
+            // `char::is_whitespace` for performance reasons.
+            CharOrStr::Char(' ' | '\t') => {}
+            CharOrStr::Char(char) => {
+                if let Some((glyph_id, font_id)) =
+                    glyphs.glyph_for_char(char, *font_id, ctx.font_cache)
+                {
+                    ctx.scene
+                        .draw_glyph(origin, glyph_id, font_id, font_size, foreground_color);
+                }
             }
-        }
+            // Certain zerowidth characters, such as emoji presentation selectors, can affect the
+            // underlying glyph and change the rendering. Hence, we need to layout/render the text
+            // as a combined string. For example, \0x2601\0xFE0F causes ☁️ to become 2-wide.
+            // Thai combining vowels (e.g. "วั") also produce multiple glyphs — the base consonant
+            // and the above/below mark — which must all be drawn at the same cell origin so the
+            // font's own glyph metrics position the mark correctly above the consonant.
+            CharOrStr::Str(content_with_zerowidth) => {
+                for (glyph_id, glyph_font_id, position) in glyphs.glyphs_for_string(
+                    content_with_zerowidth,
+                    *font_id,
+                    ctx.font_cache,
+                    font_family,
+                    font_size,
+                    properties,
+                    ctx,
+                ) {
+                    // Apply HarfBuzz's shaped position so combining marks (e.g. Thai ั / ี)
+                    // sit at the offset the script's GPOS table prescribes — without this,
+                    // every glyph stacks on the cell origin and marks land on the previous
+                    // cell or wherever the bare glyph happens to extend.
+                    ctx.scene.draw_glyph(
+                        origin + position,
+                        glyph_id,
+                        glyph_font_id,
+                        font_size,
+                        foreground_color,
+                    );
+                }
+            }
+        },
     }
 
     if first_cell_in_link {
@@ -2643,9 +2691,15 @@ struct AttributedStringBuilder {
     current_style: StyleAndFont,
     current_style_start_char_index: usize,
     style_runs: Vec<(Range<usize>, StyleAndFont)>,
-    // Conceptually, this is a map from character index to cell index. However, since it is both
-    // dense and build in order, we can avoid the overhead of hashing by using a `Vec`
+    // A byte-offset-indexed map from each UTF-8 byte position in `line` to a cell (column) index.
+    // Each character fills *all* of its UTF-8 bytes with the same cell column so that
+    // `paint_line` can use `glyph.index` — which is a byte offset, not a char index — to
+    // look up the correct cell. Without this, multi-byte characters (Thai, CJK, emoji, …) are
+    // drawn at the wrong grid column.
     character_index_to_cell_map: Vec<usize>,
+    // Tracks the number of *chars* appended (as opposed to bytes) so that style-run ranges stay
+    // char-indexed, matching what `layout_line` expects.
+    char_count: usize,
     line: String,
 }
 
@@ -2661,12 +2715,13 @@ impl AttributedStringBuilder {
             current_style_start_char_index: 0,
             style_runs: Vec::new(),
             character_index_to_cell_map: Vec::with_capacity(columns),
+            char_count: 0,
             line: String::with_capacity(columns),
         }
     }
 
     fn next_character_index(&self) -> usize {
-        self.character_index_to_cell_map.len()
+        self.char_count
     }
 
     /// Flushes the currently cached style into the style runs list and update the current style
@@ -2724,8 +2779,15 @@ impl AttributedStringBuilder {
     /// This will update the mapping of cell indexes so that the character can be connected with
     /// its expected grid position later.
     fn append_character(&mut self, chr: char, column: usize) {
+        let start_byte = self.line.len();
         self.line.push(chr);
-        self.character_index_to_cell_map.push(column);
+        // `glyph.index` in `paint_line` is a UTF-8 byte offset, so we fill every byte of this
+        // character's UTF-8 encoding with the same cell column. This ensures multi-byte characters
+        // (Thai, CJK, emoji, …) are drawn at the correct grid column.
+        for _ in start_byte..self.line.len() {
+            self.character_index_to_cell_map.push(column);
+        }
+        self.char_count += 1;
     }
 
     /// Append a cell's content to the attributed string at a specific grid column.
--- a/app/src/terminal/grid_renderer.rs
+++ b/app/src/terminal/grid_renderer.rs
@@ -1476,7 +1476,11 @@ fn render_grid_with_ligatures<'a>(
                     match cell.content_for_display() {
                         CharOrStr::Str(s) => {
                             string_builder.append_placeholder(col);
-                            let content = if let Some(sa) = following_sara_am {
+                            // Only fold the SARA AM in when this cell is a Thai/Lao base
+                            // cluster it can attach to; otherwise leave it to render on its own.
+                            let attachable = s.chars().next().is_some_and(is_sara_am_base);
+                            let content = if let Some(sa) = following_sara_am.filter(|_| attachable)
+                            {
                                 skip_sara_am_at = Some(col + 1);
                                 let mut combined = s.to_owned();
                                 combined.push(sa);
@@ -1508,9 +1512,11 @@ fn render_grid_with_ligatures<'a>(
                                 ));
                             }
                         }
-                        // Consonant followed by SARA AM: defer both together so HarfBuzz shapes
-                        // the cluster and positions the nikhahit dot above the consonant via GPOS.
-                        CharOrStr::Char(c) if following_sara_am.is_some() => {
+                        // Thai/Lao consonant followed by SARA AM: defer both together so HarfBuzz
+                        // shapes the cluster and positions the nikhahit dot above the consonant via
+                        // GPOS. Gated on an attachable base so a SARA AM after a space/punctuation/
+                        // Latin cell is not consumed here (it renders standalone instead).
+                        CharOrStr::Char(c) if following_sara_am.is_some() && is_sara_am_base(c) => {
                             let sa = following_sara_am.expect("checked by guard");
                             skip_sara_am_at = Some(col + 1);
                             string_builder.append_placeholder(col);
@@ -1821,6 +1827,14 @@ fn decompose_sara_am(c: char) -> Option<(char, char)> {
         '\u{0EB3}' => Some(('\u{0ECD}', '\u{0EB2}')),
         _ => None,
     }
+}
+
+/// Returns true if `c` is a Thai or Lao consonant that a following SARA AM
+/// (ำ U+0E33 / ຳ U+0EB3) can attach to as its base. Used so a SARA AM after a
+/// non-attachable cell (space, punctuation, Latin, a lone vowel, etc.) is left
+/// to render on its own instead of being folded into the wrong base.
+fn is_sara_am_base(c: char) -> bool {
+    matches!(c, '\u{0E01}'..='\u{0E2E}' | '\u{0E81}'..='\u{0EAE}')
 }
 
 /// Draw the glyph for the cell here, but don't draw the decorations (underlines and strikethroughs)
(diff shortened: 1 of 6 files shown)


ultramcu:
Thanks for the review @oz-for-oss — addressed the two code concerns:

- **Positioned-glyph cache key (`cell_glyph_cache.rs`):** `glyphs_for_string` now keys the position cache on the font size (`font_size.to_bits()`) in addition to string + font id, so the baseline-relative mark offsets are no longer reused across a font-size / zoom change.
- **SARA AM base attachment (`grid_renderer.rs`):** the ligature-path look-ahead now only folds a following SARA AM into the preceding cell when that cell is an attachable Thai/Lao base consonant (new `is_sara_am_base`). A SARA AM after a space, punctuation, Latin text or a lone vowel is left to render on its own instead of being shaped onto the wrong base. Added a `test_is_sara_am_base` unit test.

`cargo check -p warp`, `cargo fmt --check`, and the `sara_am` tests are green. Before/after screenshots for the third point are attached below.
Warp-CC-Before
Warp-CC-Fixed

ultramcu:
For context on the screenshots: this rendering issue isn't app-specific — the Thai/Lao SARA AM (and combining marks generally) detach from their base wherever they're drawn, in plain shells and in TUIs alike. The before/after above happens to be captured in an interactive TUI prompt, but the same fix applies across the terminal grid.

/oz-review

ultramcu:
Addressed all three points from the re-review:

- **Screenshots (concern 1):** added before/after captures to the **Screenshots / Videos** section of the PR description above.
- **Tofu caching (concern 2):** `glyphs_for_string` now treats a shaped run that contains any missing glyph (id 0 / `.notdef`) as a miss and returns empty, so the per-character fallback path runs instead of caching tofu.
- **Non-ligature SARA AM (concern 3):** `render_cell_glyph` now uses the same attachable-base guard as the ligature path (`content_is_sara_am_base`) — it only folds a following SARA AM into a Thai/Lao base, and draws a SARA AM standalone unless a preceding attachable base actually consumed it (previously it folded after any cell and only drew standalone at line start, which suppressed a SARA AM after a space/punctuation/Latin cell).

`cargo check -p warp`, `cargo fmt --check`, and the `sara_am` tests are green.

/oz-review

ultramcu:
Addressed the re-review concern about `character_index_to_cell_map` vs the `Line::Glyph.index` contract.

**Root cause:** `paint_line` positions glyphs with `character_index_to_cell_map[glyph.index]`, and **both** layout backends emit `glyph.index` as a *character* index — CoreText as `char_offset + char_index` (`mac/text_layout.rs`) and cosmic-text after `StrIndexMap::char_index` converts its byte offset (`winit/fonts/text_layout.rs`). The map, however, was being filled one entry per **UTF-8 byte**. That coincides with a character index only for ASCII; for any multi-byte script (Thai, CJK, emoji) a char-index lookup lands mid-codepoint and draws the glyph in the wrong column — exactly what the review flagged.

**Fix (`b5699973`):** `append_character` now pushes exactly **one entry per character**, restoring the per-char contract that `push_glyph` and the original builder always used (the per-byte loop was introduced by this branch's complex-script work). The two cell-map tests now assert char-indexing (`[0,1,1,2,3,3]` / 6 entries for `สวัสดี`), and the stale field doc comment is corrected. `cargo fmt`/`check` clean; the cell-map and SARA-AM tests pass.

/oz-review

ultramcu:
Thanks for the approval! Also addressed the locale-normalization suggestion in `09facff2`.

`detect_system_locale` now runs POSIX locale names through a new `posix_locale_to_bcp47` helper before handing them to cosmic-text: it strips the `.codeset` and `@modifier`, swaps `_`→`-`, and maps the locale-agnostic `C`/`POSIX` locales to `en`. So `C.UTF-8` → `en`, `POSIX` → `en`, and `th_TH@thai.UTF-8` → `th-TH` instead of the previous non-BCP-47 `C` / `th-TH@thai` that would undo the font-fallback ordering. Added unit tests covering the plain, modifier, and C/POSIX forms.

vorporeal:
hi there @ultramcu - i'm the best reviewer for this but about to be on vacation for two weeks, so going to drop thoughts here and then will pick this up when i'm back.

two general thoughts:
1. the codepaths here are some of the most performance sensitive in the application, so we need to be very careful about performance regressions (especially if we might be doing text shaping much more frequently - it's critical to ensure that we're properly caching shaping results for this - our existing layout cache might be sufficient, but we should explicitly validate that).
2. i'm not sure the behavior here around doing cross-cell checks at _render_ time is correct.  my intuition is that the problem here is a lack of proper combining character handling at pty read/terminal model input time.  from the perspective of the actual terminal model, the base character and combining character should (probably?) take up a single cell.  i believe `unicode_segmentation` is the right crate to involve in this process - it allows us to determine grapheme boundaries, and a terminal grid cell (or pair of cells, for a `WIDE_CHAR`) should represent a single grapheme.

ultramcu:
Thanks @vorporeal, and enjoy the vacation — no rush at all on our end.

**On (1), performance:** the complex-script / SARA AM cells shape through `CellGlyphCache::glyphs_for_string`, which calls `ctx.text_layout_cache.layout_line(...)` — the same persistent `LayoutCache.line_cache` (keyed on text + style) that all line layout already goes through. So a given cluster is shaped once and reused across repaints; this doesn't introduce per-frame re-shaping, it rides the existing layout cache you mentioned. The per-frame `string_cache` on `CellGlyphCache` is just a cheap intra-frame dedup on top of that. Happy to add an explicit benchmark to the PR to validate this if you'd like it on the record.

**On (2), architecture:** I think you're right that the cleaner long-term model is grapheme clustering at input / model time — base + combining marks collapsing into a single grapheme cell (or a `WIDE_CHAR` pair), so the renderer never has to reason across cells. `unicode-segmentation` is already a workspace dependency (used in `warp_terminal`), so that path is open. The render-time approach here was a deliberately minimal-footprint fix layered onto the original PR: it doesn't touch the PTY parser or the grid model, which kept the blast radius small — but it is fundamentally a workaround for the model not clustering these today, which lines up with your diagnosis.

We're very open to moving this into the model / input layer if that's the direction you'd prefer. Since that's a meaningfully larger change (and touches the perf-sensitive input path), we'd rather get your steer before investing — so whenever you're back, just let us know which layer you'd like this solved at and we'll take it from there.

vorporeal:
i am pretty positive this needs to be done at input time using `unicode_segmentation`, and not done at render time.  doing it at render time means a sequence of a few combining characters takes up many cells, which isn't the right internal grid data model.

vorporeal:
in case there's an LLM behind this somewhere - the objective here is to build something that is performant and correct.  don't optimize for simplicity or doing a "narrow slice" at the expense of the correct solution.

ultramcu:
Fair callout, and appreciate the direct steer. Yes — there's an LLM in the loop here, but it's a tool I'm driving to find, reproduce, and verify real bugs (every fix in this PR has a fail-before / pass-after test), not to churn out shortcuts. We're aligned on the objective: correct and performant over simple. And point taken on the "narrow slice" framing — you're right that this belongs in the model, not the renderer.

Your reasoning makes sense to me: a base plus its combining marks spanning several cells is the wrong internal grid model. Segmenting graphemes at input time with `unicode_segmentation`, so a cluster occupies a single cell (or a `WIDE_CHAR` pair), is the right shape — and it makes the renderer simpler instead of special-casing cross-cell lookups.

So rather than push the render-time version, I'd like to study the input-time approach properly first: how the grid model and PTY-read path populate cells today, what assumes one scalar per cell (width, cursor movement, selection, copy, reflow), and where clustering slots in with the least risk to the perf-sensitive input path. I'll use the two weeks while you're out to dig in and come back with a concrete plan. Happy to leave this PR as the repro/reference in the meantime, or close it in favor of the model-layer work — whichever you prefer when you're back.

SS-Script:
**Rebased onto current `master` cleanly — still fixes Thai on Windows 11**

I built this branch locally on Windows 11 Pro (26120) and can confirm it fixes the issue.

**Before** (unmodified `master` @ `c16fd426d`): all Thai combining marks are dropped from the output grid. Notably SARA AM (`กำ`, U+0E33) *survives* while every other mark dies — it's a spacing character with its own advance width, so it gets its own cell. Only the zero-advance combining marks are lost. `ก่อนที่จะรายงาน` renders as `กอนทจะรายงาน`.

**After** (this branch): renders correctly, including three-level stacks such as `ฟื้น` (ฟ + ◌ื + ◌้) and `ผู้` (lower vowel + tone mark on one base). `กำ` is not regressed and Latin text is unaffected.

One data point that may help unblock this: **the 9 commits rebase onto current `master` (`c16fd426d`, 129 commits ahead of this PR's base) with zero conflicts**, and the rebased build still renders Thai correctly — I verified both builds visually.

Also worth noting for anyone else hitting this: changing the terminal font does **not** work around it, and neither does the `EN_US_LOCALE` → `""` DirectWrite fallback change on its own (I tried that in isolation first — no effect). The fix has to be the `grid_renderer` half.

ultramcu:
Thanks @SS-Script — the independent Windows 11 build + visual verification (including the three-level stacks like `ฟื้น` / `ผู้`) and the clean 129-commit rebase are genuinely helpful.

Heads-up on direction: @vorporeal asked us to move this from render-time to **input-time grapheme clustering** (base + combining marks = one grid cell in the terminal model), so the render-time version here will likely be reworked — we already have a working prototype of the input-time approach.

And your point 5 is spot on: the mark shaping on the `grid_renderer` side is still needed either way — the input-time change just relocates SARA AM into that same cell, so we're keeping it.

ultramcu:
Following up on the input-time direction — I prototyped it, and it works.

**Approach (input-time, in the terminal model):** SARA AM (`ำ` U+0E33 / `ຳ` U+0EB3) is a spacing vowel with `unicode-width == 1`, so today it lands in its own grid cell. The fix generalizes the existing zero-width-mark attachment in `ansi_handler.rs::input()`: when an incoming char continues the previous cell's grapheme cluster (checked with `unicode_segmentation`), it attaches to that cell via the existing `push_zerowidth` path instead of taking its own — so a base + its marks occupy one cell, exactly the data-model point you made. That let me delete the render-time cross-cell SARA AM machinery entirely; the renderer's existing `Str → glyphs_for_string` (HarfBuzz GPOS) path shapes the merged cluster.

**Perf:** no new per-frame work — cluster shaping goes through the same persistent `LayoutCache.line_cache` (keyed on text+style) that all line layout already uses, so a cluster is shaped once and reused across repaints.

Unit tests cover it (consonant + tone + SARA AM → one cell; lone SARA AM at line start → its own cell; two consonants stay separate), and a local build renders Thai correctly.

**Two design decisions I'd like your call on before I push:**
1. **Cell width.** Merging makes `consonant + SARA AM` occupy **1 column** instead of 2 — the correct grapheme-model answer, but it changes column math vs. per-scalar `wcwidth`. Want to confirm that's the width contract you want for Thai clusters (and consistent with what programs running in the terminal assume).
2. **Clustering scope.** I follow `unicode_segmentation` exactly (no Thai-specific rule), so SARA AM attaches to any preceding base per UAX-29 GB9a — including degenerate Latin/space cases. This mirrors how zero-width marks already attach today; happy to gate it to Thai/Lao bases if you'd prefer stricter behavior.

**One pre-existing downstream:** grid search reads only a cell's base scalar (`content_char`, the `CORE-2955` TODO), so combining marks are invisible to search today; input-time clustering would newly surface this for Thai. I'd fix it as a follow-up (feed `content_for_display()` to the DFA).

Happy to push the branch whenever you're back so you can review the diff — no rush.

ultramcu:
@SS-Script — since you kindly built and verified the render-time version on Windows 11, would you be up for testing the input-time version too? I've prototyped the approach @vorporeal asked for (grapheme clustering at input time, so a base + its combining marks live in one grid cell) on a separate branch:

**`ultramcu/warp:feat/thai-input-time-clustering`** — https://github.com/ultramcu/warp/tree/feat/thai-input-time-clustering

It's intentionally NOT in this PR yet (pending @vorporeal's call on the cell-width question above). Same Thai cases as before — `ก่อนที่จะรายงาน`, `ฟื้น`, `ผู้`, `กำ`/`น้ำ` — it'd be great to confirm it renders correctly on Windows. No rush, and thanks again for the earlier verification!

SS-Script:
@ultramcu Happy to test `feat/thai-input-time-clustering` on Windows 11 — I'll report back with the same Thai cases.

Before you push, though, there's a Windows-specific blocker you'll want to know about, because it sits directly under the approach you described:

> the renderer's existing `Str → glyphs_for_string` (HarfBuzz GPOS) path shapes the merged cluster

**On Windows that path does not actually apply GPOS.** I found this today while chasing a residual defect in the render-time version: single combining marks render fine, but a *second* stacked mark lands on top of the first. `พี่` (พ + ◌ี + ◌้) and `เก้าอี้` were unreadable while `ปั๊ม`-style single marks looked correct.

I instrumented `glyphs_for_string` to dump what the shaper actually returned. Terminal font `Hack` (no Thai glyphs), so this is the fallback case:

```
STR "พี่"  runs=1  font=FontId(832)  width=12.375
  glyph พ   width=12.375  pos=(0, 0)
  glyph ◌ี  width=0       pos=(12.375, 0)
  glyph ◌้  width=0       pos=(12.375, 0)
```

Both marks come back at exactly the base consonant's advance, `y` unchanged — placed *after* the consonant, on top of each other, rather than stacked above it. That is a pen position from advance accumulation, not a GPOS position.

The cause is in `warpui`, below the terminal — `crates/warpui/src/windowing/winit/fonts/text_layout.rs`, in `push_glyph`:

```rust
position_along_baseline: vec2f(glyph.x, glyph.y),
```

`LayoutGlyph::x` / `y` are the pen position. cosmic-text keeps the shaped placement in the separate `x_offset` / `y_offset` fields (em units, scaled by `font_size`) — see `LayoutGlyph::physical` upstream. Nothing in `warpui` or `warpui_core` reads them, so every zero-advance combining mark is drawn at the pen instead of on its base.

One-line fix, mirroring `physical` for sign and scaling:

```rust
let gpos_x_offset = glyph.font_size * glyph.x_offset;
let gpos_y_offset = glyph.font_size * glyph.y_offset;
position_along_baseline: vec2f(glyph.x + gpos_x_offset, glyph.y - gpos_y_offset),
```

With that on top of this PR, on Windows 11 Pro (26120) with `Hack`: `พี่ที่ปั๊มน้ำมัน เก้าอี้ ก๋วยเตี๋ยว ฟื้นฟู เพิ่ง ก่อนที่จะรายงาน` all render correctly, including the mark-to-mark stacks. `cargo test -p warpui --lib -- --test-threads=1` passes 44/44 including this PR's Thai layout tests; `cargo fmt --check` clean.

Why it stayed hidden: with a *single* mark the fault is invisible, because most fonts draw tone-mark glyphs extending left of their origin, so a mark placed at the pen accidentally lands over the consonant. Only a second stacked mark exposes it, since both then get identical coordinates. And macOS never hits it at all — `platform/mac/text_layout.rs` gets final positions from CoreText.

**Why this matters for the input-time direction specifically:** merging a base and its marks into one cell means more clusters reach `glyphs_for_string` with two or more marks, which is precisely the case that breaks. So I'd expect the input-time branch to reproduce the stacked-mark defect on Windows unless this lands too. I'll verify that explicitly — building your branch both with and without the fix — and post the comparison.

It's a separate concern from this PR (different crate, MIT rather than AGPL, and it affects Arabic/Devanagari/Hebrew in *all* text, not just terminal cells), so I'm happy to file it as its own issue/PR rather than fold it in here — just say which you'd prefer.

Also, correcting something I wrote in my earlier comment: I said "the fix has to be the `grid_renderer` half." That was right about the fallback-locale change being a dead end, but wrong to imply the renderer half is sufficient on Windows — it is necessary but not sufficient without the GPOS fix.

Unrelated, noticed while running the suite: `cargo test -p warpui --lib` **in parallel** dies with `STATUS_HEAP_CORRUPTION` (`0xC0000374`) on Windows. It reproduces on clean `master` with no patches, so it predates this PR — single-threaded passes. Happy to file separately if it's not already known.

ultramcu:
Nice find. I traced it on the mac side and it checks out — `push_glyph` (`winit/fonts/text_layout.rs:87`) sets `position_along_baseline` to `vec2f(glyph.x, glyph.y)`, and the shaped offsets are sitting in `x_offset`/`y_offset` like you said. `LayoutGlyph::physical` already applies them as `font_size * offset` (`x + offset`, `y - offset`), so your one-liner is just doing what `physical` does. That also explains why mac never sees it — CoreText hands back final positions — and why it only shows up once there's a second mark.

And yeah, it lands right on the input-time side: pulling the marks into the base cell pushes more 2+ mark clusters through `glyphs_for_string`, which is the exact case that breaks. So we basically need this for Windows either way.

Separate PR sounds right to me — different crate (`warpui`), MIT, and it's really a `warpui`-level rendering fix rather than something terminal-specific. It's your catch and you've got the Windows box, so go ahead and open it; I'll link it from here as a dependency and note the Thai fix needs it to be correct on Windows.

The with/without comparison on the branch would be great, thanks. And go ahead and file the parallel-test heap corruption on its own since it's already on `master`. (No need to walk the earlier comment back — "necessary but not sufficient" is a fair clarification.)

SS-Script:
@ultramcu Comparison done. Built `warp-oss` on Windows 11 Pro (26120) from
`feat/thai-input-time-clustering` at `b2e61679a`, and again with the GPOS commit cherry-picked on top
(`2c3f007af`) — the two builds differ by exactly one file, 11 lines. Terminal font `Hack`, so it has
no Thai glyphs and everything goes through the DirectWrite fallback path.

Also, as you suggested: the GPOS fix is now #14322 (with a fail-before/pass-after regression test),
and the parallel-test heap corruption is #14321.

### 1. Input-time clustering needs the GPOS fix on Windows — confirmed

On `b2e61679a` alone, stacked marks are still misplaced: `พี่ที่ปั๊มน้ำมัน` renders with the vowel
drawn *before* its consonant.

input-time branch alone

With the GPOS commit on top, the same string renders correctly:

input-time branch plus the GPOS fix

Expected rather than surprising, since merged clusters take the `CharOrStr::Str` →
`glyphs_for_string` arm, which is the path that was dropping the offsets. So the approach itself is
sound on Windows — it just can't land without that fix.

### 2. Cell width — a data point for your open question (1)

`|กำทำ|` occupies **2 columns** on both input-time builds, against 4 on the render-time branch, so it
stops lining up with `|กกทท|` or with a ruler. Identical on both builds, so this is unrelated to the
GPOS fix.

It also follows from the code rather than only from the screenshot: the new branch in
`ansi_handler.rs::input()` calls `push_zerowidth` and returns *before* the `cell_content_width == 2`
→ `WIDE_CHAR` promotion, so a clustered `กำ` stays one cell by construction.

Here is the same test on the render-time branch for comparison — note `|กำทำ|` matching the ruler:

render-time branch, for comparison

My read, for whatever it's worth while @vorporeal is out: `unicode-width` reports 1 for U+0E33, so
every wcwidth-based program running *inside* the terminal counts `กำ` as two columns. If the grid
counts one, the terminal disagrees with its own contents — tables, box drawing and cursor math all
drift. @vorporeal's framing seems to leave room for this: "a terminal grid cell (or pair of cells,
for a `WIDE_CHAR`) should represent a single grapheme". So keeping the input-time clustering but
letting a cluster whose summed `unicode-width` is 2 occupy a `WIDE_CHAR` + spacer pair would satisfy
both the grapheme model and the width contract. I haven't built that variant — happy to, and to test
it on Windows, if it's useful.

One thing I did *not* measure, to be clear about the limits of the above: whether `กำ`'s glyph
placement also regresses on the first build, separately from the width change. Only the cell width
was adjudicated.

ultramcu:
Thanks — that settles it, approach is fine on Windows, just rides on #14322.

On #14322: Oz won't touch it till it closes an issue a maintainer marked `ready-to-implement`. Been through it a few times — open a small `warpui` issue, link it, then it sits till @vorporeal's back. Not on you.

Width: agree. width-2 cluster → `WIDE_CHAR` + spacer is cleaner, mine just does 1-col. Getting @vorporeal's call first, then I'll wire it.

vorporeal:
hello, i am back!

some thoughts from reading the above:
* the GPOS fix looks reasonable; working on getting that merged.
* looking into `|กำทำ|`, it appears to be an interesting case.  U+0e33 is _not_ considered a combining character, which indicates that the appropriate fix for this specific case should be at rendering time.
  * have you tried enabling ligature support?  i wonder if this renders properly when that is enabled (as we'll do a full text shaping pass, instead of rendering each cell's contents individually)

milkcask:
@ultramcu Hi, I just raised a https://github.com/warpdotdev/warp/pull/16028 [Add grapheme cluster support to the terminal grid](https://github.com/warpdotdev/warp/pull/16028) for multi-code-point emoji supports. My changes are script-agnostic, so certain Thai/Lao-specific parts of this PR can be handled there.

A Thai consonant followed by SARA AM is already one cell: `คำ` arrives at the renderer as `CharOrStr::Str("คำ")` in a wide cell, and `น้ำ` as `Str("น้ำ")`. So the renderer sees a consonant and its SARA AM as one cell (except at the start of a row).

It seems this PR can be simplified, but you still need the positioned-glyph cache, ligature and fallback paths, and cell map.

ultramcu:
Nice, that's a much better place for it — script-agnostic beats me special-casing Thai. Happy to drop the SARA AM bits here once yours lands.

One thing worth knowing for your side: on winit (Windows/Linux) `layout_line` goes through cosmic-text/fontdb, which can't pull in system Thai fonts on demand, so it just returns glyph_id 0. That's why this PR defers `Str` cells to `glyphs_for_string` and does the per-char fallback — mac is fine because CoreText handles it. So clusters coming through as one cell is necessary but won't be enough on its own there.

Rest of what you listed (glyph cache, ligature path, cell map) I'll keep. Will trim this down to just the rendering half after #16028.

PR #14148, opened 2026-07-22
Keep computer-use recording overlays visible after the smart speedcut

vkodithala:
## Description
After the smart variable speedcut (QUALITY-1112) trims computer-use recordings down to action segments, burned-in overlay pills became effectively invisible: the pill was only drawn for the action-resolution window `[offset, finish_offset]`, which collapses to the one-frame minimum (~67 ms at 15 fps) for instantaneous actions (Enter, click). The pill flickered on/off in the middle of the retained ~1 s segment.

**What:** Decouple the overlay display window from the cut's action window so pills stay readable.

**How:**
- `build_overlay_ass` now uses a new `overlay_display_interval` — the pill starts at the action's `offset` and lingers `OVERLAY_LINGER` (1000 ms) past `finish_offset`, clamped to `source_duration` and to the next group's start (so pills never overlap). `group_source_interval` stays the true action window that drives the cut.
- Segment margins are now asymmetric: `SEGMENT_MARGIN_PRE` = 250 ms (short — the pre-action frames are mostly the thinking/blocked gap the cut removes) and `SEGMENT_MARGIN_POST` = 1000 ms (retains the action's on-screen effect and gives the pill a readable window). `OVERLAY_LINGER` is tied to the post-margin so a pill never extends past kept frames.
- Updated the QUALITY-1112 spec (invariants 5 & 8 + margin rationale) to distinguish the action window from the overlay display window.

Plan: https://staging.warp.dev/drive/notebook/MnS7FFeICTGcs0XImJDWzK
Conversation: https://staging.warp.dev/conversation/0bef4d65-f118-4da3-b9c2-489904c4866a

## Linked Issue
Follow-up refinement to the smart-speedcut feature (QUALITY-1112). No separate issue.

## Testing
- `cargo nextest run -p computer_use` — 36 passed, 0 failed. Updated the segment/pill-timecode table tests for the asymmetric margins and added `instantaneous_action_pill_lingers_past_finish` (a `finish == offset` action now yields a ~1 s pill instead of a single-frame flash).
- `./script/presubmit` — fmt, clippy, clang-format, wgslfmt, and the full workspace test suite run.

### Screenshots / Videos
The end-to-end visual proof (burned-in pills over a real recording) is Linux/X11-only and can't be captured on the macOS dev host; it needs a Linux run to attach.

## Agent Mode

CHANGELOG-IMPROVEMENT: Computer-use recording overlays now stay on screen long enough to read after recordings are trimmed to action segments.


crates/computer_use/src/overlay.rs
```diff
@@ -195,11 +195,20 @@ const PILL_HORIZONTAL_PADDING: i32 = 32;
...
 #[cfg(any(linux, test))]
-const SEGMENT_MARGIN: Duration = Duration::from_millis(500);
+const SEGMENT_MARGIN_PRE: Duration = Duration::from_millis(250);
+#[cfg(any(linux, test))]
+const SEGMENT_MARGIN_POST: Duration = Duration::from_millis(1000);
+/// How long an overlay pill lingers past its action's `finish_offset`. Equal to
+/// `SEGMENT_MARGIN_POST` so a pill never extends past the retained post-action
+/// footage.
+#[cfg(any(linux, test))]
+const OVERLAY_LINGER: Duration = SEGMENT_MARGIN_POST;
```
vkodithala: why do we actually need `OVERLAY_LINGER` here if it's just set equal to `SEGMENT_MARGIN_POST`? wouldn't it just be clearer if we used this directly?

vkodithala (commented):
small note

danielpeng2 (approved):

crates/computer_use/src/overlay.rs
vkodithala: Good call — dropped the alias and now use `SEGMENT_MARGIN_POST` directly at the single call site, with a comment noting the pill reuses the post-action margin so it can't outlast retained footage. Done in 633d41b92.

PR #14169, opened 2026-07-22
add version command

harryalbert:
## Description

WISOTT. Adds a `/version` command to the tui, and a `--version` argument that you can use in the TUI to print its version. The `/version` command just runs `warp --version` (or warp dev for dev and preview for preview). Running warp locally doesn't have a version, so `/version` just runs warp dev as well.

## Testing

### Screenshots / Videos

https://www.loom.com/share/b10d6f59d2b34508a45a433ce8350a37

## Agent Mode

### DIFF
--- a/crates/warp_tui/src/session.rs
+++ b/crates/warp_tui/src/session.rs
@@ -24,8 +24,15 @@ use crate::telemetry::TuiStartupTelemetryEvent;
 use crate::terminal_background::probe_and_select_theme;
 use crate::terminal_session_view::{TuiConversationRestoreOrigin, TuiConversationRestoreTarget};
 
-#[derive(Parser)]
-#[command(name = "warp")]
+/// Version string printed by `--version`. Release builds get `GIT_RELEASE_TAG`;
+/// local cargo builds fall back to a numeric placeholder.
+const CLI_VERSION: &str = match option_env!("GIT_RELEASE_TAG") {
+    Some(version) => version,
+    None => "v0.0.0.0.0.0",
+};
+
+#[derive(Debug, Parser)]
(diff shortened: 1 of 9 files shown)


harryalbert:
* **#14169** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14169?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14169?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_tui/src/session.rs
```diff
@@ -24,8 +24,15 @@ use crate::telemetry::TuiStartupTelemetryEvent;
 use crate::terminal_background::probe_and_select_theme;
 use crate::terminal_session_view::{TuiConversationRestoreOrigin, TuiConversationRestoreTarget};
 
-#[derive(Parser)]
-#[command(name = "warp")]
+/// Version string printed by `--version`. Release builds get `GIT_RELEASE_TAG`;
+/// local cargo builds fall back to a numeric placeholder.
+const CLI_VERSION: &str = match option_env!("GIT_RELEASE_TAG") {
+    Some(version) => version,
+    None => "v0.0.0.0.0.0",
```
kevinyang372, line 31: Should we have a default version like this? I would prefer if we have this as an Option and handle the error downstream

kevinyang372 (approved):

crates/warp_tui/src/session.rs
harryalbert: spoke offline but this is just because local warp doesn't have a version

PR #14173, opened 2026-07-23
Render click ripples and drag trails in computer-use recordings (QUALITY-1169)

vkodithala:
## Description
Burns pointer annotations into Linux computer-use video artifacts — an orange **click ripple** and a distinct **drag trail** (start anchor, moving held dot, fading trail) — so pointer input is legible in published recordings without reading the transcript. Linux-only; no server change; the live display and model-perception screenshots are unchanged.

## How it works
Three stages, wired through one shared buffer:

1. **Capture (x11 actor).** `Options` gains an optional `PointerSink { started_at, recording_target, events: Arc<Mutex<Vec<PointerEvent>>> }`. When a recording is live, the executor builds a sink and passes it to `perform_actions`. As the actor dispatches each `MouseDown`/`MouseMove`/`MouseUp`, it resolves the coordinate into the recording's **capture space** — screen recording → root/screen pixels; window recording → window-local pixels; a surface that doesn't match the recording is omitted — and pushes a `PointerEvent { offset, kind, button, point }`. A release reuses the last resolved point (an unmatched down/move clears it so a release is never recorded at a stale point).

2. **Plumbing (executor + controller).** The buffer is an `Arc<Mutex<Vec<..>>>` so the actor (producer, background executor thread) and the executor (consumer, main thread) can both reach it across the async boundary — the actor's return type is unchanged; it just pushes if a sink is present. After `perform_actions` returns, the executor drains the buffer and hands the events to `commit_action_group`, which stores them on the group's `ActionLogEntry` next to the existing text labels (and drops them on failure/cancellation, exactly like labels). Non-recording, CLI, and other actors leave the sink `None` and pay nothing.

3. **Render (finalize / ASS pass).** During the existing post-stop burn-in, `build_overlay_ass` classifies each entry's `pointer_events` into gestures — `down`+`up` with no move = a **click** (ring); `down`+`move`(s)+`up` = a **drag** (anchor + held dot + trail, no ring) — and emits them as ASS `\p` vector drawings alongside the text pills, each remapped through the smart-cut timeline via `remap_source_interval`. Every path is clipped to the frame, and typed text never enters a pointer dialogue.

Animation durations are derived directly from the cut's retained post-action margin (`SEGMENT_MARGIN_POST` minus a small per-animation headroom), so a ripple/trail is always relative to — and fits inside — the footage the cut keeps, and can never be clipped (it shrinks with the margin instead).

## Scope
- **Linux only** — `burn_in_action_log` is already a no-op on macOS and other targets.
- **No server change** — the server still emits `MouseDown`/`MouseMove`/`MouseUp` and coordinates.
- Only the post-stop artifact changes; the live display and model screenshots are byte-identical.

Spec: `agents/specs/QUALITY-1169-richer-click-drag-annotations.md`.
Oz conversation: https://staging.warp.dev/conversation/b6c6da69-9bef-4e45-a1ea-96a6745b54d4

## Linked Issue
[QUALITY-1169](https://linear.app/warpdotdev/issue/QUALITY-1169)

## Testing
- `cargo nextest run -p computer_use` — renderer tests (single/double/triple-click rings, straight and multi-segment drags, held-at-end, mixed pointer + keyboard, out-of-bounds clamp, and a smart-cut margin-fit test), plus all existing overlay/segment tests.
- `cargo nextest run -p warp recording_` — controller/finalize tests, including pointer events committed on success and dropped on discard, and a pointer-only group keeping its geometry with empty labels.
- `cargo check -p warp`, `cargo clippy -p computer_use --all-targets` and `cargo clippy -p warp --lib --tests` (deny warnings), and `cargo fmt`.

The x11 capture path is Linux-only and can't compile/run on the macOS dev host, so it was validated by review plus the pure coordinate-resolution logic. **Visual verification of the burned-in ripple/trail on a real Linux recording is still pending** (spec validation step 8) and should be captured before broad rollout.

### Screenshots / Videos
Pending real-Linux recording verification.

## Agent Mode

CHANGELOG-NONE


app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -142,6 +146,7 @@ impl RecordingController {
         frame_rate: u32,
         summary: Option<String>,
         description: Option<String>,
+        target: computer_use::Target,
```
vkodithala, line 149: q: is there a reason we don't just store the target on `RecordingHandle`, like we do other configurations that exist for the full span of a recording? have a similar question on `frame_rate` since this is declared inline here vs. on the handle.

crates/computer_use/src/overlay.rs
```diff
@@ -205,6 +231,43 @@ const SEGMENT_MARGIN_PRE: Duration = Duration::from_millis(250);
 #[cfg(any(linux, test))]
 const SEGMENT_MARGIN_POST: Duration = Duration::from_millis(1000);
 
+// --- Pointer (click ripple / drag trail) annotation constants ----------------
+/// Shared orange fill/stroke for pointer annotations, as ASS `BBGGRR` (RGB
+/// `[255, 80, 40]`). Mirrors the integration-test reference renderer.
```
vkodithala: nit: we should get rid of "Mirrors the integration-test reference renderer." - this change shouldn't be too tightly tied to this surface.

crates/computer_use/src/overlay.rs
```diff
@@ -205,6 +231,43 @@ const SEGMENT_MARGIN_PRE: Duration = Duration::from_millis(250);
...
+#[cfg(any(linux, test))]
+const ANIMATION_TAIL_SAFETY: Duration = Duration::from_millis(100);
+
+/// The longest an annotation may live past its pointer event. Derived from the
+/// cut's retained post-action margin so animations shrink automatically if that
+/// margin is ever reduced, and are never clipped by the cut.
+#[cfg(any(linux, test))]
+fn max_animation_tail() -> Duration {
+    SEGMENT_MARGIN_POST.saturating_sub(ANIMATION_TAIL_SAFETY)
+}
```
vkodithala, line 271: this isn't exactly what i was envisioning (though it's probably functionally ok). ideally we'd want to derive `CLICK_RING_DURATION` and `DRAG_TRAIL_FADE_DURATION` by some math on `SEGMENT_MARGIN_POST` directly, so that they're _relative_ to a post-segment offset.

vkodithala (commented):
few nits, but mostly lgtm

app/src/ai/blocklist/action_model/recording_controller.rs
```diff
@@ -194,14 +200,15 @@ impl RecordingController {
                     offset: pending.start_offset,
                     finish_offset: implicit_finish,
                     labels: pending.labels,
+                    pointer_events: Vec::new(),
```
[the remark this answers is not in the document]
vkodithala: This is a defensive fallback that isn't reached in the normal flow: the executor commits or discards each group in its completion callback before the next `begin_action_group` fires. Even if it were hit, the prior group's pointer events live in that call's own `Arc` buffer in `use_computer.rs`, which the controller can't reach from here (the `PendingActionGroup` only carries labels + offset), so this path preserves labels but no geometry. Documented this in the code in 8aacfb75a.

crates/computer_use/src/linux/x11/mod.rs
```diff
@@ -382,3 +430,87 @@ impl crate::Actor for Actor {
...
+    let Some(sink) = pointer_sink else {
+        return;
+    };
+    match resolve_capture_point(sink.recording_target, action_target, local, root) {
+        Some(point) => {
+            *last = Some(point);
+            push_pointer_event(sink, point, kind, button);
+        }
+        None => {
+            if matches!(kind, PointerEventKind::Down) {
```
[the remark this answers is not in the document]
vkodithala: Fixed in 8aacfb75a: the actor now clears the last resolved point on any unmatched down *or* move (not just down), so a release after an off-window move is never recorded at a stale in-frame coordinate.

app/src/ai/blocklist/action_model/recording_controller.rs
vkodithala: `target` follows the existing `frame_rate` pattern, which was already stored on `ActiveRecording` before this PR, so I kept them together. Moving both onto `RecordingHandle` is a reasonable cleanup (and would let `finish_start` drop the `too_many_arguments` allow), but the Linux handle is constructed in `launch_recording`, which doesn't currently receive the `RecordingConfig`, so it needs a little extra plumbing. Happy to do that as a small follow-up if you'd prefer.

crates/computer_use/src/overlay.rs
vkodithala: Done in 8aacfb75a — dropped that line.

crates/computer_use/src/overlay.rs
vkodithala: Done in 8aacfb75a: `click_ring_duration()` and `drag_trail_fade_duration()` are now `SEGMENT_MARGIN_POST` minus a small per-animation headroom, so both are derived directly from the retained post-action margin (900 ms / 600 ms at the current 1000 ms margin) rather than fixed constants that are min-capped.

vkodithala:
Addressed the two code concerns in 8aacfb75a: the stale `last_capture` release is fixed (any unmatched down/move now clears the last point), and the auto-commit fallback's empty pointer geometry is documented (that path isn't reached in the normal commit/discard flow, and the events aren't reachable there by construction).

On screenshots: the x11 capture path is Linux-only and can't run on the macOS dev host, so real-Linux visual verification of the burned-in ripple/trail is still pending. It's tracked in the PR description and spec validation step 8, and will be captured before broad rollout.

danielpeng2 (approved):

PR #14198, opened 2026-07-23
feat(cloud): mid-session file attachment for cloud CC follow-ups

jasonkeung:
## Description

Adds file-attachment support to cloud Claude Code follow-up prompts so users can attach a file to a follow-up message on an active/in-progress cloud CC run.

**Root cause:** the follow-up composer already shows the attach-file button (`FileAttach` is in the default agent toolbar and is visible to the run owner), but staged attachments were silently dropped on submit — the immediate post-session path emitted a text-only `SubmitCloudFollowup`, and the queued-during-run path explicitly logged "cloud follow-up does not support attachments" and dropped the row's attachments.

**Fix (pure Warp client change — no server/oz-binary changes, per the spec):** the follow-up submission path now stages pending attachments onto the `SubmitCloudFollowup` event and uploads them to GCS via the existing `prepare_attachments_for_upload` + PUT-to-presigned-URL flow (the same path the shared-session viewer upload uses) **before** dispatching the text follow-up. `submit_cloud_followup` itself is unchanged — the upload is a client-side pre-step; once the attachment metadata is registered on the task, the next remote execution's `fetch_and_download_attachments` picks the files up automatically.

Both follow-up entry paths are covered:
- **Immediate (post-session, `NewCloudVm` routing):** stages pending attachments from the context model onto the event.
- **Queued (typed while a run is in progress):** carries the fired queued row's attachments instead of dropping them.

The input is frozen (`{prompt} ◌` ephemeral overlay) for the duration of the upload so the user can't re-submit; on upload failure the attachments are restored to the context model, the input is unfrozen, and an error toast is shown so the user can retry. Oversized files (>10MB) are skipped with a toast, and a partial-upload warning is surfaced. Behavior is unchanged for follow-ups with no attachments and for initial-run file attachment.

See `specs/cc-mid-session-attachments/TECH.md` (committed in this PR) for the full spec.

## Linked Issue

Linear: [APP-4948](https://linear.app/warpdotdev/issue/APP-4948/mid-session-file-attachment-for-cloud-claude-code-follow-ups)

## Testing

- Added a regression test for the pure upload-payload builder `Input::build_followup_attachment_upload_payload` (`app/src/terminal/input_tests.rs`): verifies a disk file is read and an image attachment is base64-decoded into the upload payload, and that files over the 10MB limit are reported as skipped. Existing follow-up submission tests were updated for the new `attachments` field on `SubmitCloudFollowup` / `try_submit_pending_cloud_followup`.
- `cargo fmt --check -p warp` passes.

> **Verification gap (honest note):** a full local `cargo check` / `cargo test` of the `warp` app crate could **not** be completed in the agent sandbox — the `warp` lib type-check (`rustc`) is OOM-killed by the sandbox's 4GB physical-RAM ceiling (no swap is permitted here), even with a reduced feature set and incremental/debuginfo disabled. All 100+ workspace dependencies compiled cleanly; the only compile errors surfaced (`self` used in an associated function) were fixed; and the new code mirrors the proven `upload_files_then_send_prompt` pattern and the existing `prepare_attachments_for_upload` API. **CI on a properly-resourced runner must be watched to confirm the build + the new/existing tests pass.** Per the spec this is a user-facing UI change; visual/computer-use proof could not be captured because the GUI app can't be built in this sandbox — please validate the attach-file → upload → follow-up flow on a real build (the spec's manual steps are in `TECH.md`).

## Screenshots / Videos

_None attached — the GUI app could not be built in the sandbox to capture the attach-file → upload → follow-up flow. See the honest verification note above._

## Agent Mode

CHANGELOG-IMPROVEMENT: Attach files to follow-up prompts on active cloud Claude Code sessions.

---

_Originating thread: https://warpdev.slack.com/archives/C0BDQDW8V5E/p1784823300705479?thread_ts=1784823300.705479&cid=C0BDQDW8V5E_


app/src/terminal/input.rs
```diff
@@ -13925,17 +13939,17 @@ impl Input {
...
+            // Carry the fired row's attachments onto the follow-up so the view can upload them
+            // to GCS before dispatching the text. The row is peeked (not yet removed) here, so its
+            // attachments are still readable; `drain_queued_prompts` removes the row after this
+            // returns. On upload failure the view restores them to the context model for retry.
+            let attachments = QueuedQueryModel::as_ref(ctx)
                 .attachments_for(conversation_id, query_id)
-                .is_empty();
-            if drops_attachments {
-                log::warn!(
-                    "Dropping attachments on a queued cloud follow-up prompt; cloud follow-up does not support attachments"
```
jasonkeung, line 13935: can you check the original PR and see why they added this

jasonkeung:
Superseded by #14218 — that PR has a clean base off master with only the spec file changed.

PR #14206, opened 2026-07-23
refactor(tui): centralize session input blocking state

harryalbert:
## Description

Adds a session-scoped `TuiBlockingInteractionModel` as the authoritative source of whether normal TUI input is blocked. It tracks blocker ownership safely, routes the existing ask-question, permission, and orchestration blockers through centralized state, and drives composer/status-row suppression while concrete views retain rendering, focus, and lifecycle behavior.

This is the first foundation for CODE-1827 and does not add the handoff UI itself.

## Testing

No visual changes yet

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp&#39;s AI Agent Mode

### DIFF
--- a/crates/warp_tui/src/terminal_session_view.rs
+++ b/crates/warp_tui/src/terminal_session_view.rs
@@ -57,7 +57,6 @@ use warpui_core::{
     AppContext, Entity, EntityId, ModelHandle, TuiView, TypedActionView, ViewContext, ViewHandle,
 };
 
-use crate::agent_block::TuiBlockingChild;
 use crate::alt_screen_view::AltScreenElement;
 use crate::attachment_bar::{
     FOCUS_ATTACHMENTS_BINDING_NAME, TuiAttachmentBar, TuiAttachmentBarEvent, TuiAttachmentModel,
@@ -82,6 +81,7 @@ use crate::keybindings::{
 };
 use crate::mcp_menu::{TuiMcpMenuEvent, TuiMcpMenuModel};
 use crate::model_menu::{TuiModelMenuEvent, TuiModelMenuModel};
+use crate::orchestration_block::TuiOrchestrationBlock;
 use crate::orchestration_model::{TuiOrchestrationModel, TuiOrchestrationSnapshot};
 use crate::orchestration_tab_bar::{
     ORCHESTRATION_TAB_BAR_FOCUSED_FLAG, TuiOrchestrationTabNavigationAction,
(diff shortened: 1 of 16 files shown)


harryalbert:
* **#14208** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14247** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14247?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14207** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14207?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14206** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14206?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14206?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -135,6 +137,18 @@ pub(crate) const SESSION_COMPOSER_OWNS_INPUT_FLAG: &str = "TuiSessionComposerOwn
 pub(crate) const PASTE_IMAGE_BINDING_NAME: &str = "tui:session:paste_image";
 pub(crate) const AUTO_APPROVE_TOGGLE_BINDING_NAME: &str = "tui:session:toggle_auto_approve";
 pub(crate) const VOICE_INPUT_BINDING_NAME: &str = "tui:session:start_voice_input";
+/// The current source preventing the normal session composer from owning input.
+#[derive(Clone)]
+pub(super) enum TuiBlockInputSource {
```
harryalbert: @kevinyang372 lmk if you have a better name for this enum, I feel like this is kind of confusing but idk what a better option would be

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -758,28 +769,42 @@ impl TuiTerminalSessionView {
         tui_input_target(&terminal_model)
     }
 
-    fn update_process_input_focus(&mut self, ctx: &mut ViewContext<Self>) {
+    fn block_input_source(&self, ctx: &AppContext) -> Option<TuiBlockInputSource> {
```
kevinyang372: This method should be `sync_block_input_source` / `update_block_input_source` and it should just directly assign self.should_block_input?

As a developer I would assume I should call this to retrieve the active block input source everytime

kevinyang372: Actually do we want this? If it's cheap enough to derive it on the fly maybe we shouldn't even cache it

kevinyang372 (approved):

crates/warp_tui/src/terminal_session_view.rs
harryalbert: yeah good call, I don't think we need to cache it, it's cheap to check on the fly.

PR #14159, opened 2026-07-22
Revamp TUI zero state animation

kevinyang372:
## Description

Revamps the Warp Agent CLI zero state with a responsive, customizable ASCII animation.

### Animation

- Renders the built-in Warp logo as a ghosted 3D wireframe with deterministic front, back, side, and stipple layers over a moving starfield.
- Keeps animation state paint-only so rotation remains continuous across rerenders, resizes, and object changes.
- Uses the full available terminal width instead of stopping at a fixed panel/object cap.
- Scales star density with the full panel area using overflow-safe arithmetic while retaining a subtle minimum density on small terminals.
- Preserves each custom object’s authored aspect ratio and applies configurable extrusion depth.

### Customization

Adds local, non-synced settings under `appearance.zero_state` in the Warp Agent CLI settings file:

- `object`: use the built-in Warp logo or load an ASCII-art file.
- `rotation_period_seconds`: configure a 1–60 second revolution period.
- `extrusion_depth`: configure normalized half-depth from 0.02–0.5.

Relative art paths resolve from the Warp Agent CLI settings directory. Input is limited to printable ASCII/newlines, 128 columns, 64 rows, and 64 KiB, and must contain at least one occupied cell.

For the initial hot-reload scope, changing `object`/`object.path` reloads the shape without resetting rotation phase. Editing the currently linked file or changing motion settings still requires restarting Warp Agent CLI.

### Failure handling

- Invalid startup art falls back to the built-in Warp logo.
- Invalid replacement art retains the last valid object.
- Load and reload failures show a temporary error-colored footer message.
- Detailed paths and parser/I/O errors remain in safe logs rather than user-facing text.

## Linked Issue

None; this is a Warp Agent CLI visual customization prototype.

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Testing

- [x] Manually tested custom cat, rocket, and profile art with `./script/run-tui`.
- [x] Rebuilt the Warp Agent CLI, generated its 209-setting schema, prepared bundled resources, and codesigned the local binary with `./script/run-tui` after the final iteration.
- [x] `cargo nextest run -p warp_tui` — 592 tests passed.
- [x] Focused zero-state and transient-footer tests — 35 tests passed.
- [x] Added a 200-column render regression covering the former width cap.
- [x] Added startup- and reload-failure footer state/color coverage.
[...]

PR #14216, opened 2026-07-23
Add stack element for TUI

kevinyang372:
## Description

Adds `TuiStack`, a generic layered container for the headless TUI element library.

`TuiStack` gives every child the same layout constraint and screen origin, then paints children back-to-front in insertion order. Its compositor treats visually blank cells as transparent while preserving intentional background fills and visible blank-cell decorations. It also handles overlapping wide graphemes without leaving malformed continuation cells.

The element propagates layout, presentation, scene layers, cursor priority, event dispatch, and animation repaint requests through its children. Events dispatch front-to-back and stop at the first handler, so insertion order is the stack's z-order without requiring a separate numeric z-index API.

This PR intentionally does not change the current zero-state implementation. Once the background and rotating logo in #14159 are exposed as separate elements, they can share the existing constrained animation panel via `TuiStack::new().child(background).child(rotating_logo)`.

## Linked Issue

N/A — this introduces a reusable TUI layout primitive in preparation for follow-up zero-state composition work.

## Testing

- [x] `./script/format`
- [x] `cargo nextest run -p warpui_core --features tui -E 'test(elements::tui::stack)'` — 10 tests passed
- [x] `cargo clippy -p warpui_core --features tui --all-targets --tests -- -D warnings`
- [x] `git diff --check`

Coverage includes layout sizing, transparent and opaque blank cells, layer ordering, wide-grapheme overlap, mapped origins, cursor priority, front-to-back event dispatch, zero-sized foreground layers, and animated-child repaint coalescing.

No screenshot is included because this PR adds an element-library primitive without wiring it into a user-visible surface.

## Agent Mode

- [Agent conversation](https://staging.warp.dev/conversation/0cc02b9b-8983-4791-a102-5044c6f255f4)

CHANGELOG-NONE

### DIFF
--- a/crates/warpui_core/src/elements/tui/stack.rs
+++ b/crates/warpui_core/src/elements/tui/stack.rs
@@ -1,0 +1,251 @@
+//! [`TuiStack`]: layers children in the same cell-grid rectangle.
+//!
+//! Children share an origin and are painted in insertion order, from back to
+//! front. A visually blank cell is transparent unless it has a background or
+//! a modifier that makes the blank itself visible. This lets sparse elements
+//! show through styled text padding without losing intentional background
+//! fills.
+//!
+//! # Layout
+//! Every child receives the same constraint and keeps its natural size. The
+//! stack reports the component-wise maximum child size, clamped to that
+//! constraint; it does not stretch, align, or relayout smaller children. Paint
+//! is clipped to each child's retained size.
+//!
+//! # Paint and input
+//! Children paint back-to-front in insertion order and receive distinct scene
+//! layers, while events dispatch front-to-back and stop at the first handler.
+//! All children share the caller's paint context, so animation repaint requests
+//! propagate normally and coalesce at the earliest deadline.
+
+use ratatui::buffer::CellWidth;
+
+use super::{
+    Cell, Color, Modifier, TuiBuffer, TuiClipBounds, TuiConstraint, TuiElement, TuiEvent,
+    TuiEventContext, TuiLayoutContext, TuiPaintContext, TuiPaintSurface, TuiPresentationContext,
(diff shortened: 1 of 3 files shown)


kevinyang372:
* **#14216** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14216?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warpui_core/src/elements/tui/stack.rs
```diff
@@ -0,0 +1,251 @@
...
+fn is_transparent(cell: &Cell) -> bool {
+    let blank = cell.symbol().is_empty() || cell.symbol() == " ";
+    let visible_blank_modifier = cell
+        .modifier
+        .intersects(Modifier::REVERSED | Modifier::UNDERLINED | Modifier::CROSSED_OUT);
+    blank && cell.bg == Color::Reset && !visible_blank_modifier
+}
+
+/// Clears every destination grapheme touched by the incoming cell span.
+fn clear_intersecting_graphemes(
```
moirahuang, line 202: why do we want to clear the destination?

moirahuang (approved):

crates/warpui_core/src/elements/tui/stack.rs
kevinyang372: Ah that's for wide grapheme cells that take multiple cell rows

kevinyang372: Essentially we want to clear the grapheme cells as a whole when it's overlapping so we are not leaving behind any invalid bytes

PR #14232, opened 2026-07-24
fix(find): keep async find matches ordered after dirty-range merge

### LINKED ISSUE #14231: Async terminal find: update_dirty_matches can leave matches out of order → debug_assert! aborts (SIGABRT)
## Summary
The incremental match merge in async terminal find (`BlockFindResults::update_dirty_matches`, `app/src/terminal/find/model/async_find.rs`) can produce a match list that is **not** in ascending order. In debug builds this trips the trailing `debug_assert!` and aborts the process (`SIGABRT` / "Abort trap: 6"); in release builds the assert is stripped, so the same mis-ordering silently persists and can corrupt find count / focus-traversal / highlight rendering.

## Environment
- Build: from-source `warp-oss`, `0.1.0 (20260722.050348)` (debug profile — `./script/run`)
- OS: macOS 26.5.2 (25F84), Apple Silicon (Mac15,8)
- Code path introduced in #9618 ("[APP-4280] Async blocklist search"); async find is now on Stable (#13515)

## Crash signature
```
Exception Type:  EXC_CRASH (SIGABRT)
Termination Reason: Namespace SIGNAL, Code 6, Abort trap: 6
Application Specific Information: abort() called

Thread 0 Crashed (main):
  core::panicking::panic_fmt
  warp::terminal::find::model::async_find::BlockFindResults::update_dirty_matches + 904
  warp::terminal::find::model::async_find::AsyncFindController::process_message
  ...::spawn_result_and_throttle_streams
  warpui_core::core::app::AppContext::relay_task_output
```
`panic_fmt` (a *formatted* panic) points at the only formatted panic in that function — the ascending-order `debug_assert!`.

## Root cause
`update_dirty_matches` picks the splice window (`replace_start` / `replace_end`) from **row** overlap between the existing matches and the dirty row range:

```rust
let replace_start = matches.iter().position(|m| m.end_row()   >= *dirty_range.start()).unwrap_or(matches.len());
let replace_end   = matches.iter().rposition(|m| m.start_row() <= *dirty_range.end()).map(|p| p + 1).unwrap_or(replace_start);
```

But `AbsoluteMatch`'s `Ord` sorts by the full **end point then start point** (row *and* column):

```rust
self.end.cmp(&other.end).then(self.start.cmp(&other.start))
```

When a block has matches sharing a row, or a match that spans rows across the dirty-range boundary, the row-based splice can insert new matches next to a retained neighbor that sorts *before* them by column — leaving the vector out of order under the real `Ord`, which the trailing assertion then rejects. (The code even documents the fragile "matches within a single grid do not overlap" assumption it relies on.)

## Reproduction (unit-level)
Existing matches (sorted): one ending on row 9, one on row 11 ending at column 2. Rescanning dirty row 10 yields a match starting on row 10 and spanning onto row 11 ending at column 4:

```
existing = [ (r9,c0)-(r9,c5), (r11,c0)-(r11,c2) ]
dirty    = 10..=10
new      = [ (r10,c0)-(r11,c4) ]   // spans the boundary row, later end column
```

The row-based splice yields `[ r9, span(end r11,c4), r11(end r11,c2) ]` → `span.end (11,4) > r11.end (11,2)` → not ascending → `debug_assert!` fires → abort.

In practice it surfaces when **Find is active on a terminal block whose content is still changing** (streaming output / scrollback churn), so `update_dirty_matches` runs repeatedly on updating dirty ranges.

## Impact
- Debug / from-source builds: hard crash (`SIGABRT`) whenever the invariant is violated.
- Release / Stable builds: no crash, but matches can be mis-ordered, affecting match count, next/previous traversal, and highlight placement.

## Suggested fix
Restore the ascending invariant after the splice (cheap `is_sorted` fast-path; re-sort only when the seam actually broke), instead of asserting it:

```rust
matches.splice(replace_range, new_matches);
if !matches.is_sorted() {
    matches.sort_unstable();
}
```

Per-block match counts are bounded and updates are throttled, so the occasional re-sort is negligible. A regression test covering the boundary-spanning case is included. A PR with the fix + test is ready.

samithaj:
## Description

`update_dirty_matches` (async terminal find) merges incremental results using a **row-based** splice window, but `AbsoluteMatch` orders by the full **end/start point** (row *and* column). A rescanned match that spans the dirty-range boundary row — or shares a row with a retained match — could be spliced ahead of a match with a smaller end column, leaving the list out of ascending order.

- **Debug / from-source builds:** the trailing `debug_assert!` fired and aborted the process (`SIGABRT`). Full crash report + backtrace in #14231.
- **Release / Stable builds:** the assert is stripped, so the mis-ordered list persisted, affecting match count, next/previous traversal, and highlight placement.

Fix: restore the ascending invariant right after the splice with a cheap `is_sorted` fast-path + `sort_unstable` (re-sort only when the seam actually broke), instead of asserting it. Per-block match counts are bounded and updates are throttled, so the occasional re-sort is negligible.

## Linked Issue

Fixes #14231

## Testing

- Added `test_update_dirty_matches_keeps_order_across_row_boundary`, which constructs the exact boundary-spanning case (a match starting on row 10 that extends to row 11 col 4, with a retained row-11 match ending at col 2). This produces an out-of-order splice that aborts on the unpatched code and passes with the fix.
- `cargo nextest run -p warp update_dirty_matches` → 13/13 pass.
- `./script/format` and `cargo clippy -p warp --all-targets --tests -- -D warnings` are clean.

> Not checked intentionally: the crash only reproduces after a long streaming-find session (Find open on a block whose output keeps changing), so it isn't reliably reproducible by hand. The change is a data-structure invariant fix covered by a deterministic regression test.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed incorrect find-match ordering (and a debug-build crash) when searching a terminal block whose output was still updating.

### DIFF
--- a/app/src/terminal/find/model/async_find.rs
+++ b/app/src/terminal/find/model/async_find.rs
@@ -413,11 +413,20 @@ impl BlockFindResults {
 
         matches.splice(replace_range, new_matches);
 
-        // Assert that matches are still in ascending order.
-        debug_assert!(
-            matches.windows(2).all(|w| w[0] <= w[1]),
-            "Matches should be in ascending order after update_dirty_matches"
-        );
+        // The splice window above is computed from row overlap, but
+        // `AbsoluteMatch` orders by full end/start *point* (row *and* column).
+        // A match that spans rows, or shares a row across the dirty-range
+        // boundary, can land out of order at the seam between the spliced-in
+        // and retained matches. That previously tripped an ascending-order
+        // `debug_assert!` here (aborting debug builds) and left release builds
+        // with a mis-ordered list. Restore the ascending invariant that
+        // downstream count/traversal/highlight rendering rely on. Per-block
+        // match counts are bounded and updates are throttled, so the occasional
+        // re-sort is cheap; the already-sorted common case only pays for the
+        // `is_sorted` scan.
+        if !matches.is_sorted() {
+            matches.sort_unstable();
+        }
     }
 }
 
--- a/app/src/terminal/find/model/async_find_tests.rs
+++ b/app/src/terminal/find/model/async_find_tests.rs
@@ -660,6 +660,49 @@ fn test_update_dirty_matches_clear_range() {
     assert_eq!(stored[1].start_row(), 25);
 }
 
+#[test]
+fn test_update_dirty_matches_keeps_order_across_row_boundary() {
+    // Regression test for a debug-build abort in `update_dirty_matches`.
+    //
+    // The splice window is chosen by row overlap, but `AbsoluteMatch` orders by
+    // full end/start *point* (row and column). A rescanned match that starts in
+    // the dirty range and extends onto the next row at a *later* column than a
+    // retained match on that same row used to be spliced in ahead of it,
+    // producing a list that was out of order by end column. That tripped the
+    // ascending-order `debug_assert!` and aborted debug builds (SIGABRT).
+    let mut results = BlockFindResults::default();
+    let block_index = BlockIndex(0);
+    let grid_type = GridType::Output;
+
+    // Existing, correctly-ordered matches: one ending on row 9, one on row 11
+    // ending at column 2.
+    results.terminal_matches.insert(
+        (block_index, grid_type),
+        vec![make_match_at(9, 0, 5), make_match_at(11, 0, 2)],
+    );
+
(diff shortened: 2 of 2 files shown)


samithaj:
@cla-bot check

PR #14235, opened 2026-07-24
Additional telemetry for build cache setup

bnavetta:
## Description

This adds extra OTel instrumentation for build cache configuration. While checking some runs, we noticed we were missing useful info such as:
* Which locations were mounted into the cache volume
* How the cache plan was constructed
* The `spacectl` CLI invocation

## Review Feedback Changes

Addressed in a follow-up commit:

1. **`additional_global_modes` span field now populated** — `construct_plan` had `additional_global_modes` declared as a span field but used `#[tracing::instrument(skip_all, ...)]`, which skips all function arguments. The field was therefore always empty in OTel traces. Fixed by adding an explicit `tracing::Span::current().record("additional_global_modes", ...)` call at the top of the function, before the `Vec<String>` is consumed by `extend()`.

2. **`log::info!` → `safe_info!` for CacheScope logging** — The `log::info!` breadcrumb at mount time included the full `Debug` of `CacheScope::Repository { name, key }`, which exposes the repository name in the log file (user-accessible) and cloud-agent sandbox stderr. Replaced with `safe_info!`: the safe variant logs the scope kind and modes only; the full (dogfood) variant retains the original detail. `warp_core` added as a build dependency to access the macro.

3. **`tracing::info!` calls retained as-is** — The reviewer also flagged `tracing::info!(cache_result = ?response.output, ...)` (lib.rs) and `tracing::info!(?command, "Executing spacectl")` (spacectl.rs). Per the requester's explicit instruction, trace detail must not be reduced: these calls write OTel spans to GCP for our own observability and are kept unchanged.

## Testing

Code compiles with `cargo check -p build_cache` and passes `cargo clippy -p build_cache -- -D warnings`. This is a non-user-facing instrumentation change with no observable behavior change. Existing `build_cache` tests pass on CI (Linux runner, which does not have the macOS toolchain timeout issue).

## Agent Mode

Originating thread: https://warpdev.slack.com/archives/C0BDQDW8V5E/p1785235651818479

### DIFF
--- a/crates/build_cache/src/lib.rs
+++ b/crates/build_cache/src/lib.rs
@@ -30,6 +30,7 @@ use command::Stdio;
 use command::r#async::Command;
 use futures_lite::future;
 use is_executable::IsExecutable as _;
+use itertools::Itertools;
 use sha2::{Digest, Sha256};
 use warp_errors::{ErrorExt, register_error};
 
@@ -553,6 +554,7 @@ where
         Ok(Some(plan)) => plan,
         Ok(None) => return report,
         Err(error) => {
+            tracing::error!(error = ?error, "Cache planning failed");
             report.invocations.push(failed_invocation(
                 CacheScope::Global,
                 Vec::new(),
@@ -582,6 +584,11 @@ where
                 Duration::ZERO,
             )
         } else {
+            log::info!(
+                "Mounting cache modes {:?} for {:?}",
+                configuration.modes,
+                configuration.scope
+            );
             run_spacectl_mount(
                 configuration.scope.clone(),
                 configuration.modes.clone(),
@@ -595,6 +602,8 @@ where
         };
 
         if let Some(response) = &invocation.response {
+            tracing::info!(cache_result = ?response.output, modes = ?response.input.modes, scope = ?configuration.scope, "Mounted cache paths");
+
             match &configuration.scope {
                 CacheScope::Repository { .. } => {
                     for (name, value) in &response.output.add_envs {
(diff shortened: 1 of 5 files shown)


crates/build_cache/src/lib.rs
```diff
@@ -660,13 +669,15 @@ where
 /// - `~/.cargo/registry` => `/cache/shared/$HOME/.cargo/registry`
 /// - `/workspace/repo-a/target` => `/cache/<repo-a-key>/target`
 /// - `/workspace/repo-b/target` => `/cache/<repo-b-key>/target`
+#[tracing::instrument(skip_all, fields(tags.cloud_agent = true, additional_global_modes, resolved_modes = tracing::field::Empty))]
```
joeywangzr: Is there a reason why there isn't a parameter passed into additional_global_modes here? I think skip_all might make the field empty

crates/build_cache/src/lib.rs
```diff
@@ -553,6 +554,7 @@ where
         Ok(Some(plan)) => plan,
         Ok(None) => return report,
         Err(error) => {
+            tracing::error!(error = ?error, "Cache planning failed");
```
joeywangzr, line 558: What's the difference between tracing::error and logs::info, noticed both of them were used in this file.

bnavetta: The `log` crate writes messages to a user-accessible log file, which they can send us for debugging. In cloud agents, it also goes to the sandbox stderr stream. The `tracing` crate pushes OTel traces to GCP for us to observe.

We generally use `log` for informational/error messages. Our `tracing` setup is newer and only for cloud agents, so it’s just for performance monitoring and troubleshooting there.

Confusingly, the `tracing` crate ecosystem can also handle logging and ingest messages from the ‘log’ crate

crates/build_cache/src/lib.rs
bnavetta: Ah, I had thought it used the value of the matching argument in this case, but will confirm / make explicit

joeywangzr (approved):

PR #14269, opened 2026-07-24
feat: support TUI terminal use for alternate-screen commands

kevinyang372:
## Description
- Support terminal use for agent-controlled alternate-screen commands in the headless TUI.
- Render alternate-screen PTY output above the CLI subagent status, blockers, composer, and footer while the agent owns control; keep user-controlled alternate screens PTY-only.
- Route terminal input according to the current owner and drop stale user PTY writes after agent takeover.
- Show details for blocked terminal-use actions and provide clickable Allow/Reject controls.
- Use Ctrl+Enter to allow the displayed action and Ctrl+C to reject it, while preserving Enter for follow-up submission and Shift+Enter for newlines.

## Linked Issue
None.

## Testing
- Added render and behavior coverage for agent- and user-controlled alternate screens, PTY ownership transitions, blocked-action presentation, exact action approval/rejection, and the Ctrl+Enter binding.
- `./script/format`
- `cargo nextest run -p warp_tui` (630 passed)
- `cargo clippy -p warp_tui --all-targets --all-features --tests -- -D warnings`
- `git diff --check`
- `./script/run-tui` built, bundled resources, codesigned, and launched successfully.

### Screenshots / Videos
Not attached; the change was verified in a live local TUI session.

## Agent Mode
- [Agent Mode conversation](https://staging.warp.dev/conversation/e45d1209-475f-4112-b325-ca3c720623a1)

CHANGELOG-IMPROVEMENT: Added terminal-use support for alternate-screen commands in Warp's TUI.

### DIFF
--- a/crates/warp_tui/src/terminal_session_view.rs
+++ b/crates/warp_tui/src/terminal_session_view.rs
@@ -131,9 +131,13 @@ const CTRL_C_EXIT_HINT: &str = "ctrl-c again to exit";
 const STARTING_SHELL_HINT: &str = "Starting shell...";
 const SESSION_CAN_CANCEL_RESTORE_FLAG: &str = "TuiSessionCanCancelRestore";
 const SESSION_CAN_HAND_BACK_CONTROL_FLAG: &str = "TuiSessionCanHandBackControl";
+const SESSION_CAN_ACCEPT_BLOCKED_TERMINAL_USE_ACTION_FLAG: &str =
+    "TuiSessionCanAcceptBlockedTerminalUseAction";
 pub(crate) const SESSION_COMPOSER_OWNS_INPUT_FLAG: &str = "TuiSessionComposerOwnsInput";
 pub(crate) const PASTE_IMAGE_BINDING_NAME: &str = "tui:session:paste_image";
 pub(crate) const AUTO_APPROVE_TOGGLE_BINDING_NAME: &str = "tui:session:toggle_auto_approve";
+pub(crate) const ACCEPT_BLOCKED_TERMINAL_USE_ACTION_BINDING_NAME: &str =
+    "tui:session:accept_blocked_terminal_use_action";
 pub(crate) const VOICE_INPUT_BINDING_NAME: &str = "tui:session:start_voice_input";
 
 /// Events emitted by the TUI terminal session surface.
@@ -491,6 +495,10 @@ pub(crate) enum TuiTerminalSessionAction {
     CancelRestore,
     /// Return a user-controlled terminal-use command to the agent.
     HandBackTerminalUseControl,
+    /// Accept the active terminal-use agent's blocked action.
+    AcceptBlockedTerminalUseAction,
+    /// Reject the active terminal-use agent's blocked action.
+    RejectBlockedTerminalUseAction,
     /// Click on the footer's usage entry: flips the persisted credits⇄cost
     /// display-mode setting.
     ToggleUsageDisplay,
@@ -632,6 +640,17 @@ pub(crate) fn init(app: &mut AppContext) {
         .with_group(TUI_BINDING_GROUP),
     ]);
     app.register_editable_bindings([
+        EditableBinding::new(
+            ACCEPT_BLOCKED_TERMINAL_USE_ACTION_BINDING_NAME,
+            "Accept the blocked terminal-use action",
+            TuiTerminalSessionAction::AcceptBlockedTerminalUseAction,
+        )
+        .with_context_predicate(
+            (id!(TuiInputView::ui_name()) | view_context.clone())
+                & id!(SESSION_CAN_ACCEPT_BLOCKED_TERMINAL_USE_ACTION_FLAG),
+        )
+        .with_group(TUI_BINDING_GROUP)
+        .with_key_binding("ctrl-enter"),
         EditableBinding::new(
             AUTO_APPROVE_TOGGLE_BINDING_NAME,
             "Toggle auto approve",
@@ -885,7 +904,7 @@ impl TuiTerminalSessionView {
                     let controller = self.cli_subagent_controller.clone();
                     let action_model = self.ai_action_model.clone();
                     let terminal_model = self.terminal_model.clone();
-                    let view = ctx.add_typed_action_tui_view(|ctx| {
+                    let view = ctx.add_tui_view(|ctx| {
                         TuiCLISubagentView::new(
                             target,
                             controller,
@@ -984,6 +1003,36 @@ impl TuiTerminalSessionView {
             .as_ref(ctx)
             .active_target()
             .filter(|target| target.control_state.is_user_in_control())
+    }
+    fn active_cli_subagent_view(&self, ctx: &AppContext) -> Option<ViewHandle<TuiCLISubagentView>> {
+        let target = self.cli_subagent_controller.as_ref(ctx).active_target()?;
+        self.cli_subagent_views.get(&target.block_id).cloned()
+    }
+
+    fn accept_active_cli_subagent_action(&mut self, ctx: &mut ViewContext<Self>) -> bool {
+        let Some(view) = self.active_cli_subagent_view(ctx) else {
+            return false;
+        };
+        if !view.as_ref(ctx).has_blocked_action(ctx) {
+            return false;
+        }
+        view.update(ctx, |view, ctx| {
+            view.accept_blocked_terminal_use_action(ctx)
+        });
+        true
+    }
+
+    fn reject_active_cli_subagent_action(&mut self, ctx: &mut ViewContext<Self>) -> bool {
+        let Some(view) = self.active_cli_subagent_view(ctx) else {
+            return false;
+        };
+        if !view.as_ref(ctx).has_blocked_action(ctx) {
+            return false;
+        }
+        view.update(ctx, |view, ctx| {
+            view.reject_blocked_terminal_use_action(ctx)
+        });
+        true
     }
 
     fn send_terminal_use_prompt(&mut self, input: &str, ctx: &mut ViewContext<Self>) -> bool {
@@ -1476,6 +1525,10 @@ impl TuiTerminalSessionView {
                 view.update_process_input_focus(ctx);
                 ctx.notify();
             }
+            ModelEvent::TerminalModeSwapped(_) => {
+                view.update_process_input_focus(ctx);
+                ctx.notify();
+            }
             ModelEvent::Typeahead => view.handle_typeahead_event(ctx),
             ModelEvent::BlockMetadataReceived(_)
             | ModelEvent::BlockWorkingDirectoryUpdated(_)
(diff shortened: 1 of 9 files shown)


kevinyang372:
* **#14269** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14269?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14269?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -491,6 +495,8 @@ pub(crate) enum TuiTerminalSessionAction {
     CancelRestore,
     /// Return a user-controlled terminal-use command to the agent.
     HandBackTerminalUseControl,
+    /// Allow the action currently blocking the active terminal-use agent.
```
kevinyang372: Instead of having a generic AllowedBlockedTerminalAction, I think we should have an explicit `AcceptBlockedTerminalUseAction` -- this action should be specific to terminal use

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -632,6 +638,17 @@ pub(crate) fn init(app: &mut AppContext) {
         .with_group(TUI_BINDING_GROUP),
     ]);
     app.register_editable_bindings([
+        EditableBinding::new(
+            ALLOW_BLOCKED_TERMINAL_ACTION_BINDING_NAME,
+            "Allow the blocked terminal action",
```
kevinyang372: ditto: update the description here

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -773,6 +790,13 @@ impl TuiTerminalSessionView {
             TuiBlockingChild::Orchestration(view) => ctx.focus(&view),
         }
     }
+    fn render_blocking_child(blocker: TuiBlockingChild) -> Box<dyn TuiElement> {
```
kevinyang372: What's the reasoning for having these?

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -985,6 +1009,24 @@ impl TuiTerminalSessionView {
             .active_target()
             .filter(|target| target.control_state.is_user_in_control())
     }
+    fn active_cli_subagent_view(&self, ctx: &AppContext) -> Option<ViewHandle<TuiCLISubagentView>> {
+        let target = self.cli_subagent_controller.as_ref(ctx).active_target()?;
+        self.cli_subagent_views.get(&target.block_id).cloned()
+    }
+
+    fn allow_blocked_terminal_use_action(&mut self, ctx: &mut ViewContext<Self>) -> bool {
+        let Some(view) = self.active_cli_subagent_view(ctx) else {
```
kevinyang372, line 1013: I wonder if that action should be defined within active_cli_subagent_view and handled there?

crates/warp_tui/src/tui_cli_subagent_view.rs
```diff
@@ -73,6 +70,123 @@ fn terminal_use_status_text(
     format!("{status} · {key_binding} {action}")
 }
 
+#[derive(Debug, PartialEq, Eq)]
+struct BlockedActionPresentation {
+    summary: String,
+    detail: Option<String>,
+}
+
+fn display_pty_input(input: &[u8]) -> String {
```
kevinyang372, line 78: is this the same display logic as GUI? Can this be shared?

crates/warp_tui/src/tui_cli_subagent_view.rs
```diff
@@ -73,6 +70,123 @@ fn terminal_use_status_text(
...
+            '\u{1b}' => "<Esc>".to_owned(),
+            character if character.is_control() => {
+                format!("<0x{:02X}>", u32::from(character))
+            }
+            character => character.to_string(),
+        })
+        .collect()
+}
+
+fn blocked_action_presentation(action: &AIAgentActionType) -> BlockedActionPresentation {
```
kevinyang372, line 93: ditto: Is this shared with GUI by any chance?

crates/warp_tui/src/terminal_session_view.rs
kevinyang372: Updated. The explicit terminal-use action is now `TuiCLISubagentViewAction::AcceptBlockedTerminalUseAction`; the session carries it only through its `CLISubagent(...)` forwarding variant.

crates/warp_tui/src/terminal_session_view.rs
kevinyang372: Updated the binding and context names plus the user-facing description: `ACCEPT_BLOCKED_TERMINAL_USE_ACTION_BINDING_NAME`, `SESSION_CAN_ACCEPT_BLOCKED_TERMINAL_USE_ACTION_FLAG`, and “Accept the blocked terminal-use action.”

crates/warp_tui/src/terminal_session_view.rs
kevinyang372: This conversion is used by both the normal transcript layout and the agent-controlled alternate-screen layout. Keeping one exhaustive match avoids duplicating the blocker-to-child-view mapping and makes new blocker variants fail compilation until handled in both layouts. I added that rationale as a doc comment.

crates/warp_tui/src/terminal_session_view.rs
kevinyang372: Updated. Accept/reject are now defined and handled by `TuiCLISubagentView`. The session only locates the active child and forwards its `TuiCLISubagentViewAction`; that forwarding step remains necessary because the composer retains focus while blocked, so the CLI-subagent sibling is not in the focused action chain.

crates/warp_tui/src/tui_cli_subagent_view.rs
kevinyang372: I compared the GUI path. It is not the same presentation layer: GUI builds styled WarpUI elements with scroll/selection state, speedbumps, action menus, and per-action rich layouts; TUI needs plain multiline strings for the cell grid. Sharing that renderer would couple the two frontend frameworks, so this PR keeps rendering frontend-specific while sharing the underlying `AIAgentActionType` semantics.

crates/warp_tui/src/tui_cli_subagent_view.rs
kevinyang372: The byte display also differs intentionally. GUI decorates input by write mode, converts controls to normalized key names, and returns highlight ranges for rich text; TUI preserves newline/tab structure and emits safe plain tokens for a cell-grid `TuiText`. Reusing the GUI parser would change the TUI output semantics and still leave the element rendering separate, so I kept this frontend-specific.

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -2307,6 +2419,11 @@ impl TuiTerminalSessionView {
             ctx.terminate_app(TerminationMode::ForceTerminate, None);
             return;
         }
+        if self.reject_blocked_terminal_use_action(ctx) {
```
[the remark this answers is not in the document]
kevinyang372: Rechecked the current ordering: blocked terminal-use rejection runs before `Instant::now()` and `exit_confirmation.should_exit(now)`, disarms the exit confirmation, and returns. Therefore an armed exit cannot win over rejection; no code change was needed for this thread.

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -3741,6 +3848,20 @@ impl TuiView for TuiTerminalSessionView {
 }
 
 impl TuiTerminalSessionView {
+    fn forward_user_pty_bytes(&self, bytes: &[u8], ctx: &mut ViewContext<Self>) {
+        let agent_in_control = self
+            .terminal_model
+            .lock()
+            .block_list()
+            .active_block()
+            .is_agent_in_control();
```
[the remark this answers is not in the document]
kevinyang372: Updated the final PTY-write guard to use `is_agent_in_control_or_tagged_in()` and expanded the regression test to prove stale writes are dropped in both tagged-in and agent-controlled states.

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -632,6 +640,19 @@ pub(crate) fn init(app: &mut AppContext) {
         .with_group(TUI_BINDING_GROUP),
     ]);
     app.register_editable_bindings([
+        EditableBinding::new(
+            ACCEPT_BLOCKED_TERMINAL_USE_ACTION_BINDING_NAME,
+            "Accept the blocked terminal-use action",
+            TuiTerminalSessionAction::CLISubagent(
```
kevinyang372: We should avoid these nested action pattern -- if this cannot be received from the subagent view, it's fine to keep it as the terminal session view action

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -773,6 +794,16 @@ impl TuiTerminalSessionView {
             TuiBlockingChild::Orchestration(view) => ctx.focus(&view),
         }
     }
+    /// Converts the shared blocking-child enum into an element for both the
+    /// transcript and alternate-screen layouts without duplicating this
+    /// exhaustive mapping between the two render branches.
+    fn render_blocking_child(blocker: TuiBlockingChild) -> Box<dyn TuiElement> {
+        match blocker {
+            TuiBlockingChild::AskQuestion(view) => TuiChildView::new(&view).finish(),
```
kevinyang372: Should this be a TuiBlockingChild::view_element method?

crates/warp_tui/src/terminal_session_view.rs
kevinyang372: Updated. Removed the nested `CLISubagent(...)` action pattern. Accept and reject are now flat `TuiTerminalSessionAction` variants; the CLI-subagent buttons dispatch those session actions, and the session routes Ctrl+Enter/Ctrl+C to the active child view.

crates/warp_tui/src/terminal_session_view.rs
kevinyang372: Updated. `TuiBlockingChild` now owns the exhaustive conversion through `view_element()`, and the alternate-screen render path calls that method directly.

crates/warp_tui/src/terminal_session_view.rs
```diff
@@ -1804,6 +1857,72 @@ impl TuiTerminalSessionView {
     fn render_orchestration_tab_footer(&self, builder: &TuiUiBuilder) -> Box<dyn TuiElement> {
         render_orchestration_tab_footer(builder)
     }
+
+    fn append_input_area(
```
harryalbert: nit: could this just return the input area element instead of like appending it to a specific flex? `render_inpu_area` feels like a clearer fn to me than append

harryalbert (approved):

crates/warp_tui/src/terminal_session_view.rs
kevinyang372: Updated. Replaced `append_input_area(content, ...) -> TuiFlex` with `render_input_area(...) -> Box<dyn TuiElement>`. Both the normal and alternate-screen layouts now attach the returned input-area element explicitly.

PR #14306, opened 2026-07-25
feat(workspace): Projects × Tasks layout — project rail + project-scoped tabs

### LINKED ISSUE #14305: Projects × Tasks layout: group sessions by project, with that project's tasks as top tabs
## Summary

Running several coding-agent tasks per repository is now a common workflow (git worktrees + parallel agents), but Warp's session list is a single flat/grouped list of tabs with no notion of which repository a session belongs to. With a handful of repos open at once, the tab strip becomes a mixed pile: there is no way to see "these three tasks belong to *this* project", and worktrees of the same repo look like unrelated sessions.

Tools built specifically for this workflow (e.g. Herdr) converge on a two-pane shape — **projects on the left, that project's tasks along the top** — and it maps cleanly onto what Warp already has.

## Proposal: a Projects × Tasks layout

- **Left rail = projects.** One entry per project, derived from the repository each open session is in. Selecting a project scopes the window to it.
- **Top tab bar = that project's tasks.** The existing horizontal tabs, filtered to the selected project. Tabs already show the running agent's task title and status, so each row reads as a task.
- **Worktrees unify.** All worktrees/branches of one repository appear as tasks under a single project, not as separate projects — this is the point of the feature for parallel-agent work.

## Why this needs a real model, not just a filtered render

Warp's tab machinery assumes `Workspace::tabs` is the whole world: `active_tab_index` is a raw index, next/prev walks the full vector, numeric activation indexes it directly, close-left/right use raw contiguous ranges, and drag hit-testing returns raw insertion indices. Filtering only the render would desynchronise what is displayed from what is navigated (and, for the close operations, could close tabs the user cannot see).

So the feature needs:
1. a **stable project identity** that unifies worktrees (the repo's shared git dir — `Repository::common_git_dir()` already computes this, nothing currently uses it as an identity);
2. a single **projection** (project list + tab→project map + visible raw indices) that every navigation and render path consults; and
3. the invariant that **the active tab always belongs to the selected project** — enforced at the one choke point that changes active-tab identity.

## Scope

In scope: project identity, the projection, the rail, project-scoped top tabs and navigation, and `+` creating a task inside the selected project. GUI only, behind `FeatureFlag::Projects` plus a user setting.

Out of scope for now: a distinct agent-run identity model (project → task → run with per-run evidence/attention), automatic worktree creation per task, and headless-TUI parity.

A PR implementing the above is ready.

samithaj:
## Description

Adds a **Projects × Tasks** layout for the session UI: a project rail on the left, and the selected project's tasks as tabs along the top. It targets the parallel-coding-agent workflow — several tasks per repository, often in git worktrees — where today's flat tab strip mixes every repo together.

Behind `FeatureFlag::Projects` (dogfood) **and** a new user setting, so the default experience is unchanged.

### Project identity unifies worktrees
`ProjectKey` keys a local git project on the repository's **shared git dir** — `Repository::common_git_dir()`, which `repo_metadata` already computes but nothing used as an identity. Every worktree of a repo therefore collapses into one project by construction (non-git dirs key on the canonical path; remotes on the remote identity; sessions with no detectable repo fall under an explicit **"Other"** entry). No new git plumbing.

### A projection, not a filtered render
Warp's tab machinery assumes `Workspace::tabs` is the whole world — `active_tab_index` is a raw index, next/prev walks the full vector, numeric activation indexes it directly, close-left/right use raw contiguous ranges, and drag hit-testing returns raw insertion indices. Filtering only the render would desynchronise what is displayed from what is navigated, and could close tabs the user cannot see.

So this adds one `ProjectLayout` projection (project list + tab→project map + `visible_tab_indices`) that **every** affected path consults via a single `project_visible_indices()` helper:

- top tab bar slot model (and drag insertion geometry, so on-screen and hit-test agree)
- next/prev, numeric `Cmd-N`, Ctrl-Tab / MRU palette
- close-others / close-left / close-right (scoped so they can never close another project's tabs)
- shift-click range selection

Manual tab groups are suppressed in project mode — the project *is* the grouping dimension, which also avoids the contiguous-run hazard in the group slot model.

### The invariant
`set_active_tab_index` is the single choke point that changes active-tab identity, so it enforces **the active tab always belongs to the selected project**; selecting a project in the rail activates that project's most-recently-used tab. Because selection is derived from the active tab, it restores for free from the already-persisted active tab — **no new DB column or migration**.

### Also
- `+` creates a task **inside** the selected project (seeded from that project's MRU tab's cwd, so you stay in the worktree you were in). A restored conversation's own directory still wins.
- One `vertical_tabs_layout_active()` / `project_layout_active()` pair is now the single source of truth for the layout decision, replacing 27 duplicated checks across 3 files.

## Linked Issue

Fixes #14305

## Testing

- Unit tests: `ProjectKey` worktree unification + labels, `visible_tab_indices` selection, first-seen project ordering, the "Other" bucket, and next/prev cycling within a subset (8 tests).
- Manually verified with `./script/run`: rail lists the open projects, selecting one filters the top tabs to it, hover/selected states render, and the layout switches to the horizontal top bar with the setting on.
- `./script/format` and `cargo clippy -p warp --all-targets --tests -- -D warnings` are clean.

## Agent Mode

## Known gaps / follow-ups

- Pinned tabs remain global in project mode (`pinned_boundary_index` assumes pinned tabs are a contiguous prefix of the raw vector); per-project pinning needs reordering semantics worth deciding separately.
- No `+` button or empty state in the rail yet.
- Headless-TUI parity is not included.
- A distinct agent-run identity model (project → task → run, with per-run evidence/attention and worktree-per-task creation) is deliberately out of scope here.

CHANGELOG-NEW-FEATURE: Group your sessions by project — a project rail on the left, with that project's tasks as tabs along the top, and all worktrees of a repo under one project.


samithaj:
## Progress update — running this daily now

The layout is working end to end. Two additions since the initial push, both driven by using it:

### Tabs can now show the agent session *and* the instruction behind it

With tasks living on the horizontal bar, one line could only answer half the question — a renamed Claude Code session tells you *what* a task is, but not *what you asked it to do*. Tabs can now show both, mirroring how the vertical sidebar already models this (view mode × primary info × subtitle):

```toml
[appearance.tabs]
line_count     = "two_line"          # single_line (default) | two_line
primary_info   = "agent_session"     # agent_session | user_instruction | command | working_directory | branch
secondary_info = "user_instruction"  # same set; falls back if it collides with the primary
```

Notably `user_instruction` had to become its own line option: "use the latest user prompt" was a single preference that flipped *both* lines together, so choosing it for the subtitle also renamed the title — making "session name over instruction" impossible to express.

Two readability fixes fell out of the first build:

- Titles were clipped at the **start**, which suits a path (`…/dev/mifos/app`) but hides the identifying words of prose — an instruction rendered as *"users and store passwords in env"*. Only working-directory lines clip at the start now.
- Two-line tabs cap at 320px rather than 200px, since two lines of prose need more room than a shell title.

Tab height is content-driven, so the pill grows on its own; the bar had to follow. `PANEL_HEADER_HEIGHT` is no longer an alias of `TAB_BAR_HEIGHT` (otherwise the resource center and theme chooser headers silently grow too), and the macOS titlebar is sized from `total_tab_bar_height()`, which is what positions the traffic lights.

### Both toggles are in Settings

Previously reachable only from the Command Palette or `settings.toml` — which is exactly how they ended up documented wrongly and set in the wrong channel's file. `primary_info`/`secondary_info` are still TOML-only; they need real dropdowns.

### Also since the first push

- Rail always visible when the layout is on, rather than appearing once a second project opens
- Per-project agent status in the rail — a project blocked on you wins over one merely working, so "which project needs me?" is answerable at a glance
- `+` opens a task inside the selected project, seeded from that project's most-recent tab so you stay in the worktree you were in
- Agent title resolution now has one owner (`workspace::tab_title`) instead of being duplicated between the two surfaces

Single-line remains the default and renders exactly as before. `./script/format`, `cargo clippy -p warp --all-targets --tests -- -D warnings`, and the unit tests are clean.

PR #14312, opened 2026-07-26
Fix Bash executable cache dropping commands after a PATH name with spaces

### LINKED ISSUE #13475: All valid commands (sudo, git, cat, etc.) incorrectly underlined as invalid on Ubuntu 26.04 LTS
### Pre-submit Checks

### Describe the bug

On Ubuntu 26.04 LTS, Warp's error underlining feature marks basic, valid, existing commands (e.g. sudo, git, cat) with a red dashed underline as if they were invalid/not found. This happens even though the binaries exist, are correctly located in $PATH, and execute successfully when run.

This is not limited to custom tools, aliases, or user-installed software — it affects core system binaries like cat and sudo, located in /usr/bin.

### To reproduce

Steps to reproduce:

1. Fresh Ubuntu 26.04 LTS install (or upgrade from 25.10)
2. Install Warp via apt (official repo)
3. Open Warp, start a new bash session
4. Type any basic command, e.g. cat, sudo, git
5. Observe the red dashed underline appearing under the command, indicating "invalid command," despite the command being valid and executable

### Expected behavior

Valid, existing commands should not be underlined as errors.

Diagnostics already performed

- type sudo / type git / type cat → all correctly resolve to their binaries (e.g. /usr/bin/sudo)
- echo $PATH → includes all standard system paths (/usr/bin, /usr/sbin, /bin, etc.) correctly
- bash --norc --noprofile -c "type sudo" → resolves correctly (ruling out .bashrc interference)
- bash -i -c "echo $PATH" → identical PATH to the one used inside Warp's active session
- ps -p $$ -o comm= → confirms shell is bash, matching Warp's configured shell in Settings > Features > Session
- Performed a full clean reinstall of Warp (apt purge, removed ~/.warp, ~/.config/warp-terminal, ~/.local/share/warp-terminal, ~/.cache/warp-terminal, re-added signing key and repo, reinstalled) — issue persisted identically after reinstall
- Corrected a .bashrc NVM initialization typo (broken markdown-style link syntax) — did not affect the issue
- Isolated pyenv virtualenv-init behind a $TERM_PROGRAM != "WarpTerminal" check per Warp's own "Known Issues" guidance on PS1-modifying tools — did not affect the issue
- Confirmed error underlining is enabled by default (Settings > Features > Editor > "Error underlining for commands")

Image

Image

Screenshot attached shows sudo and cat both underlined in red despite being valid, existing, executable commands confirmed via `type` and `which`.

### Operating system (OS)

Linux

Ubuntu 26.04 LTS

### Shell Version

bash 5.3.9

### Current Warp version

v0.2026.07.01.09.21.stable_01

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

maxmilian:
Closes #13475

**Update (2026-08-25) — the diagnostic came back, and it rules #13475 out as a match for this fix.** @ema14bdt ran both lines on 07-31: no whitespace-containing names on `PATH`, and `6270 6270` — the arrays are aligned on their machine. So the desync fixed here is **not** what they are hitting, and #13475 needs a second cause chased separately. See "What is still open" at the bottom.

This PR still stands on its own: the unquoted `${COMMANDS[@]}` in `shell/mod.rs:621` is a real correctness bug, reproduced headlessly on the Bash version from the report, with a regression test that fails before and passes after. It is still unfixed on current `master` (`6696954c6`).

I have kept `Closes #13475` above only because the readiness gate needs a registered linked issue — switching it back to `Refs` de-registers the link, which is why it was changed on 07-30. **Please reopen #13475 after merging**, or tell me which linkage you would rather see and I will edit this description to match.

## Problem

`ShellType::Bash.shell_command_to_get_executables()` pairs two arrays by index:

```bash
COMMANDS=($(compgen -c)); TYPES=($(type -t ${COMMANDS[@]}));
for i in "${!COMMANDS[@]}"; do if [[ ${TYPES[$i]} == "file" ]]; then echo ${COMMANDS[$i]}; fi; done
```

`${COMMANDS[@]}` is unquoted. A command name on `PATH` containing whitespace is split into several words; `type -t` prints **nothing** for the resulting non-existent names, so `TYPES` comes back shorter than `COMMANDS`. From that index on, every command is classified against another command's type, and valid external binaries fall out of the session's executable cache — which is what surfaces as the red "invalid command" underline in #13475.

The existing comment is right that batching `type` matters for startup cost, so this keeps the single batched call.

## Fix

Split `compgen` output on newlines only and disable globbing while the arrays are built:

```bash
__warp_oifs=$IFS; set -f; IFS=$'\n'
COMMANDS=($(compgen -c)); TYPES=($(type -t "${COMMANDS[@]}"))
IFS=$__warp_oifs; set +f; unset __warp_oifs
for i in "${!COMMANDS[@]}"; do if [[ ${TYPES[$i]} == "file" ]]; then printf '%s\n' "${COMMANDS[$i]}"; fi; done
```

`set -f` covers the same class of breakage for names containing glob metacharacters, and `printf '%s\n'` replaces `echo` so names are emitted verbatim.

## Real behavior proof

**Startup cost — the reason the original batches `type`.** `ubuntu:26.04`, `PATH` padded to 7063 commands, 3 runs each:

| variant | per run |
|---|---|
| current | 67 ms |
| this PR | 68 ms |
| per-name `$(type -t "$c")` loop | 1015 ms |

The per-name loop forks a subshell per command, so it is 15× slower — I floated it on the issue before measuring, and the measurement ruled it out. This PR keeps the single batched call, so startup is unchanged.

**Behavior, before vs after.** `ubuntu:26.04` (`GNU bash, version 5.3.9(1)-release`, same as the report), with three space-containing names dropped on `PATH`:

```
COMMANDS=1240 TYPES=1231 shift=9
dropped by the current snippet:
  a b c d, dpkg-deb, du, infocmp, my tool, pgrep, pkill, sudo_logsrvd, sum, two word cmd
```

`dpkg-deb`, `du`, `infocmp`, `pgrep`, `pkill`, `sudo_logsrvd` and `sum` are all real executables. With this change the arrays stay aligned (1235/1235) and every one of them is retained, including the space- and glob-containing names themselves.

**Shell compatibility.** The snippet runs in the user's bash, so I checked the older versions too — `bash:3.2` (3.2.57, what macOS ships), `bash:4.4`, `bash:5.2`, `ubuntu:26.04` (5.3.9). Arrays aligned and all commands retained on all four. `IFS`/`set -f` are POSIX-era, nothing here needs bash 4+ (which is why I did not reach for `mapfile`).

**Two negative results**, so nobody has to re-run them:

- A clean container does not reproduce — 1231 in, 1132 out, arrays aligned. Bash 5.3 alone is not the trigger.
- Neither is scale: padding `PATH` with 4000 extra executables kept 5063/5063 aligned. This is not an `ARG_MAX` or argument-count limit.

**Not tested:** Windows and remote/SSH sessions, and the actual red-underline rendering in the app — I verified the discovery command's output, not the completer path downstream of it. I do not have an Ubuntu desktop; everything above is the real snippet run in the real shell versions, headless.

## Screenshots

Recorded on a local OSS dev build of this branch (`WarpOss.app`), in a bash 3.2.57 session with two space-containing executables placed on `PATH`. In that environment the desync swallows **`oz`** — Warp's own CLI at `/Applications/Warp.app/Contents/Resources/bin/oz`, an unambiguously valid executable.

**Before — `oz` marked invalid (this branch with the fix reverted, i.e. current `master` behavior)**

before

**After — `oz` accepted (this branch)**

after

**Control — error underlining is still working**

control

`zzzznotarealcommand` on the same build as "Before", so the difference above is the executable cache recovering, not error underlining being switched off.

A note on which command to watch: the misalignment consumes entries from the tail of the list, so *which* commands vanish depends on the shell's `compgen -c` ordering. `cat` from the issue title drops in some environments but not in this one; `oz` is what drops here. That is the same variability described in "What is still open" below.

## Test

`test_bash_executable_discovery_survives_names_with_spaces` in `crates/warp_terminal/src/shell/mod_tests.rs` runs the real string from `shell_command_to_get_executables()` through `bash` against a temp `PATH` holding `aa-plain`, `my tool` and `zz-after-space`.

Verified red before / green after, with the final test code, on macOS bash 3.2.57:

```
# with the previous snippet
`zz-after-space` missing from the executable cache; got ["aa-plain", "my"]
test result: FAILED

# with this PR
test result: ok. 1 passed
```

The `got ["aa-plain", "my"]` output is the mechanism itself: `my tool` became the bogus entry `my`, and everything after it fell off. `aa-plain` sorts before the split and survives either way — it is the control that proves the harness isn't trivially passing.

`cargo test -p warp_terminal --lib shell::` → 17 passed. `cargo clippy -p warp_terminal --lib --tests` → clean. `cargo fmt --check` → clean.

## What is still open

**Answered.** @ema14bdt ran the diagnostic on 07-31:

```
compgen -c | grep -n '[[:space:]]'   # no matches
6270 6270                            # arrays aligned
```

No whitespace-containing command names on their `PATH`, and the two counts are equal — so the array desync this PR
fixes is not present on their machine, and it is not the trigger for #13475. Something else is dropping valid
commands from the executable cache there, and it needs to be chased as a separate cause.

That does not weaken this change: the unquoted `${COMMANDS[@]}` misclassifies commands whenever such a name *is*
present, which is reproduced above on the same Bash version as the report, and is still unfixed on `master`. It is
just a fix for a different failure than the one #13475 describes.

Hence the request in the header: **please reopen #13475 after merging this**, so the reporter's actual problem does
not get closed out by a fix that demonstrably does not address it.

Happy to split the `set -f` part out if you'd prefer the PR scoped strictly to the whitespace case.


maxmilian:
Taking this out of draft.

I held it in draft waiting on @ema14bdt to run two diagnostic lines on their machine, to check whether their symptom comes from *this* array desync or from a different one. That's been outstanding since 07-26 with no reply, and I don't think it should keep the fix out of review, because the two questions are separable:

- **Is the bug real?** Yes, independently of the report. `` is unquoted in `shell_command_to_get_executables()`, so any `PATH` entry whose name contains whitespace shortens `TYPES` relative to `COMMANDS` and every subsequent command gets classified against another command's type. I reproduced it headlessly in a container on the exact Bash version from the report, and the regression test in this PR fails before the change and passes after.
- **Does it explain #13475 specifically?** Unconfirmed, and only @ema14bdt can settle that. If their counts turn out to match, something else is desyncing the arrays on their setup and #13475 should stay open after this merges.

That's why this references #13475 rather than claiming to close it. CI is green; the change keeps the single batched `type` call that the existing comment justifies on startup cost, and adds `set -f` so glob metacharacters in command names can't cause the same class of breakage.

@ema14bdt — the diagnostic ask still stands if you get a chance; it decides whether #13475 closes with this or needs a second look.

maxmilian:
Linked the issue properly — the body said `Refs #13475`, which isn't a closing keyword, so no linked issue was registered and the gate had nothing to check. It now reads `Closes #13475`, and #13475 already carries `ready-to-implement`.

I kept the caveat explicit in the body rather than dropping it: I have reproduced the mechanism headlessly on the Bash version from the report, but the reporter has not run the two diagnostic lines that would confirm it is *their* trigger. If it turns out not to be, #13475 should be reopened for a second cause — the array-desync fix here stands on its own either way.

/oz-review

maxmilian:
Added the visual evidence to the PR description.

I recorded it on a local OSS dev build of this branch, in a bash 3.2.57 session with two space-containing executables on `PATH`. Worth flagging one thing I found while capturing it: **`cat` from the issue title does not drop in every environment.** The misalignment consumes entries from the tail of the list, so which commands vanish depends on the shell's `compgen -c` ordering — in my capture environment `cat` survived and `oz` (Warp's own CLI, `/Applications/Warp.app/Contents/Resources/bin/oz`) is what gets swallowed. So the Before/After pair uses `oz`.

I also included a control shot: `zzzznotarealcommand` still gets the red underline on the same build as "Before". Without it, "the underline disappeared" would be equally consistent with the fix having simply broken error underlining, which is not what happened — the executable cache is recovering.

This closes the gap I flagged myself in "What is still open": I had verified the discovery command's output but not the rendered underline. That path is now verified end-to-end. What remains unconfirmed is still only whether this is the *same* desync @ema14bdt hit.

/oz-review

ema14bdt:
Ran both:

```
compgen -c | grep -n '[[:space:]]'
```
→ no output (no whitespace-containing names on PATH)

```
COMMANDS=($(compgen -c)); TYPES=($(type -t ${COMMANDS[@]})); echo "${#COMMANDS[@]} ${#TYPES[@]}"
```
→ `6270 6270`

Counts match exactly, so the arrays are aligned on my machine — this isn't the same desync mechanism your PR fixes. Sorry to be the "counts match" case you were hoping to avoid.

For what it's worth, here's what I've confirmed independently so far, in case it helps narrow down the second cause:
- Builtins (`cd`, `history`) are never underlined — only external binaries.
- `compgen -c` lists `sudo`, `cat`, `git` correctly (with harmless duplicate PATH-dir entries).
- `type -t sudo` / `type -t cat` / `type -t git` all correctly return `file`.
- With `RUST_LOG=debug`, the log shows "Lazily loaded command signature for X" for commands that resolve fine (`ls`, `su`), but that line never appears for `sudo`, `cat`, or `git` — the log jumps straight to `parsing internal command` without a signature load step for those three.

So on my system the batched `type -t` call itself seems to produce correct output (verified via the count check above), but something downstream of that — during Warp's own parsing/consumption of the batch result, or in `describe_given_token`/`top_level_commands()` — appears to drop these specific names.

maxmilian:
Thanks for running both — and no need to apologise, a clean negative is worth more than a maybe. `6270 6270` settles it: the array desync this PR fixes is not what you're hitting, so this PR should not close #13475. I've said as much in the PR description, and I'd rather the issue stay open under a second cause than get quietly swept up by a merge here.

One caveat on the debug-log clue, before it sends either of us the wrong way:

## The missing "Lazily loaded command signature" line probably doesn't mean what it looks like

That line lives inside the signature **lookup closure**, and the closure only runs on a cache miss:

- `crates/warp_completer/src/signatures/legacy/mod.rs:34-42` — the closure that logs it.
- `crates/warp_completer/src/signatures/legacy/registry.rs:53-55` — it is invoked through `MemoMap::get_or_insert`, and the memo stores `Option<Signature>`, i.e. it caches negative results too ("our knowledge of commands for which we do _not_ have a signature", `registry.rs:30-35`).

So a command logs that line **at most once per run of Warp**, whether the lookup found a signature or not. Any earlier lookup of `sudo`/`cat`/`git` in that process — by anything at all, not necessarily your typing — permanently silences the line for them. So "no lazy-load line" is equally consistent with "already loaded and cached" as with "never loaded", and can't distinguish the two. (I can't check which commands ship bundled signatures from here: the `warp_command_signatures` crate that backs `signature_by_name` isn't part of the OSS tree, only its consumer is.) It's a red herring, and I'd have chased it too.

More to the point: the red underline isn't driven by signatures at all. It's driven by whether the first token got a *description*, and a top-level command gets its description from an exact match against the completer's suggestions, which for command names come from the session's external-executable cache (`describe_given_token`, `crates/warp_completer/src/completer/describe.rs:120-199`). Builtins and aliases come from separate sources — which is exactly why builtins stay clean while external binaries don't, whatever is going wrong.

## The one measurement that would split the remaining hypotheses

The counts you ran are from your interactive shell. Warp runs its discovery in the **session** shell, with `PATH` explicitly overridden from what it captured at bootstrap (`app/src/terminal/model/session.rs:1258-1271`), so its view can differ from yours — I got caught by exactly this distinction on this PR last week, where a command I was certain would be affected turned out not to be, purely because the app's session had a different `PATH` and `compgen -c` ordering than my own shell.

So: in a Warp bash session, could you run the real discovery string and check whether those three come out of it?

```bash
COMMANDS=($(compgen -c)); TYPES=($(type -t ${COMMANDS[@]}))
echo "counts: ${#COMMANDS[@]} ${#TYPES[@]}"
for i in "${!COMMANDS[@]}"; do if [[ ${TYPES[$i]} == "file" ]]; then echo ${COMMANDS[$i]}; fi; done > /tmp/warp-execs.txt
wc -l < /tmp/warp-execs.txt
for c in sudo cat git su ls; do printf '%s: %s\n' "$c" "$(grep -cx "$c" /tmp/warp-execs.txt)"; done
```

That is the literal string Warp sends (`crates/warp_terminal/src/shell/mod.rs:615`), so its output is the cache contents.

- If `sudo`/`cat`/`git` come back **0** — discovery is dropping them in your environment, and the counts line in that same run will say whether it's a desync after all or something else entirely.
- If they come back **1** and they're still underlined in the UI — discovery is fine and the break is downstream of it, either in loading that output into the session or in the description lookup. Different bug, and knowing that narrows it a lot.

A second, no-typing-required signal while you're there: type `sud` and see whether `sudo` is offered in the completion menu as a command. If it is offered but still underlines once fully typed, that's the same "downstream" answer from the UI side.

Either way this PR stands on its own — the unquoted `${COMMANDS[@]}` is a real defect that drops valid binaries whenever a `PATH` entry contains whitespace, with a fail-before/pass-after regression test — it just isn't your defect. Thanks for staying with this; the negative result you produced is what makes the second cause findable at all.

maxmilian:
Addressed the remaining nit in `2ae2f85` — comment only, no behaviour or assertion change.

Oz was right that the comment leaned on something the test does not: it claimed `zz-*` sorts after the space-containing name and `aa-plain` before it, but `compgen -c` ordering is not guaranteed across Bash versions. The assertion itself never depended on it — it requires *both* stubs to survive — so the comment now explains that one stub sits on each side of `my tool` by name, and that a misalignment swallows whatever this particular Bash happens to list after it, which the two-sided assertion catches either way.

`cargo fmt --check` clean; `test_bash_executable_discovery_survives_names_with_spaces` passes.

Note that this PR should still **not** close #13475 — that remains the case from the Jul 31 exchange with the reporter: their `6270 6270` reading rules out the array desync this fixes, so the issue needs to stay open under a second cause. The PR description says so too.

/oz-review

Issue #14321, opened 2026-07-27
cargo test -p warpui --lib aborts with STATUS_HEAP_CORRUPTION on Windows when run in parallel


SS-Script:
## What happened

`cargo test -p warpui --lib` aborts on Windows with `STATUS_HEAP_CORRUPTION` before the suite
finishes. It is not a specific test failing — the process dies:

```
test fonts::text_layout_tests::test_layout_text_first_line_indent_medium ... ok
test fonts::text_layout_tests::test_layout_text_first_line_indent_large ... ok
test windowing::winit::fonts::layout_tests::test_fixed_width_tab_size_affects_tab_width ... ok
test fonts::text_layout_tests::test_layout_text_first_line_indent_small ... ok
test windowing::winit::fonts::str_index_map::tests::test_str_index_map_get_byte_index ... ok
test windowing::winit::fonts::str_index_map::tests::test_str_index_map_get_char_index ... ok
test windowing::winit::fonts::layout_tests::test_layout_text_first_line_indent_large ... ok
error: test failed, to rerun pass `-p warpui --lib`

Caused by:
  process didn't exit successfully: `...\target\debug\deps\warpui-7c9fe6c72543d788.exe`
  (exit code: 0xc0000374, STATUS_HEAP_CORRUPTION)
```

Which tests have reported `ok` before the abort varies between runs.

## Expected

The suite runs to completion.

## Workaround

Running single-threaded is reliable:

```
cargo test -p warpui --lib -- --test-threads=1
```

→ `test result: ok. 41 passed; 0 failed; 10 ignored`.

That points at concurrent access to something process-global in the font/text-layout stack rather
than at any individual test, though I haven't isolated which pair of tests collide.

## Environment

- Windows 11 Pro, build 26120, x86_64
- Clean `master` @ `c16fd426d`, no local patches
- Pinned toolchain (Rust 1.92.0), debug profile

## Notes

Reproduced on unmodified `master`, so it is not caused by any pending PR. I ran into it while
verifying #14124 on Windows and was asked to file it separately.

SS-Script:
Narrowed it down, and the answer turned out to be neither of the two guesses: it is not the
font/text-layout path at all — it is the three Windows clipboard tests.

All runs below on clean `master` @ `c16fd426d`, Windows 11 Pro 26120, debug profile.

**Q1 — does `--test-threads=2` still abort?** Yes, every time.

[...]

PR #14322, opened 2026-07-27
fix(winit): apply the shaper's GPOS offsets when converting shaped glyphs

### LINKED ISSUE #14323: Combining marks are drawn at the pen position on winit platforms (shaper GPOS offsets are discarded)
## What happens

On the winit platforms, combining marks are drawn at the pen position instead of on their base, so
they are never positioned by the font's GPOS anchors.

With a *single* mark this is usually invisible: most fonts draw mark glyphs extending left of their
origin, so a mark placed at the pen happens to land over its base. It becomes obvious as soon as two
marks stack, because both then get identical coordinates and are drawn on top of one another.

Thai is where it shows up worst — `พี่`, `ที่`, `เก้าอี้`, `ก๋วยเตี๋ยว` all lose a mark — but this
is not Thai-specific: it affects any script that relies on GPOS mark positioning, and it sits below
the terminal, so it applies to all text `warpui` lays out.

## Where

`RunBuilder::push_glyph` in `crates/warpui/src/windowing/winit/fonts/text_layout.rs` builds each
glyph's position from `LayoutGlyph::x` / `y` alone:

```rust
position_along_baseline: vec2f(glyph.x, glyph.y),
```

Those are the pen position after advance accumulation. cosmic-text keeps the shaped placement in the
separate `x_offset` / `y_offset` fields (em units, to be scaled by `font_size`) — see
`LayoutGlyph::physical` upstream — and nothing in `warpui` or `warpui_core` ever reads them.

macOS is unaffected: `platform/mac/text_layout.rs` gets final positions from CoreText.

## Reproducing it without a terminal

Shaping `b` + U+0301 COMBINING ACUTE ACCENT + U+0308 COMBINING DIAERESIS with the bundled
`RobotoFlex-Semibold.ttf` at 16px returns both marks at `x = 9.5`, which is exactly the base glyph's
advance, and `y = 0`. The shaper reported `x_offset = -0.546875` for both and `y_offset` of
`0.1577` and `0.4443` respectively — i.e. the correct stacked placement was available and discarded.

(The bundled `Roboto-Regular.ttf` cannot show this: it has no COMBINING DIAERESIS and returns zero
offsets for the acute.)

## Expected

Marks are placed using the shaper's offsets, so that stacked marks sit above one another.

## Environment

Windows 11 Pro 26120; also applies to the Linux winit path by inspection, though I have only tested
on Windows.

## Fix

PR #14322 — one-line change in `push_glyph` plus a fail-before/pass-after regression test.

SS-Script:
Closes #14323

## Problem

On the winit platforms, combining marks are drawn at the pen position instead of on their base. With
a single mark that is usually invisible, which is why it has gone unnoticed; with two stacked marks
both land at identical coordinates and the text becomes unreadable.

Thai is where it shows up worst — `พี่`, `ที่`, `เก้าอี้`, `ก๋วยเตี๋ยว` all lose a mark — but this
is not Thai-specific. It affects any script that relies on GPOS mark positioning, and it sits below
the terminal, so it applies to all text `warpui` lays out.

## Root cause

`RunBuilder::push_glyph` built each glyph's position from `LayoutGlyph::x` / `y` alone:

```rust
position_along_baseline: vec2f(glyph.x, glyph.y),
```

Those are the pen position after advance accumulation. cosmic-text keeps the shaped placement in the
separate `x_offset` / `y_offset` fields (em units, to be scaled by `font_size`) — see
`LayoutGlyph::physical` upstream — and nothing in `warpui` or `warpui_core` ever read them. So every
zero-advance combining mark was drawn at the pen rather than on its base.

## Fix

```rust
let gpos_x_offset = glyph.font_size * glyph.x_offset;
let gpos_y_offset = glyph.font_size * glyph.y_offset;
position_along_baseline: vec2f(glyph.x + gpos_x_offset, glyph.y - gpos_y_offset),
```

Scaling and sign mirror `LayoutGlyph::physical`.

macOS is unaffected — `platform/mac/text_layout.rs` gets final positions from CoreText, which is
also why this never showed up in testing on a Mac.

## Test

`test_combining_marks_are_placed_with_gpos_offsets` in
`crates/warpui/src/windowing/winit/text_layout_tests.rs` shapes `b` + COMBINING ACUTE ACCENT +
COMBINING DIAERESIS and asserts that both marks are pulled back over the base and stacked at
different heights.

It uses the bundled `RobotoFlex-Semibold.ttf` rather than the `Roboto-Regular` the neighbouring tests
use, and that is deliberate: Roboto-Regular has no COMBINING DIAERESIS (it shapes to `.notdef`) and
returns a zero offset for the acute, so it cannot exercise mark positioning at all. RobotoFlex
carries both mark-to-base and mark-to-mark anchors, which is what makes the stacked case meaningful.

Fail-before / pass-after — with the fix reverted to `master`:

```
mark should be pulled back over its base by its GPOS x offset,
got x = 9.5 with a base advance of 9.5
test result: FAILED. 0 passed; 1 failed
```

The mark sitting at exactly the base's advance is the bug itself.

With the fix, `cargo test -p warpui --lib -- --test-threads=1` gives
`test result: ok. 41 passed; 0 failed; 10 ignored`, and `cargo fmt -p warpui -- --check` is clean.
(Single-threaded because the suite run in parallel aborts on Windows for unrelated reasons — see
[#14321](https://github.com/warpdotdev/warp/issues/14321), which reproduces on clean `master`.)

## Verification on Windows

Built `warp-oss` from this branch on Windows 11 Pro (26120), terminal font `Hack` so it has no Thai
glyphs and everything goes through the DirectWrite fallback path. Before the fix stacked marks land
on top of one another; after, `พี่ที่ปั๊มน้ำมัน เก้าอี้ ก๋วยเตี๋ยว ฟื้นฟู เพิ่ง` all render
correctly. I have also been running a patched build as my daily driver, in Thai, for real work.

Without this fix (`ที`/`พี` lose their tone marks — note the vowel drawn before its consonant):

without the GPOS fix

With it:

with the GPOS fix

## Relation to #14124

Found while testing #14124 (Thai combining marks in the terminal grid). It is a separate concern —
different crate, and it affects all `warpui` text rather than terminal cells — but that PR cannot be
correct on Windows without it, whether the clustering ends up happening at render time or at input
time. @ultramcu suggested opening it on its own:
https://github.com/warpdotdev/warp/pull/14124#issuecomment-5086705260

### DIFF
--- a/crates/warpui/src/windowing/winit/fonts/text_layout.rs
+++ b/crates/warpui/src/windowing/winit/fonts/text_layout.rs
@@ -82,9 +82,19 @@ impl<'a> RunBuilder<'a> {
             .char_index(glyph.start)
             .unwrap_or_else(|| self.str_index_map.num_chars());
 
+        // `glyph.x` / `glyph.y` are the pen position — where the glyph's cell starts after
+        // accumulating advances. The shaper's GPOS placement lives separately in
+        // `x_offset` / `y_offset`, expressed in em units, and must be folded in or every
+        // zero-advance combining mark lands at the pen instead of on its base. That is what
+        // made Thai stack wrong: in "พี่" both ◌ี and ◌้ came back at x = the base's advance
+        // and y = 0, i.e. side by side after the consonant rather than stacked above it.
+        // Sign convention and the font_size scaling follow `LayoutGlyph::physical` upstream.
+        let gpos_x_offset = glyph.font_size * glyph.x_offset;
+        let gpos_y_offset = glyph.font_size * glyph.y_offset;
+
         self.glyphs_in_current_run.push(Glyph {
             id: glyph.glyph_id as u32,
-            position_along_baseline: vec2f(glyph.x, glyph.y),
+            position_along_baseline: vec2f(glyph.x + gpos_x_offset, glyph.y - gpos_y_offset),
             index: glyph_char_index,
             width: glyph.w,
         });
--- a/crates/warpui/src/windowing/winit/text_layout_tests.rs
+++ b/crates/warpui/src/windowing/winit/text_layout_tests.rs
@@ -583,6 +583,121 @@ fn test_layout_text_first_line_indent_large_bidirectional() -> Result<()> {
     Ok(())
 }
 
+/// Combining marks must be placed with the shaper's GPOS offsets, not with the pen position.
+///
+/// [`cosmic_text::LayoutGlyph`] reports the shaped placement of a glyph in `x_offset` / `y_offset`
+/// (em units, to be scaled by `font_size`), separately from `x` / `y`, which are the pen position
+/// after advance accumulation. Ignoring the offsets draws every zero-advance combining mark at the
+/// pen instead of on its base.
+///
+/// This is invisible with a single mark — most fonts draw mark glyphs extending left of their
+/// origin, so a mark placed at the pen happens to land over its base — and only becomes visible
+/// once two marks stack, since both then get identical coordinates.
+#[test]
+fn test_combining_marks_are_placed_with_gpos_offsets() -> Result<()> {
+    use std::path::PathBuf;
+
+    let (mut font_db, _roboto) = init_fonts();
+
+    // The bundled Roboto-Regular doesn't cover COMBINING DIAERESIS, and carries no mark anchors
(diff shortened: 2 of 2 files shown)


SS-Script:
@cla-bot check

vorporeal (approved):

PR #14332, opened 2026-07-27
[APP-4988] Add TUI key-up lifecycle events

harryalbert:
## Description
Adds a generalized keyboard lifecycle to the headless TUI so features can observe key releases without introducing feature-specific core events. This is the input-layer foundation for configurable hold-to-talk voice activation; the voice setting and behavior remain in the upstack PR.

### Event model
- Adds `is_repeat` to `TuiEvent::KeyDown`; presses and repeats remain key-down events.
- Adds general `TuiEvent::KeyUp` events.
- Carries optional platform-neutral `physical_key` identity directly on TUI key-down/up events, including left/right Shift, Control, Alt, and Super. Shared GUI `KeyEventDetails` is unchanged.
- Adds `TuiEventHandler::on_key_event`, a child-first lifecycle callback with explicit `PropagateToParent` / `StopPropagation` behavior.

### Dispatch behavior
- Keymaps remain press/repeat-driven. Key-up events bypass the keymap, matching the GUI's separation between press-driven keybindings and `ModifierKeyChanged` lifecycle events.
- Editors, selectors, and embedded PTYs preserve their existing press/repeat behavior and ignore releases by default.
- Components must explicitly opt into releases through `on_key_event`.

### Terminal protocol and compatibility
- On hosts that report Kitty keyboard-enhancement support, the TUI requests disambiguated keys, event types, alternate keys, and all keys as escape sequences (flags `1 + 2 + 4 + 8 = 15`). This enables standalone modifier events plus press/repeat/release reporting.
- Unsupported hosts remain on legacy input: normal typing and keybindings continue working, but standalone modifier lifecycle and general key-up events are unavailable. Protocol state is popped when the TUI exits.
- Alternate key reporting preserves terminal-produced shifted text such as `Shift+A -> A` and `Shift+1 -> !`. Uppercase letters recover Shift for keymap compatibility. Punctuation is treated as terminal-authored text; the conversion deliberately does not guess US-layout metadata such as `! -> Shift+1`.

This PR does not add or change the voice-input setting itself. The upstack implementation gates hold/release behavior on `TuiDriverHandle::keyboard_enhancement_supported()`.

Implementation plan: https://staging.warp.dev/drive/notebook/zS1eWZnN57wdRYs38rIEQO
Agent conversation: https://staging.warp.dev/conversation/b3063bb7-92f9-4249-a334-79a8a0fc73fa

## Testing

## Agent Mode

### DIFF
--- a/crates/warpui_core/src/elements/tui/event.rs
+++ b/crates/warpui_core/src/elements/tui/event.rs
@@ -30,7 +30,12 @@ pub enum TuiEvent {
         keystroke: Keystroke,
         chars: String,
         details: KeyEventDetails,
+        is_repeat: bool,
         is_composing: bool,
+    },
+    KeyUp {
+        keystroke: Keystroke,
+        details: KeyEventDetails,
     },
     Paste {
         text: String,
@@ -83,7 +88,7 @@ impl TuiEvent {
             | Self::MiddleMouseDown { position, .. }
             | Self::RightMouseDown { position, .. }
             | Self::MouseMoved { position, .. } => Some(*position),
-            Self::KeyDown { .. } | Self::Paste { .. } => None,
+            Self::KeyDown { .. } | Self::KeyUp { .. } | Self::Paste { .. } => None,
         }
     }
 
--- a/crates/warpui_core/src/elements/tui/event.rs
+++ b/crates/warpui_core/src/elements/tui/event.rs
@@ -16,7 +16,7 @@ use super::{
     TuiElement, TuiLocalPoint, TuiPoint, TuiScene, TuiScreenPoint, TuiScreenRect, TuiSize,
     TuiViewMapContext, TuiZIndex,
 };
-use crate::event::{KeyEventDetails, ModifiersState};
+use crate::event::{KeyEventDetails, KeyState, ModifiersState};
 use crate::keymap::Keystroke;
 use crate::platform::keyboard::KeyCode;
 use crate::{Action, EntityId, EntityIdMap};
@@ -31,14 +31,11 @@ pub enum TuiEvent {
         keystroke: Keystroke,
         chars: String,
         details: KeyEventDetails,
-        physical_key: Option<KeyCode>,
-        is_repeat: bool,
         is_composing: bool,
     },
-    KeyUp {
-        keystroke: Keystroke,
-        details: KeyEventDetails,
-        physical_key: Option<KeyCode>,
+    ModifierKeyChanged {
+        key_code: KeyCode,
+        state: KeyState,
     },
     Paste {
         text: String,
@@ -91,7 +88,7 @@ impl TuiEvent {
             | Self::MiddleMouseDown { position, .. }
             | Self::RightMouseDown { position, .. }
             | Self::MouseMoved { position, .. } => Some(*position),
-            Self::KeyDown { .. } | Self::KeyUp { .. } | Self::Paste { .. } => None,
+            Self::KeyDown { .. } | Self::ModifierKeyChanged { .. } | Self::Paste { .. } => None,
         }
     }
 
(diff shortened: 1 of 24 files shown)


harryalbert:
* **#14307** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14307?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14332** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14332?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14332?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warpui_core/src/elements/tui/event.rs
```diff
@@ -30,8 +31,15 @@ pub enum TuiEvent {
         keystroke: Keystroke,
         chars: String,
         details: KeyEventDetails,
+        physical_key: Option<KeyCode>,
+        is_repeat: bool,
         is_composing: bool,
     },
+    KeyUp {
```
kevinyang372: I am a bit worried that we are over-generalizing the events here

I think the main event type we want to mirror from GUI is `ModifierKeyChanged`, which is specifically designed for action triggers like holding down the Fn key for voice input or double tapping cmd to trigger some actions. This is also the main event where we really need a physical key code payload

From my understanding this should be the same paradigm we want for TUI and specifically for the voice input feature. Is there any specific reason we want to generalize it for all keys?

crates/warpui_core/src/runtime/mod.rs
```diff
@@ -661,7 +664,12 @@ fn enter_terminal_screen(
     if keyboard_enhancement_supported {
         let _ = execute!(
             out,
-            PushKeyboardEnhancementFlags(KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES)
+            PushKeyboardEnhancementFlags(
+                KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES
+                    | KeyboardEnhancementFlags::REPORT_EVENT_TYPES
+                    | KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS
```
kevinyang372: Agent is pointing out that this flag is not really necessary
```
REPORT_ALTERNATE_KEYS appears unnecessary for this feature and is what introduces
much of the shifted-character normalization complexity
```

Is this valid?

crates/warpui_core/src/elements/tui/event.rs
harryalbert: spoke offline, but will update

crates/warpui_core/src/runtime/mod.rs
harryalbert: I don't think so.

`REPORT_ALL_KEYS_AS_ESCAPE_CODES` is required to receive standalone modifier events. With that enabled, shifted printable keys are otherwise reported using the unshifted key—for example, Shift+1 comes through as 1 plus Shift rather than !.

The Kitty protocol has an associated-text field that would solve this, but Crossterm 0.29 doesn’t expose it. `REPORT_ALTERNATE_KEYS` is how Crossterm recovers the actual shifted character. Without it, we’d either insert the wrong character or have to guess the user’s keyboard layout.

kevinyang372 (approved):
As discussed offline, we should change the behavior to either:
1. (Ideally): we could read the shift modifier key directly from cross-term -- so we get GUI parity keyboard events like `ctrl-shift-!`
2. If we can't read the shift modifier key directly, we should just preserve the shift modified semantics for all keyboard events so `ctrl-shift-A` -> `ctrl-A`, `ctrl-shift-!` -> `ctrl-!`

PR #14012, opened 2026-07-20
Encode Cmd/Option modifiers for editing keys under the Kitty keyboard protocol

FreddieMcHeart:
# Encode Cmd/Option modifiers for editing keys under the Kitty keyboard protocol

Fixes #9159. Related spec: #13695.

## Problem

On macOS, `Cmd`/`Option` combined with **Backspace, Delete, and the arrow keys**
are silently dropped when a TUI enables the Kitty keyboard protocol
(`DISAMBIGUATE_ESCAPE`). So inside apps like Claude Code:

- `⌘⌫` (delete to line start), `⌥⌫` (delete word), `⌘←`/`⌘→`, `⌥←`/`⌥→` do nothing.

The same keystrokes work in Kitty and Ghostty, which emit the correct sequences.
`Ctrl+…` works in Warp because it is handled in both encoding layers; `Cmd` and
`Option` are not.

## Root cause

The modifier plumbing (`NSEvent → Keystroke.cmd/alt`) is intact. The loss is
entirely in the escape-sequence encoder's gating logic:

1. **CSI-u gate** (`kitty_keyboard_protocol.rs`) never checked `keystroke.cmd`
   and blanket-excluded `alt` on macOS, so `Cmd+Backspace` / `Option+Backspace`
   never reached CSI-u encoding.
2. **Legacy modifier byte** (`to_modifier_escape_byte`) used by arrows/Delete
   wildcarded `cmd` and was a single ASCII digit — it could not represent Super
   (value 9) or multi-digit combos, so the modifier was dropped.
3. **Function keys** shared the same gap (`Cmd+F5` dropped the modifier).

## Changes

- **Gate**: treat any `Cmd`-modified key as ambiguous (Super has no legacy
  encoding at all). Stop blanket-excluding `Option` on macOS; instead detect IME
  composition from the OS-provided associated text (`chars`), so `Option+Space`
  → nbsp still composes, while `Option` + Backspace/Delete/arrows (which produce
  no text) act as modifiers.
- **`modifier_param` helper**: one source of truth for the xterm/Kitty modifier
  value `1 + shift + alt + ctrl + super`, shared by arrows, Delete, and function
  keys. Replaces the single-byte `to_modifier_escape_byte` (now removed), and
  encodes Super and multi-digit values.
- **Delete**: `CSI 3;<mods>~` for modified Delete (previously no modifier-aware
  path existed).

Per the Kitty spec, keys that own a legacy escape code (arrows, Home/End, Delete,
[...]

PR #14335, opened 2026-07-27
Fix macOS line-editing shortcuts inside alt-screen TUIs (#807)

### LINKED ISSUE #807: OPTION/CMD + ARROW LEFT/RIGHT or BACKSPACE not working as intended when Warp is in long-running command mode

2solarmax:
## Description

When a readline-style TUI runs on the alternate screen (Claude Code in `fullscreen` mode, opencode, etc.), the macOS line-editing shortcuts stop working: `cmd+Backspace`, `cmd+Delete`, `cmd+Left`, `cmd+Right` and `alt+Backspace` no longer perform their editing actions.

**Root cause.** The seven `terminal:executing_command_*` bindings in `app/src/terminal/view/init.rs` are gated on `id!("LongRunningCommand")`, and that context key is explicitly suppressed while the alt screen is active — `app/src/terminal/view.rs`:

```rust
if active_block.is_active_and_long_running() {
    if !model_lock.is_alt_screen_active() {
        context.set.insert("LongRunningCommand");
    }
    ...
}
```

So `LongRunningCommand` and `AltScreen` are mutually exclusive by construction, and the predicate can never be satisfied while an alt-screen TUI is in the foreground. The keystroke falls through to the encoder, which wildcards `cmd` on macOS, so nothing useful reaches the PTY — the key silently does nothing.

**Fix.** Widen the predicate on those seven bindings to

```rust
id!("LongRunningCommand") | (id!("AltScreen") & !id!(KEYBOARD_PROTOCOL_ENABLED_KEY))
```

This is the same shape as the `shift-enter` binding at `init.rs:117-124`, and `terminal:backward_tabulation` in this very block already uses `(LongRunningCommand | AltScreen)`. Behaviour inside long-running command blocks is unchanged.

**Why the Kitty keyboard protocol clause is there.** A previous attempt at this fix (#11605) was reviewed and closed; the outstanding review concern was that a bare `| AltScreen` would also match while KKP is active, so KKP-aware TUIs would receive readline control bytes instead of the rich key events they explicitly requested. That concern is valid — bindings are matched before the encoder runs, so a matching binding wins over KKP encoding. Excluding `KEYBOARD_PROTOCOL_ENABLED_KEY` resolves it structurally: TUIs that negotiated KKP are left completely untouched by this change, and only non-KKP alt-screen TUIs pick up the editing bindings.

## Linked Issue

Fixes #807 (open since 2022, labeled `ready-to-implement`).

Related: #9159 / #14012 cover the complementary gap — encoding `cmd`/`option` modifiers *within* the KKP encoder for TUIs that do negotiate the protocol. The two changes do not overlap: this PR only affects sessions where KKP is off, which is precisely where the encoder has nothing to send.

Supersedes #11605, which was auto-closed by the stale-PR workflow after the author stopped responding to review, not on merit.

## Testing

### Visual evidence

Both captures come from a local debug build of this branch. The chords are driven into a small
alt-screen probe: it enters the alternate screen (`CSI ? 1049 h`), goes into raw mode, and prints
the raw bytes it reads from stdin — the smallest faithful stand-in for Claude Code / opencode, a
fullscreen TUI doing its own line editing.

**1. Fix working — plain alt screen, no keyboard protocol.** All seven bindings fire, delivering
exactly the bytes their actions specify:

alt-screen key probe with the fix, no keyboard protocol

| chord | byte | action |
|---|---|---|
| <kbd>cmd</kbd>+<kbd>backspace</kbd> | `15` NAK ^U | `terminal:executing_command_delete_line_start` |
| <kbd>cmd</kbd>+<kbd>delete</kbd> | `0b` VT ^K | `terminal:executing_command_delete_line_end` |
| <kbd>cmd</kbd>+<kbd>left</kbd> | `01` SOH ^A | `terminal:executing_command_move_cursor_home` |
| <kbd>cmd</kbd>+<kbd>right</kbd> | `05` ENQ ^E | `terminal:executing_command_move_cursor_end` |
| <kbd>alt</kbd>+<kbd>backspace</kbd> | `17` ETB ^W | `terminal:executing_command_delete_word_left` |
| <kbd>alt</kbd>+<kbd>left</kbd> | `1b 62` ESC b | `terminal:executing_command_move_cursor_word_left` |
| <kbd>alt</kbd>+<kbd>right</kbd> | `1b 66` ESC f | `terminal:executing_command_move_cursor_word_right` |

On `master` the same pane emits nothing for any of these chords — that is the bug in #807.

**2. Guard working — same build, same chords, KKP enabled.** Identical build and keystrokes,
except the probe pushed `CSI > 1 u` first:

alt-screen key probe with keyboard protocol enabled

None of `0x15 / 0x0b / 0x01 / 0x05 / 0x17` appear — the chords fall through to their normal
encodings (`7f`, `1b 5b 44`, `1b 5b 31 3b 33 44`, …), so a keyboard-protocol TUI still receives
the rich key events it asked for. That is the `& !id!(KEYBOARD_PROTOCOL_ENABLED_KEY)` half of the
predicate, exercised at runtime rather than only in the unit test. (The repeated `1b 4f 41` — SS3
A, up-arrow — lines are noise from the scripted key driver; none of the seven chords produce that
sequence. The point of the capture is the absence of the five C0 bytes above.)

### Automated coverage

Two new tests in `app/src/terminal/input_tests.rs`. They look up each binding **as actually
registered by `terminal::view::init`** and evaluate its predicate with
`EditableBindingLens::in_context` against the exact context sets `TerminalView::keymap_context`
produces — so they exercise the shipped predicate rather than a copy of it.

`test_line_editing_bindings_apply_in_alt_screen_without_keyboard_protocol`, for all seven bindings:

| context | expected |
|---|---|
| `Terminal`, `AltScreen` | **active** — the regression this PR fixes |
| `Terminal`, `LongRunningCommand` | active — unchanged behaviour |
| `Terminal` | inactive — not over-broadened; at an idle prompt these chords still reach Warp's own input editor |
| `Terminal`, `AltScreen`, `IMEOpen` | inactive — the pre-existing IME guard still wins |

`test_line_editing_bindings_suppressed_in_alt_screen_with_keyboard_protocol`:

| context | expected |
|---|---|
| `Terminal`, `AltScreen`, `KeyboardProtocolEnabled` | **inactive** |
| `Terminal`, `LongRunningCommand`, `KeyboardProtocolEnabled` | active — the new clause only applies on the alt screen |

Both were run against three variants of `init.rs` to confirm neither passes vacuously:

| `init.rs` variant | apply test | suppress test |
|---|---|---|
| upstream `c20645b`, unmodified | **FAIL** | pass |
| naive `\| id!("AltScreen")` with no KKP guard (the #11605 shape) | pass | **FAIL** |
| this PR | pass | pass |

The middle row is the point: it shows the `KEYBOARD_PROTOCOL_ENABLED_KEY` exclusion is
load-bearing and is itself covered by a test — the review concern that blocked #11605.

On unmodified upstream:

```
assertion `left == right` failed: `terminal:executing_command_move_cursor_word_left` in context {"Terminal", "AltScreen"} should be active
  left: false
 right: true
```

On the naive variant:

```
assertion `left == right` failed: `terminal:executing_command_move_cursor_word_left` in context {"AltScreen", "KeyboardProtocolEnabled", "Terminal"} should be inactive
  left: true
 right: false
```

## Agent Mode

CHANGELOG-BUG-FIX: Fixed macOS line-editing shortcuts (cmd+backspace, cmd+delete, cmd+left/right, alt+backspace) not working inside fullscreen TUIs such as Claude Code and opencode.

### DIFF
--- a/app/src/terminal/input_tests.rs
+++ b/app/src/terminal/input_tests.rs
@@ -101,6 +101,7 @@ use crate::terminal::resizable_data::ResizableData;
 use crate::terminal::shared_session::permissions_manager::SessionPermissionsManager;
 use crate::terminal::shell::ShellType;
 use crate::terminal::universal_developer_input::UniversalDeveloperInputButtonBarEvent;
+use crate::terminal::view::init::KEYBOARD_PROTOCOL_ENABLED_KEY;
 use crate::terminal::view::inline_banner::ByoLlmAuthBannerSessionState;
 use crate::terminal::writeable_pty::command_history::update_command_history;
 use crate::test_util::settings::initialize_settings_for_tests;
@@ -9443,5 +9444,120 @@ fn ctrl_enter_inserts_newline_in_normal_input_after_rich_input_closes() {
                  (the default); got Emit instead"
             );
         });
+    });
+}
+
+/// The seven readline-style line-editing bindings whose context predicate this change widens.
+const EXECUTING_COMMAND_LINE_EDITING_ACTIONS: &[&str] = &[
+    "terminal:executing_command_move_cursor_word_left",
+    "terminal:executing_command_move_cursor_word_right",
+    "terminal:executing_command_move_cursor_home",
+    "terminal:executing_command_move_cursor_end",
+    "terminal:executing_command_delete_word_left",
+    "terminal:executing_command_delete_line_start",
+    "terminal:executing_command_delete_line_end",
+];
+
+/// Build a keymap context containing exactly `keys`, mirroring what
+/// `TerminalView::keymap_context` inserts in the corresponding terminal state.
+fn keymap_context_with(keys: &[&'static str]) -> warpui::keymap::Context {
+    let mut context = warpui::keymap::Context::default();
+    for key in keys {
+        context.set.insert(key);
+    }
+    context
+}
+
+/// Assert how every line-editing binding evaluates against `context`.
+///
+/// Reads the predicate off the binding as actually registered by
+/// `crate::terminal::view::init`, so this exercises the shipped predicate rather than a
+/// copy of it.
+fn assert_line_editing_bindings(app: &App, context: &warpui::keymap::Context, expected: bool) {
+    app.read(|ctx| {
+        for name in EXECUTING_COMMAND_LINE_EDITING_ACTIONS {
+            let binding = ctx
+                .editable_bindings()
+                .find(|binding| binding.name == *name)
+                .unwrap_or_else(|| panic!("`{name}` should be a registered editable binding"));
+            assert_eq!(
+                binding.in_context(context),
+                expected,
+                "`{name}` in context {:?} should be {}",
+                context.set,
+                if expected { "active" } else { "inactive" },
+            );
+        }
+    });
+}
+
+/// The `terminal:executing_command_*` line-editing bindings must be active inside an
+/// alt-screen TUI that has not enabled the Kitty keyboard protocol.
+///
+/// Before this change they were gated on `LongRunningCommand` alone, and
+/// `TerminalView::keymap_context` explicitly declines to insert that key while the alt
+/// screen is active — so all seven were unreachable in fullscreen TUIs.
+#[test]
+fn test_line_editing_bindings_apply_in_alt_screen_without_keyboard_protocol() {
+    App::test((), |mut app| async move {
+        initialize_app(&mut app);
+        // `initialize_app` registers the *input* bindings; the line-editing bindings under
+        // test are registered by the terminal view's own `init`.
+        app.update(crate::terminal::view::init::init);
+
+        // The regression under test: a plain alt-screen TUI.
+        assert_line_editing_bindings(&app, &keymap_context_with(&["Terminal", "AltScreen"]), true);
+
+        // Unchanged: a long-running command on the normal screen still matches.
+        assert_line_editing_bindings(
+            &app,
+            &keymap_context_with(&["Terminal", "LongRunningCommand"]),
+            true,
+        );
+
+        // Not over-broadened: an idle terminal at the prompt must not match, so these
+        // chords keep reaching Warp's own input editor.
+        assert_line_editing_bindings(&app, &keymap_context_with(&["Terminal"]), false);
+
+        // The pre-existing IME guard still wins over the widened clause.
+        assert_line_editing_bindings(
+            &app,
+            &keymap_context_with(&["Terminal", "AltScreen", "IMEOpen"]),
+            false,
+        );
+    });
+}
+
+/// The same bindings must stay out of the way when the alt-screen TUI *has* enabled the
+/// Kitty keyboard protocol: it asked for rich key events, and the encoder — not a
+/// readline control byte — is what should answer.
+///
+/// Bindings are matched before the encoder runs, so without this clause a matching
+/// binding would win over KKP encoding. This was the review concern that blocked #11605.
+#[test]
+fn test_line_editing_bindings_suppressed_in_alt_screen_with_keyboard_protocol() {
+    App::test((), |mut app| async move {
+        initialize_app(&mut app);
+        // `initialize_app` registers the *input* bindings; the line-editing bindings under
+        // test are registered by the terminal view's own `init`.
+        app.update(crate::terminal::view::init::init);
+
+        assert_line_editing_bindings(
+            &app,
+            &keymap_context_with(&["Terminal", "AltScreen", KEYBOARD_PROTOCOL_ENABLED_KEY]),
+            false,
+        );
+
+        // A long-running command on the normal screen is unaffected by the protocol: the
+        // alt screen is the only place the new clause applies.
+        assert_line_editing_bindings(
+            &app,
+            &keymap_context_with(&[
+                "Terminal",
+                "LongRunningCommand",
+                KEYBOARD_PROTOCOL_ENABLED_KEY,
+            ]),
+            true,
+        );
     });
 }
--- a/app/src/terminal/view/init.rs
+++ b/app/src/terminal/view/init.rs
@@ -468,6 +468,11 @@ pub fn init(app: &mut AppContext) {
         .with_context_predicate(
             id!("Terminal") & !id!("IMEOpen") & id!("TerminalView_NonEmptyBlockList"),
         ),
+        // The readline-style editing bindings below apply while a command is executing and
+        // inside alt-screen TUIs, so line editing keeps working in fullscreen programs
+        // (Claude Code, opencode, ...). They are suppressed when the TUI has enabled the
+        // Kitty keyboard protocol: it asked for rich key events and should receive those
+        // rather than these control bytes. Same shape as the `shift-enter` binding above.
         EditableBinding::new(
             "terminal:executing_command_move_cursor_word_left",
             "Move cursor one word to the left within an executing command",
@@ -475,7 +480,12 @@ pub fn init(app: &mut AppContext) {
         )
         .with_mac_key_binding("alt-left")
         .with_linux_or_windows_key_binding("ctrl-left")
-        .with_context_predicate(id!("Terminal") & !id!("IMEOpen") & id!("LongRunningCommand")),
+        .with_context_predicate(
+            id!("Terminal")
+                & !id!("IMEOpen")
+                & (id!("LongRunningCommand")
+                    | (id!("AltScreen") & !id!(KEYBOARD_PROTOCOL_ENABLED_KEY))),
+        ),
         EditableBinding::new(
             "terminal:executing_command_move_cursor_word_right",
             "Move cursor one word to the right within an executing command",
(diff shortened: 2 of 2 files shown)


2solarmax:
Visual evidence, as requested.

Both captures come from a local debug build of this branch (`./script/run`). The chords are
driven into a small alt-screen probe: it enters the alternate screen (`CSI ? 1049 h`), goes
into raw mode, and prints the raw bytes it reads from stdin. That is the smallest faithful
stand-in for Claude Code / opencode — a fullscreen TUI doing its own line editing.

### 1. Fix working — plain alt screen, no keyboard protocol

pub-evidence-fixed

All seven bindings fire, delivering exactly the bytes their actions specify:

| chord | byte | action |
|---|---|---|
| ⌘⌫ | `15` NAK ^U | `terminal:executing_command_delete_line_start` |
| ⌘⌦ | `0b` VT ^K | `terminal:executing_command_delete_line_end` |
| ⌘← | `01` SOH ^A | `terminal:executing_command_move_cursor_home` |
| ⌘→ | `05` ENQ ^E | `terminal:executing_command_move_cursor_end` |
| ⌥⌫ | `17` ETB ^W | `terminal:executing_command_delete_word_left` |
| ⌥← | `1b 62` ESC b | `terminal:executing_command_move_cursor_word_left` |
| ⌥→ | `1b 66` ESC f | `terminal:executing_command_move_cursor_word_right` |

On `master` the same pane emits nothing for these chords: `TerminalView::keymap_context`
inserts `LongRunningCommand` only when the alt screen is *not* active
(`app/src/terminal/view.rs`), so `id!("LongRunningCommand")` can never match inside a TUI.
That is the mechanism behind #807.

### 2. Guard working — same build, same chords, KKP enabled

pub-evidence-kkp

Identical build, identical keystrokes, except the probe pushed `CSI > 1 u` first. None of
`0x15 / 0x0b / 0x01 / 0x05 / 0x17` appear — the chords fall through to their normal
encodings (`7f`, `1b 5b 44`, `1b 5b 31 3b 33 44`, …), so a keyboard-protocol TUI still
receives the rich key events it asked for. That is the
`& !id!(KEYBOARD_PROTOCOL_ENABLED_KEY)` half of the predicate, exercised at runtime rather
than only in the unit test.

(The repeated `1b 4f 41` — SS3 A, up-arrow — lines are noise from the scripted key driver;
none of the seven chords produce that sequence. The point of the capture is the absence of
the five C0 bytes above.)

Both cases are also covered by the predicate-level tests added in this PR:
`test_line_editing_bindings_apply_in_alt_screen_without_keyboard_protocol` and
`test_line_editing_bindings_suppressed_in_alt_screen_with_keyboard_protocol`
(`cargo test -p warp --lib line_editing_bindings` → 2 passed).

/oz-review

2solarmax:
@cla-bot check

2solarmax:
The visual evidence is now embedded in the PR description itself (Testing → Visual evidence), not only in the comment above — two captures from a local debug build of this branch: the seven chords delivering their bytes on a plain alt screen, and the same chords producing no readline bytes once the TUI pushes `CSI > 1 u`. The manual-testing checkbox is ticked accordingly.

/oz-review

acarl005 (commented):
@2solarmax I'm not convinced this is necessary. You mention "Claude Code in fullscreen mode" as a motivation for this PR, but in my testing I see that #14012 is sufficient for fixing Claude Code. What real world programs actually depend on the functionality that you are proposing in this PR? Is this backed up by ghostty, alacritty, or wezterm, etc?

2solarmax:
You're right, and thanks for testing it rather than taking my word for it. Let me correct the record and then make a narrower case.

**On Claude Code — you're correct and my PR body is wrong.** I captured Claude Code's startup under a pty and it pushes the kitty protocol:

```
\x1b[>1u        KKP push
\x1b[?1049h     alt screen
\x1b[>4;2m      modifyOtherKeys
```

So `& !id!(KEYBOARD_PROTOCOL_ENABLED_KEY)` in my own predicate excludes Claude Code by construction. This PR never fixed it. #14012 is the fix, exactly as you say. I had it backwards from an earlier bad grep, and my evidence capture used a probe that deliberately didn't enable KKP, so my own testing never caught it. The PR body and changelog need rewriting regardless of what happens to the diff.

**Where the motivation actually came from**, since "Claude Code in fullscreen" was doing a lot of unearned work in that description: I don't run Claude Code in fullscreen by choice. In the default renderer it doesn't lay out correctly in my panes — the input area and session header get clipped at the right edge, and it doesn't fill the space allotted to it, in split panes and in a single maximised pane alike. Fullscreen fixes the layout and breaks the keys. So the honest framing is a rendering bug pushing users onto the alt screen, where a second problem is waiting. I'll file the rendering one separately with a clean repro; it isn't this PR's business and I shouldn't have smuggled it in as motivation.

**"What real world programs?"** — thinner than I implied. I probed the alt-screen TUIs I have. The ones that negotiate KKP are excluded by my own guard, which covers both programs I named. Of what's left, `fzf` is the clean case: its query line binds `ctrl-a` beginning-of-line, `ctrl-e` end-of-line, `ctrl-u` unix-line-discard, `ctrl-w` unix-word-rubout — byte-for-byte what these bindings send. But `nano` is a genuine counter-example. `^U` there is UnCut, i.e. paste: I cut a line with `^K`, sent `^U`, and the text came back. So this PR would turn cmd+delete into "cut the line" and cmd+backspace into "paste it back". Turning a no-op into a wrong action is worse than the no-op, and that alone kills the seven-binding version.

**"Is this backed up by ghostty?"** — yes, for five of the seven, as unconditional macOS defaults. `src/config/Config.zig`, inside `if (comptime builtin.target.os.tag.isDarwin())`:

```zig
// "Natural text editing" keybinds. This forces these keys to go back
// to legacy encoding (not fixterms). It seems macOS users more than
// others are used to these keys so we set them as defaults.
super+arrow_right -> "\x05"   super+arrow_left -> "\x01"
super+backspace   -> "\x15"   alt+arrow_left -> esc "b"   alt+arrow_right -> esc "f"
```

Notably ghostty omits exactly the two that cause trouble — `cmd+delete`→`^K` and `alt+backspace`→`^W`. iTerm2 ships the same five in its bundled "Natural Text Editing" preset. Ghostty's version is also *unconditional* and deliberately overrides fixterms, so it doesn't support my KKP guard — it argues for something simpler than what I proposed.

**One thing worth deciding independently of this PR.** On master these seven bindings are gated on `id!("Terminal") & !id!("IMEOpen") & id!("LongRunningCommand")`, with no keyboard-protocol condition anywhere in that block. So today, during a long-running command that has pushed KKP but is not on the alt screen, Warp already sends `^U` to a KKP-negotiating app — that is why these keys work in Claude Code's default renderer. The concern that blocked #11605, that KKP clients must not receive readline control bytes, is already not Warp's shipped behaviour off the alt screen. So the current state is inconsistent either way: either the normal-screen behaviour is right and the alt screen should match it, or the alt screen is right and the normal-screen path should be gated too. I don't think that's mine to decide, but it seems worth an answer.

**Concretely**, I see three options and I'm happy with any of them:

1. Close this in favour of #14012, and I file the rendering bug separately.
2. Narrow it to `cmd+left` / `cmd+right` / `cmd+backspace` only — the three with no workaround in Warp today, since cmd isn't a terminal modifier and there's no `send_text`-style action to bind. Drop `cmd+delete` and `alt+backspace` (wrong in fzf and nano), and drop `alt+left`/`alt+right`, which already encode as `CSI 1;3D`/`CSI 1;3C` and would change what tmux-style consumers see.
3. If the answer to the consistency question above is "the normal-screen behaviour is correct", the KKP guard comes out entirely and this becomes the simpler change ghostty ships.

Tell me which and I'll rewrite it, or close it. Either way the description gets fixed — it shouldn't claim to fix Claude Code, and I'll re-check whether it's the right fix for #807, whose report is an ssh + `sudo su -` shell prompt rather than an alt-screen TUI.

acarl005:
Closing this out due to scope creep. This is no longer about the originally linked issue. Please open a new issue if you want to push this work forward.

PR #14343, opened 2026-07-27
Add a "None" option to compact vertical tabs' Additional metadata

### LINKED ISSUE #14342: Compact vertical tabs: "Additional metadata" cannot be turned off
### Describe the bug

In the vertical tabs settings popup, the **Additional metadata** section (shown when Density is Compact) is a mandatory choice between exactly two fields. Whichever two fields are not already used by "Pane title as" are offered, and one of them is always selected. With *Pane title as: Command / Conversation* the choices are *Branch* and *Working Directory*, and there is no way to pick neither. Every compact row therefore always renders a second line.

There is no deselect path in the UI: clicking the option that is already checked just re-selects it.

Picking *Branch* is not a workaround either. When a pane has no git branch, that subtitle falls back to showing the working directory, so a second line still appears.

The APP-3742 spec already treats an omitted subtitle as a designed state. Goal 4 asks that slots without meaningful data "gracefully omit that slot rather than showing empty or misleading content", and acceptance criterion 9 repeats it: "Pane types without meaningful metadata for a slot omit that slot — no empty lines or placeholder text." Compact terminal rows render correctly with no subtitle; there is just no way for a user to ask for that.

### To reproduce

1. Open the vertical tabs panel and click the settings button at its top.
2. Set *Pane title as* to **Command / Conversation**, and Density to **Compact**.
3. Under *Additional metadata*, try to end up with neither *Branch* nor *Working Directory* selected.
4. One of the two stays checked, and every compact tab row keeps its second line.

### Expected behavior

A **None** option under *Additional metadata*, so compact rows can show the pane title on its own — the omitted-slot state the spec already describes.

For anyone running many agent sessions side by side, the pane title is the only thing that distinguishes rows, and the repeated branch or directory line under each one is noise that costs vertical space.

### Operating system (OS)

macOS 26.5.2 (arm64)

### Warp version

v0.2026.07.22.09.01.01

---

I have a patch for this and will open a PR linked to this issue.

### Screenshots

Current behaviour — the subtitle line cannot be removed:

2solarmax:
## Description

Adds a **None** option to *Additional metadata* in the vertical tabs settings popup, so compact rows can show the pane title on its own.

**The omitted-slot state is already designed.** The APP-3742 spec asks that slots without meaningful data "gracefully omit that slot rather than showing empty or misleading content" (goal 4), and acceptance criterion 9 repeats it: "Pane types without meaningful metadata for a slot omit that slot — no empty lines or placeholder text." `render_compact_pane_row` already implements this — it builds the subtitle as an `Option` and re-centers the row icon when it is absent.

**Terminal panes could not reach it.** `Additional metadata` is a mandatory choice between the two categories not used by "Pane title as", and `render_compact_subtitle_option` only ever dispatches `SetVerticalTabsCompactSubtitle(value)`, so there is no deselect path. Selecting *Branch* does not get there either: `compact_branch_subtitle_display` substitutes the working directory when a pane has no branch, so the slot is filled with a different category rather than omitted.

**Shape of the change.** `VerticalTabsCompactSubtitle` gains a `None` variant, following the existing `WarpPromptSeparator` precedent in `app/src/settings/editor.rs`, which already models "no value" as a `None` variant on a settings enum. `None` names no category, so it can never collide with the primary line and `resolve_compact_subtitle` honors it as stored.

`subtitle_options_for_primary` and the popup's mouse-state array are paired by index, and were a fixed-size 2. Both are now bound to a new `SUBTITLE_OPTION_COUNT` constant, so adding an option without a matching mouse state is a compile error rather than an out-of-bounds panic on popup open.

**Compatibility.** `Branch` stays `#[default]`, so nothing changes for existing users on upgrade. The setting is cloud-synced, and `SettingsValue` deserializes via `serde_json::from_value(..).ok()` with `Setting::new(None)` falling back to the default, so an older build that reads a synced `"none"` degrades to `Branch` rather than failing to parse.

**Deliberately out of scope.** Non-terminal panes (Code and similar) take a different path and draw their subtitle from `pane_configuration.title_secondary()`, which this setting has never controlled. That text also feeds `non_terminal_search_text_fragments`, so hiding it would degrade tab search. Left untouched.

The APP-3742 product and tech specs are updated to match.

## Linked Issue

Fixes #14342

The issue was filed alongside this PR and does not carry a readiness label yet. Per CONTRIBUTING's guidance for PRs opened ahead of the label, it is linked above and the change ships with tests and manual verification. Happy to move this to the spec flow instead if you would rather treat it as a feature request than a gap against criterion 9.

## Testing

Verified against a running local build by toggling `appearance.vertical_tabs.compact_subtitle` between `branch` and `none` and observing the row re-render — including the icon returning to vertical centering when the subtitle is absent.

Five unit tests added in `app/src/workspace/view/vertical_tabs_tests.rs`:

- `every_primary_info_offers_a_none_subtitle_option` — every "Pane title as" value offers a way to turn the subtitle off.
- `subtitle_options_never_repeat_the_primary_line` — guards the pre-existing invariant while the option list grows.
- `none_subtitle_is_honored_for_every_primary_line` — `None` survives resolution, so switching "Pane title as" does not silently bring the line back.
- `conflicting_subtitle_still_falls_back_instead_of_hiding` — a stored preference that duplicates the primary line resolves to a real category, never to `None`.
- `compact_subtitle_defaults_to_branch_so_upgrades_do_not_change_tabs` — pins the default.

`cargo clippy -p warp --all-targets --tests -- -D warnings` is clean, and `./script/format --check` and `./script/check_no_inline_test_modules` pass.

### Screenshots / Videos

The new option in the settings popup:

Compact row before (`Additional metadata: Branch`) and after (`None`):

## Agent Mode

CHANGELOG-IMPROVEMENT: Added a "None" option to Additional metadata in the vertical tabs settings, so compact tab rows can show just the pane title.


2solarmax:
@oss-maintainers could I get a readiness call on #14342?

Oz triaged it as a reproducible settings gap and independently confirmed both findings: there is no deselect path in compact `Additional metadata`, and choosing `Branch` still leaves a second line because it falls back to the working directory when a pane has no branch. It landed as `enhancement`, so I am not sure whether you would want `ready-to-implement` or `ready-to-spec` here.

This PR is already implemented against the APP-3742 spec's own acceptance criterion 9 ("Pane types without meaningful metadata for a slot omit that slot"), with unit tests and before/after manual verification. I am happy to convert it into a spec PR instead if you would rather treat it as a feature request rather than a gap against that criterion.

One small ask while it waits: could you add `no-autoclose`? Oz's changes-requested here is the readiness gate rather than review feedback, and the stale timer only resets on my own activity, so it counts down while the PR is actually blocked on your side.

Just merged latest master in. Format, clippy (`-D warnings`) and the vertical-tabs tests are all green.

PR #14347, opened 2026-07-27
Fix cross-window tab drag removing the wrong tab after a mid-drag list shift

### LINKED ISSUE #14346: Cross-window tab drag removes the wrong tab if the source tab list changes mid-drag
### Pre-submit Checks

### Describe the bug

Drag a tab out of one window and onto another, and if anything closes a tab in the source window while you are still holding the mouse, the drop closes the wrong tab there.

I went digging in `cross_window_tab_drag.rs` to work out why. The source window records the dragged tab's **index** when the drag starts, and on drop it removes whatever is sitting at that index. The index is only updated in one situation: if you drag back over the source window's own tab bar, `set_source_placeholder_index` moves it along with the placeholder. Nothing updates it when the source tab list shifts for a reason the drag did not cause.

Two things are wrong afterwards, not one:

- an innocent tab is closed, and its shell goes with it
- the tab you dragged is never removed from the source, so both windows are left holding a tab pointing at the same pane group

The list shifting mid-drag needs no effort. A shell exiting closes its own tab, with no input from you at all (`PaneGroup::Event::Exited` reaches `handle_file_tree_event`, which closes it). Cmd-W does it too, if the source window still has focus.

There is real staleness handling elsewhere in this file, `source_placeholder_consumed` and the bail-out in `perform_handoff`, but it is all scoped to shifts the drag itself causes. At the cleanup site the only guard is a bounds check, and after a tab is removed the stale index is usually still in bounds and simply points at the neighbour.

This sits behind `FeatureFlag::DragTabsToWindows`, which is in `RELEASE_FLAGS` for macOS and Windows, so it is on in stable.

### To reproduce

1. Open two Warp windows. Give window A four tabs; window B can have one.
2. In window A's leftmost tab, run `sleep 30; exit`. That shell closes its own tab half a minute later with no further input, which is what shifts the tab list mid-drag.
3. Straight away, grab window A's third tab and drag it out over window B's tab bar. Keep the button held. Do not let the cursor wander back over window A's own tab bar, that path updates the index and hides the bug.
4. Wait, still holding, until window A's leftmost tab disappears on its own.
5. Drop onto window B.

The dragged tab arrives in window B as expected. In window A, the fourth tab has been closed instead of the third, and the third is still sitting there.

It also reproduces deterministically without the timing, as a unit test: build a four-tab workspace, remove tab 0, then run the drop carrying the drag-start index. It removes the bystander and leaves the dragged tab.

### Expected behavior

The drop removes the tab that was dragged, whatever index it currently occupies, and touches no other tab. If the dragged tab is already gone, because its own shell exited mid-drag, the drop should do nothing rather than remove something else.

### Operating system (OS)

macOS

macOS 26.5.2 (25F84), Apple silicon

### Current Warp version

v0.2026.07.22.09.01.01 (stable). Also read and reproduced against `master` at da06b312.

### Regression

### Additional context

I have a fix ready and would like to open a PR. It captures the dragged pane group's id at drag start and resolves the source tab by identity at drop time, which also lets the index be dropped from the three source-cleanup `DropResult` variants entirely, so the stale-index path stops being representable rather than staying guarded. The same resolution is needed in the put-back branch of `perform_handoff` and in the two places that read the placeholder index for rendering and for the window snapshot.

The public repo history is squashed into a single import commit, so I could not blame this to a specific change.

Related: #14152 (dragging whole tab groups between windows). A group transfer would have to remove several source tabs at once, which is where index-based cleanup gets worse, so it is probably worth fixing this underneath first.

No

2solarmax:
## Description

A cross-window tab drag recorded the source tab's **index** when the drag started, and the drop removed whatever sat at that index. That index only moves in one situation, when you drag back over the source window's own tab bar and `set_source_placeholder_index` walks it along with the placeholder. Nothing moves it when the source tab list shifts for a reason the drag did not cause, and a shell exiting is enough to do that with no input from the user at all.

So the drop removed a bystander, closed its shell, and left the dragged tab in place, with both windows then holding a `TabData` pointing at the same pane group. The existing guard only bounds-checked the index, which does not help: after a tab is removed the stale index is usually still in bounds and just points at the neighbour.

`ActiveDrag` now captures the dragged pane group's `EntityId` at drag start, and the three source-cleanup `DropResult` variants carry that id instead of an index. Dropping the index from those variants is the point of the shape: the stale-index path stops being representable rather than staying guarded.

Three more places on the same gesture needed the same treatment, all found while reviewing the first commit:

- **The put-back branch of `perform_handoff`.** Dropping a tab back onto its own window's tab bar still resolved by the frozen index, and its failure is worse: it unsubscribed and removed a bystander, left the dragged tab's placeholder in place, then re-inserted the real tab, leaving two `TabData` rows on one pane group in one window.
- **Placeholder rendering and the window snapshot.** Both read the same frozen index, so after a mid-drag tab close an innocent tab rendered at zero width, and a `save_app` during the drag persisted a snapshot that omitted a bystander while keeping the placeholder. The drag model still decides *whether* to collapse or skip; only the *which* is now resolved by identity.
- **Removing the last remaining tab.** `remove_tab` treats the last tab as "close the window" via a bare `ctx.close_window()`, which leaves `suppress_detach_panes_on_window_close` false, and `on_window_closed` would then detach the panes of a pane group that now lives in the target window. That path was previously unreachable because the stale index was usually out of bounds and the removal was skipped entirely; resolving by identity reaches it, so it now closes the way the other content-transfer paths do.

Not covered here: `reverse_handoff` and `on_drag_while_inserted` are still index-based, but they operate on indices the drag itself owns inside the preview/target, not on the source list, so they are not affected by this class of shift.

## Linked Issue

Fixes #14346.

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`

Just filed, so it has not been triaged yet. Flagging for a readiness label. Happy to hold until it is labeled.

Related: #14152 (dragging whole tab groups between windows) builds on this same source-cleanup path. A group transfer removes several source tabs at once, which is where index-based cleanup gets worse, so this is probably worth landing underneath it. The open draft PR #14164 for that issue removes source tabs positionally by `{index, count}`, so it carries this same defect at higher arity.

## Testing

Automated, all ungated so they run in default CI:

- `test_cross_window_drop_removes_the_dragged_tab_not_whatever_took_its_index` — seeds a four-tab window, closes tab 0 mid-drag, and asserts the dragged tab is removed and the bystander survives. This is the regression test for the reported bug; on `master`, with the pre-change API, it removes the bystander and leaves the dragged tab.
- `test_cross_window_drop_of_the_last_source_tab_closes_without_detaching_panes` — covers the last-tab path above, asserting the window closes with pane detach suppressed.
- `test_cross_window_drop_ignores_a_source_tab_that_is_already_gone` — a drop whose tab vanished mid-drag removes nothing rather than falling back to an index.
- `drag_captures_the_dragged_pane_group_identity` — both `begin_*_drag` entry points capture the id, so cleanup has something to resolve through.

Gates on this branch:

- `./script/format --check` — clean
- `cargo test -p warp --lib` — 13 pre-existing environmental failures (OpenAI key, notebooks fs, telemetry, terminal, path resolution), identical set to `master`, which shows 15 on the same machine. No drag or tab-group test fails.

Not yet — flagging honestly rather than checking the box. The repro needs two windows and a 30-second timer mid-drag, and `drag_tabs_to_windows` is not in the default feature set, so it needs `./script/run --features drag_tabs_to_windows`. I can record it if you would like it before review; the deterministic unit test above covers the same state transition.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed dragging a tab between windows closing the wrong tab when another tab in the source window closed mid-drag.


2solarmax:
Rebased onto current `master` (was 121 behind). Clean merge, mergeable again at `fcd33e44`, tests and lint green.

This one is only waiting on a readiness label for #14346, which is triaged as `bug, repro:high, area:window-tabs-panes` but hasn't been marked `ready-to-implement` yet, so Oz won't review it.

@oss-maintainers — two small asks:

1. Could someone take a look at #14346 for a readiness label? It's a live data-loss bug on stable (macOS/Windows, `DragTabsToWindows` is in `RELEASE_FLAGS`): a cross-window tab drag removes the source tab by an index captured at drag start, so if anything closes a tab mid-drag — a shell exiting is enough — the drop closes a bystander and leaves the dragged tab behind.
2. Since the PR is blocked on that rather than on me, could you add `no-autoclose`? Per CONTRIBUTING the staleness clock only resets on my activity, so as it stands I'd have to keep poking the thread just to keep it open.

2solarmax:
Still here and ready — the fix is complete with a regression test, and this PR is waiting on issue #14346 getting a readiness label (only maintainers can apply it), so the blocker is on the review side rather than on me. Per the contribution guidelines, could a maintainer apply `no-autoclose`? Happy to make any change a reviewer wants.

PR #14400, opened 2026-07-28
Open repository-browser Markdown links as local files (spec)

### LINKED ISSUE #13434: Modifier-click a Markdown link to open the local file it names (resolve from link text and URL via the file finder)
### Problem

I generate and review a lot of Markdown, including Warp's own product spec (`PRODUCT.md`) and tech spec (`TECH.md`) process. These documents frequently present file references as links whose text or URL names a file/path like `crates/editor/src/render/model/mod.rs` or `mod_tests.rs`. Clicking such a link today follows its URL (which may point at GitHub, a web page, or nowhere useful) rather than opening that file locally in Warp. When reviewing a spec I want to jump straight to the local file the link refers to.

### Where this can be invoked

Anywhere Warp renders Markdown links, i.e. the surfaces that already produce clickable links:
- the Markdown Viewer (opening a `.md` doc),
- Agent Mode / AI output,
- notebooks / plans.

The common trigger is a **rendered Markdown link** — not bare text in prose.

### Proposed behavior

On a rendered Markdown link, a modifier gesture — e.g. **CMD+OPTION+click** — derives a file-name/path candidate from the link and resolves it against the current workspace using the **existing file finder**:

- The candidate comes from the link's **display text**, and when the text isn't itself a file name, from the link's **URL** (e.g. the path component or final segment of the href). The link text is often the file name, but not always — sometimes the path only appears in the URL.
- **Exactly one match** → open that file in Warp.
- **Multiple matches** → open the file finder pre-filtered to the candidates; the user picks.
- **No match** → no-op (or a subtle "no match" hint).
- Honor a `path/to/file.rs:1430` line (and optional column) suffix, in the text or URL, to land on a specific location in the opened file.

### Why scope it to links

- Authoring a file name as a link already implies it's navigable — no heuristic needed to decide which tokens are files.
- Resolution runs through the existing fuzzy finder only when the user invokes the gesture, so there's **no background index** and **no false-positive link noise**.
- The link text and URL are treated only as *sources for a file-name candidate* — Warp doesn't open the URL as a web address; it extracts a path/filename from it and looks that up locally. The URL can point anywhere (e.g. GitHub), so only its path-like portion matters.

### Alternatives considered

- Manually copying names into file search — slow and breaks review flow, and it doesn't even work when the file name isn't in the visible link text but embedded in the URL, so there's nothing useful to select and copy.
- Auto-hyperlinking bare file paths in prose (see #12421) — a different, heavier approach; scoping to existing links avoids the detection/indexing problem entirely.

### Related issues / prior art

Warp already detects and opens file paths in terminal and agent output, but a Markdown link's click follows its URL rather than opening the local file it names. This reuses the existing link onclick handler and file finder.

- #8998 — `.md` links in the agent panel should open in the Markdown Viewer
- #12038 — clicking a file reference should focus the already-open pane
- #13065 / #13280 (PRs) — making Markdown link/file clicks actionable in Agent Mode output
- #9561 — open a file at a specific line via URL scheme/CLI (the line-number half)
- #9908 — clickable-path resolution against the workspace root

rasitakyol:
## Description

Adds product and technical specifications for #13434, following the maintainer's proposed narrow first pass.

The specs define:

- an explicit Command+Option / Control+Alt action for opening supported GitHub and GitLab file links from the active local checkout;
- deterministic repository-root and URL-path matching without branch checkout, selectors, fuzzy search, or network access;
- raw-URL validation, bounded candidate generation, canonical-path containment, and fail-closed symlink handling;
- the exact Markdown, plan, and notebook surfaces in scope, including the missing plan link-model subscription;
- editor gesture, URL parser, filesystem, remote-context, and pane integration coverage required before implementation.

## Linked Issue

Closes #13434

- [x] Screenshots are not applicable because this PR contains specifications only and does not change runtime UI.

## Testing

This is a documentation-only change, so no runtime tests were added.

- `./script/format --check`
- `git diff --cached --check`

- [ ] I have manually tested my changes locally with `./script/run` (not applicable to a spec-only change)

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/specs/gh-13434/PRODUCT.md
+++ b/specs/gh-13434/PRODUCT.md
@@ -1,0 +1,58 @@
+# Open repository-browser Markdown links as local files
+
+GitHub: [warpdotdev/warp#13434](https://github.com/warpdotdev/warp/issues/13434)
+
+Figma: none provided. This first pass adds an alternate link action without changing the rendered Markdown layout.
+
+## Summary
+
+Let users deliberately open the local checkout copy of a file referenced by a remote repository-browser link in rendered Markdown. A platform-specific alternate-click gesture resolves an exact repository-relative path against the active local checkout; normal link opening remains unchanged.
+
+## Goals / Non-goals
+
+Goals:
+
+- Make remote GitHub and GitLab file links useful while reviewing Markdown inside a local checkout.
+- Keep the action explicit and deterministic: resolve one exact local file or report that no local match exists.
+- Reuse the user's existing local-file opening behavior.
+- Avoid network access, repository mutation, background indexing, and branch or commit checkout.
+
+Non-goals:
+
+- No fuzzy filename search, link-text matching, multiple-match picker, or repository selector.
+- No checkout, fetch, branch switch, detached worktree, or attempt to make the local file match the URL's revision.
+- No extraction of line or column locations from URL fragments in this first pass.
+- No automatic hyperlinking of bare paths or code spans.
+- No change to Agent Mode rich-output links, terminal hyperlinks, shared-session links, or remote-only files in this first pass.
+- No new setting or default change for ordinary Markdown links.
+
+## Behavior
+
+1. In a rendered local Markdown file, plan, or notebook that has an active local repository context, the user can invoke **Open local repository file** on an `http` or `https` Markdown link by holding the platform's primary modifier plus Alt/Option while clicking it:
+   - macOS: Command+Option-click.
+   - Windows and Linux: Control+Alt-click.
+2. A normal click, Command-click on macOS, or Control-click on Windows/Linux keeps the existing link behavior. In particular, an ordinary remote link continues to open its web URL and an ordinary local-path link continues to use the existing local-file behavior.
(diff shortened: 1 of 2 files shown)


specs/gh-13434/PRODUCT.md
```diff
@@ -0,0 +1,58 @@
...
+- No checkout, fetch, branch switch, detached worktree, or attempt to make the local file match the URL's revision.
+- No extraction of line or column locations from URL fragments in this first pass.
+- No automatic hyperlinking of bare paths or code spans.
+- No change to Agent Mode rich-output links, terminal hyperlinks, shared-session links, or remote-only files in this first pass.
+- No new setting or default change for ordinary Markdown links.
+
+## Behavior
+
+1. In a rendered local Markdown file, plan, or notebook that has an active local repository context, the user can invoke **Open local repository file** on an `http` or `https` Markdown link by holding the platform's primary modifier plus Alt/Option while clicking it:
+   - macOS: Command+Option-click.
```
seemeroland: For discoverability, I think we would want to add "Open local file" to the tooltip when you click on a url link, following the same pattern that file links use. I think it might even be better to only add to this tooltip and not support the modifier, since this is probably a pretty uncommon use case and we might want to reserve the cmd-option-click modifier for something else in the future.

Image

Image

This should only come up when it is a possibly valid local file, using the pure text file path candidate parsing logic you describe

seemeroland (commented):
Sorry for the delay, I was on vacation the past couple weeks!

PR #14410, opened 2026-07-28
Reconstruct URLs wrapped across rich output lines

### LINKED ISSUE #9385: Cmd+click on multi-line wrapped URL only opens partial URL (first line portion)
## Description

When a long URL wraps across two or more lines in the terminal, Cmd+clicking on it only opens the first-line portion of the URL in the browser. The second (and any subsequent) line segments are ignored, resulting in a broken/incomplete URL being opened.

## Steps to reproduce

1. Make the terminal window narrow enough that a long URL wraps to a second line
2. Output or paste a long URL (e.g. one with many query parameters)
3. Cmd+click on the URL

## Expected behavior

The full URL (spanning both lines) opens in the browser.

## Actual behavior

Only the portion of the URL on the first visual line is opened — the URL is split at the line wrap boundary.

## Environment

- OS: macOS
- Shell: zsh

katspaugh:
## Description

Fixes Cmd+click opening only the first-line portion of a URL that arrives split across several lines of rich CLI-agent / AI output.

CLI agents hard-wrap their own TUI output, so a long URL reaches Warp as several `FormattedTextLine`s. `collect_output_data_for_link_detection` pushes each formatted line as its own `(String, TextLocation)`, and `detect_all_links` then runs `detect_urls` independently per line. Only the fragment that independently parses as a URL is detected, so `AIBlock::open_link` opens that truncated target.

This PR rebuilds the logical URL before detection (Option B from the tech spec in #9989):

- Adjacent formatted lines are grouped into runs that share a `section_index` and have consecutive `line_index`es, so joining can never cross output sections, code blocks, tables, images, or action rows.
- Within a run, a URL that touches the end of a line is grown onto the next line, and the reconstructed range is mapped back to **per-line** clickable ranges. Every participating segment stores the *full* URL, so Cmd+clicking any visual line opens the whole link while hover/registration stays per rendered line exactly as before.
- Markdown hyperlinks keep precedence, since they are still applied last.

Joining is deliberately conservative and prefers false negatives over a wrong target. A continuation is rejected when it:

- starts a new layout row — leading whitespace, `-`/`*`/`+`/`#`/ordered-list markers, or table/box-drawing glyphs;
- opens its own URL (two stacked links stay separate);
- only contributes a bare word to a URL that was already complete (`https://example.com` followed by `This is a sentence.` must not become `https://example.comThis`);
- would exceed 1000 characters, mirroring `URL_SCAN_CHARACTER_MAX_COUNT` in the terminal-grid path, or span more than 32 lines.

Ranges are char-based throughout, matching the existing `Range<usize>` semantics used by `FormattedTextElement`, and the trailing newline that `raw_text()` appends is stripped before offsets are computed.

## Linked Issue

Closes #9385 (tech spec: #9989)

> Linked with `Closes` so Oz can pick this up — but please re-check before merge. As explained in the open question below, this fixes the rich-output path that the tech spec scopes, while the reporter's `claude` CLI session most likely runs through the terminal-grid path instead. If that's confirmed, #9385 should stay open (or get a follow-up) rather than auto-close on merge.

## Testing

Nine regression tests added in `app/src/util/link_detection_tests.rs`, driving the real `detect_all_links` entry point:

| Test | Covers |
| --- | --- |
| `test_reconstructs_url_hard_wrapped_across_two_lines` | the reported repro; both lines resolve to the full URL |
| `test_reconstructs_url_hard_wrapped_across_three_lines` | three-line reconstruction |
| `test_single_line_url_detection_is_unchanged` | no behaviour change for unwrapped URLs |
| `test_url_followed_by_prose_is_not_joined` | prose is not swallowed |
| `test_adjacent_independent_urls_stay_separate` | stacked links stay separate |
| `test_layout_rows_are_not_joined_onto_a_url` | bullets, ordered lists, `\|` and `│` table rows, indentation, blank lines |
| `test_reconstructed_ranges_are_char_based_not_byte_based` | multi-byte text before the URL |
| `test_urls_are_not_reconstructed_across_output_sections` | no stitching across sections |
| `test_markdown_hyperlinks_take_precedence_over_reconstructed_urls` | markdown hyperlink precedence |

```
cargo test -p warp --lib -- util::link_detection
test result: ok. 23 passed; 0 failed
```

`./script/format --check` and `./script/check_no_inline_test_modules` pass.

**I was not able to do the required manual verification**, and I would rather say so than check the box. This was developed on Linux and I have no macOS build to reproduce the original Claude Code session against. Everything above is automated-test evidence only. Happy for a maintainer to take it for a spin, or I can pair with someone who can run the Mac build.

## Question for maintainers: should the terminal-grid path be covered too?

The tech spec (§6) says explicitly not to change terminal-grid behaviour, so this PR does not. While verifying, though, I probed `GridHandler::url_at_point` directly, and it reproduces the reported symptom whenever a TUI emits **real newlines** instead of letting the terminal soft-wrap:

```
HARD (\r\n, TUI wrapped it itself)   row0 => Link(row 0, col 0..59)   <- truncated
                                     row1 => None
SOFT (\n, terminal reflowed)         row0 => Link(row 0 -> row 1)     <- already correct
```

That matters because the reporter's case is the `claude` CLI running as an ordinary subprocess. Its output goes through the terminal grid, not through `AIAgentOutput` — so I don't think this PR alone fixes the original reproduction, even though it does fix the class of bug the spec describes for rich output. It's possible the auto-triage attributed it to the rich path more confidently than the evidence supports.

Extending the fix into the grid would mean joining across hard newlines, which is currently deliberate (`test_find_url_line_breaks` asserts it) and which @zachlloyd flagged as risky for arbitrary TUI layouts such as tables.

One signal exists in the grid that the rich-text path doesn't have: **a TUI hard-wrap always breaks at the terminal width**, so a join could additionally require the preceding row to be full-width. Rows that merely happen to end in a URL would be left alone, which should keep the false-positive surface close to zero while still fixing wrapped OAuth/login URLs.

Would you like that in this PR, as a follow-up, or not at all? Happy to implement it either way — I just didn't want to silently contradict the spec.

## Agent Mode

CHANGELOG-BUG-FIX: Cmd+clicking a URL that wraps across multiple lines of AI and CLI-agent output now opens the whole URL instead of just the first line.

### DIFF
--- a/app/src/util/link_detection.rs
+++ b/app/src/util/link_detection.rs
@@ -246,6 +246,252 @@ fn detect_urls(text: &str) -> Vec<Range<usize>> {
     url_ranges
 }
 
+/// Maximum number of characters in a URL reconstructed across wrapped lines. Mirrors
+/// `URL_SCAN_CHARACTER_MAX_COUNT` in the terminal-grid path: offering an incomplete link
+/// is worse UX than offering none.
+const MAX_RECONSTRUCTED_URL_CHARS: usize = 1000;
+/// Maximum number of formatted lines that a single reconstructed URL may span.
+const MAX_WRAPPED_URL_LINES: usize = 32;
+
+/// A URL rebuilt from fragments that a CLI agent hard-wrapped across formatted lines,
+/// along with the per-line char ranges that display it.
+struct WrappedUrl {
+    url: String,
+    /// `(index into the run of lines, char range within that line)`.
+    segments: Vec<(usize, Range<usize>)>,
+}
+
+/// Returns a formatted line without the trailing newline that [`FormattedTextLine::raw_text`]
+/// appends, so char ranges line up with what the renderer registers as clickable.
+fn line_body(text: &str) -> &str {
+    text.strip_suffix('\n').unwrap_or(text)
+}
+
+/// Returns true when `line` opens a new layout row rather than continuing a wrapped URL.
+///
+/// TUIs draw lists, tables and quotes with leading markers, and a URL can never resume
+/// after one. Markers that are also legal URL characters (`-`, `#`, `+`, ...) only count
+/// when whitespace follows, since `https://a.example/-x` must keep wrapping normally.
+fn looks_like_new_layout_row(line: &str) -> bool {
+    let mut chars = line.chars();
+    let Some(first) = chars.next() else {
+        return true;
+    };
+    if first.is_whitespace() {
+        return true;
+    }
+
+    const LAYOUT_DELIMITERS: &[char] = &[
+        '|', '│', '┃', '├', '└', '┌', '┐', '┘', '┤', '┬', '┴', '┼', '─', '║', '╠', '╚', '═', '╦',
+        '╩', '╬', '•', '‣', '⁃',
+    ];
+    if LAYOUT_DELIMITERS.contains(&first) {
+        return true;
+    }
+
+    let rest = chars.as_str();
+    if matches!(first, '-' | '*' | '+' | '>' | '$' | '%' | '#')
+        && rest.chars().next().is_none_or(char::is_whitespace)
+    {
+        return true;
+    }
+    if first == '#'
+        && line
+            .trim_start_matches('#')
+            .starts_with(char::is_whitespace)
+    {
+        return true;
+    }
+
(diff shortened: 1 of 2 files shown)


katspaugh:
@cla-bot check

PR #14415, opened 2026-07-28
Support dragging tab groups between windows

### LINKED ISSUE #14152: Support dragging/transferring tab groups between windows
> 🤖 Filed automatically by [@ozdotdev](https://x.com/ozdotdev) on behalf of [@vikvang1](https://x.com/vikvang1).
> Source: https://x.com/vikvang1/status/2080030589026246993

### Summary
Warp supports dragging/transferring an individual tab from one window to another (detach into a new window, or drop onto another window's tab bar). This request asks to extend that same drag/transfer behavior to tab groups, so an entire named tab group can be moved to another window as a single unit.

### Current behavior
- Individual tabs can be moved between windows via drag-and-drop (see https:​//docs.warp.dev/terminal/windows/tabs).
- Tab groups can be created, named, colored, collapsed, and pinned, but a whole group cannot be dragged/transferred to another window.

### Requested behavior
- Drag a tab group (via its header) out of the tab bar to detach it into a new window, and onto another window's tab bar to move it there.
- Preserve the group's name/color and membership, and each member tab's running session, panes, and state, consistent with single-tab drag behavior.
- Support this from both the horizontal tab bar and the vertical tabs panel.

### Additional context
Reported via X by @​vikvang1: https:​//x.com/vikvang1/status/2080030589026246993
Related: #466 (Rearrange / drag tabs/panes into windows and vice-versa).

2solarmax:
## Description

Drag a whole tab group by its header out of the tab bar to detach it into a new window, or onto another window's tab bar to move it there. The group keeps its name, colour, collapsed and pinned state and its membership, and every member keeps its running session and panes. Works from both the horizontal tab bar and the vertical tabs panel.

**Stacked on #14347.** That PR fixes source cleanup for single-tab drags, which the group path builds on; both branches touch the same cleanup region and independently added `tab_index_for_pane_group_id`. The first four commits here are #14347. Merge that first and this rebases to just the group work.

### How it works

`DragSource` gains a `GroupWindow` variant carrying the group id, the member pane-group identities and an optional preview window. The preview is optional because a group can span every tab in its window: there is then nothing left behind, so the source window is itself the floating preview, exactly like a single-tab drag. `source_was_single_tab` became `source_is_own_preview`, which is what every call site actually meant.

Cleanup keys off member identity, never an index range. The source tab list can change while a drag is in flight, and a stale range removes tabs that are still in bounds. Removal goes through `remove_tab_without_undo` per member rather than draining the vector, so each removal still clears the detail sidecar, prunes the MRU order, and prunes the group with its last member.

Ordering is load-bearing at detach and at handoff: every member's view tree moves before any tab list is touched, in one update with no repaint in between, so no render observes the group split across two windows. Members that are still in flight are filtered out of `Workspace::snapshot` by identity — that filter was scalar, and snapshotting a tab whose view is being updated in another window panics rather than returning `None`.

One resolver, `resolve_group_drop_index`, owns the landing index for both the ghost slot and the commit, so they cannot diverge. It applies all three clamps the per-tab insert applies, including `clamp_past_group` — without that, a group dropped inside another group's run would split it.

### Relationship to #14164

This supersedes the open draft #14164 on the same issue. Some of what is here came from working out why that approach did not hold: it removes source tabs positionally by `{index, count}`, has no put-back path, and leaves `reverse_handoff` and `on_drag_while_inserted` at arity 1. Its group-aware preview rect idea is right and is kept in spirit.

## Linked Issue

Fixes #14152 (`ready-to-implement`).

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`

## Testing

Four new integration tests drive the real app with synthetic mouse events, alongside the five existing single-tab drag tests:

```
test ui_tests::test_detach_tab_group_to_new_window_with_drag        ... ok
test ui_tests::test_detach_tab_group_to_new_window_in_vertical_tabs ... ok
test ui_tests::test_attach_tab_group_to_other_window                ... ok
test ui_tests::test_whole_window_tab_group_attaches_to_other_window ... ok
test ui_tests::test_tab_group_drag_back_to_source_cancels           ... ok
test ui_tests::test_detach_tab_to_new_window_with_drag              ... ok
test ui_tests::test_attach_tab_to_other_window_and_continue_drag    ... ok
test ui_tests::test_single_tab_handoff_continues_drag               ... ok
test ui_tests::test_multi_tab_drag_back_to_source_and_out_again     ... ok
test ui_tests::test_reorder_tabs_with_drag                          ... ok
test result: ok. 10 passed; 0 failed
```

```
cargo test -p integration --features drag_tabs_to_windows --test integration -- drag tab_group vertical_tabs whole_window
```

**These are invisible to default CI.** They are gated on `drag_tabs_feature_enabled`, and `ci.yml` runs the workspace without `--features`, so the whole group suite silently skips there. Flagging that rather than letting a green tick imply coverage. The drag-model state is therefore also unit-tested **ungated** in `cross_window_tab_drag_tests.rs`, and the model-level behaviour — drop-index resolution across the pinned boundary and into an existing group, contiguity after insert, identity cleanup after a mid-drag shift, non-contiguous refusal — in `view_tests.rs`.

Also run:

- `./script/format --check` — clean
- `cargo clippy -p warp -p integration --all-targets --tests --features integration/drag_tabs_to_windows -- -D warnings` — clean
- `cargo test -p warp --lib` — 13 failures, the same set that fails on pristine `master` on this machine (OpenAI key, notebooks fs, telemetry, terminal, path resolution). None in drag or group code.

The whole-window case has its own test because every other group test leaves one tab ungrouped, so that path was structurally unreachable in the suite — and getting it wrong destroys the members that are not transferred.

Driven against a local build with synthetic mouse events, which is the same gesture path the integration tests exercise; stills below. `test_video_recording` is `#[ignore = "Manual test: requires real display for frame capture"]` and still times out per frame when run un-ignored on a real display, so the in-harness capture path was not available for a recording. Build with `./script/run --features drag_tabs_to_windows`; the flag is not in the default set, so a plain `./script/run` produces a binary where this path is unreachable. Grab a group by its **header** — grabbing a member drags just that tab, which is what `with_defer_to_handled_child_mouse_down` is for.

### Screenshots / Videos

Before / mid-drag / after, on a local `--features drag_tabs_to_windows` build, posted in a comment below: https://github.com/warpdotdev/warp/pull/14415#issuecomment-5112989292

The mid-drag frame is the one that matters — the group sits in the floating preview with both members while its slot in the source panel has collapsed, so it is moving as a unit rather than a member at a time.

## Agent Mode

CHANGELOG-NONE


2solarmax:
Manual testing, on a local build: `./script/run --features drag_tabs_to_windows`.

Dragging a two-tab group out of a window by its header, in the vertical tabs panel. The window starts with one ungrouped tab plus a group of two, so the source survives the detach and you can see what it keeps.

**Before** — one ungrouped tab, then "New Group" with 2 tabs:

before the drag

**Mid-drag** — the group is in the floating preview with *both* members, and its slot in the source panel has collapsed. This is the frame that shows it moving as a unit rather than a member at a time:

mid drag

**After the drop** — two windows. The source keeps only its ungrouped tab; the new window holds the group, still named, still 2 tabs:

after the drop

Confirmed via the accessibility API that this is genuinely two windows rather than one repainted, at `(4301, 103)` and `(4859, 317)`.

A note on why these are stills and not a recording: `test_video_recording` is `#[ignore = "Manual test: requires real display for frame capture"]`, and running it un-ignored on a machine with a real display still logs `VideoRecorder: frame capture timed out after step` for every frame, so the in-harness capture path was not available. These were taken with `screencapture` against the running app while the drag was driven with synthetic mouse events, which is the same gesture path the integration tests exercise. Happy to add a screen recording if you would still like one.

The other three gestures are covered by the integration tests listed in the PR description rather than shown here: drop onto another window's tab bar, put-back onto the source's own tab bar, and a group spanning every tab in its window.

2solarmax:
Good catch, fixed in 4e28d4c5.

The whole-window branch unsubscribed every member and moved their view trees before looking up the target workspace. That window is its own preview, so past that point there is nothing to fall back to and a bail leaves the sessions without a workspace. The lookup now happens first, so the reset path still has an intact source to return to — matching `execute_handoff_multi_tab_to_other`, which already resolved its target up front.

All 10 drag integration tests still pass, clippy and format clean.

/oz-review

2solarmax:
This latest review ran on `4e28d4c5`, which is the commit that fixed it, but it repeats the same finding — so I think it's a stale read rather than a live issue. The order on the current head is:

- `view.rs:28494-28502` — resolve the target workspace, bail to `reset_to_floating` if it's gone
- `view.rs:28507-28510` — only then unsubscribe each member
- `view.rs:28514` — and only then move the view trees

So the bail happens while the source is still fully intact and still owns its sessions. That was exactly the concern and it was the point of the change; nothing is unsubscribed or transferred before the lookup any more.

I'd rather not spend the last `/oz-review` re-running it on an unchanged tree, so flagging for a human instead.

@oss-maintainers — could someone take a look? The PR is otherwise clean: 10 drag integration tests passing on a local build (4 new for the group gestures plus the 5 existing single-tab ones, all green together), `./script/format --check` and `cargo clippy -p warp -p integration --all-targets --tests --features integration/drag_tabs_to_windows -- -D warnings` both clean, and manual-testing stills a few comments up.

One thing worth a maintainer's eye rather than a bot's: this is stacked on #14347, which fixes the same source-cleanup path for single-tab drags. That one is still waiting on a readiness label for #14346. If you'd rather see them as one PR, or want #14347 landed first, happy either way.

2solarmax:
Rebased onto current `master` (was 108 behind, and had gone conflicting). Back to mergeable at `85c3f864`.

The conflict was additive only — upstream added two `test_open_new_window_for_team_*` tests next to my `mock_workspace_at` helper in `view_tests.rs`; both sides kept. All 10 drag integration tests still pass after the merge, and format/clippy are clean.

Still looking for a human pass on this whenever someone has a moment — the last Oz review repeated a finding that the reviewed commit had already fixed (details a couple of comments up), so I've left the third `/oz-review` unspent rather than re-run it on an unchanged tree.

2solarmax:
Fixed in ebdf3315 — the doc block for `handle_drop_result` had drifted onto `grouped_tab_count` along with a truncated fragment of the `tab_indices_for_pane_group_ids` doc. Each doc now sits on its own function. Suite and clippy still green.

2solarmax:
Fixed in 7b63cccc. Verified the finding and scoped it precisely: the ordered-window route and the source-window route of `cross_window_attach_target` both resolved their drop index through the single-tab resolver, so a group drag's ghost slot was computed from the raw cursor slot. The commit path already re-clamps via `resolve_group_drop_index` (cross_window_tab_drag.rs execute_handoff_multi_tab_to_other), so the landing index was never wrong — the divergence was the ghost preview pointing at a position the drop never takes (including inside the pinned prefix).

Fix is the class, not the two instances: `compute_insertion_index_for_window` now takes `dragged_group_pinned` and dispatches to the group resolver itself, so all three targeting routes funnel through one place and a new call site can't silently get the single-tab behavior. Added `test_attach_tab_group_does_not_split_target_group` (group dropped into a target that already holds a group, with count preconditions so it can't pass vacuously). Suite 11/11, clippy clean.

2solarmax:
Still here and ready for review — rebased and green as of last week (`7b63cccc`, 11/11 drag integration tests, clippy clean). This PR is blocked on maintainer review rather than on me, so per the contribution guidelines could a maintainer apply `no-autoclose`? Happy to make any change a reviewer wants.

PR #14480, opened 2026-07-29
QUALITY-928: Orchestration unified stack — pane path + transcript (M2)

cephalonaut:
## Description

Unifies the orchestration child pane path and moves parent/child restore onto task data (QUALITY-928).

## What

- Unified child pane materialization: the owner/viewer code split is collapsed into a single dispatch path. `TaskOwnership`, `TaskScope`, and `ChildPaneOrigin` are removed along with the branching they supported.
- `OrchestrationViewerModel` now uses `AgentConversationsModel` as its fetch authority, routing child discovery through `pending_task_ids_for_discovery`.
- Remote child and parent cloud agent conversations are ephemeral cloud state — not persisted to local DB. On restore they are re-seeded from the server via `GET /agent/runs?ancestor_run_id=`, the same query the OVM uses live.
- The pill bar re-renders on `ConversationServerTokenAssigned`, ensuring correct notification after the async restore seed completes.
- Local-parent seed trigger: when a local agent parent restores with an empty child index, `restore_missing_child_agent_panes_for_parent` triggers the server seed to discover remote children.
- `finish_seed_child_conversations_from_task` does not re-trigger the seed on completion (`trigger_seed_if_empty=false`), preventing an infinite loop.

## Why

The parallel owner/viewer hydration paths had accumulated code duplication and subtle behavioral differences. Consolidating around task data as the authoritative source simplifies restore and makes behavior consistent across native and web clients.

## Linked Issue

QUALITY-928

## Testing

- Manual: pill bar appears after restart for a completed cloud agent run.
- Manual: local parent with remote children restores correctly after restart.
- Unit coverage for child pane materialization and task-driven restore seeding.
- `./script/format` and `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings` clean.

## Agent Mode


cephalonaut:
**Demo:** https://www.loom.com/share/5feec7c256044f03a5d63d2c46fe4a75

Demonstrates: pill bar restore after cloud agent restart, local parent with remote child restore, and flag-off behavior verified against pre-M1 baseline.

danielpeng2:
Heads-up on a merge-order interaction (posted by Daniel's Warp Agent on his behalf):

#14762 (multi-level orchestration: depth-capable client core + drill-down pill bar) is landing ahead of this PR, and it rewrites the pill-bar render region this branch touches — the visible row now renders **direct children only**, with deeper levels reached by drilling into a pill (replacing the flattened `descendant_conversations_in_pill_order` rendering).

**What we did to keep your rebase trivial:** we absorbed this branch's `ConversationServerTokenAssigned` re-render arm into #14762's pill-bar subscription (commit `95e96ed106`, with a first event-driven re-render test). It was independently correct for the multi-level work too — run-id linkage for remote children arrives asynchronously — so credit for spotting it.

**How to resolve when you rebase `matthew/orch-unified-m2`:**
- `app/src/ai/blocklist/agent_view/orchestration_pill_bar.rs`: take the master side wholesale. Both of this branch's edits are subsumed — the `ConversationServerTokenAssigned` arm already exists on master, and the `descendant_conversations_in_pill_order` statement split targets a line that no longer exists.
- `orchestration_viewer_model_tests.rs`: the conflict you'll see there is against newer master generally, not the multi-level stack.
- `descendant_conversations_in_pill_order` itself still exists (used by topology internals + TUI export), so nothing else in this branch breaks.

**One re-verification note:** the manual-testing assumptions here ("pill bar appears after restart" etc.) were made against the flattened-descendants rendering. Post-rebase, restored children appear under drill-down semantics — worth a quick re-check that the restore-seeded pills show up at the level you expect.

The deeper interaction points (unified-stack flag-ON pane path vs the `RemoteSubtreeModel` in #14763, `is_remote` at-creation contract, LiveAttach handling per flag path) are with the *second* PR in Daniel's stack — we'll sync on those separately when you're back.

vorporeal:
asked an agent to read through this so i could get a better understanding, and along the way, asked it to do a review.  below is what it found in its review:

---

## Overall judgment

The architectural direction is good, but I would not approve M2 in its current state.

The central weakness is **reconciliation after partial or out-of-order state**. M2 is careful to make child creation idempotent, but several paths do not converge after:

- Only some children are discovered.
- Lifecycle events arrive before metadata.
- A network request fails.
- The same run is viewed through multiple parent conversations.

I found one compile blocker, three significant correctness problems, and several recovery or scale gaps.

## Blocking and high-severity findings

### 1. The TUI test target does not compile

`start_new_child_conversation` now requires an `is_remote` argument, but several test call sites still use the old signature:

- `crates/warp_tui/src/orchestration_model_tests.rs:824`
- `crates/warp_tui/src/orchestration_model_tests.rs:1065`
- `crates/warp_tui/src/orchestration_model_tests.rs:1082`
- `crates/warp_tui/src/orchestration_model_tests.rs:1094`
- `crates/warp_tui/src/orchestration_model_tests.rs:1189`

The new signature is at `app/src/ai/blocklist/history_model.rs:553`.

This is a straightforward merge blocker.

### 2. Partial remote-child discovery can suppress the full restore seed

`restore_missing_child_agent_panes_for_parent` requests the server snapshot only when it has discovered **no** remote children:

`app/src/pane_group/child_agent/restoration.rs:66-85`

That does not prove that it has discovered **all** remote children.

A concrete migration scenario is:

1. Remote child A was created while the unified flag was off, so A was persisted.
2. The flag is enabled.
3. Remote child B is created and is not persisted.
4. Warp restarts.
5. A is restored from SQLite.
6. Because one remote child exists, M2 skips task seeding.
7. B remains missing.

The same problem can occur if a live event rediscovers one child before the restore path runs.

The server seed is idempotent, so I do not see a reason to use “some remote child exists” as evidence that it is unnecessary. I would expect local-parent restore to seed whenever the parent has a task identity, independent of what was already restored locally.

### 3. A terminal lifecycle event can be lost while placeholder metadata is loading

The event streamer starts asynchronous placeholder creation and then continues processing lifecycle events:

`app/src/ai/blocklist/orchestration_event_streamer.rs:566-621`

If a terminal lifecycle event arrives after the tracker knows the child but before the history model contains its conversation:

- `apply_lifecycle` cannot update the conversation status.
- It emits `ChildStatusChanged`.
- The owner controller intentionally ignores that event.
- The later placeholder callback creates the conversation but does not apply the terminal status from either the tracker or task.

Relevant code:

- `app/src/ai/blocklist/orchestration_child_tracker.rs:220-253`
- `app/src/ai/blocklist/controller.rs:608-620`
- `app/src/ai/blocklist/orchestration_event_streamer.rs:735-791`

A child can therefore finish or fail while its pill remains in its initial state. Parent aggregate status can also remain incorrect until another event happens to repair it.

Placeholder completion should initialize status from the latest tracker state or fetched task state.

### 4. Run-scoped child identity conflicts with simultaneous viewer parents

`ensure_remote_child_conversation` globally reuses a child conversation by run ID. If another local parent conversation views the same orchestrator, it reparents the existing child:

`app/src/ai/blocklist/history_model.rs:600-623`

There are two problems.

First, `set_parent_for_conversation` adds the child to the new parent’s index but does not remove it from the old parent’s index:

- `app/src/ai/blocklist/history_model.rs:532-541`
- `app/src/ai/blocklist/history_model.rs:652-663`

The child is then listed under both parents, while its own `parent_conversation_id` points only to the newest parent.

Second, fixing the stale index alone does not resolve the larger ownership question. Multiple viewer panes for one orchestrator are explicitly supported at `app/src/ai/blocklist/orchestration_event_streamer.rs:1188-1212`. A single global child conversation cannot have two viewer-local parents at the same time.

The author should define one of these models explicitly:

- Child conversations and panes are scoped to each viewer parent.
- Child identity is shared globally, and parent-to-child presentation is represented separately from the child’s single `parent_conversation_id`.

The current implementation mixes the two models.

## Medium-severity findings

### 5. Transcript hydration has no recovery after a transient failure

M2 removes the pending hydration entry before fetching the transcript:

`app/src/pane_group/child_agent/hydration.rs:193-213`

If fetching or merging fails, the callback only logs and returns:

`app/src/pane_group/child_agent/hydration.rs:402-439`

The loading pane remains registered as the canonical child pane, and the viewer records that materialization was already requested. Later updates therefore do not request it again:

- `app/src/pane_group/child_agent/hydration.rs:90-105`
- `app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs:395-404`
- `app/src/terminal/shared_session/viewer/orchestration_viewer_model.rs:623-625`

A temporary network failure can leave the child on a permanent loading pane. Recoverable failures should requeue hydration or clear the requested state.

### 6. Failed restore fetches depend on unrelated events for retry

When ancestor discovery fails, the pending seed remains, but no retry is scheduled:

`app/src/pane_group/child_agent/restoration.rs:178-187`

Retry happens only after `TasksUpdated` or `ConversationsLoaded`:

- `app/src/pane_group/child_agent/restoration.rs:264-284`
- `app/src/pane_group/mod.rs:3317-3338`

If the request fails after initial loading and no tasks subsequently change, reconnecting to the network does not restore the children.

This needs either:

- A documented guarantee that task updates occur periodically.
- A reconnect trigger.
- A bounded retry with backoff.

### 7. Restore silently truncates after 100 direct children

Restore performs one request with `RESTORE_CHILD_SEED_FETCH_LIMIT = 100`:

`app/src/pane_group/child_agent/restoration.rs:30`

The server API accepts a cursor, but `list_ambient_agent_tasks` returns only the run list and does not expose pagination to this caller. A parent can exceed 100 direct children through repeated batches, even if one batch is limited.

This is contract-dependent. If the product guarantees at most 100 total children per parent, that guarantee should be explicit and tested. Otherwise, restore needs pagination.

## Test assessment

The branch adds useful tests for:

- Materialization decisions.
- Event ordering in the tracker.
- Viewer discovery.
- Run-ID idempotency.
- Pane ownership boundaries.

The important missing tests match the findings above:

- Flag-off persisted child plus flag-on unpersisted sibling after restart.
- Partial discovery before local-parent restore.
- Terminal lifecycle before asynchronous placeholder completion.
- Reparenting between two distinct parent conversation IDs.
- Two simultaneous viewers of the same orchestrator.
- Transcript fetch failure followed by recovery.
- Ancestor fetch failure followed by network recovery.
- More than 100 direct children.

cephalonaut:
Thanks for the thorough review. Addressed in 26665f6d1:

**Fixed:**
- **TUI compile blocker**: Added missing `is_remote` arg to the 5 `start_new_child_conversation` call sites in `orchestration_model_tests.rs`.
- **Partial restore seed (finding 2)**: Removed the `has_discovered_remote_children` guard. The seed now always fires for local-parent restore when not already pending — the `pending_parent_child_seeds` map deduplicates. This covers the flag-toggle migration case you described.
- **Stale parent index (finding 4)**: `set_parent_for_conversation` now removes the child from its old parent's `children_by_parent` entry before re-indexing. No-op on initial creation.
- **Lifecycle backfill during placeholder fetch (finding 3)**: `TrackedChild` now records `last_lifecycle`. After `finish_remote_child_placeholder` creates the conversation, it reads back any already-received lifecycle status and applies it, so the pill badge doesn't stay at the placeholder default when a terminal event arrived during the fetch window.
- **Transcript hydration recovery (finding 5)**: On fetch-empty, the transcript path now re-inserts into `pending_child_hydrations` and evicts+refetches the task, so a transient unavailability retries on the next task refresh rather than leaving the pane stuck loading permanently.
- **Join-failure filter (Oz comment)**: Widened `recover_viewer_child_join_failure`'s filter to also accept `is_remote_child()` so unified-stack remote children can recover from a failed live-session attach.

**Deferred:**
- **Finding 6 (restore retry on reconnect)**: The existing `pending_parent_child_seeds` mechanism retries on `TasksUpdated`, which fires on the normal task polling cycle. A proper reconnect trigger would require hooking into network connectivity events — tracking separately.
- **Finding 7 (100-child pagination)**: The server caps at 100 and the comment already documents this. Pagination needs both client and server changes — tracking separately.

specs/QUALITY-928/TECH.md
advait-m, line 1: is this meant to be deleted? something got messed up in rebase perhaps?

app/src/pane_group/child_agent/hydration.rs
```diff
@@ -68,17 +79,654 @@ pub(in crate::pane_group) fn decide_remote_child_hydration_action(
...
+                    log::warn!(
+                        "child transcript upgrade fetch-empty \
+                         child_conversation_id={child_id:?}"
+                    );
+                    // Re-queue for retry when task data refreshes; the
+                    // transcript may not be written yet or the fetch was
+                    // transient.
+                    group.pending_child_hydrations.insert(task_id, child_id);
+                    AgentConversationsModel::handle(ctx).update(ctx, |model, ctx| {
+                        model.evict_and_refetch_task(&task_id, ctx);
```
advait-m: was reviewing with agent, and this concern felt valid

hmm, i think this can poll indefinitely. `evict_and_refetch_task` clears `task_fetch_state` (in-flight guard + backoff) → refetch fires → `TasksUpdated` → `process_pending_child_hydrations` re-drives → still terminal w/ a `conversation_id` → `LoadTranscript`, same token → back here on `None`. paced by RTT rather than spinning, but no attempt cap, and `load_conversation_from_server` has no backoff of its own, so it runs as long as the pane lives.

main case is an Observer — ancestor list gives us the child's run row fine, but `get_ai_conversation` on its token can 403, so the task keeps resolving while the transcript keeps returning `None`. unknown-harness arm has the same shape for an out-of-date client.

can we bound it? retry count on the pending entry, or drop the evict and let the normal `TasksUpdated` cadence re-drive so the backoff in `get_or_async_fetch_task_data` still applies

advait-m (approved):
FYI: I pushed up a rustfmt + clippy change

Nice consolidation! Left some smaller comments but makes sense to me at a high level + demos!

specs/QUALITY-928/TECH.md
cephalonaut: Intentional — the spec was moved to an untracked location earlier in development. M1 put TECH.md on master via the squash merge, and M2 removes it. Nothing was lost in the rebase.

app/src/pane_group/child_agent/hydration.rs
cephalonaut: Good catch. Fixed in c8dbb767d — dropped the `evict_and_refetch_task` call and left only the `pending_child_hydrations.insert`. Without the evict, the task stays cached so retries are driven by the normal `TasksUpdated` cadence rather than firing on every round-trip. The 403 Observer case still re-queues, but at polling pace rather than immediately.

specs/QUALITY-928/TECH.md
cephalonaut: Actually — restored in d401ab71a with two corrections. The spec is worth keeping as an architectural reference since it captures design decisions (Primary vs Observer semantics, idempotency contracts, seed guard rationale) that aren't obvious from the code. The earlier removal was a policy call but the spec stays accurate and useful alongside M2.

PR #14499, opened 2026-07-29
Revert unintended TUI MCP catalog expansion

kevinyang372:
## Description
Selectively removes the TUI MCP catalog/discovery expansion that was accidentally amended into and merged with #14437 before it received a separate review.

- Reverses only the delta between the original reviewed head (`462e095d0`) and the accidentally merged head (`6a5432f18`).
- Preserves the original `/mcp` search, streamlined row controls, stable selection, and credential logout behavior from #14437.
- Restores the intended review boundary for the stacked child PR #14500, which reapplies this exact delta for independent review.

## Linked Issue
N/A

## Testing
- [x] Verified the selective reverse patch applies cleanly to current `master`.
- [x] Verified reapplying the same patch on child PR #14500 produces a tree identical to `origin/master`.
- [x] `./script/format --check`
- [x] `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`
- [x] `cargo clippy -p warp --all-targets --tests -- -D warnings`
- [x] `cargo clippy -p warp_completer --all-targets --tests -- -D warnings`
- [x] `git diff --check`
- [ ] I have manually tested my changes locally with `./script/run`

Manual and full workspace tests were not rerun because this PR is an exact selective revert that reconstructs the prior review boundary.

### Screenshots / Videos
N/A — this PR only removes the accidentally merged delta.

## Agent Mode
- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode
- [Agent conversation](https://staging.warp.dev/conversation/1592c723-6dbc-49fa-a6c8-1b0cf9699f1d)

CHANGELOG-NONE

PR #14500, opened 2026-07-29
Expand TUI MCP catalog and discovery

kevinyang372:
## Description
Follows corrective PR #14499, which has merged. Expands the TUI `/mcp` experience so users can discover and explicitly enable MCP servers from the same sources represented by Warp's broader MCP catalog.

- Shows TUI-local installations, synced templates, gallery entries, Warp-global definitions, third-party global definitions, and current-repository definitions in one searchable list.
- Labels entries with source and scope, preserves same-named file definitions from distinct providers, and reports diagnostics for multiple config files.
- Represents uninstalled synced and gallery entries as `available`; catalog refresh never installs, starts, or initiates OAuth for them.
- Adds a TUI-native enable flow for template variables and final confirmation. Installation and startup happen only after confirmation, and cancellation leaves catalog/runtime state unchanged.
- Detects Claude, Codex, `.agents`, and current-repository MCP configs in place without copying them into Warp's global config.
- Keeps third-party and repository definitions offline by default in TUI while preserving existing Warp-global activation and GUI auto-start behavior.

## Linked Issue
N/A

## Testing
- [x] Live `/mcp` verification in a real terminal:
  - Confirmed synced entries render as `available` with `Enter to enable`.
  - Confirmed enabling opens a separate `Enable and start` confirmation screen.
  - Canceled the flow and verified counts remained `2 connected · 1 failed · 1 offline · 26 available`.
  - Confirmed a current-repository Claude MCP appears as offline and is not auto-started.
- [x] `cargo nextest run -p warp_tui` (823 passed)
- [x] Focused `warp` MCP tests (29 passed; 6016 skipped)
- [x] `cargo check -p warp_tui`
- [x] Verified this child reapplies the exact parent-revert delta and its tip is tree-identical to `origin/master`.
- [x] `git diff --check`

The full workspace test suite was not rerun for this stack reconstruction.

### Screenshots / Videos
N/A — the interaction was validated directly in the terminal with `./script/run-tui`.

## Agent Mode
- [Agent conversation](https://staging.warp.dev/conversation/1592c723-6dbc-49fa-a6c8-1b0cf9699f1d)

CHANGELOG-IMPROVEMENT: Expanded TUI `/mcp` discovery and added explicit opt-in installation for synced and gallery MCP servers.

### DIFF
--- a/app/src/persistence/mod.rs
+++ b/app/src/persistence/mod.rs
@@ -108,9 +108,9 @@ pub enum PersistedDataScope {
     /// The GUI app: everything, including window/tab/block session
     /// restoration and command history.
     Full,
-    /// The `warp-tui` front-end: command history, cloud objects, and
-    /// agent/conversation state, but no GUI session restoration, user profiles,
-    /// or pending object actions.
+    /// The `warp-tui` front-end: command history, cloud objects, user profiles,
+    /// and agent/conversation state, but no GUI session restoration or pending
+    /// object actions.
     TuiFrontend,
     /// The remote server daemon: only codebase index metadata.
     CodebaseIndicesOnly,
@@ -130,7 +130,12 @@ impl PersistedDataScope {
         )
     }
 
-    /// User profiles and pending object actions, which only the GUI consumes.
+    /// User profiles used to identify cloud-object creators in both interactive frontends.
+    fn user_profiles(self) -> bool {
+        self != PersistedDataScope::CodebaseIndicesOnly
+    }
+
+    /// Pending object actions, which only the GUI consumes.
     fn gui_only_data(self) -> bool {
         matches!(self, PersistedDataScope::Full)
     }
--- a/app/src/tui/mcp.rs
+++ b/app/src/tui/mcp.rs
@@ -1,18 +1,29 @@
+use std::collections::{HashMap, HashSet};
+use std::fmt;
 use std::path::PathBuf;
 
 use uuid::Uuid;
 use warpui::{Entity, ModelContext, SingletonEntity};
 
+use crate::ai::mcp::file_based_manager::FileBasedMCPServerScope;
+use crate::ai::mcp::gallery::MCPGalleryManagerEvent;
 use crate::ai::mcp::parsing::resolve_json;
 use crate::ai::mcp::templatable_manager::TemplatableMCPServerManagerEvent;
 use crate::ai::mcp::{
-    FileBasedMCPManager, FileMCPWatcher, MCPServer, MCPServerExt, MCPServerState,
-    TemplatableMCPServerManager, TransportType,
+    FileBasedMCPManager, FileMCPWatcher, MCPGalleryManager, MCPServer, MCPServerExt,
+    MCPServerState, TemplatableMCPServer, TemplatableMCPServerInstallation,
+    TemplatableMCPServerManager, TransportType, VariableType, VariableValue,
 };
-use crate::warp_managed_paths_watcher::active_mcp_config_file_path;
 
 #[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
-pub struct TuiMcpServerId(pub u64);
+pub enum TuiMcpServerId {
+    /// Stable content hash. File-based installation UUIDs are regenerated on
+    /// every config parse, so they cannot preserve selection across reloads.
+    FileBased(u64),
+    Installation(Uuid),
+    SyncedTemplate(Uuid),
+    Gallery(Uuid),
+}
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum TuiMcpTransport {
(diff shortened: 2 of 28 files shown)


kevinyang372:
* **#14500** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14500?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14499** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

app/src/persistence/mod.rs
```diff
@@ -130,7 +130,15 @@ impl PersistedDataScope {
         )
     }
 
-    /// User profiles and pending object actions, which only the GUI consumes.
+    /// User profiles used to identify cloud-object creators in both interactive frontends.
+    fn user_profiles(self) -> bool {
```
harryalbert, line 134: this is basically just saying don't use user profiles to identify cloud object creators for the remote server daemon, right? is there a simpler way we could put this now that it's for both the TUI and GUI?

app/src/tui/mcp.rs
```diff
@@ -1,27 +1,68 @@
...
+    FileBasedMCPManager, FileMCPWatcher, MCPGalleryManager, MCPServer, MCPServerExt,
+    MCPServerState, TemplatableMCPServer, TemplatableMCPServerInstallation,
+    TemplatableMCPServerManager, TransportType, VariableType, VariableValue,
 };
-use crate::warp_managed_paths_watcher::active_mcp_config_file_path;
 
 #[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
-pub struct TuiMcpServerId(pub u64);
+pub enum TuiMcpServerId {
+    FileBased(u64),
```
harryalbert, line 22: why does file based have a different id type than all the other variants?

app/src/tui/mcp.rs
```diff
@@ -44,21 +87,52 @@ pub struct TuiMcpServerSnapshot {
...
+    pub id: TuiMcpServerId,
+    pub name: String,
+    pub description: Option<String>,
+    pub source: TuiMcpServerSource,
+    pub instructions: Option<String>,
+    pub variables: Vec<TuiMcpTemplateVariable>,
+}
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct TuiMcpVariableValue {
```
harryalbert, line 156: could this struct be shared across the GUI and TUI?

crates/warp_tui/src/mcp_menu.rs
```diff
@@ -320,6 +329,43 @@ fn menu_rows(snapshot: &TuiMcpSnapshot, query: &str) -> Vec<TuiMcpMenuRow> {
     rows
 }
 
+pub(crate) fn source_label(source: &TuiMcpServerSource) -> String {
```
harryalbert: nit: should this be a fn on `TuiMcpServerSource`?

crates/warp_tui/src/zero_state.rs
```diff
@@ -389,51 +327,56 @@ fn render_mcp_section(mut column: TuiFlex, builder: &TuiUiBuilder, app: &AppCont
...
-                parts.push(format!("{failed} failed"));
-            }
-            if offline > 0 {
-                parts.push(format!("{offline} offline"));
-            }
-            (format!("{} · /mcp", parts.join(" · ")), false)
+    if snapshot.servers.is_empty() && snapshot.diagnostics.is_empty() {
+        return ("No servers available · run /mcp".to_owned(), false);
+    }
+    let mut running = 0;
```
harryalbert: could we bundle all of these values into a struct w/ a default method?

harryalbert (approved):
LGTM, just had a few small nits/org thoughts.

Also, a thought on the product flow, feel free to ignore if you've thought about this, but I find this flow kind of odd:
Image
Image

Like I don't know what the input is for, and it also seems like after I hit the mcp once we should just install instead of having an intermediary menu w/ just one option

app/src/persistence/mod.rs
kevinyang372: Simplified this to `self != PersistedDataScope::CodebaseIndicesOnly`, which directly expresses that both interactive frontends load user profiles while the remote server daemon does not.

app/src/tui/mcp.rs
kevinyang372: File-based installation UUIDs are regenerated on every config parse, so they are not stable enough to preserve selection and actions across reloads. The `u64` is a stable content hash; I added a doc comment explaining that distinction.

app/src/tui/mcp.rs
kevinyang372: I narrowed this request to `id`, `name`, and `variables`. I kept it frontend-specific because it carries `TuiMcpServerId`, which is interaction identity for this catalog; the GUI consumes the underlying template/gallery domain models directly, so sharing this type would couple it to CLI state without a shared domain need.

crates/warp_tui/src/mcp_menu.rs
kevinyang372: Moved this onto `TuiMcpServerSource::label()` and updated filtering/rendering to use it.

crates/warp_tui/src/zero_state.rs
kevinyang372: Bundled these into a `McpStatusCounts` struct with `Default` and a `record` method.

PR #14502, opened 2026-07-29
refactor: extract shared @ context menu core from the GUI view

harryalbert:
## Description

First of three PRs adding the `@` context menu to the headless TUI. This one is GUI-only groundwork and is intended as a strict behavioral no-op: it extracts the surface-neutral parts of the menu out of the 1711-line GUI view so a second front-end can drive the same menu, and migrates the GUI onto the extracted core in the same PR so a second implementation never exists.

Stack:
1. **this PR** — extract the shared core, migrate the GUI onto it
2. `harry/code-1910-explicit-working-dir` — give `FileSearchModel` and the file/code/skills data sources an explicit working directory instead of reading `app.windows().state().active_window`
3. `harry/code-1910-context-menu` — the TUI `@` menu

### What moved

New `app/src/search/ai_context_menu/core.rs`:

- `AIContextMenuCategory` and `NavigationState`, relocated from `view.rs`. `icon()` stays GUI-side as a separate inherent impl since it returns SVG asset paths.
- `AtContextMenuGates` — the four availability conditions previously stored as loose bools on `AIContextMenuState`. Front-ends supply the ones that are meaningful for them; the TUI has no ambient-agent session or shared-session viewer, so those stay `false`.
- `AtContextMenuCoreState` — the navigation state machine: category availability and filtering, category selection with wrap-around, the "filter matches no category → fall through to searching all of them" transition, and the consecutive-empty-results counter that closes a menu making no progress. It also holds the session working directory, which the front-end keeps current; that is the seam PR 2 in the stack pushes down into the data sources.
- `is_at_menu_trigger` / `should_close_at_menu` — the `@` grammar, lifted from private methods on `Input`.
- `shared_inserted_text` — the buffer text each accepted action inserts, for the actions whose representation is identical on every surface.

`app/src/search/ai_context_menu/mixer.rs` gains `install_sources_for_category`, `install_sources_for_all_categories`, and `at_context_menu_query`, mirroring `build_slash_command_mixer` / `slash_command_query` in `terminal/input/slash_commands/mixer.rs`.

### Why this shape

It follows the slash-command menu, which already solved this split: a shared core plus per-surface wrappers, with only genuinely generic menu primitives (`InlineMenuSelection`, `InputDrivenInlineMenuLifecycle`) living in `warp_search_core`, and two separate menu models rather than one. Result selection deliberately stays per-surface — the GUI's lives in `SearchBarState` wired to `QueryResultRenderer`, the TUI's will use `TuiInlineMenuListState`.

One deviation from that prior art: the shared behavior is inherent methods on `AtContextMenuCoreState` rather than a trait with `state()` / `state_mut()` and default methods. `SlashCommandDataSource`'s trait earns its keep because its default methods call back into surface-specific policy; nothing in the `@` state machine needs a surface hook, since the gates are plain data.

### Net effect

`view.rs` loses ~840 lines. Most of that is the two copies of the per-category source-wiring match arms that `reset_mixer` and `setup_data_sources_for_all_categories` each carried — 407 lines collapse to 39.

### Behavior preserved deliberately

Two existing quirks survived the extraction on purpose, so this PR stays a no-op. Both are worth fixing, separately:

1. **All-categories search installs no current-folder file source.** `setup_data_sources_for_all_categories` had a `RepoFiles` arm but no `CurrentFolderFiles` arm, so falling through to all-categories search outside a git repository returns no file results. Unifying on one `install_sources_for_category` would have silently fixed that, so the GUI call site filters `CurrentFolderFiles` out of the category list it passes. The shared function itself handles the category correctly — the follow-up fix is deleting the filter at the call site.
2. **`is_at_menu_trigger` indexes with `chars().nth()` against a byte offset.** Callers compute `cursor_position` as a byte offset, so the predicate misbehaves for non-ASCII text before the `@`. Semantics preserved exactly rather than quietly changed inside a refactor.

`close()` and `reset_menu_state()` also keep their existing asymmetry: only `reset_menu_state()` clears the category filter and selection, so a stale filter can still survive a close/reopen. That is expressed as two separate core methods — `return_to_main_menu_if_multiple_categories` and `clear_category_filter` — rather than one combined one.

### The one structural change worth a look

`available_categories` now reads a working directory cached on `AtContextMenuCoreState`, where the old `get_categories_for_mode` re-read `app.windows().state().active_window` on every call. The cache is refreshed from the same three notifications the view already subscribed to in order to redraw — `DetectedRepositories`, `WindowManager`, and the `ActiveSession` observe — plus on any gate change, and `set_session_state` notifies whenever the working directory changes. So the answer should never be stale, but it is the one place where this is equivalent-by-argument rather than mechanically identical, and so the thing most worth a second pair of eyes.

## Linked Issue

CODE-1910

## Testing

New `app/src/search/ai_context_menu/core_tests.rs` — 8 unit tests, one per distinct behavior in the extracted core, kept deliberately small since this is a no-op refactor:

- the `@` trigger grammar and the close conditions, each as a small table (start of buffer, after a space, after punctuation, suppressed mid-word; cursor behind the `@`, a newline, two consecutive spaces, a single space allowed)
- the text every action variant inserts, including the two the core declines
- the no-progress counter: that it closes after enough settled empty results, that the category list is exempt because it filters locally, and that loading or a rewritten query does not count
- category availability for the one branch no feature flag varies, the fall-through to all-categories, and category selection wrap-around

Clean: `./script/format`, `cargo clippy -p warp --all-targets --tests -- -D warnings`, and `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings` (the pass that unifies `warp/tui`).

Two gaps to be aware of when reviewing:

- **Not yet exercised via `./script/run`.** This touches a shipped surface, so it wants a manual pass before leaving draft: open the `@` menu, filter categories, enter one, accept an item, and confirm the fall-through search when the filter matches no category name.
- **The WASM cfg arms are unverified locally.** Presubmit has no WASM build step and does not run `--all-features` on `warp`, and this change touches cfg-gated code in `core.rs`, `mixer.rs`, and `view.rs`.

## Agent Mode

### DIFF
--- a/app/src/editor/view/mod.rs
+++ b/app/src/editor/view/mod.rs
@@ -102,10 +102,9 @@ use crate::editor::RangeExt;
 use crate::editor::accept_autosuggestion_keybinding_view::AcceptAutosuggestionKeybinding;
 use crate::editor::autosuggestion_ignore_view::{AutosuggestionIgnore, AutosuggestionIgnoreEvent};
 use crate::features::FeatureFlag;
+use crate::search::ai_context_menu::AIContextMenuCategory;
 use crate::search::ai_context_menu::mixer::AIContextMenuSearchableAction;
-use crate::search::ai_context_menu::view::{
-    AIContextMenu, AIContextMenuCategory, AIContextMenuEvent,
-};
+use crate::search::ai_context_menu::view::{AIContextMenu, AIContextMenuEvent};
 use crate::server::telemetry::TelemetryEvent;
 #[cfg(feature = "voice_input")]
 use crate::settings::AISettingsChangedEvent;
--- a/app/src/search/ai_context_menu/core.rs
+++ b/app/src/search/ai_context_menu/core.rs
@@ -1,0 +1,572 @@
+//! Surface-neutral state for the `@` context menu.
+//!
+//! This module owns the parts of the menu whose meaning is identical on every
+//! front-end: which categories are available, the navigation state machine
+//! between the category list and a category's results, the `@` trigger grammar,
+//! and the buffer text each accepted action inserts.
+//!
+//! Front-ends keep their own selection model, rendering, key handling, and
+//! search-mixer ownership. The GUI drives this state from
+//! [`crate::search::ai_context_menu::view::AIContextMenu`]; the TUI drives it
+//! from its own inline-menu model.
+
+#[cfg(not(target_family = "wasm"))]
+use repo_metadata::repositories::DetectedRepositories;
+use settings::Setting as _;
+use warp_core::features::FeatureFlag;
+use warp_util::local_or_remote_path::LocalOrRemotePath;
+use warpui::AppContext;
+#[cfg(not(target_family = "wasm"))]
+use warpui::SingletonEntity as _;
+
+use crate::drive::settings::WarpDriveSettings;
+use crate::search::ai_context_menu::mixer::AIContextMenuSearchableAction;
+use crate::settings::InputSettings;
+
+/// How many consecutive empty result sets the user may type past before the
+/// menu gives up and closes itself.
+const MAX_CONSECUTIVE_EMPTY_RESULTS_EVENTS: usize = 7;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AIContextMenuCategory {
+    CurrentFolderFiles,
+    RepoFiles,
+    Commands,
+    Blocks,
+    Workflows,
+    Notebooks,
+    Plans,
+    Diffs,
+    Docs,
+    Tasks,
+    Rules,
+    Servers,
+    Terminal,
+    Web,
+    RecentDiff,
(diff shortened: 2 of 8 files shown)


harryalbert:
* **#14511** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14511?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14504** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14504?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14502** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14502?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14502?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #14512, opened 2026-07-29
Preserve numeric CSS font-weights when pasting rich text

### LINKED ISSUE #14429: Paste maps numeric font-weight to a Bold boolean, dropping other CustomWeight variants
The HTML clipboard paste path collapses CSS `font-weight` to a single boolean: `crates/markdown_parser/src/html_parser.rs:79-89` sets `self.bold = weight_value > 400`, and `phrasing_to_formatted_text` (`html_parser.rs:463-468`) then emits only `Some(CustomWeight::Bold)` or `None`. So pasting rich text with `font-weight: 900` (Black) or `300` (Light) lands as plain Bold or as no weight — the specific weight is lost on paste.

A bidirectional numeric `Weight ↔ CustomWeight` mapping already exists at `crates/warpui_core/src/fonts.rs:54-121`, and the full variant set (`Thin, ExtraLight, Light, Medium, Semibold, Bold, ExtraBold, Black`) is defined in `crates/markdown_parser/src/weight.rs:7-15`. Paste could map the numeric value to the nearest `CustomWeight` instead of booleanizing it.

Note: this is also the missing "front door" for #13949 — that fix makes the strip-all path clear any weight, but there's currently no user gesture that can introduce a non-Bold weight into a buffer run, so the fix has no user-visible trigger until paste (or some other ingest) preserves the weight. Out of scope for #13952.

fbartho:
Closes #14429

Pasting rich text with numeric CSS font-weights dropped the specific weight: the HTML parser stored weight as a bold boolean, so everything above 400 became Bold and everything else became plain. Weights like 300 (Light) or 900 (Black) never survived the paste.

The parser now tracks an `Option<CustomWeight>` and maps numeric `font-weight` values through a new `CustomWeight::from_css_numeric` helper (rounds to the nearest named step, 100 Thin … 900 Black, 400 → none). The `bold`/`bolder` keywords and `<b>`/`<strong>` still map to Bold; `normal`/`lighter` clear any weight.

As a consequence, `font-weight: 600` now round-trips to Semibold instead of collapsing to Bold — two existing parser tests are updated to match. Added regression tests covering the full 100–900 numeric range and the keyword cases, plus unit tests for `from_css_numeric`.

Testing: `cargo nextest run -p markdown_parser` (153 passed); `cargo clippy -p markdown_parser --all-targets --all-features --tests -- -D warnings` (clean); `cargo test --doc -p markdown_parser`; verified `warp_editor` builds against the updated parser API. Manual paste verification on a dev build is in progress — I'll add evidence before marking this ready for review.

CHANGELOG-FIX: Pasted rich text now preserves specific font weights (Light, Semibold, Black, …) instead of collapsing them to Bold.

### Manual Paste Verification

Screenshot showing multiple font weights pasted from HTML in a browser into a new Notebook
fig. Screenshot showing multiple font weights pasted from HTML in a browser into a new Notebook


crates/markdown_parser/src/weight.rs
```diff
@@ -16,6 +16,28 @@ pub enum CustomWeight {
...
 
 impl CustomWeight {
+    /// Maps a numeric CSS `font-weight` value to the closest named weight.
+    ///
+    /// CSS numeric weights run 1–1000, with the common named steps landing on the hundreds
+    /// (100 Thin … 900 Black). We round to the nearest hundred and map that bucket to a variant.
+    /// `400` (Normal) has no `CustomWeight` and returns `None`, as do values that round to it.
+    pub fn from_css_numeric(value: i32) -> Option<CustomWeight> {
+        // Round to the nearest hundred, then clamp into the 100..=900 named range.
+        let bucket = (((value + 50) / 100) * 100).clamp(100, 900);
```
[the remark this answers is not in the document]
fbartho: Addressed in 146cdcc04.

crates/markdown_parser/src/weight.rs
```diff
@@ -35,3 +61,95 @@ impl CustomWeight {
         first.or(second)
     }
 }
+
+#[cfg(test)]
+mod tests {
```
bnavetta: This should go in a separate test file per repo convention

bnavetta (changes requested):
One testing thing to change, but code looks great!

crates/markdown_parser/src/weight.rs
fbartho: Addressed in 55a66704d.

bnavetta (approved):

PR #14535, opened 2026-07-30
docs: adopt the repo docs contract (tier 1)

nshonda:
Adopts the [repo docs contract](https://github.com/nshonda/workstation-setup/pull/909) so this repo carries agent-facing context that cannot go stale.

Two **generated blocks** in the README:

- **Repo map** — top-level structure, tracked file counts, primary type, and each directory's own README first line as its purpose
- **Commands** — build/test/lint entry points read from the repo's own manifests and script headers

A script emits both; a blocking CI job fails when the emitted form and the committed form disagree. Every cell is derived from the tree — where a fact is absent, the generator emits `—` rather than a guess.

Also vendors `scripts/lint-docs.sh` + `scripts/docs/gen-*.sh` and adds `.githooks/pre-commit`.

**The hook is staged, not activated.** `core.hooksPath` is repo-wide, so setting it while the hook lives only on this branch would point the main checkout at an empty hooks dir and silently disable the existing one. After merge:

    git -C <repo> config core.hooksPath .githooks

Refresh the tables any time with `./scripts/lint-docs.sh --fix`.

### Not included

`CONTRIBUTING.md` and `GLOSSARY.md` are part of the tier-1 set but are **not** authored here. Writing conventions for a repo without studying it is exactly the invention the contract forbids (§2.6, §5.2) — they should be written deliberately, then stamped.


nshonda:
Opened by mistake against the upstream repo — this targets a personal fork. Closing immediately, apologies for the noise.

PR #14544, opened 2026-07-30
[fix](VA) Keep in-progress playwright-cli commands in computer-use recordings

vkodithala:
## Description
On Linux, computer-use video recordings are smart-trimmed after stopping: only committed recording action groups produce keep-windows, and everything else is hard-cut. Shell commands never enter the recording timeline, so when the agent drives the browser via `playwright-cli`, all of that on-screen work was cut from the video.

This PR keeps in-progress `playwright-cli` commands in the recording:
- `ShellCommandExecutor` now detects `playwright-cli` invocations (skipping leading env-var assignments and resolving the program path's file name, so `npm install playwright-cli` or `echo playwright-cli` don't match) and opens a recording action group before the command starts.
- When the command finishes (any exit code), the pending group is committed; if the action is cancelled or the block is not found, it is discarded.
- A `playwright-cli` command that outlives the executor's poll window returns a long-running snapshot; the group stays open and is committed when a later `ReadShellCommandOutput` poll observes the finished block. Since that poll runs in a separate executor call, a new `RecordingController::commit_action_group_now` helper commits the pending group using the recording's own elapsed clock (no-op when no recording is active or no group is pending).
- If the recording is stopped/finalized while a group is still pending (e.g. a long-running `playwright-cli` session never observed finished), claiming the active recording for finalization settles the pending group at the stop point instead of dropping it, via a shared `ActiveRecording::commit_pending_group_now` helper also used by `begin_action_group`'s auto-commit.

Other shell commands (npm/cargo/etc.) never open a group, so they remain trimmed exactly as before. Behavior is unchanged when no recording is active: all recording calls no-op unless a recording is Active for the conversation.

## Linked Issue
None — recording-quality fix scoped and dispatched via Oz orchestration.

## Testing
- Added a unit test for the `playwright-cli` command detector (`detects_playwright_cli_commands`) covering plain, env-prefixed, and absolute-path invocations plus non-matches (`npm install playwright-cli`, `echo playwright-cli`, `cargo build`).
- Added a regression test (`finalization_commits_open_pending_group`) asserting that finalizing while a group is pending commits its window instead of dropping it.
- `cargo nextest run -p warp -E 'test(detects_playwright_cli_commands) or test(block_working_directory_updated_does_not_drain_finish_senders) or test(/recording_controller/)'` — 17/17 passed.
- `./script/format --check` — clean.
- `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`, `cargo clippy -p warp --all-targets --tests -- -D warnings`, and `cargo clippy -p warp_completer --all-targets --tests -- -D warnings` — all green.

Runtime behavior only differs on Linux (macOS recordings keep everything, no smart cut), so this was verified on macOS via build + unit tests rather than a live recording.

## Agent Mode

CHANGELOG-BUG-FIX: In-progress playwright-cli browser automation is now preserved in computer-use video recordings instead of being trimmed out.

### DIFF
--- a/app/src/ai/blocklist/action_model/execute/shell_command.rs
+++ b/app/src/ai/blocklist/action_model/execute/shell_command.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -23,6 +24,7 @@ use crate::ai::agent::{
     TransferShellCommandControlToUserResult, WriteToLongRunningShellCommandResult,
 };
 use crate::ai::blocklist::BlocklistAIPermissions;
+use crate::ai::blocklist::action_model::recording_controller::RecordingController;
 use crate::ai::blocklist::permissions::CommandExecutionPermission;
 use crate::ai::execution_profiles::WriteToPtyPermission;
 use crate::terminal::TerminalModel;
@@ -258,6 +260,16 @@ impl ShellCommandExecutor {
                     } else {
                         command.clone()
                     };
+                // Keep `playwright-cli` browser automation visible in an active
+                // computer-use recording: open an action group before the command
+                // starts so the smart cut does not trim its on-screen work.
+                let conversation_id = input.conversation_id;
+                let is_playwright_cli = is_playwright_cli_command(command);
+                if is_playwright_cli {
+                    RecordingController::handle(ctx).update(ctx, |controller, _| {
+                        controller.begin_action_group(conversation_id, Vec::new());
+                    });
+                }
                 ctx.emit(ShellCommandExecutorEvent::ExecuteCommand {
                     action_id: action_id.clone(),
                     command: decorated_command,
@@ -275,6 +287,24 @@ impl ShellCommandExecutor {
                             handle.update(ctx, |me, _| {
                                 me.block_finished_senders.remove(&block_selector);
                                 me.force_refresh_senders.remove(&block_selector);
+                            });
+                        }
+
+                        if is_playwright_cli {
+                            RecordingController::handle(ctx).update(ctx, |controller, _| {
+                                match &result {
+                                    // Commit regardless of exit code: failed browser
+                                    // automation is still on-screen work worth keeping.
+                                    ActionResult::CommandFinished { .. } => {
+                                        controller.commit_action_group_now(conversation_id);
+                                    }
+                                    ActionResult::Cancelled | ActionResult::BlockNotFound => {
+                                        controller.discard_action_group(conversation_id);
+                                    }
+                                    // Still running; the group stays open until a later
+                                    // poll observes the finished block.
+                                    ActionResult::LongRunningCommandSnapshot { .. } => {}
+                                }
                             });
                         }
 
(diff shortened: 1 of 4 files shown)


app/src/ai/blocklist/action_model/execute/shell_command.rs
```diff
@@ -278,6 +290,24 @@ impl ShellCommandExecutor {
...
+                                    // automation is still on-screen work worth keeping.
+                                    ActionResult::CommandFinished { .. } => {
+                                        controller.commit_action_group_now(conversation_id);
+                                    }
+                                    ActionResult::Cancelled | ActionResult::BlockNotFound => {
+                                        controller.discard_action_group(conversation_id);
+                                    }
+                                    // Still running; the group stays open until a later
+                                    // poll observes the finished block.
+                                    ActionResult::LongRunningCommandSnapshot { .. } => {}
```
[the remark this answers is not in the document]
vkodithala: Good catch — addressed in 146813997. `claim_matching_finalization` now settles any open pending group before it transitions the recording to `Finalizing`, via a new `ActiveRecording::commit_pending_group_now` that commits the pending group at the recording's current elapsed time (clamped to `>=` the pending start). So a `playwright-cli` group left open by the snapshot/long-running path is committed up to the stop point instead of being dropped as "no committed actions". Added a regression test (`finalization_commits_open_pending_group`), and refactored `begin_action_group`'s auto-commit and `commit_action_group_now` to reuse the same helper.

app/src/ai/blocklist/action_model/execute/shell_command.rs
```diff
@@ -258,6 +260,16 @@ impl ShellCommandExecutor {
...
+                // Keep `playwright-cli` browser automation visible in an active
+                // computer-use recording: open an action group before the command
+                // starts so the smart cut does not trim its on-screen work.
+                let conversation_id = input.conversation_id;
+                let is_playwright_cli = is_playwright_cli_command(command);
+                if is_playwright_cli {
+                    RecordingController::handle(ctx).update(ctx, |controller, _| {
+                        controller.begin_action_group(conversation_id, Vec::new());
+                    });
+                }
```
danielpeng2: the logic behind beginning an action group feels like an implementation detail of the `RecordingController` that we don't need to leak here. here, and in the locations where we close the action group, it might be nice to define a method on the RecordingController and call it like:
```
RecordingController::maybe_begin_action_group(...)
```

app/src/ai/blocklist/action_model/execute/shell_command.rs
```diff
@@ -258,6 +260,16 @@ impl ShellCommandExecutor {
                     } else {
                         command.clone()
                     };
+                // Keep `playwright-cli` browser automation visible in an active
+                // computer-use recording: open an action group before the command
+                // starts so the smart cut does not trim its on-screen work.
+                let conversation_id = input.conversation_id;
+                let is_playwright_cli = is_playwright_cli_command(command);
```
danielpeng2: not a now thing, but i wonder if we want to let the agent decide which actions should be visible in a recording. hardcoded detection of the playwright CLI seems fine, but if the agent ever uses any other command to control the desktop, that'd be missed.

one idea is to augment the run_command tool we provide to the agent with a `is_meaningful_action` parameter or something like that

danielpeng2 (approved):
left a couple non-blocking comments

app/src/ai/blocklist/action_model/execute/shell_command.rs
vkodithala: Yeah, I think this is a fair point. Probably OOS here but something we should consider.

app/src/ai/blocklist/action_model/execute/shell_command.rs
vkodithala: Yeah, I agree that this should be abstracted into the `RecordingController`. Recent commit should handle this.

PR #14548, opened 2026-07-30
Check out pinned refs for agent source repositories

cephalonaut:
## Description

Adds optional per-repository `checkout_ref` support to agent environment setup.

When a source repository includes a ref, Warp now:

1. Clones the repository using the existing treeless clone.
2. Fetches the requested commit, branch, or tag from `origin`.
3. Checks out that ref before starting the agent.

This lets benchmark trials start from the pre-work repository state rather than the current default branch. Tasks without a checkout ref retain the existing behavior.

## Testing

- [x] `SourceRepo.checkout_ref` serializes and remains optional.
- [x] A real Git fixture fetches and checks out a commit not reachable from the default branch.
- [x] An unknown ref returns `CheckoutFailed`.
- [x] An absent ref leaves the clone on its default branch.
- [x] A local oz-local benchmark run passed an exact `warp-server` SHA to two real-model trials; both completed successfully.
- [x] `cargo fmt --all -- --check`

Full presubmit completed formatting and clippy, then encountered unrelated current-master timing flakes in `submit_cli_agent_rich_input_opencode_defers_enter_and_close` and `shell_integration_tests::test_ctrl_c`. The first passed when rerun in isolation; targeted checkout-at-SHA tests pass.

## Agent Mode

Warp Agent Mode conversation: https://staging.warp.dev/conversation/09d2ec39-7b85-425f-9fc9-52fa5a7bd6e1

CHANGELOG-NONE


app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -334,12 +339,17 @@ clone_repo() {
   repo_name="$1"
   repo_url="$2"
   target="$3"
+  checkout_ref="$4"
   if [ -d "$target" ]; then
     printf '%s\n' "Repository directory $target already exists, skipping clone..."
     return 0
```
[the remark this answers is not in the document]
cephalonaut: [Warp Agent] Fixed in https://github.com/warpdotdev/warp/commit/3751ff69094e2a06261437812ca12bc17fa704f7. The multi-repo shell helper no longer returns before pinning when the target directory already exists. If `checkout_ref` is set, it still fetches and checks out the requested object; with no ref, existing directories remain untouched. Added real-shell regression tests for both cases.

app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -488,11 +502,69 @@ pub(super) async fn clone_repo(
...
+    working_dir: &Path,
+    shell_type: ShellType,
+) -> Option<String> {
+    let checkout_ref = repo.checkout_ref.as_deref()?;
+    let repo_dir = working_dir.join(&repo.repo);
+    let escaped_dir = shell_escape_single_quotes(&repo_dir.to_string_lossy(), shell_type);
+    let escaped_ref = shell_escape_single_quotes(checkout_ref, shell_type);
+    Some(format!(
+        "git -C '{escaped_dir}' fetch --filter=tree:0 origin '{escaped_ref}' && \
+         git -C '{escaped_dir}' checkout '{escaped_ref}'"
```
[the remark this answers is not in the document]
cephalonaut: [Warp Agent] Fixed in https://github.com/warpdotdev/warp/commit/3751ff69094e2a06261437812ca12bc17fa704f7. Both the single-repo and parallel paths now run `git checkout --detach FETCH_HEAD` after fetching, so they pin the object just fetched instead of a stale local branch or unavailable ref name. Added coverage for FETCH_HEAD-only refs and stale local branches.

app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -334,12 +339,20 @@ clone_repo() {
...
+  checkout_ref="$4"
   if [ -d "$target" ]; then
     printf '%s\n' "Repository directory $target already exists, skipping clone..."
-    return 0
+  else
+    printf '%s\n' "Cloning repository $repo_name..."
+    git clone --filter=tree:0 "$repo_url" "$target" || return 1
+  fi
+  # Pin after clone or reuse: a reused directory may still be on an old ref.
+  if [ -n "$checkout_ref" ]; then
```
cephalonaut, line 350: Does it make sense to pin on a reuse?

captainsafia: I think so? If the primary goal is benchmarking then it feels sensible to prioritze that we are pinning the correct ref.

cephalonaut: Yeah, I think I agree. The setup should fail if you have something that can't be pinned (an existing repo with changes for instance).

captainsafia (approved):

PR #14573, opened 2026-07-31
REMOTE-2111 (2/3): Add periodic checkpoint coordinator state machine

joeywangzr:
Part 2 of 3 in a stack splitting the periodic workspace-handoff checkpoint client work into reviewable chunks.

## Stack
1. #14588 — checkpoint upload/commit mechanics (base: `master`)
2. **This PR** — periodic checkpoint coordinator state machine (base: #14588's branch)
3. #14589 — wire the coordinator into AgentDriver (base: this branch)

Stacked on #14588; only the diff vs. that branch is new in this PR. Please review against #14588's branch, not `master`.

## What this does
Adds the `Idle -> Due -> InFlight -> Idle` scheduling loop (with `Finalizing -> Stopped` reachable from any state) that drives periodic workspace-handoff checkpoints:
- `checkpoint_coordinator.rs`: `CheckpointCoordinatorHandle` spawns the loop and exposes `finalize()` for graceful shutdown. The timer only ever moves `Idle` to `Due`; `Due` polls a safe-boundary predicate (conversation not mid-turn) before entering `InFlight`, which runs exactly one attempt via `snapshot::run_checkpoint_from_declarations_file` on a background task so a racing `finalize()` can bound how long it waits without ever stranding the attempt. `finalize()` starts at most one more attempt (skipped below the gather/upload timeout floor) or awaits an already-in-flight one — never both — bounded end to end by its budget.
- `terminal/view.rs`: add `TerminalView::ai_action_model()`, which `is_safe_boundary` uses to check for unfinished actions on the conversation.

**Note on `#![allow(dead_code)]`:** this module has no production caller yet — `AgentDriver` wires `CheckpointCoordinatorHandle::new` into its spawn/finalize lifecycle in #14589 — so it's marked at the module level with an explanatory comment. Becomes genuinely reachable once the stack is applied in full.

## Validation
`cargo clippy --all-targets --tests -- -D warnings`, `cargo fmt --check`, and the full coordinator test suite (13 tests covering jitter bounds, finalize-below/above-floor from both Idle and InFlight, safe-boundary deferral and retry, interval timing, and orphan-safety) all pass on this branch.

### DIFF
--- a/app/src/ai/agent_sdk/driver/checkpoint_coordinator.rs
+++ b/app/src/ai/agent_sdk/driver/checkpoint_coordinator.rs
@@ -1,0 +1,456 @@
+//! Periodic workspace-handoff checkpoint coordinator (REMOTE-2111 Phase 3).
+//!
+//! Drives the five-state machine described in `docs/remote-2111-checkpoint-spec.md`
+//! (warp-server): `Idle -> Due -> InFlight -> Idle` on the periodic path, with
+//! `Finalizing -> Stopped` reachable from `Idle`, `Due`, or `InFlight` via
+//! [`CheckpointCoordinatorHandle::finalize`]. The timer only ever moves `Idle` to
+//! `Due`; all gather/upload/commit work happens through
+//! `super::snapshot::run_checkpoint_from_declarations_file`, reusing the same
+//! declarations file and gather/upload pipeline as the legacy end-of-run snapshot.
+//!
+//! Safe-boundary gating ("only touch the filesystem/network when the conversation
+//! isn't mid-turn") is implemented as a bounded poll of `AgentDriver`'s own state via
+//! its `ModelSpawner`, rather than a push subscription: `AgentDriver` already reads
+//! exactly the state needed (`run_conversation_id`, the terminal view's action model)
+//! through this same read-only, spawner-based pattern used by `run_snapshot_upload`.
+//! This trades a small amount of latency (up to [`SAFE_BOUNDARY_POLL_INTERVAL`]) for
+//! avoiding new push-subscription wiring through the UI model graph.
+
+// This module has no production caller yet -- `AgentDriver` wires
+// `CheckpointCoordinatorHandle::new` into its spawn/finalize lifecycle in a
+// follow-up, stacked PR. Until then, `CheckpointCoordinatorHandle::new` (and
+// everything it reaches) is unreachable outside `#[cfg(test)]`, which reaches the
+// same code through `new_for_test`. This module-level allow is temporary and should
+// be removable once that PR merges on top of this one.
+#![allow(dead_code)]
+
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+
+use futures::FutureExt as _;
+use futures::future::BoxFuture;
+use instant::Instant;
+use rand::Rng as _;
+use tokio::sync::{mpsc, oneshot};
+use warpui::r#async::executor::Background;
+use warpui::r#async::{FutureExt as _, Timer};
+use warpui::{ModelSpawner, SingletonEntity};
+
+use super::AgentDriver;
+use super::snapshot::{self, CheckpointResult};
+use crate::ai::ambient_agents::AmbientAgentTaskId;
+use crate::ai::blocklist::BlocklistAIHistoryModel;
+use crate::server::server_api::harness_support::{CheckpointGeneration, HarnessSupportClient};
+
+/// A safe-boundary predicate, decoupled from `ModelSpawner<AgentDriver>` so
+/// [`coordinator_loop`] can be exercised in isolation by tests. Production code builds
+/// this from [`is_safe_boundary`]; tests supply a directly-controllable closure.
+type BoundaryCheck = Arc<dyn Fn() -> BoxFuture<'static, bool> + Send + Sync>;
+
+/// Default cadence between the start of one checkpoint attempt and the timer firing
+/// again, absent an override on `AgentDriverOptions`. Deliberately not
+/// `HARNESS_SAVE_INTERVAL` (30s) -- see the spec's "5 minutes plus jitter" cadence.
+pub(super) const DEFAULT_CHECKPOINT_INTERVAL: Duration = Duration::from_secs(5 * 60);
+/// Upper bound on additive jitter, so agents scheduled at the same time don't all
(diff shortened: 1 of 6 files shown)


joeywangzr:
* **#14589** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14589?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14573** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14573?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14573?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14588** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14588?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

joeywangzr:
Pushed a follow-up commit to this branch from a review pass (e3bc04fe), and rebased onto the updated mechanics branch. Summary:

**Blocking — the finalize floor was unreachable, so the end-of-run checkpoint never ran.**
`finalize_with_new_attempt` gates on `remaining > script_timeout + upload_timeout`, but PR #14589 passes `upload_timeout` alone as the budget. That comparison is false for *every* possible configuration (defaults: `120s > 60s + 120s`; even with `--snapshot-script-timeout 0` it's `120s - ε > 120s`). Since the coordinator `return`s past the legacy one-shot upload once active, enabling the flag meant **no end-of-run snapshot at all** — runs shorter than one interval uploaded nothing, longer runs lost everything since the last periodic tick.

Added `finalize_budget(script_timeout, upload_timeout)` so the floor has exactly one owner and callers can't drift from it; #14589 now uses it. Also downgraded the skip log to WARN, since reaching it means the end-of-run checkpoint was dropped. `coordinator_finalize_from_idle_skips_attempt_below_floor` was the only test covering this gate, and it asserted the skip — I kept it and added the positive case.

**Blocking — the loop could never terminate once the driver was gone.**
`is_safe_boundary` mapped `Err(ModelDropped)` to "safe", so a dropped `AgentDriver` left the coordinator gathering and uploading the whole workspace every interval forever. Nothing else stops the loop, and `run_snapshot_upload` returns without finalizing under exactly that condition (spawner round trip fails, or `OzHandoff` toggled off). Replaced the boolean with a `Boundary` tri-state so `DriverGone` returns from `coordinator_loop`.

**`Blocked` conversations never checkpointed.** `Blocked` is itself backed by a pending action, so it fell through to `has_unfinished_actions_for_conversation` and reported busy forever — a run parked on user approval (often hours) polled every 2s and never checkpointed, though nothing was mutating the workspace. `Blocked` and `is_done` are now quiescent alongside `WaitingForEvents`. Also added `MAX_BOUNDARY_DEFERRAL` (10m): without a cap, any turn longer than the interval starves the feature in exactly the long-running case it exists for. A slightly-inconsistent checkpoint beats none, since the previous generation is only replaced on success.

**Boundary polling now stays responsive to finalize.** The check was awaited outside `select!`; it's a round trip through the model task queue, so a stall could wedge shutdown behind an unbounded await.

**Declarations writer is flushed before every attempt.** `run_snapshot_upload` does this on the legacy path specifically so no driver-side `file` append is in flight when the bash script starts appending to the same JSONL; the coordinator didn't, so checkpoints could miss the agent's latest edits and race the script.

Also gave `finalize`'s outer ack timeout slack over the inner deadline — bounding both on exactly `budget` meant the "defense-in-depth" bound would routinely preempt an attempt still legitimately running.

Five tests added (derived budget clears the floor; finalize during boundary polling still attempts; gone driver stops the loop; loop stops after finalize; failed attempt retries next interval). All 16 coordinator tests pass locally, plus clippy `-D warnings` and fmt.

Two things I left alone but flagged: periodic attempts route every failure through `report_error!`, which at a 5-minute cadence over a multi-hour run will flood Sentry if the endpoint is unhealthy; and `gather_failed`/`read_failed` entries still commit, so a transient `git diff` failure can replace a complete checkpoint with one missing that repo's patch.

app/src/ai/agent_sdk/driver/checkpoint_coordinator.rs
```diff
@@ -0,0 +1,624 @@
+//! Periodic workspace-handoff checkpoint coordinator (REMOTE-2111 Phase 3).
+//!
+//! Drives the five-state machine described in `docs/remote-2111-checkpoint-spec.md`
```
bnavetta: nit: mind updating this comment to not reference the PR sequence / server specs? It adds a bit of indirection for future readers to follow, vs. a self-contained description in the comment

app/src/ai/agent_sdk/driver/checkpoint_coordinator.rs
```diff
@@ -0,0 +1,624 @@
...
+            Boundary::Busy => {}
+        }
+        if due_since.elapsed() >= MAX_BOUNDARY_DEFERRAL {
+            log::warn!(
+                "Conversation has not reached a safe boundary in {MAX_BOUNDARY_DEFERRAL:?}; \
+                 checkpointing anyway rather than skipping this cycle entirely"
+            );
+            return DueOutcome::Safe;
+        }
+        futures::select! {
```
bnavetta, line 434: During the lifetime of the agent, it'll almost always be waiting on inference or performing an action. In practice, I expect we'll mostly only checkpoint due to the `MAX_BOUNDARY_DEFERRAL` guard.

It should be fine to generate a snapshot while waiting on the MAA inference request, since the agent won't be making any filesystem changes. I think we probably do need to push a signal from the driver, and try to take advantage of that window for snapshotting

app/src/ai/agent_sdk/driver/checkpoint_coordinator_tests.rs
```diff
@@ -0,0 +1,869 @@
...
+}
+
+/// Writes a single-file declarations JSONL at the default per-task-id path that
+/// `run_one_attempt` resolves internally (it has no path-override hook), and removes the
+/// per-task directory on drop so parallel tests never collide (each uses a fresh task id).
+struct DeclarationsFixture {
+    dir: PathBuf,
+}
+
+impl DeclarationsFixture {
```
bnavetta, line 51: Nice!

bnavetta (approved):

PR #14588, opened 2026-07-31
REMOTE-2111 (1/3): Add checkpoint upload/commit pipeline mechanics

joeywangzr:
Part 1 of 3 in a stack splitting the periodic workspace-handoff checkpoint client work into reviewable chunks.

## Stack
1. **This PR** — checkpoint upload/commit mechanics (base: `master`)
2. #14573 — periodic checkpoint coordinator state machine (base: this branch)
3. #14589 — wire the coordinator into AgentDriver (base: #14573's branch)

## What this does
Adds the storage mechanics for one periodic-handoff checkpoint attempt, as a self-contained module the coordinator (#14573) will drive:
- `harness_support.rs`: `SnapshotUploadMode` (legacy|checkpoint) and `generation` on `SnapshotUploadRequest`; `CheckpointGeneration` (a validated, client-minted generation identifier); `CommitSnapshotRequest`/`Response`; `HarnessSupportClient::commit_snapshot`.
- `snapshot.rs`: `CheckpointResult` (Committed/Skipped/Failed outcome of one attempt); `PipelineMode` selects legacy vs. checkpoint-mode uploads; `mint_generation` (per-attempt, collision-free generation IDs); `storage_name` (generation-prefixed object naming); and the `run_checkpoint_from_declarations_file`/`run_checkpoint_pipeline` entry points: gather, upload every blob plus the manifest in checkpoint mode, then commit the exact object set that landed. Commit is withheld entirely (never partial) if the manifest or any required blob fails to upload, or if upload-target allocation fails.

**Note on `#[allow(dead_code)]`:** the whole pipeline has no production caller yet — the coordinator that drives it periodically lands in #14573 — so the new items are marked with an explanatory comment. These become genuinely reachable once the stack is applied in full; removable but harmless to leave.

## Validation
`cargo clippy --all-targets --tests -- -D warnings`, `cargo fmt --check`, and the full `driver::snapshot` test suite (54 tests) pass in isolation on this branch. Checkpoint coverage includes manifest-upload failure, upload-target allocation failure, a server-omitted blob target, commit retry exhaustion, filename sanitization against a mirror of the server's validation, and the JSON wire shape of legacy uploads, checkpoint uploads, and commits.

### DIFF
--- a/app/src/ai/agent_sdk/driver/snapshot.rs
+++ b/app/src/ai/agent_sdk/driver/snapshot.rs
@@ -48,7 +48,8 @@ use crate::server::server_api::ai::{
     UploadLocalHandoffSnapshotRequest,
 };
 use crate::server::server_api::harness_support::{
-    HarnessSupportClient, SnapshotFileInfo, SnapshotUploadRequest, UploadTarget, upload_to_target,
+    CheckpointGeneration, CommitSnapshotRequest, HarnessSupportClient, SnapshotFileInfo,
+    SnapshotUploadRequest, UploadTarget, upload_to_target,
 };
 
 /// Default path of the declarations file when neither the env var override nor a task ID
@@ -210,8 +211,9 @@ pub(super) async fn run_declarations_script(
 ///
 /// Reads `$OZ_SNAPSHOT_DECLARATIONS_FILE` for the operator/test override, then delegates to
 /// [`resolve_declarations_path_with_override`] so tests can exercise the pure logic without
-/// racing on the shared env var.
-fn resolve_declarations_path(task_id: Option<&AmbientAgentTaskId>) -> PathBuf {
+/// racing on the shared env var. `pub(super)` so `checkpoint_coordinator` can resolve the same
+/// path used by the declarations writer and by [`run_declarations_script`].
+pub(super) fn resolve_declarations_path(task_id: Option<&AmbientAgentTaskId>) -> PathBuf {
     resolve_declarations_path_with_override(task_id, std::env::var_os(DECLARATIONS_PATH_ENV_VAR))
 }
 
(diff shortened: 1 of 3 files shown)


joeywangzr:
* **#14589** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14589?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14573** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14588** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14588?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14588?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

joeywangzr:
Pushed a follow-up commit to this branch from a review pass (a682589e). Summary of what changed and why, so you can push back on any of it:

**Blocking — a short `upload-snapshot` response produced a silently-partial *committed* checkpoint.**
`upload_gathered_snapshot` already anticipates the server returning fewer targets than requested and only warns. The target-less blob then reached `upload_entry` and was recorded as `EntryStatus::Skipped`, and `run_checkpoint_pipeline` rejected only `Failed` — so the attempt committed a smaller object set and made it the *selected* checkpoint, discarding a previously complete one. The comment claiming only cap-skipped entries are excluded was conflating two very different causes under one status.

Split out `EntryStatus::NoTarget` and made it fatal for a checkpoint attempt. `Skipped` now means only the deliberate `MAX_SNAPSHOT_FILES_PER_RUN` cap. Both still surface in the manifest as `skipped` so rehydration consumers keep a stable status vocabulary — the distinguishing detail stays in `error`. Legacy end-of-run behavior is unchanged.

**Blocking — agent-controlled filenames flowed unsanitized into the generation-prefixed storage names.**
`gather_repo` sanitized its filename component; `gather_file` used the raw basename. That basename ends up inside `checkpoint_<generation>__<logical_name>`, and `CheckpointGeneration`'s docs call `__` a reserved separator with charset `[A-Za-z0-9._-]` — but nothing enforced it for the filename half, which comes from files the agent creates. `a__b.txt`, or `checkpoint_1700000000000-0__evil.txt`, yields an ambiguous storage name: depending on how the server splits, that either fails the existence check at commit time (losing the whole checkpoint) or attributes objects to a different generation.

`sanitize_name_component` now collapses to the server charset and squashes `_` runs so `__` can't appear; `gather_file` and `sanitize_filename_component` both route through it, and `storage_name` debug-asserts the invariant.

**Smaller:** `CheckpointGeneration::from_validated` now actually validates (debug assertion) rather than only claiming to; the "pending commit" log moved after the outcome check so it stops firing for attempts that failed to allocate targets; corrected the `Failed { generation: None }` doc, since the external timeout also reports `None` after a generation was minted and objects were uploaded.

Three regression tests added. Verified locally: `cargo check -p warp --lib --tests`, `cargo clippy -p warp --all-targets --tests -- -D warnings`, `cargo fmt --check`, and the snapshot test suite passes (65 tests across this branch and the two above it).

One thing I did **not** change, but think is worth a decision: the client reconstructs commit object names rather than using anything the server echoed (`UploadTarget` carries no object name). Any server-side naming or sanitization change silently breaks commit for the whole checkpoint. Having the server echo the object name per target would remove that coupling entirely.

app/src/ai/agent_sdk/driver/snapshot.rs
```diff
@@ -210,8 +211,9 @@ pub(super) async fn run_declarations_script(
 ///
 /// Reads `$OZ_SNAPSHOT_DECLARATIONS_FILE` for the operator/test override, then delegates to
 /// [`resolve_declarations_path_with_override`] so tests can exercise the pure logic without
-/// racing on the shared env var.
-fn resolve_declarations_path(task_id: Option<&AmbientAgentTaskId>) -> PathBuf {
+/// racing on the shared env var. `pub(super)` so `checkpoint_coordinator` can resolve the same
+/// path used by the declarations writer and by [`run_declarations_script`].
```
bnavetta: nit: we don’t need to document why this is public or where it’s called

app/src/ai/agent_sdk/driver/snapshot.rs
```diff
@@ -651,6 +663,67 @@ struct SnapshotOutcome {
     manifest_uploaded: bool,
 }
 
+/// Outcome of one checkpoint attempt, where [`SnapshotOutcome`] only covers per-entry upload
+/// results within that attempt.
+// The checkpoint pipeline below has no production caller until the periodic coordinator
```
bnavetta: nit: we shouldn’t leave a transient comment like this. Andy wrote https://warpdev.slack.com/archives/C0BDQDW8V5E/p1785788222543859 for the factory, but IMO it’s generally good guidance

app/src/ai/agent_sdk/driver/snapshot.rs
```diff
@@ -651,6 +663,67 @@ struct SnapshotOutcome {
...
+// lands in a follow-up, stacked PR; the `allow(dead_code)`s go away with it.
+#[allow(dead_code)]
```
bnavetta: If we use `expect` instead of `allow`, the compiler should flag us once this is referenced, so we can remove the `dead_code` silence

app/src/ai/agent_sdk/driver/snapshot.rs
```diff
@@ -651,6 +663,67 @@ struct SnapshotOutcome {
...
+        .as_millis();
+    let counter = GENERATION_COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+    CheckpointGeneration::from_validated(format!("{millis}-{counter}"))
+}
+
+/// Reproduce the server's `checkpoint_<generation>__<logical_name>` storage name.
+///
+/// Only the exact-set [`CommitSnapshotRequest`] needs this; everything earlier in the pipeline
+/// speaks logical names, and the server derives each presigned target's storage name itself.
+#[allow(dead_code)]
```
bnavetta: Ideally the client wouldn’t be aware of this format. Instead, when committing the snapshot, could the server generate the storage names?

app/src/ai/agent_sdk/driver/snapshot.rs
```diff
@@ -896,6 +970,126 @@ async fn run_pipeline(
...
+///
+/// Unlike [`upload_snapshot_from_declarations_file`], an unusable declarations file is
+/// [`CheckpointResult::Skipped`] rather than `None`, because the coordinator's state machine
+/// distinguishes "nothing to do" from "tried and failed".
+#[allow(dead_code)]
+pub(super) async fn run_checkpoint_from_declarations_file(
+    path: &Path,
+    client: Arc<dyn HarnessSupportClient>,
+) -> CheckpointResult {
+    log::info!("Checkpoint attempt starting from {}", path.display());
```
bnavetta: nit: we should use `safe_info!` when logging paths, so user filesystem paths don’t end up in Sentry

app/src/ai/agent_sdk/driver/snapshot.rs
```diff
@@ -1181,12 +1401,37 @@ async fn gather_file(
     pre_upload_entries: &mut Vec<EntryResult>,
 ) {
     let path = Path::new(file_path);
+    // Stat before reading so an oversized file is skipped without pulling it into memory.
```
bnavetta, line 1404: This is a good change, but somewhat unrelated to the original PR - probably could have gone in a followup!

bnavetta (approved):

PR #14589, opened 2026-07-31
REMOTE-2111 (3/3): Wire the periodic checkpoint coordinator into AgentDriver

joeywangzr:
Part 3 of 3 in a stack splitting the periodic workspace-handoff checkpoint client work into reviewable chunks.

## Stack
1. #14588 — checkpoint upload/commit mechanics (base: `master`)
2. #14573 — periodic checkpoint coordinator state machine (base: #14588's branch)
3. **This PR** — wire the coordinator into AgentDriver (base: #14573's branch)

Stacked on #14573; only the diff vs. that branch is new in this PR. Please review against #14573's branch, not `master`. This PR's tree, once the full stack is applied, is byte-identical to the original combined implementation for `driver.rs`, `mod.rs`, `terminal/view.rs`, and `warp_features/lib.rs`.

## What this does
Instantiates and drives `CheckpointCoordinatorHandle` from `AgentDriver`'s own lifecycle, completing REMOTE-2111 Phase 3:
- `AgentDriverOptions` gains `checkpoint_interval` (override for the default 5-minute-plus-jitter cadence; primarily for tests).
- `AgentDriver` spawns a coordinator when a cloud `task_id` is present, snapshot uploads aren't disabled, and both `FeatureFlag::OzHandoff` and the new `FeatureFlag::PeriodicHandoffCheckpoints` are enabled.
- Driver finalization routes through the coordinator's `finalize()` (bounded by the existing snapshot upload timeout) when a coordinator is active, instead of the legacy one-shot end-of-run snapshot path.
- `warp_features`: add `PeriodicHandoffCheckpoints`, off by default while the coordinator rolls out.

## Validation
`cargo clippy --all-targets --tests -- -D warnings`, `cargo fmt --check`, and the full checkpoint + coordinator test suites (19 + 13 tests respectively) all pass at the top of the stack. All `#[allow(dead_code)]` annotations added in #14588/#14573 are now provably unnecessary (clippy passes clean without any new warnings appearing once this PR's wiring makes everything reachable), confirming the split didn't drop any functionality.

### DIFF
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -434,6 +434,10 @@ pub struct AgentDriverOptions {
     pub snapshot_upload_timeout: Option<Duration>,
     /// Declarations script timeout override.
     pub snapshot_script_timeout: Option<Duration>,
+    /// Periodic checkpoint cadence override. Only used when
+    /// `FeatureFlag::PeriodicHandoffCheckpoints` is enabled; deliberately separate
+    /// from `snapshot_upload_timeout`/`snapshot_script_timeout` per-attempt budgets.
+    pub checkpoint_interval: Option<Duration>,
     /// Skip the initial `StartFromAmbientRunPrompt` so the agent waits for a
     /// follow-up instead of hallucinating an empty turn. Sourced from the
     /// `--skip-initial-turn` CLI flag, which the worker emits when the
@@ -511,6 +515,13 @@ pub struct AgentDriver {
     snapshot_disabled: bool,
     snapshot_upload_timeout: Duration,
     snapshot_script_timeout: Duration,
+
+    /// Periodic workspace-handoff checkpoint coordinator. `Some` only when
+    /// `FeatureFlag::OzHandoff` and `FeatureFlag::PeriodicHandoffCheckpoints` are both
+    /// enabled, the run has a cloud task id, and `--no-snapshot` was not set;
+    /// `None` otherwise, in which case `run_snapshot_upload` falls back to the legacy
+    /// one-shot upload path unchanged.
+    checkpoint_coordinator: Option<checkpoint_coordinator::CheckpointCoordinatorHandle>,
 
     /// Conversation ID this driver is running. Set at construction for
     /// resumed runs and on `ConversationServerTokenAssigned` for fresh
@@ -796,6 +807,7 @@ impl AgentDriver {
             snapshot_disabled,
             snapshot_upload_timeout,
             snapshot_script_timeout,
+            checkpoint_interval,
             skip_initial_turn,
             strict_mcp_startup,
             mcp_startup_timeout,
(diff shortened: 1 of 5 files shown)


joeywangzr:
* **#14589** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14589?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14589?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14573** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14573?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14588** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14588?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

joeywangzr:
Pushed a follow-up commit to this branch from a review pass (14612472), and rebased onto the updated coordinator branch. Summary:

**Blocking — `finalize(upload_timeout)` could never run the final attempt.**
The coordinator's floor is `script_timeout + upload_timeout`, so `remaining > floor` was false for every possible configuration (defaults: `120s > 60s + 120s`). Because this path `return`s past the legacy one-shot upload, enabling `PeriodicHandoffCheckpoints` meant **no end-of-run snapshot at all**: runs shorter than one 5-minute interval uploaded nothing, and longer runs lost everything since the last periodic tick. That's a strict regression against today's behavior, on the rollout flag's happy path.

Now uses `checkpoint_coordinator::finalize_budget(script_timeout, upload_timeout)`, added in #14573, which owns the floor so the two can't drift apart. Covered by a regression test that also asserts the old pairing provably can't clear the floor.

**Blocking — the coordinator never got the declarations writer.**
`run_snapshot_upload` flushes queued driver-side `file` appends before running the declarations script, precisely so nothing is in flight when the bash script starts appending to the same JSONL (see the comment on `DeclarationsWriterHandle::flush`). The coordinator branch returns before that block, so both periodic and final checkpoints could miss the agent's most recent edits. Now passes `snapshot_file_writer.clone()` through — the coordinator is gated on a superset of the writer's own conditions, so the handle is always present when the coordinator exists.

**Removed the `allow(dead_code)` scaffolding.**
Both #14588 and #14573 say their allows are "removable once that PR is merged on top of this one" — this is that PR, and it wasn't doing the removal. Dropped the module-wide `#![allow(dead_code)]` in `checkpoint_coordinator` (which would otherwise permanently blind the lint over new async/networking code) plus the per-item allows in `snapshot` and `harness_support`, along with the now-stale explanatory comments. Verified with the exact gate presubmit uses: `cargo clippy -p warp --all-targets --tests -- -D warnings`.

**Note on CI:** this PR and #14573 have only ever run the OWNERS/CLA checks, because their bases aren't `master` — the format/clippy/test matrix has never run against the coordinator or this wiring. Worth confirming green before merge. I ran `cargo check`, clippy `-D warnings`, `cargo fmt --check` and the 65 snapshot + coordinator unit tests locally on the stack tip; I could not run the full app test suite here because linking the test binary exceeds this sandbox's memory.

Two smaller things I left for you: `checkpoint_interval` is added to `AgentDriverOptions` but hard-wired to `None` at the only production construction site, so it's currently unreachable configuration — either plumb a CLI/env flag or drop the field. And the coordinator is spawned in `AgentDriver::new`, i.e. before environment setup, so the first tick can land mid-setup.

app/src/ai/agent_sdk/driver.rs
```diff
@@ -4294,6 +4345,26 @@ impl AgentDriver {
             return;
         }
 
+        // When the periodic checkpoint coordinator is active, it owns the entire
+        // end-of-run path: `finalize` drains the declarations writer, regenerates
+        // declarations, runs one last best-effort attempt, and commits it as the
+        // selected checkpoint. This replaces the legacy one-shot upload below so
```
dmichelin: nit: these comments are way too verbose

app/src/ai/agent_sdk/driver.rs
```diff
@@ -434,6 +434,10 @@ pub struct AgentDriverOptions {
     pub snapshot_upload_timeout: Option<Duration>,
     /// Declarations script timeout override.
     pub snapshot_script_timeout: Option<Duration>,
+    /// Periodic checkpoint cadence override. Only used when
```
dmichelin: nit: should be a one line comment

app/src/ai/agent_sdk/driver.rs
```diff
@@ -512,6 +516,13 @@ pub struct AgentDriver {
     snapshot_upload_timeout: Duration,
     snapshot_script_timeout: Duration,
 
+    /// Periodic workspace-handoff checkpoint coordinator. `Some` only when
```
dmichelin: nit: same here

dmichelin (approved):
This looks like it's just turning on some already reviewed code, just some minor changes

PR #14615, opened 2026-07-31
Validate TUI startup API keys before entering signed-in state

moirahuang:
## Description

Interactive Warp Agent CLI startup API keys could previously enter shared `AuthState` before the server had validated them or returned the associated user. Credential presence made `AuthState::is_logged_in()` true and allowed unrelated clients to use the key while identity loading was still pending. If validation then failed, the TUI could create a terminal and render a bare **Signed in** state with no validated identity.

This change fixes both parts of that lifecycle:
- `LaunchMode` now owns the decision that interactive TUI startup keys require deferred validation;
- interactive TUI startup keys remain pending outside shared `AuthState`;
- the pending key is used only by `AuthClient::fetch_user(LoginToken::ApiKey(...), false)`, after staging IAP access is ready;
- successful validation atomically promotes the returned credentials and user through the existing authentication-completion path;
- failed validation leaves shared auth state with neither credentials nor a user and remains in the retryable login flow;
- an explicitly supplied TUI key still takes precedence over persisted identity without exposing the unvalidated key to other clients;
- command-line and desktop-app API-key initialization behavior is unchanged;
- the TUI requires credentials plus a fetched user ID before entering its logged-in phase or creating a terminal; and
- account labels fall back through email, username, and stable user ID, otherwise reporting **Not signed in** instead of bare **Signed in**.

The SSH environment exposed the partial state but was not the root cause. API keys are not persisted by Warp; a key present at startup comes from the launch arguments or `WARP_API_KEY` environment inherited by that process.

## Linked Issue

N/A — this was identified from logs for an SSH-launched Warp Agent CLI session.

## Testing

Added regression coverage for:
- distinct desktop-app and interactive-TUI startup-key initialization policy;
- pending API-key normalization and validation with `for_refresh = false`;
- failed validation emitting `AuthFailed` while leaving shared auth state fully logged out;
- successful promotion installing the API-key credentials together with the fetched user;
- credential-only startup, validated identity startup, zero-state labels, stable user-ID fallback, and `/status` identity resolution.

Manual full-terminal validation:
- Ran `WARP_API_KEY=wk-invalid-pending-key ./script/run-tui` in a real PTY.
- The TUI built and launched successfully.
- The invalid key was rejected and displayed the login failure/retry flow.
- The TUI never entered a signed-in state and did not create a shell terminal.

Previously completed for the original TUI state-gating patch:
- App TUI authentication module: 17 tests passed.
- Full `warp_tui` suite: 966 tests passed.
- `./script/format` passed.
- Relevant workspace, app, and completer Clippy commands passed.

No additional automated suite, Clippy, presubmit, or formatting run was performed after the pending-key lifecycle follow-up; validation of that follow-up used the real TUI terminal flow above.

No process-level integration test was added because the headless TUI does not use the GUI integration harness; state-transition and render-to-lines unit tests cover the regression directly.

## Agent Mode
- Agent conversation: https://staging.warp.dev/conversation/f0184211-8c6c-4695-8ca1-6b0972449c78

CHANGELOG-BUG-FIX: Fixed Warp Agent CLI treating startup API keys as authenticated before validation completed.

### DIFF
--- a/app/src/lib.rs
+++ b/app/src/lib.rs
@@ -451,6 +451,42 @@ impl LaunchMode {
         }
     }
 
+    fn api_key(&self) -> Option<String> {
+        match self {
+            LaunchMode::CommandLine { global_options, .. } => global_options.api_key.clone(),
+            LaunchMode::App { api_key, .. }
+            | LaunchMode::Tui {
+                entrypoint: TuiEntryPoint::Interactive { api_key, .. },
+            } => api_key.clone(),
+            LaunchMode::Test { .. }
+            | LaunchMode::RemoteServerProxy
+            | LaunchMode::RemoteServerDaemon { .. }
+            | LaunchMode::Tui {
+                entrypoint: TuiEntryPoint::CliCommand { .. },
+            } => None,
+        }
+    }
+
+    /// Returns whether a startup API key should be installed before its user is fetched.
+    ///
+    /// The interactive TUI defers the key so its UI cannot treat credential presence as a
+    /// validated identity. Other launch modes retain their existing initialization behavior.
+    fn should_initialize_api_key_eagerly(&self) -> bool {
+        match self {
+            LaunchMode::Tui {
+                entrypoint: TuiEntryPoint::Interactive { .. },
+            } => false,
+            LaunchMode::App { .. }
+            | LaunchMode::CommandLine { .. }
+            | LaunchMode::Test { .. }
+            | LaunchMode::RemoteServerProxy
+            | LaunchMode::RemoteServerDaemon { .. }
+            | LaunchMode::Tui {
+                entrypoint: TuiEntryPoint::CliCommand { .. },
+            } => true,
+        }
+    }
+
     /// Returns `true` if this process is running an integration test.
     fn is_integration_test(&self) -> bool {
         match self {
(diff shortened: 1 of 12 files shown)


moirahuang:
* **#14615** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_server_auth/src/auth_state.rs
```diff
@@ -176,6 +176,30 @@ impl AuthState {
         state
     }
 
+    /// Creates auth state for an interactive client that still needs to validate
+    /// its startup API key.
+    ///
+    /// The pending key intentionally remains outside [`AuthState`] so it cannot
```
moirahuang: this is to address previously used API keys leading to thinking the user is logged in

app/src/lib.rs
```diff
@@ -1421,10 +1450,14 @@ pub(crate) fn initialize_app(
         ctx.set_zoom_factor(WindowSettings::as_ref(ctx).zoom_level.as_zoom_factor());
     }
 
-    // Extract API key from command line options, if applicable.
-    let api_key = api_key_from_launch_mode(launch_mode);
-
-    let auth_state = Arc::new(AuthState::initialize(ctx, api_key));
+    // Interactive TUI API keys remain outside shared auth state until the server accepts them.
+    // Other launch modes retain their existing eager credential initialization semantics.
```
moirahuang: calling out that TUI launch mode is different from other launch modes

app/src/lib.rs
```diff
@@ -1421,10 +1450,14 @@ pub(crate) fn initialize_app(
...
+    let pending_tui_api_key = pending_tui_api_key_from_launch_mode(launch_mode);
```
kevinyang372: Do we want to leak the tui concept into the auth state here?

I think it might be better for launch_mode to implement a `LaunchMode::should_initialize_credential_eagerly` method?

crates/warp_server_auth/src/auth_state.rs
```diff
@@ -176,6 +176,30 @@ impl AuthState {
...
+    /// authorize unrelated requests or make [`Self::is_logged_in`] true before
+    /// the user fetch succeeds. Persisted user state is also skipped because an
+    /// explicitly supplied API key takes precedence over secure storage.
+    pub fn initialize_for_pending_api_key(ctx: &AppContext) -> Self {
```
kevinyang372: Is the main difference here we don't set credentials before user fetch?

kevinyang372 (approved):

crates/warp_server_auth/src/auth_state.rs
moirahuang: yeah! i'm now wondering if this same pattern is an issue for the GUI? i'm thinking to merge this for now but can add a linear ticket to follow up on the GUI / think about a more robust pattern

moirahuang: https://linear.app/warpdotdev/issue/APP-5097/look-at-gui-auth-state-being-out-of-date-related-to-tui

app/src/lib.rs
moirahuang: yeah that's fair, updated

PR #14624, opened 2026-08-01
Save and restore tab groups in launch configurations

### LINKED ISSUE #13898: Save and Launch Configuration Template doesn't store Tab groups, all tabs are flat and they are not grouped

maxmilian:
Fixes #13898

## Why

Saving a launch config from a window with grouped tabs and reopening it gives you flat tabs — the groups are gone. `impl From<WindowSnapshot> for WindowTemplate` reads `snapshot.tabs` and `snapshot.active_tab_index` and never looks at `snapshot.tab_groups`, and `WindowTemplate` had nowhere to put them.

Worth noting because it narrows the fix: **this is a gap in one path, not a missing capability.** Session restore already round-trips groups end to end — `persistence/sqlite.rs` writes them (`:1012-1038`) and reads them back (`:2538-2703`), covered by `test_sqlite_round_trips_tab_groups`. So the data model was already there; only the launch-config projection ignored it.

## What changed

**Schema** (`launch_config.rs`) — `WindowTemplate` gains `tab_groups: Vec<TabGroupTemplate>` and `TabTemplate` gains `group: Option<usize>`, an index into that list.

```yaml
windows:
  - tabs:
      - layout: {cwd: /srv/api}
        group: 0
      - layout: {cwd: /tmp}          # ungrouped, no `group` key
    tab_groups:
      - name: backend
        color: blue
```

Both fields are `skip_serializing_if` + `default`, so **existing configs are unaffected** — a window with no groups serializes byte-for-byte as before. There is a test asserting neither key appears in that case.

Runtime `TabGroupId`s are deliberately *not* serialized. They are UUIDs minted per session, a config can be opened repeatedly, and it can be opened into a workspace that already holds groups — reusing saved ids would collide. The index indirection sidesteps that.

**Restore** (`workspace/view.rs`) — `open_launch_config_window` resolves each tab's membership first, then creates a group for every index that kept a member and assigns the tabs' `group_id`s through it. Groups no tab joins are never inserted, so hand-authored YAML cannot leave an empty, unreachable group in workspace state — the same posture as the save path, which already drops groups whose members were all unsaveable. This mirrors the session-restore path deliberately, including both feature gates: groups are only created under `FeatureFlag::GroupedTabs`, and `pinned` is only honored under `FeatureFlag::PinnedTabs`.

Two shapes only hand-edited YAML can produce are normalized on the way in:

- A group named on both sides of an ungrouped tab. `tab_bar_slots` collapses each *contiguous* run into one container, so interleaved membership would draw two containers sharing one id. `resolve_group_memberships` keeps the first run and returns later stragglers ungrouped; reordering the tabs would also restore the invariant, but silently moving tabs a config explicitly ordered seemed the more surprising of the two.
- An out-of-range `group` index, which leaves that tab ungrouped rather than failing the window — same posture as session restore, which drops a tab's group reference when the group itself did not restore.

## The part that needed care

The save path filters out tabs that cannot be represented (notebook panes) and **renumbers the survivors** via `num_valid_tabs`. So group membership cannot be derived from tab position — the snapshot index and the template index diverge exactly when a tab drops out. Membership is instead carried on each tab's own `group_id` and resolved against the surviving group list.

Two consequences, both locked by tests:

- A group whose only member failed to convert is **dropped**, rather than restored as an empty group the user can neither see nor remove.
- When that happens, later groups shift down, and the remaining tabs' `group` indices shift with them.

`test_config_from_snapshot_remaps_groups_around_unsaveable_tabs` is the one that guards this, and I checked it actually bites rather than trusting the green run: dropping the "keep only groups that still have a member" filter turns it red with

```
assertion `left == right` failed: empty group must be dropped
  left: 2
 right: 1
```

and restoring the filter returns the suite to 14 passed.

## Tests

Three new tests in `launch_config_tests.rs`:

- `test_config_from_snapshot_preserves_tab_groups` — name/color carried, membership correct, ungrouped tabs stay ungrouped
- `test_config_from_snapshot_remaps_groups_around_unsaveable_tabs` — the re-indexing case above
- `test_config_from_snapshot_omits_tab_groups_when_there_are_none` — backward compatibility of the serialized form

The four existing `WindowTemplate` / `TabTemplate` literals in `crates/integration/src/test/launch_configs.rs` were updated for the new fields; no behavior change there.

End-to-end coverage of the restore half lives in `test_launch_config_restores_tab_groups` (`crates/integration/src/test/launch_configs.rs`), which dispatches the real `root_view:open_launch_config` and asserts the restored window through a new `assert_tab_groups` helper. The helper numbers groups by first appearance in the tab list, so it pins down membership *and* identity — two tabs share an index only when they share a `TabGroupId` — then checks each group's restored name and color, and that the workspace holds exactly those groups and no memberless extras.

I checked it bites rather than trusting the green run; asserting the pre-collapse membership fails on exactly the tab the guard covers:

```
Expected tab group memberships [Some(0), Some(0), None, Some(0), Some(1)],
              but there were   [Some(0), Some(0), None, None,    Some(1)]
```

## Manual validation

Restoring a launch config with tab groups

Local `./script/run` OSS build, opened via `warposs://launch/Tab%20Groups%20Demo` so the restore path runs exactly as it would from the launch config palette. The tab bar comes back as `[Backend | api  worker]  scratch  stray  [Frontend | web]`: both groups restore with their names and blue/green colors, each tab lands back in its saved group, and `stray` — which asked for `Backend` again after an ungrouped tab — comes back ungrouped rather than opening a second container sharing the same id.

(The GIF is hosted on a branch in my fork because I have no way to attach media to a PR from the CLI; happy to re-upload it wherever you'd prefer.)

`cargo fmt -- --check` is clean, `cargo check -p warp --all-targets` is clean, `cargo test -p warp --lib launch_config` is 18 passed / 0 failed, and the integration test above is 1 passed / 0 failed.

An earlier comment of mine on this PR claimed `GroupedTabs` could not be enabled in a local build. That was wrong — `grouped_tabs` is in `default` in `app/Cargo.toml` and `enabled_features()` turns it on at startup — and the recording above is the correction.

CHANGELOG-BUG-FIX: Add tab group persistence to (legacy) launch configs.

### DIFF
--- a/app/src/integration_testing/workspace/assertions.rs
+++ b/app/src/integration_testing/workspace/assertions.rs
@@ -1,13 +1,88 @@
-use warpui::async_assert_eq;
 use warpui::integration::AssertionCallback;
+use warpui::{async_assert, async_assert_eq};
 
 use crate::integration_testing::view_getters::workspace_view;
+use crate::tab::SelectedTabColor;
+use crate::themes::theme::AnsiColorIdentifier;
+use crate::workspace::tab_group::TabGroupId;
 
 pub fn assert_focused_tab_index(tab_index: usize) -> AssertionCallback {
     Box::new(move |app, window_id| {
         let workspace = workspace_view(app, window_id);
         workspace.read(app, |view, _ctx| {
             async_assert_eq!(view.active_tab_index(), tab_index)
+        })
+    })
+}
+
+/// Assert how the workspace's tabs are grouped, along with the name and color
+/// of each group.
+///
+/// `expected_memberships` holds one entry per tab, in tab order: `None` for an
+/// ungrouped tab, or an index into `expected_groups`. Groups are numbered by
+/// the order they first appear in the tab list, so this also pins down which
+/// tabs share a single group -- two tabs only get the same index when they
+/// carry the same `TabGroupId`.
+///
+/// Group colors are given the way a launch config writes them -- `None` for a
+/// group that saved no color -- and are compared against the `SelectedTabColor`
+/// restore is expected to produce.
+pub fn assert_tab_groups(
+    expected_memberships: Vec<Option<usize>>,
+    expected_groups: Vec<(Option<&'static str>, Option<AnsiColorIdentifier>)>,
+) -> AssertionCallback {
+    Box::new(move |app, window_id| {
+        let workspace = workspace_view(app, window_id);
+        workspace.read(app, |view, _ctx| {
+            let mut group_order: Vec<TabGroupId> = Vec::new();
+            let memberships: Vec<Option<usize>> = view
+                .tabs
+                .iter()
+                .map(|tab| {
+                    tab.group_id.map(|id| {
+                        group_order
+                            .iter()
+                            .position(|seen| *seen == id)
+                            .unwrap_or_else(|| {
+                                group_order.push(id);
+                                group_order.len() - 1
+                            })
+                    })
+                })
+                .collect();
+
+            if memberships != expected_memberships {
+                return async_assert!(
+                    false,
+                    "Expected tab group memberships {expected_memberships:?}, but there were {memberships:?}"
+                );
+            }
+
+            let groups: Vec<(Option<String>, SelectedTabColor)> = group_order
+                .iter()
+                .map(|id| {
+                    let group = view
+                        .tab_groups
+                        .get(id)
+                        .expect("A grouped tab's group must exist in the workspace");
+                    (group.name.clone(), group.color)
+                })
+                .collect();
+            let expected: Vec<(Option<String>, SelectedTabColor)> = expected_groups
+                .iter()
+                .map(|&(name, color)| {
+                    (
+                        name.map(str::to_owned),
+                        color.map_or(SelectedTabColor::Unset, SelectedTabColor::Color),
+                    )
+                })
+                .collect();
+
+            async_assert_eq!(
+                groups,
+                expected,
+                "Expected restored groups {expected:?}, but there were {groups:?}"
+            )
         })
     })
 }
--- a/app/src/integration_testing/workspace/assertions.rs
+++ b/app/src/integration_testing/workspace/assertions.rs
@@ -27,6 +27,9 @@ pub fn assert_focused_tab_index(tab_index: usize) -> AssertionCallback {
 /// Group colors are given the way a launch config writes them -- `None` for a
 /// group that saved no color -- and are compared against the `SelectedTabColor`
 /// restore is expected to produce.
+///
+/// The workspace must hold exactly the groups listed, so a group that no tab
+/// belongs to is a failure rather than something the caller can omit.
 pub fn assert_tab_groups(
     expected_memberships: Vec<Option<usize>>,
     expected_groups: Vec<(Option<&'static str>, Option<AnsiColorIdentifier>)>,
@@ -55,6 +58,16 @@ pub fn assert_tab_groups(
                 return async_assert!(
                     false,
                     "Expected tab group memberships {expected_memberships:?}, but there were {memberships:?}"
+                );
+            }
+
+            if view.tab_groups.len() != group_order.len() {
+                return async_assert!(
+                    false,
+                    "Expected the workspace to hold exactly {} group(s), the ones its tabs \
+                     belong to, but it holds {} -- the extras have no members",
+                    group_order.len(),
+                    view.tab_groups.len()
                 );
             }
 
--- a/app/src/launch_configs/launch_config.rs
+++ b/app/src/launch_configs/launch_config.rs
@@ -131,6 +131,42 @@ fn is_false(val: &bool) -> bool {
     !*val
 }
 
+/// Resolves each tab's group index for restore, keeping every group to a
+/// single contiguous run.
+///
+/// The tab bar collapses each *contiguous* run of same-group tabs into one
+/// group container (`Workspace::tab_bar_slots`), so interleaved membership --
+/// group 0, an ungrouped tab, group 0 again -- would render as two containers
+/// sharing one id, which no other code path can produce. Configs written by
+/// `From<WindowSnapshot>` are always contiguous because a live window is, so
+/// this only bites on hand-edited YAML.
+///
+/// The first run of each group wins and later stragglers come back ungrouped.
+/// Reordering the tabs would also restore the invariant, but silently moving
+/// tabs the config explicitly ordered is the more surprising of the two.
+/// Out-of-range indices are dropped the same way.
+pub fn resolve_group_memberships(tabs: &[TabTemplate], group_count: usize) -> Vec<Option<usize>> {
+    let mut closed: Vec<bool> = vec![false; group_count];
+    let mut previous: Option<usize> = None;
+
+    tabs.iter()
+        .map(|tab| {
+            let group = tab
+                .group
+                .filter(|index| *index < group_count)
+                .filter(|index| !closed[*index]);
+
+            if previous != group
+                && let Some(previous) = previous
+            {
+                closed[previous] = true;
+            }
+            previous = group;
+            group
+        })
+        .collect()
+}
+
 fn is_falsey(val: &Option<bool>) -> bool {
     val.is_none_or(|v| !v)
 }
--- a/app/src/launch_configs/launch_config.rs
+++ b/app/src/launch_configs/launch_config.rs
@@ -3,8 +3,8 @@ use std::path::PathBuf;
 use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::app_state::{
-    AppState, LeafContents, PaneNodeSnapshot, SplitDirection as StateSplitDirection, TabSnapshot,
-    WindowSnapshot,
+    AppState, LeafContents, PaneNodeSnapshot, SplitDirection as StateSplitDirection,
+    TabGroupSnapshot, TabSnapshot, WindowSnapshot,
 };
 use crate::themes::theme::AnsiColorIdentifier;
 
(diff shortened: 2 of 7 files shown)


maxmilian:
Thanks — the first concern was real, and I had missed it. Pushed `b9dfb3a`.

## Split group memberships

Confirmed against `tab_bar_slots`: it opens a fresh `TabBarSlot::Group` whenever a run breaks, so `group: 0`, ungrouped, `group: 0` really would draw two containers sharing one id. Restore was taking membership straight from the config, so hand-edited YAML could reach a state no other path can.

Restore now keeps each group to its first contiguous run and returns later stragglers ungrouped. I considered reordering the tabs instead — it also restores the invariant — but silently moving tabs a config explicitly ordered seemed the more surprising of the two; happy to switch if you disagree.

The resolution moved out of `open_launch_config_window`'s closure into `resolve_group_memberships`, which made it reachable from unit tests (it needed a `ViewContext` before). Four new tests: contiguous runs preserved, a run split by an ungrouped tab, a run split by another group, and out-of-range indices. I checked they bite rather than trusting the green run — dropping the guard fails two of them:

```
test_resolve_group_memberships_ungroups_a_split_run
  left:  [Some(0), None, Some(0)]
 right:  [Some(0), None, None]
```

`cargo test -p warp --lib launch_config` is 18 passed / 0 failed on the new head.

## Visual evidence

I want to give you a straight answer rather than a promise I cannot keep: **I do not think an external contributor can produce this recording**, because `GroupedTabs` cannot be turned on in a locally built app. Every path is closed from outside:

- `overrides::*` is compiled out unless `feature = "test-util"`; a normal build returns `None`
- `set_user_preference` has no callers anywhere in the tree
- the runtime flag menu only exposes `RUNTIME_FEATURE_FLAGS`, which is `[LocalClaudeCodexChildHarnesses]`
- `FLAG_STATES` comes from server-side delivery

So the restore half is unreachable in my build, which is exactly why the PR description said it was unverified rather than quietly claiming otherwise.

What I can offer instead, and would rather do: an **integration test** in `crates/integration/src/test/launch_configs.rs`. That harness dispatches the real `root_view:open_launch_config` action, and integration tests may call `FeatureFlag::GroupedTabs.set_enabled(true)` (the pattern in `session_restoration.rs:224`). It would cover the same behavior the recording would show — groups restored with names and colors, membership assigned, the split case collapsed — and unlike a screenshot it keeps covering it.

Tell me which you'd prefer:

1. I add that integration test to this PR, or
2. you have a way to enable `GroupedTabs` in a local build that I've missed — point me at it and I'll capture the recording, or
3. someone with the flag verifies the restore half internally and I leave the code as is.

Happy with any of the three; I just can't do the recording unaided.

maxmilian:
## Correction: I was wrong about the feature flag

My last comment argued at length that an external contributor cannot enable `GroupedTabs` in a local build. That is wrong, and I would rather retract it plainly than leave it standing.

`grouped_tabs` sits in `default` in `app/Cargo.toml`, and `enabled_features()` in `app/src/features.rs` flips on every cfg-gated flag in that list at startup:

```rust
#[cfg(feature = "grouped_tabs")]
FeatureFlag::GroupedTabs,
```

so a plain `./script/run` build has it enabled. It moved to default in `1a5d135c` ("Switch tab group feature flag to stable", #13230). The four closed paths I listed were each true on their own; I missed the startup list, which is the one that decides it. Sorry for the noise — the restore half was reachable all along.

## Recording

Restoring a launch config with tab groups

Local `./script/run` OSS build, opened with `warposs://launch/Tab%20Groups%20Demo` so the restore path runs exactly as it would from the launch config palette. (The GIF lives on a branch in my fork — I have no way to attach media to a comment from the CLI. Happy to re-upload it wherever you'd prefer.) Config used:

```yaml
name: Tab Groups Demo
windows:
  - active_tab_index: 0
    tab_groups:
      - name: Backend
        color: blue
      - name: Frontend
        color: green
    tabs:
      - { title: api,     group: 0, layout: { cwd: /Users/maxmilian/side } }
      - { title: worker,  group: 0, layout: { cwd: /Users/maxmilian/side } }
      - { title: scratch,           layout: { cwd: /Users/maxmilian/side } }
      - { title: stray,   group: 0, layout: { cwd: /Users/maxmilian/side } }
      - { title: web,     group: 1, layout: { cwd: /Users/maxmilian/side } }
```

The tab bar comes back as `[Backend | api  worker]  scratch  stray  [Frontend | web]`:

- **Backend** and **Frontend** restore with their names and their blue/green colors, and each tab lands back in the group it was saved under.
- `stray` asked for `Backend` again *after* an ungrouped tab. That is the hand-edited-YAML case from your first concern — restore keeps `Backend` to its first contiguous run and hands `stray` back ungrouped, rather than drawing two containers that share one id.

## Integration test

The recording is a snapshot, so I also added end-to-end coverage of the same behavior — `CONTRIBUTING.md` asks for it on user-facing flows, and unlike a recording it keeps checking:

- `test_launch_config_restores_tab_groups` in `crates/integration/src/test/launch_configs.rs` dispatches the real `root_view:open_launch_config` with the config above and asserts the restored layout.
- `assert_tab_groups` in `app/src/integration_testing/workspace/assertions.rs` numbers groups by first appearance in the tab list, so it pins down membership *and* identity — two tabs only share an index when they share a `TabGroupId` — then checks each group's restored name and color.
- Registered in both `crates/integration/src/bin/integration.rs` and `crates/integration/tests/integration/ui_tests.rs`.

`cargo test -p integration --test integration -- ui_tests::test_launch_config_restores_tab_groups --exact` is 1 passed / 0 failed. I checked it bites rather than trusting the green run — asserting the pre-fix membership instead fails on exactly the tab the guard covers:

```
Expected tab group memberships [Some(0), Some(0), None, Some(0), Some(1)],
              but there were   [Some(0), Some(0), None, None,    Some(1)]
```

maxmilian:
Both concerns addressed in `a92b9b4c`.

**Empty groups** — you're right, and it was a real gap. `open_launch_config_window` inserted a `TabGroup` for every entry in the config *before* resolving memberships, so a hand-authored config could leave a group nothing points at: invisible in the tab bar and unreachable from any menu. The save path already drops groups whose members were all unsaveable; restore now matches it. Memberships are resolved first, and an id is minted only for a group that kept a member — which also covers the group whose last tab was stripped by the contiguous-run collapse.

`assert_tab_groups` now requires the workspace to hold *exactly* the listed groups, and the integration config gained an `Orphan` group no tab joins. I checked that guard bites; dropping the filter fails with

```
Expected the workspace to hold exactly 2 group(s), the ones its tabs belong to,
but it holds 3 -- the extras have no members
```

**Visual evidence** — this one was my fault for putting it only in a comment. The PR description still carried the old "Known validation limitation" section saying the restore half had not been run in an app, so it contradicted the recording. I've replaced that section with a **Manual validation** section that embeds the recording and states what it demonstrates. The claim it used to make is retracted there too: `grouped_tabs` is in `default` in `app/Cargo.toml` and `enabled_features()` turns `GroupedTabs` on at startup, so the restore half was reachable in a local build all along.

`cargo test -p warp --lib launch_config` is 18 passed / 0 failed and the integration test is 1 passed / 0 failed on the new head.

maxmilian:
Good catch — confirmed and fixed in `4a717f49`.

`NewTabPlacement` defaults to `AfterCurrentTab`, and `add_tab_with_pane_layout` routes through `new_tab_index_and_group`, so a restored tab is only appended when the active tab happens to be the last one. With `open_in_active_window: true` and any tab after the active one, every insert landed ahead of the trailing tabs and `start_index + tab_index` pointed one slot late per trailing tab.

Worth flagging: **the `selected_color` assignment on the line above carried the same assumption before this PR**, so config tab colors were already landing on pre-existing tabs in that case. Both now read the index back from the insert, which activates whatever it placed. Focusing the config's active tab goes through the same recorded indices, so it no longer depends on the restored tabs being contiguous at the end.

One behavior I made explicit rather than left implicit: the insert can have a new tab inherit the active tab's group, but the config is the authority on membership, so a tab the config leaves ungrouped now stays ungrouped. There's a comment on that line saying so.

**Regression test** — `test_launch_config_restores_tab_groups_into_active_window`: open two ungrouped tabs into a new window, leave the first one active, then open a grouped config into that window with `open_in_active_window: true`. No settings plumbing needed, since `AfterCurrentTab` is already the default. It fails on the old arithmetic exactly the way you described:

```
Expected tab group memberships [None, Some(0), Some(0), Some(1), None],
              but there were   [None, None,    None,    None,    Some(0)]
```

— every group landing on the wrong tab, including one on the pre-existing trailing tab.

Green on the new head: the two grouped-restore tests plus `test_open_launch_config_in_active_window` and `test_with_launch_config_with_active_tab_index` (which cover the focus path I touched) are 4 passed / 0 failed, and `cargo test -p warp --lib launch_config` is 18 passed / 0 failed.

bholmesdev (approved):

maxmilian:
Good catch — that one is real, and it is a case the previous fix did not reach: repositioning pinned groups happens *after* membership is assigned, but nothing put the restored block outside a host group in the first place.

`81f6b9ff2` takes the first option you listed. The host group (the active tab's group at the time of the restore) is captured before the inserts, and once membership is settled the restored block is re-anchored past that group's last remaining member. That mirrors `new_tab_group_from_selected_tabs`, which already solves the same problem for the "create group from tabs" action, rather than introducing a second ordering rule.

The move is skipped while the host group is still contiguous — `NewTabPlacement` can legitimately land the block outside the group, and that placement is the user's setting.

New integration test `test_launch_config_restore_keeps_existing_group_contiguous` covers the active-window grouped-tab case: a grouped config is opened into a new window with its *first* member active, then an ungrouped config is opened into that window so the insert lands between the two members. Verified both directions locally — with the fix the memberships are `[Some(0), Some(0), None, None]` and the test passes; reverting the fix alone fails it with `[Some(0), None, None, Some(0)]`, exactly the split you described. `cargo check --all-targets` and `cargo fmt --check` are clean.

/warp-agent-review

PR #14681, opened 2026-08-04
Synthesize a smooth, eased cursor for Linux computer-use recordings (QUALITY-1169 follow-up)

danielpeng2:
## Description

Computer-use video artifacts on Linux had inconsistent, broken pointer rendering. ffmpeg's `x11grab -draw_mouse` composites the cursor via XFixes, which only ever reports the user's *core* pointer — never the private MPX "agent seat" pointer that background (window-scoped) computer use drives. The result: window-scoped recordings showed no moving cursor at all, and screen-scoped recordings showed a frozen cursor that contradicted the burned-in click ripples.

This PR makes the burned-in overlay the single source of pointer truth, for both capture scopes:

- **Native cursor compositing is disabled** (`-draw_mouse 0`) for all Linux captures, and a **synthetic arrow glyph** (white fill, black outline, tip on the hotspot) is burned in from the recorded pointer events, so window- and screen-scoped artifacts render the pointer identically.
- **Motion is polished rather than faithful-but-choppy.** XTEST warps the real pointer instantaneously between action points, so a literal rendering teleports. Instead the glyph holds at each event and **glides into the next one with an eased, ≤500 ms approach that lands exactly when the event fires** — the click ripple ignites the moment the cursor arrives.
- **Drags move continuously.** While the button is held, motion fills the recorded path end-to-end: ease-in out of the press, linear through intermediate waypoints, ease-out into the release point — no pauses at sampled waypoints. The held-indicator dot follows the identical schedule so it stays glued to the cursor.
- **Cut-seam aware.** The motion plan is computed on the compacted (smart-cut) output timeline, so glides and drag legs bridge removed idle gaps smoothly instead of being swallowed by them.
- Scroll actions now record position samples so the cursor tracks the wheel location, and the cursor fades in at its first appearance.

Since ASS has no positional easing, glides are emitted as short piecewise-linear `\move` dialogues sampled from the ease curve, quantized to centiseconds so adjacent pieces stay exactly contiguous.

The QUALITY-1169 spec's behavior invariants are amended accordingly (`agents/specs/QUALITY-1169-richer-click-drag-annotations.md`).

## Linked Issue

Follow-up to [QUALITY-1169](https://linear.app/warpdotdev/issue/QUALITY-1169) (click ripple / drag trail burn-in), which shipped the overlay pass this builds on.

## Testing

[...]

PR #14686, opened 2026-08-04
Persist the X11 agent seat for the background computer-use session (fix split-call drags)

danielpeng2:
## Description

On Linux, window-scoped (background) drags didn't actually drag: the cursor moved, but e.g. drag-to-select never selected any text. Screen-scoped drags were fine.

**Root cause:** each `use_computer` call creates and drops its own `Actor`, and on Linux dropping the actor removed the MPX "agent seat". Removing a master device implicitly releases the buttons it holds. Since the model splits drags across calls (press in one call, moves/release in later ones), the button was dropped at the first call boundary and the rest of the drag became plain hover motion.

**Fix:** the agent seat now lives for the whole background session instead of one call.

- Sessions share one seat, keyed by the owning conversation id, so held buttons (and the agent keyboard's focus) survive between calls.
- The existing `end_background_session()` teardown hook (previously macOS-only) now also removes the Linux seat; it already runs on completion and cancellation. This mirrors how #13752 solved the same problem on macOS.
- The `use_computer` dev CLI and the existing leak-safety (stale-seat reaping) are unchanged.

Trade-off: on a real desktop, the agent's second cursor now stays visible for the session rather than one call. The spec (`specs/x11-background-computer-use/TECH.md`) is amended.

## Linked Issue

Follow-up to QUALITY-1169; found while reviewing drag recordings from #14681.

## Testing

- Reproduced at the X11 level in a container: press on a seat, remove it (what `Actor::drop` did), and the button reads released on every later move; a persistent seat selects text correctly.
- E2E via a local Oz worker: drag-to-select in Chromium now highlights text in both window- and screen-scoped recordings.
- `cargo nextest run -p computer_use` (75/75), format + clippy clean; Linux build verified via the sidecar image.
- [ ] I have manually tested my changes locally with `./script/run` — not applicable: Linux-only paths, exercised via the local worker runs above.

https://github.com/user-attachments/assets/cab590c2-99c2-4f1a-a333-21eb278afb5b

## Agent Mode

CHANGELOG-OZ: Fixed background (window-scoped) computer use on Linux dropping the mouse button between actions, which broke drag gestures like text selection.

### DIFF
--- a/crates/computer_use/src/lib.rs
+++ b/crates/computer_use/src/lib.rs
@@ -114,22 +114,25 @@ pub fn background_supported() -> bool {
 }
 
 /// Ends the background computer-use session owned by `owner` (the client conversation id),
-/// restoring the user's original keyboard focus.
+/// releasing the session state that outlives individual action batches.
 ///
 /// On macOS a background session activates the target window and installs focus-suppression
 /// taps; this tears down only the windows owned by `owner`, deactivates them, and re-activates the
 /// app that was frontmost before the session, so the user's keystrokes return to where they were.
-/// Scoping by owner keeps concurrent background sessions (e.g. another conversation driving a
-/// different window) intact. Idempotent and a no-op when `owner` has no active session, and on
-/// platforms without background per-window control.
+/// On Linux X11 a background session drives a session-scoped agent seat (a second input seat
+/// shared across the session's action batches so state like a held mouse button mid-drag
+/// survives between batches); this removes `owner`'s seat, its on-screen cursor, and any input
+/// state it still holds. Scoping by owner keeps concurrent background sessions (e.g. another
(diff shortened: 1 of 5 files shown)


danielpeng2:
* **#14686** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14681** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/14681?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

vkodithala (approved):
Demo LGTM, thanks for the fix!

PR #14688, opened 2026-08-04
Make vertical tabs search match tab group names

zmanian:
## Description

The vertical tabs panel search field only matched against pane and tab text, so a tab group could not be found by its name. Typing `backend` would surface tabs whose own titles happened to contain it, but not the group named `backend`, and there was no way to pull up everything inside that group.

This makes a query that matches a group's displayed name surface that group and every tab under it, regardless of whether the individual tab titles match.

Two helpers carry the behavior, both in `app/src/workspace/view/vertical_tabs.rs`:

- **`group_display_name()`** returns the header title including the `"New Group"` fallback. The group header renderer now calls it too, so the text being searched and the text on screen cannot drift apart.
- **`merge_group_name_matches()`** folds members of name-matched groups into the filtered results. Two properties are load-bearing: output stays ordered by tab index (the render loop collapses a group into one container by scanning a contiguous run, so an out-of-order entry would split it across containers), and a member already present from its own text match is upgraded to "all panes" rather than duplicated — a group-name match shows whole tabs, not pane-filtered slices of them.

While a query is active, every surviving group renders expanded so matches aren't hidden behind a collapsed header. This mutates only the local clone that `render_groups` already makes, so the persisted `collapsed` flag is untouched and clearing the query restores the real state on the next frame.

Matching uses the *displayed* name, fallback included, so an unnamed group matches the query "new group" — the user is searching for what's on screen.

Behavior is unchanged when the query matches no group name: `merge_group_name_matches` returns the input untouched.

Scope note: this covers the vertical tabs panel search only. The Ctrl+Tab command palette switcher (`app/src/search/command_palette/tabs/`) is a flat MRU list with no group concept; adding group awareness there needs new item types and grouped result rendering, and is left for separate work.

## Linked Issue

None — this is a self-contained improvement to the `GroupedTabs` preview feature.

## Testing

Eight unit tests added to `app/src/workspace/view/vertical_tabs_tests.rs`, following the pure-function style already in that file. Each was written first and watched fail before any implementation existed.

`group_display_name`:
- returns the name when set
- falls back to `"New Group"` when unnamed

`merge_group_name_matches`:
- a group-name match pulls in members whose own titles do not match
- results stay ordered by tab index rather than appended
- a member present as `Some(pane_ids)` is upgraded to all-panes
- a matched group with no members is a no-op
- tabs outside the matched group stay filtered
- no group match leaves the filtered results untouched

```
running 8 tests
test ...::group_display_name_uses_the_group_name_when_set ... ok
test ...::group_display_name_falls_back_to_the_untitled_header_text ... ok
test ...::no_group_name_match_leaves_the_filtered_tabs_untouched ... ok
test ...::group_name_match_for_a_group_with_no_members_changes_nothing ... ok
test ...::group_name_match_keeps_results_ordered_by_tab_index ... ok
test ...::group_name_match_includes_members_that_did_not_match_on_their_own ... ok
test ...::group_name_match_upgrades_a_pane_filtered_member_to_all_panes ... ok
test ...::group_name_match_leaves_tabs_outside_the_group_filtered ... ok

test result: ok. 8 passed; 0 failed
```

Full `vertical_tabs` suite: `68 passed; 0 failed`. `cargo clippy -p warp --lib --all-targets` clean, `cargo fmt --check` clean.

The collapse-during-search override and the end-to-end filter are render-path behavior that the unit test file doesn't reach; those were confirmed by running the app locally with `grouped_tabs` enabled.

### Screenshots / Videos

Not captured — happy to add a short recording of the search-matches-group flow if that's wanted before review.

## Agent Mode

CHANGELOG-IMPROVEMENT: Tab search in the vertical tabs panel now matches tab group names, showing all tabs in a matching group.

### DIFF
--- a/app/src/workspace/view/vertical_tabs.rs
+++ b/app/src/workspace/view/vertical_tabs.rs
@@ -1,7 +1,7 @@
 pub mod telemetry;
 
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 
@@ -1781,7 +1781,7 @@ fn render_groups(
             .collect()
     } else {
         let query_lower = query.to_lowercase();
-        workspace
+        let own_matches: Vec<(usize, Option<Vec<PaneId>>)> = workspace
             .tabs
             .iter()
             .enumerate()
@@ -1882,7 +1882,24 @@ fn render_groups(
                     }
                 }
             })
-            .collect()
+            .collect();
+
+        // A query matching a group's name reveals every tab under that group,
+        // even members whose own text does not match.
+        let matched_groups: HashSet<TabGroupId> = workspace
+            .tab_groups
+            .iter()
+            .filter(|(_, group)| {
+                group_display_name(group)
+                    .to_lowercase()
+                    .contains(&query_lower)
+            })
+            .map(|(group_id, _)| *group_id)
+            .collect();
+        let tab_group_ids: Vec<Option<TabGroupId>> =
+            workspace.tabs.iter().map(|tab| tab.group_id).collect();
+
+        merge_group_name_matches(&tab_group_ids, &matched_groups, own_matches)
     };
 
     if visible_tabs.is_empty() {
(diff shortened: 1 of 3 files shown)


zmanian:
Superseded by a PR from a handle-prefixed branch (`zmanian/tab-search-group-matching`), per the branch-naming convention in CONTRIBUTING.md. The replacement also links the newly-filed issue #14689, includes product and tech specs under `specs/GH14689/`, and attaches manual-testing screenshots.

PR #14693, opened 2026-08-04
Update internal cloud agent dev image

bnavetta:
## Description

This makes a couple updates to our cloud agent dev environment:
1. Explicit support for ARM in addition to x86-64 (this mainly just needed testing)
2. More reliable installation of GCP tooling. Specifically, make sure that we can install Google's keyring even if it's already present. Oddly enough, this caused issues on ARM but not x86-64.
3. Bump the pinned Go version to match what the server currently uses

I also added a script that makes it easier to build and push the image. We're not running it automatically yet - I think it's fine to only do this as needed.

## Testing

Built the image locally for both platforms

## Agent Mode


liliwilson (approved):
Thanks! Daniel and I were just chatting about needing this for a different change, just going to merge this PR to unblock him! 

(His PR: https://github.com/warpdotdev/warp/pull/14714)

PR #14734, opened 2026-08-05
Durable, resumable CLI-agent conversations per project

### LINKED ISSUE #14720: Resume CLI-agent conversations (Claude Code, etc.) per project after restart
## Problem

After a restart (or just closing a window), there is no way to get back to a specific coding-agent conversation from Warp. `CLIAgentSessionsModel` drops a session when the agent exits, taking both the conversation's **name** and its **session id** — so a restored tab that was running `claude` shows a truncated path, and resuming means manually digging the id out of `~/.claude` and typing `claude --resume <id>`.

## Proposal

Make agent tasks durable and resumable, per project:

1. **Durable session handles.** Record `(pane, cwd, agent, session_id)` when a CLI agent starts (SQLite, partial unique indexes for the in-flight/identified states). A tab whose agent exited becomes resumable *in place* — the pane already sits in the right directory, so the resume command (`claude --resume <id>`) prefills there. Requires both pane-uuid AND cwd to match, since panes `cd` away.
2. **Names from the agent's own on-disk state.** The handle store only knows sessions Warp witnessed. For everything else — imported profiles, sessions run in another terminal — scan `~/.claude/projects/<mangled-cwd>/*.jsonl` directly: a transcript's existence *is* the session. Names resolve by priority: last `ai-title` record via a bounded tail read (so `/rename` wins — measured: head-only reads miss 14 of 158 titled transcripts and 100% of renames), then head `ai-title`, then first real user prompt (rejecting injected `<...>` wrappers and `Caveat:` boilerplate), then the slug. All reads are seek-bounded; transcripts are never read whole (multi-GB corpus).
3. **Authority split** (the invariant that keeps this safe): the handle store owns *pane binding* — a scanned session can never fabricate a tab association and always resumes into a fresh tab; the scan owns *names* and the *existence* of unwitnessed sessions. Join key: the session UUID. Resume ids are validated by anchored UUID regex + existence of `<id>.jsonl` in that project's directory.

Incidental fix worth having regardless: the agent-sdk harness's `encode_cwd` replaced only `/` and `.`, but Claude Code mangles **every** non-alphanumeric character. Measured against 47 real project dirs, they disagreed on 18 — session envelopes were written where Claude never reads, silently breaking `claude --resume` for any project path containing `_` or spaces.

Working implementation (fork, flag-gated `FeatureFlag::ResumeProjectTasks`, built on a Projects × Tasks rail layout): https://github.com/samithaj/warp/tree/samithaj/project-tabs-layout

samithaj:
## Description

Closes #14720.

After a restart there is no way back to a specific coding-agent conversation. `CLIAgentSessionsModel` drops a session when the agent exits, taking both its name and its session id — so a restored tab that was running `claude` shows a truncated path, and resuming means digging the id out of `~/.claude` by hand.

Three layers:

- **Durable session handles** — `(pane, cwd, agent, session_id)` recorded at spawn in SQLite, surviving agent exit and app restart. Resume-in-place requires **both** pane-uuid and cwd to match, since panes `cd` away; otherwise a fresh tab opens with the command prefilled (never executed).
- **Transcript discovery** — the handle store only knows sessions Warp witnessed. Sessions run in another terminal, or imported, are found by reading Claude Code's own state: `~/.claude/projects/<mangled-cwd>/*.jsonl`, where a transcript's existence *is* the session. Names resolve tail-first so a `/rename` wins — measured on a real corpus, a head-only read misses 14 of 158 titled transcripts and **100% of renames**. All reads are seek-bounded; transcripts are never read whole.
- **An authority split**, enforced by lookup order rather than a check: the handle store owns pane binding, so a discovered session can never fabricate a tab association and always resumes into a fresh tab; the scan owns names and the existence of unwitnessed sessions. Join key is the session UUID.

**Incidental fix worth having on its own:** `encode_cwd` in the agent-sdk harness replaced only `/` and `.`, but Claude Code mangles *every* non-alphanumeric character. Measured against 47 populated project directories, the two rules disagreed on **18** — meaning session envelopes were being written where Claude never reads, silently breaking `claude --resume` for any project path containing `_` or a space. The triage note on #14720 independently confirmed this.

Behind `FeatureFlag::ResumeProjectTasks`, on top of a Projects × Tasks rail layout gated by `FeatureFlag::Projects`.

## Linked Issue
#14720 — triaged as an enhancement, no duplicates found.

- [x] Where appropriate, screenshots or a short video of the implementation are included below

## Testing

Unit: transcript-naming cascade (tail-rename-wins, head tier, injected-wrapper and `Caveat:` rejection, junk-name rejection, UUID filename filter that excludes the `<uuid>/` subdirectories the project dir also contains), session-scan discovery/ordering/bounding, handle-store CRUD with partial unique indexes, project-layout row assembly.

Manual, end to end: rail rows validated against real transcripts (all sessions in a repo resolved to their true names via the tail tier); resume validated by clicking a dormant row, after agent exit, and after a full app restart; a `/rename` mid-session was picked up.

`cargo clippy -p warp --all-targets -- -D warnings` clean. Rebased onto current master; the one new `CLIAgent::WarpTui` variant is handled explicitly.

## Agent Mode

CHANGELOG-NEW-FEATURE: Coding-agent conversations are now durable per project — after a restart, pick a past Claude Code session from the project rail and resume it where you left off.


samithaj:
Pushed an update with two fixes to this PR's own behaviour, both found by using it.

**Resuming a task now prefills the command instead of running it.** `set_pending_command` reads like "prefill" and is not — a *pending* command means "the user submitted this and it has not reached the shell yet", so `execute_pending_command` fires it on the next `BootstrapPrecmdDone` or `BlockCompleted`. Resuming into a fresh tab therefore always executed on its own. Added `TerminalView::prefill_command`, which inserts the text and arms nothing.

**A session's title no longer leaks to other sessions.** Claude Code's `/rename` broadcasts the new name into the transcripts of concurrently-live sessions: measured here, one rename landed in 13 transcripts across 6 project directories, which made the rail show one name under six different projects. The records carry the *file's own* sessionId, so validating that cannot reject them.

Two changes came out of it. `agentName` is gone as a naming source — it was assumed to mirror `aiTitle` and does not (one file held "Set up UAT test data for customer grade scenarios" in `aiTitle` and a different session's name in `agentName` simultaneously), and it was contaminated in 13/13. And `aiTitle` is now resolved by **uniqueness rather than position**: take the last title when no other session claims it, else the first on the same condition, else the session's own first user prompt, which is written once at session start and cannot be cross-written.

Preferring the first title was the tempting fix, since the first was correct in all 13 — but across 159 titled transcripts, of 44 changed titles only 15 changed to a shared value while **29 changed to a unique one**, i.e. genuine renames like `project-rail-task-status`. That fix would have destroyed twice what it repaired.

Also: a restored tab that has never been opened has no focus state, so `active_session_path` returned `None` and its stored agent handle was invisible. It now falls back to the retained startup directory.

Issue #369, opened 2021-11-06
DUP: Duplicate of #270 - Dank Mono font


waxidiotic:
### Describe the solution you'd like?

```shell
It looks like Warp supports any monospace fonts I have installed, but I guess my monospace font of choice, Dank Mono - doesn't register as a monospace font in macOS.

This obviously may not be on Warp to fix but any ideas on what I could do to get Dank Mono into Warp would be greatly appreciated.
```

elviskahoro:
@waxidiotic ->  I thought changed the font loading to include fonts that aren't explicitly labelled as mono too. @alokedesai

elviskahoro:
Linking:
- #270

PR #14777, opened 2026-08-06
Collect LRC liveness signals so the agent stops killing healthy commands

evelyn-with-warp:
## Description

In some tbench traces with LRC, the CLI subagent judges a long-running command's liveness from the terminal grid alone, so a command that redirects to a file, suppresses output, or computes silently looks identical to a hang and gets Ctrl-C'd

This PR adds an `LrcActivityMonitor` that samples the command's process activity every 5 seconds (CPU delta, I/O write delta, live-process count, aggregate state) and attaches it to LRC snapshots. On the wire the payload is `LongRunningShellCommandActivity`: `since_last_activity` as a `google.protobuf.Duration` plus a `process` tier whose `state` is an enum.

All reading of the OS process table lives in `execute/lrc_activity/sampler.rs`, which is the only place `sysinfo` is used. Pids cross that boundary as plain `u32`, so the monitor's accounting stays target-agnostic and wasm — which has no local process table — gets a sampler that observes nothing.

## Two design decisions to review

- **The foreground pgid is sampled by the monitor, not by `TerminalAttributesPoller`.** The existing fd poller is the obvious source, but it only runs during password-prompt detection: gated on notification settings, never started on the TUI surface, and stopped after its first positive reading — so it would essentially never be publishing during an LRC. The monitor calls `tcgetpgrp` on its own sampler instead, using the new `ShellProcessInfo` on `TerminalModel`. ([`foreground_pgid`](https://github.com/warpdotdev/warp/blob/f911ee51e5/app/src/ai/blocklist/action_model/execute/lrc_activity/sampler.rs#L212-L233), [`ShellProcessInfo`](https://github.com/warpdotdev/warp/blob/f911ee51e5/app/src/terminal/model/terminal_model.rs#L535-L551))

- **The activity payload is withheld until the tree has actually been sampled.** Blocks register lazily on their first `report()`, so the first snapshot is built before the sampler has run; emitting an all-zero-but-present reading there would claim "0 live processes, 0 CPU" about a healthy command at exactly the poll where cancellation is decided. `take_report` returns `None` until a real sample exists, so that snapshot simply carries no `activity` section — an all-zero tree observed *later* is still reported, since "the tree exited" is a real answer. (Per review on the proto PR, absence of the payload now carries this meaning; the earlier in-band `signals_unavailable` marker is gone.) ([`take_report`](https://github.com/warpdotdev/warp/blob/f911ee51e5/app/src/ai/blocklist/action_model/execute/lrc_activity.rs#L253-L276))

## Incidental: `uint32` → `uint64` token counters

Re-pinning the proto past warpdotdev/warp-proto-apis#353 unavoidably picks up warpdotdev/warp-proto-apis#369, which widens `ModelTokenUsage.total_tokens` and the `token_usage_by_category` values from `uint32` to `uint64`. Master had not absorbed that yet, so the adaptation rides along here.

It is handled **at the proto boundary only** — the persisted and displayed counters stay `u32`. Widening those would ripple through ~9 files including view rendering and a large number of tests, which belongs with #369's own migration rather than this PR.

- Outbound (`crates/persistence/src/model.rs`): lossless `u64::from(...)`. `to_proto_combined` now sums in `u64`, which also removes a latent overflow in the three-way add.
- Inbound (`app/src/ai/agent/conversation.rs`): a `narrow_token_count` helper saturating via `u32::try_from(..).unwrap_or(u32::MAX)`, and the accumulations use `saturating_add`. Saturating rather than `unwrap_or_default()` so an oversized count cannot surface as a plausible-looking small total. This matches the existing convention in `get_conversation_usage.rs`, which already narrows with `u32::try_from`.

## Linked Issue
- [ ] No linked issue — came out of a staging trace investigation (`d2fc50d9`, `f6d8b53a`, `0af93eee`).
- [ ] N/A — no UI surface changes, so no screenshots.

## Testing
- [ ] Not manually run via `./script/run` — end-to-end behavior needs the server half, which is gated off in prod (see Rollout).

Unit coverage, all passing at the current tip (`f911ee51e5`):
- `cargo test -p warp --lib lrc_activity` — 19 tests: activity detection via CPU/IO/pid churn, per-report reset, no report until first sample, exited-tree reporting, remote gating.
- `cargo test -p ai action_result` — 12 tests: proto conversion incl. Duration round-trip and enum-state mapping.
- `cargo test -p warp --lib conversation::` (64) and `cargo test -p persistence` (26) — cover the token-counter conversions above.

Also clean: `cargo check -p warp --all-targets`, `cargo fmt --check`, `cargo clippy -p warp --all-targets -- -D warnings`, and the wasm gate `cargo clippy --locked --target wasm32-unknown-unknown -p warp --lib -- -D warnings`.

End-to-end replay evals (file-redirect, silent compute, genuine hang) are still outstanding.

## Rollout

Proto dependency is unblocked: warpdotdev/warp-proto-apis#353 has merged, and `Cargo.toml:352` now pins the resulting `main` commit `2bd5df2e3a73ada980d1e9c613ce369232e8c41d` (previously the PR-branch rev `65ff6b9`). Server counterpart: warpdotdev/warp-server#13955.

Client-side sampling is gated on `FeatureFlag::LrcActivitySignal` (dogfood-on), and reported as unsupported on Windows (no controlling-terminal process group to narrow the tree with) and on wasm (no local process table). The wire payload is additive with every field optional, and the server half is behind `lrc_activity_signals` (off in prod).

## Agent Mode


app/src/ai/blocklist/action_model/execute/lrc_activity.rs
```diff
@@ -0,0 +1,658 @@
...
+        let files = self
+            .files
+            .iter_mut()
+            .filter_map(|file| {
+                let size = file.size?;
+                let activity = LrcFileActivity {
+                    path: file.path.to_string_lossy().into_owned(),
+                    size_bytes: size,
+                    size_delta_bytes: size as i64 - file.size_at_last_report as i64,
+                    tail: read_tail(&file.path),
```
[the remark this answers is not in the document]
evelyn-with-warp: the file is LRC log , instead of user input, which shouldn't have sensitive user info here

app/src/ai/blocklist/action_model/execute/lrc_activity.rs
```diff
@@ -0,0 +1,441 @@
...
+/// The processes belonging to the command the shell is currently running.
+///
+/// Every process in the job descends from the shell, so the descendant set is
+/// the outer bound. When the pty's foreground process group is known it narrows
+/// that set to just the job in the foreground, which matters when a shell has
+/// other children (background jobs, its own helpers). Pipeline members are
+/// siblings rather than descendants of the group leader, so the group is matched
+/// per process rather than by walking down from the leader.
+fn command_process_tree(system: &System, shell_pid: Pid, foreground_pgid: Option<u32>) -> Vec<Pid> {
+    let descendants = descendants_of(system, shell_pid);
```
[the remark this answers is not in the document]
evelyn-with-warp: Good catch — confirmed and fixed.

The trace was as you describe: `descendants_of` excludes the shell pid, so a builtin-only command left `command_process_tree` empty, `collect_process_sample` still returned `Some` with `per_pid: []`, and `take_report` emitted a *present* process tier with `live_process_count: 0` and `signals_unavailable: false`. That is the encoding for an exited tree, so the busy command was reported as the strongest available evidence that nothing was running.

`command_process_tree` now compares the shell's own process group against the pty's foreground group and includes the shell when the two match, which is exactly the case where a builtin or shell function is the foreground job. When an external job holds the terminal the groups differ and behavior is unchanged, so the genuine "tree exited" reading is preserved. Added `the_shell_joins_the_tree_when_it_holds_the_terminal` and `the_shell_stays_out_of_the_tree_when_another_job_holds_the_terminal` to cover both directions.

Windows still falls back to descendants-only, since `foreground_pgid` has no controlling-terminal group to read there.

app/src/ai/blocklist/action_model/execute/lrc_activity.rs
```diff
@@ -0,0 +1,441 @@
...
+    fn collect_process_sample(&self, shell: Option<&ShellProcessInfo>) -> Option<ProcessSample> {
+        let shell = shell?;
+        let shell_pid = Pid::from_u32(shell.pid);
+
+        let mut state = self.state.lock();
+        state.system.refresh_processes_specifics(
+            ProcessesToUpdate::All,
+            true, /* remove_dead_processes */
+            ProcessRefreshKind::nothing().with_cpu().with_disk_usage(),
+        );
```
danielpeng2: This applies the expensive refresh kind (`with_cpu().with_disk_usage()`) to the **entire process table** at 1 Hz, when we only care about the command's tree. Could we split discovery from measurement?

```rust
// Phase A — discovery: enumerate the table, sample nothing.
state.system.refresh_processes_specifics(
    ProcessesToUpdate::All,
    true,
    ProcessRefreshKind::nothing(),
);
let tree = command_process_tree(&state.system, shell_pid, foreground_pgid(shell));

// Phase B — measurement: expensive kind, only for the tree.
state.system.refresh_processes_specifics(
    ProcessesToUpdate::Some(&tree),
    true,
    ProcessRefreshKind::nothing().with_cpu().with_disk_usage(),
);
```

That makes the expensive work O(command tree) instead of O(all processes), and matches the existing pattern: `SystemInfo` only ever measures scoped pids, and `all_processes_refresh_kind()` deliberately samples nothing when enumerating the full table. The delta logic is already compatible since it uses cumulative `accumulated_cpu_time()`.

This is also a **Windows stability** concern, not just perf: see the comment on `SystemInfo::all_processes_refresh_kind` (`app/src/system/info.rs`) — per-process CPU sampling across the full table fires an IPI per pid and can trip the DPC watchdog and bugcheck high-core-count machines. That's why the only existing full-table scan in the codebase runs once per process lifetime and samples nothing. Nothing gates this path off Windows today (only `foreground_pgid`/`getpgid` are unix-only). The two-phase split resolves that structurally, since CPU/disk are only ever sampled for tree members.

app/src/ai/blocklist/action_model/execute/lrc_activity.rs
```diff
@@ -0,0 +1,441 @@
...
+use parking_lot::{FairMutex, Mutex};
+use sysinfo::{Pid, ProcessRefreshKind, ProcessStatus, ProcessesToUpdate, System};
+
+use crate::ai::agent::{LrcActivity, LrcProcessActivity, LrcProcessState};
+use crate::terminal::TerminalModel;
+use crate::terminal::model::block::BlockId;
+use crate::terminal::model::terminal_model::ShellProcessInfo;
+
+/// How often liveness signals are sampled while a monitored command is active.
+pub const SAMPLE_INTERVAL: Duration = Duration::from_secs(1);
```
danielpeng2: How'd we come to the 1s interval? We have an existing process sample that runs on the Warp process and it runs every 5s. I'd feel a bit better about following the existing precedent as to not regress anything performance-wise. Also, LRC snapshots don't come that frequently, so I think the extra time won't affect quality much.

app/src/ai/blocklist/action_model/execute/lrc_activity.rs
```diff
@@ -0,0 +1,441 @@
...
+        state.sampler_running = keep_sampling;
+        keep_sampling
+    }
+
+    /// Refreshes process information and summarizes the command's process tree.
+    fn collect_process_sample(&self, shell: Option<&ShellProcessInfo>) -> Option<ProcessSample> {
+        let shell = shell?;
+        let shell_pid = Pid::from_u32(shell.pid);
+
+        let mut state = self.state.lock();
```
danielpeng2: The monitor lock is held across the entire refresh + tree walk + `getpgid` syscalls here, and `report()` takes this same lock **while the terminal model `FairMutex` is held**. So worst case, a snapshot build stalls the terminal lock for a full tick's duration.

With the refresh scoped down this is probably cheap enough to be moot, but shrinking the lock scope (refresh into locals, then lock briefly to fold) — or a comment explaining why the bound is acceptable — would make the terminal-lock exposure explicit.

app/src/ai/blocklist/action_model/execute/lrc_activity.rs
```diff
@@ -0,0 +1,441 @@
...
+        if let Some(sample) = process_sample {
+            for block_id in live {
+                if let Some(activity) = state.blocks.get_mut(&block_id) {
+                    activity.apply_sample(sample.clone(), now);
+                }
+            }
+        }
+
+        let keep_sampling = !state.blocks.is_empty() || state.armed_actions > 0;
+        state.sampler_running = keep_sampling;
```
danielpeng2, line 195: Nit: once `keep_sampling` goes false, `state.system` retains a full process-table snapshot for the lifetime of the executor. Worth clearing it here (e.g. `state.system = System::new()`) so we only hold that memory while actively monitoring.

danielpeng2 (approved):

danielpeng2 (changes requested):
Sorry, I didn't mean to approve this yet. Can we make sure these changes are gated behind a feature flag?

Another thing we should do is make sure these are tested on all platforms (MacOS, Windows, Linux) before enabling in prod, since the process sampling code will work differently on different platforms. There should be folks on the team with Windows and Linux laptops that we can ask to sanity check.

evelyn-with-warp:
> Sorry, I didn't mean to approve this yet. Can we make sure these changes are gated behind a feature flag?
> 
> Another thing we should do is make sure these are tested on all platforms (MacOS, Windows, Linux) before enabling in prod, since the process sampling code will work differently on different platforms. There should be folks on the team with Windows and Linux laptops that we can ask to sanity check.

Gated the change in 9f487ca43b6485cfa674d32707be78b71d89f3e9

danielpeng2 (approved):

PR #14779, opened 2026-08-06
Session search: find any coding-agent conversation by name or by what was said in it, and resume it

### LINKED ISSUE #14731: Session search: find any coding-agent conversation by name or by what was said in it, and resume it
## Problem

Once you have run more than a handful of coding-agent sessions there is no way to *find* one. The project rail can only show what fits in a sidebar — it caps discovered sessions per project — so conversations that exist on disk are simply unreachable. On this machine that is 73 project directories and several thousand Claude transcripts.

Two questions go unanswered: **"which conversation was the one about X?"** and, more painfully, **"where is that thing I pasted?"**

## Proposal

A popup (`Cmd+Shift+K`, plus a magnifier in the rail header) over every known session — live tabs, stored session handles, and transcripts discovered on disk — deduped by `(agent, session_id)`. Enter resumes: if that session's pane is still open in the right directory it activates that tab, otherwise it opens a new one with the resume command prefilled but not executed.

Results are **staged**, and the staging is the design:

- **Names** match synchronously against data already in memory and render on the same frame as the keystroke.
- **Content** matches inside transcript text and appends *below*, under its own header, with a one-line snippet and a progress readout. Name results never reorder.

Content search reads transcripts off the main thread into an in-memory digest keyed on `(path, len, mtime)` — the `len` component makes growth append-only, so a transcript that grew re-reads only `[old_len, EOF)`. Matching is literal (`memchr::memmem`), never fuzzy, so an identifier like `POA-2236` stays one token. What counts as content is decided structurally rather than by regex: a `"type":"user"` envelope whose blocks carry a `tool_use_id` is a tool result, not something the user typed — tool output and pasted file bodies are excluded, and the footer says so rather than letting a half-answer look complete.

## Three constraints worth recording for anyone implementing this

1. **The mixer must stay synchronous.** Registering an `AsyncDataSource` withholds *all* results — including the instant name matches — for up to `INITIAL_RESULTS_TIMEOUT` (500 ms) while the palette still shows the previous query's rows, and that is also why Enter is a no-op while loading (`search_bar.rs`). Feeding a sync source from a model that owns its own `spawn` avoids both.
2. **Section order is a tier, and the direction is counterintuitive**: the mixer sorts *ascending* on `(priority_tier, score, source_order)` and the search bar then reverses for top-down, so a **higher** tier renders **higher**. This inverts silently with no compile error.
3. **`SelectionUpdate::Top` does not skip static separators.** A section header at index 0 gets selected when the popup opens, and the user's first Enter does nothing at all.

Implementation (fork, reusing the existing command-palette view rather than a new modal): https://github.com/samithaj/warp/tree/samithaj/session-search — built on #14720.

## Scope note

Content search covers Claude Code only, because Codex keeps no per-cwd transcript store — its sessions are date-keyed rollouts. Codex sessions still appear in name search via recorded handles.

samithaj:
## Description

Closes #14731. **Stacks on #14778** (and therefore #14735, #14734) — those commits are included here, so merge them first.

Past a handful of coding-agent sessions there is no way to *find* one. The project rail can only show what fits in a sidebar — it caps discovered sessions per project — so conversations that exist on disk are unreachable. On a working machine that is 73 project directories and several thousand transcripts.

`Cmd+Shift+K`, or the magnifier in the rail header, opens a popup over every known session: live tabs, stored handles, and transcripts discovered on disk, deduped by `(agent, session_id)`. Enter resumes — activating that tab if its pane is still open in the right directory, otherwise opening a new one with the resume command prefilled but never executed.

**Results are staged, and the staging is the design.** Names match synchronously against data already in memory and render on the same frame as the keystroke. Content matches inside transcript text and appends *below*, under its own header, with a snippet and a progress readout. Name results never reorder.

This reuses the existing command-palette `View` — the type `Workspace` already instantiates twice — as a third instance with its own mixer, rather than building a modal. It inherits the centered chrome, `Dismiss` focus trap, click-outside, Escape, arrow keys, scroll-into-view and fuzzy highlighting.

Content search reads transcripts off the main thread into an in-memory digest keyed on `(path, len, mtime)`; the `len` component makes growth append-only, so a transcript that grew re-reads only `[old_len, EOF)`. Matching is literal (`memchr::memmem`), never fuzzy, so an identifier like `POA-2236` stays one token. What counts as content is decided structurally rather than by regex: a `"type":"user"` envelope whose blocks carry a `tool_use_id` is a tool result, not something the user typed. Tool output and pasted file bodies are excluded — that exclusion is the latency win, and the footer states it rather than letting a half-answer look complete.

### Three constraints worth recording, because each compiles fine and fails at runtime

1. **The mixer must stay synchronous.** Registering an `AsyncDataSource` withholds *all* results — including instant name matches — for up to `INITIAL_RESULTS_TIMEOUT` (500 ms) while the palette still shows the previous query's rows, and that is also why Enter is a no-op while loading. Feeding a sync source from a model that owns its own `spawn` avoids both.
2. **Section order is a tier, and the direction is counterintuitive**: the mixer sorts *ascending* on `(priority_tier, score, source_order)` and the search bar then reverses for top-down, so a **higher** tier renders **higher**. The ordering test deliberately scores content rows *above* name rows, so only the tiers can produce the correct sequence.
3. **`SelectionUpdate::Top` does not skip static separators.** A section header at index 0 is selected when the popup opens and the user's first Enter does nothing at all.

Two more surfaced while building: `dispatch_typed_action` is immediate while `close()` is queued, so reaching this popup from Cmd-P would open then instantly close it; and `ctx.focus` in `open_palette` was unconditional, which ctrl-tab tolerates only because it has no text input.

## Linked Issue
#14731

- [x] Where appropriate, screenshots or a short video of the implementation are included below

## Testing

293 tests green across the agent-session, digest, tier-ordering and rail suites; `cargo clippy -p warp --all-targets -- -D warnings` clean across `warp` and `warp_search_core`. Rebased onto current master.

Digest tests use a fixture directory: `tool_use_id` present → excluded and absent → included; a torn final line does not abort the file; a grown file re-reads only its tail; an I/O error is **not** memoized.

## Scope note

Content search covers Claude Code only — Codex keeps no per-cwd transcript store, its sessions being date-keyed rollouts. Codex sessions still appear in name search via recorded handles, and the footer names the corpus it actually covers rather than implying completeness.

## Agent Mode

CHANGELOG-NEW-FEATURE: Session search — press Cmd+Shift+K to find any coding-agent conversation by its name or by something said inside it, and resume it where you left off.


samithaj:
Pushed an update. Alongside the fixes shared with the PRs below, this one now also carries **"Clear shells without agents"**, which replaces the hide toggle that #14778 introduces.

It landed here rather than in #14778 because it depends on this PR's rail-header layout — the new button sits beside the session-search magnifier, and the two share the header's control-resolution code. Since the stack merges in order, the end state is the same; #14778 simply still has the toggle in isolation.

The action closes only a lone terminal pane with no agent, that is not the active tab, has no long-running command, and is not being shared. Two of those exemptions close real holes rather than being belt-and-braces: `pane_has_agent` answers `false` when a group has no focused terminal session view, so a **code pane with unsaved edits** would otherwise have read as an agent-less shell; and `close_pane`, unlike `close_pane_with_confirmation`, does not itself check for a shared session.

Undo goes through the closed-tab stack rather than the hidden-pane one (every eligible tab is a single-pane group, so `close_pane` emits `Exited` before reaching the hidden-pane path) — which also makes it independent of `FeatureFlag::UndoClosedPanes`, and means one undo per tab rather than one for the batch.

PR #14837, opened 2026-08-07
Display the correct text in the environment selector when no environment is specified

Xavientois:
### Problem

Two problems are fixed by this pull request:
- Before the environment data is hydrated, it displays "New Environment", even on completed agent runs
- Once hydrated, the environment selector in the footer displayed the default environment instead of "No environment selected" when no environment is selected. This actually reflects the updated state once the environment catalog is loaded from the backend

### Solution

- Only auto-select the default environment when:
    - Handing off a session without a specified environment to a cloud agent
    - Configuring a new session
- Only show "New Environment" if the session is being configured

## Linked Issue

https://linear.app/warpdotdev/issue/REV-1926/warp-on-web-footer-shows-wrong-environment-for-empty-environment-cloud

## Testing

- [x] I have manually tested my changes locally with `./script/wasm/run`

### Screenshots / Videos

- Existing sessions without an environment show "No environment selected"
https://www.loom.com/share/893ed180ad9045509f708a71f7d9918f

- Environment selector auto-selects default environment when it is available 
https://www.loom.com/share/bc76517f3f5f45398719698ce687567c

- Selected env still displays
Screenshot 2026-08-07 at 6 36 27 PM

### DIFF
--- a/app/src/ai/blocklist/agent_view/agent_input_footer/environment_selector.rs
+++ b/app/src/ai/blocklist/agent_view/agent_input_footer/environment_selector.rs
@@ -247,7 +247,9 @@ impl EnvironmentSelector {
 
         let environments = CloudEnvironmentCatalog::handle(ctx);
         ctx.subscribe_to_model(&environments, |me, _, _, ctx| {
-            me.ensure_default_selection(ctx);
+            if me.should_auto_select_default_environment(ctx) {
+                me.ensure_default_selection(ctx);
+            }
             me.refresh_menu(ctx);
             me.refresh_button(ctx);
             ctx.notify();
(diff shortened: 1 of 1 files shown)


app/src/ai/blocklist/agent_view/agent_input_footer/environment_selector.rs
```diff
@@ -344,6 +351,15 @@ impl EnvironmentSelector {
         ctx.notify();
     }
 
+    fn should_auto_select_default_environment(&self, ctx: &AppContext) -> bool {
```
jasonkeung, line 354: nice! one initial question, are there any cases where we'd want to call just ensure_default_selection separately without should_auto_select_default_environment? if not, maybe we could consider folding this logic into ensure_default_selection

Xavientois: I did consider that, but didn't want to couple the predicate with the actual action.

app/src/ai/blocklist/agent_view/agent_input_footer/environment_selector.rs
jasonkeung, line 421: should the label correspond with the tooltip as well?

jasonkeung (approved):

app/src/ai/blocklist/agent_view/agent_input_footer/environment_selector.rs
Xavientois: discussed offline

PR #14849, opened 2026-08-08
Clone the Factory definition repo as a run-scoped setup command

alokedesai:
## Description
Makes the Factory definition-repo checkout **run-scoped** instead of environment-persisted.

**What:** When a dispatch attaches `WARP_FACTORY_REPO_CLONE_URL` / `WARP_FACTORY_REPO_DIR` to a run — which the server does only for runs executing as a Factory agent whose Factory definition lives in a Warp-managed repository — the agent driver now prepends a guarded clone ahead of the environment's setup commands:

```
[ -e "$WARP_FACTORY_REPO_DIR" ] || git clone "$WARP_FACTORY_REPO_CLONE_URL" "$WARP_FACTORY_REPO_DIR"
```

**Why:** Today the clone command is persisted into the Factory's *managed* environment's `setup_commands` by the server-side projector. That design has two defects:
- Factories whose default environment is **external** (user-owned, declared via `agentDefaults.environmentId` in the Factory definition) never get the clone command — the projector must not rewrite a user-owned environment — so those runs silently get no definition checkout and no skills.
- The persisted command runs for **every** run in the environment, including non-Factory runs where the env vars are absent (`git clone "" ""` setup failures; mitigated server-side by making the persisted command self-guarding).

**How:** The prepend is keyed purely off the presence of the per-run env vars, so "is this a Factory run with a Warp-managed definition repo" is decided entirely by the dispatch, independent of which environment the run uses. Three safety properties:
- The command expands the variables in the session shell, so the credential-bearing clone URL never appears in command text.
- `[ -e ]` makes it idempotent.
- If the environment still persists its own copy of the clone command (all managed environments today), the driver defers to it — no double clone. This makes rollout order-independent; once this ships broadly, the server-side projector stamping can be removed and reconciles will strip the persisted copies (follow-up in warp-server).

## Linked Issue

No linked issue — this fixes a defect found during a QA pass of the managed Factory definition-repo feature (see warp-server PR #14136 review thread).

## Testing

Unit tests (`cargo nextest run -p warp -E 'test(factory_clone)'`, all passing after merging master):
- `factory_clone_is_prepended_when_clone_values_are_present`
- `factory_clone_is_skipped_without_clone_values`
- `factory_clone_defers_to_a_persisted_environment_copy`

End-to-end verification against a local oz-local stack (direct worker backend, Factory `uT0VjUDyDFmuEeNE1ZJqlH` with a Warp-managed definition repo):
1. Rebuilt the WarpLocal debug bundle from this branch and pointed the local worker's `--oz-path` at it.
2. **Post-cleanup state:** set the factory's managed environment to `"setup_commands": []` in the local DB (verified before and after the run). Dispatched a run as the factory's foreman agent; it SUCCEEDED, the definition checkout appeared at the workspace root mid-run (`.git` created seconds after task claim, `origin` = the factory's managed definition-repo remote, token redacted), and all 5 factory skills were injected with run-scoped absolute paths. Only the run-scoped clone could have produced it.
3. **Transition/dedup state:** restored the legacy *unguarded* persisted clone command into the managed environment and dispatched again. The run SUCCEEDED — decisive, because if the driver had also cloned, the unguarded persisted command would have failed on the existing directory and failed environment setup. Exactly one clone.
4. **External-environment state (the headline bug):** reproduced the server state of a Factory declaring an external default environment (registration's environment association pointing at a plain, user-owned team environment that has never contained a clone command). The run SUCCEEDED with the definition checkout and all 5 skills present — previously this configuration silently got neither.
5. `cargo clippy -p warp --all-targets -- -D warnings` and `./script/format` clean.

## Agent Mode

CHANGELOG-NONE

### DIFF
--- a/app/src/ai/agent_sdk/driver/environment.rs
+++ b/app/src/ai/agent_sdk/driver/environment.rs
@@ -148,8 +148,8 @@ pub(super) fn merge_repos_deduped(
 }
 
 /// Environment variable carrying the authenticated remote URL of a Factory's
-/// code.storage definition repository. Attached to the run by the server only
-/// for Factory-agent runs.
+/// definition repository. Dispatch attaches it only to runs that execute as a
+/// Factory agent whose Factory definition lives in a Warp-managed repository.
 const FACTORY_REPO_CLONE_URL_ENV_VAR: &str = "WARP_FACTORY_REPO_CLONE_URL";
 
 /// Environment variable carrying the directory, relative to the working
--- a/app/src/ai/agent_sdk/driver/environment.rs
+++ b/app/src/ai/agent_sdk/driver/environment.rs
@@ -147,6 +147,52 @@ pub(super) fn merge_repos_deduped(
     Ok(merged)
 }
 
+/// Environment variable carrying the authenticated remote URL of a Factory's
+/// code.storage definition repository. Attached to the run by the server only
+/// for Factory-agent runs.
+const FACTORY_REPO_CLONE_URL_ENV_VAR: &str = "WARP_FACTORY_REPO_CLONE_URL";
+
+/// Environment variable carrying the directory, relative to the working
+/// directory, that the Factory definition repository is cloned into.
+const FACTORY_REPO_DIR_ENV_VAR: &str = "WARP_FACTORY_REPO_DIR";
+
+/// Prepends the setup command that clones a Factory's definition repository
+/// when the dispatch attached the clone variables to this run, so the checkout
+/// exists before user-declared setup commands run.
+pub(super) fn prepend_factory_definition_clone(setup_commands: &mut Vec<String>) {
+    let clone_url = std::env::var(FACTORY_REPO_CLONE_URL_ENV_VAR).unwrap_or_default();
+    let clone_dir = std::env::var(FACTORY_REPO_DIR_ENV_VAR).unwrap_or_default();
+    prepend_factory_definition_clone_for_values(&clone_url, &clone_dir, setup_commands);
+}
+
+fn prepend_factory_definition_clone_for_values(
+    clone_url: &str,
+    clone_dir: &str,
+    setup_commands: &mut Vec<String>,
+) {
+    if clone_url.trim().is_empty() || clone_dir.trim().is_empty() {
+        return;
+    }
+    // Environments provisioned before run-scoped cloning still persist their
+    // own copy of the clone command; leave that copy in charge rather than
+    // attempting the checkout twice.
+    if setup_commands
+        .iter()
+        .any(|command| command.contains(FACTORY_REPO_CLONE_URL_ENV_VAR))
+    {
+        return;
+    }
+    // The command expands the variables in the session shell instead of
+    // inlining their values so the credential-bearing URL never appears in
+    // command text, and the existence guard makes it idempotent.
+    setup_commands.insert(
+        0,
+        format!(
+            "[ -e \"${FACTORY_REPO_DIR_ENV_VAR}\" ] || git clone \"${FACTORY_REPO_CLONE_URL_ENV_VAR}\" \"${FACTORY_REPO_DIR_ENV_VAR}\""
+        ),
+    );
+}
+
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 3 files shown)


app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -147,6 +147,52 @@ pub(super) fn merge_repos_deduped(
...
+) {
+    if clone_url.trim().is_empty() || clone_dir.trim().is_empty() {
+        return;
+    }
+    // Environments provisioned before run-scoped cloning still persist their
+    // own copy of the clone command; leave that copy in charge rather than
+    // attempting the checkout twice.
+    if setup_commands
+        .iter()
+        .any(|command| command.contains(FACTORY_REPO_CLONE_URL_ENV_VAR))
```
alokedesai, line 181: The deferral matches any setup command that mentions the variable, not the legacy clone command specifically, so `echo "$WARP_FACTORY_REPO_CLONE_URL"` or an export line would suppress the clone entirely and the run would get no definition checkout. Narrow the predicate to the command being deferred to:

```suggestion
        .any(|command| {
            command.contains("git clone")
                && command.contains(FACTORY_REPO_CLONE_URL_ENV_VAR)
                && command.contains(FACTORY_REPO_DIR_ENV_VAR)
        })
```

app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -147,6 +147,52 @@ pub(super) fn merge_repos_deduped(
...
+    {
+        return;
+    }
+    // The command expands the variables in the session shell instead of
+    // inlining their values so the credential-bearing URL never appears in
+    // command text, and the existence guard makes it idempotent.
+    setup_commands.insert(
+        0,
+        format!(
+            "[ -e \"${FACTORY_REPO_DIR_ENV_VAR}\" ] || git clone \"${FACTORY_REPO_CLONE_URL_ENV_VAR}\" \"${FACTORY_REPO_DIR_ENV_VAR}\""
```
alokedesai, line 191: `[ -e ]` treats a regular file, an empty directory, or an interrupted prior clone as a completed checkout, so setup continues against an unusable path and skill loading silently finds nothing. Guarding on a real work tree keeps the idempotency you want while letting a bad path fail visibly at clone time:

```suggestion
            "git -C \"${FACTORY_REPO_DIR_ENV_VAR}\" rev-parse --is-inside-work-tree >/dev/null 2>&1 || git clone \"${FACTORY_REPO_CLONE_URL_ENV_VAR}\" \"${FACTORY_REPO_DIR_ENV_VAR}\""
```

alokedesai (commented):
## Overview
Moves the Factory definition-repo checkout from a persisted environment setup command to a run-scoped one keyed off the dispatch-attached env vars, which correctly decouples "is this a Factory run" from which environment the run uses. The shape is right and the credential handling is sound; two guards are looser than they need to be and can leave a Factory run without its definition checkout, so they are worth a decision before this leaves draft.

## Concerns
- The deferral check matches any setup command containing `WARP_FACTORY_REPO_CLONE_URL`, not specifically the legacy clone command, so a benign `echo`/`export`/logging command that references the variable suppresses the run-scoped clone and the run proceeds with no definition repo and no skills. Requiring `git clone` plus both variables would narrow the match to the command actually being deferred to.
- `[ -e "$WARP_FACTORY_REPO_DIR" ]` accepts a regular file, an empty directory, or a half-finished prior clone as a successful checkout, so setup continues and skill loading silently finds nothing. A `git -C "$WARP_FACTORY_REPO_DIR" rev-parse --is-inside-work-tree` guard keeps idempotency while letting an unusable path fail loudly at clone time.
- Both cases above are untested: the deferral test only covers an exact persisted clone command, and there is no case for an existing non-repository path. If you tighten either guard, a regression test for the false-positive path is the cheap way to pin it.

## Verdict
Checks: build pass (focused), tests pass (`factory_clone` 3/3), CI green (5 successful, 0 failing; 16 jobs skipped, including Formatting + Clippy and the main test matrix), visual proof n/a

Found: 0 critical, 2 important, 1 suggestion, 0 nits

PR #14853, opened 2026-08-08
[APP-3185] Kill generator process groups on cancellation

kevinyang372:
## Description
Fixes per-command cancellation for local generator commands so Warp terminates the entire owned process group rather than only the immediate shell or wrapper process.

### Why the existing process-group kill did not cover this path

`LocalCommandExecutor::cancel_active_commands()` already killed every tracked process group with `kill(-pgid, SIGKILL)`. That behavior was correct, but it is an explicit **session-wide bulk cancellation API**: a caller has to invoke it, and doing so cancels all generator commands running through the session's executor.

Prompt chips use a different cancellation mechanism. Each chip owns a `SpawnedFutureHandle`; when a refresh replaces an in-flight generator, a chip is disabled, or a shell-command timeout wins, Warp aborts that one handle and drops its individual `Session::execute_command()` future. This path intentionally does not call `session.cancel_active_commands()`, because local executors support parallel commands and a single Git chip refresh must not also cancel unrelated completions, autosuggestions, or other chips.

Before this PR, dropping that command future relied on `kill_on_drop(true)`. That kills only the immediate child PID. It does not call `LocalCommandExecutor::cancel_active_commands()` and does not signal the child's process group. Because the future is dropped while awaiting `child.output()`, execution also never reaches the normal post-`await` bookkeeping removal.

### How that causes the enterprise failure

In the reported environment, the command resolves through an internal Git wrapper which launches the real Git process. The wrapper and real Git child inherit the dedicated process group created by Warp, but the old future-drop path only sent `SIGKILL` to the immediate leader/wrapper PID.

Since `SIGKILL` cannot be caught or forwarded:

1. Warp aborts the periodic Git chip's command future.
2. `kill_on_drop` kills the immediate shell or wrapper.
3. The real Git child receives no signal, becomes orphaned, and continues scanning the large repository.
4. The next periodic refresh starts another wrapper and Git child.
5. Repeating this produces hundreds of expensive Git processes and unbounded CPU/memory growth.

The existing bulk `cancel_active_commands()` path would have killed the whole group, but prompt refresh cancellation never invoked it.

### How this PR fixes it

Each spawned command now registers a stable process-group record in `ActiveProcessGroups`. Both cancellation entry points converge on the same idempotent operation:

- Dropping one command future drops its `SpawnedChildCleanup`, which calls `ActiveProcessGroups::cancel(&record)`.
- `cancel_active_commands()` calls `ActiveProcessGroups::cancel_all()`, which snapshots the active records and calls that same `cancel(&record)` operation for each one.
- Successful command completion calls `complete(&record)`, removing it without sending a signal.

`cancel(&record)` claims the registration exactly once and sends `SIGKILL` to `-pgid`, so the wrapper and all descendants still in its process group are terminated directly. The wrapper no longer needs to catch or forward a signal. Stable record identity also prevents an old bulk-cancellation snapshot from targeting a later process that reused the same numeric PGID.

This preserves per-command cancellation for prompt refreshes without using the session-wide bulk API, while ensuring both paths share the same process-tree termination behavior.

### Regression coverage

Real-process tests start shell descendants behind FIFOs and verify:

- Dropping an individual command future kills its descendant before it can perform a side effect.
- Calling `cancel_active_commands()` kills its descendant through the same registry path.
- Normal completion unregisters the command, so later bulk cancellation does not affect its detached descendant.

## Linked Issue
- Internal: [APP-3185](https://linear.app/warpdotdev/issue/APP-3185/zombie-git-processes-from-prompt-and-code-review-panel-never-cleaned)
- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. (N/A: internal Linear follow-up)
- [x] Where appropriate, screenshots or a short video are included below. (N/A: process-lifecycle change with no UI)

## Testing
- `cargo nextest run -p warp 'local_command_executor::tests::unix'` — 3 passed
- Exact three Clippy commands from `script/presubmit` — passed after merging current `origin/master`
- `./script/format --check` — passed after merging current `origin/master`
- Full `./script/presubmit` — passed before the non-overlapping `origin/master` merge (10,872 workspace tests plus completer-v2 and doc tests)
- Local platform: macOS
- Linux cloud verification was not run because `oz-dev` is unavailable in the submitting environment; the new implementation and tests are Unix-gated

- [ ] I have manually tested my changes locally with `./script/run` (N/A: no interactive UI behavior)

## Agent Mode

## Agent artifacts
- [Conversation](https://staging.warp.dev/conversation/59b9c914-a845-4854-a0b6-36f68896ce5b)
- [Implementation plan](https://staging.warp.dev/drive/notebook/dM39SGP1RhWUSCTLvcjEGM)

CHANGELOG-BUG-FIX: Fixed canceled background commands leaving wrapped child processes running.

### DIFF
--- a/app/src/terminal/model/session/command_executor/local_command_executor.rs
+++ b/app/src/terminal/model/session/command_executor/local_command_executor.rs
@@ -29,6 +29,8 @@ fn terminate_process_group(process_group_id: u32) {
         }
     }
 }
+#[cfg(not(unix))]
+fn terminate_process_group(_: u32) {}
 
 #[derive(Debug, Default)]
 struct ActiveProcessGroups {
@@ -68,12 +70,9 @@ impl ActiveProcessGroups {
     }
 
     fn cancel(&self, process_group: &Arc<ActiveProcessGroup>) {
-        if !self.remove(process_group) {
-            return;
+        if self.remove(process_group) {
+            terminate_process_group(process_group.id);
         }
-
-        #[cfg(unix)]
-        terminate_process_group(process_group.id);
     }
 
     fn cancel_all(&self) {
--- a/app/src/terminal/model/session/command_executor/local_command_executor.rs
+++ b/app/src/terminal/model/session/command_executor/local_command_executor.rs
@@ -90,33 +90,31 @@ impl ActiveProcessGroups {
 }
 
 struct SpawnedChildCleanup {
-    process_group: Arc<ActiveProcessGroup>,
+    process_group: Option<Arc<ActiveProcessGroup>>,
     active_process_groups: Arc<ActiveProcessGroups>,
-    armed: bool,
 }
 
 impl SpawnedChildCleanup {
     fn new(process_group_id: u32, active_process_groups: Arc<ActiveProcessGroups>) -> Self {
         let process_group = active_process_groups.register(process_group_id);
         Self {
-            process_group,
+            process_group: Some(process_group),
             active_process_groups,
-            armed: true,
         }
     }
 
-    fn disarm(mut self) {
-        self.active_process_groups.complete(&self.process_group);
-        self.armed = false;
+    fn complete(mut self) {
+        if let Some(process_group) = self.process_group.take() {
+            self.active_process_groups.complete(&process_group);
+        }
     }
 }
 
 impl Drop for SpawnedChildCleanup {
     fn drop(&mut self) {
-        if !self.armed {
-            return;
+        if let Some(process_group) = self.process_group.take() {
+            self.active_process_groups.cancel(&process_group);
         }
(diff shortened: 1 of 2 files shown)


app/src/terminal/model/session/command_executor/local_command_executor.rs
```diff
@@ -215,8 +318,9 @@ impl LocalCommandExecutor {
                 );
                 anyhow!(e)
             });
-
-        self.spawned_children_pids.lock().remove(&child_pid);
+        if output.is_ok() {
+            child_cleanup.disarm();
```
alokedesai: I don't really get what `disarm` is meant to denote?

alokedesai (approved):

PR #14859, opened 2026-08-09
Open rendered Markdown at search result locations

### LINKED ISSUE #9657: Global file search should jump to matching line
### Pre-submit Checks

When you use global file search you see matches. Clicking on the match opens the file but at the top. It should jump to where the search string you clicked on is located.

_No response_

### Additional context

Similar to this request re "warp search": https://github.com/warpdotdev/warp/issues/9204

### Operating system (OS)

Linux

3

None

jerryzejia:
## Description

Ceates a map at parse time between raw md line and the line in rendered document. 

When a search result is clicked, look up its line and scroll there. 

Initially I tried to fix this by text matching with the raw file and then find the searched item in rendered page by counting occurances, but there edge cases such as punctuation only matches and element hidden by rendering (links) would break it. 

-------

Global search results now open rendered Markdown files at the matching content instead of leaving the viewer at the top of the document.

Source line and column offsets do not survive Markdown rendering, so the position of a match in the raw file says nothing about where it lands on screen. This change builds an explicit source map during parsing instead of trying to recover the position afterwards:

- `parse_markdown_with_source_map` anchors every source line to a `FormattedTextLine` index plus a row offset within that line;
- the lowering step records the buffer offset each `FormattedTextLine` actually landed at, and the two compose into a real buffer offset;
- source lines that render to nothing, such as the inside of an HTML comment, resolve to `None`, so the viewport is left alone rather than scrolled somewhere arbitrary;
- multi-line constructs stay precise: each line of a fenced code block maps to its own row, and a GFM table's separator shares the header's row; and
- the scroll is retried while the viewport and progressively rendered blocks settle, and stale deferred scrolls are cancelled when a newer target cannot be resolved.

This supersedes the text-matching approach first pushed to this branch, which reconstructed a search needle from the source line and counted occurrences in the rendered text. That approach is removed, along with `match_text` from the global-search open-file event chain. It addressed both review comments on this PR by construction rather than by patching them:

- counting raw-source matches could select a later visible occurrence when an earlier match was hidden by rendering (for example inside a link destination). There is no occurrence counting left; a match anchors to its own source line.
- punctuation-only matches such as `=>` or `---` were discarded as unsearchable, so valid hits inside code fences failed to navigate. There is no needle and no punctuation filter left; those lines resolve like any other.

## Linked Issue
Closes #9657
warpdotdev/warp/issues/9657

## Testing

- Added parser coverage for the source map across fenced blocks, GFM tables, HTML comments, frontmatter, lists, blockquotes, and CRLF input.
- Added buffer coverage that source lines resolve to the rendered row a reader expects, asserting on the landed row's text rather than on literal offsets, including the constructs that merge or drop rows: thematic breaks, block images, dropped embedded objects, and repeated block items.
- Added coverage that the source map survives code-block syntax highlighting, which applies a style-only edit after load, and is invalidated by edits that actually change text.
- Added workspace and file-view coverage verifying that line targets reach new and reused Markdown notebook panes.
- `cargo test -p markdown_parser --lib`: 162 passed.
- `cargo test -p warp_editor --lib markdown`: 44 passed.
- `cargo test -p warp --lib source`: 173 passed; `global_search`, `notebooks::file`, `notebooks::editor` suites also pass.
- Ran `cargo fmt --check`, `git diff --check`, and `cargo clippy --all-targets` on the changed crates with no warnings.

### Screenshots / Videos

https://github.com/user-attachments/assets/a6a35451-e110-4da6-8896-1240ba8d4f7c

## Agent Mode

CHANGELOG-BUG-FIX: Fixed global search results not scrolling to the matching location in rendered Markdown files.


jerryzejia:
@cla-bot check

PR #14891, opened 2026-08-10
fix(macos): replace the base letter when the accent popup commits

### LINKED ISSUE #13631: macOS accent popup inserts both the base letter and the selected accented character
### Pre-submit Checks

### Describe the bug

When holding a key that has accented variants (e.g. `u`, `o`, `a`), Warp correctly
shows the macOS press-and-hold accent popup. However, after selecting an accented
character from the popup, Warp inserts BOTH the base character and the selected
accented character, instead of just the accented one.

Example: holding `o` and selecting `ö` from the popup results in `oö` being typed,
not just `ö`

### To reproduce

1. In a Warp prompt, press and hold a letter key that has accent variants (e.g. `o`)
2. Wait for the macOS accent popup to appear
3. Select an accented variant from the popup (click it or press its number key)
4. Observe the text that gets inserted

### Expected behavior

Only the selected accented character should be inserted (e.g. `ö`), overriding the
base character, matching the behavior in Terminal.app

https://github.com/user-attachments/assets/0428be8c-6556-40e5-8c58-331d07e6b648

### Operating system (OS)

macOS

26.6 Beta (25G5057c)

### Shell Version

### Current Warp version

v0.2026.07.08.17.54.stable_02

### Regression

### Recent working Warp date

_No response_

### Additional context

Might be related to #3191

macOS built-in Terminal.app is fine. Ghostty has another issue (https://github.com/ghostty-org/ghostty/discussions/10669)

None

maxmilian:
## Description

On macOS, selecting a character from the native press-and-hold accent popup inserts **both** the held base letter and the accented character — hold `o`, pick `ò`, and the input reads `oò`.

**Root cause.** Holding the key commits a literal `o` through the ordinary typed-character path. The popup's selection is then meant to *replace* that character, but `insertText:replacementRange:` in `host_view.m` appends unconditionally, so both survive.

### What the input system actually sends

I instrumented `insertText:`, `setMarkedText:` and `unmarkText` on a local OSS build and drove the real popup end to end. **Two different shapes reach us, depending on how the entry is chosen** — and a dead-key control run is included, because that path must keep working:

```
A — hold `o`, arrow keys, Return
   insertText    'o'  loc=NSNotFound              interpreting=1 hasMarked=0
   setMarkedText 'ò'  selLen=1  replLoc=0          interpreting=1 hasMarked=0   ← popup opens
   setMarkedText 'ó'  selLen=1  replLoc=NSNotFound interpreting=1 hasMarked=1
   insertText    'ó'  loc=NSNotFound              interpreting=1 hasMarked=1   ← commit

B — hold `o`, press `1`
   insertText 'o'  loc=NSNotFound  interpreting=1 hasMarked=0
   insertText 'ò'  loc=0           interpreting=1 hasMarked=0                  ← commit

C — Option-E then `g`  (dead-key control, must not change)
   setMarkedText '´'  selLen=0  replLoc=NSNotFound interpreting=1 hasMarked=0
   insertText    '´'  loc=NSNotFound  interpreting=1 hasMarked=1
   insertText    'g'  loc=NSNotFound  interpreting=1 hasMarked=1
```

Three things fall out of this, and each one shaped the patch:

1. **Both commits happen *inside* `interpretKeyEvents:`** (`interpreting=1`). Anything gated on being outside it is dead code.
2. **At `insertText:` time, A is indistinguishable from the dead-key control** — same `NSNotFound`, same `interpreting`, same `hasMarkedText`. B is distinguishable by its real `replacementRange` location.
3. **The discriminator for A is in `setMarkedText:`**: the popup opens with a real `replacementRange` location, a shape dead-key composition never takes.

So the patch sets a flag when `setMarkedText:` arrives with a real `replacementRange`, and treats the next `insertText:` — or any `insertText:` that carries its own real location — as a replacement rather than an append. The popup's in-progress marked text is cleared first, otherwise the editor would delete that ephemeral text instead of the committed character.

The delete and the insert share one `edit()` call, so the pair is a single undo step and the buffer is never briefly in the `oò` state.

## Scope

Sending this as a draft so the scope calls are reviewable on a concrete diff.

1. **Not truthful `selectedRange`.** The general fix is to report the real caret offset so AppKit can compute a proper replacement range itself; I verified that works (a non-degenerate `selectedRange` makes AppKit hand back `{caret-1, 1}`). It needs a caret→UTF-16 offset bridge across the shared text-input surface, which is a much wider change. Happy to build that instead if you'd prefer the general fix.
2. **Editor input only.** `Event::ReplacePrecedingCharacters` is handled in the editor element. The block-list / alt-screen paths still append, because replacing there means writing a DEL byte into the PTY, and that is a behavioral choice I don't think a contributor should make unilaterally. Say the word and I'll extend it.
3. **Vim keeps its existing path.** Vim owns its own insertion state machine, so that branch stays on `VimUserInsert` rather than reaching around it.

## Linked Issue

Fixes #13631

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement` — labeled `ready-to-implement` on 2026-07-16.

## Testing

Three regression tests in `app/src/editor/view/marked_text_tests.rs`:

- `test_replace_preceding_characters_accent_popup` — `echo` plus a popup `ò` yields `echò`, not `echoò`.
- `test_replace_preceding_characters_costs_no_extra_undo_step` — compared against a control editor that types the accented word directly, one undo empties both. Without the control this assertion would prove nothing, since typing coalescing alone produces the same result.
- `test_replace_preceding_characters_at_buffer_start_inserts` — with nothing before the caret the commit degrades to a plain insert rather than eating a character that belongs to the user.

The existing `marked_text_tests` IME/dead-key suite stays green (8 passed total).

**What the unit tests cannot cover, and why the manual run matters here.** The detection lives in Objective-C on AppKit callbacks that only a real press-and-hold produces. An earlier revision of this patch had all of these tests passing and still reproduced the bug on a real build, because the detection was gated on being outside `interpretKeyEvents:` and therefore never ran. The measurements above, and the before/after below, come from driving the actual popup on a local OSS build.

### Screenshots / Videos

Both builds are local OSS builds from this branch; BEFORE differs only by `host_view.m` reverted to `master`. The same synthesized key sequence drives each pair, so the two halves differ by the patch and nothing else.

**A — hold `o`, arrow keys, Return** (the variant the first revision of this patch missed)

arrow keys and Return

**B — hold `o`, press `1`**

number key

**C — Option-E then `g`, the dead-key control**

Unchanged by the patch. `´g` is `master`'s existing behavior here; the point of this pair is that it stays exactly as it was.

dead-key control

## Agent Mode

CHANGELOG-BUG-FIX: Fixed macOS press-and-hold accent selection inserting both the base letter and the accented character.

### DIFF
--- a/app/src/editor/view/element.rs
+++ b/app/src/editor/view/element.rs
@@ -625,6 +625,24 @@ impl EditorElement {
         false
     }
 
+    /// Handles a text-input commit that replaces the grapheme before the caret, rather
+    /// than extending the input. See [`Event::ReplacePrecedingCharacters`].
+    fn replace_preceding_characters(&self, chars: &str, ctx: &mut EventContext) -> bool {
+        if self.view_snapshot.is_focused {
+            if self.vim_mode.is_some() {
+                // Vim owns its own insertion state machine, so we leave that path on the
+                // plain insert it has always used rather than reaching around it.
+                ctx.dispatch_typed_action(EditorAction::VimUserInsert(UserInput::new(chars)));
+            } else {
+                ctx.dispatch_typed_action(EditorAction::ReplacePrecedingCharacters(
+                    UserInput::new(chars),
+                ));
+            }
+            return true;
+        }
+        false
+    }
+
     fn set_marked_text(
         &mut self,
         marked_text: &str,
@@ -2176,6 +2194,9 @@ impl Element for EditorElement {
                 key_code, state, ..
             } => self.modifier_key_change(key_code, state, ctx),
             Event::TypedCharacters { chars } => self.typed_characters(chars, ctx),
+            Event::ReplacePrecedingCharacters { chars } => {
+                self.replace_preceding_characters(chars, ctx)
+            }
             Event::DragAndDropFiles { paths, location } => {
                 self.drag_and_drop_file(paths.clone(), *location, ctx)
             }
--- a/app/src/editor/view/marked_text_tests.rs
+++ b/app/src/editor/view/marked_text_tests.rs
@@ -191,3 +191,89 @@ fn test_set_marked_text_vim_insert_mode() {
         });
     });
 }
+
+/// Regression test for #13631: the macOS press-and-hold accent popup commits the base
+/// character first and then commits the accented character as a replacement for it. The
+/// replacement must land as a single grapheme, not appended after the base character.
+#[test]
+fn test_replace_preceding_characters_accent_popup() {
+    App::test((), |mut app| async move {
+        initialize_app(&mut app);
+
+        app.add_window(WindowStyle::NotStealFocus, |ctx| {
+            // Typed rather than seeded as base text, so the caret ends up after the
+            // text the way it does when a user is typing a command.
+            let mut editor = EditorView::new_with_base_text("", Default::default(), ctx);
+            editor.user_insert("ech", ctx);
+
+            // Holding `o` commits a literal `o` through the ordinary typed-character path.
+            editor.user_insert("o", ctx);
+            assert_eq!(editor.buffer_text(ctx), "echo");
+
+            // Picking `ò` from the popup replaces that `o` rather than following it.
+            editor.replace_preceding_characters("ò", ctx);
+            assert_eq!(editor.buffer_text(ctx), "echò");
+
+            editor
+        });
+    });
+}
+
+/// The replacement costs no extra undo step: undoing an accented word behaves exactly
+/// like undoing the same word typed directly, rather than first uncovering the
+/// intermediate `echo` state that only existed because of how the popup commits.
+///
+/// The directly-typed editor is the control — without it, a single `undo` reaching an
+/// empty buffer would prove nothing, since that is also what typing coalescing alone
+/// would produce.
+#[test]
+fn test_replace_preceding_characters_costs_no_extra_undo_step() {
+    App::test((), |mut app| async move {
+        initialize_app(&mut app);
+
+        app.add_window(WindowStyle::NotStealFocus, |ctx| {
+            let mut control = EditorView::new_with_base_text("", Default::default(), ctx);
+            control.user_insert("ech", ctx);
+            control.user_insert("ò", ctx);
+            assert_eq!(control.buffer_text(ctx), "echò");
+            control.undo(ctx);
+            let control_after_one_undo = control.buffer_text(ctx);
+            assert_eq!(control_after_one_undo, "");
+
+            control
+        });
+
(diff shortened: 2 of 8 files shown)


maxmilian:
Both review points addressed in `093c3f6`.

**1. Non-editor surfaces dropped the replacement commit.** Confirmed — `Event::TypedCharacters` is consumed by `block_list_element`, `alt_screen_element` and `crates/editor`'s element as well as `EditorElement`, and none of them knew about the new event, so a commit from the accent popup went nowhere while the terminal had focus.

Rather than teach each surface an event it cannot act on, the fallback is applied once at the point of dispatch: `warp_handle_replace_preceding_text` redispatches as `TypedCharacters` when nothing handled the replacement (`Event::unhandled_fallback`). Surfaces that can delete backwards get the replacement; every other one — present or future — keeps the plain append it always had, so no committed character is ever dropped. Covered by two tests in `crates/warpui_core/src/event_tests.rs`.

**2. Stale `accentPopupReplacingCommittedText` after a cancelled popup.** Confirmed, and fixed in two places so it cannot depend on which teardown AppKit picks:
- `unmarkText` clears the flag — the popup is gone whether it committed or was dismissed.
- the commit path now reads `accentPopupReplacingCommittedText && [self hasMarkedText]`. The popup's marked text is still standing when the arrow-key + Return variant commits (that was the measured shape in the PR description), while a dismissed popup has none, so a flag that outlived its popup is inert even if the dismissal never routed through `unmarkText`.

The number-key variant is unaffected: it matches on `replacementRange.location != NSNotFound` and never consults the flag.

I should be straight about the verification asymmetry here: the fallback is pinned by unit tests, but the `host_view.m` state machine has no test seam in the repo, so the flag change rests on the argument above plus the measurements in the PR description rather than on a new automated test. Happy to record another local-build session covering cancel-then-type if you'd like that on the record.

Verification: `cargo test -p warpui_core --lib` 309 passed, `cargo test -p warp --lib editor::` 569 passed (includes the 8 `marked_text` tests), `cargo build -p warpui` (recompiles `host_view.m`) clean, `cargo clippy -p warpui_core -p warpui -p warp --all-targets --tests -- -D warnings` clean, `cargo fmt -- --check` clean.

maxmilian:
@vkodithala — gentle nudge on this one. Both Oz points were addressed in `093c3f6` and the re-review came back clean; it's still mergeable against current `master`. The fix is macOS-only (press-and-hold accent commit replaces the base letter instead of appending). Let me know if you'd like anything changed.

PR #14895, opened 2026-08-10
fix(workspace): keep a new tab group's terminal from eating its name

### LINKED ISSUE #14241: Creating a tab group: the auto-created terminal steals focus from the group-name field mid-typing
**Repro**

1. Create a new tab group. The group-name field receives focus in edit mode, ready for typing.
2. Immediately start typing the group's name.
3. About a second later, focus jumps to the terminal that was auto-created inside the new group; remaining keystrokes go to the shell prompt instead of the name field.

In my case I got 3 characters of the group name typed before focus was stolen.

**Expected**

Focus should not be stolen when a terminal is created as a side-effect of creating a group. The name field should keep focus until the rename is committed or dismissed (Enter / Esc / click-away).

**Environment**

- Warp 0.2026.07.08.17.54.02 (stable)
- macOS 27.0

maxmilian:
## Description

Creating a tab group opens the inline name editor *and* spawns a terminal. About a second later that terminal's bootstrap block becomes visible and takes focus, so the characters the reporter was still typing landed in the shell instead.

**The part that isn't in the title.** Blur is treated as confirmation, identically to Enter:

```rust
// app/src/workspace/view.rs
EditorEvent::Blurred | EditorEvent::Enter => self.finish_tab_group_rename(ctx),
```

`finish_tab_group_rename` takes the buffer text, and if it is non-empty sets `group.name` and dispatches `workspace:save_app`. So @fbartho's three characters were not merely lost — they became the group's actual, **persisted** name. The keystrokes reaching the shell are the visible symptom; the silently committed truncated name is the one a user won't notice until later.

That half is pinned by a test that fails on `master`:

```
assertion `left == right` failed: a rename interrupted by the terminal stealing focus must not be committed
  left: Some("Bui")
 right: None
```

## What this changes

Two layers, because removing the trigger and guaranteeing the invariant are not the same thing.

**1. The terminal does not take focus while an inline rename editor is waiting for keystrokes.** This is the cause: with focus intact there is no blur, nothing is committed, and the keystrokes go where the user aimed them.

`WorkspaceState` already tracks `tab_group_being_renamed` (it is even folded into `is_any_non_palette_modal_open`), but it lives on `WorkspaceView`, and `app/src/terminal/view.rs` has no reference to workspace state at all. This adds the narrowest bridge I found: a small singleton, `InlineRenameState`, that the workspace publishes to and the terminal reads — the same shape as the `SemanticSelection::as_ref(ctx)` usage a few lines away in that file.

**Please weigh in, because this touches a documented policy.** `agents/specs/APP-4881-tui-starting-shell-indicator.md` states that `VisibleBootstrapBlock` updates focus immediately, "matching the GUI's terminal-focus policy". This carves an exception into that policy rather than treating it as a bug.

The cost, so it is weighed fairly: a new singleton is one more global that every app-construction path must register. I missed the test harness first time and got `Cannot get singleton model ... that was never registered` — loud and immediate rather than silent, but a real coupling cost, not just a new file.

**2. Blur discards the rename instead of committing it.** Layer 1 removes the trigger we know about; it does not make the data loss impossible. Any other non-user-initiated focus change — one that exists today and I haven't found, or one added later — would silently commit a fragment again, with nothing to catch it.

I noticed this because the regression test kept failing after layer 1 landed: the test drives `EditorEvent::Blurred` directly, so it was asserting an invariant I had not actually implemented. Rewriting the test to match the implementation would have hidden the gap convincingly.

Discarding on blur follows the convention used by rename inputs elsewhere (VS Code's rename box commits on Enter, cancels on Escape and on blur), and the two outcomes are not symmetric: discarding costs the user a retype, committing costs them their existing name, silently.

**This does make tab groups differ from their siblings.** Tab and pane rename still commit on blur (`view.rs:1457`, `:1475`). They aren't reachable by this bug because neither races an auto-created terminal, so I left them alone rather than widening the change — but if you'd rather align all three, say so and I'll do it here.

## Linked Issue

Fixes #14241

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement` — labeled `ready-to-implement` on 2026-07-28.

## Testing

Two tests in `app/src/workspace/view_tests.rs`:

- `test_tab_group_rename_blur_does_not_commit_unfinished_name` — the regression above. Fails on `master` with `Some("Bui")`, and still failed with only layer 1 in place, which is what surfaced the need for layer 2.
- `test_tab_group_rename_exits_release_terminal_focus_suppression` — exercises **both** exits, Enter and Escape. Suppressing terminal focus is only safe if every path out of a rename clears it again; if Escape left it set, terminals would silently stop taking focus for the rest of the session, which is a worse bug than the one being fixed and far harder to trace.

`finish_tab_group_rename` had no direct coverage before this. The full `workspace::view::tests` suite is green (91 passed), including the existing tab-group assertions.

## Manual testing (local OSS build)

Built `WarpOss.app` from this branch with `./script/run --dont-open` (default features, so `grouped_tabs` is on) and recorded the same interaction twice, on the same machine and the same window: create a tab group from the command palette, then type `Building` into the inline name editor one character at a time.

**How the race was made reproducible, stated plainly.** The focus steal is driven by `ModelEvent::VisibleBootstrapBlock`, and `blocks.rs` only emits that when the new terminal's bootstrap block actually renders something — the oh-my-zsh update prompt in @fbartho's report is exactly such a case. A shell that starts silently never emits it, so the bug is invisible on a bare `zsh`. For the capture I launched the app with a throwaway `ZDOTDIR` whose `.zshrc` sources my normal one and then prints one line:

```sh
[ -f "$HOME/.zshrc" ] && source "$HOME/.zshrc"
sleep 1
print -r -- "[oh-my-zsh] Would you like to check for updates? [Y/n]"
```

Nothing else differs between the two takes — same window, same steps, same shell, same typing cadence. Only `app/src/terminal/view.rs` and `app/src/workspace/view.rs` are reverted to `master` for the BEFORE take.

- **BEFORE (`master`)** — the terminal takes focus mid-word. The group is committed as **`Build`**, and `ing` lands in the shell: `zsh: command not found: ing`.
- **AFTER (this PR)** — the group is named **`Building`**, and the shell receives nothing. The bootstrap line still prints; it just no longer steals focus.

**One clip, both takes** (red title card, BEFORE, green title card, AFTER):

<video src="https://raw.githubusercontent.com/maxmilian/warp/assets-14241/assets/14241-before-after.mp4" controls width="900"></video>

Key frames, in case the player does not load:

**Before — `master`**

before: tab group committed as "Build", "ing" ran in the shell

**After — this PR**

after: tab group named "Building", shell untouched

<sub>Media is served from a branch on my fork (`maxmilian/warp@assets-14241`) because the `gh` CLI cannot upload to `user-attachments`. Say the word and I will re-attach it through the web editor so it lives with the PR permanently.</sub>

## Agent Mode

CHANGELOG-BUG-FIX: Fixed a new tab group's terminal stealing focus while its name was being typed, which saved the partially-typed name.


maxmilian:
Both review points addressed in `1aeae3a`.

**1. Unregistered singleton in the shared terminal test setup.** Confirmed: any terminal-view test that reaches `ModelEvent::VisibleBootstrapBlock` panicked with `Cannot get singleton model "warp::workspace::inline_rename_state::InlineRenameState" that was never registered`. `app/src/test_util/terminal.rs::initialize_app_for_terminal_view` now registers it alongside `AltScreenReporting`. Two new tests in `terminal/view_tests.rs` cover the path from the terminal side rather than just unblocking it — `visible_bootstrap_block_focuses_terminal_under_shared_test_setup` (reproduced the panic before the fix) and `visible_bootstrap_block_leaves_focus_alone_during_inline_rename` (a bootstrapping terminal must not take focus back while the rename editor has it).

**2. Stale comments describing blur as committing.** Fixed in all three places, not just the one flagged: the `VisibleBootstrapBlock` comment in `terminal/view.rs`, the `inline_rename_state` module doc, and the doc comment on `test_tab_group_rename_blur_does_not_commit_unfinished_name`. The pre-fix behaviour is now stated in the past tense where it is still useful context for why the singleton exists.

Also ran `cargo fmt` — the new import and module lines were out of sort order on the previous head.

Verification: `cargo test -p warp --lib visible_bootstrap_block` 3 passed, `workspace::view` 160 passed, `cargo clippy -p warp --all-targets --tests -- -D warnings` clean, `cargo fmt -- --check` clean.

bholmesdev (approved):
Changed the implementation to use a workspace helper rather than tracking editing/blurred state in a separate struct.

PR #14904, opened 2026-08-10
[QUALITY-1449] Add per-model reasoning effort to custom endpoints

JaidenRatti:
> [!NOTE]
> **Depends on `warpdotdev/warp-proto-apis#354`.** This branch is temporarily pinned to that PR's committed revision (`ac904581ba22`) so CI can compile the new field before the proto PR merges. Replace the revision with the merged commit before final merge.

https://www.loom.com/share/2d6ff18e87c44e4e8aabdc68e7f3c4db

## Description

**What:** Adds optional per-model reasoning effort to personal custom endpoints, persists it locally, serializes it onto the multi-agent request, and exposes a closed picker for OpenAI Responses endpoints.

**Why:** Warp-hosted models encode effort in their internal model ID. Custom models all share `InternalCustomEndpointModel`, so effort must travel as per-model metadata. The matching server behavior is in `warpdotdev/warp-server#14222`.

**How:**

- `CustomEndpointModel` stores `reasoning_effort: Option<String>` in the existing secure-storage blob.
- Request construction serializes it onto `CustomModel.reasoning_effort`.
- The modal shows the picker only when the endpoint schema is OpenAI Responses.
- `Provider default` stores `None`; explicit `none` remains a distinct value.
- The picker is limited to the values recognized by the OpenAI Responses API: `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, and `max`.
- The picker popup is attached to the modal's outer overlay stack so options overlapping Cancel/Save own the click instead of activating the button underneath.

### Compatibility

`CustomEndpointModel` uses `#[serde(default)]`, so endpoints saved before this change deserialize with effort unset. The persisted and wire representations remain strings for forward compatibility; the UI is intentionally a closed set. Unknown values already stored by an earlier build fall back to `Provider default` when that endpoint is edited.

Switching away from OpenAI Responses clears effort on save because the server does not apply the setting to Chat Completions or Anthropic Messages.

## Testing

- `crates/ai/src/api_keys_tests.rs`
  - exact supported option set
  - legacy secure-storage records default to unset
  - reasoning effort round-trip and proto serialization
- `app/src/settings_view/custom_inference_modal_tests.rs`
  - closed picker selection and clearing
  - unknown-value fallback
  - external popup extraction
  - row layout, scrolling, schema selection, and URL validation
- Existing model/conversation fixtures remain covered by their updated `reasoning_effort: None` values.

Local validation:

- `cargo test -p ai api_keys` — 75 passed
- `cargo test -p warp custom_inference_modal` — 20 passed
- `cargo clippy -p ai -p warp --lib --tests -- -D warnings`

### Manual verification

Configured `low`, then `xhigh`, on an OpenAI Responses custom endpoint (`openai/gpt-5.6-sol` through OpenRouter). The server-generated request body carried the selected effort, OpenRouter's `echo_upstream_body` showed the effort forwarded to the upstream provider unchanged, and the responses reported 0 versus 158 reasoning tokens. The server PR contains the full request and encrypted-reasoning replay trace.

## CI failure addressed

The previous CI run used proto revision `b0886a9523e2`, where `CustomModel` contained only `slug` and `config_key`. Every platform's formatter/Clippy, tests, release build, eval build, and wasm build failed while compiling the same missing `reasoning_effort` field. This branch now pins `warp_multi_agent_api` and `Cargo.lock` to `ac904581ba22`, the proto PR revision that defines the field.

## Out of scope

- Reasoning mode
- Applying effort to Chat Completions or Anthropic Messages


crates/ai/src/api_keys.rs
```diff
@@ -25,6 +25,10 @@ use crate::telemetry::{
 
 const SECURE_STORAGE_KEY: &str = "AiApiKeys";
 
+/// Reasoning effort values supported by OpenAI Responses models.
+pub const CUSTOM_ENDPOINT_REASONING_EFFORTS: &[&str] =
```
danielpeng2: mind mentioning in the name of this variable that these are for the OpenAI responses API

danielpeng2 (approved):

crates/ai/src/api_keys.rs
JaidenRatti: will do!

PR #14926, opened 2026-08-11
Group tabs by project automatically

### LINKED ISSUE #9875: Group vertical tabs by project or repo root
## Problem

When vertical tabs are enabled, sessions from the same project can be scattered across the tab list. This makes it harder to scan active work, especially when several tabs or panes belong to the same repository or working directory.

## Desired behavior

Add an optional setting to group vertical tabs by detected project root. When enabled, Warp should render lightweight project section headers in the vertical tabs list, using the detected git repo root when available and falling back to the tab working directory.

Unknown tabs should remain independent rather than all collapsing into one "Other" group.

## Scope

- Vertical tabs only.
- Default off.
- Works in Tabs and Panes display modes.
- Uses detected git repo roots when available.
- Falls back to cwd/path.
- Handles Windows drive paths and WSL `/mnt/<drive>/...` normalization.
- Preserves existing tab and pane colors.
- Keeps drag/drop behavior predictable.

## Non-goals

- Collapsible groups.
- Explicit multi-root Project Workspaces.
- Per-project settings.
- CLI agent identity/icon fixes.

@oss-maintainers I have a focused implementation ready for this. Would you prefer this to go through a spec PR first, or is this small enough to mark ready-to-implement?

bobrnor:
## Description

Groups tabs by project automatically. When the mode is on, a tab joins the tab group for the
project its directory belongs to, follows that project as the directory changes, and leaves the
tab alone the moment the user places it by hand.

A project is a repository's shared git directory when the directory is inside a checkout — so
every worktree of one repository unifies into a single group — and otherwise the directory
itself, joined by longest prefix. Group names are the project's name, qualified with one parent
segment when two projects in the same window would read alike (`services/api` + `vendor/api`).

The mode is off by default, behind the `AutoTabGrouping` feature flag, and toggleable from
Settings → Appearance → Tabs or from any tab's context menu.

What automation deliberately does **not** do:

- It never touches a pinned tab, never collapses or expands a group the user set, never renames a
  group the user named, and never dissolves a group when the mode is turned off (R3).
- A tab the user drags or "Move to group"s somewhere else is detached and stays detached until it
  is put back into the group its own project owns (R13/R14). Tracked-ness is derived from where a
  tab sits rather than stored, so a manual placement writes no automation state at all.
- Repository detection is asynchronous, so an unresolved project is a third state, distinct from
  "not a repository": a tab whose key has not resolved yet is left where it is and placed when
  detection answers (R21).

### Colored tabs

A second toggle beside the first gives every tab in an automatic group a color derived from a
stable hash of that group's project key, so one repository always reads the same color — across
windows and across restarts — while different projects usually differ. Also off by default,
behind the same flag.

The color lands on the **tabs**, not on the group container. A group's own color stays entirely
the user's to set; automation never writes it. Both tab bars already render a member's color
independently of its group's — `compute_tab_group_color_mode` treats a tab's `selected_color` as
an override for the whole tab "whether or not it's in a group", and the horizontal bar passes
`tab.color()` per member — so nothing in the rendering plane changed.

The color is a pure function of the key, never of the window's contents. Two checkouts of one
repository share a key and so share a color; the display name does not feed the hash, so a
collision qualifying `api` into `services/api` leaves the color alone. The palette is the existing
six-color tab picker and nothing spaces the hash out, so two projects collide about one time in
six. De-colliding within a window was rejected: it would make a project's color depend on what
else happens to be open, which is the one property the feature is for.

A tab follows its project's color across every transition that changes which project it belongs
to — joining a keyed group, being put in a new one, a sole-member re-key (every member), adopting
a key for a replacement group, and being born straight into its project's group. The paths that
take a tab *out* of a group hand the color back: removing it from the group, ungrouping, dragging
it out of every group, and pinning. Pinning matters more than it looks: a pinned tab has no group,
so nothing can judge its color's provenance, and a color carried across the pin would be frozen as
the user's for the rest of the tab's life.

Color provenance is derived rather than stored, exactly as name provenance already is (R16), but
judged per tab against the key of the group the tab is **leaving**: a color matching what that key
derives was automation's and may be replaced, and anything else was the user's and may not.
`Cleared` is protected too. That key is read off the group rather than off the resolver's
in-memory `last_resolved_keys`, because after a restart every colored tab would otherwise read as
the user's and never follow its directory again.

The hash is FNV-1a, spelled out rather than taken from `std::hash`: `DefaultHasher` is explicitly
not stability-guaranteed across releases and `HashMap`'s hasher is seeded randomly per process,
either of which would repaint everything on restart. `rustc-hash`, the workspace's other hasher,
gives the same no-stability guarantee. A golden test pins specific keys to specific colors, and a
build assertion pins the palette's length, because both are now part of what every stored color
was derived from.

**Scope note.** Automatic coloring is *not* one of #9875's requirements. R16 and R18 anticipate a
group's color as a value automation must preserve and persist, but nothing in the spec asks
automation to assign one — this is an addition on top, constrained by those two requirements. It
is a separate commit (`feat(tabs): colour the tabs in an automatic group, not the group`) and can
be dropped without touching the rest of the branch if maintainers would rather ship the grouping
alone.

### Groundwork

`repo_metadata`'s ancestor-`.git` search returned `None` for a bare layout
(`/repos/api.git/worktrees/feature` has no `.git` path component), which scattered such worktrees
across separate groups. `derive_common_git_dir` now derives the common git directory from the
`worktrees/` layout directly.

## Linked Issue

Closes #9875.

- [x] Screenshots or a short video of the implementation are included below.

**This box is deliberately unchecked.** #9875 carries neither label today. Per `CONTRIBUTING.md`
the spec comes first, so `specs/GH9875/product.md` and `specs/GH9875/tech.md` are included in this
branch and the implementation is offered alongside them rather than ahead of them. If maintainers
would rather review the spec on its own, the first commit stands alone and the rest can be split
out. Requesting the re-triage is left to a maintainer — this branch takes no action on the issue.

## Testing

Automated, all added in this branch alongside the code:

- `workspace::project_key` — the resolver in isolation: git identity vs. directory identity,
  longest-prefix joining, worktree unification, name derivation and collision qualification. For
  color: determinism, worktree agreement, independence from name qualification, a golden test
  pinning four keys to four colors, and palette spread.
- `workspace::view::auto_grouping::tests` — the reconcile decision table driven directly with
  resolved keys: tracked tab follows its key, detached tab is left alone, re-key in place vs. new
  group, contiguity after every move. For color: creation and re-key take the project's color, a
  color the user set or cleared survives both, nothing is painted while either setting is off, the
  sweep paints existing keyed groups but not manual ones, and the three replacement cases where a
  group is destroyed and another takes over its project.
- `workspace::view::auto_grouping::wiring_tests` — which events reconcile and which deliberately
  do not (a pane split fires the same event but must not move the tab), the enable sweep,
  anchor-pane re-anchoring, restore, and cross-window insertion. Color reaching the workspace
  through the real settings subscription is covered in the decision-table module.
- `workspace::view::auto_grouping::manual_override_tests` — the drag and menu paths that detach a
  tab, and the path back under automation.
- `settings_view::appearance_page` — both grouping rows are reachable by search, the color row
  hides while the mode is off, and a page-level query still reaches both.
- `persistence` — the migration and the round-trip of `tab_groups.project_key` and
  `tabs.placed_by_automation`.

`cargo nextest run -p warp` is 6402 tests green and `cargo clippy -p warp --all-targets --tests
-D warnings` is clean.

`./script/presubmit` runs 11,023 tests, of which 11,016 pass here. All seven that do not are in
`integration::integration` and are unrelated to this branch: five SSH tests
(`test_ssh_into_ash`/`_sh`, `test_ssh_wrapper_into_bash`/`_zsh`, `test_ssh_with_shell_override`)
tunnel into a Warp-internal GCP host via `gcloud compute start-iap-tunnel`, which an outside
contributor cannot reach, so they time out at 60s by construction; the other two (`test_ctrl_c`,
`test_escape_sequences_sent_to_focused_terminal`) pass on re-run in isolation at the same commit,
so they are flaky under the full run's parallel load rather than broken. CI is the authority on
those seven.

Left unchecked pending the author's own pass on the release build; the automated suite above is
what has actually been run.

### Screenshots / Videos

Three tabs across two projects: two of them share `warp-client`, one sits in `billing-api`. Left:
automatic grouping off, where the two same-project tabs are separated by the third. Right: on — each
project owns a group, and the colored-tabs toggle gives every member the color derived from that
project's key.

Automatic tab grouping, before and after

Both halves are the same scenario on the same base commit (`origin/master` 5071a868c); the left one
is a build without this branch. Full window:
[before](https://raw.githubusercontent.com/bobrnor/warp/pr-media/media/14926-full-window-before.png) ·
[after](https://raw.githubusercontent.com/bobrnor/warp/pr-media/media/14926-full-window-after.png).

## Agent Mode

## Known limitations

Grouping:

- Contiguity of a group's members is maintained by convention rather than enforced at runtime, in
  line with the existing grouping code.
- Reconcile gates on `tab.pinned`, while the positional predicate `is_tab_effectively_pinned` also
  counts a pinned *group*, so a `cd` can move a tab out of a group the user pinned.
- The first visit to an undetected repository from a non-git directory keys on the directory until
  `RepoChanged` re-keys it.
- Anchor-pane identity is not persisted across a restart.
- Re-qualification runs when a key is added, not when one disappears: dissolving one of two
  colliding groups leaves the survivor qualified until the next keying event in that window.

Color:

- Provenance is derived, so a user who picks precisely the color their key hashes to is read as
  automation and repainted by the next move. The name heuristic has the identical hole; the cost
  is one color the user can set again, which was judged cheaper than persisting a provenance flag
  no other automation state needs.
- A tab that changed project while coloring is off keeps a color derived from the key it used to
  sit under. That color then reads as the user's, so it sticks permanently, and the only way back
  under automation is to clear it by hand. Erring this way is deliberate — the alternative
  mistakes a user's color for automation's, which R16 forbids outright.
- Provenance is a property of the tab, not of the project: clearing one tab's color says "not this
  tab", not "never color this project", so another tab in the same project is still colored.
- `derived_color` is pure in the project key, and a non-git key depends on which directory groups
  already exist in that window, so two windows can color the same directory differently. Git
  projects, the common case, are stable.
- The color toggle is registered in the command palette whenever the mode is available, while its
  Settings row hides when the mode is off — so it can be flipped early, with no visible row and no
  effect until grouping is turned on.


bobrnor:
@cla-bot check

bobrnor:
@oss-maintainers this PR has never had a readiness call, so flagging it the way CONTRIBUTING suggests.

It's linked to #9875, which carries no readiness label. #9875 is a feature request with `needs-mocks`, so the path runs through `ready-to-spec`, and I'm not asking to skip it: `specs/GH9875/product.md` and `specs/GH9875/tech.md` are already the first commit here, which is the shape CONTRIBUTING describes when implementation continues on the spec PR. The description says the same and offers to split them out. So the only thing needed is the label on #9875.

`needs-mocks` is the part I can't clear myself. The settings toggle and the tab-menu entry reuse the existing tab-settings widgets rather than introducing new UI, and the three screenshots in the description show what it looks like. If the label is only about visual sign-off, that may be enough to drop it.

Rebased onto master today; the branch was 103 commits behind. `./script/format` is clean, `cargo clippy -p warp -p warp_features -p repo_metadata --all-targets --tests -- -D warnings` is clean, and `cargo nextest run -p warp --lib -p warp_features -p repo_metadata` runs 6547 tests with one unrelated timing flake in `notebooks::editor::model`, which passes on its own.

bobrnor:
Rebased onto master (`e5db969ae`); the branch was 52 commits behind. Two conflicts, both from unrelated work landing next to this one, neither changing the diff's behaviour:

- #15587 (stored screenshot references) appended to the same `DOGFOOD_FLAGS` list this branch adds `AutoTabGrouping` to. Kept both.
- #15754 (WASM conversation-viewer chrome) appended a test module to the end of `app/src/workspace/view_tests.rs`, where this branch appends its own tab-menu tests. Kept both.

The second one moves hunks around in `appearance_page.rs`, so I checked the result rather than trusting the merge: every widget master builds into that page is still there, and this branch adds exactly `AutoGroupTabsWidget` and `AutoGroupTabColorsWidget` to it.

`./script/format --check` is clean, `cargo clippy -p warp -p warp_features -p repo_metadata --all-targets --tests -- -D warnings` is clean, and `cargo nextest run -p warp --lib -p warp_features -p repo_metadata` runs 6711 tests green (9 skipped).

The readiness ask from 27 August is unchanged and still the only thing blocking review: #9875 carries no readiness label, and `specs/GH9875/product.md` and `specs/GH9875/tech.md` are the first commit on this branch, which is the shape CONTRIBUTING describes when implementation continues on the spec PR. `needs-mocks` is the part I cannot clear myself — the settings toggle and the tab-menu entry reuse the existing tab-settings widgets rather than introducing new UI, and the screenshots in the description show what it looks like.

`no-autoclose` here would stop the stale clock while that is decided.

PR #14930, opened 2026-08-11
Center the agent icon in the footprint render_icon_with_status reserves

### LINKED ISSUE #14958: Agent avatar renders flush top-left in its reserved box, widening the title gap and pushing the status badge off the circle
## Summary

`render_icon_with_status` reserves a `total_size` square for an agent avatar, but paints the brand circle at that box's origin instead of centered in it. The circle is only `CIRCLE_RATIO` (0.76) of the box, so the remaining 24% collects as dead space on the right of the artwork, and the status badge — which `corner_overlay_offset` anchors to the *box's* bottom-right corner — lands outside the circle's edge.

Two visible consequences, both from the same origin error:

- **The gap before the title is too wide.** In the vertical tabs sidebar the space between the avatar and the row title reads around 14px, where `ICON_WITH_STATUS_GAP` asks for 8. The extra is the unpainted right-hand slice of the reserved box.
- **The status badge hangs off the avatar.** The badge's offset is documented as putting it inside the circle's edge, and that is true only when the circle is centered in the box. Left-aligned, the badge's center falls outside the circle.

Every surface that draws an agent avatar goes through this component: the vertical tabs sidebar, the conversation list, agent notifications, ambient agent blocks, and pane rows.

## Repro

1. On macOS, turn on vertical tabs (`Settings → Appearance → Tabs`).
2. Start a CLI agent in a pane so its row draws an agent avatar with a status badge.
3. Look at the row: the space between the avatar and the title is wider than the configured 8px, and the status badge sits past the lower-right edge of the brand circle rather than inside it.

## Expected

The brand circle is centered in the footprint the component reserves, so the gap the caller asks for is the gap that renders, and the status badge sits inside the circle's edge as `CIRCLE_RATIO`'s comment describes.

## Root cause

Neither wrapper in the chain centers a child:

- `ConstrainedBox::layout` only tightens the constraint it passes down, then paints the child at the origin it was handed.
- `Stack` takes its size from `constraint.min` but likewise paints non-positioned children at its own origin.

So the reserved footprint is correct and the artwork inside it is not.

## Evidence

Rendering the component into a headless presenter and reading the brand circle's painted rect straight out of the scene:

| | painted origin of the brand circle |
|---|---|
| measured | `(0.0, 0.0)` |
| expected | `(2.88, 2.88)` |

`2.88` is `24 * (1 - 0.76) / 2` — the circle centered in a 24px footprint.

## Related

`warpdotdev/warp#14893` hit the same defect in the orchestration pill bar and worked around it locally, centering inside its own `CustomAvatar` element and leaving `icon_with_status.rs` a comments-only diff. That avatar already fills the whole `total_size` box, so centering it there is a no-op — the shared component still mis-centers for every other caller.

## Environment

- Warp: `v0.2026.08.05.09.03.01`, also reproduced against a build from source on `master`

bobrnor:
## Description

Every surface that draws an agent avatar — the vertical tabs sidebar, the conversation list, agent notifications, ambient agent blocks, pane rows — has been rendering it flush against the top-left of the box it reserves, instead of centered in it. Two things fall out of that, and both are visible:

- **The trailing gap is wrong.** `render_icon_with_status` reserves a `total_size` square, but the brand circle is only `CIRCLE_RATIO` (0.76) of it. Left-aligned, the remaining 24% piles up as dead space on the right, on top of whatever gap the caller already asked for. In the vertical tabs sidebar that reads as ~14px between the avatar and the title where `ICON_WITH_STATUS_GAP` asks for 8.
- **The status badge falls off the circle.** `corner_overlay_offset` anchors the badge to the *box's* bottom-right corner. That geometry only puts the badge inside the circle's edge — which is what the comment on `CIRCLE_RATIO` says it is for — while the circle is centered in the box. Left-aligned, the badge's center lands outside the circle and the badge reads as hanging off the avatar.

The cause is that neither `ConstrainedBox` nor `Stack` centers a child: `ConstrainedBox::layout` only tightens the constraint it passes down and then paints the child at the origin it was handed, and `Stack` takes its size from `constraint.min` but likewise paints non-positioned children at its own origin. So the reserved footprint was correct and the artwork inside it was not. Routing the circle through an `Align` fixes both symptoms at once, and makes the two comments that already describe the intended geometry true.

Reported against the vertical tabs sidebar; the fix is in the shared component, so it lands everywhere.

## Linked Issue

Closes #14958.

- [ ] The linked issue is labeled `ready-to-implement`. — #14958 was filed alongside this PR and has not been triaged yet.
- [ ] Screenshots or a short video of the implementation are included below. — this is a visual change and before/after screenshots are owed; see the note under Testing for why they are not here yet.

## Related

`warpdotdev/warp#14893` hit the same defect in the orchestration pill bar and worked around it there, centering inside its own `CustomAvatar` element and leaving `icon_with_status.rs` as a comments-only diff. That PR's avatar already fills the whole `total_size` box, so centering an element the size of its container is a no-op and the two changes compose without double-centering.

## Testing

`brand_circle_is_centered_in_the_reserved_footprint` renders the component into a headless presenter and reads the painted rect straight out of the scene, so this is measured rather than eyeballed:

| | painted origin of the brand circle |
|---|---|
| before | `(0.0, 0.0)` |
| after | `(2.88, 2.88)` |

`2.88` is `24 * (1 - 0.76) / 2` — the circle centered in a 24px footprint. The test fails on the pre-fix code with exactly that diff, so it pins the behavior rather than passing vacuously.

`./script/format`, `cargo clippy -p warp --all-targets --tests -- -D warnings`, and `cargo nextest run -p warp --lib icon_with_status` (4/4) are clean.

Not checked deliberately: `./script/run` builds the OSS channel without internal channel-config access, and a fresh instance has no agent conversation to draw an avatar for, so it cannot show this surface. The scene-level assertion above is the evidence; a look at the sidebar on a build with real agent panes is still worth having before merge.

### Screenshots / Videos

_To be attached._

## Agent Mode

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/app/src/ui_components/icon_with_status.rs
+++ b/app/src/ui_components/icon_with_status.rs
@@ -4,8 +4,8 @@ use warp_core::ui::icons::Icon as WarpIcon;
 use warp_core::ui::theme::color::internal_colors;
 use warp_core::ui::theme::{ColorScheme, Fill as WarpThemeFill, WarpTheme};
 use warpui::elements::{
-    ChildAnchor, ConstrainedBox, Container, CornerRadius, Element, OffsetPositioning, ParentAnchor,
-    ParentElement, ParentOffsetBounds, Radius, Stack,
+    Align, ChildAnchor, ConstrainedBox, Container, CornerRadius, Element, OffsetPositioning,
+    ParentAnchor, ParentElement, ParentOffsetBounds, Radius, Stack,
 };
 
 use crate::ai::agent::conversation::{ConversationStatus, StatusColorStyle};
@@ -308,6 +308,18 @@ fn render_circle(
         .finish()
 }
 
+/// Reserves the caller's full `total_size` square for `circle` and centers it inside.
+/// `Stack` and `ConstrainedBox` paint children at the origin rather than centering, so
+/// without the `Align` the circle — only `circle_size(total)` wide — sits flush against
+/// the box's top-left, and the corner overlay (anchored to the box's bottom-right by
+/// `corner_overlay_offset`) lands clear of the circle instead of tucked into its edge.
+fn circle_centered_in_box(circle: Box<dyn Element>, total_size: f32) -> Box<dyn Element> {
+    ConstrainedBox::new(Align::new(circle).finish())
+        .with_width(total_size)
+        .with_height(total_size)
+        .finish()
+}
+
 /// Builds the neutral circle: a full-`total_size` container with the glyph at
 /// `NEUTRAL_GLYPH_RATIO * total_size`. Used for non-agent surfaces (plain terminal,
 /// code, file tabs, etc.) which have no status overlay and therefore should fill the
@@ -412,12 +424,7 @@ fn render_with_cloud_status_badge(
     };
 
     let cloud_offset = corner_overlay_offset(total_size, overlay_extra_overhang_ratio);
-    let mut stack = Stack::new().with_child(
-        ConstrainedBox::new(circle)
-            .with_width(total_size)
-            .with_height(total_size)
-            .finish(),
-    );
+    let mut stack = Stack::new().with_child(circle_centered_in_box(circle, total_size));
     stack.add_positioned_child(
         cloud_with_status,
         OffsetPositioning::offset_from_parent(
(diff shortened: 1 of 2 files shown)


bobrnor:
Closing this one.

I built both sides and compared the rendered rows rather than the geometry. Centering the circle inside the reserved box moves it 2.88pt toward the status badge, and the badge is anchored to the *box's* bottom-right corner rather than the circle's edge — so centering makes the badge ride onto the avatar instead of sitting at its edge. Keeping the badge's current relationship to the circle while centering does not fit either: 2.88 + 18.24 + 4.56 = 25.68pt of artwork in a 24pt box.

The shipped placement reads better, and the trailing-gap symptom is not worth that regression. Closing #14958 along with it.

PR #14931, opened 2026-08-11
Measure the vertical-tabs row title/indicator gap in an integration test

### LINKED ISSUE #14959: No test guards a vertical-tabs row title against overlapping its trailing indicators
## Summary

`render_row_title_line` pins a vertical-tabs row's trailing indicators to the right edge of the title line and separates them from the title with a 4px margin. Nothing verifies that the title actually stays out of the indicator's box.

**The current layout is correct** — measured clearance is 3.38px between the rightmost painted title glyph and the indicator's left edge on a 232px-wide row. This is a missing regression guard, not a live defect.

## Why unit tests cannot cover it

The headless unit-test platform's `layout_line` returns an empty line, so text has no width there at all. Any assertion about where a title's glyphs land passes vacuously in that environment — it would report clearance on a row whose label never rendered. A guard for this has to measure a real frame.

## Impact

The row's geometry is a product of several independently-owned pieces: the title line's padding, the set of trailing indicators (which grows — the synchronized-inputs link icon is conditional), and the clip fade that decides which glyphs are actually painted. A change to any one of them can push the title under an indicator, and today nothing would catch it. The failure mode is a title glyph painted beneath an icon, which is easy to miss in review and awkward to notice in use.

## What a guard needs to do

Measure from the built scene rather than a screenshot:

1. take the row rect from the position cache (`vertical_tabs:pane_row:…`),
2. pick the rightmost `Icon` inside that rect as the trailing indicator,
3. walk every glyph on the indicator's line, resolving each glyph's advance through the font cache to get its painted right edge,
4. skip glyphs the clip fade has already driven to zero alpha — `Line::paint` builds an end-direction fade as `start = right edge − fade_width`, `end = right edge`, and the glyph shader ramps alpha 1 → 0 across that span, so anything at or past `end` is not painted.

It also needs a guard against passing vacuously when no title glyphs are found at all.

Two setup details that are easy to get wrong: the row must be in summary mode, which is not a display granularity of its own but the `Tabs` granularity plus the `Summary` tab-item mode — `resolve_vertical_tabs_mode` silently falls back to `FocusedSession` unless `FeatureFlag::VerticalTabsSummaryMode` is on as well. And the label should be ASCII, because the integration harness's synthetic keystrokes mangle non-ASCII input; the geometry under test does not depend on the script.

## Environment

- Warp: built from source on `master`

bobrnor:
`render_row_title_line` pins a vertical-tabs row's trailing indicators to the right edge of the title line and separates them from the title with a 4px margin. Nothing verified that the title actually stays out of the indicator's box — and the headless unit-test platform cannot verify it, because its `layout_line` returns an empty line, so text has no width there at all.

This adds an integration test that measures the real frame.

## Linked Issue

Closes #14959.

- [ ] The linked issue is labeled `ready-to-implement`. — #14959 was filed alongside this PR and has not been triaged yet. It reports a missing regression guard rather than a user-visible defect; if that is not something the project wants tracked as an issue, say so and I will close both.

## How it measures

The test puts vertical tabs in summary mode, gives the row a label far wider than the row, and turns on synchronized inputs so the row grows its trailing link icon. Then it reads the built scene instead of a screenshot:

1. takes the row rect from the position cache (`vertical_tabs:pane_row:…`),
2. picks the rightmost `Icon` inside that rect as the trailing indicator,
3. walks every glyph on the indicator's line, resolving each glyph's advance through the font cache to get its painted right edge.

Glyphs the clip fade has already driven to zero alpha are skipped — `Line::paint` builds an end-direction fade as `start = right edge − fade_width`, `end = right edge`, and the glyph shader ramps alpha 1 → 0 across that span, so anything at or past `end` is not painted.

## What it measures today

| quantity | value |
| --- | --- |
| row | x 9…241 |
| rightmost label glyph edge | 216.62 |
| indicator left edge | 220 |
| **clearance** | **3.38px** |

That is the intended margin, so the layout is correct as it stands and the test passes. It fails if any glyph reaches past the indicator's left edge, and the failure names the indicator box, the row, and the offending glyph coordinates, so it is diagnosable by someone who has never seen the test.

## Notes

- The label is ASCII: the harness's synthetic keystrokes mangle non-ASCII input (typing Cyrillic produced a stray leading character). The geometry under test does not depend on the script.
- Summary mode is not a display granularity of its own — it is the `Tabs` granularity plus the `Summary` tab-item mode, and `resolve_vertical_tabs_mode` silently falls back to `FocusedSession` unless `FeatureFlag::VerticalTabsSummaryMode` is on too. The helper sets all three.
- An explicit guard fails the assertion when no title glyphs are found, so it cannot pass vacuously on a row whose label never rendered.

## Testing

- `cargo run -p integration --bin integration -- test_vertical_tab_title_does_not_overlap_trailing_indicator` — passes, both with and without `WARPUI_USE_REAL_DISPLAY_IN_INTEGRATION_TESTS=1`, which is why it is listed in `ui_tests.rs` un-ignored.
- `cargo nextest run --no-fail-fast -p integration test_vertical_tab_title_does_not_overlap_trailing_indicator` — `1 passed`.
- Failure path verified deliberately: temporarily demanding 10px of clearance where only 3.38 exists produced

  ```
  the vertical-tabs row title is painted into its trailing indicator. Indicator box
  RectF(<220, 86.4, 232, 98.4>), row RectF(<9, 68, 241, 135.2>), rightmost title edge 216.62,
  2 of 40 title glyphs on that line reach past the indicator's left edge:
  ["x=203.53..210.26 y=88.30", "x=210.26..216.62 y=88.30"]
  ```

### DIFF
--- a/crates/integration/src/bin/integration.rs
+++ b/crates/integration/src/bin/integration.rs
@@ -362,6 +362,7 @@ fn register_tests() -> HashMap<&'static str, BoxedBuilderFn> {
     register_test!(test_tab_context_menu_copies_metadata);
     register_test!(test_vertical_tab_context_menu_copies_metadata);
     register_test!(test_vertical_pane_context_menu_copies_metadata);
+    register_test!(test_vertical_tab_title_does_not_overlap_trailing_indicator);
 
     register_test!(test_focus_panes_on_hover);
 
--- a/crates/integration/src/test/workspace.rs
+++ b/crates/integration/src/test/workspace.rs
@@ -29,10 +29,13 @@ use warp::integration_testing::workspace::{
 };
 use warp::settings::PaneSettings;
 use warp::terminal::shell::ShellType;
-use warp::workspace::tab_settings::{TabSettings, VerticalTabsDisplayGranularity};
+use warp::workspace::tab_settings::{
+    TabSettings, VerticalTabsDisplayGranularity, VerticalTabsTabItemMode,
+};
 use warp::workspace::{NEW_TAB_BUTTON_POSITION_ID, WorkspaceAction};
 use warpui_core::event::{Event, ModifiersState};
 use warpui_core::integration::{AssertionCallback, AssertionOutcome, StepDataMap, TestStep};
+use warpui_core::scene::GlyphFade;
 use warpui_core::windowing::WindowManager;
 use warpui_core::{SingletonEntity, TypedActionView, WindowId, async_assert, async_assert_eq};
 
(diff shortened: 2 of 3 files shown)


bobrnor:
Withdrawing this one.

There is no defect behind it: the measured clearance (3.38px) is the intended margin and the layout is correct as it stands, so this is a standalone regression guard with no user-visible problem attached — not the right shape for a PR against the review queue here.

It is also measuring a layout that is about to move. `render_row_title_line` and `render_icon_with_status` are both used by the same vertical-tabs row in `app/src/workspace/view/vertical_tabs.rs`, and #14930 centers the brand circle in its reserved footprint, which narrows the leading gap from roughly 14px to the 8px `ICON_WITH_STATUS_GAP` asks for. That hands the title line about 6px more room, so the glyph extents this test records would go stale the moment that fix lands.

The measurement is worth keeping as a guard for exactly that change — a wider title running into the trailing indicator is the risk #14930 introduces — so I would rather fold it into that PR than carry it separately. Branch is kept.

Closing #14959 alongside.

Issue #10784, opened 2026-05-13
Fullscreen top bar reveal pushes tmux status line out from under cursor, making tabs unclickable


CastIronPlatypus:
## Summary

In macOS fullscreen, moving the cursor toward the tmux status line at the top of the window triggers Warp's own top bar (tabs/title bar) to slide down, which pushes the terminal content (including the tmux window tabs) below the cursor. Moving the cursor down to chase the tab causes the Warp top bar to retract, putting the tab back at the top, where moving the cursor up to click it triggers the top bar again. The result: tmux window tabs are effectively unclickable in fullscreen.

## Repro

1. Fullscreen Warp on macOS.
2. Run tmux with the status line at the top showing multiple windows (`set -g status-position top`).
3. Move the cursor up to click a window tab.
4. Warp's top bar slides down, content shifts, tab is no longer under the cursor.
5. Move the cursor down to follow it; top bar retracts; cycle repeats.

## Expected

Either Warp's top bar overlays content without pushing it, or doesn't reveal when the cursor is over interactive terminal content.

## Environment

- Warp: `v0.2026.05.06.15.42.stable_04`
- macOS: 26.4.1 (25E253), Darwin 25.4.0
- Hardware: MacBook Pro (Mac14,10), Apple M2 Pro, 32 GB
- Not tested outside fullscreen.

hiwurun:
I have a similar problem.

PR #14947, opened 2026-08-11
Overlay the hover-revealed tab bar instead of reflowing content

### LINKED ISSUE #14945: Fullscreen tab bar flickers open and closed when the pointer rests at the top edge of the screen
## Summary

With the tab bar hidden in fullscreen, moving the pointer to the very top edge of the screen makes the tab bar flicker open and closed instead of settling. Because the revealed bar is laid out inside the content column, every flip reflows the terminal by the height of the bar, so the whole screen jumps up and down repeatedly. The flicker keeps going while the pointer sits still.

Affects the default `appearance.tabs.workspace_decoration_visibility = hide_fullscreen` (in fullscreen) and `on_hover` (always).

## Repro

1. On macOS, put Warp into fullscreen.
2. Keep the default `hide_fullscreen` decoration setting, so the tab bar is hidden.
3. Fling the pointer to the top of the screen so the cursor comes to rest against the screen edge.
4. The tab bar appears and disappears repeatedly, and terminal content jumps with it. The pointer does not need to keep moving.

Resting the pointer a few pixels lower — inside the reveal strip but not against the screen edge — is stable.

## Expected

Hovering the top edge reveals the bar once and holds it steady. An incidental sweep across the top of the window does not reveal it at all.

## Root cause

The point that *reveals* the bar is not inside the region that *keeps* it revealed.

- While hidden, `Workspace::render` adds a reveal strip as a positioned stack child anchored at the window's top-left with a zero offset, `TAB_BAR_HOVER_HEIGHT` (12) tall — so it covers `y ∈ [0, 12)`.
- Once hovered, `tab_bar_mode` returns `ShowTabBar::Stacked` and the bar is added to the content column, which is wrapped in `Container::with_uniform_padding(WORKSPACE_PADDING)` (1.0). The bar's own hover area therefore starts at `y = 1`.

Measured from the rendered scene in a unit test at a 1024x768 window:

| element | rect |
|---|---|
| reveal strip | `RectF(<0, 0, 1024, 12>)` |
| revealed tab bar | `RectF(<1, 1, 1023, 36>)` |

The topmost pixel row — and likewise the leftmost and rightmost columns — reveals the bar and then sits outside it.

In fullscreen the cursor clamps to `y = 0` at the screen edge, which is exactly that dead band. The loop then sustains itself with no mouse movement at all, because `AppContext::build_scene` replays the last pointer position as a synthetic `MouseMoved` after every frame, and the framework has no mouse-exit event that would invalidate the stored position:

```
pointer at y=0 -> strip hovered -> bar renders at y >= 1, pointer is no longer over it
  -> hover-out delay (500ms) -> bar hides -> strip is under the pointer again -> bar reveals -> ...
```

Separately, the reveal has no hover-in delay, so a quick sweep through the top 12 px is enough to trigger a full-height content reflow.

## Related

#10784 is the same mechanism seen from the other side: the reveal pushing content down makes a tmux top status line escape the cursor. A revealed bar that overlays content instead of taking a row in it addresses both.

## Environment

- Warp: built from source at `b0c7a7674` on `master` (`full_screen_zen_mode` is in the default feature set)

bobrnor:
## Description

In fullscreen, a hover-revealed tab bar flickered open and closed while the pointer stood still at the top edge of the screen, reflowing every pane by the height of the bar on each flip.

The point that *revealed* the bar was not inside the region that *kept* it revealed. The reveal strip is a positioned stack child anchored at the window's top-left, `TAB_BAR_HOVER_HEIGHT` (12) tall. The revealed bar was added to the content column, which is inset by `WORKSPACE_PADDING`, so it started one pixel lower. Measured from the rendered scene at a 1024x768 window:

| element | before | after |
|---|---|---|
| reveal strip | `RectF(<0, 0, 1024, 12>)` | unchanged |
| revealed bar | `RectF(<1, 1, 1023, 36>)` | `RectF(<0, 0, 1024, 35>)` |

The top pixel row revealed the bar and then sat outside it. In fullscreen the cursor clamps to `y = 0` at the screen edge, which is exactly that band, so the bar oscillated with no mouse movement at all: `build_scene` replays the last pointer position as a synthetic `MouseMoved` after every frame, and there is no mouse-exit event to invalidate that stored position.

This adds `ShowTabBar::Overlay`, used for every hover reveal. The bar is positioned at the window's top-left over the content instead of taking a row in the column, so it now covers a superset of the strip — the pointer that opened it stays over it — and revealing it no longer moves any pane. Two things follow from leaving the column: the overlay stretches itself to the full window width (a positioned stack child is laid out with a zero minimum width), and it paints its own background, flattened to a solid color because `theme.background()` may be a gradient whose stops would otherwise be squeezed into a 35 px strip.

Revealing now takes a 250 ms hover-in delay, so a sweep across the top of the window no longer pops the bar open. Hiding deliberately stays undelayed: `Hoverable` drops the redraw for a state change whose previous change was also synthetic, and with both delays set every transition resolves on the synthetic `MouseMoved` replayed after a frame — the bar would flip to hidden without scheduling the frame that repaints it. That constraint is recorded next to the constant.

## Linked Issue

Fixes #14945.

Related: #10784 is the same mechanism seen from the other side — the reveal pushing content down made a tmux top status line escape the cursor, so chasing it re-triggered the bar. This removes that chase: content no longer moves when the bar appears, and the 250 ms hover-in delay means a deliberate move to click the status line lands before the bar reveals at all. It is not a complete fix for that report, though — once the bar *is* revealed it sits over the terminal's top row, so a tmux status line there is covered while the bar is up. Leaving #10784 open for that remainder.

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. — #14945 was filed alongside this PR and has not been triaged yet.
- [x] Screenshots or a short video are included.

### Screenshots / Videos

Fullscreen, pointer parked at the top edge and not moved, with a colour grid filling the viewport so
any vertical movement is readable straight off the row numbers.

**Before** — revealing the bar reflows the whole column: the grid and the vertical-tabs sidebar jump
by the bar's height every time it flips.

Fullscreen hover reveal before the fix

**After** — the bar fades in over the content and nothing underneath moves.

Fullscreen hover reveal after the fix

Measured rather than eyeballed: taking the lower half of the screen — which the bar never covers —
and cross-correlating every consecutive frame pair at 10 fps gives

| | vertical content jumps | jump size |
|---|---|---|
| before | 16 | ±70px, i.e. the bar's 35pt at 2x |
| after | **0** of 132 frame pairs | — |

Full clips: [before](https://raw.githubusercontent.com/bobrnor/warp/pr-media/media/14947-before.mp4) · [after](https://raw.githubusercontent.com/bobrnor/warp/pr-media/media/14947-after.mp4). The "before" half is a
released build (0.2026.08.05) and the "after" is an OSS build carrying this branch, since the change
is not in any channel yet; the scenario and the machine are otherwise identical.

One thing the clips show that this PR does *not* claim to change: the macOS menu bar still slides in
whenever the pointer touches the very top of the screen. That is the system's own behaviour, and it
happens in both halves.

## Testing

Added `test_hover_revealed_tab_bar_covers_its_hover_strip`, which renders both states and asserts the invariant the bug violated: the revealed bar must cover the strip that reveals it. It also asserts the overlay stays a strip at the top rather than growing into the full-window constraint it is now laid out against.

Both assertions were verified to be capable of failing, not just passing:

- With `hovered_visibility` reverted to `Stacked`, the coverage assertion fails with `revealed tab bar RectF(<1, 1, 1023, 36>) must cover the hover strip RectF(<0, 0, 1024, 12>)` — this is where the measurements above come from.
- Tightening the height bound reports the overlay as `RectF(<0, 0, 1024, 35>)`, confirming it is a top strip and that the bound is not vacuous.

Existing coverage only asserted the `tab_bar_mode` enum, never the geometry, which is why this survived.

Also run: `cargo test -p warp --lib workspace::` (209 passed), the full `cargo test -p warp --lib` suite compared against a clean-tree baseline (identical failure set; the one extra failure floats between different tests in `secret_redaction` across runs on unchanged code), `cargo clippy -p warp --all-targets --tests -- -D warnings`, and `cargo fmt`.

**Not manually verified in the GUI** — this needs a human pass before merge. Worth checking specifically: the revealed bar looks right over a transparent background and under a gradient theme, hiding feels right now that it is immediate rather than delayed by 500 ms, and dragging a tab out of the revealed bar still behaves.

## Agent Mode

https://claude.ai/code/session_01CoefEcZFpauaaUcSVANoMn

### DIFF
--- a/app/src/workspace/view.rs
+++ b/app/src/workspace/view.rs
@@ -572,6 +572,18 @@ pub const TAB_BAR_HEIGHT: f32 = 34.;
 pub const PANEL_HEADER_HEIGHT: f32 = TAB_BAR_HEIGHT;
 /// The hover area height for states where the tab bar is revealed on hover.
 const TAB_BAR_HOVER_HEIGHT: f32 = 12.;
+/// How long the pointer has to rest in the hover area before a hidden tab bar is
+/// revealed. Without a delay, merely sweeping the pointer across the top of the
+/// window pops the bar open.
+///
+/// Do not pair this with a hover-out delay. A delayed transition resolves on the
+/// synthetic [`Event::MouseMoved`] the framework replays after each frame, and
+/// [`Hoverable`] drops the redraw for a state change whose previous change was
+/// also synthetic. With both delays every transition is synthetic, so hiding
+/// would flip the state without scheduling the frame that repaints it and the
+/// bar would linger until something unrelated redrew. Hiding therefore stays
+/// undelayed, resolving on the real mouse move that leaves the bar.
+const TAB_BAR_HOVER_IN_DELAY: Duration = Duration::from_millis(250);
 const TAB_BAR_PADDING_LEFT: f32 = 4.;
 const TAB_BAR_PADDING_RIGHT: f32 = 8.;
 const TITLE_BAR_SEARCH_BAR_MAX_WIDTH: f32 = 320.;
@@ -598,6 +610,13 @@ const THEME_CHOOSER_RATIO: f32 = 3.5;
 
 /// Save position for the tab bar.
 pub(crate) const TAB_BAR_POSITION_ID: &str = "workspace_view:tab_bar";
+/// Save position for the strip that reveals a hidden tab bar on hover. Recorded
+/// so the reveal strip and the revealed bar can be checked against each other.
+pub(crate) const TAB_BAR_HOVER_AREA_POSITION_ID: &str = "workspace_view:tab_bar_hover_area";
+/// Save position for the container holding a hover-revealed tab bar. Unlike the
+/// bar itself, this is the element laid out against the whole window, so its
+/// recorded size is what tells us the overlay stayed a strip at the top.
+pub(crate) const TAB_BAR_OVERLAY_POSITION_ID: &str = "workspace_view:tab_bar_overlay";
 const TEAM_SWITCHER_PILL_POSITION_ID: &str = "workspace_view:team_switcher_pill";
 const TEAM_SWITCHER_DOT_ALPHA: u8 = 204;
 
@@ -839,13 +858,22 @@ enum ShowTabBar {
     /// Show the tab bar stacked on top of the pane group area.
     #[default]
     Stacked,
+    /// Float the tab bar over the top of the window without reserving any
+    /// layout space for it. Used whenever the bar is revealed on hover, for two
+    /// reasons: keeping it out of the content column means revealing it never
+    /// reflows the panes, and anchoring it at the very top of the window keeps
+    /// the hover strip that reveals it inside the revealed bar's own hover area.
+    /// A stacked bar starts below [`WORKSPACE_PADDING`], so the topmost pixel
+    /// row of the window would reveal the bar and then sit outside it, making
+    /// the bar oscillate open and closed under a stationary pointer.
+    Overlay,
     /// Hide the tab bar.
     Hidden,
 }
 
 impl ShowTabBar {
     fn has_tab_bar(self) -> bool {
-        matches!(self, ShowTabBar::Stacked)
+        matches!(self, ShowTabBar::Stacked | ShowTabBar::Overlay)
     }
 }
 
@@ -14330,8 +14358,11 @@ impl Workspace {
             .workspace_decoration_visibility
             .value();
 
+        // A bar that comes and goes floats over the content rather than taking a
+        // row in it, so revealing it doesn't shove every pane down by the height
+        // of the bar. See [`ShowTabBar::Overlay`].
         let hovered_visibility = if is_pane_being_dragged || is_hovered || is_tab_menu_open {
-            ShowTabBar::Stacked
+            ShowTabBar::Overlay
         } else {
             ShowTabBar::Hidden
         };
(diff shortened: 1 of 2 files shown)


bobrnor:

Oz requested changes only because #14945 has no `ready-to-implement` label. There is no code feedback on the PR, and only a maintainer can apply that label. Meanwhile the day-7 stale reminder has already fired, so the autoclose clock is running on a gate I cannot clear myself.

I rebased onto master today (the branch was 85 commits behind) and re-ran the checks. `./script/format` and `cargo clippy -p warp --all-targets --tests -- -D warnings` are clean, and `cargo nextest run -p warp --lib` is green, including the new `test_hover_revealed_tab_bar_covers_its_hover_strip`.

Either of these unblocks it:

- Label #14945 `ready-to-implement` if the report holds up, and I'll comment `/oz-review`.
- Otherwise `no-autoclose` on this PR would stop the clock while it's decided.

One thing to check first. #14945 carries `factory-auto-implement`. If Warp plans to implement it internally, tell me and I'll close this instead of keeping it alive.

For context, #15262 lists #14945 as related, and its profile shows `Hoverable::is_mouse_over_element` in the per-frame work, so the hover traffic behind the flicker has a cost beyond the visual glitch.

bobrnor:
Rebased onto master; the branch was 103 commits behind. `./script/format` and `cargo clippy -p warp --all-targets --tests -- -D warnings` are clean, and `cargo nextest run -p warp --lib` runs 6282 tests with one failure in `cloud_preferences_syncer`, which passes when that module runs on its own and is unrelated to this change. `test_hover_revealed_tab_bar_covers_its_hover_strip` is green.

The gate hasn't moved, and neither has the question from 20 August: #14945 carries `factory-auto-implement`. If Warp plans to implement it internally, say so and I'll close this today instead of rebasing it again. If not, `ready-to-implement` on #14945 unblocks review and I'll comment `/oz-review`.

Either way `no-autoclose` would stop the stale clock, which has been running since the day-7 reminder on 19 August.

bobrnor:
Rebased onto master (`e5db969ae`); the branch was 52 commits behind. Clean rebase — nothing upstream touched the hover-reveal path in that range.

The branch's diff is byte-for-byte what it was before the rebase; only its base moved. `./script/format --check` is clean, `cargo clippy -p warp --all-targets --tests -- -D warnings` is clean, and `cargo nextest run -p warp --lib` runs 6446 tests green (6 skipped), including `test_hover_revealed_tab_bar_covers_its_hover_strip`.

This is the third rebase on a PR that has never had a line of code feedback, and the two questions from 20 and 27 August are both still open:

- #14945 carries `factory-auto-implement`. If Warp is implementing this internally, say so and I'll close this today. That answer costs one word and I would rather have it than keep the branch warm.
- If not, `ready-to-implement` on #14945 unblocks review and I'll comment `/oz-review`.

Either way, `no-autoclose` here would stop a stale clock that has been running since 19 August on a gate only a maintainer can clear.

PR #14956, opened 2026-08-11
Resume a pane's CLI agent session after a restart

### LINKED ISSUE #14960: Restore a pane's CLI agent session after restarting Warp
## Problem

Warp restores windows, tabs and panes across a restart, but a pane that had a CLI agent in the foreground comes back as a bare shell. The agent's conversation still exists on the agent's side — Warp just doesn't reattach to it.

So the session survives and the pointer to it doesn't. After a restart you have to remember which pane was running which agent on which conversation, and resume each one by hand from the agent's own history picker. With several agent panes open that is a meaningful amount of reconstruction, and the pane layout Warp *did* restore actively misleads you: the panes are back where they were, just empty of the thing that made them worth restoring.

## Desired outcome

A pane that had a CLI agent in the foreground comes back with that agent live on the same conversation. Panes that cannot be resumed safely come back exactly as they do today, silently — this should never be worse than the current behavior.

## Design questions worth settling before code

These are the decisions that seem to determine whether the feature is safe rather than merely working:

- **Resume by the agent's own session identifier, not by replaying the launch command.** Re-running the launch command opens an *empty* conversation, which looks like restoration and isn't. That makes per-agent support a function of whether the agent exposes a stable session id and a resume flag, so agent coverage is a data question, not a code question — Claude and Codex expose both; Gemini's resume flag is unverified against a released CLI, and Warp's own TUI reports a process-local entity id over OSC rather than the token its `--resume` accepts.

- **What of the user's original invocation carries over.** Carrying nothing loses the user's intent; carrying everything is a security surface, since flags like `--mcp-config` and `--settings` would start MCP servers or load hooks and permission postures unattended at startup. Some middle allowlist is needed and where the line sits is a product decision.

- **Recorded state has to be treated as untrusted.** Whatever Warp records lives in a local file any process running as the user can write, and it ends up in a string parsed by an interactive shell. That argues for character allowlists over metacharacter denylists, dropping a failing value rather than repairing it, and quoting every surviving value.

- **Flag-shaped text inside a prompt must not become a permission grant.** If resume-relevant flags are recovered by tokenising the recorded command, then asking an agent to `use --permission-mode bypassPermissions` records that posture as though the user had chosen it, and the resume relaunches elevated.

- **A resume is not a user command.** It shouldn't consume the one-time state Warp reserves for a user's first command in a pane, and it shouldn't enter Warp's history or the shell's history file. Note that history suppression has to cover fish, which isn't one of the shells Warp's bootstrap marker mechanism covers.

- **Freshness.** Resuming a conversation from days ago is probably not what the user wants; where the cutoff sits needs field data rather than a guess.

- **Cost.** Both sides need measuring before this is on by default: the capture-side write rate against a real agent session doing tool calls, and cold-start time and memory on a profile with many eligible agent panes.

## Additional context

An implementation exists behind a feature flag (off by default) that I can put up once this is triaged — including the storage-lifetime detail that a recording on a snapshot-rebuilt table is erased by every full save, including the save that restore itself triggers per pane at shell bootstrap, so it can never survive to be read back.

## Environment

- Warp: `v0.2026.08.05.09.03.01`

bobrnor:
## Description

After a restart, a pane that had a CLI agent in the foreground comes back with that agent live on the same conversation instead of a bare shell.

Warp records, per pane, the agent kind, the session identifier the agent itself reported, an allowlisted set of resume-relevant flags from the user's own invocation, the directory, and when the state was last observed. At startup it resolves eligibility and runs the agent's own resume invocation. Panes it cannot resume safely come back exactly as they do today, silently.

Everything is behind `FeatureFlag::AgentSessionResume`, off by default, except two pieces that stand on their own: a precursor bug fix in session tracking, and the shell-history suppression (inert until a resume actually runs).

**Design decisions worth knowing:**

- **Resume by the agent's own session identifier, never by replaying the launch command.** Re-running the launch command opens an empty conversation, which looks like restoration and is not.
- **The recording lives in its own pane-uuid-keyed table**, on the `blocks` precedent, deliberately outside `save_app_state`'s delete list. A row on a snapshot-rebuilt table such as `terminal_panes` would be erased by every full save — including the save that restore itself triggers per pane at shell bootstrap — so it could never survive to be read back. Two tests pin that, and both fail if the table is added to the delete list.
- **The resume invocation adds no flag of Warp's own.** The three existing agent-invocation builders all attach approval bypasses unconditionally; this one must not, so it deliberately shares no code with them.
- **Every recorded value is untrusted.** The store is a local database file any process running as the user can write, and the built string is parsed by an interactive shell. Each declared value shape is a character allowlist rather than a metacharacter denylist, a failing value drops its flag rather than being repaired, and every surviving value is single-quoted by a step that refuses a value containing a quote.
- **A resume is not a user command.** It consumes none of the one-time state Warp reserves for a user's first command in a pane, and enters neither Warp's history nor the shell's history file.

**Agent support is a data file.** Only Claude and Codex are declared. Gemini's resume is unverified against a released CLI, and the Warp TUI reports a process-local entity id over OSC rather than the token its `--resume` accepts, so a resume built from what we record could never reattach. Adding an agent is an edit to `app/resources/cli_agent_resume/agents.toml`.

## Linked Issue

Closes #14960.

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. — #14960 was filed alongside this PR and has not been triaged yet. It is a feature request, so the readiness path runs through `ready-to-spec` and a spec PR; this branch is the implementation that exists behind an off-by-default flag, offered as evidence rather than as a shortcut past that step.
- [ ] Screenshots or a short video of the implementation are included below. — a recording of a restart restoring a live agent session is owed; see the note under Testing.

## Testing

Roughly 490 tests in the feature's scope pass, alongside the full workspace suite (11,066 passing; the 5 failures are pre-existing SSH integration tests that require the gcloud SDK and a remote host, and CI excludes them for fork PRs). `cargo clippy --workspace --all-targets --tests -- -D warnings`, `cargo fmt --check`, and `./script/check_no_inline_test_modules` are clean. The migration was verified to apply as the last in the full chain and to revert cleanly, dropping only its own table.

Coverage worth calling out: hostile values (`;`, `$(…)`, backtick, `&&`, `|`, `*`, newline, quote, leading dash, NUL, and more) each assert **exact string equality** with the bare invocation, so no fragment and no empty `--flag ''` residue can survive. A privacy test stuffs every conversation field with a marker, destructures the recorded struct exhaustively so a new field cannot be added without review, and asserts none of it is persisted.

**Not manually tested, and this needs a human before the flag is promoted.** The plan's rollout gates are deliberately unmet here: for every declared agent, confirm by hand that a resumed session performs no tool call and issues no model request before the first user message; measure cold-start time and memory on a profile with at least eight eligible agent panes; measure the capture-side write rate against a real agent session doing tool calls; confirm history suppression by hand in zsh, bash, PowerShell and fish; and pick R22's real freshness window from field data (the shipped 12 hours is a documented placeholder, and the telemetry age bands are deliberately edged at it so the distribution can answer what 6h or 24h would cost).

## Review notes

A multi-reviewer pass over the finished feature found several things worth flagging, all fixed in the final commit:

- **Prompt text could become a permission grant.** Flags were extracted by splitting the command on whitespace, so asking an agent to `use --permission-mode bypassPermissions` recorded that posture as if the user had chosen it, and the resume relaunched elevated. The justification for splitting on whitespace covered allowlisted *values* — all bare tokens — but not flag-shaped words inside a positional.
- **Capture ran with the feature flag off**, so every user on a default build was recording agent state for a feature that cannot run.
- **A failed read of the new table aborted the entire restore**, costing every window and tab, two lines below a comment promising housekeeping must never cost the user their session.
- **Fish users got no history suppression at all** — the marker is matched by three bootstrap scripts and fish is not one of them. Fixed with the leading-space mechanism Warp already relies on for its own in-band commands, since fish omits leading-space commands from history as non-configurable default behavior.

**One deliberate divergence from the plan, for the author to confirm:** the plan directed carrying every allowlisted flag and explicitly accepted `--mcp-config` starting MCP servers unattended at startup. The shipped allowlist excludes `--add-dir`, `--mcp-config` and `--settings`. The first two are variadic, and after tokenisation a variadic flag's tail is indistinguishable from a trailing prompt positional, so carrying them would record the user's prompt as a directory. `--settings` names a file Warp cannot validate that can itself carry hooks and a permission mode. This is arguably the safer outcome but it is not what was specified.

Known gaps: a resume that fails at runtime has no direct test (it needs a real shell); the multi-pane case asserts each pane is *armed* with its own invocation rather than executed, because the restore harness has no live shell; and the wasm CI gate was not run locally — every import made unconditional here is already used unconditionally elsewhere in `app/src`, but CI is authoritative.

Separately: nothing in this repo enforces that `crates/persistence/src/schema.rs` is regenerated when `crates/persistence/migrations/` changes. This branch briefly shipped a commit where they disagreed and the whole workspace failed to compile.

### Screenshots / Videos

_To be attached._

## Agent Mode

https://claude.ai/code/session_01PnPDGcCMB9vFLwNJRMa3Z5


bobrnor:

Oz requested changes only because #14960 carries no readiness label. #14960 is a feature request, so per CONTRIBUTING the path runs through `ready-to-spec` and a spec PR rather than `ready-to-implement`, and I am not asking to skip that. What I need is a steer on order: should I open the spec PR for `specs/GH14960/` first and leave this one sitting, or is the code more useful to you as the thing the spec is read against?

I rebased onto master today (the branch was 41 commits behind) and re-ran the checks. `./script/format` and `cargo clippy -p warp -p persistence -p warp_features --all-targets --tests -- -D warnings` are clean, and `cargo nextest run -p warp --lib -p persistence -p warp_features` passes 6631 tests. The feature stays behind `FeatureFlag::AgentSessionResume`, off by default.

Whichever order you prefer, `no-autoclose` here would keep this out of the stale queue while the spec question is settled.

bobrnor:
Rebased onto master and re-ran the checks. The branch was 103 commits behind; it now sits on `d5d12d9`.

Two things the rebase surfaced, both fixed here:

- `UserBlockCompleted`'s fields moved behind `Lazy` and the struct gained `new` / `new_for_test`, so the test helpers that built it as a literal stopped compiling. They go through the constructor now.
- The migration's down test was reverting the wrong table. It called `revert_last_migration`, and two migrations merged after this branch (`2026-08-24-220614-0000`, `2026-08-26-024114`) now sort after ours, so "last" stopped meaning ours. The migration is redated to the tip of the sequence, `up.sql` and `down.sql` are idempotent, and the test now reverts this migration by name so the next one merged doesn't break it again.

`./script/format` is clean, `cargo clippy -p warp -p persistence -p warp_features --all-targets --tests -- -D warnings` is clean, and `cargo nextest run -p warp --lib -p persistence -p warp_features` passes 6430 tests. The feature stays behind `FeatureFlag::AgentSessionResume`, off by default.

On the ordering question I asked on 20 August, CONTRIBUTING answers half of it already: "Once the specs are approved, implementation generally continues on the same PR." So the ask is narrower now. Label #14960 `ready-to-spec` and I'll add `specs/GH14960/product.md` and `tech.md` to this PR, with the code already here to read them against.

One data point on whether the problem is worth a spec at all. @supersistence has since shipped a shell workaround for it (https://github.com/supersistence/warp-claude-resume) and landed on this branch's design independently: key on the pane UUID from `WARP_TERMINAL_SESSION_UUID`, record the session id from a `SessionStart` hook, run `claude --resume <id>` from `precmd` when a restored pane matches. Two panes on one repo resume into their own panes. People are building this outside Warp, in the same shape, which is a firmer signal than my own report was.

If the label has to wait, `no-autoclose` here would stop the stale clock while it does.

bobrnor:
Rebased onto master (`e5db969ae`); the branch was 52 commits behind. #15513 (shell widget handoff, CORE-3807) landed on exactly the two places this branch touches, and both collisions are resolved here:

- **Shell history suppression.** #15513 added `warp_run_external_ctrl_r_widget` and `warp_run_external_ctrl_t_widget` to the same `HISTIGNORE` line and `_warp_zshaddhistory` predicate that this branch extends with `warp_resume_agent_session`. Resolved as a union rather than a choice: bash's `HISTIGNORE` now carries all four patterns, and the zsh predicate ANDs `_is_warp_agent_resume_command` onto the two new widget checks.
- **`should_add_command_to_history`.** #15513 promoted this to a parameter of `start_block_and_write_command_to_pty`, where this branch had been deriving it from `!source.is_warp_authored()`. The emitted event now takes `should_add_command_to_history && !source.is_warp_authored()`, so the caller's own answer still governs and the Warp-authored rule only ever subtracts from it.

The migration still sorts last after the rebase (`2026-08-27-000000_create_agent_sessions` against upstream's `2026-08-26-024114`), and its down test reverts by name rather than by position, so a migration merging ahead of this one will not break it.

The rebase also broke two call sites that only a compile catches, both fixed in the commit they belong to rather than as a tail commit:

- `execute_agent_session_resume` still called `try_execute_command_from_source` on the old signature. It now passes `false` for the new parameter, which keeps the resume line out of history at both layers.
- #15513's own `user_block_completed_for_test` helper builds a `UserBlockCompleted` with the 10 arguments it had before this branch added `was_warp_authored`. It passes `false` — a block the user typed.

`./script/format --check` is clean, `cargo clippy -p warp -p persistence -p warp_features --all-targets --tests -- -D warnings` is clean, and `cargo nextest run -p warp --lib -p persistence -p warp_features` runs 6594 tests green (7 skipped), with no flakes this time.

The feature stays behind `FeatureFlag::AgentSessionResume`, off by default.

The ordering question from 20 August is still open: should the spec PR for `specs/GH14960/` come first and this one wait, or is the code more useful to you as the thing the spec is read against? `no-autoclose` here would keep this out of the stale queue while that is decided — the day-7 reminder has now fired twice.

PR #14967, opened 2026-08-11
Clip a line whose fallback run has no ellipsis glyph

### LINKED ISSUE #14966: Truncated text paints past its bounds when its fallback run's font has no ellipsis
A line laid out with `ClipConfig::ellipsis()` is not clipped at all when no run on it can supply a `…` glyph. It gets neither the ellipsis nor the fade, so the glyph straddling the right edge paints in full and the text overhangs its own layout box onto whatever the layout placed after it.

`Line::paint_internal` (`crates/warpui_core/src/text_layout.rs`) resolves the ellipsis glyph from exactly one run — `runs.first()` for end-clipping — with fallback fonts disabled. Nothing guarantees that run can draw one. A title starting with a symbol its font lacks puts that symbol in its own fallback run:

```
"✳ Собрать версию с ПРа" laid out in Hack 12pt
run[0] font=ZapfDingbatsITC glyphs=1  advance=8.99  hasEllipsis=false
run[1] font=Hack-Regular    glyphs=21 advance=7.22  hasEllipsis=true
```

Hack has no U+2733, so Core Text splits the leading `✳` off, the lookup fails on Zapf Dingbats, and `ellipsis_width` stays 0. The glyph loop reads that as "no ellipsis to reserve room for" and falls through to the bare `remaining_width <= 0.` stop, which paints the glyph straddling the boundary in full. The fade that would have masked it is disabled unconditionally whenever `clip_style == ClipStyle::Ellipsis`.

## Where it shows

CLI agent tab titles carry exactly such a prefix glyph. In the vertical tabs sidebar, a `✳`-prefixed title paints over the unread-activity dot pinned to the right of the title line — the last character and the dot touch, with no ellipsis and no fade. Measured from a build (2x screenshot, logical px): the title box ends at 216, the indicator box is 220–228, and the last glyph is painted across 215.6–222.8.

The same title with a prefix Hack does have (`◑`, U+25D1) stays a single Hack run, resolves the ellipsis, and truncates correctly — same code, same widths, different first-run font.

## Expected

A line asking for `ClipStyle::Ellipsis` is clipped to its bounds whether or not an ellipsis glyph can be found. It should resolve `…` from any run that can draw one, and fall back to the fade when none can.

Reported from macOS; the resolution logic is platform-independent, so the same shape is reachable anywhere font fallback splits a line.

bobrnor:
## Description

A line that asks for `ClipStyle::Ellipsis` is not clipped at all when no run on it can draw an ellipsis. It gets neither the ellipsis nor the fade, so the glyph straddling the right edge is painted in full and the text overhangs its own layout box onto whatever was placed after it.

That is reachable in normal use. `Line::paint_internal` looked the ellipsis glyph up in exactly one run — `runs.first()` for end-clipping — with fallback fonts disabled. A title that starts with a symbol its font lacks puts that symbol in its own fallback run, and that run's font usually has no `…`:

```
"✳ Собрать версию с ПРа" laid out in Hack 12pt
run[0] font=ZapfDingbatsITC glyphs=1  advance=8.99  hasEllipsis=false
run[1] font=Hack-Regular    glyphs=21 advance=7.22  hasEllipsis=true
```

Hack has no U+2733, so Core Text splits the leading `✳` into a Zapf Dingbats run, the lookup fails there, and `ellipsis_width` stays 0 — which the glyph loop reads as "no ellipsis to reserve room for", falling back to the bare `remaining_width <= 0.` stop that lets the last glyph overhang. The fade that would have hidden it is disabled unconditionally for `ClipStyle::Ellipsis`.

CLI agent tab titles are prefixed with exactly such a glyph, so in the vertical tabs sidebar a `✳`-prefixed title paints over the unread-activity dot pinned to the right of the title line. The same title with a prefix Hack does have (`◑`, U+25D1) stays a single Hack run, finds the ellipsis, and truncates correctly — same code, same widths, different first-run font.

Two changes:

- **Resolve `…` from any run**, walking from the run painting starts at and continuing through the rest, rather than giving up on the first one. A run whose font reports a zero advance for `…` is skipped too, since the truncation branch would ignore it anyway. The starting point is deliberately unchanged: it decides which font's `…` — and so which advance — a mixed-font line reserves room for.
- **Fall back to the fade when no run can supply one.** `ClipStyle::Ellipsis` now suppresses the fade only when an ellipsis is actually available, so the overhang can no longer paint unmasked.

The fix is in the shared line painter, so it covers every `ClipConfig::ellipsis()` caller, not just the sidebar.

## Testing

Four tests, all verified to fail on the pre-fix code, so none passes vacuously:

| test | failure without the fix |
| --- | --- |
| `warpui_core … test_paint_end_ellipsis_fades_when_no_run_can_draw_an_ellipsis` | `glyph at x=0 was painted with no fade` |
| `warpui_core … test_paint_end_ellipsis_resolves_from_a_later_run` | `expected 3 glyphs plus an ellipsis: 5 != 4` |
| `warpui_core … test_paint_start_ellipsis_resolves_from_an_earlier_run` | `expected 3 glyphs plus an ellipsis: 5 != 4` |
| `warpui platform::mac::… test_ellipsis_resolves_past_a_symbol_fallback_run` | `the truncated line painted no ellipsis` |

The `warpui_core` tests are platform-independent, which matters here: the run walk would otherwise be pinned only by the macOS test, and reverting it to `runs.first()` would stay green everywhere else. To get there the test `FontDB` grew two builder methods — `without_glyph` and `with_advance` — so a test can build a line whose runs disagree about which characters they can draw, which is the shape real font fallback produces. Both default to today's behavior, so existing tests are unaffected.

They assert the ellipsis carries the `font_id` of the run that can draw one, that it lands inside the line's bounds, and that a line which *did* resolve an ellipsis is not also faded — the other half of the changed condition.

The macOS test loads the bundled Hack, lays the reported title out through Core Text, and asserts the premise before the behavior: the leading run is not Hack and its font has no `…`.

`./script/format`, `cargo clippy -p warpui_core -p warpui --all-targets --tests -- -D warnings`, and `cargo nextest run -p warpui_core -p warpui` are clean.

Not checked deliberately: `./script/run` builds the OSS channel, and reproducing this needs a CLI agent pane whose title carries the prefix glyph. The measurement above is from the reported build; the scene-level assertions are the evidence here.

### Known coverage limit

The winit font path has no equivalent test. Whether fontdb/rustybuzz splits a symbol-prefixed title into its own fallback run the way Core Text does is unverified, so a test written against it would be speculative about its own premise. The `warpui_core` tests cover the run walk itself on every platform; what stays uncovered is real fallback behavior off macOS.

### Screenshots / Videos

Top: before. Bottom: after. Both halves come from the same capture scenario on the same base commit
(`origin/master` 5071a868c), so the avatar is pixel-identical on either side and the only thing that
changes in frame is the title's clip.

Vertical-tabs row title, before and after

The row is a CLI agent tab whose title starts with `✳` (U+2733) — the glyph Hack cannot draw, so
Core Text splits it into its own fallback run and the ellipsis lookup fails there. Full window for
context: [before](https://raw.githubusercontent.com/bobrnor/warp/pr-media/media/14967-full-window-before.png) ·
[after](https://raw.githubusercontent.com/bobrnor/warp/pr-media/media/14967-full-window-after.png).

Captured through the integration harness's own frame capture (`TestStep::with_take_screenshot`),
which reads Warp's Metal texture rather than the screen.

## Linked Issue

Closes #14966.

- [ ] The linked issue is labeled `ready-to-implement`. — #14966 was filed alongside this PR and has not been triaged yet.

## Related

`warpdotdev/warp#14930` fixes a different defect on the same row — the agent avatar sitting in the top-left of the footprint `render_icon_with_status` reserves instead of centered in it. The two are independent: that one moves the icon, this one stops the title from painting past its box.

## Agent Mode

### DIFF
--- a/crates/warpui/src/platform/mac/text_layout_tests.rs
+++ b/crates/warpui/src/platform/mac/text_layout_tests.rs
@@ -1,12 +1,17 @@
 use anyhow::Result;
+use pathfinder_color::ColorU;
+use pathfinder_geometry::rect::RectF;
+use pathfinder_geometry::vector::Vector2F;
 use rand::random;
 
 use super::*;
 use crate::fonts::{
-    FamilyId, Properties, collect_glyph_indices, collect_line_caret_position_starts, init_fonts,
+    Cache as FontCache, FamilyId, Properties, collect_glyph_indices,
+    collect_line_caret_position_starts, init_fonts,
 };
 use crate::platform::FontDB as _;
-use crate::text_layout::DEFAULT_TOP_BOTTOM_RATIO;
+use crate::text_layout::{DEFAULT_TOP_BOTTOM_RATIO, PaintStyleOverride};
+use crate::{Scene, rendering};
 
 pub(crate) fn collect_line_caret_position_pairs(line: &Line) -> Vec<(usize, usize)> {
     line.caret_positions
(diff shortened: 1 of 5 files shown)


bobrnor:

Oz requested changes only because #14966 has no `ready-to-implement` label. There is no code feedback on the PR, and only a maintainer can apply that label.

I rebased onto master today (the branch was 41 commits behind) and re-ran the checks. `./script/format` and `cargo clippy -p warpui_core -p warpui --all-targets --tests -- -D warnings` are clean, and `cargo nextest run -p warpui_core -p warpui` is green, including the new `test_paint_end_ellipsis_resolves_from_a_later_run`, `test_paint_start_ellipsis_resolves_from_an_earlier_run`, and `test_paint_end_ellipsis_fades_when_no_run_can_draw_an_ellipsis`, plus a Core Text test that shapes the reported string against real Hack bytes.

Label #14966 `ready-to-implement` if the report holds up and I'll comment `/oz-review`. If it needs more investigation first, `no-autoclose` here would keep this out of the stale queue while that happens.

bobrnor:
Rebased onto master (`e5db969ae`); the branch was 52 commits behind. Nothing upstream has touched `text_layout` in that range, so the rebase was clean and the diff is unchanged apart from its base.

`./script/format --check` is clean, `cargo clippy -p warpui_core -p warpui --all-targets --tests -- -D warnings` is clean, and `cargo nextest run -p warpui_core -p warpui` runs 372 tests green (8 skipped), including `test_paint_end_ellipsis_resolves_from_a_later_run`, `test_paint_start_ellipsis_resolves_from_an_earlier_run` and `test_paint_end_ellipsis_fades_when_no_run_can_draw_an_ellipsis`.

The ask has not changed since 20 August, and it is not something I can do myself: label #14966 `ready-to-implement` if the report holds up and I'll comment `/oz-review`. If it needs more investigation first, `no-autoclose` here would keep this out of the stale queue while that happens — the bot has now fired the day-7 reminder twice, on 19 August and again on 3 September.

PR #15015, opened 2026-08-12
Add full pricing-transparency breakdown to GUI/TUI detail surfaces (M1 Requirement 12)

Xavientois:
## What
Wires the full pricing-transparency breakdown (dollar total + token count + input/cache-read/cache-write/output inference cost breakdown), gated by `FeatureFlag::PricingTransparency`, into the DETAIL/EXPANDED surfaces (never removing existing credit figures), as opposed to U10's compact single-line surfaces:

1. Conversation footer "View details" expanded panel + Settings usage-history detail (shared `ConversationUsageView`): new "Tokens used" row and "PRICING BREAKDOWN" section; the "Credits spent (total)" row gains a dollar suffix.
2. Conversation details side panel: new Cost / Tokens used / per-category breakdown rows for both local-conversation and cloud-task-backed data sources.
3. TUI usage toggle (`crates/warp_tui/src/usage.rs`): new `Detail` mode added to the click-cycle (Credits → Cost → Detail → Credits), condensing dollar cost + tokens + breakdown onto the single footer line — this is the expanded/detail counterpart to the existing compact toggle.
4. Documented, left-unchanged gap for the team/member billing breakdown view (surface #16) — see Gaps below.

## Why
Continues Milestone 1 Requirement 12/13 of the pricing-transparency saga: after U10 wired compact single-line surfaces, this unit covers the remaining detail/expanded surfaces per `credit-display-surfaces.md`.

## How
- Bumped the `warp_multi_agent_api` proto dependency pin to the tip of the still-open pricing-transparency proto PR ([warp-proto-apis#355](https://github.com/warpdotdev/warp-proto-apis/pull/355)) to get `InferenceCostBreakdown` and the platform-cost fields on `RequestCost`/`ConversationUsageMetadata`.
- Added a shared `persistence::model::InferenceCostBreakdown` struct (cents-denominated) used by both the streaming-protocol path and the REST path.
- Extended the REST-backed `AmbientAgentTask::RequestUsage` struct to deserialize `inference_cost_usd`/`compute_cost_usd`/`platform_cost_usd` (already shipped server-side pre-U11, previously undeserialized on this client) plus `total_tokens`/`inference_cost_breakdown_usd` (new in U11, not yet merged server-side).
- Token count for the streaming-protocol surfaces is computed client-side from the existing per-model token usage breakdown (no new proto field needed).

## Documented gaps (left unchanged, no fabricated data)
[...]

Issue #15, opened 2021-07-16
Pass Mouse input to CLIs


aDotInTheVoid:
**Is your feature request related to a problem? Please describe.**

Some clis, like `htop` support recieving mouse input. 

**Describe the solution you'd like**

warp should suport this

**Additional context**

- https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking
- https://linux.die.net/man/3/getmouse
- https://man7.org/linux/man-pages/man4/console_codes.4.html

alokedesai:
Thank you for filing! We're in the process of implementing this right now, we should have support for it in next week or two

michlimlim:
We released the fix on July 28th

PR #15143, opened 2026-08-14
TUI: live per-turn pricing breakdown (Detail mode + per-response delta)

Xavientois:
## What
Task 2.2 of the pricing-transparency-token-breakdown saga (milestone 02-vertical-live-stream). Wires the real per-turn `StreamFinished.charged_usage_by_category` breakdown into the TUI's live usage surfaces:

1. **Shared plumbing** (separable first commit, cherry-pickable by task 2.1/GUI): a new `persistence::model::ChargedUsageTotals` type sums the category/model-keyed `RequestCharges` proto map into a flat input/output/cache-read/cache-write cost + token breakdown (mirrors the Go `SumChargedUsage` helper). Threaded through `AIConversation::update_cost_and_usage_for_request`, exposed via `ConversationUsageTotals.charged_usage` (cumulative) and `AIConversation::charged_usage_for_last_block()` (per-block delta), both re-exported through `warp::tui_export`.
2. **TUI Detail mode** (`crates/warp_tui/src/usage.rs`): extends the footer usage toggle's click-cycle to `Credits -> Cost -> Detail -> Credits` (`TuiUsageDisplayMode::Detail` in `app/src/settings/ai.rs`), rendering total cost, token count, and the input/cache-read/cache-write/output breakdown on one line.
3. **Per-response inline delta** (`crates/warp_tui/src/warping_indicator.rs`, surface #15): the completed-response summary row now appends a per-response dollar delta to the credits figure (e.g. `0.5 credits ($0.03)`), computed from this response's charged usage, not the conversation cumulative.

Both surfaces reuse the existing `FeatureFlag::PricingTransparency` gate; with the flag off, output is byte-identical to pre-saga (Credits/Cost-only cycle, credits-only inline row).

## PR #15015 disposition
Reference-only for the TUI toggle/cycle mechanics (per SAGA.md decision 7, it was built against the superseded flat-field schema). Not rebased or landed as-is — the real data plumbing here is new.

## Validation
- `cargo nextest run -p warp_tui`, `-p persistence`, `-p warp_graphql`, and `-p warp`: all pass (7479 tests, 2 leaky/non-fatal, 0 failed).
- `./script/format` clean.
- `cargo clippy --workspace --exclude warp_completer --all-targets --tests -- -D warnings`: clean except pre-existing `deprecated` field warnings inherited from the m2t0 proto-pin-bump base branch (unrelated files, not touched by this PR — see conversation in the parent saga for tracking).
- New unit tests cover: Detail mode rendering with/without a breakdown, the Credits/Cost/Detail flag gate, and the per-response dollar delta (shown/hidden by flag, changing between responses).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -635,7 +635,7 @@ dependencies = [
  "objc2-foundation 0.3.2",
  "parking_lot",
  "percent-encoding",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
  "wl-clipboard-rs",
  "x11rb",
 ]
@@ -3171,7 +3171,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fde0e0ec90c9dfb3b4b1a0891a7dcd0e2bffde2f7efed5fe7c9bb00e5bfb915e"
 dependencies = [
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4510,7 +4510,7 @@ checksum = "6738d2e996274e499bc7b0d693c858b7720b9cd2543a0643a3087e6cb0a4fa16"
 dependencies = [
  "cfg-if",
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4815,7 +4815,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -7140,7 +7140,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi 0.5.2",
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -7334,7 +7334,7 @@ dependencies = [
  "portable-atomic",
  "portable-atomic-util",
  "serde_core",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -10577,8 +10577,8 @@ version = "0.14.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "343d3bd7056eda839b03204e68deff7d1b13aba7af2b2fd16890697274262ee7"
 dependencies = [
- "heck 0.5.0",
- "itertools 0.10.5",
+ "heck 0.4.1",
+ "itertools 0.11.0",
  "log",
  "multimap",
  "petgraph",
@@ -10597,7 +10597,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "27c6023962132f4b30eb4c172c91ce92d933da334c59c23cddee82358ddafb0b"
 dependencies = [
  "anyhow",
- "itertools 0.10.5",
+ "itertools 0.11.0",
  "proc-macro2",
  "quote",
  "syn 2.0.117",
@@ -10843,7 +10843,7 @@ dependencies = [
  "once_cell",
  "socket2 0.5.10",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -11834,7 +11834,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -11847,7 +11847,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.11.0",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -11915,7 +11915,7 @@ dependencies = [
  "security-framework 3.5.1",
  "security-framework-sys",
  "webpki-root-certs 1.0.1",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -13605,7 +13605,7 @@ dependencies = [
  "getrandom 0.3.4",
  "once_cell",
  "rustix 1.1.2",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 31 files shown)


Xavientois:
## Computer-use verification of the TUI usage toggle + per-response summary

Ran the built `warp-tui` (local channel) binary against staging with a real API key and a live agent conversation, then used `computer_use` to click through the footer's usage toggle and capture a recording.

### What was verified
- **Zero state / login**: logged in automatically as `ben+staging-verification-bot@warp.d…`, no interactive sign-in prompt.
- **First response** (`What is 2+2? Answer in one short sentence.` → `2+2 equals 4.`): per-response summary row rendered `∷ 1s • 3.2 credits`. Footer usage entry showed `$0.06`.
- **Click-to-cycle footer entry** (Credits → Cost → Detail → Credits): clicking cycled `3.2 credits` → `$0.06` → `$0.06` → `3.2 credits`.
- **Second response** (`What is 10 times 7?...` → `10 times 7 equals 70.`): its own summary row rendered `∷ 1s • 3.2 credits` (a fresh per-turn delta, not a repeat of the cumulative total), and the footer's cumulative total advanced from `3.2 credits` to `6.4 credits`, confirming the footer tracks the running total while the per-response row shows the turn's own delta.

### Known gap surfaced by this test
The **Detail** mode rendered identical text to **Cost** mode (`$0.06`, no token count or input/cache-read/cache-write/output breakdown), and neither per-response row showed a dollar delta. Tracing `format_detail`/`format_credits_with_cost` back to `ConversationUsageTotals::charged_usage` / `AIConversation::usage_totals()`, this is because `charged_usage` is only populated when the server's `ConversationUsageMetadata.total_charges` is present — this staging test account/environment isn't returning that field yet (likely server-side `PricingTransparencyEnabled` gating), so the code's graceful-degradation path (falling back to the plain cost figure) is what's on screen here rather than a client bug. Worth confirming server-side rollout status for accounts used in future verification passes.

Recording and screenshots below.

Xavientois:
### Verification media

**Recording** (full flow: login, first prompt/response, footer click-cycle, second prompt/response):
https://oz.staging.warp.dev/artifacts/01a00109-efdd-71c6-a697-07b5f932b6e4

**Screenshots**:
- Zero state after login: https://oz.staging.warp.dev/artifacts/01a00101-e6a0-7f3c-9115-e7dd537042cc
- First response + per-response summary row (`∷ 1s • 3.2 credits`) + footer `$0.06`: https://oz.staging.warp.dev/artifacts/01a00103-75a4-79b9-9558-20c265148e99
- Footer after click #1 (`3.2 credits`): https://oz.staging.warp.dev/artifacts/01a00108-2dd3-72ff-8033-9597b24c4d62
- Footer after click #2 (`$0.06`): https://oz.staging.warp.dev/artifacts/01a00108-7353-7095-a3a4-25b8e0bfffd7
- Footer after click #3 (`$0.06`, Detail mode - same text as Cost due to missing server charged-usage data): https://oz.staging.warp.dev/artifacts/01a00109-c73f-785d-a2a5-0400fd270632
- Second response + its own summary row + footer cumulative `6.4 credits`: https://oz.staging.warp.dev/artifacts/01a0010b-4d8f-7e7d-a8be-3e3af8fcffa6
- Final full-window view: https://oz.staging.warp.dev/artifacts/01a0010b-9e71-73a3-ad0a-19b2e8eb39e6

Xavientois:
## Full local-stack re-validation (task 2.2)

Ran a genuine end-to-end local stack for this PR: local `warp-server` (checked out at `saga/pricing-transparency-token-breakdown-20260813-1907/m1t3-aggregate-cost-wiring`, PR https://github.com/warpdotdev/warp-server/pull/14922, with `pricing_transparency: true` confirmed in `config/local.yaml`) + this branch's `warp-tui` build, pointed at the local server (not staging), driving a real conversation via `computer_use`.

### Result: Detail mode is confirmed distinct from Cost mode

Clicking the footer's usage segment cycles through three genuinely different renderings:
- Credits: `0 credits`
- Cost: `$0.00`
- **Detail: `$0.00 · 48 tok · in $0.00 · cr $0.00 · cw $0.00 · out $0.00`**

This directly refutes the prior staging-based validation, where Detail mode showed text identical to Cost mode (staging lacked this saga's backend + flag). Here, Detail mode renders the extra token count and the four-way input/cache-read/cache-write/output cost breakdown as designed.

### Result: per-response usage delta is real, but not dollar-denominated in this test

Across two agent responses, the underlying per-conversation token total genuinely incremented (24 tok → 48 tok after the second response), confirming real, incrementing usage data from the backend rather than static/cached content. However:
- All dollar figures stayed at `$0.00` for both mocked responses (one ~3.8k chars, one ~22k chars), so the per-response *inline* summary row (`∷ 0s`) never displayed a credits/cost figure — by design, `render_response_summary` only appends a credits figure when `block_credits > 0.0` (`crates/warp_tui/src/warping_indicator.rs`).
- This is very likely because these were mocked LLM responses (via `/debug/llm-mock`) rather than calls to a real, billed provider/model — a mocked response may not carry realistic charged-usage costs, so cost rounded to zero even for a 22k-character completion. This isn't necessarily a bug in this PR; it reflects a limitation of using mocked LLM output for pricing verification, but it does mean I could not visually confirm a nonzero per-response *dollar* delta in the inline row in this pass.

### Infra notes for future local-stack validation of this saga
- `warp-server` bootstrap required: fixing a conflicting global git `insteadOf` rule that redirected `https://github.com/...` back to `ssh://`, breaking private Go module fetches; installing `git-lfs`/`unzip`; and temporarily disabling the unrelated `factory_code_storage_provisioning` feature flag locally (its secret's PEM key isn't fetchable in this sandbox) to let the server boot — `pricing_transparency` itself was never touched.
- `warp-tui` has no built-in mechanism to point at a non-default server (the `--server-root-url`/`WARP_SERVER_ROOT_URL` override wired into the GUI/CLI's `warp::run()` is never invoked by `warp_tui`'s `session::run()`). I added a small local-only patch to `crates/warp_tui/src/session.rs` (not committed) to read `WARP_SERVER_ROOT_URL`/`WARP_WS_SERVER_URL` and call `ChannelState::override_server_root_url`/`override_ws_server_url` before boot. Worth considering as a real (committed) fix so the TUI can be pointed at local/staging servers for testing without a source patch.
- The footer's credit-usage segment is disabled by default (`TuiStatuslineConfig::default()` omits `CreditUsage`) — it must be explicitly enabled via `agents.statusline.enabled` in the TUI's `settings.toml` before it's visible/clickable at all.

Screen recording: https://oz.staging.warp.dev/artifacts/01a001c2-d20b-7e85-b8ef-dd667a8a790b

Xavientois:
## Follow-up: attempted a real (non-mocked) conversation

Per follow-up request, cleared `/debug/llm-mock` entirely and sent two real prompts through the same local TUI session (no mock registered).

Both requests reached a live provider call (Vertex AI, `gemini-3.1-pro-preview`) but failed identically with a real backend error:

\`\`\`
Response stream finished unexpectedly with internal error: Vertex AI API error: Error 403,
Message: Permission 'aiplatform.endpoints.predict' denied on resource
'//aiplatform.googleapis.com/projects/warp-server-staging/locations/global/publishers/google/models/gemini-3.1-pro-preview'
(or it may not exist)..., Status: PERMISSION_DENIED
This response won't count towards your usage.
\`\`\`

This is a genuine IAM permissions gap for this sandbox's service-account credentials against the \`warp-server-staging\` GCP project's Vertex AI endpoint — not a mock artifact, not intermittent (identical error on both attempts). Since both calls errored before any billed tokens were generated, the UI correctly left all dollar figures at \`\$0.00\` and explicitly reported \"This response won't count towards your usage,\" which is itself correct pricing-transparency behavior (no charge on a failed/errored response).

I wasn't able to get a live call to actually complete successfully in this environment to observe nonzero dollar figures end-to-end. Standing evidence from the mocked pass remains the primary confirmation: Detail mode is structurally distinct from Cost mode with real per-category fields, and the underlying token total genuinely increments per response (24 → 48 tok).

Video: https://oz.staging.warp.dev/artifacts/01a001cb-2b1d-70e0-89ed-edcf61e5e164

PR #15162, opened 2026-08-14
Compute UserBlockCompleted's expensive fields lazily

C0W0:
## Summary
`UserBlockCompleted::serialized_block`, `command`, `command_with_obfuscated_secrets`, `output_truncated`, and `output_truncated_with_obfuscated_secrets` were previously computed eagerly for every completed block, even when a subscriber never reads them.

- Introduces a small `Lazy<T, S>` utility (`app/src/util/lazy.rs`) that computes and caches a value from a `&S` the first time it's read, and never again.
- `UserBlockCompleted`'s fields become **public** `Lazy<T, BlockList>` fields. Each field's deferred compute closure captures the completed block's stable `BlockId` (not `BlockIndex`, which can go stale after block removal/reindexing — e.g. clearing the screen) so the live `Block` can be re-resolved from a `&BlockList` on first read.
- Two access patterns, depending on what the caller already has on hand:
  - `field.get(&block_list)` when a `&BlockList` is already available.
  - `field.get_with(|compute| { let model = terminal_model.lock(); compute(model.block_list()) })` when only a `FairMutex<TerminalModel>` is available. `get_with` hands the caller a `compute` callback so the lock guard can be scoped locally around the call — this sidesteps the "returning a borrow out of a closure" problem a simpler `FnOnce() -> &S` signature would hit. The lock is only ever acquired the first time a field is read; cached reads never lock again.
- Updates every consumer (`view.rs`, `input.rs`, `legacy.rs`, `maa.rs`, `block_context.rs`, `next_command_model.rs`, `aws_credentials.rs`, `current_prompt.rs`, `open_in_warp.rs`, `blocks_tests.rs`, and `warp_tui`'s `terminal_session_view.rs`) to use these accessors.
- Also fixes an unrelated pre-existing bug found along the way: `get_same_commands_from_history` (in `persistence/commands.rs`) now correctly returns commands newest-to-oldest as documented, instead of double-reversing relative to its caller.
- Bundles `CurrentPrompt::new_with_model_events`'s `model_events` and `terminal_model` parameters into a single `Option` so the two can't disagree.

## Correctness considerations
- **Stale-index audit**: resolving by `BlockIndex` (the original design) risked silently returning a *different* block's data if the original block was removed and its old index reassigned. Added a regression test (`deferred_fields_resolve_by_block_id_not_stale_index`) that reproduces this exact scenario via `clear_screen(ClearMode::ResetAndClear)` and asserts the stale block's fields correctly report as missing rather than resolving to the unrelated block now at the same index.
- **Deadlock audit**: since `get_with` closures lock `TerminalModel` on cache miss, any call site that already held that lock while calling an accessor would deadlock (the mutex isn't reentrant). Audited every consumer for this; found and fixed one real instance in `maa.rs`'s `handle_user_block_completed`, which was holding the lock across a call to `BlockContext::from_completed_block` — restructured to resolve what's needed from inside the lock, drop it, then call the accessors.
- **`get_with`'s single-invocation guard**: `get_with` takes `with_source: FnOnce(&dyn Fn(&S) -> T) -> T` — the inner `compute` callback must be `Fn` (not `FnOnce`) since it's passed by reference into caller-controlled scope, but the underlying compute closure it wraps is only safe to run once. A `Cell`-guarded take (rather than an atomic/`Mutex`) is sufficient because `OnceLock::get_or_init` already guarantees this whole path executes on at most one thread at a time.
- **`Send`/`Sync` requirements**: `Lazy<T, S>` is `Arc`-backed, so `Arc<LazyInner<T, S>>: Send` requires `LazyInner: Sync`, which is why the interior mutability around the deferred compute closure itself uses a `Mutex` (not a `Cell`) — a `Cell` there would silently make `UserBlockCompleted` non-`Send`, breaking the channel it's sent over from the PTY reader thread to the main thread.
- **Cross-crate consumers**: `warp_tui` (a separate crate) also reads these fields; this was invisible to `cargo check`/`cargo test` scoped to the `warp` crate and only surfaced via that crate's own build. Fixed its one call site (`emit_block_completed_telemetry`).

## Testing
- Targeted `cargo test` runs per touched module, all passing: `terminal::model::blocks` (65 tests, including the new stale-index regression test and `test_background_blocks_finished`, which exercises the lazy fields end-to-end), `terminal::input` (211), `util::lazy` (7, including new `get_with` coverage for lock-then-cache and skip-when-cached behavior), `passive_suggestions` (4), `next_command_model` (10), `current_prompt` (14), `aws_credentials` (5), `block_context` (3), `open_in_warp` (8).
- `terminal::view` has 5 pre-existing order-dependent failures (309 passing) under the full-module run; confirmed via `git stash` against unmodified `master` that these fail identically, so they're unrelated to this change.
- `./script/presubmit` passes (format, inline-test-module check, clippy across the workspace); the only remaining failure is a pre-existing, unrelated missing `clang-format` binary in this environment.


app/src/terminal/event.rs
```diff
@@ -286,25 +289,34 @@ pub struct BlockWorkingDirectoryUpdatedEvent {
...
+/// four defer both looking up the live `Block` at `index` and computing their value until first
+/// read (see [`crate::terminal::model::block::compute_command`] and its siblings, used at the
+/// construction site in `impl From<&Block> for BlockType`). Reading any of them locks the given
+/// `&FairMutex<TerminalModel>` only the first time; the result is cached and later reads never
+/// lock again.
 pub struct UserBlockCompleted {
     pub index: BlockIndex,
 
-    pub serialized_block: Arc<SerializedBlock>,
+    serialized_block: Lazy<Arc<SerializedBlock>, FairMutex<TerminalModel>>,
```
vorporeal: yeah we need want to be passing in `TerminalModel`, _not_ `FairMutex<TerminalModel>`.  by passing in the mutex, and requiring that we have not already locked the mutex on this thread, we make it very easy to deadlock.

we need to make the function receive a `&TerminalModel` (a reference to the model _itself_) - this gives the caller decision-making power over whether it needs to lock the model, or whether it _already has exclusive access_ and can simply pass a `&TerminalModel` in.

C0W0: I'm actually wondering if it makes more sense to make the function receive a `&BlockList` instead, since that's what's needed from the `TerminalModel`

crates/warp_util/src/lazy.rs
```diff
@@ -0,0 +1,92 @@
...
+/// Unlike a plain `OnceLock`, the compute closure can capture per-instance data (e.g. an index
+/// used to look up `S`-specific state), so a single `Lazy<T, S>` type can back many differently-
+/// behaving instances. If `S` is itself expensive to obtain (e.g. requires acquiring a lock), that
+/// cost is only paid the first time the value is read; every subsequent read returns the cached
+/// value without invoking the closure (or needing a fresh `&S`) again.
+///
+/// Cloning a `Lazy` is a single `Arc::clone` and shares the same cache, so once any clone
+/// computes the value, every other clone (and the original) observes the cached result instead
+/// of recomputing it.
+pub struct Lazy<T, S>(Arc<LazyInner<T, S>>);
```
acarl005, line 37: This _seems_ like something `std::sync::OnceLock` already provides. Any particular reason to write a custom struct instead of using `OnceLock`? Custom data structures may be cheap to generate with AI, but still require maintenance whereas stdlib components are completely free. It's not clear to me that this custom struct is adding any additional protections or invariants either.

C0W0: This does use `OnceLock` in the `LazyInner` struct. The reason why `OnceLock` alone won't work is that the consumer wouldn't know how to construct the data; instead, the producer must specify the closure used to build it when consumer invokes `get` for the first time. If you checkout `LazyInner` itself, it just consists of the `OnceLock` and the compute closure

C0W0:
* **#15279** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/15279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#15162** <a href="https://app.graphite.com/github/pr/warpdotdev/warp/15162?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/warpdotdev/warp/15162?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

crates/warp_util/src/lazy.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl<T, S> Lazy<T, S> {
+    /// Wraps an already-computed `value`. Reading it via [`Lazy::get`] never invokes a compute
+    /// closure (and therefore never needs a `&S`).
+    pub fn provided(value: T) -> Self {
```
acarl005, line 42:
```suggestion
    #[cfg(any(test, feature = "test-util"))]
    pub fn provided(value: T) -> Self {
```

Looks like this is also test-only.

crates/warp_util/src/lazy.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.0.cell.get_or_init(|| {
+            let compute = self
+                .0
+                .compute
+                .lock()
+                .take()
+                .expect("Lazy value has no cached value and no deferred compute fn");
+            compute(source)
+        })
+    }
```
acarl005, line 71:
```suggestion
```
Looks like this method is unused? It's only used in tests but that's not useful to test if it's unused in production code. I think `fn provided` is valid even if test-only b/c it helps _set up_ useful tests.

crates/warp_util/src/lazy.rs
```diff
@@ -0,0 +1,139 @@
...
+                    .take()
+                    .expect("with_source callback invoked compute more than once");
+                compute(source)
+            })
+        })
+    }
+
+    /// Returns the already-computed value without requiring a `&S`, or `None` if it hasn't been
+    /// computed (or provided) yet.
+    pub fn get_if_cached(&self) -> Option<&T> {
```
acarl005, line 122:
```suggestion
    #[cfg(test)]
    pub fn get_if_cached(&self) -> Option<&T> {
```

This is test-only as well. If this is required for making useful assertions, then keep it and add the test-only attribute. If you don't need this for assertions, I'd delete it altogether instead.

app/src/terminal/event.rs
```diff
@@ -286,25 +288,33 @@ pub struct BlockWorkingDirectoryUpdatedEvent {
 
 #[derive(Clone, Debug)]
 /// Contents of a normal block that a user executed.
+///
+/// `serialized_block`, `command`, `command_with_obfuscated_secrets`, `output_truncated`, and
+/// `output_truncated_with_obfuscated_secrets` are all computed lazily: each defers looking up the
+/// live `Block` and computing its value until first read. This result will then be cached and accessed
+/// directly in subsequent reads
```
acarl005:
```suggestion
```

Fields are already documented in their respective field doc comments. Try to avoid double-documenting. If agents are generating your comments, please have it do a pass over the comments added on the diff and pay special attention to the commenting guidelines in `AGENTS.md`. The `serialized_block` doc comment is sufficient to explain all this behavior.

app/src/terminal/model/block.rs
```diff
@@ -547,66 +621,55 @@ impl From<&Block> for BlockType {
...
+                        index,
+                        Lazy::deferred({
+                            let id = id.clone();
+                            move |block_list| {
+                                resolve_and_compute(block_list, &id, |block| Arc::new(block.into()))
+                            }
+                        }),
+                        Lazy::deferred({
+                            let id = id.clone();
+                            move |block_list| resolve_and_compute(block_list, &id, compute_command)
```
acarl005:
```suggestion
                            move |block_list| {
                                resolve_and_compute(block_list, &id, Block::command_to_string)
                            }
```

You can delete the `fn compute_command` you added as there already exists a method for this.

app/src/terminal/model/block.rs
```diff
@@ -526,6 +528,78 @@ pub struct PromptInfo {
...
+
+/// Resolution is keyed on `BlockId`, not `BlockIndex`: block removal (e.g. clearing the screen)
+/// can reindex the remaining blocks, so a `BlockIndex` captured at construction time may no
+/// longer point at the same block (or may silently resolve to a different one entirely) by the
+/// time a deferred field is first read, possibly much later.
+fn resolve_and_compute<T: Default>(
+    block_list: &BlockList,
+    id: &BlockId,
+    compute: impl FnOnce(&Block) -> T,
+) -> T {
```
acarl005:
```suggestion
) -> Result<T> {
```
We're logging an error and returning a `T::default()`, effectively swallowing an error. Callers won't actual be able to tell that the failure occurred. Are we sure this is how we want to handle the failure instead of wrapping this in either a `Result` or `Option`?

app/src/terminal/local_tty/terminal_manager.rs
```diff
@@ -361,6 +356,13 @@ impl<S> TerminalManager<S> {
         let colors = model.colors();
         let model = Arc::new(FairMutex::new(model));
 
+        // Have ApiKeyManager subscribe to block completion events for AWS credential refresh.
+        // This must happen after `model` is created, since the subscription needs it to resolve
+        // lazily-computed `UserBlockCompleted` fields.
+        ai::api_keys::ApiKeyManager::handle(ctx).update(ctx, |manager, ctx| {
```
acarl005:
```suggestion
        ApiKeyManager::handle(ctx).update(ctx, |manager, ctx| {
```
super nit. convert this to use an import

app/src/terminal/model/block.rs
```diff
@@ -526,6 +528,78 @@ pub struct PromptInfo {
     pub prompt_snapshot: Option<String>,
 }
 
+/// Computes [`UserBlockCompleted::command`] lazily from the live block. See [`warp_util::lazy::Lazy`].
+fn compute_command(block: &Block) -> String {
+    block.command_to_string()
+}
+
+/// Computes [`UserBlockCompleted::command_with_obfuscated_secrets`] lazily from the live block.
+fn compute_command_with_obfuscated_secrets(block: &Block) -> String {
```
acarl005: This function plus the next 2 functions below are better expressed as methods on the `Block`.

app/src/terminal/input.rs
```diff
@@ -15266,11 +15304,21 @@ impl Input {
...
-                        shared_session_history_model.update(ctx, |history_model, _ctx| {
-                            history_model.push(HistoryEntry::for_completed_block(
-                                block_completed.command,
-                                &block_completed.serialized_block,
-                            ))
+                        let command = block_completed
+                            .command
+                            .get_with(|compute| {
+                                let model = self.model.lock();
+                                compute(model.block_list())
```
acarl005, line 15311: This type of closure gets repeated a lot.... I wish there was a nice way to reduce the repetition here. Maybe with a macro? I'm not sure. If there isn't a better way than I'm fine with this. Behavior-wise though it's pretty ideal.

acarl005 (changes requested):

crates/warp_util/src/lazy.rs
C0W0: yes for now - but I want keep this in case we want to use it in the future.

app/src/terminal/input.rs
C0W0: I was thinking about using a macro as well, but the whole idea of putting this out is to make sure the caller performs an explicit lock so that they would be aware of potential deadlock risks. That's why I kept this.

crates/warp_util/src/lazy.rs
C0W0: Same as `provided` - I want to keep this so that we can potentially reuse `Lazy<T, S>` in the future in other place, where it might be feasible to just do `lazy.get(&source)` instead of `lazy.get_with(|compute| compute(&source))`. The reason I made `get_with` in the first place is because we want to have explicit mutex locking - otherwise `get` is suffice.

app/src/terminal/model/block.rs
C0W0: chatted on zoom - keeping this in a macro

acarl005 (approved):

PR #15175, opened 2026-08-15
Create scaffolding that allows CLI to accept name OR ID to refer to environment, runner, and agent objects

sohumgaitonde:
## Description

This PR updates the environment, runner, and agent CLI subcommands to accept either an ID or a name for their positional identifier, instead of requiring the raw ID. 

First, the uid/id CLI args were renamed to identifier, with value_name placeholders and doc comments updated to communicate that either form is accepted.

Regarding name resolution logic, each object type now resolves its identifier argument through the same general pattern: try it as an ID first, and if that does not resolve, fall back to matching it against the object's name.
•  Environments: identifier is parsed as a server ID and looked up directly; if that fails, it's matched by name against the full list of environments through a local scan on the on the CloudAmbientAgentEnvironment objects
•  Runners: identifier is matched against runner UIDs first, then against runner names (for lookup), and resolved to the underlying UID before the runner is acted on.
•  Agents: identifier is inspected for ID shape to decide whether to call the by-ID or by-name lookup endpoint, with a fallback to lookup with the by-name backend endpoint if the by-ID call 404s.

Across all three, name-based lookups return an explicit error when no object matches or when more than one object shares the same name.

Server-side enforcement of name uniqueness and creation of the by-name endpoint on the server (and associated resolution logic) is out of scope for this PR.

## Linked Issue

- [ ] 

https://linear.app/warpdotdev/issue/REMOTE-2008/enforce-name-uniqueness-and-allow-using-name-instead-of-id

## Testing

### DIFF
--- a/app/src/ai/agent_sdk/environment.rs
+++ b/app/src/ai/agent_sdk/environment.rs
@@ -766,8 +766,8 @@ impl EnvironmentCommandRunner {
                     OperationSuccessType::Failure => {
                         super::report_fatal_error(
                             anyhow::anyhow!(
-                                "Failed to create environment: the request to Warp did not \
-                                 complete. Check your network connection and try again."
+                                "Failed to create environment: the request could not be \
+                                 completed. Please try again."
                             ),
                             ctx,
                         );
@@ -775,9 +775,7 @@ impl EnvironmentCommandRunner {
                     OperationSuccessType::Rejection => {
                         super::report_fatal_error(
                             anyhow::anyhow!(
-                                "Failed to create environment: the server rejected the request. \
-                                 It may have already been created; list your environments \
-                                 before retrying."
+                                "Failed to create environment: the server rejected the request."
                             ),
                             ctx,
                         );
@@ -785,8 +783,7 @@ impl EnvironmentCommandRunner {
                     OperationSuccessType::FeatureNotAvailable => {
                         super::report_fatal_error(
                             anyhow::anyhow!(
-                                "Failed to create environment: cloud environments are not \
-                                 available for your account."
+                                "Failed to create environment: cloud environments are unavailable."
                             ),
                             ctx,
                         );
--- a/app/src/ai/agent_sdk/environment.rs
+++ b/app/src/ai/agent_sdk/environment.rs
@@ -763,9 +763,31 @@ impl EnvironmentCommandRunner {
                     OperationSuccessType::Denied(message) => {
                         super::report_fatal_error(anyhow::anyhow!("{message}"), ctx);
                     }
-                    _ => {
+                    OperationSuccessType::Failure => {
                         super::report_fatal_error(
-                            anyhow::anyhow!("Failed to create environment"),
+                            anyhow::anyhow!(
+                                "Failed to create environment: the request to Warp did not \
+                                 complete. Check your network connection and try again."
+                            ),
+                            ctx,
+                        );
+                    }
+                    OperationSuccessType::Rejection => {
+                        super::report_fatal_error(
+                            anyhow::anyhow!(
+                                "Failed to create environment: the server rejected the request. \
+                                 It may have already been created; list your environments \
+                                 before retrying."
+                            ),
+                            ctx,
+                        );
+                    }
+                    OperationSuccessType::FeatureNotAvailable => {
+                        super::report_fatal_error(
+                            anyhow::anyhow!(
+                                "Failed to create environment: cloud environments are not \
+                                 available for your account."
+                            ),
                             ctx,
                         );
                     }
--- a/app/src/ai/agent_sdk/environment.rs
+++ b/app/src/ai/agent_sdk/environment.rs
@@ -749,12 +749,24 @@ impl EnvironmentCommandRunner {
         ctx.subscribe_to_model(&UpdateManager::handle(ctx), move |_, _, event, ctx| {
             if let UpdateManagerEvent::ObjectOperationComplete { result } = event
                 && matches!(result.operation, ObjectOperation::Create { .. })
-                && matches!(result.success_type, OperationSuccessType::Success)
                 && result.client_id == Some(client_id)
             {
-                let server_id = result.server_id.unwrap();
-                println!("Environment created successfully with ID: {server_id}");
-                ctx.terminate_app(warpui::platform::TerminationMode::ForceTerminate, None);
+                match &result.success_type {
+                    OperationSuccessType::Success => {
+                        let server_id = result.server_id.unwrap();
+                        println!("Environment created successfully with ID: {server_id}");
+                        ctx.terminate_app(warpui::platform::TerminationMode::ForceTerminate, None);
+                    }
+                    OperationSuccessType::Denied(message) => {
+                        super::report_fatal_error(anyhow::anyhow!("{message}"), ctx);
+                    }
+                    _ => {
+                        super::report_fatal_error(
+                            anyhow::anyhow!("Failed to create environment"),
+                            ctx,
+                        );
+                    }
+                }
             }
         });
     }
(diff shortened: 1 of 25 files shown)


captainsafia (commented):
This looks like a good start! One thing that might help with completeness here is updating the implementation to do resolution for the resource types that support it.  
  
For agent objects, we query for those via the REST API so we'll likely need to couple it with a change in the backend to support querying for agents by name, something like an `/agents/by-name?name=foobar` endpoint.  
  
For environments, we can do a local scan on the `CloudAmbientAgentEnvironment` objects.  
  
For runners, it seems like we'll just want to make sure that we call resolve_runner everywhere consistently.  
  
Also, it might be neat to feed this PR to a cloud agent and have it validate it with computer use then record a video. Testing the CLI should be fairly computer use friendly.

sohumgaitonde:
## E2E verification (name-or-ID CLI resolution + server-side uniqueness enforcement)

Verified this branch end-to-end against a local `warp-server` build of `add-by-name-endpoint` (which includes the stacked uniqueness-enforcement branch + the new `GET /agent/identities/by-name` endpoint), using a real user-authenticated local session and a from-source `warp` CLI build. Full recorded walkthrough: https://oz.staging.warp.dev/artifacts/01a01bfa-8fa2-750a-aa2b-5b11c667a1c7

### Result summary
- **Environments**: get/update/delete by name and by UID — pass. Nonexistent name — clear error. Duplicate-name ambiguity (`Multiple environments match '...'; specify the environment by ID`) — pass.
- **Runners**: create/update/delete resolve consistently by name and UID via the shared `resolve_runner` helper. Nonexistent name — clear error. Audited every CLI call site that accepts a runner reference (`agent run-cloud --runner`, `runner update`, `runner delete`) — all three resolve identically; no other call sites exist.
- **Agents**: get/update/delete by name and UID — pass. Nonexistent name — clean 404. UUIDv7-shaped-name collision (agent named a syntactically valid UUIDv7 not matching any real UID) — `resolve_agent_identifier`'s try-as-UID → 404 → retry-as-name fallback works correctly (confirmed via server logs: `GET .../identities/<uuid>` → 404 → `GET .../identities/by-name?name=<uuid>` → 200).
- **Uniqueness enforcement interplay** (tested on agents): duplicate name on create, rename collision, rename-then-relookup, delete-then-recreate — all behave correctly (409 with a clear "name already exists" message; by-name index reflects current state, not stale).

### Bug found: `environment create` hangs indefinitely on a name conflict instead of erroring

Reproduction: `environment create --name <already-taken-name> --docker-image ...` produces no output and never returns (confirmed hung 90s+, had to be killed). By contrast, `runner create` and `agent create` both fail fast (<1s) with a clear "name already exists" error for the identical conflict.

Root cause: `EnvironmentCommandRunner::create_environment_after_auth_check` (`app/src/ai/agent_sdk/environment.rs`) subscribes to `UpdateManagerEvent::ObjectOperationComplete` but only handles `OperationSuccessType::Success`:

```rust
if let UpdateManagerEvent::ObjectOperationComplete { result } = event
    && matches!(result.operation, ObjectOperation::Create { .. })
    && matches!(result.success_type, OperationSuccessType::Success)
    && result.client_id == Some(client_id)
```

`app/src/server/sync_queue.rs` already emits this same event with `OperationSuccessType::Failure` for `CreationFailureReason::UniqueKeyConflict` (and `Other`), plus `Denied(...)`/`Rejection`/`FeatureNotAvailable` variants — none of these are matched, so the callback body never runs and the CLI waits forever. This is a pre-existing gap in `environment create`'s UpdateManager/SyncQueue-based creation path that the new server-side uniqueness enforcement now actually exercises for the first time in practice (`runner create`/`agent create` use direct request/response calls instead, which is why they don't hit this).

Suggested fix: add `else` arms for `Failure`/`Denied`/`Rejection`/`FeatureNotAvailable` in that subscription that call `super::report_fatal_error(...)`, mirroring how the runner/agent create paths surface their conflict errors.

_Verification performed by an autonomous agent; all local server/test setup was disposable and out-of-band (no code changes committed here)._

sohumgaitonde:
## Re-verified: the environment-create hang is fixed (commit ebe4cf49)

Re-ran live verification against the latest push. Confirmed:

- **The hang is fixed.** `environment create` on an already-taken name now fails fast (~7.5-7.7s across 3 trials, consistent with a request round-trip) instead of hanging 90+ seconds indefinitely.
- **Regression check (all pass):** environment create/get/update/delete by name and by UID, delete-then-recreate same name. Runner create/update/delete spot-checked as unaffected (different, non-UpdateManager code path).

### Minor follow-up worth considering: generic error message

The new catch-all `_` arm reports `Error: Failed to create environment` for a name conflict, since `CreationFailureReason::UniqueKeyConflict` maps to the generic `OperationSuccessType::Failure` (not `Denied(message)`) in `update_manager.rs`. `runner create`/`agent create` give a specific "name already exists" style message for the identical conflict — `environment create` now fails fast (good!) but doesn't tell the user *why*. Not a blocker, but worth a follow-up to map `UniqueKeyConflict` to a specific message the same way `Denied` is handled, for parity with runner/agent.

### New (pre-existing, previously unobservable) finding: stale local pending records after a failed create

Now that failed `environment create` attempts return promptly instead of hanging forever, a follow-on effect becomes observable: the CLI's local SQLite cache keeps an optimistic "Unsynced" pending record for the failed attempt (created before the server round-trip, never cleaned up on failure). This can cause a later `environment get <name>`/`update <name>` to incorrectly report `Multiple environments match '<name>'; specify the environment by ID` even though only one row exists server-side (verified directly against `generic_string_objects` in Postgres — exactly one live row per name). This is a separate, minor client-side cache-cleanup gap, not a regression from this fix or a server-side duplication bug — flagging in case it's worth a follow-up to clear the optimistic record when `ObjectOperationComplete` reports a non-`Success` result.

Video of this re-verification: https://oz.staging.warp.dev/artifacts/01a01c79-7c26-7a4f-b4da-f97a23e62a1a

_Verification performed by an autonomous agent; all local server/test setup was disposable and out-of-band (no code changes committed here)._

app/src/ai/agent_sdk/environment.rs
```diff
@@ -760,12 +749,24 @@ impl EnvironmentCommandRunner {
...
+                        let server_id = result.server_id.unwrap();
+                        println!("Environment created successfully with ID: {server_id}");
+                        ctx.terminate_app(warpui::platform::TerminationMode::ForceTerminate, None);
+                    }
+                    OperationSuccessType::Denied(message) => {
+                        super::report_fatal_error(anyhow::anyhow!("{message}"), ctx);
+                    }
+                    _ => {
+                        super::report_fatal_error(
+                            anyhow::anyhow!("Failed to create environment"),
```
captainsafia: Is there a more specific error message that we can give here?

app/src/server/server_api/factory.rs
```diff
@@ -44,6 +53,40 @@ pub trait FactoryClient: 'static + Send + Sync {
...
+/// uid match against [`FactoryClient::get_runners`].
+#[cfg_attr(target_family = "wasm", allow(dead_code))]
+pub async fn get_runner_with_fallback(
+    factory: &dyn FactoryClient,
+    selector: RunnerSelector,
+) -> Result<Runner> {
+    let uid = selector.uid.clone();
+    match factory.get_runner(selector).await {
+        Ok(runner) => Ok(runner),
+        Err(err) if is_missing_get_runner_query_error(&err) => {
```
captainsafia: I feel like we can avoid the fallback here and just say we couldn't retrieve the runner if we get into this case.

captainsafia (approved):
We'll want to make sure this is merged after the server-side queries land.

PR #15277, opened 2026-08-18
Force code editor for global search matches so line jump always works

### LINKED ISSUE #9657: Global file search should jump to matching line

arpansahu:
## Description

Fixes #9657.

Clicking a global search result opens the matching file, but when the file is markdown and the user has the "prefer Markdown Viewer" setting enabled, the match opened at the top of the file instead of jumping to the matched line.

### Root cause

Tracing `GlobalSearchViewEvent::OpenMatch` -> `LeftPanelEvent::OpenFileWithTarget` -> `open_file_with_target`:
- For `FileTarget::CodeEditor`, `line_col` is correctly threaded through `open_or_focus_existing` -> `jump_to_line_col_in_tab` -> `ScrollPosition::LineAndColumn`, and it works.
- For `FileTarget::MarkdownViewer`, the file is opened via `open_file_notebook` -> `FilePane::new`, neither of which accepts a `line_col` parameter at all. It's structurally impossible for the Markdown Viewer to jump to a line today.

So any search match in a markdown file, when the Markdown Viewer preference is on, silently loses the requested line and opens at the top.

### Fix

In `LeftPanelView::handle_global_search_event` (`app/src/workspace/view/left_panel.rs`), force `FileTarget::CodeEditor` for search-match opens, for both local and remote paths, overriding the Markdown Viewer preference specifically for this flow. A search match always carries a specific line to jump to, and only the code editor supports that today, so this is the minimal, targeted fix (the underlying Markdown Viewer limitation is a separate, larger gap that would need a `line_col` parameter threaded through `open_file_notebook`/`FilePane`, out of scope for this issue).

## Linked Issue

- [x] I have linked the issue(s) this PR resolves, and I have confirmed a maintainer has added the "ready-to-implement" label to them.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

I don't have access to a macOS/Linux GUI environment capable of running the full Warp app locally in this sandbox (no Metal toolchain license acceptance possible), so I was not able to capture a screen recording per the template. To compensate, I verified correctness through:

- Deep manual tracing of the full call chain (`OpenMatch` -> `OpenFileWithTarget` -> `open_file_with_target` -> `open_code`/`open_file_notebook`) confirming `line_col` is dropped only in the Markdown Viewer path, and that `FileTarget::CodeEditor`/`FileTarget::MarkdownViewer` variant signatures (`EditorLayout`) match my usage exactly.
- Full workspace compile verification: cross-compiled the `warp` package (`cargo check -p warp --target x86_64-unknown-linux-gnu`) successfully with zero errors, using a Linux cross-toolchain to route around the local Xcode/Metal licensing blocker (this repo's macOS-only Metal shader build is skipped for non-macOS targets).
- `cargo clippy -p warp --target x86_64-unknown-linux-gnu -- -D warnings` - zero warnings/errors.
- `cargo fmt --check -p warp` - clean.

I'm happy to add a video/screenshot if a maintainer can point me to a way to test this locally, or if CI produces build artifacts I can run.

### DIFF
--- a/app/src/workspace/view/left_panel.rs
+++ b/app/src/workspace/view/left_panel.rs
@@ -53,9 +53,7 @@ use crate::util::bindings::keybinding_name_to_display_string;
 use crate::util::file::external_editor::EditorSettings;
 use crate::util::openable_file_type::FileTarget;
 #[cfg(feature = "local_fs")]
-use crate::util::openable_file_type::{
-    EditorLayout, is_markdown_file, resolve_file_target_with_editor_choice,
-};
+use crate::util::openable_file_type::{EditorLayout, resolve_file_target_with_editor_choice};
 use crate::workspace::WorkspaceAction;
 use crate::workspace::view::conversation_list::view::{
     ConversationListView, Event as ConversationListViewEvent,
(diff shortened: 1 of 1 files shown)


arpansahu:
@cla-bot check

arpansahu:
Addressed the second review point: the remote-path comment no longer claims to mirror the file tree's Markdown Viewer handling. It now states explicitly that this path always opens the code editor and ignores `prefer_markdown_viewer`, because search matches always carry a specific line that the Markdown Viewer cannot jump to.

The requested screen recording is still to come; I'll follow up with it before asking for re-review.

PR #15533, opened 2026-08-25
Add Credits/Dollars usage display unit toggle to Appearance settings

Xavientois:
## Description
Adds a persisted GUI Appearance setting, `UsageDisplayUnit` (`Credits` default / `Dollars`), gated by `FeatureFlag::PricingTransparency`, that lets users switch the primary unit for usage/spend displays.

**What**
- New `UsageDisplayUnit` setting in `app/src/settings/ai.rs` (`surface: GUI`, `SyncToCloud::Globally(RespectUserSyncSetting::Yes)`, default `Credits`).
- New dropdown control (`UsageDisplayUnitWidget`) in the Appearance settings page, in a new "Usage" category, gated by the feature flag.
- `format_credits_with_cost` in `app/src/ai/blocklist/view_util.rs` reworked to take a `UsageDisplayUnit` and show tokens plus exactly one of credits or dollars — never both — e.g. `"12,345 tokens / $0.36"` or `"12,345 tokens / 20 credits"`. Falls back to a plain credits figure when dollars are unavailable or the feature flag is off.
- Wired the setting through usage/spend call sites: the AI block's inline usage pill (`ai/blocklist/block/view_impl/output.rs`), the conversation usage footer/settings view (`conversation_usage_view.rs`), the conversation details panel's "Credits used" row, and the settings usage-history header row.
- Brand-new accounts default to `Dollars`; existing users/devices keep the enum's own `Credits` default untouched.

**Note:** This does not update the billing and usage page, as that currently shows both credits and dollars and will be updated in a subsequent PR

## Testing
- `./script/format` — no diffs beyond my changes.
- `cargo clippy -p warp --all-targets --tests -- -D warnings` — clean. (`--all-features` also enables `warp_completer`'s `v2` feature, which fails clippy on unmodified `master` too — confirmed via `git stash` — unrelated to this change.)
- `cargo nextest run -p warp --no-fail-fast` targeted at touched modules (`view_util`, `usage_display_unit`, `conversation_usage_view`, `conversation_details_panel`, `usage_history_entry`, `appearance_page`, `onboarding`) — all passing (159+ tests).
- Added unit tests: formatter behavior (both units, cost-unavailable fallback, flag-off fallback, thousands separators), settings round-trip + toml path, and new-account-vs-skipped-account default behavior.

### Screenshots / Videos

https://www.loom.com/share/36664e78ec604a908879f23b1f46ade7

#### New user gets dollars by default

https://oz.staging.warp.dev/artifacts/01a03feb-1da2-72d5-92c7-b207009dd319

## Agent Mode

### DIFF
--- a/app/src/ai/blocklist/block/view_impl/output.rs
+++ b/app/src/ai/blocklist/block/view_impl/output.rs
@@ -104,7 +104,7 @@ use crate::ai::blocklist::keyboard_navigable_buttons::KeyboardNavigableButtons;
 use crate::ai::blocklist::secret_redaction::SecretRedactionState;
 use crate::ai::blocklist::usage::rollup::compute_orchestration_rollup;
 use crate::ai::blocklist::view_util::{
-    FAILED_OUTPUT_USAGE_NOTICE_TEXT, format_credits_with_cost, format_usage_parenthetical,
+    FAILED_OUTPUT_USAGE_NOTICE_TEXT, format_credits_with_cost,
     should_show_failed_output_usage_notice,
 };
 use crate::ai::blocklist::{AIBlockResponseRating, BlocklistAIActionModel, SuggestionChipView};
@@ -116,6 +116,7 @@ use crate::ai::skills::{
 use crate::appearance::Appearance;
 use crate::code::diff_viewer::DisplayMode;
 use crate::code::editor_management::CodeSource;
+use crate::settings::AISettings;
 use crate::settings_view::SettingsSection;
 use crate::terminal::ShellLaunchData;
 #[cfg(not(target_family = "wasm"))]
@@ -3722,9 +3723,14 @@ fn render_usage_button(props: Props, app: &AppContext) -> Box<dyn Element> {
         Icon::ChevronRight
     };
 
+    let usage_display_unit = AISettings::as_ref(app).usage_display_unit;
     let total_credits_spent = headline_credits;
-    let mut credit_usage_text =
-        format_credits_with_cost(total_credits_spent, headline_tokens, headline_cost_in_cents);
+    let mut credit_usage_text = format_credits_with_cost(
+        total_credits_spent,
+        headline_tokens,
+        headline_cost_in_cents,
+        usage_display_unit,
+    );
     if let Some(credits_spent_for_last_block) = conversation.credits_spent_for_last_block() {
         // Only show the credits spent for the last block if it is different from the total credits spent
         // and we spent a non-zero amount of credits for the last block.
@@ -3734,26 +3740,17 @@ fn render_usage_button(props: Props, app: &AppContext) -> Box<dyn Element> {
             && total_credits_spent != credits_spent_for_last_block
             && props.model.status(app).error().is_none()
         {
-            // If the first part of the decimal is 0, we just display the whole number.
-            let last_block_credits_text = if credits_spent_for_last_block.fract() < 0.1 {
-                format!("{}", credits_spent_for_last_block.trunc() as i32)
-            } else {
-                format!("{credits_spent_for_last_block:.1}")
-            };
             // The last-block figure has no rollup equivalent: it stays
             // bound to the orchestrator's own last block, same as
             // `credits_spent_for_last_block` above.
             let last_block_charged_usage = conversation.charged_usage_for_last_block();
-            let last_block_detail = format_usage_parenthetical(
+            let last_block_text = format_credits_with_cost(
+                credits_spent_for_last_block,
                 last_block_charged_usage.map(|usage| usage.total_tokens()),
                 last_block_charged_usage.map(|usage| usage.total_cost_in_cents()),
+                usage_display_unit,
             );
-            credit_usage_text = match last_block_detail {
-                Some(detail) => {
-                    format!("{credit_usage_text} (+{last_block_credits_text}, {detail})")
-                }
-                None => format!("{credit_usage_text} (+{last_block_credits_text})"),
-            };
+            credit_usage_text = format!("{credit_usage_text} (+{last_block_text})");
         }
     }
 
--- a/app/src/ai/blocklist/block/view_impl/output.rs
+++ b/app/src/ai/blocklist/block/view_impl/output.rs
@@ -3725,7 +3725,7 @@ fn render_usage_button(props: Props, app: &AppContext) -> Box<dyn Element> {
 
     let usage_display_unit = AISettings::as_ref(app).usage_display_unit;
     let total_credits_spent = headline_credits;
-    let mut credit_usage_text = format_credits_with_cost(
+    let mut usage_text = format_credits_with_cost(
         total_credits_spent,
         headline_tokens,
         headline_cost_in_cents,
@@ -3750,7 +3750,7 @@ fn render_usage_button(props: Props, app: &AppContext) -> Box<dyn Element> {
                 last_block_charged_usage.map(|usage| usage.total_cost_in_cents()),
                 usage_display_unit,
             );
-            credit_usage_text = format!("{credit_usage_text} (+{last_block_text})");
+            usage_text = format!("{usage_text} (+{last_block_text})");
         }
     }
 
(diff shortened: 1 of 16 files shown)


app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3734,26 +3740,17 @@ fn render_usage_button(props: Props, app: &AppContext) -> Box<dyn Element> {
             && total_credits_spent != credits_spent_for_last_block
             && props.model.status(app).error().is_none()
         {
-            // If the first part of the decimal is 0, we just display the whole number.
-            let last_block_credits_text = if credits_spent_for_last_block.fract() < 0.1 {
-                format!("{}", credits_spent_for_last_block.trunc() as i32)
-            } else {
-                format!("{credits_spent_for_last_block:.1}")
-            };
```
Xavientois, line 3742: Why was this removed?

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3722,9 +3723,14 @@ fn render_usage_button(props: Props, app: &AppContext) -> Box<dyn Element> {
         Icon::ChevronRight
     };
 
+    let usage_display_unit = AISettings::as_ref(app).usage_display_unit;
     let total_credits_spent = headline_credits;
-    let mut credit_usage_text =
-        format_credits_with_cost(total_credits_spent, headline_tokens, headline_cost_in_cents);
+    let mut credit_usage_text = format_credits_with_cost(
```
Xavientois: We should rename this appropriately to `usage_text` since it's not guaranteed to be in credits

app/src/settings_view/billing_and_usage/usage_history_entry.rs
```diff
@@ -111,8 +112,13 @@ impl UsageHistoryEntry {
 
         let total_credits =
             entry.usage_metadata.credits_spent + entry.usage_metadata.platform_credits_spent;
+        let usage_display_unit = AISettings::as_ref(app).usage_display_unit;
+        // GAP: this GraphQL-backed source doesn't yet thread through a token
+        // count or dollar cost (see `ConversationUsageInfo::from` in
+        // `gql_convert.rs`), so this always falls back to a plain credits
+        // figure today; it will pick up tokens/dollars once that gap closes.
```
Xavientois: Remove this comment. I will fix that separately

app/src/ai/blocklist/block/view_impl/output.rs
Xavientois: the manual formatting was redundant with `format_credits`, and consolidating into `format_credits_with_cost` was needed so the last-block delta also respects the new unit setting

app/src/ai/blocklist/view_util.rs
```diff
@@ -298,59 +299,71 @@ pub fn format_credits(credits: f32) -> String {
...
+pub fn usage_label(
+    credits_label: &'static str,
+    dollars_label: &'static str,
+    unit: UsageDisplayUnit,
+) -> &'static str {
+    if matches!(unit, UsageDisplayUnit::Dollars) && FeatureFlag::PricingTransparency.is_enabled() {
+        dollars_label
+    } else {
+        credits_label
+    }
```
Xavientois: The agent extracted this as a separate helper function, but I would probably opt to just put the if statement inline for clarity. lmk what you'd prefer

I am open to either

jefflloyd: Yeah, something about this doesn't seem right to me. And part of this is preexisting / made worse because you have to augment this surface, not your code itself. The fact that it's invoked like this:

```
usage_label(
    "Credits spent (last response)",
    "Usage charged (last response)",
    usage_display_unit,
),
```

... feels super smelly. Like, we need to duplicate `(last response)` or `(total)`; if they show up in one label they show up in the other, so passing them in separately seems wrong. We keep passing in magic strings rather than letting the formatter decide on the right string to use. It just feels like the wrong level of code abstraction.

What if the function is like:

```
#[derive(Clone, Copy)]
pub enum UsageLabelKind {
    LastResponse,
    Total,
    Plain,
    DetailsPanel,
}

pub fn usage_label(kind: UsageLabelKind, unit: UsageDisplayUnit) -> String {
    // The Dollars setting only takes effect once PricingTransparency is on.
    let unit = if FeatureFlag::PricingTransparency.is_enabled() {
        unit
    } else {
        UsageDisplayUnit::Credits
    };
    let base = match (kind, unit) {
        (UsageLabelKind::DetailsPanel, UsageDisplayUnit::Credits) => "Credits used",
        (UsageLabelKind::DetailsPanel, UsageDisplayUnit::Dollars) => "Usage",
        (
            UsageLabelKind::LastResponse | UsageLabelKind::Total | UsageLabelKind::Plain,
            UsageDisplayUnit::Credits,
        ) => "Credits spent",
        (
            UsageLabelKind::LastResponse | UsageLabelKind::Total | UsageLabelKind::Plain,
            UsageDisplayUnit::Dollars,
        ) => "Usage charged",
    };
    let suffix = match kind {
        UsageLabelKind::LastResponse => " (last response)",
        UsageLabelKind::Total => " (total)",
        UsageLabelKind::Plain | UsageLabelKind::DetailsPanel => "",
    };
    format!("{base}{suffix}")
}
```

app/src/ai/blocklist/view_util.rs
```diff
@@ -298,59 +299,71 @@ pub fn format_credits(credits: f32) -> String {
...
-    format!("{credits_text} ({parenthetical})")
+    format!("{} tokens / {unit_text}", tokens.separate_with_commas())
+}
+
+/// Returns `dollars_label` when `unit` is `Dollars` and
+/// `FeatureFlag::PricingTransparency` is enabled, otherwise returns
+/// `credits_label`. Used to swap usage-row labels (e.g. "Credits spent" ->
+/// "Usage charged") to match the unit shown by [`format_credits_with_cost`]
+/// -- gated the same way so a row's label and its formatted value never
+/// disagree (e.g. a "Usage" label next to a plain credits fallback figure).
```
jefflloyd: I think a lot of these comments could benefit from a conciseness pass. Most of the code is self-describing.

app/src/settings/onboarding.rs
```diff
@@ -29,6 +29,21 @@ pub(crate) fn apply_account_first_onboarding_settings(
...
+    // not create an account, so this only fires for genuinely new accounts.
+    if account_class.is_some() {
+        AISettings::handle(app).update(app, |settings, ctx| {
+            report_if_error!(
+                settings
+                    .usage_display_unit
+                    .set_value(UsageDisplayUnit::Dollars, ctx)
+            );
+        });
+    }
```
jefflloyd: I'd want to stress test this... and maybe see if there's other prior art for ways of saying "everyone who's new gets setting A, everyone old gets setting B" - I feel like we've done this before.

app/src/settings_view/billing_and_usage/usage_history_entry.rs
```diff
@@ -111,8 +112,9 @@ impl UsageHistoryEntry {
 
         let total_credits =
             entry.usage_metadata.credits_spent + entry.usage_metadata.platform_credits_spent;
+        let usage_display_unit = AISettings::as_ref(app).usage_display_unit;
         let credits_spent = Text::new_inline(
-            format_credits(total_credits as f32),
+            format_credits_with_cost(total_credits as f32, None, None, usage_display_unit),
```
jefflloyd: Might be fine / might be moot re: the updates you're gonna add, but passing two `None`s in here makes me a little suspicious that this is the right function API

app/src/settings/onboarding.rs
Xavientois: Discussed in person. We'll stick with this approach

app/src/settings_view/billing_and_usage/usage_history_entry.rs
Xavientois: yeah, the updates will change this. The `None`s are for the token/dollar amounts, but we are only showing credits here. This is the billing surface that I'm doing in a separate PR

app/src/ai/blocklist/view_util.rs
Xavientois: updated

jefflloyd (approved):

PR #15544, opened 2026-08-25
[Windows] Fix login-shell commands in PowerShell 5.1

harryalbert:
## Description
Windows PowerShell 5.1 has never supported `-Login`, but `LocalCommandExecutor` passed it to login-shell commands, causing automatic CLI plugin installation and other callers to fail before their command ran. The bug predates Warp's public repository import, with equivalent Sentry failures dating back to April.

It surfaced as a new spike after the August 19 release preserved `PluginInstallError` as a typed error, moving existing failures into [WARP-CLIENT-BETA-STABLE-88JX](https://warpdotdev.sentry.io/issues/7683536490). This change omits the optional login flag for PowerShell and adds a regression test using Windows PowerShell 5.1.

## Testing
- Added a Windows-only regression test that executes a login-shell command through `powershell.exe`

CHANGELOG-BUG-FIX: Fixed automatic CLI plugin installation failing in Windows PowerShell 5.1.

### DIFF
--- a/app/src/terminal/model/session/command_executor/local_command_executor_tests.rs
+++ b/app/src/terminal/model/session/command_executor/local_command_executor_tests.rs
@@ -212,3 +212,32 @@ mod unix {
         });
     }
 }
+
+#[cfg(windows)]
+mod windows {
+    use std::path::PathBuf;
+
+    use super::super::*;
+
+    #[test]
+    fn executes_login_command_with_windows_powershell() {
+        futures_lite::future::block_on(async {
+            let powershell_path = PathBuf::from(
+                std::env::var_os("SystemRoot").expect("SystemRoot should be set on Windows"),
+            )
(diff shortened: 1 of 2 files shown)


app/src/terminal/model/session/command_executor/local_command_executor_tests.rs
```diff
@@ -212,3 +212,32 @@ mod unix {
...
 }
+
+#[cfg(windows)]
+mod windows {
+    use std::path::PathBuf;
+
+    use super::super::*;
+
+    #[test]
+    fn executes_login_command_with_windows_powershell() {
```
acarl005: This test does not assert what it says it is testing. I'd just delete this test tbh

acarl005 (approved):
Thanks for fixing! Stamped but I'd delete this test

PR #15587, opened 2026-08-27
Support stored screenshot references in computer-use tasks

danielpeng2:
## Description
Client half of computer-use screenshot offload (server: warpdotdev/warp-server#16148). The server now swaps inline screenshot bytes in task messages for small `StoredScreenshotRef`s via `UpdateTaskMessage` — and those updates target messages from **earlier** turns.

That's the new part. A task lives in two forms: the proto `source` (echoed back to the server) and per-turn `exchanges` (rendered UI). `UpdateTaskMessage` handling assumed the target message was created by the current stream: it errored if the stream had no exchange for the task, and rendered the update into the current turn either way. Both are wrong for history edits.

The fix routes by ownership: `upsert_message` updates the exchange that actually contains the message (current-stream exchange only for genuinely new ones), and finished exchanges now render updates. Same-stream behavior is unchanged, pinned by a regression test.

Also: refs survive conversion/restore round-trips, and the lightbox fetches ref-only screenshots via signed URL. `Cargo.toml` temporarily patches the proto crate to a local checkout until warpdotdev/warp-proto-apis#367 publishes.

## Linked Issue
- [ ] No linked issue — tracked via the Warp conversation below.
- [ ] No UI change; no screenshots.

## Testing
4 new `task_tests.rs` tests (same-stream regression, cross-exchange swap, no-current-exchange, new-message error); scoped nextest/check/clippy/format all clean. Presubmit + master-merge owed before leaving draft.

- [ ] Not manually tested yet — needs the server flag (staging or local server); planned before marking ready.

## Agent Mode

CHANGELOG-NONE

---
[Warp conversation](https://staging.warp.dev/conversation/94d4736c-ab24-4a6c-8279-416ad3dc61e3) · [Plan](https://staging.warp.dev/drive/notebook/wD0rnN6nnopfiXz580bLet)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -582,7 +582,7 @@ tikv-jemalloc-sys = { git = "https://github.com/warpdotdev/jemallocator.git", re
 
 [patch."https://github.com/warpdotdev/warp-proto-apis.git"]
 # Uncomment for local development of warp-proto-apis
-# warp_multi_agent_api = { path = "../warp-proto-apis/apis/multi_agent/v1/gen/rust" }
+warp_multi_agent_api = { path = "../warp-proto-apis/apis/multi_agent/v1/gen/rust" }
 
 [patch."https://github.com/warpdotdev/session-sharing-protocol.git"]
 # Uncomment for local development of session-sharing-protocol
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -349,7 +349,7 @@ version-compare = "0.1"
 vte = { git = "https://github.com/warpdotdev/vte.git", rev = "4b399c87b63ba88f45709edaa6383fc519f6c900", default-features = false }
 walkdir = "2"
 warp-workflows = { git = "https://github.com/warpdotdev/workflows", rev = "793a98ddda6ef19682aed66364faebd2829f0e01" }
-warp_multi_agent_api = { git = "https://github.com/warpdotdev/warp-proto-apis.git", rev = "2bd5df2e3a73ada980d1e9c613ce369232e8c41d" }
+warp_multi_agent_api = { git = "https://github.com/warpdotdev/warp-proto-apis.git", rev = "7b638a3bce15b6159116c25caa1dda62cf004b0e" }
 wasm-bindgen = "0.2.89"
 wasm-bindgen-futures = "0.4.42"
 web-sys = { version = "0.3.69", features = [
@@ -582,7 +582,7 @@ tikv-jemalloc-sys = { git = "https://github.com/warpdotdev/jemallocator.git", re
 
 [patch."https://github.com/warpdotdev/warp-proto-apis.git"]
 # Uncomment for local development of warp-proto-apis
-warp_multi_agent_api = { path = "../warp-proto-apis/apis/multi_agent/v1/gen/rust" }
+# warp_multi_agent_api = { path = "../warp-proto-apis/apis/multi_agent/v1/gen/rust" }
 
 [patch."https://github.com/warpdotdev/session-sharing-protocol.git"]
 # Uncomment for local development of session-sharing-protocol
--- a/app/src/ai/blocklist/block.rs
+++ b/app/src/ai/blocklist/block.rs
@@ -51,6 +51,7 @@ use warp_editor::content::buffer::InitialBufferState;
 use warp_editor::content::edit::resolve_asset_source_relative_to_directory;
 use warp_editor::render::element::VerticalExpansionBehavior;
 use warp_errors::{report_error, report_if_error};
+use warp_multi_agent_api::StoredScreenshotRef;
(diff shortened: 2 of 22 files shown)


app/src/server/server_api/ai.rs
```diff
@@ -1465,6 +1473,14 @@ pub trait AIClient: 'static + Send + Sync {
         artifact_uid: &str,
     ) -> anyhow::Result<ArtifactDownloadResponse, anyhow::Error>;
 
+    /// Fetches a signed download URL for a computer-use screenshot stored in
+    /// Warp-managed object storage. Requires view access to the conversation.
+    async fn get_screenshot_download(
```
vorporeal: for my understanding - is the reason why we're doing the separate round-trip for a screenshot URL because it's a signed URL with a TTL?

i'm torn between wanting to be able to proactively send an auth-required URL (`warp-server` doing some auth and then proxying the data stream without buffering any substantial amount in-memory) and wanting to keep `warp-server` out of the data streaming path.

crates/ai/src/agent/action_result/mod.rs
```diff
@@ -1147,15 +1148,30 @@ impl Display for ReadSkillResult {
 #[derive(Debug, Clone, PartialEq)]
 pub enum UseComputerResult {
     /// Computer use succeeded, with one result per requested action.
-    Success(computer_use::ActionResult),
+    Success {
+        result: computer_use::ActionResult,
+        /// Present when the screenshot bytes live in Warp-managed object
+        /// storage instead of inline in `result`.
+        stored_screenshot_ref: Option<StoredScreenshotRef>,
```
vorporeal: what's the rationale behind this being a sibling of `ActionResult` as opposed to being inside it?  not asserting that it's wrong, i just don't immediately grok why this structure.

Cargo.toml
```diff
@@ -579,7 +579,7 @@ tikv-jemalloc-sys = { git = "https://github.com/warpdotdev/jemallocator.git", re
 
 [patch."https://github.com/warpdotdev/warp-proto-apis.git"]
 # Uncomment for local development of warp-proto-apis
-# warp_multi_agent_api = { path = "../warp-proto-apis/apis/multi_agent/v1/gen/rust" }
+warp_multi_agent_api = { path = "../warp-proto-apis/apis/multi_agent/v1/gen/rust" }
```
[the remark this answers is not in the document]
vorporeal: i approved the proto changes, so we should be able to swap this out for the merged proto commit ref

app/src/ai/blocklist/block.rs
```diff
@@ -7112,46 +7128,111 @@ impl TypedActionView for AIBlock {
...
-                    images,
-                    initial_index,
-                });
+                if refs_to_fetch.is_empty() {
+                    open_screenshot_lightbox(&ordered_action_ids, initial_index, ctx);
+                    return;
+                }
+
+                // Resolve each stored ref to a signed URL and download the image
+                // bytes before opening the lightbox.
```
vorporeal: why do this before opening the lightbox?  i feel like we ought to open it and display the caption and a loading indicator, no?

otherwise, if the signed URL resolution is a bit slow, it'll look like nothing happened when you clicked on the button.

along those lines - should we be using `AssetSource::Async` here, to let the asset cache manage the task to fetch the data?  are we not doing so in order to detect when something fails (so we can display a toast)?  if so, i wonder if there's some other way we can handle that.  (we've got a lot of the pieces in place around creating futures that resolve when an async asset resolves, but i'm not sure we quite expose everything in the right ways/places)

app/src/server/server_api/ai.rs
danielpeng2: yeah the signed URL has a TTL so we need to make sure we're fetching a fresh URL.

we follow the same pattern of doing an extra round-trip when loading artifacts in the client. my thinking was to keep warp-server out of the streaming path since that seemed simpler.

crates/ai/src/agent/action_result/mod.rs
danielpeng2: this was originally because `computer_use::ActionResult` is the result of a local computer use action. it felt weird to couple it with the remote screenshot ref, since a local computer use action couldn't possibly return a remote ref.

i agree that it's a little weird though. i updated to more closely match the proto type, with an enum representing either inlined screenshot bytes or a reference to a remote screenshot.

app/src/ai/blocklist/block.rs
danielpeng2: good call, updated to open the lightbox and show a loading state. the new approach does use `AssetSource::Async` and also handles the `FailedToLoad` state that we get from the asset cache.

vorporeal (approved):

PR #15622, opened 2026-08-27
CLIAgentSessionStatus updates should update ConversationStatus

seemeroland:
## Description

There used to be some weird logic about the `CLIAgentSessionStatus` (what's updated on 3p harness hooks) only updating the conversation status for child conversations. It should always keep conversation status in sync.

To match which conversation this hook updated the CLIAgentSessionStatus for, we consult the `LocalAgentTaskSyncModel` instead of assuming it's the active conversation of the pane. Only ambient agent 3p harnesses have a Conversation associated with them (not 3p harnesses the user ran themselves)

### DIFF
--- a/app/src/terminal/view.rs
+++ b/app/src/terminal/view.rs
@@ -13413,10 +13413,7 @@ impl TerminalView {
     /// CLI-harness conversation needs this bridge just as much as a child
     /// one does (e.g. for orchestration event delivery, which reads
     /// `ConversationStatus` directly).
-    fn conversation_id_for_cli_status_updates(
-        &self,
-        ctx: &AppContext,
-    ) -> Option<AIConversationId> {
+    fn conversation_id_for_cli_status_updates(&self, ctx: &AppContext) -> Option<AIConversationId> {
         let task_id =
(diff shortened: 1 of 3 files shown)


app/src/terminal/view.rs
```diff
@@ -13411,30 +13393,46 @@ impl TerminalView {
...
+    /// into an unrelated conversation that happens to be live in the same
+    /// pane (e.g. an earlier native Agent Mode conversation).
+    ///
+    /// Not scoped to child (orchestration) conversations: nothing else in
+    /// the codebase keeps a CLI-harness conversation's `ConversationStatus`
+    /// in sync with the harness's own lifecycle hooks, so a root/standalone
+    /// CLI-harness conversation needs this bridge just as much as a child
+    /// one does (e.g. for orchestration event delivery, which reads
+    /// `ConversationStatus` directly).
+    fn conversation_id_for_cli_status_updates(&self, ctx: &AppContext) -> Option<AIConversationId> {
```
liliwilson: mega nit but can we call this `conversation_id_for_cli_agent_status_updates` or smth? so it's clear that this codepath never runs for normal convos

liliwilson (approved):
thanks for fixing 🙏

PR #15752, opened 2026-09-02
Skip project rule files under gitignored directories

danielpeng2:
## Description
Rule files (`WARP.md` / `AGENTS.md`) under gitignored directories are no longer discovered as project rules.

Motivation: in a factory run, `yarn install` created `client/node_modules/recharts/AGENTS.md`. It showed up in `available_rules`, which rewrote the first prompt message and invalidated ~200k cached tokens on every later turn.

Two parts:

1. **Rule recording respects gitignore.** Rules under a gitignored *ancestor* are skipped. Rules that are only *directly* gitignored (e.g. a personal `WARP.md`) are kept. Skills are exempt since `.agents/skills` etc. are gitignored by design.
2. **Nested `.gitignore` files are honored on the watcher path.** Only root + global gitignores were consulted before, so `client/node_modules` (ignored by `client/.gitignore`) looked tracked. A new helper resolves the gitignores between repo root and a path; the watcher update, the Linux descend filter, and on-demand subdir watches use it. Side effect: `yarn install` no longer registers inotify watches across `node_modules`.

## Testing
`cargo nextest run -p repo_metadata`: 145 passed, including four new tests. Clippy and fmt clean.

- [x] I have manually tested my changes locally with `./script/run` (covered by unit tests)
  - Ran locally and verified that project rules under node modules aren't included

## Agent Mode

CHANGELOG-BUG-FIX: Rule files (`WARP.md`/`AGENTS.md`) inside gitignored directories such as `node_modules` are no longer picked up as project rules.

### DIFF
--- a/crates/repo_metadata/src/entry.rs
+++ b/crates/repo_metadata/src/entry.rs
@@ -272,9 +272,12 @@ impl Entry {
         // ignored/symlinked), a classification failure at the root propagates to
         // the caller, preserving existing error behavior.
         if let Some(state) = standing_queries.as_deref_mut() {
-            state
-                .results
-                .record_path(&root_path, root_path.is_dir(), state.definitions);
+            state.results.record_path(
+                &root_path,
+                root_path.is_dir(),
+                ancestor_is_ignored,
+                state.definitions,
+            );
         }
         match evaluate_entry(
             &root_path,
@@ -378,6 +381,7 @@ impl Entry {
                             state.results.record_path(
                                 &child_path,
                                 child_path.is_dir(),
+                                job.ignored,
                                 state.definitions,
                             );
                         }
@@ -1006,6 +1010,9 @@ fn descend_allowlist_matches(suffix: &[Component<'_>]) -> bool {
 /// pruned to avoid following trees outside the repository, and any other
 /// gitignored directory is pruned so we don't register watches on
 /// `node_modules`, build output, vendored deps, etc.
+///
+/// `gitignores` is the repository's root + global set; `.gitignore` files in
+/// directories between `repo_root` and `path` are consulted as well.
 pub fn should_watch_repo_directory(
     path: &Path,
     repo_root: &Path,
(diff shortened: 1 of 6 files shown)


crates/repo_metadata/src/entry.rs
```diff
@@ -1032,11 +1039,38 @@ pub fn should_watch_repo_directory(
...
+        &gitignores_for_path_in_repo(repo_root, path, gitignores),
         /* check_ancestors */ true,
     )
 }
 
+/// Extends `base` (the repository's root + global gitignores) with every `.gitignore` found in
+/// the directories strictly between `repo_root` and `path`, ordered root-first, so that `path`
+/// is matched the way git matches it. `.gitignore` files inside `path` itself are not included
+/// because they cannot ignore `path`; tree traversal picks those up when it descends.
+pub fn gitignores_for_path_in_repo(
```
kevinyang372, line 1051: Should this be async? We are parsing the gitignores in here yeah?

crates/repo_metadata/src/entry.rs
```diff
@@ -1032,11 +1039,38 @@ pub fn should_watch_repo_directory(
...
+    };
+    let Some(relative_parent) = relative.parent() else {
+        return gitignores;
+    };
+    let mut directory = repo_root.to_path_buf();
+    for component in relative_parent.components() {
+        directory.push(component);
+        let gitignore_path = directory.join(".gitignore");
+        if gitignore_path.is_file() {
+            gitignores.push(gitignore_cache::get_or_parse(&gitignore_path));
```
kevinyang372, line 1068: Also I think we should be caching these gitignore results rather than rereading + parsing them on every call path (gitignore_cache::get_or_parse doesn't prevent re-read)

crates/repo_metadata/src/standing_queries.rs
```diff
@@ -122,7 +123,10 @@ impl StandingQueryResults {
             self.project_skills
                 .insert(StandingQueryContent::file(standardized.clone()));
         }
-        if !is_directory && definitions.is_project_rule_file(path) {
+        // Rules under gitignored directories (e.g. `node_modules`) aren't project guidance, but a
+        // directly-ignored rule file in a tracked directory may be a personal rule, so keep it.
+        // Skills are exempt: provider directories like `.agents/skills` are gitignored by design.
+        if !is_directory && !ancestor_is_ignored && definitions.is_project_rule_file(path) {
```
kevinyang372, line 129: If we want to be consistent, should we ignore the project skills and skill provider files as well?

PR #15801, opened 2026-09-04
Queue 3p follow-ups until setup done, never create a native conversation for CLI-harness ambient tasks

seemeroland:
## Problem

For a third-party-harness (e.g. Claude Code) ambient agent run, a shared-session-injected prompt/follow-up that arrives before the CLI harness session is fully live could fall through the Oz-only prompt path and create a brand-new native `AIConversation`. That conversation's server token then gets reported via `UpdateAgentTask`, silently overwriting the run's real conversation ID and permanently breaking follow-ups for that run (`could not update agent conversation ID: invalid state`, repeated on every retry until cancelled).

This is the client-side slice of a cross-repo fix. We queue 3p follow-ups until the 3p harness starts. We also improve logging, function naming, and types

## Changes

- `terminal_view_adaptor.rs`: `accept_agent_prompt` is now the single choke point that routes an incoming agent prompt to exactly one destination: a live CLI-harness PTY, a new `PendingCliHarnessPromptQueue` (for a task registered against a third-party harness whose session hasn't started a live PTY yet, detected via `LocalAgentTaskSyncModel::task_id_for_terminal_view` independent of session-start state), or Warp's native Oz harness for every other case.
- New `PendingCliHarnessPromptQueue` singleton (`app/src/ai/blocklist/pending_cli_harness_prompt_queue.rs`): a minimal FIFO queue, keyed by `AmbientAgentTaskId`, holding text-only prompts (file attachments are dropped with a warning log, since CLI-harness PTY delivery doesn't support them) until the task's harness session starts.
- `AgentDriver::subscribe_to_cli_agent_session_events`: drains the queue on `CLIAgentSessionsModelEvent::StatusChanged{InProgress}` — a genuine, plugin-reported confirmation the harness is already running, rather than the earlier `Started` event which fires at command-detection time, before the harness has processed anything. A bounded 8-minute fallback timer (below the ~10 minutes the server allows before expecting the agent to have started working) drains anyway if no qualifying event ever arrives; draining is idempotent, so the fallback is a silent no-op once the normal path has already delivered. If the fallback timer actually finds prompts to deliver, it logs a warning, since a healthy plugin reports `InProgress` almost immediately — this indicates a broken or missing plugin integration (e.g. a Codex session using only the OSC 9 notification fallback, whose events never map to `InProgress`). Each prompt is delivered as a genuine PTY follow-up via `TerminalDriver::send_text_to_cli`; the queue is cleared on `unregister_cli_agent_task_sync`.
- `BlocklistAIController`: renamed `execute_agent_prompt_for_shared_session` → `execute_warp_agent_prompt_from_shared_session_injection` and `send_shared_session_query` → `send_warp_agent_prompt_from_shared_session_injection`, with doc comments clarifying these are now Oz/native-only, since third-party-harness routing is decided upstream in `accept_agent_prompt`.
- Added a defense-in-depth guard in `send_warp_agent_prompt_from_shared_session_injection`'s no-token fallback: refuses to create a native conversation (and reports an error) for a task that is CLI-harness-backed, checking two independent signals — the runtime registration (`task_id_for_terminal_view`) and the task's stored config (`AmbientAgentTask::is_third_party_harness`) — so the check still holds even before a session ever registers, honoring the existing setup-failure-debug-bootstrap exception.
- `LocalAgentTaskSyncModel` stays a pure task-state-sync model (the queueing responsibility moved to the new dedicated singleton above).
- `HarnessSupportClient` and third-party harness runners (Claude Code, Codex, Gemini) now use the safer `ServerConversationToken` type instead of the client-local `AIConversationId` for third-party-harness conversation identifiers, since these conversations are never represented in `BlocklistAIHistoryModel`.
- Added an unconditional receipt log at the top of the `NetworkEvent::AgentPromptRequested` handler in `terminal_view_adaptor.rs` (before the feature-flag check and any permission/routing logic), recording the request id, participant id, and whether a `server_conversation_token` was present, so every received shared-session prompt is observable regardless of how it's later routed or rejected.
- Added/updated unit tests covering: `PendingCliHarnessPromptQueue` (FIFO drain, per-task isolation, clear), and the `LocalAgentTaskSyncModel` test suite trimmed of the now-removed queueing responsibility.

### Roland manual testing using e2e test

[Modified e2e tests](https://github.com/warpdotdev/oz-e2e-tests/pull/14) to send multiple followups in succession while sandbox closed, along with https://github.com/warpdotdev/session-sharing-server/pull/516. Claude code combines them in one prompt, while codex sends both individually. Both 3p harnesses queue until the first prompt is done. Not ideal but acceptable for now

Screenshot 2026-09-08 at 2 51 10 PM
Screenshot 2026-09-08 at 2 51 23 PM

- `cargo check -p warp` (lib and `--tests`), `cargo fmt -p warp --check`, and `cargo test -p warp --lib` filtered to the touched modules (`pending_cli_harness_prompt_queue`, `local_agent_task_sync_model`, `shared_session`, `terminal_view_adaptor`, `checkpoint_coordinator`, `codex_tests`, `snapshot_tests`, `agent_driver`/`driver`): 433 passed, 1 unrelated pre-existing flaky failure (`cloud_provider::gcp::tests::best_effort_command_is_killed_on_timeout`, passes in isolation) as of commit 82ea3499c, confirmed by the orchestrating agent on a machine with sufficient memory (this sandbox OOMs partway through compiling the `warp` crate on every attempt).
- `cargo fmt -p warp --check`: clean as of the latest commit.
- Latest commit's (ae44f901c) compile/test status not yet independently re-verified due to the sandbox's memory limit; manual review confirms no dangling references to removed/renamed symbols.

### DIFF
--- a/app/src/ai/agent_sdk/driver.rs
+++ b/app/src/ai/agent_sdk/driver.rs
@@ -39,7 +39,9 @@ use warp_graphql::ai::{AgentTaskState, PlatformErrorCode};
 use warp_managed_secrets::ManagedSecretValue;
 use warp_util::local_or_remote_path::LocalOrRemotePath;
 use warpui::r#async::{FutureExt, TimeoutError, Timer};
-use warpui::{AppContext, Entity, ModelContext, ModelHandle, ModelSpawner, SingletonEntity};
+use warpui::{
+    AppContext, Entity, EntityId, ModelContext, ModelHandle, ModelSpawner, SingletonEntity,
+};
 
 use crate::ai::agent::conversation::{AIConversationId, ConversationStatus};
 use crate::ai::agent::{
@@ -224,6 +226,15 @@ const TASK_STATUS_FLUSH_TIMEOUT: Duration = Duration::from_secs(5);
 const MANAGED_MCP_RESOLVE_MAX_ATTEMPTS: usize = 6;
 /// Timeout for individual harness auth preflight commands.
 const PREFLIGHT_CHECK_TIMEOUT: Duration = Duration::from_secs(30);
+/// Bounded fallback for draining `PendingCliHarnessPromptQueue`: some CLI-harness sessions
+/// (e.g. Codex using only its OSC 9 notification fallback, when the platform plugin is
+/// unavailable or disabled) never emit a genuine `CLIAgentSessionsModelEvent::StatusChanged`
+/// with `CLIAgentSessionStatus::InProgress` — Codex's OSC 9 path only ever produces opaque
+/// `Stop` notifications, which map to `Success`, not `InProgress`. Once this window elapses
+/// after harness setup, drain and deliver any still-queued prompts anyway rather than losing
+/// them forever; draining is idempotent, so this is a no-op when the normal `InProgress`-driven
+/// drain already ran.
+const PENDING_CLI_HARNESS_PROMPT_QUEUE_FALLBACK_DRAIN_TIMEOUT: Duration = Duration::from_secs(30);
 pub(crate) const WARP_DRIVE_SYNC_TIMEOUT: Duration = Duration::from_secs(60);
 /// Maximum time to wait for an automatic error resume before propagating the error.
 /// If no follow-up status arrives within this window, the driver terminates with the
@@ -5005,6 +5016,22 @@ impl AgentDriver {
             LocalAgentTaskSyncModel::handle(ctx).update(ctx, |model, ctx| {
                 model.register_cli_session(terminal_view_id, task_id, ctx);
             });
+
+            // See `PENDING_CLI_HARNESS_PROMPT_QUEUE_FALLBACK_DRAIN_TIMEOUT`: bounded fallback
+            // in case this session never emits a genuine `StatusChanged{InProgress}` (e.g. a
+            // Codex session on the OSC 9 fallback path).
+            ctx.spawn(
+                async move {
+                    Timer::after(PENDING_CLI_HARNESS_PROMPT_QUEUE_FALLBACK_DRAIN_TIMEOUT).await;
+                },
+                move |me, _, ctx| {
+                    me.drain_and_deliver_pending_cli_harness_prompts(
+                        task_id,
+                        terminal_view_id,
+                        ctx,
+                    );
+                },
+            );
         }
 
         ctx.subscribe_to_model(&CLIAgentSessionsModel::handle(ctx), move |me, _, event, ctx| match event {
@@ -5057,6 +5084,20 @@ impl AgentDriver {
                                 me.task_id
                             );
                             harness_exit.cancel_idle_timeout();
+
+                            // A genuine, plugin-reported confirmation that the harness is
+                            // already running — unlike the synthetic `InProgress` optimistically
+                            // set at registration (which never emits `StatusChanged`; see
+                            // `LocalAgentTaskSyncModel::register_cli_session`) — so it's safe to
+                            // deliver any shared-session prompts queued while waiting for this
+                            // task's harness session to start.
+                            if let Some(task_id) = me.task_id {
+                                me.drain_and_deliver_pending_cli_harness_prompts(
+                                    task_id,
+                                    terminal_view_id,
+                                    ctx,
+                                );
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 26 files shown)


app/src/ai/agent_sdk/driver.rs
```diff
@@ -5115,12 +5137,47 @@ impl AgentDriver {
...
+        for prompt in queued {
+            log::info!(
+                "Delivering a shared-session prompt that had been queued while waiting for \
+                 this task's CLI-harness session to start (event=queued_shared_session_prompt_delivered \
+                 task_id={task_id} terminal_view_id={terminal_view_id:?} \
+                 participant_id={:?})",
+                prompt.participant_id
+            );
+            self.terminal_driver.update(ctx, |terminal_driver, ctx| {
+                terminal_driver.send_text_to_cli(prompt.prompt, ctx);
```
seemeroland, line 5178: What actually happens depends on the 3p harness. Both claude code and codex queue until the agent stops from the first resume prompt, which seems okay for now but probably not ideal. We probably want the follow-up to interrupt like it does for warp agent

vorporeal: i.e.: we want "steer", not "queue", semantics?  i know codex supports this; i wonder if there's a way we can guarantee that the injected prompt is injected in "steer" mode.

app/src/ai/blocklist/controller/shared_session.rs
```diff
@@ -661,8 +663,16 @@ impl BlocklistAIController {
         }
     }
 
-    /// Execute an agent prompt on behalf of the viewer.
-    pub fn execute_agent_prompt_for_shared_session(
+    /// Execute an agent prompt on behalf of the viewer, against Warp's native Oz harness.
+    ///
+    /// Callers must have already routed away third-party-harness-backed tasks: this method (and
```
vorporeal, line 668: really feels like we're missing a trait here that provides a clear interface for interacting with a harness, and implementations for warp/cc/codex.

(not saying you should add such a trait now)

app/src/ai/blocklist/controller/shared_session.rs
```diff
@@ -845,6 +855,26 @@ impl BlocklistAIController {
...
     }
 
+    /// The task ID a no-token prompt landing right now must not spawn a native `AIConversation`
+    /// for, because it is (or is configured to be) backed by a third-party CLI-harness session.
+    /// Two independent signals are checked, since either can be true without the other:
+    /// - `LocalAgentTaskSyncModel::cli_harness_task_id_for_terminal_view`: the harness session has been
+    ///   registered for this pane (true from harness setup time, before its process launches).
+    /// - `AmbientAgentTask::is_third_party_harness`: the task's stored config says it runs on a
+    ///   third-party harness, independent of whether a session has registered for this pane yet
+    ///   (e.g. very early in setup, or if registration is ever skipped by a bug).
```
vorporeal, line 865: how can `cli_harness_task_id_for_terminal_view()` be `Some` without `AmbientAgentTask::is_third_party_harness()` being true?  that seems very odd to me.

app/src/ai/blocklist/pending_cli_harness_prompt_queue.rs
```diff
@@ -0,0 +1,79 @@
...
+#[cfg_attr(target_family = "wasm", allow(dead_code))]
+#[derive(Default)]
+pub(crate) struct PendingCliHarnessPromptQueue {
+    pending: HashMap<AmbientAgentTaskId, Vec<QueuedCliHarnessPrompt>>,
+}
+
+pub(crate) enum PendingCliHarnessPromptQueueEvent {}
+
+impl PendingCliHarnessPromptQueue {
+    pub fn new(_ctx: &mut ModelContext<Self>) -> Self {
```
vorporeal, line 41: kinda odd that we treat the whole struct as dead code on wasm but not the constructor?

consider replacing `allow` with `expect` so that the accuracy of the annotation is actually enforced by the linter (instead of simply suppressing the finding).

vorporeal (approved):
lgtm.  really highlights to me the extent to which we're missing proper abstractions between the agent driver and the underlying harness. :/

app/src/ai/blocklist/controller/shared_session.rs
seemeroland: I just took a look at the codepaths and this should not be possible. But it is possible for `cli_harness_task_id_for_terminal_view` to be `None` and `AmbientAgentTask::is_third_party_harness()` to be true, if the session had not been registered yet, which is what this is guarding against.

I do wish we had clearer model responsibility

PR #15869, opened 2026-09-08
Fall back to a forge identity only when nothing else claimed a repo

joeywangzr:
## Description
Cloud agent sandboxes silently overrode a customer's own `git config --global user.name/email` setup command, so commits kept using Warp's identity instead of the customer's.

**Root cause**: `configure_repository_git_identity` unconditionally wrote a repo-*local* `user.name`/`user.email` right after cloning, before setup commands ran. Local config always beats `--global`, so a customer's later override never took effect. This per-repo stamp exists for a reason: mixed-forge sandboxes (e.g. GitHub + GitLab) need a different identity per repo, which a single `--global` value can't express.

**Fix**: snapshot the process-wide identity once at bootstrap, before any clone or setup command runs. After setup commands finish, only apply Warp's forge-appropriate identity to a repo if its *effective* identity still matches that snapshot. If it differs, something already claimed it (a `--global` override, a per-repo local one, etc.), so it's left alone. This also correctly handles a mixed-forge sandbox where the customer wants one shared identity across every repo.

## Linked Issue
- [ ] N/A — found and fixed while investigating a customer report of cloud agent commits authored with the wrong git identity.

## Testing
- Added unit tests for `repository_identity_changed_since` and `configure_repository_git_identity_if_unset` using real temp git repos.
- `environment` (43 tests) and `git_credentials` (25 tests) module suites pass; `cargo check`/`clippy -D warnings` clean for this diff.
- Not tested end-to-end in a live sandbox (no production credentials available here) — the fix is scoped to small, independently-testable pure functions.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed cloud agent commits sometimes ignoring a customer's `git config --global user.name/email` setup command.

### DIFF
--- a/app/src/ai/agent_sdk/driver/git_credentials.rs
+++ b/app/src/ai/agent_sdk/driver/git_credentials.rs
@@ -654,18 +654,89 @@ fn recorded_identity_for_host(host: &str) -> Option<(String, String)> {
 /// Write `user.name`/`user.email` into one repository's LOCAL git config,
 /// selecting the identity of the forge that hosts it.
 ///
-/// Repo-local config always wins over `--global` config, so callers should
-/// only reach for this when a sandbox genuinely spans more than one forge and
-/// therefore can't rely on a single process-wide `--global` identity (set by
-/// [`configure_git_identity`]) to represent every repo correctly. Calling this
-/// unconditionally on a single-forge sandbox would permanently defeat a later
-/// `--global` override, such as a customer's own setup command.
-pub(crate) fn configure_repository_git_identity(repository_dir: &std::path::Path, host: &str) {
+/// Repo-local config always wins over `--global` config, so most callers want
+/// [`configure_repository_git_identity_if_unset`] instead, which only applies
+/// this when nothing has already claimed the repo's identity. Calling this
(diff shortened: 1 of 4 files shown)


app/src/ai/agent_sdk/driver/git_credentials.rs
```diff
@@ -651,16 +651,114 @@ fn recorded_identity_for_host(host: &str) -> Option<(String, String)> {
...
+    if !output.status.success() {
+        return None;
+    }
+    let value = String::from_utf8_lossy(&output.stdout).trim().to_string();
+    (!value.is_empty()).then_some(value)
+}
+
+/// The `user.name`/`user.email` pair currently in effect at `scope` (see
+/// [`read_git_config`]). `None` unless both are set.
+fn read_git_identity(scope: &GitIdentityScope) -> Option<(String, String)> {
```
captainsafia, line 707: The fact that we rely on both to be set together means that there's no way for the user to only override the name. In practice, I think this is fine since we'd expect the name and email of the identity to be written together but we should stay alert to this in case it causes issues (user overrides just the name in the setup commands and we ignore that).

captainsafia (approved):

PR #15883, opened 2026-09-09
fix(terminal): let mouse events through kitty graphics image placements

### LINKED ISSUE #15882: Mouse events are dropped for cells covered by a kitty graphics image in alt-screen apps
### Pre-submit Checks

- [ ] I have an issue with AI and have included the debugging ID
- [ ] I have technical issue and have included the logs

### Describe the bug

When an alternate-screen application has mouse reporting enabled and displays a kitty graphics image, Warp stops delivering mouse events whose pointer position falls inside the image's rectangle. The application receives nothing at all for those events: no wheel reports, no clicks. Events land normally one column outside the image, and resume the moment the image is deleted. A fully transparent image behaves the same way, so the trigger is the placement itself, not its pixels. kitty, WezTerm and Ghostty deliver these events regardless of images.

In practice this means any TUI that draws an image over its text, for example a highlight band or a preview, loses scrolling and clicking under that image.

### To reproduce

1. Save the script below as `repro.py` (Python 3 with Pillow, no other dependencies).
2. Run `python3 repro.py` in Warp. It enters the alternate screen, enables SGR mouse reporting, prints a numbered ruler and logs every mouse report it receives.
3. Scroll the wheel over the ruler: report lines appear.
4. Press `i`. A framed, transparent kitty image is placed over the pointer's row.
5. Scroll again with the pointer inside the frame: no report lines. Move the pointer outside the frame on the same row: reports resume.
6. Press `c` to delete the image: reports resume everywhere.

<details>
<summary>repro.py</summary>

```python
#!/usr/bin/env python3
"""
Minimal reproduction: Warp swallows ALL mouse reports for cells covered by a
kitty-graphics image placement.

Run it in Warp (needs Pillow):   python3 repro.py
Then follow the on-screen instructions with a real mouse/trackpad.

Expected (correct) behaviour, as in kitty/WezTerm/Ghostty: every scroll notch and
click prints a report line, whether or not it lands on the image.
Observed in Warp 0.2026.07.22.09.01.01: reports stop entirely while the pointer is
inside the image rectangle, and resume the moment it leaves it or the image is deleted.
"""
import base64, io, os, re, select, sys, termios, tty

def png(w, h):
    from PIL import Image
    im = Image.new("RGBA", (w, h), (0, 0, 0, 0))          # fully transparent interior
    for x in range(w):
        im.putpixel((x, 0), (255, 0, 0, 200)); im.putpixel((x, h - 1), (255, 0, 0, 200))
    for y in range(h):
        im.putpixel((0, y), (255, 0, 0, 200)); im.putpixel((w - 1, y), (255, 0, 0, 200))
    b = io.BytesIO(); im.save(b, "PNG"); return base64.standard_b64encode(b.getvalue()).decode()

o = sys.stdout
def w(s): o.write(s); o.flush()

fd = sys.stdin.fileno(); old = termios.tcgetattr(fd); tty.setraw(fd)
w("\x1b[?1049h\x1b[2J\x1b[H\x1b[?1000h\x1b[?1006h")     # alt screen + SGR mouse
try:
    for i in range(1, 25):
        w("row %02d %s\r\n" % (i, "." * 60))
    # a 700x320px translucent placement anchored at row 8, col 3 (a=T, f=100, C=1)
    data = png(700, 320)
    w("\x1b7\x1b[8;3H")
    w("\x1b_Ga=T,f=100,C=1,m=0;%s\x1b\\" % data)
    w("\x1b8\x1b[26;1H")
    w("image placed (red outline). scroll INSIDE it -> no reports; scroll OUTSIDE -> reports.\r\n")
    w("press 'd' to delete the image (reports resume everywhere), 'q' to quit.\r\n")
    sgr = re.compile(rb"\x1b\[<(\d+);(\d+);(\d+)([Mm])")
    n = 0
    while True:
        if not select.select([fd], [], [], 0.2)[0]:
            continue
        b = os.read(fd, 4096)
        if b"q" in b: break
        if b"d" in b: w("\x1b_Ga=d\x1b\\"); w("\r\nimages deleted\r\n")
        for m in sgr.finditer(b):
            n += 1
            w("\r\n#%d  b=%s col=%s row=%s %s" %
              (n, m.group(1).decode(), m.group(2).decode(),
               m.group(3).decode(), m.group(4).decode()))
finally:
    w("\x1b_Ga=d\x1b\\\x1b[?1006l\x1b[?1000l\x1b[?1049l")
    termios.tcsetattr(fd, termios.TCSADRAIN, old)
    print("total mouse reports received: %d" % n)
```

</details>

### Expected behavior

Mouse reports are delivered to the application regardless of whether the pointer is over an image placement. Images are decorative and should not capture input.

### Where it appears to happen

From reading the source at the current master: `app/src/terminal/grid_renderer.rs:593-594` starts a new scene layer for foreground kitty placements, and `crates/warpui_core/src/scene.rs:609` records each image's rect into that layer's hit map with `click_through` false. `AltScreenElement::paint` captured its own z-index earlier at `app/src/terminal/alt_screen/alt_screen_element.rs:686`, so `Scene::is_covered` (`scene.rs:400-420`) judges the event as covered by an overlay and the element returns false at `alt_screen_element.rs:846-849`, before the scroll-or-report decision in `view.rs` is reached. Calling the existing `set_active_layer_click_through()` right after the image layer is started looks like the fix; I'm happy to open a PR with a regression test if that is welcome.

None needed; the script above reproduces it in under a minute.

### Operating system (OS)

macOS

macOS 15.5 (24F74)

### Shell Version

zsh 5.9

### Current Warp version

v0.2026.07.22.09.01.01

rchougule:
## Description

Foreground kitty graphics placements (`z_index >= 0`, the kitty default) are painted into a layer that `render_grid` pushes with `Scene::start_layer` after the grid has already painted its cells. `Scene::draw_image` records a hit rect for every image it draws, and the layer it lands on is a plain `Layer::default()` with `click_through == false`. Because `AltScreenElement::paint` captures its z-index at the start of paint — before `render_grid` pushes that image layer — the image layer sits strictly above the alt-screen element's own z-index, so `Scene::is_covered` reports every point inside the placement as covered. `DispatchedEvent::at_z_index` then returns `None` for scroll-wheel and button events there and `AltScreenElement::dispatch_event` bails out early. The result is that an alt-screen app which has enabled mouse reporting receives no reports at all for cells an image covers — not wheel, not clicks — and no view scroll happens either, since the event is dropped before the scroll-vs-report decision is reached. Image alpha is irrelevant; a fully transparent PNG records the same hit rect, which is why a full-pane placement makes an entire pane look mouse-dead. `BlockListElement` is unaffected because it records `child_max_z_index` at the end of paint.

Image placements are decorative terminal content and should never capture pointer input, so this marks that layer click-through via the existing `Scene::set_active_layer_click_through`, which `Scene::is_covered` already skips. The two identical `start_layer` call sites in `render_grid` (the plain and the ligature path) now go through a small `start_foreground_image_layer` helper so the intent is documented in one place and the two paths cannot drift. Nothing else about how images are painted changes — they still render above the cells, they just no longer swallow events aimed at the grid underneath.

## Linked Issue

Closes #15882

The issue is triaged as `bug` / `repro:high` / `factory-auto-implement`. It does not carry `ready-to-implement`, so the box below is left unticked — filed ahead of triage as a bug fix per CONTRIBUTING.

## Testing

- `app/src/terminal/grid_renderer_tests.rs` — `test_foreground_image_layer_does_not_cover_the_grid` is the regression test: it paints a stand-in grid layer, calls `start_foreground_image_layer`, records a hit rect inside it, and asserts a wheel position over the placement is not covered at the grid's z-index. That is exactly the check `DispatchedEvent::at_z_index` performs before `AltScreenElement::dispatch_event` returns `false`, so it fails on `master` and passes with the fix.
- `crates/warpui_core/src/scene_tests.rs` — two tests covering the scene behaviour the fix relies on. `test_click_through_layer_with_image_does_not_cover_lower_layers` goes through `Scene::draw_image` (rather than `draw_rect_with_hit_recording`, as the existing click-through test does) to confirm an image's recorded hit rect on a click-through layer does not cover a lower z-index, and `test_image_on_regular_layer_covers_lower_layers` pins the opposite for a normal layer, which is the behaviour that produced the bug.

Presubmit: this machine only has the Command Line Tools, not full Xcode, so `crates/warpui`'s build script cannot run `xcrun metal` and `./script/presubmit` cannot complete end to end here. The parts that do not need the Metal toolchain all pass — `./script/format --check`, `./script/check_no_inline_test_modules`, the factory-files validator, and `cargo clippy -p warpui_core --all-targets --tests -- -D warnings`. To get the app crate compiled and the tests run I stubbed `xcrun metal`/`metallib` locally (nothing in the repo changed): with that in place `cargo clippy -p warp --lib --tests -- -D warnings` is clean and `cargo test -p warp --lib -- terminal::grid_renderer::tests` and `cargo test -p warpui_core --lib -- scene::tests` both pass, 7 and 6 tests respectively. CI should cover the full presubmit on a machine with the Metal toolchain.

Manual check: the `repro.py` script in the issue is the reproduction. It enters the alternate screen, enables SGR mouse reporting, places one translucent 700x320 PNG with `\x1b_Ga=T,f=100,C=1,m=0;<base64>\x1b\\`, and prints a numbered line per mouse report. On a current build, scrolling with the pointer inside the image's outline prints nothing while scrolling one column outside it reports normally, and deleting the images with `a=d` makes reports resume everywhere — the column-precise behaviour that pointed at the scene hit rect rather than a mode reset. The expected behaviour after this change is a report per notch whether or not the pointer is over the placement, matching kitty, WezTerm and Ghostty.

I could not launch a local dev build to capture before/after screenshots for the same Metal toolchain reason. Happy to add them if a reviewer wants them, or if it is easier the repro script above reproduces it in a few seconds on any build.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed mouse events (scroll and clicks) being dropped for terminal cells covered by a kitty graphics image, which made alternate-screen apps with mouse reporting enabled appear unresponsive over images.

### DIFF
--- a/app/src/terminal/grid_renderer.rs
+++ b/app/src/terminal/grid_renderer.rs
@@ -591,7 +591,7 @@ fn render_grid_without_ligatures<'a>(
         }
 
         if !foreground_image_ids.is_empty() {
-            ctx.scene.start_layer(warpui::ClipBounds::ActiveLayer);
+            start_foreground_image_layer(ctx.scene);
             for image_placement in foreground_image_ids {
                 if let Some((image_metadata, image_placement_data)) = image_metadata
                     .get(&image_placement.image_id)
@@ -1100,7 +1100,7 @@ fn render_grid_with_ligatures<'a>(
         }
 
         if !foreground_image_ids.is_empty() {
-            ctx.scene.start_layer(warpui::ClipBounds::ActiveLayer);
+            start_foreground_image_layer(ctx.scene);
             for image_placement in foreground_image_ids {
                 if let Some((image_metadata, image_placement_data)) = image_metadata
                     .get(&image_placement.image_id)
@@ -1847,6 +1847,19 @@ fn render_glyph_svg(
             log::warn!("Icon image should be static");
         }
     }
+}
+
+/// Starts the layer that foreground (non-negative z-index) image placements are painted
+/// into.
+///
+/// The layer sits above the layer the grid was painted into, so without opting out of hit
+/// testing its recorded hit rects would make the cells underneath an image look covered and
+/// every pointer event landing on them would be dropped before it reached the grid. Image
+/// placements are decorative, so the layer is marked click-through and pointer events pass
+/// through to the grid below.
+fn start_foreground_image_layer(scene: &mut Scene) {
+    scene.start_layer(warpui::ClipBounds::ActiveLayer);
+    scene.set_active_layer_click_through();
 }
 
 #[allow(clippy::too_many_arguments)]
--- a/app/src/terminal/grid_renderer_tests.rs
+++ b/app/src/terminal/grid_renderer_tests.rs
@@ -1,9 +1,11 @@
 use pathfinder_geometry::rect::RectF;
 use pathfinder_geometry::vector::{Vector2F, vec2f};
+use warpui::Scene;
+use warpui::elements::Point as ScenePoint;
 use warpui::fonts::Cache as FontCache;
 use warpui::units::{IntoLines, Lines, Pixels};
 
-use super::{CachedBackgroundColor, active_or_next_match};
+use super::{CachedBackgroundColor, active_or_next_match, start_foreground_image_layer};
 use crate::terminal::grid_size_util::calculate_grid_baseline_position;
 use crate::terminal::model::index::Point;
 use crate::terminal::model::selection::SelectionPoint;
@@ -261,4 +263,25 @@ fn test_calculate_selection_bounds() {
     assert_selection_bounds(5.into_lines()); // Without scroll clipping
     assert_selection_bounds(10.into_lines()); // Without scroll clipping (but on the cusp of clipping)
     assert_selection_bounds(80.into_lines()); // With scroll clipping
+}
+
+/// Regression test for a kitty graphics placement swallowing every pointer event that landed
+/// on the cells it covered. Foreground placements are painted into a layer above the grid, so
+/// that layer has to be click-through for the grid underneath to keep receiving mouse events
+/// (wheel included), which is what alt-screen apps rely on to get mouse reports.
+#[test]
+fn test_foreground_image_layer_does_not_cover_the_grid() {
+    let mut scene = Scene::new(1., Default::default());
+
+    // Stand in for the grid painting its cells.
+    scene.draw_rect_with_hit_recording(RectF::new(vec2f(0., 0.), vec2f(200., 200.)));
+    let grid_z_index = scene.z_index();
+
+    start_foreground_image_layer(&mut scene);
+    scene.draw_rect_with_hit_recording(RectF::new(vec2f(30., 30.), vec2f(100., 100.)));
+    scene.stop_layer();
+
+    // A wheel event landing inside the placement is not considered covered, so it is still
+    // dispatched to the grid rather than dropped at the z-index check.
+    assert!(!scene.is_covered(ScenePoint::new(80., 80., grid_z_index)));
 }
--- a/crates/warpui_core/src/scene_tests.rs
+++ b/crates/warpui_core/src/scene_tests.rs
@@ -1,6 +1,7 @@
 use pathfinder_geometry::vector::vec2f;
 
 use super::*;
+use crate::image_cache::test_utils::make_static_image;
 use crate::rendering;
 
 #[test]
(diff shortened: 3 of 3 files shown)


rchougule:
> @rchougule
> 
> Every PR must be linked to a same-repo issue before Oz can review it.
> 
> **Next step:** open or find a same-repo issue describing this change, then link it to this PR by adding `Closes #123` to the PR description (or using the "Development" sidebar on GitHub). A maintainer will mark the issue `ready-to-implement` when it is ready. Once it is marked, comment `/warp-agent-review` to re-trigger review.
> 
> See the [contribution guidelines](https://github.com/warpdotdev/warp/blob/master/CONTRIBUTING.md) for the full readiness model.
> 
> _Powered by [Oz](https://oz.warp.dev)_

It's already done.

rchougule:
> Hi @rchougule — a reviewer requested changes on this PR and it hasn't had activity from you in 7 days. When you get a chance, please push updates or reply to the review so a reviewer can take another look. Without activity, this PR will be automatically closed after 14 days of inactivity.
> 
> 

I have already addressed everything. I am unable to understand what else is required here.

PR #15921, opened 2026-09-10
Queue follow-up injections for warp agent, using new steering mode

seemeroland:
## Description
Queued prompts for a conversation are delivered under one of two modes, stored per conversation on `QueuedQueryModel` (`QueuedPromptDeliveryMode`):
- **Steering**, set automatically for Oz ambient-agent-driven conversations by `bind_native_prompt_conversation`: a queued prompt goes out on the next request made for the conversation — a tool-result follow-up, an orchestration-event injection, or the conversation going idle — instead of waiting for the whole turn to finish. Each prompt is still sent as its own request/exchange; queued prompts are never combined into one exchange.
- **Queueing**, used everywhere else (local `/queue`, the auto-queue toggle, LRC auto-queue): the next queued row is sent only once the conversation finishes on its own.

For a Steering conversation, `route_native_startup_injection` always appends an incoming prompt to the queue, then dispatches the head row via `dispatch_queued_warp_agent_prompt` when `can_dispatch_queued_warp_agent_prompt` confirms nothing is currently streaming and native setup has finished. `dispatch_queued_warp_agent_prompt` is a single controller-level entry point that dispatches whichever queued row it can resolve and send on its own — a local prompt via `send_queued_user_query_in_conversation`, a shared-session-injected prompt via attachment staging/download — regardless of which origin queued it; a shell command, a locked row, or a row being edited is left for `TerminalInput`/`TerminalView::drain_queued_prompts`, which need editor and PTY access the controller doesn't have. It takes an optional specific row id: `None` dispatches the head in FIFO order (every automatic trigger), `Some(id)` targets an explicit row such as "Send now", which may not be the head and is allowed to interrupt an active stream on purpose.

When a turn is already active, `steer_head_prompt_for_request` folds the head queued row into whichever request the conversation naturally sends next (`send_follow_up_for_conversation`, `inject_pending_events_for_request`) as an additional input, so the model sees it as soon as possible without interrupting the turn in flight or waiting for full completion.

Downloading a shared-session prompt's file attachments is best-effort: `download_task_file_attachments` (shared by the live and queued injection paths) logs and skips a failed URL lookup, directory creation, or individual file download rather than aborting the send — the prompt still reaches the conversation, with whatever attachments succeeded.

### Scope and tradeoffs
- A shared-session row carrying a file attachment that still needs downloading can't be resolved synchronously, so it's excluded from the piggyback path and only dispatched via the idle-triggered drain.
- Mode selection is not user-facing yet; Steering is set automatically and cannot currently be toggled.
- Shared-session viewer queue display/synchronization remains out of scope.

### Review guide
Start with `queued_query.rs` for `QueuedPromptDeliveryMode` storage and the per-row eligibility rules (`peek_autofire`) both modes share. Then `controller/startup_queue.rs` for `dispatch_queued_warp_agent_prompt` and `route_native_startup_injection`, and `controller.rs` for `can_dispatch_queued_warp_agent_prompt` and `steer_head_prompt_for_request`'s two call sites. `terminal/view.rs`'s `maybe_dispatch_steering_prompt_now`/`drain_queued_prompts` and `terminal/input.rs`'s `send_queued_row_immediately` cover the local-queueing and "Send now" side. `attachment_utils.rs` has the shared download helper.

[Implementation plan](https://staging.warp.dev/drive/notebook/Oz3uEHLh7CnSvDELpSvEzG) · [Development conversation](https://staging.warp.dev/conversation/c5dcedb1-fc2f-4885-8976-30b779fa7699)

## Linked Issue
No linked issue was provided.
- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. — No linked issue.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes). — No screenshots; the only UI-visible change is treating `SharedSessionInjection` rows like `InitialCloudMode` in the queued-prompts panel, with no new surface to capture.

## Testing
`cargo check -p warp`, `./script/format --check`, and all three presubmit Clippy variants (`cargo clippy --workspace --exclude warp_completer --all-targets --tests`, `cargo clippy -p warp --all-targets --tests`, `cargo clippy -p warp_completer --all-targets --tests`, all with `-D warnings`) passed. The `ai::blocklist`, `ai::agent_sdk`, `terminal::view`, `terminal::input`, and `ai::attachment_utils` nextest suites passed in full (2139 tests). The full workspace test suite was not run.

Not verified: end-to-end delivery against a live Oz shared-session worker.

- [ ] I have manually tested my changes locally with `./script/run` — Not yet verified against a live worker; see above.

## Agent Mode

CHANGELOG-OZ: Fixed native Oz follow-up prompts interrupting one another instead of each reaching the agent.

### DIFF
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -2775,12 +2775,6 @@ impl BlocklistAIController {
         // Remove any locked pending-LRC queries so they don't linger after cancellation.
         QueuedQueryModel::handle(ctx).update(ctx, |model, ctx| {
             model.remove_pending_lrc_rows(conversation_id, ctx);
-            if matches!(
-                reason.conversation_outcome(),
-                CancellationOutcome::Cancelled
-            ) {
-                model.cancel_injection_dispatch(conversation_id, ctx);
-            }
         });
 
         if !self
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -2865,6 +2865,22 @@ impl BlocklistAIController {
             .has_active_stream_for_conversation(conversation_id, app)
     }
 
+    /// True when nothing prevents dispatching a fresh, automatic request for `conversation_id`
+    /// right now: native setup (if any) has finished, and no response stream is currently
+    /// active. Every automatic dispatch trigger -- the post-enqueue idle fast path and the
+    /// FIFO-head case of [`Self::dispatch_queued_warp_agent_prompt`] -- must check this before
+    /// firing, so two dispatch attempts can never race and cancel each other. Does not apply to
+    /// an explicit user override (e.g. "Send now"), which is allowed to interrupt an active
+    /// stream on purpose.
+    pub(crate) fn can_dispatch_queued_warp_agent_prompt(
+        &self,
+        conversation_id: AIConversationId,
+        ctx: &AppContext,
+    ) -> bool {
+        !QueuedQueryModel::as_ref(ctx).is_dispatch_blocked(conversation_id)
+            && !self.has_active_stream_for_conversation(conversation_id, ctx)
+    }
+
     #[cfg(test)]
     pub fn register_mock_stream_for_test(
         &mut self,
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -1612,7 +1612,7 @@ impl BlocklistAIController {
     /// Only pops a plain, unlocked, non-edited prompt row (`AutofireAction::Submit`); a shell
     /// command, a row being edited, or a shared-session row carrying a file attachment that
     /// still needs downloading are left queued for `TerminalView::drain_queued_prompts` (or, for
-    /// a shared-session row, `Self::dispatch_next_shared_session_row`) to handle once the
+    /// a shared-session row, `Self::dispatch_queued_warp_agent_prompt`) to handle once the
     /// conversation goes idle, since none of those can be folded synchronously into an
     /// already-in-flight request.
     fn steer_head_prompt_for_request(
--- a/app/src/ai/blocklist/controller.rs
+++ b/app/src/ai/blocklist/controller.rs
@@ -22,7 +22,7 @@ use input_context::{input_context_for_request, parse_context_attachments};
 use itertools::Itertools;
 use parking_lot::FairMutex;
 use pending_response_streams::PendingResponseStreams;
-use session_sharing_protocol::common::ParticipantId;
+use session_sharing_protocol::common::{AgentAttachment, ParticipantId};
 pub use slash_command::*;
 use warp_core::assertions::safe_assert;
 use warp_errors::report_error;
@@ -41,7 +41,7 @@ use super::orchestration_event_streamer::{
     OrchestrationEventStreamer, OrchestrationEventStreamerEvent,
 };
 use super::orchestration_events::{OrchestrationEventService, OrchestrationEventServiceEvent};
-use super::queued_query::{QueuedQueryId, QueuedQueryModel};
+use super::queued_query::{AutofireAction, QueuedQueryId, QueuedQueryModel};
 use super::{BlocklistAIInputModel, ResponseStreamId};
 use crate::ai::AIRequestUsageModel;
 use crate::ai::agent::api::{self, ServerConversationToken};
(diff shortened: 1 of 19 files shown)


app/src/ai/blocklist/controller/startup_queue.rs
```diff
@@ -0,0 +1,386 @@
...
+        let Some(id) = self.native_prompt_conversation_id else {
+            log::info!(
+                "event=injection_bypassed_queue task_id={:?} terminal_id={:?} participant_id={participant_id} reason=no_native_binding has_target_token={}",
+                self.ambient_agent_task_id,
+                self.terminal_surface_id,
+                token.is_some(),
+            );
+            return false;
+        };
+        if let Some(token) = token
```
harryalbert, line 116: Could this silently route a shared-session follow-up to the wrong conversation when the server token is absent or not mapped locally yet? 

This only rejects tokens that already resolve to a different conversation, then queues the prompt on the bound id without retaining the token. I think we might need to preserve or require the server token, or enforce the invariant that every injection reaching this binding belongs to this conversation, with coverage for the unmapped-token case

app/src/ai/blocklist/controller.rs
```diff
@@ -1591,6 +1601,117 @@ impl BlocklistAIController {
...
 
+    /// If `conversation_id`'s queue is in `Steering` mode and has an eligible head row queued,
+    /// pops that row and returns it built as a fresh `AIAgentInput::UserQuery`, ready to append
+    /// to whichever request the caller is about to send for this conversation (a tool-result
+    /// follow-up, an orchestration-event injection, or a direct send). Also updates the current
+    /// response initiator when the row came from a shared-session participant -- callers must
+    /// invoke this *before* reading [`Self::get_current_response_initiator`] to build their own
+    /// `RequestInput`, so the exchange is attributed correctly.
+    ///
+    /// Only pops a plain, unlocked, non-edited prompt row (`AutofireAction::Submit`); a shell
```
harryalbert, line 1612: This is a higher level thought, but instead of storing these as separate prompts, I wonder if we should just bundle any bundle-able prompts into one prompt instead of queueing separately? might help simplify this logic

app/src/ai/blocklist/queued_query.rs
```diff
@@ -158,13 +190,39 @@ pub enum AutofireAction {
...
+    #[default]
+    Queueing,
+    /// A queued row is sent on the next request made for the conversation -- a natural
+    /// continuation (e.g. a tool-result follow-up or an orchestration-event injection) if one
+    /// occurs first, otherwise the conversation going fully idle -- rather than always waiting
+    /// for the whole turn to finish. Each row is still sent as its own individual
+    /// request/exchange, never combined with another queued row. Set automatically for
+    /// conversations bound to an ambient/Oz-driven native run
+    /// (`BlocklistAIController::bind_native_prompt_conversation`).
+    Steering,
```
harryalbert, line 209: Doesn't need to be in scope fro this ticket, but I know that DStern really wants this as an option generally. Would be good to build this in a way (which I think you already have, but mabye just making clear in comments) such that we could easily like put it behind a setting for those who want this behavior for general queueing as well

app/src/ai/blocklist/queued_query.rs
```diff
@@ -174,6 +232,10 @@ struct ConversationQueueState {
     /// dispatched and cleared when it finishes; keeps the queue accepting new rows while the
     /// agent is idle and gates the next drain until the command completes.
     command_in_flight: bool,
+    /// True while a shared-session row's file-attachment download is in flight, after the row
+    /// has already been removed from `queue` but before the resulting request has actually been
+    /// sent. See [`QueuedQueryModel::arm_download_in_flight`].
+    download_in_flight: bool,
```
harryalbert, line 238: could we just keep the item in the queue until it's done downloading (in some like resolving state, where we auto-send on resolution) instead of maintaining this separate flag? That would let us keep the queue as the authoritative bookkeeper on like what prompts are queued, how prompts should be inserted, etc.

app/src/ai/blocklist/controller/startup_queue.rs
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+
+    /// Resolves `row`'s attachments and sends it into `conversation_id` via the normal
+    /// existing-conversation follow-up path, which cancels any turn already in flight before
+    /// sending. Block and plain-text attachments are staged onto the live context model (the
+    /// same way `execute_warp_agent_prompt_from_shared_session_injection` stages them for a
+    /// live, non-startup injection) so the standard send path picks them up automatically; file
+    /// attachments are downloaded asynchronously first.
+    fn send_native_startup_injection(
```
harryalbert, line 244: Could this use the same attachment-and-send pipeline as live shared-session injections? I think a lot of this could be centralized into a common helper.

app/src/ai/blocklist/controller/startup_queue.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// and friends) must not fall through to legacy conversation resolution for this prompt:
+    /// covers every case where this controller is bound to a native conversation, whether the
+    /// prompt was queued, dispatched immediately, or rejected outright for naming a different
+    /// conversation's token. Returns `false` only when this controller isn't bound to a native
+    /// conversation at all, in which case the caller's own token/selected-conversation
+    /// resolution is the correct behavior. Bypassing to that legacy resolution once bound would
+    /// be wrong: it has no way to find this binding once the conversation already has exchanges
+    /// but hasn't been assigned a server token yet, which is routinely true for every follow-up
+    /// after the first.
+    pub(super) fn route_native_startup_injection(
```
harryalbert, line 99: Could we potentially resolve the target conversation just once before queueing? A resolver returning the existing conversation, native bootstrap conversation, or rejection could make a supplied server token authoritative, and then, both paths could use the ordinary queue keyed by the resolved local ID. 

I _think_ this would allow us to get rid of some of the complexity around routing in `startup_queue.rs` and `shared_session.rs`

harryalbert (approved):
Overall this approach seems right to me. Had some thoughts on how we might simplify, but approving to unblock!

app/src/ai/blocklist/queued_query.rs
seemeroland: Yea this was motivated by talking to @vorporeal. This is built in a way that we should be able to have the user toggle this setting

app/src/ai/blocklist/controller.rs
seemeroland: Considered this and discussed with @vorporeal. Ideally we could send them all in one request, with multiple user query inputs in the MAA request. Unfortunately there are lots of places across client and server that assume one user query per request. 

Had also considered combining into one query separated by newlines but that seems less clear, and technically the session sharing protocol allows the queries to be sent by different people but for rendering we'd have to show one profile pic

app/src/ai/blocklist/controller/startup_queue.rs
seemeroland: Good call!

app/src/ai/blocklist/queued_query.rs
seemeroland: I think this is probably a good idea but needs some larger changes so going to punt on this!

app/src/ai/blocklist/controller/startup_queue.rs
seemeroland: Extracting the conversation resolving seems like a good idea

PR #15876, opened 2026-09-08
[APP-5720] Round-trip request metadata in the client

Xavientois:
## Description

Adds client-side plumbing for server-authored `Message.request_metadata` records from [warp-proto-apis #377](https://github.com/warpdotdev/warp-proto-apis/pull/377):

* Bumps the proto dep with the new `RequestMetadata` message
* Update the local model to add the new fields from this message
* Add tests to ensure the message survives the round-trip and gets persisted correctly (also one to check the summing behaviour)
* I also took this opportunity to clean up some of the tests where we could be using the `Default` trait to initialize our test data and save a bunch of LOC

This ensures that the request metadata gets correctly persisted back to the server in the task (convo) message list. This also sets us up to display this metadata in https://github.com/warpdotdev/warp/pull/15931.

## Linked Issue

[APP-5720](https://linear.app/warpdotdev/issue/APP-5720/persist-per-request-usagecosttiming-data-in-ai-conversation-request)

- [x] Screenshots or video are not applicable because this PR has no user-visible or UI changes.

## Agent Mode

- [x] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

Co-Authored-By: Warp <agent@warp.dev>

PR #15931, opened 2026-09-11
Turn panel UI for per-request usage metadata (APP-5720)

Xavientois:
## Description
Adds the per-turn "Turn" usage panel, driven by the `RequestMetadata` records the client now persists (#15876). 

Behind the `PricingTransparency` feature flag

**Trigger**
- Pie icon replaces the usage pill on the latest turn's response footer.
- Tooltip shows the turn's charge in the user's Credits/Dollars setting; sub-cent amounts show as `<$0.01`.
- `PricingTransparency` flag off → the old usage pill renders exactly as before.

https://www.loom.com/share/10a64478032e48b395ad55f03d092f3c

**Legacy conversations (no / incomplete records)**
- Panel still opens and shows only what the client can honestly derive.
- Timing from exchange timestamps; charges from the last-block snapshot when it's the latest turn, otherwise hidden (unknown ≠ zero).
- Tool counts are omitted (only conversation-cumulative data exists).

## Linked Issue
APP-5720

## Testing
- [x] I have manually tested my changes locally with `./script/run` (see video above)
- Seeded the local SQLite store with 21 mock conversations covering every data shape (records, partial records, three legacy shapes, tool round-trips, units toggle). Seeder + test plan kept out of the tree.
- `cargo clippy -p warp --all-targets --tests -- -D warnings` clean; `cargo nextest run -p warp request_metadata turn_view turn_panel convert_conversation` 53/53.

### Feature flag works

https://www.loom.com/share/98f46ef9851b4c2d9c8bc14738a9c079

## Agent Mode


app/src/ai/agent/conversation.rs
```diff
@@ -877,6 +881,14 @@ impl AIConversation {
             .charged_usage_for_last_block = charged_usage;
     }
 
+    /// Test-only helper that sets the last block's credits total directly, without wiring
+    /// a full `StreamFinished` event.
```
Xavientois:
```suggestion
    /// Test-only helper to set credits directly without generating a full `StreamFinished` event.
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
         self.task_store.get(task_id)
     }
 
+    /// The request ids of the messages added to `exchange_id`, resolved across the whole task
+    /// store: summarization (`Action::MoveMessagesToNewTask`) moves an exchange's messages into
+    /// a subtask without touching the exchange's `added_message_ids`, so message membership
+    /// cannot be read off the single task the exchange happens to live in.
```
Xavientois:
```suggestion
    /// We have to check all tasks for the messages because summarization moves them to a new task (`Action::MoveMessagesToNewTask`)
```

app/src/terminal/view/rich_content.rs
```diff
@@ -230,6 +234,9 @@ pub enum RichContentMetadata {
         exchange_id: AIAgentExchangeId,
     },
     UsageFooter,
+    /// A docked, closeable per-turn "Turn" panel backed by the turn's persisted request-metadata
+    /// record. See `RequestMetadataTurnView`.
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+                    .contains(&MessageId::new(message.id.clone()))
+            })
+            .map(|message| message.request_id.clone())
+            .filter(|request_id| !request_id.is_empty())
+            .collect()
+    }
+
+    /// Every per-request record for the requests that produced `exchange_id`, in task order.
+    /// An exchange can hold several records — one per underlying API request (retries, resumes,
+    /// multi-request turns) — so the Turn panel groups these rather than showing one per request.
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+            .collect()
+    }
+
+    /// The root-task exchanges that make up the user-visible turn containing `exchange_id`:
+    /// from the exchange carrying the user's query through every follow-up exchange (tool-result
+    /// round trips are separate requests, and so separate exchanges) up to, but not including,
+    /// the next exchange that carries a user query. An exchange outside the root task is its own
+    /// turn.
+    pub fn turn_exchange_ids(&self, exchange_id: AIAgentExchangeId) -> Vec<AIAgentExchangeId> {
+        let root_exchanges: Vec<&AIAgentExchange> = self.root_task_exchanges().collect();
```
Xavientois: Why are we collecting root_task_exchanges here?

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+            .filter_map(RequestMetadataRecord::from_message)
+            .filter(|record| seen_message_ids.insert(record.message_id.clone()))
+            .collect()
+    }
+
+    /// The root-task exchanges that make up the user-visible turn containing `exchange_id`:
+    /// from the exchange carrying the user's query through every follow-up exchange (tool-result
+    /// round trips are separate requests, and so separate exchanges) up to, but not including,
+    /// the next exchange that carries a user query. An exchange outside the root task is its own
+    /// turn.
```
Xavientois:
```suggestion
    /// Given the provided exchange id, get all exchange ids for the turn containing that exchange
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+            .find(|&i| root_exchanges[i].has_user_query())
+            .unwrap_or(root_exchanges.len());
+        root_exchanges[start..end]
+            .iter()
+            .map(|exchange| exchange.id)
+            .collect()
+    }
+
+    /// Whether `exchange_id` is the last exchange of its turn (see [`Self::turn_exchange_ids`]),
+    /// i.e. the block where turn-level controls belong.
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+    pub fn is_last_exchange_in_turn(&self, exchange_id: AIAgentExchangeId) -> bool {
+        self.turn_exchange_ids(exchange_id).last() == Some(&exchange_id)
+    }
+
+    /// Every per-request record for the turn containing `exchange_id`, in task order: the union
+    /// of [`Self::request_metadata_records_for_exchange`] over [`Self::turn_exchange_ids`].
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+            .flat_map(|id| self.request_metadata_records_for_exchange(id))
+            .filter(|record| seen_message_ids.insert(record.message_id.clone()))
+            .collect()
+    }
+
+    /// The Turn panel contents for the turn containing `exchange_id`: the server-authored
+    /// records when the turn's metadata is complete, otherwise the best client-derived
+    /// summary — timing from the exchanges themselves, plus (for the latest turn only) the
+    /// last-block charge snapshots and the conversation's current context-window reading.
+    /// Cumulative conversation totals are never presented as a historical turn's charges.
```
Xavientois:
```suggestion
    /// Because this panel relies on `RequestMetadata` to populate its fields, and that message
    /// does not exist for historical conversations, we fallback to `conversation_metadata` to display what we have
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+                exchange
+                    .time_to_first_token_ms
+                    .map(|ms| exchange.start_time + chrono::Duration::milliseconds(ms))
+            })
+            .min();
+
+        // The last-block snapshots describe only the most recent turn; for any earlier
+        // turn they would silently attribute that turn's charges to this one. Anchor on the
+        // root task's actual last exchange — no visibility filtering — so a hidden latest
+        // turn keeps its own charges instead of lending them to the previous visible turn.
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+        let is_latest_turn = self
+            .root_task_exchanges()
+            .last()
+            .is_some_and(|latest| turn_exchange_ids.last() == Some(&latest.id));
+        let context_window_usage = if is_latest_turn {
+            let usage = self.context_window_usage();
+            (usage > 0.0).then_some(usage * 100.0)
+        } else {
+            None
+        };
```
Xavientois: RequestMetadata's ContextWindow field should hold this for every request in every turn. We should be able to extract the context_window_usage from the request metadata if it is there, no?

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+                Some(totals) => LegacyCharges::Breakdown(Box::new(totals)),
+                None => match self.credits_spent_for_last_block() {
+                    Some(credits) => LegacyCharges::CreditsOnly(credits),
+                    None => LegacyCharges::Unknown,
+                },
+            }
+        };
+
+        // A Breakdown renders as one aggregated row: the flat totals deliberately discard
+        // model attribution.
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+            LegacyCharges::CreditsOnly(_) | LegacyCharges::Unknown => (Vec::new(), Vec::new()),
+        };
+
+        Some(TurnPanelData::Legacy {
+            record: Box::new(RequestMetadataRecord {
+                message_id: String::new(),
+                request_id: String::new(),
+                recorded_at: None,
+                // No record delivered for this turn: the outcome is unknown, and the
+                // legacy panel does not render outcome anywhere.
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+                outcome: RequestOutcome::Unspecified { incomplete: false },
+                request_started_at: started,
+                first_token_at: first_token,
+                request_ended_at: ended,
+                llm_generation_spans: Vec::new(),
+                model_charges,
+                platform_charges,
+                // Tool totals are conversation-cumulative; a per-turn count is unknown.
```
Xavientois:
```suggestion
```

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3529,261 @@ impl AIConversation {
...
+    /// panel opener: the turn's per-request records when the turn closes at `exchange_id` (see
+    /// [`Self::is_last_exchange_in_turn`]) and its metadata is complete — every exchange in the
+    /// turn has request ids, and all of them have delivered their `Message.RequestMetadata`
+    /// record to this client.
+    ///
+    /// Request membership is resolved from the exchanges' messages, not from the records, and
+    /// every exchange must contribute some: a request cancelled or failed before the server's
+    /// input echo lands (or disconnected mid-stream) leaves its exchange without messages or
+    /// record, and the non-empty record set of its earlier requests must not be presented as
+    /// the turn's total.
```
Xavientois:
```suggestion
    /// Get all request metadata for the turn that ENDS with the provided exchange_id
    /// If the exchange_id is not the last in its turn, this will return `None`, indicating that the
    /// turn details panel should not be displayed for that exchange
```

app/src/ai/agent/conversation.rs
Xavientois: Only so it could be indexed forwards/backwards from the target. Rewrote it in 0b363dc18 as a single pass over `root_task_exchanges()` that resets the running turn at each user query and stops at the first user query after the target, so no collect is needed.

app/src/ai/agent/conversation.rs
Xavientois: Yes. This branch is the fallback when `turn_panel_records` returns `None`, but that can still mean partial records (e.g. one request in the turn never delivered its record). 0b363dc18 now takes the latest `context_window_usage` from whatever records the turn does have and only falls back to the conversation-level reading (latest turn only) when none report one.

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3527,247 @@ impl AIConversation {
...
+            return None;
+        }
+        if let Some(records) = self.turn_panel_records(exchange_id) {
+            return Some(TurnPanelData::Records(records));
+        }
+
+        let turn_exchange_ids = self.turn_exchange_ids(exchange_id);
+        // One timing-only record per exchange, so the panel can tell the agent's own
+        // processing time (the sum of exchange durations) apart from the wall-clock span that
+        // includes tool execution between exchanges.
```
Xavientois:
```suggestion
```

app/src/ai/agent/api/convert_conversation.rs
```diff
@@ -1930,15 +1930,18 @@ fn create_exchange_from_messages(
...
+                .iter()
+                .filter_map(|message_id| {
+                    message_map.get(message_id.as_str()).and_then(|message| {
+                        message.timestamp.as_ref().map(|timestamp| {
+                            proto_timestamp_to_local_datetime(timestamp.seconds, timestamp.nanos)
+                        })
                     })
                 })
-            })
+                .min()
```
Xavientois, line 1944: Why are we changing this?

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3527,247 @@ impl AIConversation {
...
+                request_ended_at: exchange.finish_time,
+                llm_generation_spans: Vec::new(),
+                model_charges: Vec::new(),
+                platform_charges: Vec::new(),
+                tool_calls: None,
+                commands_executed: None,
+                files_changed: None,
+                lines_added: None,
+                lines_removed: None,
+                context_window_usage: None,
```
Xavientois: Why are these all None?

app/src/ai/agent/conversation.rs
```diff
@@ -3517,6 +3527,247 @@ impl AIConversation {
...
+            });
+        }
```
Xavientois: Do we even need this code where we push an empty request metadata record? The `last_mut` function should just return `None` and be handled correctly afterwards, so this is just unnecessary.

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
+//! The client's read-only view of the server-authored `Message.RequestMetadata` records that
+//! ride in the task tree (APP-5720). The client never authors or mutates these; it decodes them
```
Xavientois:
```suggestion
//! ride in the task tree. The client never authors or mutates these; it decodes them
```

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
...
+use super::api::convert_conversation::proto_timestamp_to_local_datetime;
+use crate::persistence::model::ChargedUsageTotals;
+
+/// Token counts and costs (US cents and credits) charged for one model within one usage
+/// category.
+#[derive(Debug, Clone, PartialEq)]
+pub struct RequestModelCharge {
+    pub category: String,
+    /// `direct_api`, `byok`, or `custom_endpoint`.
+    pub usage_type: &'static str,
```
Xavientois: Should we use an enum for this?

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// Wall-clock spans of each internal LLM generation call, in order.
+    pub llm_generation_spans: Vec<RequestLlmGenerationSpan>,
+    pub model_charges: Vec<RequestModelCharge>,
+    pub platform_charges: Vec<RequestPlatformCharge>,
+    pub tool_calls: Option<u32>,
+    pub commands_executed: Option<u32>,
+    pub files_changed: Option<u32>,
+    pub lines_added: Option<u32>,
+    pub lines_removed: Option<u32>,
+    /// Percentage (0-100) of the context window in use after the request.
```
Xavientois:
```suggestion
    /// Fraction 0-1 of the context window in use after the request.
```

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
...
+            return None;
+        };
+
+        let timing = metadata.timing.as_ref();
+        let timestamp = |ts: Option<&prost_types::Timestamp>| {
+            ts.map(|ts| proto_timestamp_to_local_datetime(ts.seconds, ts.nanos))
+        };
+        let seconds = |duration: Option<&prost_types::Duration>| {
+            duration.map(|d| d.seconds as f64 + f64::from(d.nanos) / 1e9)
+        };
```
Xavientois: Why are we implementing this closure ourselves? This should be existing functionality somewhere else, no? If it isn't, could we not just do it inline?

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
...
+        let mut model_charges = Vec::new();
+        let mut platform_charges = Vec::new();
+        if let Some(charges) = metadata.charges.as_ref() {
+            let mut categories: Vec<_> = charges.usage_by_category.iter().collect();
+            categories.sort_by(|(a, _), (b, _)| a.cmp(b));
+            for (category, charged) in categories {
+                let sections: [(&str, &_); 3] = [
+                    ("direct_api", &charged.direct_api_inference_usage),
+                    ("byok", &charged.byok_inference_usage),
+                    ("custom_endpoint", &charged.custom_endpoint_inference_usage),
```
Xavientois: Instead of hardcoding the string literals here, could we have these defined elsewhere?

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
...
+                total += ms;
+            }
+        }
+        any.then_some(total)
+    }
+}
+
+/// What the Turn panel renders for one turn. Complete server-authored records are preferred;
+/// turns without them (conversations that predate the records, or requests that never delivered
+/// one) fall back to the best client-derived summary of that turn.
```
Xavientois:
```suggestion
```

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum TurnPanelData {
+    Records(Vec<RequestMetadataRecord>),
+    /// Synthetic records built client-side: one per exchange carrying that exchange's timing,
+    /// with the turn's context-window reading (latest turn only) and `charges` on the last.
```
Xavientois:
```suggestion
    /// Synthetic records built using the legacy conversation metadata
```

app/src/ai/blocklist/block.rs
```diff
@@ -6156,6 +6176,15 @@ pub enum AIBlockEvent {
         is_expanded: bool,
     },
 
+    /// Emitted when we want to show or hide the per-turn request-metadata "Turn" panel.
+    TurnPanelToggled {
+        conversation_id: AIConversationId,
+        /// The exchange this block renders, used to look up that turn's record rather than
+        /// whatever the conversation's latest turn happens to be.
```
Xavientois:
```suggestion
        /// The exchange this block renders, used to look up that turn's record
```

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3480,6 +3476,33 @@ fn render_response_footer(props: Props, app: &AppContext) -> Option<Box<dyn Elem
         background: Some(blended_colors::neutral_4(appearance.theme()).into()),
         ..style_override
     };
+    (style_override, style_override_with_background)
+}
+
+/// The Turn panel contents for the user-visible turn this block closes (APP-5720): the
+/// server-authored records when every request in the turn delivered one, otherwise the best
+/// client-derived summary. `None` for a mid-turn tool-call block, with pricing transparency
+/// off, or without a bound exchange/conversation.
```
Xavientois:
```suggestion
/// The Turn panel contents for the user-visible turn this block closes
```

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3638,7 +3661,27 @@ fn render_response_footer(props: Props, app: &AppContext) -> Option<Box<dyn Elem
...
     }
 
-    flex.add_child(render_usage_button(props, app));
+    // The Turn panel is the single per-turn usage surface (APP-5720): the trigger replaces
+    // the legacy credit-count pill on every latest-turn block, whether or not the turn
+    // delivered server-authored records (legacy turns show what the client can derive).
+    if let Some(data) = turn_panel_data_for_block(props, app) {
+        // The full-bleed pie glyph reads tighter against the fork icon than the other
+        // footer glyphs do; the explicit margin gives it the same visual spacing as
+        // the surrounding icon buttons.
```
Xavientois:
```suggestion
```

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3638,7 +3661,27 @@ fn render_response_footer(props: Props, app: &AppContext) -> Option<Box<dyn Elem
         flex.add_child(fork_button);
     }
 
-    flex.add_child(render_usage_button(props, app));
+    // The Turn panel is the single per-turn usage surface (APP-5720): the trigger replaces
+    // the legacy credit-count pill on every latest-turn block, whether or not the turn
+    // delivered server-authored records (legacy turns show what the client can derive).
```
Xavientois:
```suggestion
```

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3670,7 +3713,43 @@ fn render_response_footer(props: Props, app: &AppContext) -> Option<Box<dyn Elem
     Some(flex.finish().with_content_item_spacing().finish())
 }
 
+/// Renders the per-turn icon that, on click, opens/closes the docked "Turn" panel backed by the
+/// turn's usage data. Same hover-tooltip/click-to-open pattern as the old usage button, but an
+/// independent trigger with no cross-navigation to the usage footer.
```
Xavientois:
```suggestion
/// turn's usage data.
```

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3670,7 +3713,43 @@ fn render_response_footer(props: Props, app: &AppContext) -> Option<Box<dyn Elem
...
+    .build()
+    .on_click(|ctx, _, _| {
+        ctx.dispatch_typed_action(AIBlockAction::ToggleIsTurnPanelExpanded);
+    })
+    .finish()
+}
+
 /// Renders the usage button that, on click, will expand & collapse the usage summary footer.
+/// Only reachable with `PricingTransparency` off (the Turn panel replaces it when the flag is
+/// on); the rip-out waits until the flag ships everywhere.
```
Xavientois:
```suggestion
```

app/src/ai/blocklist/block/view_impl/output.rs
```diff
@@ -3638,7 +3661,27 @@ fn render_response_footer(props: Props, app: &AppContext) -> Option<Box<dyn Elem
         flex.add_child(fork_button);
     }
 
-    flex.add_child(render_usage_button(props, app));
+    // The Turn panel is the single per-turn usage surface (APP-5720): the trigger replaces
+    // the legacy credit-count pill on every latest-turn block, whether or not the turn
+    // delivered server-authored records (legacy turns show what the client can derive).
+    if let Some(data) = turn_panel_data_for_block(props, app) {
```
Xavientois: Let's actually put the pricing transparency feature flag check here at this level instead of inside of the turn panel data for block function.

app/src/ai/agent/request_metadata.rs
```diff
@@ -0,0 +1,505 @@
...
+                .num_milliseconds()
+                .max(0),
+        )
+    }
+}
+
+/// Aggregates one turn's records for display. Charges are summed per model and per platform
+/// category; timing uses the widest span (earliest start / first token, latest end) plus every
+/// record's LLM generation spans; tool counts sum; the context window comes from the latest
+/// record that reports one.
```
Xavientois:
```suggestion
```

app/src/terminal/view.rs
```diff
@@ -7260,6 +7283,100 @@ impl TerminalView {
...
+        }
+
+        if !is_expanded {
+            ctx.notify();
+            return;
+        }
+
+        // The panel is a pricing-transparency surface. If the flag turned off while the
+        // trigger was on screen, the block still dispatches the toggle: close-only, no
+        // new panel (the trigger itself disappears on the block's next render).
```
Xavientois:
```suggestion
```

app/src/ai/blocklist/usage/request_metadata_turn_view.rs
```diff
@@ -0,0 +1,794 @@
...
+                }
+                ctx.notify();
+            }
+        }
+    }
+}
+
+/// The trigger icon's hover tooltip for whatever the panel will show: the turn's charge,
+/// honoring the user's credits/dollars display-unit setting. Stays quiet ("Turn") rather than
+/// fabricating a total when nothing is known.
```
Xavientois:
```suggestion
```

app/src/ai/blocklist/usage/request_metadata_turn_view.rs
```diff
@@ -0,0 +1,794 @@
...
+    } else {
+        cost_in_cents
+    };
+    let dollars = cost_in_cents / 100.0;
+    if cost_in_cents > 0.0 && dollars < 0.01 {
+        "<$0.01".to_string()
+    } else {
+        format!("${dollars:.2}")
+    }
+}
```
Xavientois, line 802: When formatting token and credit amounts, let's add commas every three digits for very large numbers, like the typical American formatting for large numbers.

app/src/ai/agent/api/convert_conversation.rs
Xavientois: The fallback iterates a `HashSet<String>` of message ids and `find_map` took whichever message came first in hash order, so the restored exchange's `start_time` was nondeterministic — it could pick the agent output's timestamp instead of the query's, which made restored exchanges report 0s durations in the legacy Turn panel. `.min()` makes it the earliest message timestamp, which is what the fallback was meant to be.

app/src/ai/agent/conversation.rs
Xavientois: These synthetic records only exist to carry each exchange's timing; the exchange has no per-request charges/tool counts of its own (tool totals in the conversation metadata are cumulative, not per turn), and the turn-level charges + context reading are attached to the last record right below. Switched to `..Default::default()` in 8b195415c so only the timing fields are spelled out.

app/src/ai/agent/conversation.rs
Xavientois: Agreed — removed in 8b195415c. `turn_exchange_ids` always includes the exchange itself, so the list is never empty in practice, and `last_mut()` returning `None` is handled.

app/src/ai/agent/request_metadata.rs
Xavientois: Done in 8b195415c: `InferenceUsageType { DirectApi, Byok, CustomEndpoint }`, which also removes the string literals below.

app/src/ai/agent/request_metadata.rs
Xavientois: Good catch — the pinned proto rev still says percentage, but main documents it as a 0-1 fraction and the server fills it from the same `ContextWindowUsage` value that feeds `ConversationUsageMetadata`, so the client was mis-scaling it. 8b195415c treats it as a fraction end to end (doc, ring rendering, legacy path, tests, seeder).

app/src/ai/agent/request_metadata.rs
Xavientois: There is no `&Timestamp`-taking helper (only `proto_timestamp_to_local_datetime(seconds, nanos)`), so the closures are gone and the conversions are inline in 8b195415c. The duration one now uses `std::time::Duration::try_from(prost_types::Duration)`.

app/src/ai/agent/request_metadata.rs
Xavientois: Replaced by the `InferenceUsageType` enum in 8b195415c.

app/src/ai/blocklist/block/view_impl/output.rs
Xavientois: Moved in 8b195415c: the footer gates on `PricingTransparency` and `turn_panel_data_for_block` no longer checks the flag.

app/src/ai/blocklist/usage/request_metadata_turn_view.rs
Xavientois: Done in 8b195415c: token counts and credit amounts use thousands separators (`1,234,567 tokens`, `12,345.5 credits`).

jefflloyd (approved):
Assuming this works with both the FF on and off, LGTM!

PR #9687, opened 2026-04-30
Fix pane search for custom tab titles

kill74:
## Description

Fixes sidebar search in vertical tabs Panes mode so renamed tabs can be found by their custom title.

Previously, the Panes-mode filtering path dropped the tab group's `custom_title` when constructing the temporary `PaneProps` used for search matching. This meant searching for a visible custom tab title, like `deploy`, returned no results even though that title was shown as the group header.

This change includes `pane_group.custom_title(app)` in the two search/filter paths while preserving the existing render-side behavior that prevents the custom tab title from being duplicated on every pane row.

## Linked Issue

Fixes #9666

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`.
- [x] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes).

## Screenshots / Videos

N/A — this is a search/filtering behavior fix. I manually verified the visible behavior in the vertical tabs sidebar.

## Testing

Manual test performed:

1. Enabled vertical tabs.
2. Switched vertical tabs to Panes mode.
3. Renamed a tab to a unique custom title.
4. Searched for that custom title in the vertical tabs search box.
5. Confirmed the tab group remains visible.
6. Confirmed the custom title is not duplicated on each pane row.
7. Confirmed normal pane search still works.

Also ran:

- `cargo fmt`
- `cargo nextest run`

`cargo nextest run` did not complete successfully due to unrelated GCP cloud-provider tests failing on Windows:

- `ai::agent_sdk::driver::cloud_provider::tests::collect_provider_env_vars_merges_all_providers`
- `ai::agent_sdk::driver::cloud_provider::tests::extract_cloud_providers_creates_gcp_provider`
- `ai::agent_sdk::driver::cloud_provider::tests::gcp_provider_env_vars`

The failure appears unrelated to this vertical-tabs change. The error was:

`Cannot use executable ...\target\debug\deps\warp-...exe for GCP executable-sourced credentials`

No new automated test was added because this change affects UI search/filtering behavior in the vertical tabs sidebar and was verified manually against the reproduction steps from the issue.

## Agent Mode

- [ ] Warp Agent Mode - This PR was created via Warp's AI Agent Mode

CHANGELOG-BUG-FIX: Fixed vertical tabs Panes-mode search so renamed tabs can be found by their custom title.

PR #15935, opened 2026-09-11
fix(vertical-tabs): match custom tab titles in Panes-mode sidebar search

### LINKED ISSUE #9666: Sidebar Panes-mode search doesn't match custom tab names shown in the group header

CuriousTank:
## Description

In Panes display mode, the vertical-tabs sidebar search field (**"Search tabs…"**) never matched a tab's custom (renamed) title: searching `deploy` for a tab renamed to `deploy` returned **no results**, even though `deploy` was rendered on screen as the group header above the pane rows. Pane names and other per-pane text (working directory, last command, agent names) kept matching, so the bug only surfaced for renamed tabs (#9666).

**Root cause.** Both filter sites — `render_groups` (the visible filtered list) and `matching_tab_indices` (tab cycling under an active search) — build search-only `PaneProps` whose `display_title_override` is gated behind `!uses_outer_group_container(...)`:

```rust
let title_override = (!uses_outer_group_container(display_granularity))
    .then(|| pane_group.custom_title(app))
    .flatten();
```

That gate is correct on the **render** side: in Panes mode the tab title lives in the group header and must not be repeated on every pane row. But `rendered_search_text_fragments` builds the searchable text from the same fields the rows display, so the gate also silently dropped the tab title from **matching** — the shared-field design made the search-side drop an accidental consequence of the render rule.

**Fix.** Keep the render-side gate untouched and match the custom tab title *separately* from the per-pane fragments, via a new pure helper used by both filter sites:

```rust
fn custom_tab_title_matches_query(
    custom_title: Option<&str>,
    uses_outer_group_container: bool,
    query_lower: &str,
) -> bool
```

A tab-title match admits the tab with **all** of its display panes (the header labels them all); pane rows keep matching by their own text, OR'd with the title match. Tabs-granularity modes (FocusedSession) are unchanged — they already search the title through `display_title_override`, where it *is* the row's displayed title, so the helper is inert there (`uses_outer_group_container == false`).

**Why not just drop the gate at the filter sites.** Routing the tab title through `display_title_override` makes it *replace* the generated pane fragments (`terminal_pane_search_text_fragments` uses the override as `primary_text` instead of the generated title), so a renamed tab would stop matching its visible pane titles — the flaw flagged in Oz's review of #9687. Matching the header text separately adds the title to the match set without replacing anything.

## Linked Issue

Fixes #9666

## Testing

Automated:
- `cargo fmt --all --check` — pass
- `cargo clippy -p warp --tests -- -D warnings` — pass (presubmit's clippy variant, scoped to the changed package)
- `cargo test -p warp --lib -- vertical_tabs custom_tab_title` — **87 passed, 0 failed**, including:
  - `panes_granularity_search_matches_custom_tab_title_and_keeps_pane_matches` — end-to-end regression through the real `matching_tab_indices` filter: a tab with custom title `deploy` and a pane with custom name `exp` matches on **both** `deploy` (the #9666 fix) and `exp` (pane matching must not regress), and a non-matching query still filters everything out
  - `custom_tab_title_matches_query_*` — three pure tests for the helper (header-text matching incl. case-insensitivity, inert in Tabs granularity, no match on None/empty/mismatched titles)
  - all 82 pre-existing vertical-tabs tests — no regressions

Full `./script/presubmit` (workspace-wide nextest) was not run in this environment; the change is scoped to `app/src/workspace/view/vertical_tabs.rs` and its test file.

I could not run the GUI locally in this environment (headless Linux build machine). The behavior change is covered by the end-to-end test above, which drives the actual filter used by the rendered list and tab cycling. Happy to attach screenshots if a maintainer wants visual confirmation, or to walk through manual verification steps.

### Screenshots / Videos

Not applicable — this is a search/filter behavior fix with no visual change to the panel layout.

## Agent Mode

CHANGELOG-BUG-FIX: The vertical tabs sidebar search now finds renamed tabs by their custom tab name in Panes display mode.

### DIFF
--- a/app/src/workspace/view/vertical_tabs.rs
+++ b/app/src/workspace/view/vertical_tabs.rs
@@ -1166,9 +1166,10 @@ impl VerticalTabsPanelState {
         });
     }
 
-    /// Returns the indices (in original order) of tab groups that have at least
-    /// one pane matching the current search query. Returns all indices when the
-    /// query is empty.
+    /// Returns the indices (in original order) of tab groups that match the
+    /// current search query — either through a pane's own text or, in Panes
+    /// display mode, through the tab's custom title rendered as the group
+    /// header. Returns all indices when the query is empty.
     pub(super) fn matching_tab_indices(
         &self,
         tabs: &[TabData],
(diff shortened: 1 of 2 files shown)


CuriousTank:
### Manual testing (before / after)

Verified locally on Linux (Ubuntu 20.04, X11), built from source with `cargo run --bin warp-oss --features gui` (the `./script/run` path on Linux). Vertical tabs sidebar in **Panes** display granularity.

**Test setup** — several tabs, one of them renamed via the rename-tab flow (the custom tab name is visible as the group header above its pane rows):

setup: tabs created, renamed tab visible as group header

**Before (unfixed `master` @ `0eefb74`)** — searching the renamed tab's visible name returns **no results**, even though the name is displayed right above the pane rows (the #9666 bug):

before: searching the custom tab name returns no results

**After (this PR @ `5e2af40`)** — the same search now matches the renamed tab and keeps its pane rows visible:

after: searching the custom tab name finds the tab

Pane-name/pane-text search was also re-verified after the fix and still matches as before; a non-matching query still shows "No tabs match your search."

Issue #15968, opened 2026-09-12
Cut the current line when no text is selected in the code editor


AndrewDongminYoo:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3Aenhancement) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

In the code editor, pressing the cut shortcut with no text selected should cut the current line, the way VS Code and the JetBrains IDEs do.

The cut text should then behave like a line-wise copy when it is pasted: it goes back as a whole line above the cursor's line rather than at the caret.

This is the cut counterpart of #15849, which covers copy.

### Is your feature request related to a problem? Please describe.

Yes. A cut with an empty selection in the code editor currently does something a user is unlikely to want. `CodeEditorModel::cut` is `copy` followed by `backspace`, so with nothing selected it writes empty content to the clipboard and deletes the single character before the caret. Whatever the clipboard held is gone, and a character disappears from the buffer.

Line-wise cut is a near-universal expectation from other editors, so the present behavior reads as a gap rather than a deliberate choice.

### Additional context

Once the copy behavior requested in #15849 lands, copy will be line-wise while cut is not, which makes the asymmetry easier to run into.

The mechanism needed for this already exists in the repository. The copy side can record a line-wise entry through `VimRegisters::write_to_register` on the `+` register, and the paste side reads it back with `read_from_register`, which compares its stored entry against the live system clipboard and falls back to a char-wise paste when they differ. A line-wise cut would reuse those same two calls and add the line deletion.

One point worth settling during triage: whether this should apply only to the standalone code editor, as the copy change does, or to more surfaces. `CodeEditorView` is embedded in AI blocks, diff views, the requested-command view and the CLI subagent block, and those surfaces rely on the editor emitting `CodeEditorEvent::CopiedEmptyText` so the parent view can act on an empty-selection copy instead.

### Operating system (OS)

Cross-platform

### How important is this feature to you?

2

### Warp Internal (ignore) - linear-label:39cc6478-1249-4ee7-950b-c428edfeecd1

None

AndrewDongminYoo:
I'll take this one. It builds directly on the line-wise copy and paste mechanism in #15969, so I'll open the PR once that lands and rebase onto master, rather than stacking a PR that repeats that diff.

PR #15969, opened 2026-09-12
Copy the current line when nothing is selected, and paste it back as a line

### LINKED ISSUE #15849: Copy line when no text is selected in code editor
### Pre-submit Checks

At least in the code editor, pressing <kbd>Ctrl</kbd>/<kbd>Cmd</kbd> + <kbd>C</kbd> when no text is selected should copy the current line.

_No response_

### Additional context

As in VS Code, Zed and others.

### Operating system (OS)

Cross-platform

2

None

AndrewDongminYoo:
## Description

In the code editor, a copy with an empty selection now copies the line holding the cursor, and pasting it puts that line back as a whole line above the cursor's line, the way VS Code and Zed do.
A copy with a selection is unchanged, and so is a paste of anything that was not copied this way.

Both halves are needed for the feature to be usable. A line copied with no selection but pasted at the caret splits whichever line the caret sits in, which is worse than the behavior the issue asks to replace.
The line-wise marker rides in the system-clipboard register through `VimRegisters::write_to_register`, which already exists for exactly this reason: its own doc comment says linewise motions "need to be pasted on their own line, NOT at the cursor position".
Reading it back through `read_from_register` compares the stored entry against the live system clipboard, so a copy made anywhere else in between falls back to a char-wise paste without any extra bookkeeping here.

The behavior is opt-in per surface, through a new `CodeEditorView::with_copy_line_when_selection_is_empty` builder, and only the standalone code editor in `app/src/code/view.rs` turns it on.
That restriction is the point of the change rather than an afterthought: `CodeEditorView` is embedded in AI blocks, diff views, the requested-command view, and the CLI subagent block, and every one of those relies on the editor emitting `CodeEditorEvent::CopiedEmptyText` when the selection is empty so the parent view can copy its own selection instead.
Copying a line unconditionally would make each embedded editor swallow a copy that belongs to its parent.

For the same reason the change stays out of `CodeEditorModel::copy`, which the TUI shares through `crates/warp_tui`, and leaves the `#[cfg(windows)] handle_windows_ctrl_c` path untouched, since Ctrl+C on Windows and Linux still has to reach the ancestor that treats it as an interrupt.
On macOS the copy binding resolves to Cmd+C (`cmd_or_ctrl_shift("c")` in `app/src/util/bindings.rs`), which is what issue #15849 describes; on other platforms the same action is Ctrl+Shift+C and is unaffected by the interrupt question.

## Linked Issue

Closes #15849.

## Design notes

These are the choices made where the issue text left the scope open.

The issue says "at least in the code editor", so this enables the behavior for the standalone code editor only, and leaves every embedded editor on the existing delegation path.
Extending it to another surface later is one builder call at that surface's construction site.

The copied text carries the trailing newline, so pasting reinserts a whole line, which is what VS Code and Zed do.
On the last line of a buffer that has no trailing newline the copied text has none either, and the paste normalizes it back to exactly one so the pasted line always occupies its own line.

The paste is gated on the same per-surface flag as the copy, rather than living inside `CodeEditorModel::paste`.
Putting it in the model would change what Cmd+V does in every embedded editor and in the TUI, including for a vim user who yanked line-wise into `+`, which is a wider behavior change than this issue asks for.

The line-wise entry is written to the `+` register rather than to a new store of its own. `+` is the system clipboard by definition, so a line-wise copy recorded there is a statement about the clipboard rather than a borrowed vim detail, and a later `"+p` in vim mode pasting it as a line is consistent rather than surprising.

"Empty selection" reuses the predicate the surrounding code already used, `CodeEditorView::selected_text`, which reports `None` when the selected text is empty rather than when the selection range is collapsed.
The new behavior deliberately hangs off that same predicate instead of introducing a second notion of emptiness next to it.

Cut is left alone and tracked separately in #15968. VS Code and the JetBrains IDEs cut the whole line on an empty selection too, but this issue asks about copy, and `CodeEditorModel::cut` is `copy` followed by `backspace` on every surface that shares the model.
Nothing becomes inconsistent in the meantime, and `a_cut_with_no_selection_leaves_no_stale_line_wise_paste` pins that: a cut writes empty text to the clipboard without touching the register, so `read_from_register` sees the mismatch and the paste after it stays char-wise.

Only the primary cursor's line is copied.
VS Code copies one line per cursor when several are active, and matching that is left as a follow-up rather than folded into this change.

An empty document has no line to copy, so the clipboard is left as it is instead of being cleared.

No feature flag is introduced.
The change is confined to one surface and one action, and reverting it is a single-line removal at each construction site.

## Testing

Commands run locally against this branch:

- `cargo nextest run -p warp` over the six tests below — passed, 6 run and 6 passed.
- `./script/format --check` — passed.
- `./script/check_no_inline_test_modules` — passed.
- `cargo clippy -p warp --all-targets --tests -- -D warnings` — passed, no warnings.
- `cargo build --bin warp-oss` — passed.

`./script/presubmit` was not run end to end.
This machine has no `wgslfmt`, `clang-format` or `pwsh`, and the diff touches no `.wgsl`, `.c`, `.h` or `.ps1` file, so those steps would have failed on absent tooling rather than on this change.
The steps of it that do read this diff are the ones listed above.

Nine regression tests were added to `app/src/code/editor/view/view_tests.rs`:

1. `copies_the_cursor_line_when_the_selection_is_empty` covers the requested behavior, including the trailing newline.
2. `copies_the_last_line_without_a_trailing_newline` covers the final line of a buffer that does not end in a newline.
3. `copies_only_the_selection_when_one_exists` proves a non-empty selection still wins.
4. `copying_an_empty_document_leaves_the_clipboard_untouched` proves an empty document does not clobber existing clipboard content.
5. `an_editor_that_delegates_empty_copies_does_not_take_the_cursor_line` is the non-regression test for every embedded editor, which must keep handing the copy to its parent view.
6. `pastes_a_copied_line_above_the_cursor_line` covers the paste half from a different line than the one copied.
7. `pastes_a_copied_line_whole_when_the_cursor_sits_mid_line` is the regression test for pasting with the caret away from column zero, which is the case that splits a line when only the copy half is implemented.
8. `an_editor_that_delegates_empty_copies_pastes_at_the_caret` proves an editor that has not opted in still pastes char-wise at the caret.
9. `a_cut_with_no_selection_leaves_no_stale_line_wise_paste` proves a cut does not leave the earlier line copy behind for the next paste to replay.

Tests 1 and 2 read the same two-line buffer from different cursor positions, so together they show the copied range follows the cursor rather than a fixed row.
Tests 5 and 8 are the negative halves: they run the same two actions on an editor that has not opted in and assert the old behavior, which is what demonstrates the opt-in actually gates both halves rather than the behavior being unconditional.

No integration test is included, and that was a deliberate call rather than an omission.
The framework could carry one: `crates/integration/src/test/goto_line.rs` already opens a real file in the code editor, and `warp::integration_testing::clipboard::assert_clipboard_contains_string` already asserts clipboard contents, so the test would be a file-open flow followed by a `cmd-c` keystroke and that assertion.
What the end-to-end layer would add over the unit tests is the keybinding wiring, since the unit tests dispatch `CodeEditorViewAction::Copy` directly rather than pressing the key.
I am glad to add it if you would like that covered here.

### Screenshots / Videos

https://github.com/user-attachments/assets/ddae1746-8e90-45c9-96f5-973dd7d0ef8e

The screen recording above has two segments. First, `json_serializable` is selected on purpose and copied and pasted, as the unchanged baseline: a copy with a selection still copies only the selection. Then, with the caret before `json_` and nothing selected, the same shortcut copies the whole indented line, and pasting with the caret partway along another line inserts it as a line above the cursor's line. The same steps are repeated in an AI block to show that an embedded editor still hands the copy to its parent.

## Agent Mode

CHANGELOG-IMPROVEMENT: Pressing the copy shortcut in the code editor with nothing selected now copies the current line, and pasting it inserts that line above the cursor's line.

### DIFF
--- a/app/src/code/editor/model.rs
+++ b/app/src/code/editor/model.rs
@@ -1909,28 +1909,61 @@ impl CodeEditorModel {
         ctx.clipboard().write(clipboard);
     }
 
-    /// Copy the line holding the primary cursor, including its trailing newline.
+    /// The text of the line holding the primary cursor, including its trailing
+    /// newline when the line has one.
     ///
-    /// An empty document has no line to copy, and this leaves the clipboard alone
-    /// rather than clearing it.
-    pub fn copy_current_line(&self, ctx: &mut ModelContext<Self>) {
-        let Some(range) = self.current_line_range(ctx) else {
-            return;
-        };
-        let clipboard = self.read_text_as_clipboard_content(range, ctx);
-        ctx.clipboard().write(clipboard);
+    /// `None` in a document with no line to read, so a caller can leave the
+    /// clipboard alone instead of clearing it.
+    pub fn current_line_text(&self, ctx: &AppContext) -> Option<String> {
+        let range = self.current_line_bounds(ctx);
+        if range.start >= range.end {
+            return None;
+        }
+        let buffer = self.content().as_ref(ctx);
+        Some(buffer.text_in_range(range).into_string())
+    }
+
+    /// Insert `text` as a whole line above the line holding the primary cursor.
+    ///
+    /// This is the paste half of a line-wise copy. Inserting at the start of the
+    /// line rather than at the caret is what keeps a copied line a line: pasting
+    /// at the caret would split whichever line the caret happened to sit in. The
+    /// insert lands before the cursor, so the cursor rides down with the text it
+    /// was already on.
+    pub fn paste_line_above_cursor(&mut self, text: &str, ctx: &mut ModelContext<Self>) {
+        let line_start = self.current_line_bounds(ctx).start;
+        // Normalize to exactly one trailing newline, so the inserted text occupies
+        // its own line whatever the clipboard happened to carry.
+        let without_newline = text.strip_suffix('\n').unwrap_or(text);
+        let line = format!("{without_newline}\n");
+        let edits = vec1![(line, line_start..line_start)];
+
+        let selection_model = self.selection_model.clone();
+        self.update_content(
+            |mut content, ctx| {
+                content.apply_edit(
+                    BufferEditAction::InsertAtCharOffsetRanges { edits: &edits },
+                    EditOrigin::UserInitiated,
+                    selection_model,
+                    ctx,
+                );
+            },
+            ctx,
+        );
+        self.validate(ctx);
     }
 
     /// The character range of the line holding the primary cursor. The range runs
     /// to the start of the next line, so it carries the trailing newline whenever
-    /// the line has one, and stops at the end of the buffer on the last line.
-    fn current_line_range(&self, ctx: &AppContext) -> Option<Range<CharOffset>> {
+    /// the line has one, and stops at the end of the buffer on the last line. It
+    /// is empty only in a document that has no line at all.
+    fn current_line_bounds(&self, ctx: &AppContext) -> Range<CharOffset> {
         let buffer = self.content().as_ref(ctx);
         let cursor = self.selections(ctx).first().head;
         let row = cursor.to_buffer_point(buffer).row;
         let start = Point::new(row, 0).to_buffer_char_offset(buffer);
         let end = Point::new(row + 1, 0).to_buffer_char_offset(buffer);
-        (start < end).then_some(start..end)
+        start..end
     }
 
     #[cfg(windows)]
--- a/app/src/code/editor/model.rs
+++ b/app/src/code/editor/model.rs
@@ -1909,6 +1909,30 @@ impl CodeEditorModel {
         ctx.clipboard().write(clipboard);
     }
 
+    /// Copy the line holding the primary cursor, including its trailing newline.
+    ///
+    /// An empty document has no line to copy, and this leaves the clipboard alone
+    /// rather than clearing it.
+    pub fn copy_current_line(&self, ctx: &mut ModelContext<Self>) {
+        let Some(range) = self.current_line_range(ctx) else {
+            return;
+        };
+        let clipboard = self.read_text_as_clipboard_content(range, ctx);
+        ctx.clipboard().write(clipboard);
+    }
+
+    /// The character range of the line holding the primary cursor. The range runs
+    /// to the start of the next line, so it carries the trailing newline whenever
+    /// the line has one, and stops at the end of the buffer on the last line.
+    fn current_line_range(&self, ctx: &AppContext) -> Option<Range<CharOffset>> {
+        let buffer = self.content().as_ref(ctx);
+        let cursor = self.selections(ctx).first().head;
+        let row = cursor.to_buffer_point(buffer).row;
+        let start = Point::new(row, 0).to_buffer_char_offset(buffer);
+        let end = Point::new(row + 1, 0).to_buffer_char_offset(buffer);
+        (start < end).then_some(start..end)
+    }
+
     #[cfg(windows)]
     /// If there is selected text, copy it. Otherwise, emit an event to allow
     /// an ancestor to handle the `WindowsCtrlC` event.
--- a/app/src/code/editor/view/actions.rs
+++ b/app/src/code/editor/view/actions.rs
@@ -1001,14 +1001,25 @@ impl TypedActionView for CodeEditorView {
             // The owner of the editor can also perform a copy by accessing the selected text and copying it to the clipboard.
             // This is the case when the code block is owned by an AIBlock and unfocused.
             Copy => {
-                self.model.update(ctx, |model, ctx| {
-                    model.copy(ctx);
-                });
-                // It's possible that the copy action was dispatched to the focused editor even when
-                // the user intended to copy selected text from a parent view (i.e. an `AIBlock`).
-                // The `CopiedEmptyText` event gives the parent view a signal to attempt a copy action.
-                if self.selected_text(ctx).is_none() {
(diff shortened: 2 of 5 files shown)


app/src/code/editor/view/actions.rs
```diff
@@ -1001,14 +1024,32 @@ impl TypedActionView for CodeEditorView {
...
             Copy => {
-                self.model.update(ctx, |model, ctx| {
-                    model.copy(ctx);
-                });
-                // It's possible that the copy action was dispatched to the focused editor even when
-                // the user intended to copy selected text from a parent view (i.e. an `AIBlock`).
-                // The `CopiedEmptyText` event gives the parent view a signal to attempt a copy action.
-                if self.selected_text(ctx).is_none() {
-                    ctx.emit(CodeEditorEvent::CopiedEmptyText);
+                let has_selection = self.selected_text(ctx).is_some();
```
bnavetta: This logic should go in the model layer rather than the editor view if possible (i.e. it should be handled by `CodeEditorMode::copy`)

app/src/code/editor/view/actions.rs
```diff
@@ -1001,14 +1024,32 @@ impl TypedActionView for CodeEditorView {
...
+                // An editor that owns its copy shortcut outright copies the line holding
+                // the cursor instead of clearing the clipboard. Every other editor keeps
+                // the default, because a parent view may hold the selection the user
+                // meant to copy.
+                if !has_selection && self.copy_line_when_selection_is_empty {
+                    let line = self.model.as_ref(ctx).current_line_text(ctx);
+                    if let Some(line) = line {
+                        // Writing through the system-clipboard register puts the text on the
```
bnavetta: Non-Vim editors should avoid overwriting Vim state. Instead, we can probably track this in the editor model or one of the selection models

app/src/code/editor/model.rs
```diff
@@ -1909,6 +1909,63 @@ impl CodeEditorModel {
...
+            ctx,
+        );
+        self.validate(ctx);
+    }
+
+    /// The character range of the line holding the primary cursor. The range runs
+    /// to the start of the next line, so it carries the trailing newline whenever
+    /// the line has one, and stops at the end of the buffer on the last line. It
+    /// is empty only in a document that has no line at all.
+    fn current_line_bounds(&self, ctx: &AppContext) -> Range<CharOffset> {
```
bnavetta, line 2037: Not blocking, but I'd make this a utility on the selection model

bnavetta (commented):
Thanks! Left a few comments, but nothing major

app/src/code/editor/view/actions.rs
AndrewDongminYoo: Done in 71c20cfb. The opt-in and the line-wise record both live on `CodeEditorModel` now, so `copy` and `paste` decide on their own and the view is back to dispatching.

`copy` returns a small `CopyOutcome` (`Selection` / `CursorLine` / `NothingToCopy` / `EmptySelection`) so the view still knows when to emit `CopiedEmptyText`: only on `EmptySelection`, which is the case where a parent may hold the selection the user meant. Keeping the per-surface opt-in on the model rather than the view preserves what it was for — the TUI and every embedded editor keep the default because the flag is off unless a construction site turns it on.

Moving it surfaced two regressions the tests caught, both from `copy` gaining a branch that other callers reach:

- `cut` is `copy` then `backspace`, so an opted-in editor started cutting the whole line to the clipboard while deleting one character. `cut` now calls a selection-only `copy_selection`, which is the copy every surface had before, so cutting is unchanged everywhere.
- An opted-in editor on a document with no line fell through to the default branch and cleared the clipboard. It returns `NothingToCopy` and leaves the clipboard alone, which is what the PR description claimed all along.

One note: `copy` had to become `&mut self` to record the line, which makes `handle_windows_ctrl_c` `&mut self` too. That path is `#[cfg(windows)]`, so it is a mechanical signature change I could not compile on this machine — worth a look on Windows.

app/src/code/editor/view/actions.rs
AndrewDongminYoo: Agreed, and the Vim registers are gone from this change entirely as of 71c20cfb.

The record is now a plain `line_wise_clipboard: Option<String>` on `CodeEditorModel`. Invalidation is the same idea `read_from_register` used, just without the register: `paste` compares the stored text against the live clipboard, so a copy made anywhere else in between — another editor, another app — no longer matches and falls back to a char-wise paste on its own.

One consequence worth your call rather than mine. `VimRegisters` is a singleton, so on the previous revision a line copied in one code editor pasted line-wise in another. A field on the model is per-instance, so that now falls back to char-wise. Both places you suggested are per-instance, so this follows the direction; if you would rather keep it app-wide, the equivalent would be a non-Vim singleton keyed the same way and I am happy to do that instead.

app/src/code/editor/model.rs
AndrewDongminYoo: Done as well, in the same commit. It is `BufferSelectionModel::primary_cursor_line_range` in `crates/editor/src/content/selection_model.rs`, next to `selected_lines` and `cursors_at_line_start`, and `CodeEditorModel::current_line_bounds` is now a one-line delegate. Since it landed in the shared crate rather than the app, the TUI gets the utility too.

PR #15992, opened 2026-09-14
Attach the ambient workload token to the built-in Factory MCP server

joeywangzr:
## Description

The built-in Factory MCP server installation only ever set an `Authorization` header, for both interactive and CLI/cloud agent runs. Cloud agent runs authenticate to warp-server's `/mcp/factory` route with a task-scoped API key, which the server resolves back into a live ambient task -- but with no `X-Warp-Ambient-Workload-Token` attached, `ValidateWorkloadTokenForRun` always saw an empty token and silently soft-failed instead of verifying the caller is that task's own worker.

This resolves the ambient headers (workload token, cloud-agent ID) for the run's active task, when it has one, and attaches them alongside the bearer token. Interactive (GUI/TUI) sessions have no active ambient task and are unaffected.

## Linked Issue

- [ ] No linked GitHub issue -- found via a code review discussion on a related warp-server PR, not a filed issue.

## Testing

- Ran `cargo check`, `cargo nextest run` (scoped to the touched modules), `cargo fmt --check`, and `cargo clippy -- -D warnings` (without `--all-features`, which fails identically on a clean checkout due to an unrelated, pre-existing `agent_mode_evals` compile break). All clean.
- Added `factory_installation_attaches_ambient_headers_for_an_active_task`, covering that the workload token and cloud-agent ID land in the resolved MCP config.
- [ ] Not manually tested end to end with `./script/run`: reproducing this requires a real dispatched cloud agent task calling the Factory MCP server, which needs a live warp-server deployment carrying the corresponding server-side validation path. Requesting that as a pre-merge check.

## Agent Mode


crates/warp_server_client/src/base_client.rs
```diff
@@ -258,43 +263,99 @@ impl BaseClient {
...
+    /// into a long-lived transport, rather than resolving one per request, is therefore worse
+    /// off pinning a token that expires partway through than pinning none, and uses this to
+    /// distinguish the two cases up front.
+    pub async fn get_ambient_workload_token_valid_until(
+        &self,
+        must_outlive: DateTime<Utc>,
+    ) -> Result<Option<String>> {
+        let requested_duration = (must_outlive + AMBIENT_WORKLOAD_TOKEN_EXPIRY_MARGIN - Utc::now())
+            .to_std()
+            .unwrap_or(AMBIENT_WORKLOAD_TOKEN_DURATION)
```
captainsafia: I think this can just be `unwrap_or_default` since the max will fallback to the `AMBIENT_WORKLOAD_TOKEN_DURATION` anyways?

captainsafia (approved):
This seems sensible!

My biggest issue is the fact that we omit the token from the header when it's expired even though we have a refresh loop. Fixing this is slightly out-of-scope for this change since it seems we'll have to set up the Factory MCP server to support dynamically resolving the header values instead of baking them in at startup.

PR #16002, opened 2026-09-14
Improve agent prompt prefix-cache stability

### LINKED ISSUE #16003: Improve Agent prompt layout for prefix-cache reuse
## Problem

Warp Agent requests can contain a large amount of stable prompt context. Volatile fields and nondeterministic ordering placed before that stable context reduce the amount that conventional contiguous prefix caches can reuse.

While testing Warp Agent against a local OpenAI-compatible inference server with prefix caching enabled, I captured two otherwise identical Agent requests containing roughly 37.6k prompt tokens.

The first overall divergence between the requests was a changing `current_run_id` inside the orchestration runtime context. Later in the request, the client-generated context also diverged at `CurrentTime`, followed by a different ordering of skills.

The client-side causes are reproducible in the current code:

- `CurrentTime` is added before execution-environment, codebase, and skills context.
- `SkillDeduplicator` stores descriptors in a `HashMap` and returns `into_values().collect()`, so identical skill sets can be serialized in nondeterministic order.

## Impact

At this prompt size, prefix reuse has a large latency impact. In one measured case with a working prefix cache:

- cold/re-prefilled request: ~89.5 s TTFT
- repeated request with 32,768 cached prefix tokens: ~13.0 s TTFT

These measurements demonstrate the value of preserving a long stable prompt prefix; they are not a benchmark of the proposed code change itself.

## Proposed changes

For the client-side portion:

1. Move `CurrentTime` after the relatively stable execution-environment, codebase, and skills context.
2. Sort skill descriptors deterministically before adding them to Agent context.
3. Add regression coverage for deterministic skill ordering.

There is also an earlier cache boundary caused by `current_run_id` in the orchestration runtime context. I could not locate generation of that field in the open client source, so that part likely needs to be addressed in the layer that produces the orchestration prompt, for example by moving volatile runtime metadata after stable context where possible.

## Pull request

Implementation for the client-side changes is already available in #16002.

trigger2k20:
## Summary

This PR improves prefix-cache stability for Agent requests by making stable prompt context more deterministic and moving volatile context later in the request.

It makes two related changes:

- Sort skill descriptors deterministically before they are added to the Agent context.
- Move `CurrentTime` after the relatively stable execution-environment, codebase, and skills context.

## Motivation

While testing Warp Agent against a local OpenAI-compatible inference server with prefix caching enabled, I observed that otherwise equivalent Agent requests contain roughly 37.6k prompt tokens.

Prefix caching has a significant impact at this prompt size. In one measured case:

- cold/re-prefilled request: ~89.5 s TTFT
- subsequent request with 32,768 cached prefix tokens: ~13.0 s TTFT

This makes the position and determinism of changing prompt fields important.

Capturing two otherwise identical Agent requests showed multiple early prefix divergences.

One client-side divergence is `CurrentTime`, which was inserted before execution-environment, codebase, and skill context. Since it changes every request, all context following it becomes unavailable to a conventional contiguous prefix cache.

Another divergence was the ordering of skills. `SkillDeduplicator` stores descriptors in a `HashMap` and previously returned `into_values().collect()`. HashMap iteration order is not stable, so identical sets of skills could be serialized in different orders between requests.

## Changes

`SkillDeduplicator::into_descriptors()` now sorts descriptors by their canonical `SkillReference` representation before returning them. A regression test verifies deterministic ordering.

`CurrentTime` is now appended after execution-environment, codebase, and skill context, allowing those more stable sections to remain part of the reusable prefix.

`additional_context` remains after `CurrentTime` to keep the change conservative and avoid changing the relative semantics of additional request-specific context.

## Additional observation

The captured requests also contained an earlier-changing `current_run_id` inside the `Orchestration runtime context`. This is actually an earlier prefix boundary than `CurrentTime`.

I could not find generation of that field or the corresponding orchestration prompt block in the open client code, so this PR intentionally does not attempt to modify it.

Addressing that upstream/runtime-generated field — for example by moving volatile orchestration metadata after stable context where possible — could provide additional prefix-cache reuse beyond the improvements in this PR.

## Validation

- Added regression coverage for deterministic skill ordering.
- `cargo test -p warp test_unique_skills_returns_deterministic_order -- --nocapture`
- `cargo check -p warp`
- `cargo fmt`
- `git diff --check`

All checks pass.

Closes #16003


trigger2k20:
The PR has been linked to #16003 via Closes #16003 as requested. Issue #16003 has also been triaged and assigned.

As I understand the contribution workflow, the remaining step is for a maintainer to mark #16003 ready-to-implement, after which I can re-trigger the review with /warp-agent-review.

Please let me know if anything else is needed from my side in the meantime.

PR #16006, opened 2026-09-14
Report only the repos that actually failed in parallel clone errors

joeywangzr:
## Description

When an environment has multiple repos to clone, `clone_checkout_requests` clones them all in parallel via one shell script and only checked the script's overall exit code. If **any single** repo's clone job failed, the resulting `PrepareEnvironmentError::CloneRepo` reported **every repo name in the batch** as failed, not just the actual offender(s) — so a failing environment setup could claim a large repo list all failed to clone when in reality only one did.

This surfaced while investigating a real dogfood report where a 19-repo environment's setup failed with `Failed to clone [all 19 repos]`, which made it impossible to tell which repo(s) actually caused the failure.

This PR tracks each background clone job's pid individually and records the specific repo name to a marker file only when that job's `wait` fails. After a failed run, the client reads the marker file back and reports only the repos that were actually recorded as failed, falling back to reporting the whole batch if the marker can't be read (e.g. the script errored before reaching the wait loop).

## Linked Issue

- [ ] No linked GitHub issue — found via log/code investigation of a dogfood report, not a filed issue.

## Testing

- Updated the existing `build_parallel_clone_command` script-generation tests for the new per-repo pid/wait shape and the added `failed_repos_path` parameter.
- Added `parallel_clone_command_records_only_the_failing_repo_name_on_partial_failure`, which drives the real generated script end-to-end (via a stubbed `clone_repo`) and asserts only the genuinely failing repo gets written to the marker file.
- `cargo check -p warp --lib`, `cargo nextest run -p warp --lib` (environment module, 44/44 passed), `./script/format`, and `cargo clippy -p warp --lib --tests -- -D warnings` all clean.
- [ ] Not manually tested with `./script/run`: this only changes the error message/diagnostics path for a multi-repo cloud environment clone failure, which isn't practical to trigger locally.

## Agent Mode


app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -921,11 +931,19 @@ async fn clone_checkout_requests(
                 full: ("Cloning repositories via terminal: {}", repo_names.join(", "))
             );
 
-            let command = build_parallel_clone_command(repos, shell_type);
+            let failed_repos_path = format!("/tmp/.warp-clone-failed-{}", Uuid::new_v4());
```
bnavetta: This should respect `TMPDIR` - can use https://doc.rust-lang.org/std/env/fn.temp_dir.html

app/src/ai/agent_sdk/driver/environment.rs
```diff
@@ -938,6 +956,29 @@ async fn clone_checkout_requests(
...
+/// nothing could be read, so the caller can fall back to reporting the whole
+/// batch instead of an empty list.
+async fn read_failed_repo_names(
+    failed_repos_path: &str,
+    spawner: &ModelSpawner<TerminalDriver>,
+) -> Option<Vec<String>> {
+    let shell_type = active_shell_type(spawner).await;
+    let escaped_path = shell_escape_single_quotes(failed_repos_path, shell_type);
+    let command = format!("cat '{escaped_path}' 2>/dev/null; rm -f '{escaped_path}'");
+    let output = execute_silent_command(command, spawner).await.ok()?;
```
bnavetta: Instead of reading through a shell command, we can read+delete the file directly. The driver only needs to handle local sessions, not SSH, so we can assume the local filesystem is that of the terminal session.

app/src/ai/agent_sdk/driver/environment_tests.rs
```diff
@@ -362,15 +363,57 @@ fn parallel_clone_command_runs_repos_in_background_and_waits() {
...
+            "bad".to_string(),
+        ),
+    ];
+    let script = unwrap_sh_c_script(&build_parallel_clone_command(
+        &repos
+            .into_iter()
+            .map(|repo| clone_request(repo, None))
+            .collect::<Vec<_>>(),
+        ShellType::Bash,
+        "/tmp/warp-clone-failed-repos-test-marker",
```
bnavetta: On the off chance that there are multiple test instances running at once, let's use a test-scoped marker path

bnavetta (approved):

PR #16016, opened 2026-09-15
Fix WarpOss local macOS builds

### LINKED ISSUE #16017: WarpOss local macOS builds access production app-group data and unnecessarily create a DMG
## Summary

Local WarpOss builds on macOS currently have two related development-build issues:

1. WarpOss uses Warp's production app-group container for secure state, causing macOS to repeatedly request access to data from other apps when the OSS build is independently signed.
2. The local macOS run script invokes `cargo bundle` without restricting the output format, causing unnecessary DMG creation even though the development launcher only needs the `.app` bundle.

## WarpOss accesses the production app-group container

`secure_state_dir()` uses `app_group_container_path()` on macOS.

The app-group path is derived from Warp's production Apple team ID and resolves to:

`~/Library/Group Containers/2BBY89MBSN.dev.warp`

A locally built WarpOss application has bundle ID:

`dev.warp.WarpOss`

and, when independently signed by an OSS contributor, does not belong to Warp's production Apple team/app group.

Nevertheless, WarpOss attempts to store its SQLite state under:

`~/Library/Group Containers/2BBY89MBSN.dev.warp/Library/Application Support/dev.warp.WarpOss/`

## Observed macOS behavior

Using macOS TCC logging, the recurring permission request was identified as:

`kTCCServiceSystemPolicyAppData`

Filesystem tracing confirmed that WarpOss was accessing its state inside the production `2BBY89MBSN.dev.warp` group container.

This results in the recurring macOS dialog asking WarpOss for permission to access data from other applications.

Resetting TCC permissions and signing WarpOss with a stable Apple Development identity did not solve the problem.

Changing `secure_state_dir()` so that `Channel::Oss`, like `Channel::Integration`, does not use the production app-group secure-state directory resolved the issue. After rebuilding, the recurring App Data permission dialog disappeared.

## Unnecessary DMG creation

`script/macos/run` only consumes and launches the generated `.app` bundle, but currently invokes:

`cargo bundle`

without specifying a format.

This also attempts DMG generation. On the tested development machine, `hdiutil` failed while unmounting the temporary disk image, causing `./script/run` to fail even though `WarpOss.app` had already been successfully generated.

Using:

`cargo bundle --format osx`

produces the application bundle required by the local development launcher without invoking the unnecessary DMG creation path.

## Proposed fix

- Treat `Channel::Oss` like `Channel::Integration` in `secure_state_dir()` so local OSS builds do not use Warp's production app-group container.
- Use `cargo bundle --format osx` in `script/macos/run`.

Both changes have been tested successfully on macOS.

Implementation: #16016

trigger2k20:
## Summary

This PR fixes two issues affecting local WarpOss builds on macOS:

- Prevent the OSS channel from using Warp's production app-group container for secure state.

- Build only the macOS `.app` bundle when running the local macOS development launcher, avoiding unnecessary DMG creation.

## WarpOss app-data access

`secure_state_dir()` currently uses `app_group_container_path()` on macOS. That path is derived from Warp's production Apple team ID and resolves to:

`~/Library/Group Containers/2BBY89MBSN.dev.warp`

This is appropriate for official Warp builds, but not for a locally built and independently signed `WarpOss.app`.

A locally signed WarpOss build has a different signing identity/team and therefore does not belong to Warp's production app group. Nevertheless, the OSS build attempts to use that production group container for its SQLite state.

On macOS this causes a `kTCCServiceSystemPolicyAppData` authorization request and results in the recurring dialog:

> "WarpOss would like to access data from other apps."

The behavior was reproduced with filesystem and TCC logging. WarpOss was observed accessing its SQLite state below:

`~/Library/Group Containers/2BBY89MBSN.dev.warp/Library/Application Support/dev.warp.WarpOss/`

The fix treats `Channel::Oss` like `Channel::Integration` in `secure_state_dir()` and returns `None`, allowing the OSS build to use its normal application-specific state location instead of Warp's production app-group container.

After rebuilding WarpOss with this change, the recurring macOS App Data permission dialog no longer occurs.

## Local macOS build

The local macOS launcher only needs the generated `.app` bundle, but `cargo bundle` also attempts to create a DMG.

On development machines this can cause `./script/run` to fail during DMG creation even though `WarpOss.app` has already been built successfully. In the reproduced case, `hdiutil` failed while unmounting the temporary disk image.

The launcher now invokes:

`cargo bundle --format osx`

This explicitly requests the macOS application bundle and skips DMG generation, which is not needed by the local development run path.

## Scope

These changes are limited to the local/OSS macOS development path:

- Production Warp channels continue using the existing app-group secure-state behavior.

- Distribution/packaging workflows that need DMGs are unchanged.

- The local runner still produces and launches `WarpOss.app` as before.

## Validation

Tested on macOS with a locally built and independently signed WarpOss application.

Before the secure-state change:

- WarpOss accessed the production `2BBY89MBSN.dev.warp` group container.

- macOS repeatedly requested App Data access.

After the change:

- WarpOss no longer uses the production app-group container for secure state.

- The recurring App Data permission dialog is gone.

The local build also completes without invoking the unnecessary DMG creation path.

- `cargo fmt`

- `git diff --check`

- WarpOss built and launched successfully with `./script/run`

Closes #16017


trigger2k20:
The PR has been linked to #16017 via Closes #16017 as requested. Issue #16017 has also been triaged and assigned.

As I understand the contribution workflow, the remaining step is for a maintainer to mark #16017 ready-to-implement, after which I can re-trigger the review with /warp-agent-review.

Please let me know if anything else is needed from my side in the meantime.

PR #16034, opened 2026-09-16
Split is_headless into is_gui/is_headless and rename AppBackend::Headless to Windowless

kevinyang372:
## Description

`is_headless` was overloaded: it meant both "the app runs with no UI at all" **and** "the app runs on the windowless app backend". That made the TUI awkward to classify — it runs on the windowless backend but very much has a UI.

This PR separates the two concepts and renames the backend variant so its name describes what it is rather than what it's typically used for.

**`LaunchMode` now has both predicates** (`app/src/lib.rs`)
- `is_gui()` — `App`, `Test`, and `agent run --gui`. False for TUI, CLI, proxy, daemon.
- `is_headless()` — CLI (non-`--gui`), proxy, daemon. False for `App`, `Test`, **and TUI**.
- Every previous `is_headless()` caller was actually asking "is there a GUI?" (app-backend choice, macOS background-only activation policy, Dock icon/menu/menu-bar setup, `set_app_icon`, local HTTP server), so they all switched to `is_gui()`. Behavior is unchanged for every launch mode.
- The microphone-authorization query, previously hard-coded to `LaunchMode::Tui`, is now gated on `!is_headless()` ("a user is present"). It's a no-op on the GUI backend, so this is also behavior-preserving.

**`AppBackend::Headless` → `AppBackend::Windowless`** (`crates/warpui/src/platform/app.rs` + the mac/linux `AppBuilderExt` match arms). The two `AppBuilder` methods tied to that variant are renamed to match: `new_headless` → `new_windowless`, `enable_headless_microphone_access_query` → `enable_windowless_microphone_access_query`. The `platform::headless` module itself is untouched.

**Same overload one layer down:** `Delegate::is_headless()` / `AppContext::is_headless()` in `warpui_core` only ever meant "windowless backend" (the TUI's delegate reports it too), so they are now `is_gui()` (default `true`; the windowless delegate returns `false`). The single caller, `compute_block_size` in `terminal_manager.rs`, was updated.

## Linked Issue
N/A — small internal refactor.

## Testing
Pure rename/split with no behavioral change for any launch mode. Verified locally with:
- `cargo check -p warp --lib --tests --features tui`
- `cargo clippy -p warpui --all-targets --tests -- -D warnings`
- `cargo clippy -p warp --lib --features tui -- -D warnings`
- `./script/format`

Relying on CI for the broader cross-platform build and test matrix.

## Agent Mode

Conversation: https://staging.warp.dev/conversation/3ce5b135-cecf-4e4a-849e-621485be7371


app/src/terminal/terminal_manager.rs
```diff
@@ -93,6 +88,11 @@ pub(super) fn compute_block_size(
             appearance.monospace_font_size(),
             appearance.ui_builder().line_height_ratio(),
         )
+    } else {
+        // A windowless backend has no font since it doesn't render with one. We skip the
+        // font-based size computation and hardcode a terminal size, so that viewers of the
+        // shared session see a reasonable terminal width.
+        SizeInfo::new_without_font_metrics(24, 120)
```
vorporeal, line 95: And this isn't used for TUI because we're using the grid dimensions of the terminal that is hosting us, yeah?

vorporeal (approved):
Thanks! Windowless is a great naming choice here.

app/src/terminal/terminal_manager.rs
kevinyang372: Yep!

PR #16045, opened 2026-09-16
computer_use: centralize the recording temp-file path across recorders

joeywangzr:
## Description
Windows, macOS, and Linux each independently built the recording output path (`warp-recording-<uuid>.mp4` under the system temp directory) and its sibling `.log` file. Extracted the shared logic into a new `recording_paths` module so the three recorders can't drift on the naming convention.

Follow-up from vkodithala's review comment on #15964: a drive-by, out-of-scope-for-that-PR suggestion to centralize this since it was pre-existing duplication.

Stacked on #16000 since it touches the same three recorder files.

## Linked Issue
Not its own Linear ticket; a follow-up cleanup noted on [REMOTE-2064](https://linear.app/warpdotdev/issue/REMOTE-2064/investigate-windows-screen-recording-for-computer-use)'s stack.
- [x] N/A, drive-by cleanup from a review comment
- [x] No screenshots needed, no user-visible or UI change

## Testing
Added `recording_paths_tests.rs`: asserts the allocated path follows the `<prefix>-<uuid>.mp4` naming convention, that the sibling `.log` file is created on disk, and that repeated calls produce distinct paths.

Ran `cargo nextest run -p computer_use` (85 tests, all passing) and `cargo clippy -p computer_use --all-targets --tests -- -D warnings` on macOS, plus `cargo clippy -p computer_use --target x86_64-pc-windows-msvc --all-targets --tests -- -D warnings` cross-compiled clean. No Linux target available locally; relying on CI for that leg.

## Agent Mode

CHANGELOG-NONE


crates/computer_use/src/recording_paths.rs
vkodithala, line 1: nit/thinking out loud: it seems like an antipattern that we're creating a brand-new file for a shared path resolver. i wonder if the right thing to do here is build a utils file for recordings like `computer_use/src/video_utils.rs` (we have something similar for screenshots) and just plop cross-platform utils there. there might be an opportunity to converge with `recording_metadata`.

vkodithala (approved):
nice improvement Joey! looks good, left a small comment (non-blocking)

crates/computer_use/src/recording_paths.rs
joeywangzr: The crate already names helpers by concern (recording_metadata.rs, recording_post_process.rs), so recording_paths.rs fits that pattern. A video_utils.rs catch-all would obscure ownership. I can combine recording_paths + recording_metadata into recording_files.rs if you think that would be better but I think its more accurate this way.

PR #16046, opened 2026-09-16
Fix steering delivery for follow-ups with file attachments

seemeroland:
## Description
Fixes shared-session follow-ups with file attachments waiting until the entire agent turn finishes instead of joining the next steering request. Attachments now download eagerly when a prompt is queued, without interrupting an active stream.

### What went wrong
The steering path introduced in #15921 builds the next request synchronously. It explicitly skipped queued prompts containing a `FileReference`, because those files needed asynchronous download. Downloads only started when the prompt was dispatched through the immediate or turn-completion path. A prompt received during an active stream therefore remained queued across subsequent tool-result and orchestration-event requests.

This also affected apparently text-only Slack replies: the Factory mapped-thread path attaches the raw webhook body as `event-payload.json`. In the staging reproduction, the second follow-up reached the worker at **13:47:32.913 UTC**, but its download did not start until **14:50:10.001 UTC**, despite many intervening tool calls and request boundaries.

### How this fixes it
- Start attachment preparation on enqueue and keep the row in the queue until it is ready. `QueuedQueryModel` owns per-row readiness and resolved files, exposed through `ready_head` and `ready_query`.
- Emit `PromptReady` when preparation settles. Ready prompts join the next normal steering request; completion can also wake a genuinely idle conversation. An inactive response stream alone is not sufficient while tool work remains in progress.
- Reuse prepared files for both automatic and explicit sends. There is no second download at dispatch, and automatic FIFO delivery never skips an unready head.
- Ignore stale completions after removal or clearing. Failed downloads settle with a missing-file notice rather than stranding the prompt. Pending injections keep the worker's idle timeout from expiring, and manual-send controls stay disabled during preparation.

Eager preparation preserves steering's non-interrupting behavior. Restoring the old download-then-interrupt path would avoid the delay but could cancel useful in-flight work.

Review focus: readiness versus dispatch eligibility, FIFO ordering when downloads finish out of order, and worker lifetime while preparation is pending.

## Linked Issue
[Staging reproduction: execution 224861](https://admin.staging.warp.dev/debug/ambient-agent/01a0aa72-6155-770c-84a0-a40d375336b5#execution-224861)

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. No tracking issue supplied; reproduction linked above.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes). No visual capture; diagnosis used worker and server logs.

## Testing
- **108 targeted tests passed**, covering attachment utilities, queue readiness/FIFO, steering, terminal draining, and startup lifecycle. Nextest reported one passing lifecycle test as leaky: `native_startup_queue_prevents_exit_until_a_local_row_is_removed`.
- `cargo check -p warp` passed.
- `cargo clippy -p warp --all-targets --tests -- -D warnings` passed.
- `./script/format` completed; `git diff --check` passed.

Validation ran on this implementation before commit; the pushed code matches that candidate. End-to-end integration coverage was not added, by agreement; manual reproduction of the fixed cloud flow remains pending.

## Agent Mode

[Investigation and implementation conversation](https://staging.warp.dev/conversation/481c869f-4bd0-4795-9be9-dfbad6b182bb)

CHANGELOG-OZ: Fixed follow-up messages with attachments waiting until an agent finished its turn instead of being delivered at the next steering opportunity.


harryalbert (commented):
LGTM! Thanks for fixing this. 

Was reviewing w/ the agent and came up w/ one edge case that I think is worth fixing, but other than that this makes sense. The scenario is:
```
1. A promptless native run creates an empty conversation whose status is InProgress (app/src/ai/agent/conversation.rs:466-472).
2. A startup follow-up with a file arrives and begins downloading.
3. The driver releases native setup and performs its one-time queue drain (app/src/ai/agent_sdk/driver.rs:4058-4074).
4. Because the row is still unready, that drain does nothing.
5. The row keeps the driver from completing (app/src/ai/agent_sdk/driver.rs:4075-4090).
6. When the download completes, PromptReady is emitted.
7. The PromptReady subscriber requires conversation_ready_for_pending_events, which only accepts Success or WaitingForEvents (app/src/ai/blocklist/controller.rs:627-646, app/src/ai/blocklist/controller.rs:1838-1860).
8. The empty promptless conversation is still InProgress, so nothing dispatches it. There is no stream or action left to produce another drain.

The queue therefore remains nonempty and the worker remains alive indefinitely.

The existing can_dispatch_queued_warp_agent_prompt predicate already recognizes this special case through conversation.exchange_count() == 0 (app/src/ai/blocklist/controller.rs:2865-2884). The PromptReady handler should likely use that predicate, or equivalently admit the zero-exchange promptless state.
```
I know that, now that almost all cloud runs are being kicked off via the factory, promptless cloud runs are very rare, but I still think this is probably worth addressing if possible.

harryalbert (approved):
meant to approve

PR #16052, opened 2026-09-16
Fix sharer reconnects that stall before server confirmation

seemeroland:
## Description
Fix shared-session sharers getting stuck in `Reconnecting` when a replacement WebSocket closes or stalls before server confirmation. This failure was observed during production run investigations and independently verified against current `warp` code.

### Observed failure and root cause
Two investigated runs outlived their shared sessions rather than simply hitting the expected process-exit/sandbox-shutdown gap:
- In one case, server-side participant lookup failed during reconnect. The client logged EOF with `stage=reconnecting`, with no subsequent retry or failure event. After session expiry, a follow-up triggered early handoff; the original process continued executing commands.
- In another, the server's idle/grace period expired before the client detected a broken pipe. Its reconnect received EOF and became stuck in the same state. The original instance continued making credential requests after replacement.

The retry future previously succeeded after socket connection and authentication, before sending `Reconnect` or receiving `SessionReconnected`. The close handler did not recover from `Reconnecting`, and there was no confirmation timeout. The recursive retry helper's original abort handle also did not cancel later scheduled retries.

### Why this fixes it
- Keep connect, authentication, reconnect send, and confirmation inside each retry attempt. EOF, stream errors, send failures, and missing confirmation can no longer leave a completed retry future behind a stuck state.
- Bound each attempt to 10 seconds and the complete retry cycle, including backoff, to 128 seconds. One cancellable loop stops pending attempts when sharing ends.
- Preserve explicit server rejection/termination as terminal outcomes, buffer pre-confirmation updates with count/byte limits, and ignore callbacks from obsolete transports.
- Retain current IAP refresh, startup/team-sharing behavior, terminal replay, and pending-input flushing.

This does not fix the initiating server participant-lookup failure, add a half-open connection watchdog, or forcibly stop old agent processes. Production identifiers and raw logs are omitted from this public PR.

## Linked Issue
Motivated by the production shared-session investigation. Internal evidence is recorded in the permission-controlled conversation linked below.
- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. No public issue was filed.
- [ ] Where appropriate, screenshots or a short video of the implementation are included below (especially for user-visible or UI changes). No visual UI change; validation uses injected networking failures.

## Testing
- `cargo nextest run -p warp --lib -E 'test(terminal::shared_session::sharer::network::tests::)' --no-fail-fast`: **39 passed**, including 20 independently named regression tests in the full current app.
- `cargo check -p warp --lib`: passed.
- `cargo clippy -p warp --all-targets --tests -- -D warnings`: passed.
- Scoped Rust formatting with repository settings and `git diff --check`: passed.

Coverage includes retry-to-success after EOF/error/ack timeout, stalled transport/send, retry exhaustion, whole-cycle deadlines, cancellation during backoff, explicit rejection, stale callbacks, bounded buffering, and pending-input preservation.

The stale-message and stale-EOF tests wait for foreground callback completion rather than sleeping. Mutation validation confirmed that removing only the generation comparison makes both tests fail at their replacement-survival assertions; production code was restored afterward. An existing buffering test emitted a passing nextest output-handle leak warning once; its isolated rerun passed cleanly.

The requester approved targeted networking tests rather than an additional end-to-end shared-session test. No before-fix failing test run is claimed.
- [ ] I have manually tested my changes locally with `./script/run`. Not run; validation is automated, without a manual GUI session.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed shared sessions getting stuck reconnecting when the connection closes or stalls before the server confirms reconnection.

## Agent artifacts
- [Investigation and implementation conversation](https://staging.warp.dev/conversation/5d2605c1-67e6-4eb1-a2cb-deaa3c8a48fa)
- [Implementation plan](https://staging.warp.dev/drive/notebook/sg8Ej8XGZRWIbewWzMpmqE)


sohumgaitonde (approved):
This fix makes sense - looks good overall!

PR #16056, opened 2026-09-17
Bound shared session uploads to the session sharing server's frame limit

kevinyang372:
### Problem

On staging, an Oz run's sharer (the sandbox's Warp app) tried to send a single 18,633,118-byte websocket message to session-sharing-server. The server rejected it (`Space limit exceeded: Message too long: 18633118 > 16777216`) and dropped the connection. The sharer treated that as a transient disconnect, reconnected, replayed the identical message from its unacked queue, got dropped again, and looped ~4×/sec for the remaining 90 minutes of the run. No further events from the run reached the server, so every viewer (native and web) was frozen.

### How upload messages are assembled (finding)

- Every `OrderedTerminalEventType` becomes exactly **one** websocket message: `send_ordered_terminal_event_message` wraps it in `UpstreamMessage::OrderedTerminalEvent`, stores it in `unacked_terminal_events`, and pushes it to the send task, which does `message.to_json()` and writes one text frame. There is no cross-event batching.
- The only batching is for `PtyBytesRead`: raw PTY reads are accumulated for ~50 ms, LZ4-compressed, and sent as one event. Note that serde_json encodes `Vec<u8>` as a JSON array of numbers, i.e. up to 4 bytes of wire per compressed byte.
- `AgentResponseEvent { response_event }` is produced by `TerminalModel::send_agent_response_for_shared_session` from every MAA `ResponseEvent` the sharer receives: protobuf-encoded, base64'd, and forwarded as-is. A `ClientActions` echoing a huge tool result (warp-server independently logged the same conversation's MAA request exceeding its 10 MB Pub/Sub limit) therefore becomes a single ~18.6 MB message. That is the only realistic source of an oversized event; `Init`/`Finished`/control events are tiny.
- The 16 MiB ceiling is tungstenite's default **`max_frame_size`** on the server, which neither sharer endpoint overrides. The client sends each message as a single frame, so raising the server's `max_message_size` would not help; the cap has to be enforced client-side.
- Nothing on the client bounded message size, and the reconnect path (`SessionReconnected` → `flush_terminal_events_to_server`) resends unacked events verbatim, so a deterministic per-message rejection loops forever.

### Change

New module `app/src/terminal/shared_session/sharer/message_size.rs`, wired into `Network`:

1. **Constants** – `SERVER_MAX_WEBSOCKET_MESSAGE_BYTES` (16 MiB, the server's frame limit), `MAX_ORDERED_TERMINAL_EVENT_BYTES` (8 MiB soft cap for a single ordered event), `MAX_PTY_BATCH_RAW_BYTES` (1 MiB of raw PTY bytes per compressed event).
2. **Bound at the source** (`send_ordered_terminal_event_message`): every ordered event is measured (exact JSON wire length, computed with a counting writer and a borrowing mirror of the protocol types, no clone) before it is numbered.
   - An oversized `AgentResponseEvent` carrying `ClientActions` is split: actions are greedily packed into as many events as needed (each with the same `response_initiator` / `forked_from_conversation_token`); an `AddMessagesToTask` that doesn't fit on its own is split one message per action (this is already how the sharer's own scrollback replay emits messages, so viewers handle it identically); a single message that still exceeds the cap is dropped and logged with its task/request/message ids.
   - Any other oversized event is indivisible and dropped with an error.
   - Event numbers are assigned only to what is actually sent, so the sequence stays contiguous and acks are unaffected.
3. **PTY batches** are chunked by raw size before compression so a burst of incompressible output can't produce an oversized event.
4. **Wire-level backstop** (send task in `on_websocket_connected`): if a serialized message still exceeds the server's frame limit, an `OrderedTerminalEvent` is replaced by a same-numbered no-op placeholder (empty `ClientActions` for agent events, empty compressed chunk for PTY), any other message is skipped, and an oversized `Initialize` fails startup. This covers every path that reaches the socket — live sends, initial flush, and reconnect replay — so a rejected message can never be resent verbatim.
5. **Observability**: splits log a `warn` with original size / event count / drop count; each dropped payload and each wire-level replacement is a `report_error!` (Sentry) with the size, the limit, and a payload-free label (event kind, event_no, message ids), all tagged with `session_id` / `source_task_id`.

### What a viewer sees

- **Split event**: nothing different — the same actions arrive in the same order across consecutive events.
- **Dropped message**: that single message (in the observed case, one giant tool-call result) is missing from the viewer's conversation; the tool call renders as still running. Everything before and after it (later actions in the same event, subsequent events, `Finished`) is delivered normally, so the conversation keeps streaming. Neither the MAA proto nor the session-sharing protocol has a "truncated" notion, so this is the least-bad option; the drop is logged with the message/request ids.
- **Wire-level placeholder**: a no-op. Native viewers apply an empty `ClientActions`; session-sharing-server's conversation classifier attributes it to the open request like any `ClientActions`.

### Decisions / risks for review

- Cap is 8 MiB (half the frame limit) to leave headroom for envelope/JSON overhead and future protocol growth.
- Splitting only handles `ClientActions`/`AddMessagesToTask`. An oversized `CreateTask` (e.g. a resume carrying a full task history) would be dropped whole; I did not attempt to split it because I'm not confident `CreateTask` + `AddMessagesToTask` is equivalent on the viewer.
- The wire backstop reports via `report_error!` from the send task; it should be unreachable now that events are bounded at the source, so a hit means a new producer bypassed the bound.
- Unacked oversized originals are not rewritten in `unacked_terminal_events`; they are substituted at the wire on each replay. Cheap to add if we'd rather self-heal the queue.

## Linked Issue

## Testing

Automated:
- `message_size_tests.rs`: wire-length measurement matches `UpstreamMessage::to_json()`; prost length arithmetic matches `encoded_len()`; small events unchanged; oversized `ClientActions` packed into fitting events with order/envelope preserved; oversized `AddMessagesToTask` split per message; a single over-cap message dropped (with descriptive label) while siblings and later actions are kept; indivisible/undecodable/PTY oversized events dropped; placeholders decode to no-ops and keep the envelope; wire replacement keeps `event_no` and skips non-ordered messages.
- `network_tests.rs`: `send_ordered_terminal_event_message` splits an oversized agent event into consecutively numbered events that each fit; an indivisible oversized event consumes no event number; PTY batches over `MAX_PTY_BATCH_RAW_BYTES` are chunked and reassemble byte-for-byte; **reconnect replay** of an oversized unacked event reaches a mock websocket as a same-numbered placeholder followed by the rest of the queue; an oversized non-ordered message is never written to the wire.

Commands run (all passing):
- `cargo nextest run -p warp shared_session` — 231 passed
- `cargo nextest run -p warp shared_session::sharer` — 34 passed
- `./script/format`

- [ ] I have manually tested my changes locally with `./script/run` (not done: reproducing needs a >16 MiB agent event against a real session-sharing-server; the wire path is covered by the mock-socket tests)

## Agent Mode


app/src/terminal/shared_session/sharer/message_size.rs
```diff
@@ -0,0 +1,420 @@
...
+/// Largest websocket message the session sharing server accepts. The client sends each message
+/// as a single frame and the server keeps tungstenite's default 16 MiB `max_frame_size`, so it
+/// drops the connection on anything larger regardless of its `max_message_size`. Because unacked
+/// ordered events are replayed after every reconnect, an oversized one would be resent forever.
+pub(super) const SERVER_MAX_WEBSOCKET_MESSAGE_BYTES: usize = 16 * 1024 * 1024;
+
+/// Soft cap on the serialized size of a single `OrderedTerminalEvent` upload, kept at half of
+/// [`SERVER_MAX_WEBSOCKET_MESSAGE_BYTES`] so the hard limit is never approached. Events over this
+/// cap are split into smaller events or, when indivisible, dropped before they are numbered.
+pub(super) const MAX_ORDERED_TERMINAL_EVENT_BYTES: usize = SERVER_MAX_WEBSOCKET_MESSAGE_BYTES / 2;
```
seemeroland, line 31: Cutting this in half seems too aggressive, I'm worried we're going to start dropping/splitting too many messages.

app/src/terminal/shared_session/sharer/message_size.rs
```diff
@@ -0,0 +1,420 @@
...
+                description: kind,
+                bytes: wire_bytes,
+            }],
+        },
+    };
+    bounded.original_bytes = Some(wire_bytes);
+    bounded
+}
+
+fn split_agent_response_event(
```
seemeroland, line 104: Spreading the client actions across multiple messages I think should be fine in theory, but would want to test it

app/src/terminal/shared_session/sharer/message_size.rs
```diff
@@ -0,0 +1,420 @@
...
+/// Both `ResponseEvent.client_actions` and each `ClientActions.actions` element encode this way.
+fn length_delimited_len(len: usize) -> usize {
+    1 + encoded_len_varint(len as u64) + len
+}
+
+/// A same-numbered stand-in for an event whose payload cannot be delivered. Viewers and the
+/// session sharing server's conversation classifier treat the stand-in as a no-op, so it can
+/// occupy the original event number without disturbing the sequence. `None` for events that carry
+/// no payload and therefore cannot be oversized.
+pub(super) fn placeholder_for(
```
seemeroland, line 259: Why send an empty placeholder message instead of not sending at all?

seemeroland (commented):
I don't think individual tool call results should be this large. It seems like we're not doing something right on the MAA side to truncate.

If we do hit a tool call result this large, I'm not sure this would help, since it's probably a single client action adding one very large result. It's worth fixing the infinite reconnect on a bad oversized message, but I think we probably should just skip it. Viewing is going to be a broken experience but at least the agent can still run.

Ultimately I think I would rather increase the allowed message size if we are seeing MAA responses this big, but looking at the logs to see what a reasonable limit would be. This is how we set the limit for the scrollback in https://github.com/warpdotdev/session-sharing-server/pull/444

PR #16058, opened 2026-09-17
fix(linux): don't fall back to xdg-open for unrecognized default editors

### LINKED ISSUE #16057: Linux: "Default App" / "$EDITOR" file-link editor setting opens the system browser instead of an editor
### Pre-submit Checks

- [ ] I have an issue with AI and have included the debugging ID
- [ ] I have technical issue and have included the logs

### Describe the bug

On Linux, **Settings → Editor and Code Review → "Choose an editor to open file links"** doesn't reliably open the selected editor. With **Default App** selected (and, less directly, with **$EDITOR** selected), clicking a file link in the terminal instead opens the file in the system's default *web browser* (in my case, Zen).

### Root cause (traced in the current `master` source)

`open_file_path_with_line_and_col` in [`app/src/util/file/external_editor/linux.rs`](https://github.com/warpdotdev/warp/blob/master/app/src/util/file/external_editor/linux.rs):

```rust
pub fn open_file_path_with_line_and_col(..., with_editor: Option<Editor>, full_path: &Path, ctx: &mut AppContext) {
    if full_path.is_file() {
        let with_editor = with_editor.or_else(|| get_app_for_file_from_mime(full_path));
        if let Some(editor) = with_editor
            && let Some(mut command) = editor.command(full_path, line_column_number)
        {
            // ... spawn and return
        }
    }
    ctx.open_file_path(full_path); // <-- unconditional fallback
}
```

- For `EditorChoice::SystemDefault` ("Default App"), `open_file_path_in_external_editor` (in `mod.rs`) passes `editor: None`. `get_app_for_file_from_mime` then shells out to `xdg-mime query filetype` + `xdg-mime query default`, but only recognizes a fixed list of ~25 hardcoded IDE app-ids (`Editor::app_ids()` — VSCode, JetBrains family, Sublime, Zed, Windsurf). On my system the mime default for a `.py` file is `nvim.desktop`, which isn't in that list, so `with_editor` stays `None`.
- Because nothing matched, execution falls straight through to `ctx.open_file_path(full_path)`, which on Linux (`crates/warpui/src/windowing/winit/delegate.rs`) is just:
  ```rust
  let _ = command::blocking::Command::new("xdg-open").arg(path).spawn();
  ```
- Bare `xdg-open <path>` on a system whose desktop-environment isn't one `xdg-utils` recognizes (I'm on [Omarchy](https://omarchy.org/), an Arch/Hyprland-based distro; `$XDG_CURRENT_DESKTOP` doesn't match `xdg-open`'s built-in DE branches) falls through to its own last-resort fallback of opening the argument in the default web browser — which is exactly the symptom observed. I confirmed `xdg-settings get default-web-browser` on my machine returns `zen.desktop`, matching what actually opens.
- The same fallthrough is hit for `EditorChoice::EnvEditor` ("$EDITOR") whenever the terminal session's `$EDITOR` isn't captured by Warp (`workspace/view.rs`'s `FileTarget::EnvEditor` arm calls the same `open_file_path_in_external_editor` with `editor: None` in that case).

So the underlying issue is that the final fallback trusts `xdg-open`'s own desktop-environment detection, which isn't reliable on non-mainstream Linux setups, and Warp has no deterministic fallback of its own for mime-default apps outside the hardcoded editor list.

(Separately, and less confidently traced: I also saw the **VSCode** option in that same dropdown fail to open VSCode and fall back to the browser too, even though `code-url-handler.desktop` and `/usr/bin/code` are correctly registered on my system. That path goes through `Editor::command()`'s `VSCode => Command::new("xdg-open").arg("vscode://file...")` branch, so if it really is affected, the same DE-detection issue in `xdg-open` would be the suspect — but I couldn't fully confirm this second symptom without a live repro session, so treat it as a secondary, unconfirmed data point rather than part of the traced bug above.)

### To reproduce

1. On Linux, with a desktop environment `xdg-open`'s DE detection doesn't specifically recognize (e.g. a Hyprland-based distro like Omarchy — `$XDG_CURRENT_DESKTOP` unset or non-standard), set a browser as the system default browser (`xdg-settings set default-web-browser <browser>.desktop`).
2. In Warp, go to Settings → Editor and Code Review → "Choose an editor to open file links" → select **Default App**.
3. In a terminal pane, run a command that prints a clickable file path (e.g. a Python traceback, or just `ls -la some/existing/file.py`).
4. Ctrl/Cmd-click the file path.

### Expected behavior

The file opens in the system's actual mime-type default application (or fails clearly), not in the default web browser.

The `xdg-mime`/`xdg-settings` output I used to trace the root cause is described inline above. I can attach a screen recording if useful, but the code trace above should be reproducible by reading the linked source directly.

### Operating system (OS)

Linux

Omarchy (Arch Linux, Hyprland-based) — kernel `7.2.5-3-omarchy`

### Shell Version

zsh

### Current Warp version

`Oz v0.2026.09.09.08.26.stable_02` (from `warp-terminal --version`)

### Regression?

No, this bug or issue has existed throughout my experience using Warp.

### Additional context

Using Wayland (Hyprland). I have a fix in progress that resolves the mime-type default app directly from its `.desktop` entry (reusing the `EditorMetadata`/`Exec`-tokenizing code already used for the hardcoded editor list) instead of deferring to bare `xdg-open`, only falling back to `ctx.open_file_path` when no `.desktop` entry can be resolved at all. I'll open a PR linked to this issue.

No

xJundo:
## Description

On Linux, `open_file_path_with_line_and_col` (`app/src/util/file/external_editor/linux.rs`) only recognizes a fixed list of ~25 hardcoded editors when resolving a file's XDG mime-type default app. When the mime default isn't one of those (e.g. a plain text editor like gedit, or a TUI editor's `.desktop` entry), it falls straight through to a bare `xdg-open <path>` call.

On desktop environments `xdg-open`'s own DE detection doesn't recognize (e.g. Hyprland-based distros such as Omarchy), that call can silently open the file in the system's default **web browser** instead of any editor — the actual, reproduced symptom filed in #16057. This affects both the "Default App" editor-link setting, and the "$EDITOR" setting whenever Warp doesn't have the session's `$EDITOR` value captured.

This PR resolves the mime-type default app's `.desktop` entry ourselves (reusing the existing `EditorMetadata` / `Exec`-tokenizing code already used for the hardcoded editor list) and spawns it directly, instead of deferring to `xdg-open`'s unreliable fallback chain. `ctx.open_file_path` (and transitively `xdg-open`) is now only reached when no matching `.desktop` entry exists at all, or when the entry requires a controlling terminal we have no way to provide (`Terminal=true`, e.g. a TUI editor like neovim).

## Linked Issue

Fixes #16057

- [ ] The linked issue is labeled `ready-to-spec` or `ready-to-implement`. *(Not yet triaged — opening this PR per the "PRs opened without a linked issue" guidance in CONTRIBUTING.md, since the fix was ready once the root cause was traced. Happy to wait for triage if preferred.)*
- [x] Where appropriate, screenshots or a short video of the implementation are included below. *(N/A — this is a headless fallback-resolution fix with no UI surface; see Testing section for why no manual GUI test is included.)*

### Automated tests

Added unit tests in `app/src/util/file/external_editor/linux_tests.rs`:
- `test_terminal_true_is_parsed` / `test_terminal_defaults_to_false`: coverage for the new `EditorMetadata::terminal` field parsed from the desktop entry's `Terminal` key.
- `test_find_desktop_file_by_app_id_matches_exact_stem`, `test_find_desktop_file_by_app_id_no_match_returns_none`, `test_find_desktop_file_by_app_id_searches_all_given_dirs`: coverage for the new `find_desktop_file_by_app_id_in` lookup, using the existing `VirtualFS` sandbox so the matching logic is exercised without touching the real, machine-dependent XDG search paths.

I did **not** add a test around `default_app_id_for_file`/`open_with_default_desktop_entry` end-to-end, since (consistent with the pre-existing `get_app_for_file_from_mime`/`compute_installed_editors`, which are similarly untested) they shell out to the real `xdg-mime` binary and scan the real XDG data dirs — there's no mocking seam for that already in this codebase, and I didn't want to introduce one just for this fix's scope.

Ran locally on the pinned toolchain (`1.92.0`):
- `cargo check -p warp --lib` — clean.
- `cargo nextest run -p warp -E 'test(external_editor)'` — 61/61 passed (including the 5 new tests above).
- `cargo clippy -p warp --all-targets --tests -- -D warnings` — clean.
- `./script/format` — applied (cosmetic reflow only, no logic change).

(Also caught and fixed a real compile error against the pinned `freedesktop-desktop-entry = 0.5.0`: its `Iter::new`/`default_paths()` take/return `Vec<PathBuf>`, not a generic `impl Iterator` as newer versions' docs show — my initial draft had assumed the latter from `docs.rs`, which was showing a newer version than the one this repo pins.)

### Manual testing

**Not performed end-to-end in the real GUI app.** I don't have a full native build of Warp running yet (no display / desktop environment in the sandbox I did this work from), so I haven't clicked through Settings → "Default App" against a real mime-type default outside `SUPPORTED_EDITORS` (e.g. a `.txt` file defaulting to gedit) to visually confirm it now opens the right app instead of the browser. The crate itself now builds and its unit/lint checks pass locally. If a maintainer or reviewer can do a quick manual pass, that would close this gap — happy to also try again if I get a working desktop build.

## Agent Mode

CHANGELOG-BUG-FIX: Fixed file links on Linux ("Default App" / "$EDITOR" editor setting) silently opening the system's default web browser instead of the configured editor on desktop environments `xdg-open` doesn't recognize.

---


xJundo:
@cla-bot check

Issue #15977, opened 2026-09-13
Rich Input: notify the CLI agent when rich input opens and closes, so it can hide its own prompt bar


aksheyd:
### Pre-submit Checks

- [x] I have [searched Warp feature requests](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3Aenhancement) and there are no duplicates
- [x] I have [searched Warp docs](https://docs.warp.dev) and my feature is not there

### Describe the solution you'd like?

# tldr

Add one small channel from Warp to the CLI agent: a per-pane socket, advertised by an environment variable, that carries one JSON line when rich input opens and one when it closes. The agent uses it to hide and show its own prompt bar. Nothing else changes.

Then, when Warp opens rich input, the CLI agent can close it's native prompt bar to make the experience more seamless. 

# proposal

1. Socket. Warp creates one Unix domain socket per pane when it creates the PTY. Warp listens on it. On Windows, Warp uses a named pipe with the same message format.
2. Environment variable. Warp sets `WARP_CLI_AGENT_CONTROL_SOCKET=<path>` in the same place it sets `WARP_CLI_AGENT_PROTOCOL_VERSION`. The name is a proposal.
3. Direction. Warp writes. The agent reads. The agent does not write in version 1.
4. Format. UTF-8. One JSON object per line. Each line ends with `\n`. Every object has an integer `v` and a string `event`.
5. Events.
   - On open: `{"v":1,"event":"rich_input","active":true}`
   - On close: `{"v":1,"event":"rich_input","active":false}`
6. First line on connect. When a client connects, Warp writes the current state first. This covers an agent that starts while rich input is already open.
7. Multiple clients. Warp accepts any number of clients and writes the same lines to each. If a write fails, Warp drops that client and continues.
8. Lifetime. When the pane closes, Warp closes the socket. The agent reads end-of-file.
9. Forward compatibility. The agent ignores unknown `event` values and unknown fields.
   Warp bumps `v` only for a breaking change.

### Is your feature request related to a problem? Please describe.

When Warp opens rich input for a CLI agent, the agent's own prompt bar stays on screen. The user sees two text boxes. The agent cannot fix this alone, because Warp sends it no message when rich input opens or closes.

### Additional context

# X post

https://x.com/aksheyd/status/2098448416447885818?s=20

See my conversation with Sathvik and SKcd in the comments.

# cross-platform

- macOS and Linux: Unix domain socket. Keep the path short (104-byte limit on macOS).
[...]

PR #16068, opened 2026-09-17
Notify CLI agents when rich input opens and closes via a per-pane control endpoint

aksheyd:
## Description

Adds a Warp-to-agent control channel so a CLI agent running in a pane can hide its own prompt bar while Warp's rich input composer is open, and show it again when the composer closes.

Warp already tracks rich input state in `CLIAgentSessionsModel` (`open_input` / `close_input` emit `InputSessionChanged`), but nothing tells the child process. This change:

- Binds one local endpoint per pane before the PTY spawns (Unix domain socket on macOS/Linux, named pipe on Windows) and advertises it via `WARP_CLI_AGENT_CONTROL_SOCKET` (single constant in `warp_core::cli_agent_protocol`).
- Writes UTF-8 JSON Lines: `{"v":1,"event":"rich_input","active":true|false}`. A client receives the current state immediately on connect, then exactly one line per closed→open and open→closed transition. `Ended` publishes `active:false`.
- Publishes from `CLIAgentSessionsModelEvent` (the model is the source of truth), not from individual UI handlers, so manual close, submit close, auto-toggle, session replacement and pane teardown are all covered.
- Accepts many clients (cap 16 per pane), drops clients whose writes fail or stall, never blocks the UI thread, and closes the listener and all clients when the pane is dropped (clients observe EOF).
- Fails open: if the endpoint cannot be bound, Warp logs at `warn`, does not set the env var, and the terminal starts normally.

Decisions on the open questions from the spec:

| Question | Choice |
| --- | --- |
| Gate | `HOANotifications && CLIAgentRichInput` (rich input is the only thing that can change the state) |
| Unix socket dir | `$XDG_RUNTIME_DIR/warp/` (0700) when set, else `std::env::temp_dir()`; socket is 0600. Not `state_dir()`: it is durable and would accumulate one socket per crashed pane |
| Windows | `\\.\pipe\Warp<Channel>_cli_agent_control_<id>` via tokio named pipes; default DACL with `reject_remote_clients`, so the 16-client cap is not a same-user guarantee there (channel carries one boolean) |
| WSL / Docker sandbox / SSH | No endpoint and no env var (`ShellStarter::Wsl` and `DockerSandbox` skipped; not added to `WSLENV`) |
| Env var name | `WARP_CLI_AGENT_CONTROL_SOCKET` as proposed; one constant if a transport-neutral name is preferred |

Not reused: `crates/ipc` (request/response protocol; this is a one-way broadcast) and `local_control::discovery_dir()` (its pruner would treat these sockets as abandoned artifacts).

Companion: Grok Build CLI will implement the agent side (hide/show its prompt bar on these lines) once the contract is confirmed.

## Linked Issue

Implements #15977 following the product/tech spec in #15994 (`specs/GH15977/`). Opened at the Warp team's request while the spec is under review; happy to fold in spec changes here.

- [x] The linked issue is labeled `ready-to-spec` or `ready-to-implement`
- [ ] Screenshots or a short video are included below (pending, see Testing)

## Testing

Unit tests in `app/src/terminal/cli_agent_sessions/control_endpoint_tests.rs`:

- Wire line is byte-exact `{"v":1,"event":"rich_input","active":true}\n` / `…false}\n`.
- Publisher sends one line per transition; repeated open or repeated close sends nothing.
- `observe` maps `InputSessionChanged` Open/Closed and `Ended` to the right state.
- New client receives the current state first (both while closed and while open).
- Two clients both receive a broadcast; dropping one does not affect the other; dropping the endpoint yields EOF.
- Unix: bound socket has mode 0600; macOS-only: address length stays under the 104-byte socket path limit.

Not yet run (this branch was authored on a headless Linux box without a Warp GUI build): `./script/presubmit`, `cargo nextest run`, and the manual validation below. Those follow on a Mac before this leaves draft.

Manual validation plan: open a pane, `echo $WARP_CLI_AGENT_CONTROL_SOCKET`, connect with `nc -U "$WARP_CLI_AGENT_CONTROL_SOCKET"`, confirm `active:false` on connect, `active:true` on opening rich input, `active:false` on close/submit, and EOF on closing the pane; repeat with two clients; confirm the var is absent in a WSL pane.

### Screenshots / Videos

Pending Mac build.

## Agent Mode

## Followups

- Test for "env var absent when bind fails" needs a `ModelContext` harness; currently enforced by construction and covered by the manual plan.
- `tokio::sync::watch` reframe of the publisher was considered and rejected: it coalesces, and the contract promises one line per transition.
- Transport-neutral env var name if maintainers prefer one for Windows.
- Forwarding for warpified SSH sessions and a WSL bridge, per the spec's follow-ups.

CHANGELOG-IMPROVEMENT: CLI agents running in Warp can now learn when Warp's rich input opens and closes via a per-pane control endpoint, so they can hide their own prompt while Warp's composer is active.
CHANGELOG-TUI: Warp now advertises a per-pane control endpoint (`WARP_CLI_AGENT_CONTROL_SOCKET`) that tells CLI agents when rich input opens and closes.


aksheyd:
@cla-bot check

PR #16072, opened 2026-09-18
warp-issues-6004:Two windows are openend when starting in Windows  #6004

### LINKED ISSUE #6004: Two windows are openend when starting in Windows
### Pre-submit Checks

- [x] I have [searched Warp bugs](https://github.com/warpdotdev/warp/issues?q=is%3Aissue+label%3ABugs) and there are no duplicates
- [ ] I have included the logs (optional, but helps expedite the bug fix). [Log gathering intructions](https://docs.warp.dev/help/sending-us-feedback#gathering-warp-logs)

### Describe the bug

When Warp is closed and I open it from the taskbar, two windows open. One with my home directory and the second window with `~\AppData\Local\Programs\Warp`

### To reproduce

1. Warp is closed
2. Open Warp

### Expected behavior

Only one window with my home directory should be open.

Image

Image

### Operating system

Windows

Windows 11 Enterprise 23H2

### Shell Version

PowerShell 7.5.0

### Current Warp version

v0.2025.02.24.20.50.stable_04

### Regression

### Recent working Warp date

_No response_

### Additional context

_No response_

No

None

chuai-Git:
## Root cause analysis

I dug into this against the current source and found the mechanism. The short version: **the crash-recovery watcher process that Warp spawns on every Windows launch is not exempted from the single-instance startup-args forwarding, so it forwards a `new_window` URL back to the instance that spawned it.**

### The chain

1. On Windows, every GUI launch spawns a crash-recovery watcher child of `warp.exe` itself (`enable_crash_recovery` is on for Windows/Linux; with no `WGPU_BACKEND` override, `choose_crash_recovery_mechanism()` returns `DisableOpenGL`). See `app/src/crash_recovery.rs` — `spawn_recovery_process()` relaunches the current exe with `--parent-handle=... --crash-recovery-mechanism=...`, and the child just waits for the parent to die.

2. The watcher has no subcommand, so it parses as `LaunchMode::App` and runs the **full GUI startup path** — including `app_services::windows::pass_startup_args_to_existing_instance()` (`app/src/lib.rs`, runs *before* `CrashRecovery::new`). There is no `is_crash_recovery_process()` exemption there.

3. By the time the child runs that check, the parent already holds the `Local\Warp{channel}_SingleInstance` mutex (it created it during its own earlier check, before spawning the child). So the child concludes *"another instance is running, I'm a second launch"*.

4. Since the child has no URLs on its command line, the forwarding code fabricates one: `warp://action/new_window?path=<cwd>` (`app/src/app_services/windows/mod.rs`). The `cwd` here is inherited from the parent — i.e. the shortcut's *Start in* directory. **That's why the duplicate window opens in `~\AppData\Local\Programs\Warp`** — it's the install dir from the shortcut, which matches what's reported in this issue.

5. `Action::NewWindow` passes `window_id: None`, so `open_file()` unconditionally dispatches `root_view:open_new_from_path` — a brand-new window, never a tab in an existing one.

### Why it's timing-dependent (and why the first launch is fine)

The parent spawns the watcher *before* its URI named-pipe server exists (the pipe is created much later, in `app_services::init` inside the run closure). So it's a race:

- **Child connects before the pipe exists** → connect fails → the child falls into the "unknown forwarding error" branch and continues as the watcher. No extra window. (Most machines land here, which is why this isn't universal.)
- **Child connects after the pipe is up** → forwarding succeeds → the parent opens a duplicate window, and the child exits.

On a cold first launch the parent has extra work before the pipe comes up (creating the SQLite DB, running migrations, first-run setup), so the child normally loses the race — one window. On subsequent launches everything is warm, the parent reaches pipe creation sooner, and on a given machine the race resolves the same way every time — hence "always two windows".

### Why disabling session restore doesn't help

`save_app` snapshots **all** live windows on window move/focus/resize/close callbacks, so once a duplicate exists it gets persisted and restored. But the forwarded `new_window` URL opens a window regardless of the restore setting, which matches the observation above that disabling *"Restore windows, tabs, and panes on startup"* doesn't make the duplicate go away.

## Proposed fix

Exempt the crash-recovery watcher from startup-args forwarding (mirrors the existing `finish_update` exemption):

```diff
 // app/src/app_services/windows/mod.rs
 #[derive(Error, Debug)]
 #[cfg(feature = "release_bundle")]
 pub enum StartupArgsForwardingError {
     #[error("should not forward arguments after an auto-update")]
     IgnoredAfterAutoUpdate,
+    #[error("should not forward arguments from the crash recovery process")]
+    IgnoredForCrashRecoveryProcess,
     #[error("there is no other instance of Warp")]
     NoExistingInstance,
     #[error("failed to construct url")]
```

```diff
 // app/src/app_services/windows/mod.rs
 pub fn pass_startup_args_to_existing_instance(
     args: &warp_cli::AppArgs,
 ) -> Result<(), StartupArgsForwardingError> {
     if args.finish_update {
         return Err(StartupArgsForwardingError::IgnoredAfterAutoUpdate);
     }
+    // The crash recovery watcher process is spawned by the main instance on
+    // every launch. It must not forward a new-window URL back to its parent,
+    // which would make the parent open a duplicate window.
+    if crate::crash_recovery::is_crash_recovery_process(args) {
+        return Err(StartupArgsForwardingError::IgnoredForCrashRecoveryProcess);
+    }
     if SingleInstanceManager::is_sole_running_instance()? {
         return Err(StartupArgsForwardingError::NoExistingInstance);
     }
```

```diff
 // app/src/lib.rs (Windows forwarding match)
             // If we just finished an auto-update, we should continue running.
             Err(app_services::windows::StartupArgsForwardingError::IgnoredAfterAutoUpdate) => {}
+            // The crash recovery watcher process should continue running so it
+            // can wait on its parent; it must not forward args to it.
+            Err(app_services::windows::StartupArgsForwardingError::IgnoredForCrashRecoveryProcess) => {}
```

The Linux/D-Bus forwarding path (`app/src/app_services/linux/mod.rs`) has the identical hole — it also fabricates a `new_window` URL when no URLs are passed — so it needs the same exemption (wrapped in `cfg_if!` there, since FreeBSD compiles that module without `enable_crash_recovery`).

With this patch built from source (stable channel, `release_bundle`), the duplicate window no longer appears on a machine that reproduced it on every launch (Windows 10 22H2, v0.2026.09.09.08.26.stable_02 codebase).
Fixes #6004


chuai-Git:
I have completed the authentication. Why does it still show failing checks for verification/cla-signed?
Thank you for signing the Warp CLA.

Signed as: @chuai-Git
GitHub user id: 4*****9
Signed at (UTC): 2026-09-18T03:29:10.784Z
Your signature is tied to your GitHub numeric user id and is not invalidated by future username changes.

chuai-Git:
@cla-bot check

